PR #6, opened 2022-04-12
Bug 578871 - prevent UI freeze during UNDO operation + autobuild

jukzi:
While JDT's refactoring correctly tries to interrupt autobuild the UNDO
of the same operation did not. In general all UNDO Operations did not
try to interrupt autobuild.

This adds dependency from ui to resources to make the History operations
become aware of resource scheduling rule.

from https://git.eclipse.org/r/c/platform/eclipse.platform.ui/+/192458

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
@@ -21,15 +21,20 @@ import org.eclipse.core.commands.operations.IOperationHistoryListener;
 import org.eclipse.core.commands.operations.IUndoContext;
 import org.eclipse.core.commands.operations.IUndoableOperation;
 import org.eclipse.core.commands.operations.OperationHistoryEvent;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.jobs.IJobManager;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.LegacyActionTools;
 import org.eclipse.jface.operation.IRunnableWithProgress;
 import org.eclipse.osgi.util.NLS;
+import org.eclipse.swt.custom.BusyIndicator;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IPartListener;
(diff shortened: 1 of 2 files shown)


mickaelistria:
As mentioned on Gerrit, I'm not really convinced that this a proper fix to the right problem. The actual problem is trying to begin a rule in the UI Thread should show some progress dialog. That would cover all cases of that kind without need for the operation to care about the progress report, and instead to focus on actually doing its job.
So I would rather not merge it. I will have more time for development in May and think I can implement a more general solution then.

ghost:
How about submit this and **just revert** if it makes any trouble or when you found a better solution?
(I have doubts you will find a general solution in time)

mickaelistria:
OK, let's do that.

mickaelistria:
I opened #8 about a general solution.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
```diff
@@ -298,7 +303,27 @@ public final void run() {
...
+			} else {
+				// prevent UI freeze during AutoBuild as in 
+				// org.eclipse.jdt.internal.ui.refactoring.RefactoringExecutionHelper#perform
+				final IJobManager manager = Job.getJobManager();
+				// pessimistic scheduling rule:
+				final ISchedulingRule rule = ResourcesPlugin.getWorkspace().getRoot(); 
+				try {
+					try {
+						// interrupt autobuild or show Wait/Cancel Dialog:
+						Runnable r = () -> manager.beginRule(rule, null);
```
iloveeclipse, line 317: This change causes a regression. If the Undo task joins on a workspace rule, it hangs now.
So starting a rule here is not OK.
I would propose to revert the change.

jukzi: > If the Undo task joins on a workspace rule

how to reproduce that?

iloveeclipse: Have custom `IUndoableOperation` (we have one that extends `AbstractOperation`) that runs operation in a different thread and there takes the workspace rule.

```
    @Override
    public IStatus undo(IProgressMonitor monitor, IAdaptable info) throws ExecutionException {

        IProgressService progressService = PlatformUI.getWorkbench().getProgressService();
        try {
            progressService.run(true, false, new IRunnableWithProgress() {
                @Override
                public void run(IProgressMonitor monitor) throws InvocationTargetException, InterruptedException {
                    ISchedulingRule rule = ResourcesPlugin.getWorkspace().getRoot();
                   Job.getJobManager().beginRule(rule, monitor); // here we hang now
                   // ... do something that needs workspace
                }
            });
        } catch (InvocationTargetException | InterruptedException e) {
        }

        return Status.OK_STATUS;
    }
```

iloveeclipse: Basically, locking **anything accessible to 3rd party code** before starting undo is not OK, because "undo" in 3rd party code can require the same lock.

Also interesting (I haven't noticed that first), that the org.eclipse.ui.workbench bundle in question had no references to org.eclipse.core.resources bundle before, I think for a good reason.

PR #12, opened 2022-04-17
Bug 531785 - Auto infix search in Open Resource

pbodnar:
## Basic feature description

A new match rule `RULE_SUBSTRING_MATCH` is introduced in the
`SearchPattern`. The constant itself is copied as-is from JDT's variant
of `SearchPattern`, yet its usage and purpose differs.

When used and prefix search is not enforced, then:

* search is done as if "*" was specified at the pattern start
* CamelCase search doesn't require the 1st pattern char to be the very
1st char of the matched string

Use this in Resource search dialogs for the default and file name
pattern, while also list "prefix matches" firstly and extend the matched
characters highlighting to still work correctly.
Path (container) and extension searches should not be affected by this.

Further changes:

* users can start a search pattern with ">" to enforce the old
prefix-only search

For seeing how autoInfixSearch differs from the old one, one can compare "SearchPatternAuto.java" with "InfixSearchPatternAuto.java".

## Yet to be investigated / implemented

* [x] present the SearchPattern's autoInfixSearch feature on UI layer: change the Open Resource dialog label
* [x] API: introduce `SearchPattern.RULE_SUBSTRING_MATCH` or similar instead of `SearchPattern.autoInfixSearch`
* [x] update corresponding user documentation (see eclipse-platform/eclipse.platform.common#89)

## Historical issues identified that could / should be solved later on

1. `SearchPattern.RULE_CASE_SENSITIVE` - this flag seems to be ignored by the actual SearchPattern implementation
2. `SearchPattern.camelCaseMatch()` could be optimized to check strings' lengths firstly. Something like:

		int patternLength = patternEnd - patternStart;
		int nameLength = nameEnd - nameStart;
		if (nameLength < patternLength) {
			return false;
		}

3. Constants like `SearchPattern.END_SYMBOL` are just private and special chars checking is hard-coded in dependent classes like `FilteredResourcesSelectionDialog`
4. Inefficient matches highlighting: `ResourceItemLabelProvider.getMatchPositions()` and related seem to contain some tricky code and logic. -> see #184

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
@@ -102,6 +102,19 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 	private static final String SHOW_DERIVED = "ShowDerived"; //$NON-NLS-1$
 	private static final String FILTER_BY_LOCATION = "FilterByLocation"; //$NON-NLS-1$
 
+	private static final char START_SYMBOL = '>';
+
+	private static final char END_SYMBOL = '<';
+
+	private static final char BLANK = ' ';
+
+	// this is hard-coded, as a UI option is most probably not necessary
+	private final boolean autoInfixSearch = true;
+
+	private int getDefaultMatchRules() {
+		return SearchPattern.DEFAULT_MATCH_RULES | (autoInfixSearch ? SearchPattern.RULE_SUBSTRING_MATCH : 0);
+	}
+
 	private ShowDerivedResourcesAction showDerivedResourcesAction;
 
 	private ResourceItemLabelProvider resourceItemLabelProvider;
@@ -443,6 +456,7 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 
 			if (pattern != null) {
 				int patternDot = pattern.lastIndexOf('.');
+				// Prioritize names matching the whole pattern
 				String patternNoExtension = patternDot == -1 ? pattern : pattern.substring(0, patternDot);
 				boolean m1 = patternNoExtension.equals(n1);
 				boolean m2 = patternNoExtension.equals(n2);
@@ -452,6 +466,21 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 					}
 					if (m2) {
 						return 1;
+					}
+				}
+				// Prioritize names starting with the pattern
+				char patternFirstChar = getFirstFileNameChar(pattern);
+				if (patternFirstChar != 0) {
+					patternFirstChar = Character.toLowerCase(patternFirstChar);
+					m1 = patternFirstChar == Character.toLowerCase(s1.charAt(0));
+					m2 = patternFirstChar == Character.toLowerCase(s2.charAt(0));
+					if (!m1 || !m2) {
+						if (m1) {
+							return -1;
+						}
+						if (m2) {
+							return 1;
+						}
 					}
 				}
 			}
@@ -493,6 +522,22 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 
 			return comparability;
 		};
+	}
+
+	/**
+	 * @param pattern
+	 * @return the first character from the given string which <em>could</em> be
+	 *         considered a part of a file name. Returns <code>0</code> if there is
+	 *         no such character found.
+	 */
+	private char getFirstFileNameChar(String pattern) {
+		for (int i = 0; i < pattern.length(); i++) {
+			char ch = pattern.charAt(i);
+			if (ch != '*') {
+				return ch;
+			}
+		}
+		return 0;
 	}
 
 	/**
@@ -722,6 +767,9 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 			}
 
 			// Pre-process the matching pattern
+			if (matching.charAt(0) == '>') {
+				matching = matching.substring(1);
+			}
 			char lastChar = matching.charAt(matching.length() - 1);
 			if (lastChar == ' ' || lastChar == '<') {
 				matching = matching.substring(0, matching.length() - 1);
@@ -756,8 +804,11 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 			if (regionsDontMatch) {
 				// We should get here only when CamelCase nor wildcard matching succeeded
 				// A simple comparison of the whole strings should succeed instead
-				if (string.toLowerCase().startsWith(matching.toLowerCase())) {
-					positions.add(new Position(0, matching.length()));
+				int matchingIndex = autoInfixSearch
+						? string.toLowerCase().indexOf(matching.toLowerCase())
+						: (string.toLowerCase().startsWith(matching.toLowerCase()) ? 0 : -1);
+				if (matchingIndex > -1) {
+					positions.add(new Position(matchingIndex, matching.length()));
 				}
 			}
 			return positions;
@@ -959,7 +1010,7 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 		 * @param typeMask    filter type mask. See {@link IResource#getType()} types.
 		 */
 		public ResourceFilter(IContainer container, boolean showDerived, int typeMask) {
-			super();
+			super(new SearchPattern(getDefaultMatchRules()));
 			this.filterContainer = container;
 			this.showDerived = showDerived;
 			this.filterTypeMask = typeMask;
@@ -977,21 +1028,22 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 		private ResourceFilter(IContainer container, IContainer searchContainer, boolean showDerived, int typeMask) {
 			this(container, showDerived, typeMask);
 
-			String stringPattern = getPattern();
-			int matchRule = getMatchRule();
+			final String stringPattern = patternMatcher.getInitialPattern();
 			String filenamePattern;
 
 			int sep = stringPattern.lastIndexOf(IPath.SEPARATOR);
 			if (sep != -1) {
+				// This means that we primarily check (via `patternMatcher`) just the resource
+				// _name_ part and when there is some actual _container_ part (`sep > 0`), we
+				// also do checks for that part (via `containerPattern` and optional
+				// `relativeContainerPattern`).
 				filenamePattern = stringPattern.substring(sep + 1, stringPattern.length());
-				if ("*".equals(filenamePattern)) //$NON-NLS-1$
-					filenamePattern = "**"; //$NON-NLS-1$
 
 				if (sep > 0) {
 					if (filenamePattern.isEmpty()) // relative patterns don't need a file name
 						filenamePattern = "**"; //$NON-NLS-1$
 
-					String containerPattern = stringPattern.substring(0, sep);
+					String containerPattern = stringPattern.substring(isMatchPrefix(stringPattern) ? 1 : 0, sep);
 
 					if (searchContainer != null) {
 						relativeContainerPattern = new SearchPattern(
@@ -1001,6 +1053,8 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 					}
 
 					if (!containerPattern.startsWith(Character.toString('*'))) {
+						// bug 552418 - make the search always "root less", so that users don't need to
+						// type the initial "*/"
 						if (!containerPattern.startsWith(Character.toString(IPath.SEPARATOR))) {
 							containerPattern = IPath.SEPARATOR + containerPattern;
 						}
@@ -1010,35 +1064,29 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 							| SearchPattern.RULE_PREFIX_MATCH | SearchPattern.RULE_PATTERN_MATCH);
 					this.containerPattern.setPattern(containerPattern);
 				}
-				boolean isPrefixPattern = matchRule == SearchPattern.RULE_PREFIX_MATCH
-						|| (matchRule == SearchPattern.RULE_PATTERN_MATCH && filenamePattern.endsWith("*")); //$NON-NLS-1$
-				if (!isPrefixPattern)
-					// Add '<' again as it was removed by SearchPattern
-					filenamePattern += '<';
-				else if (filenamePattern.endsWith("*") && !filenamePattern.equals("**")) //$NON-NLS-1$ //$NON-NLS-2$
-					// Remove added '*' as the filename pattern might be a camel case pattern
-					filenamePattern = filenamePattern.substring(0, filenamePattern.length() - 1);
+				if (isMatchPrefix(stringPattern)) {
+					filenamePattern = '>' + filenamePattern;
+				}
 				patternMatcher.setPattern(filenamePattern);
-				// Update filenamePattern and matchRule as they might have changed
-				filenamePattern = getPattern();
-				matchRule = getMatchRule();
 			} else {
 				filenamePattern = stringPattern;
 			}
 
 			int lastPatternDot = filenamePattern.lastIndexOf('.');
 			if (lastPatternDot != -1) {
-				if (matchRule != SearchPattern.RULE_EXACT_MATCH) {
-					namePattern = new SearchPattern();
-					namePattern.setPattern(filenamePattern.substring(0, lastPatternDot));
-					String extensionPatternStr = filenamePattern.substring(lastPatternDot + 1);
-					// Add a '<' except this is a camel case pattern or a prefix pattern
-					if (matchRule != SearchPattern.RULE_CAMELCASE_MATCH && matchRule != SearchPattern.RULE_PREFIX_MATCH
-							&& !extensionPatternStr.endsWith("*")) //$NON-NLS-1$
-						extensionPatternStr += '<';
-					extensionPattern = new SearchPattern();
-					extensionPattern.setPattern(extensionPatternStr);
+				// This means we primarily check resource name as _name_ and _extension_ part
+				// and only when we don't succeed, we try the default whole-name check (via
+				// `patternMatcher`).
+				namePattern = new SearchPattern(getDefaultMatchRules());
+				String namePatternStr = filenamePattern.substring(0, lastPatternDot);
+				if (isMatchSuffix(stringPattern) && !namePatternStr.endsWith("*")) { //$NON-NLS-1$
+					// This means extension part will end with '<' (or ' ')
+					// and we should apply the same for the name part.
+					namePatternStr += "<"; //$NON-NLS-1$
 				}
+				namePattern.setPattern(namePatternStr);
+				extensionPattern = new SearchPattern();
+				extensionPattern.setPattern(filenamePattern.substring(lastPatternDot + 1));
 			}
 
 		}
@@ -1216,6 +1264,31 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 			resourceFactory.saveState(element);
 		}
 
+	}
+
+	/**
+	 * Returns whether prefix matching is enforced in the given search pattern.
+	 */
+	private static boolean isMatchPrefix(String pattern) {
+		if (pattern.length() == 0) {
+			return false;
+		}
+
+		char first = pattern.charAt(0);
+		return pattern.length() > 1 && first == START_SYMBOL;
+	}
+
+	/**
+	 * Returns whether suffix matching is enforced in the given search pattern.
+	 */
+	private static boolean isMatchSuffix(String pattern) {
+		if (pattern.length() <= 1) {
+			return false;
+		}
+
+		char last = pattern.charAt(pattern.length() - 1);
+		boolean matchPrefix = isMatchPrefix(pattern);
+		return pattern.length() > (matchPrefix ? 2 : 1) && (last == END_SYMBOL || last == BLANK);
 	}
 
 }
--- a/bundles/org.eclipse.ui.workbench/.settings/.api_filters
+++ b/bundles/org.eclipse.ui.workbench/.settings/.api_filters
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<component id="org.eclipse.ui.workbench" version="2">
+    <resource path="Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java" type="org.eclipse.ui.dialogs.SearchPattern">
+        <filter id="336658481">
+            <message_arguments>
+                <message_argument value="org.eclipse.ui.dialogs.SearchPattern"/>
+                <message_argument value="DEFAULT_MATCH_RULES"/>
+            </message_arguments>
+        </filter>
+        <filter id="336658481">
+            <message_arguments>
+                <message_argument value="org.eclipse.ui.dialogs.SearchPattern"/>
+                <message_argument value="RULE_SUBSTRING_MATCH"/>
+            </message_arguments>
+        </filter>
+    </resource>
+</component>
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
@@ -46,6 +46,9 @@ public class SearchPattern {
 	 * <p>
 	 * Unless the pattern ends with ' ' or '>', search is performed as if '*' was
 	 * specified at the end of the pattern.
+	 * <p>
+	 * When {@link #RULE_SUBSTRING_MATCH} is in effect, search is performed as if
+	 * '*' was specified at the start of the pattern.
 	 */
 	public static final int RULE_PATTERN_MATCH = 0x0002;
 
@@ -84,8 +87,32 @@ public class SearchPattern {
 	 * <code>"NPE"</code> string pattern, search will only use Camel Case match
 	 * rule, but with <code>N*P*E*</code> string pattern, it will use only Pattern
 	 * match rule.
+	 * <p>
+	 * Unless {@link #RULE_SUBSTRING_MATCH} is in effect, it is required that
+	 * the 1st pattern's char must match the very 1st char of the search result.
 	 */
 	public static final int RULE_CAMELCASE_MATCH = 0x0080;
+
+	/**
+	 * Match rule: The search pattern is a string placed anywhere in the search
+	 * result. When in effect, this flag also affects some of the other match rules
+	 * (see their documentation for details).
+	 * <p>
+	 * Prefix search may still be enforced by placing '>' at the beginning of the
+	 * pattern. Analogically, suffix search may be enforced by placing ' ' or '&lt;'
+	 * at the end of the pattern.
+	 *
+	 * @since 3.128
+	 */
+	public static final int RULE_SUBSTRING_MATCH = 0x0200;
+
+	/**
+	 * The default set of match rules as used by the no-argument constructor.
+	 *
+	 * @since 3.128
+	 */
+	public static final int DEFAULT_MATCH_RULES = RULE_EXACT_MATCH | RULE_PREFIX_MATCH | RULE_PATTERN_MATCH
+			| RULE_CAMELCASE_MATCH | RULE_BLANK_MATCH;
 
 	private int matchRule;
 
@@ -95,6 +122,8 @@ public class SearchPattern {
 
 	private TextMatcher stringMatcher;
 
+	private static final char START_SYMBOL = '>';
+
 	private static final char END_SYMBOL = '<';
 
 	private static final char ANY_STRING = '*';
@@ -103,26 +132,33 @@ public class SearchPattern {
 
 	private int allowedRules;
 
+	private boolean substringSearch;
+
+	private boolean matchPrefix;
+
+	private boolean matchSuffix;
+
 	/**
-	 * Creates a new instance of SearchPattern with the following match rules
-	 * configured: {@link #RULE_EXACT_MATCH} | {@link #RULE_PREFIX_MATCH} |
-	 * {@link #RULE_PATTERN_MATCH} | {@link #RULE_CAMELCASE_MATCH} |
-	 * {@link #RULE_BLANK_MATCH}.
+	 * Creates a new instance of SearchPattern with {@link #DEFAULT_MATCH_RULES
+	 * default set of rules} configured.
 	 */
 	public SearchPattern() {
-		this(RULE_EXACT_MATCH | RULE_PREFIX_MATCH | RULE_PATTERN_MATCH | RULE_CAMELCASE_MATCH | RULE_BLANK_MATCH);
+		this(DEFAULT_MATCH_RULES);
 	}
 
 	/**
 	 * Creates a search pattern with a rule or rules to apply for matching index keys.
 	 *
 	 * @param allowedRules one of {@link #RULE_EXACT_MATCH},
-	 *                     {@link #RULE_PREFIX_MATCH}, {@link #RULE_PATTERN_MATCH},
+	 *                     {@link #RULE_PREFIX_MATCH},
+	 *                     {@link #RULE_SUBSTRING_MATCH},
+	 *                     {@link #RULE_PATTERN_MATCH},
 	 *                     {@link #RULE_CAMELCASE_MATCH},
 	 *                     {@link #RULE_CASE_SENSITIVE}, or their combination in
-	 *                     order to enable more types of matching. Note that
-	 *                     {@link #RULE_CASE_SENSITIVE} is special in that it
-	 *                     just affects how the other match rules
+	 *                     order to enable more types of matching. Note that rules
+	 *                     {@link #RULE_CASE_SENSITIVE} and
+	 *                     {@link #RULE_SUBSTRING_MATCH} are special in that they
+	 *                     generally just affect how the other match rules
 	 *                     behave.<br>
 	 *                     Examples: {@link #RULE_EXACT_MATCH} |
 	 *                     {@link #RULE_CASE_SENSITIVE} if an exact and case
@@ -133,6 +169,7 @@ public class SearchPattern {
 	 */
 	public SearchPattern(int allowedRules) {
 		this.allowedRules = allowedRules;
+		this.substringSearch = (allowedRules & RULE_SUBSTRING_MATCH) != 0;
 	}
 
 	/**
@@ -142,6 +179,16 @@ public class SearchPattern {
 	 */
 	public String getPattern() {
 		return this.stringPattern;
+	}
+
+	/**
+	 * Gets the initial (input) string pattern.
+	 *
+	 * @return pattern
+	 * @since 3.128
+	 */
+	public String getInitialPattern() {
+		return this.initialPattern;
 	}
 
 	/**
@@ -181,50 +228,65 @@ public class SearchPattern {
 			}
 			//$FALL-THROUGH$
 		default:
-			return startsWithIgnoreCase(text, stringPattern);
+			// apply RULE_PREFIX_MATCH / RULE_SUBSTRING_MATCH
+			boolean doMatchPrefix = matchPrefix || !substringSearch;
+			if (doMatchPrefix && !matchSuffix) {
+				return startsWithIgnoreCase(text, stringPattern);
+			}
+			if (!doMatchPrefix && matchSuffix) {
+				return endsWithIgnoreCase(text, stringPattern);
+			}
+			if (doMatchPrefix && matchSuffix) {
+				// the same as RULE_EXACT_MATCH
+				return stringPattern.equalsIgnoreCase(text);
+			}
+			return text.toLowerCase().contains(stringPattern.toLowerCase());
 		}
 	}
 
 	private void initializePatternAndMatchRule(String pattern) {
-		int length = pattern.length();
-		if (length == 0) {
+		if (pattern.length() == 0) {
 			matchRule = RULE_BLANK_MATCH;
 			stringPattern = pattern;
 			return;
 		}
-		char last = pattern.charAt(length - 1);
+
+		// pre-process the string pattern
+		char first = pattern.charAt(0);
+		char last = pattern.charAt(pattern.length() - 1);
+		// note: a file name might start with a space => we can't use it for enforcing prefix match
+		matchPrefix = pattern.length() > 1 && first == START_SYMBOL;
+		matchSuffix = pattern.length() > (matchPrefix ? 2 : 1) && (last == END_SYMBOL || last == BLANK);
+		stringPattern = pattern;
+		if (matchPrefix) {
+			stringPattern = stringPattern.substring(1);
+		}
+		if (matchSuffix) {
+			stringPattern = stringPattern.substring(0, stringPattern.length() - 1);
+		}
 
 		if (pattern.indexOf('*') != -1 || pattern.indexOf('?') != -1) {
 			matchRule = RULE_PATTERN_MATCH;
-			switch (last) {
-			case END_SYMBOL:
-			case BLANK:
-				stringPattern = pattern.substring(0, length - 1);
-				break;
-			case ANY_STRING:
-				stringPattern = pattern;
-				break;
-			default:
-				stringPattern = pattern + ANY_STRING;
+			if (substringSearch && !matchPrefix && first != ANY_STRING) {
+				stringPattern = ANY_STRING + stringPattern;
+			}
+			if (!matchSuffix && last != ANY_STRING) {
+				stringPattern += ANY_STRING;
 			}
 			return;
 		}
 
-		if (validateMatchRule(pattern, RULE_CAMELCASE_MATCH) == RULE_CAMELCASE_MATCH) {
+		if (validateMatchRule(stringPattern, RULE_CAMELCASE_MATCH) == RULE_CAMELCASE_MATCH) {
 			matchRule = RULE_CAMELCASE_MATCH;
-			stringPattern = pattern;
 			return;
 		}
 
-		if (last == END_SYMBOL || last == BLANK) {
+		if ((!substringSearch || matchPrefix) && matchSuffix) {
 			matchRule = RULE_EXACT_MATCH;
-			stringPattern = pattern.substring(0, length - 1);
 			return;
 		}
 
 		matchRule = RULE_PREFIX_MATCH;
-		stringPattern = pattern;
-
 	}
 
 	/**
@@ -242,6 +304,20 @@ public class SearchPattern {
 				return false;
 		}
 		return true;
+	}
+
+	/**
+	 * @param text
+	 * @param suffix
+	 * @return true if text ends with given suffix, ignoring case; false in other
+	 *         way
+	 */
+	private boolean endsWithIgnoreCase(String text, String suffix) {
+		int textLength = text.length();
+		int suffixLength = suffix.length();
+		if (textLength < suffixLength)
+			return false;
+		return startsWithIgnoreCase(text.substring(textLength - suffixLength), suffix);
 	}
 
 	/**
@@ -425,32 +501,41 @@ public class SearchPattern {
 		// check first pattern char
 		if (name.charAt(nameStart) != pattern.charAt(patternStart)) {
 			// first char must strictly match (upper/lower)
-			return false;
+			if (!this.substringSearch || this.matchPrefix) {
+				return false;
+			}
+			nameStart = name.indexOf(pattern.charAt(patternStart), nameStart + 1);
+			if (nameStart < 0) {
+				return false;
+			}
 		}
-
-		int patternLength = patternEnd;
-
-		if (pattern.charAt(patternEnd - 1) == END_SYMBOL || pattern.charAt(patternEnd - 1) == BLANK)
-			patternLength = patternEnd - 1;
 
 		char patternChar, nameChar;
 		int iPattern = patternStart;
 		int iName = nameStart;
 
-		// Main loop is on pattern characters
+		// Main loop starts from the 2nd characters...
 		while (true) {
 
 			iPattern++;
 			iName++;
 
 			if (iPattern == patternEnd) {
-				// We have exhausted pattern, so it's a match
+				// We have exhausted pattern, so it might be a match
+				if (!this.matchSuffix) {
+					return true;
+				}
+				for (int i = iName; i < nameEnd; i++) {
+					if (isNameCharAllowed(name.charAt(i))) {
+						// There is another uppercase further in the name, but the pattern doesn't allow
+						// this
+						return false;
+					}
+				}
 				return true;
 			}
 
 			if (iName == nameEnd) {
-				if (iPattern == patternLength)
-					return true;
 				// We have exhausted name (and not pattern), so it's not a match
 				return false;
 			}
@@ -466,24 +551,15 @@ public class SearchPattern {
 			if (!isPatternCharAllowed(patternChar))
 				return false;
 
-			// patternChar is uppercase, so let's find the next uppercase in
+			// patternChar is uppercase, so let's find the next patternChar-matching uppercase in
 			// name
 			while (true) {
 				if (iName == nameEnd) {
-					if ((iPattern == patternLength) && (patternChar == END_SYMBOL || patternChar == BLANK))
-						return true;
+					// We have exhausted name (and not pattern), so it's not a match
 					return false;
 				}
 
 				nameChar = name.charAt(iName);
-
-				if ((iPattern == patternLength) && (patternChar == END_SYMBOL || patternChar == BLANK)) {
-					if (isNameCharAllowed(nameChar)) {
-						return false;
-					}
-					iName++;
-					continue;
-				}
 
 				if (Character.isDigit(nameChar)) {
 					// nameChar is digit => break if the digit is current pattern character
@@ -653,6 +729,9 @@ public class SearchPattern {
 	 * @return true if the given pattern is a sub pattern of this search pattern
 	 */
 	public boolean isSubPattern(SearchPattern pattern) {
+		if (this.initialPattern.length() == 1 && this.initialPattern.charAt(0) == START_SYMBOL) {
+			return false;
+		}
 		return trimWildcardCharacters(pattern.initialPattern).startsWith(trimWildcardCharacters(this.initialPattern));
 	}
 
--- a/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench; singleton:=true
-Bundle-Version: 3.127.100.qualifier
+Bundle-Version: 3.128.0.qualifier
 Bundle-ClassPath: .
 Bundle-Activator: org.eclipse.ui.internal.WorkbenchPlugin
 Bundle-ActivationPolicy: lazy
--- a/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.ide; singleton:=true
-Bundle-Version: 3.20.0.qualifier
+Bundle-Version: 3.20.100.qualifier
 Bundle-ClassPath: .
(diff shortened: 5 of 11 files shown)


bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench; singleton:=true
-Bundle-Version: 3.125.100.qualifier
+Bundle-Version: 3.126.0.qualifier
```
pbodnar: TODO: Once confirmed, also update versions in MANIFEST.MF of depending projects.

mickaelistria:
For consistency, couldn't it be a `SearchPattern.RULE_INFIX_MATCH` or something like that?

pbodnar:
> For consistency, couldn't it be a `SearchPattern.RULE_INFIX_MATCH` or something like that?

Possibly yes, I will have to look at how it could be incorporated into the existing code base, but also how to make it easy to use for the client code. Maybe a related question to be sure: We want to keep the behavior backwards compatible for existing clients (so that they won't have to change their code in order to keep the old behavior), don't we?

mickaelistria:
> We want to keep the behavior backwards compatible for existing clients (so that they won't have to change their code in order to keep the old behavior), don't we?

As much as possible yes. If some important feature does require to break some behavior, this needs to be discussed with the PMC for a decision.

pbodnar:
So, as can be seen from the commits, I have implemented some other points from the todo-list in the description. The next big thing in the list is this one:

> * make sure found resources are properly highlighted (see existing `ResourceItemLabelTest`)

After this, I would probably move on to the last:

> * present the SearchPattern's autoInfixSearch feature on UI layer somehow

Plus of course, some documentation pages should be updated...

---

BTW The Jenkins build is currently failing on some error I don't quite understand. It is possible that the error is quite unrelated to commits in this PR, but I don't know really...

---

A final thought, after digging into this search topic more, I think that **replacing** the current "wildcards / camelCase / etc.-based searches" with a brand-new "fuzzy search" as known from the other IDEs out there (i. e. when writing _simply & easily_ "illexc" finds "**Ill**egalArgument**Exc**eption") is the way to go. Yet, even finishing this one (which is rather an **extension** of the existing searches) could be a good start before moving on to a more "radical" change like that?

As usual, feedback and hints are welcome. :)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -52,7 +56,9 @@ public class SearchPattern {
 	/**
 	 * Match rule: The search pattern matches the search result only if cases are
 	 * the same. Can be combined with previous rules, e.g. {@link #RULE_EXACT_MATCH}
-	 * | {@link #RULE_CASE_SENSITIVE}
+	 * | {@link #RULE_CASE_SENSITIVE}.
+	 * <p>
+	 * TODO The actual code seems to ignore this flag when performing searches.
```
pbodnar: I've noticed this when going through the code, I'm not sure what to think about it / what to do about it. IMHO luckily nobody probably cares about this case-sensitivity flag in their client code...?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -549,6 +625,33 @@ public final int getMatchRule() {
 		return this.matchRule;
 	}
 
+	/**
+	 * @return whether to automatically expect that infix search should be done.
+	 *         Prefix search may be still enforced by placing " " or ">" at the
```
pbodnar: TODO: `" " or` should be removed from this comment.

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
```diff
@@ -436,6 +441,21 @@ protected Comparator getItemsComparator() {
 						return 1;
 					}
 				}
+				// Prioritize names starting with the pattern
```
pbodnar, line 471: This could be hopefully sufficient?

The existing comparator code is written in a way which doesn't seem to let us cover the code with corresponding unit tests easily. But hopefully it doesn't matter that much in this case.

pbodnar:
Status update: I've tried some real-world manual testing, finding I need to fix handling of some edge cases in / with the new code. Approximately these:

1. when entering just " " or "> ": ArrayIndexOutOfBoundsException
2. nothing is found when typing-in (not copy-pasting) ">abc", because ">" already finds nothing (and "subFilter concept" kicks in, filtering an empty set on entering ">a" etc.)
3. ResourceFilter should probably work with SearchPattern.initialPattern in order to handle dots (".") in the search string correctly, while pattern is being typed in the search dialog

pbodnar:
Wow, it looks like everything is fine again :), I guess the previous build failures were given by the June release / freeze period somehow.

Anyway, as can be seen, I have (hopefully) coped with the 3 points described above - by the 3 follow-up commits. So if someone could review and give a **feedback**, it would be great. Mainly the 3rd point was quite challenging, as I had to drill through many historical tickets to find out what is going on there in the `ResourceFilter` class & retesting various scenarios manually (no easy tests automation in this area without a huge refactoring, it seems).

I'm also not sure if I have correctly selected to merge the master branch into mine, or if I should do a rebase instead. So please let me know, provided I should do this differently.

HannesWell:
> I'm also not sure if I have correctly selected to merge the master branch into mine, or if I should do a rebase instead. So please let me know, provided I should do this differently.

A rebase is perfectly fine. We try to avoid merge commits.
Usually it does not matter when a change happened at your computer and what rounds you spinned until the change is completed, what matters is what changes were finally committed to the master. And this is best tracked with a simple linear history where each commit is a self-contained change. So unless you have unrelated changes in this PR or large changes that can be understood more easily if applied incrementally you can squash your commits into one or at least few self-contained commits.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Removed (within freshly rebased commits).

pbodnar:
@HannesWell, thanks for your quick info. So while still probably quite far from being a fully implemented PR, I have at least master-rebased, reorganized & squashed the 8 commits into 3 relatively "self-contained" ones, which I believe can somewhat simplify the reviews.

pbodnar:
OK, so I have finished some smaller amendments and the code is ready for review & testing now (except for the 2 remaining tasks from PR description). I think I will take a small break from this one. ;)

So @mickaelistria, @vogella or whoever is watching this issue, can you please have a look? It would be great to have some concrete suggestions before moving on, so that we avoid a major rework after everything is implemented. Thanks in advance. :)

mickaelistria:
I'm quite busy with a handful of high priority tasks. I'll try to review it next week if no other committer does look at it in the meantime (but having someone else to review would be very welcome!)

vogella:
I will be soon on vacation for a few weeks so I cannot review at the moment. I suggest to squash the commits into one commit using interfactive rebase and force push to your branch. Also I would suggest to remove the WIP from the PR, WIP is usually something committers ignore as WIP indicates that the change is not yet complete but to me it looks (on a quick glance) that this change is actually ready for review and that you expecting feedback on it.

Thanks @pbodnar for working on this, it is really nice to see an issue being worked on that makes IntelliJ better than Eclipse

vogella:
@pbodnar this one needs manual rebase.

pbodnar:
OK, freshly rebased on the latest master, squashed into 1 commit, updated the description and removed prefix "WIP". All ready for a review, hopefully. :)

mickaelistria (changes requested):
This should really be a MatchRule like others. Keeping this consistencyy would reduce the amount of new API to be created.

mickaelistria:
Also, I've looked a bit on the web to find a formal definiton of `infix search` and couldn't find something relatively standard. Can you please properly define it, or use another more standard term?

pbodnar:
> This should really be a MatchRule like others. Keeping this consistencyy would reduce the amount of new API to be created.

OK, I can see we could remove those 2 new constructors and (currently unused) `isAutoInfixSearch()` afterwards, right? I expect that for backwards compatibility reasons, we won't make the new rule the default one in the no-argument constructor, so can someone please suggest the best "Eclipse code-base" way of how to make the API developer-friendly, so that for example `RULE_EXACT_MATCH | RULE_PREFIX_MATCH | RULE_PATTERN_MATCH | RULE_CAMELCASE_MATCH | RULE_BLANK_MATCH | RULE_INFIX_MATCH` doesn't have to be copied to every client code which would like to use the new behavior? Maybe introduce a dedicated public constant(s) for the various rule sets? I can also imagine introducing a builder for the `SearchPattern` class, but that would probably go above the scope of this PR...?

> Also, I've looked a bit on the web to find a formal definiton of `infix search` and couldn't find something relatively standard. Can you please properly define it, or use another more standard term?

Regarding the `infix search` term, there are many pages using this term, but I can see it doesn't have to be necessarily the best in our context (and it would deserve an explanatory sentence in the javadoc). What to use instead though? Page [What is it called when you search the middle of a string instead of the beginning?](https://dba.stackexchange.com/questions/187658/what-is-it-called-when-you-search-the-middle-of-a-string-instead-of-the-beginnin) suggests `unanchored search`, which comes from the world of regex, but maybe `contains search` is a "more natural" candidate, although it has a verb in its name?

mickaelistria:
JDT's SearchPattern calls it [SUBSTRING_MATCH](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/7eeb9b4342d4d83f3a5a1d078a4e4dff0ab83bb2/org.eclipse.jdt.core/search/org/eclipse/jdt/core/search/SearchPattern.java#L253) if I understand correctly. That's a good term to reuse.

pbodnar:
OK, I will try to refactor the code - to introduce a new flag `RULE_SUBSTRING_MATCH`...

pbodnar:
@mickaelistria, as can be seen above, I have [force-pushed](https://github.com/eclipse-platform/eclipse.platform.ui/compare/3dba2df70f44988abb75be11cafaf9a7c3f0df25..0b51e283e93b1295c68e8d6f68f312b4bd0e2514) the required changes, together with amending the commit message accordingly (the 2nd force push is just another necessary rebase on master). Is it at least close to what you have expected? :)

Actual problem: Eclipse as well as probably the builds here tell me that introducing the new class field (constant) is an "API breaking change". So according to the docs, it looks that either module's version needs to be increased (from 3.x to 4.0.0) [1], or a "compatibility problem filter" (".settings/.api_filters") might be created [2]. Which way should we choose here?

[1] https://wiki.eclipse.org/Version_Numbering
[2] https://wiki.eclipse.org/PDE/User_Guide#Creating_filters

merks:
@pbodnar 

Use a filter.   

Downstream bundles typically use this type of version range:

image

So a major version increment would make a great many downstream bundles stop working.

pbodnar:
@mickaelistria, @merks, thank you for your feedback. I think the code is ready for another review now. :)

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/SearchPatternAuto.java
```diff
@@ -111,66 +91,42 @@ public void testPatternMatch2() {
...
-	 * Every string starting with an upperCase character should be recognized as camelCase pattern match rule.
-	 * Result for "AbCd " SearchPattern should be similar to regexp pattern "Ab[^A-Z]*Cd[^A-Z]*".
 	 */
 	@Test
-	public void testCamelCaseForcedEndMatch() {
-		Pattern pattern = Pattern.compile("Ab[^A-Z]*Cd[^A-Z]*");
-		assertMatches("AbCd ", SearchPattern.RULE_CAMELCASE_MATCH, pattern);
-		// alternative ending character:
-		assertMatches("AbCd<", SearchPattern.RULE_CAMELCASE_MATCH, pattern);
+	public void testCamelCaseChangingCaseMatch() {
```
mickaelistria, line 109: Why are those tests modified?
It would be better to keep them, to emphasize that previous use-cases are not broken; and to add extra ones.

pbodnar: @mickaelistria, I see your point. The reason behind the changes is that I found it more "friendly" to have similar sets of tests for original vs. new way of search, so they are more easily comparable.

So I slightly amended the original tests, so that they cover the various scenarios in a better / broader way. For example this one - testing `"AbCd "` - missed the check that for example "`ZzAbCd`" won't be found by the original search (if I remember it correctly, all the generated `resources` basically start with `A`). Of course, I can revert some of the changes as you propose, but I'm not sure if it is worth it...

vogella:
@[pbodnar](https://github.com/pbodnar) can you have a look at the feedback from @mickaelistria

pbodnar:
@vogella, I think I have already done so? Maybe if you could more specific? :)

bundles/org.eclipse.ui.workbench/.settings/.api_filters
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
mickaelistria, line 1: You shouldn't need this file if you set proper baseline in your IDE (should now be 4.25 release)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
mickaelistria: This TODO shouldn't be in Java doc, but a regular comment (accessible only to anyone who reads the code).

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -85,10 +95,33 @@ public class SearchPattern {
...
+	/**
+	 * Match rule: The search pattern is a string placed anywhere in the search
+	 * result. When in effect, this flag also affects some of the other match rules
+	 * (see their documentation for details).
+	 * <p>
+	 * Prefix search may still be enforced by placing '>' at the beginning of the
+	 * pattern. Analogically, suffix search may be enforced by placing ' ' or '&lt;'
+	 * at the end of the pattern.
+	 *
+	 * @since 3.126
```
mickaelistria: Should now be 3.127 as 3.126 was released in the meantime.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -149,6 +190,16 @@ public String getPattern() {
 		return this.stringPattern;
 	}
 
+	/**
+	 * Gets the initial (input) string pattern.
+	 *
+	 * @return pattern
+	 * @since 3.126
```
mickaelistria: Should now be 3.127 as 3.126 was released in the meantime.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -549,6 +630,22 @@ public final int getMatchRule() {
 		return this.matchRule;
 	}
 
+	/**
```
mickaelistria: Can't we do without those 2 new APIs? as we have access to the rule one can query it with bit masks. We try to minimize the amount of API we create/maintain in Platform.

bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
mickaelistria: Should now be 3.127.0 as 3.126 was released in the meantime.

mickaelistria (changes requested):
I'd like to merge it soon; can you please address the comments I submitted inline?

bundles/org.eclipse.ui.workbench/.settings/.api_filters
pbodnar: OK, but after removing this file (and after upgrading the IDE & baseline & Git rebase), I still get this error message on MANIFEST.MF, similarly to what happened to me previously:

    The major version should be incremented in version 3.127.0, since API breakage occurred since version 3.126.100

And the same goes for the new constants in SearchPattern:

    The field org.eclipse.ui.dialogs.SearchPattern.RULE_SUBSTRING_MATCH has been added to a class

Can you please give advice on how to cope with this then? Or maybe this file is actually still necessary?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Well, I see your point / goal, but I'm not quite sure here. If we were to "query the pattern's rule with bit mask", I think the code is not ready for this scenario yet? I.e. a) there is no `RULE_SUFFIX_MATCH` to use for such a query, b) just one of the `RULE_*` constants is set into the inferred `matchRule` field now, so it doesn't expect to be a combination of multiple flags.

pbodnar:
> I'd like to merge it soon; can you please address the comments I submitted inline?

That is good news. :) I have prepared some of the suggested amendments, but I have questions to some others if you could please have a look at them before I do a fresh force push. Thank you.

bundles/org.eclipse.ui.workbench/.settings/.api_filters
mickaelistria: OK, I see. Let's keep it then. The rule is defined in https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_Classes
```
Adding an API field to an API class that is extended by Clients is dangerous in two respects:

    It may break API contract compatibility similar to [Evolving_Java-based_APIs#Example_4_-_Adding_an_API_method](https://wiki.eclipse.org/Evolving_Java-based_APIs#Example_4_-_Adding_an_API_method) in case of name clashes.
    It may cause linkage errors in case an instance (respectively static) field hides a static (respectively instance) field, see [JLS8 13.4.8](https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.4.8). This can not happen if all field declarations follow the usual naming conventions from [JLS8 6.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html#jls-6.1) and classes only have API fields that are either
        "static final" (and hence have a name that doesn't contain any lowercase letters), or
        non-static and non-final (and hence have a name that contains a least one lowercase letter)

Apart from the binary compatibility issues, it is generally good software engineering practice that API classes should not expose any non-constant fields.
```
and I believe we're in the case where all API (non-private) fields in SearchPattern are `static final`

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
mickaelistria: We don't really care about a `RULE_SUFFIX_MATCH`; also I wonder can't pattern already deal with suffix? Why do you need particular consideration of suffix here; can't this be translated into a regular pattern without extra state?
Anyway the javadoc of `getMatchRule` says we can have the rule combined using bit masks.
In any case, it would be great if we can just keep this methods private, or package protected.

mickaelistria (dismissed):
I've added some answers, hopefully helpful.
Please also update the Javadoc for the constructor and getMatchRule to mention the new RULE_SUBSTRING_MATCH

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: I have updated the javadocs in the new commit to better describe the reality. I think they were historically wrong, so you might have also been mislead by them...

> Can't we do without those 2 new APIs?

I will recheck, it will probably require to clone the suffix/prefix-search-checking code into the client code (which is currently just the Resource dialog)...

bundles/org.eclipse.ui.workbench/.settings/.api_filters
pbodnar: OK. Note that by coincidence, this file emerged in the meantime in master, so I have put the corresponding lines to it after doing Git rebase.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, done.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, done.

pbodnar:
> I've added some answers, hopefully helpful.
> Please also update the Javadoc for the constructor and getMatchRule to mention the new RULE_SUBSTRING_MATCH

Thank you. So I've updated the javadocs, see my comment [above](https://github.com/eclipse-platform/eclipse.platform.ui/pull/12#discussion_r978103989) for details.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, done. I thought it wouldn't be bad to also let the clients know directly, but OK, never mind.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: @mickaelistria, so I have "moved" the new `isMatch...()` methods into the Dialog code in the new commit. Can you please check it out? It can't be probably done much simpler, provided we want to keep all the changes small?

pbodnar:
Getting closer? :) We've still got some unfinished tasks - see "Yet to be investigated / implemented" in the description. Ideas / directions are still welcome...

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -39,21 +39,30 @@ public class SearchPattern {
 
 	/**
 	 * Match rule: The search pattern is a prefix of the search result.
+	 * <p>
+	 * This is the <em>default</em> match rule if no other applies.
```
mickaelistria: I don't think this should be documented here, but more on the method/constructor that do take this decision.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -39,21 +39,30 @@ public class SearchPattern {
...
 	/**
 	 * Match rule: The search pattern contains one or more wild cards ('*' or '?').
 	 * A '*' wild-card can replace 0 or more characters in the search result. A '?'
 	 * wild-card replaces exactly 1 character in the search result.
+	 * <p>
+	 * Unless the pattern ends with ' ' or '>', '*' is automatically added to the
+	 * end of the pattern.
+	 * <p>
+	 * When {@link #RULE_SUBSTRING_MATCH} is in effect, then '*' is automatically
+	 * added to the start of the pattern.
```
mickaelistria: I don't think this should be documented here; it's more internal technical decision, we don't want to make it an API contract.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -85,10 +94,33 @@ public class SearchPattern {
 	 * <code>"NPE"</code> string pattern, search will only use Camel Case match
 	 * rule, but with <code>N*P*E*</code> string pattern, it will use only Pattern
 	 * match rule.
-	 *
+	 * <p>
+	 * Unless {@link #RULE_SUBSTRING_MATCH} is in effect, it is required that
+	 * the 1st pattern's char must match the very 1st char of the search result.
```
mickaelistria, line 92: This doesn't need to be spec'd; this is the behavior one would expect according to other existing documentation.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -105,39 +139,48 @@ public class SearchPattern {
...
+
+	private boolean matchSuffix;
+
 	/**
 	 * Creates a new instance of SearchPattern with the following match rules
 	 * configured: {@link #RULE_EXACT_MATCH} | {@link #RULE_PREFIX_MATCH} |
 	 * {@link #RULE_PATTERN_MATCH} | {@link #RULE_CAMELCASE_MATCH} |
-	 * {@link #RULE_BLANK_MATCH} )
-	 *
+	 * {@link #RULE_BLANK_MATCH}.
```
mickaelistria: Documentation should mention this uses `DEFAULT_RULES`

mickaelistria (commented):
Code looks ready; just some notes about documentation (trying to find the right grain about what we actually want to specify and where, without restraining from other pieces from changing later if need be)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Are you sure about this? I couldn't find anything like that in the public documentation of this class. There is an example saying `NPE` will match `NullPointerException`, but it doesn't say it won't match for example `InvalidNullPointerException`.

Generally I think it really should be documented how concretely the new flag `RULE_SUBSTRING_MATCH` affects this and other behaviors (as `RULE_PATTERN_MATCH` above), somewhere - maybe moving this just to the new flag itself would be better? You say stuff like this shouldn't be part of the API contract - do I get it right that it means that we want to keep the flags "general" and kind of "not to propagate" how concretely the SearchPattern class behaves while doing concrete searches? Of course, if someone wants, they can find that info in the unit tests, but I'm still not sure if that suffices...

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Well, I see that talking about concrete characters is kind of "internal technical decision", yet I think it also currently fills the gap of describing how exactly this flag behaves when talking about prefix and suffix searches. So do you think this description should be removed altogether (as discussed in the other, related thread), or maybe just reformulating it in a general way (i. e. not mentioning concrete characters used) would help here?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Good point, this goes against DRY. So simply change this e.g. to `Creates a new instance of SearchPattern with {@link #DEFAULT_MATCH_RULES default set of rules} configured.`?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, I think I get the point here. At one of the constructors, it is de-facto mentioned?:

```java
        /** ...
	 *                     Note also that default behavior for generic types/methods
	 *                     search is to find prefix matches.
	 */
	public SearchPattern(int allowedRules) {
		this.allowedRules = allowedRules;
		this.substringSearch = (allowedRules & RULE_SUBSTRING_MATCH) != 0;
	}
```

I'm not sure what exactly this sentence means by "generic types/methods" though, it is pretty old code already. So what do you suggest? Maybe to move this info just to the `matches()` method, to keep it simple, less confusing and just at one place?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
mickaelistria: The lest we specify, the easier maintenance is. Something that was specified to work in a particular way cannot easily be changed later to work differently. So let's specify the purpose of the flag, but not the implementation.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
mickaelistria: ok

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
mickaelistria: Yes, I think that would be good.

bundles/org.eclipse.ui.workbench/.settings/.api_filters
pbodnar: Considering this done.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/SearchPatternAuto.java
pbodnar: @mickaelistria, so is it OK from you under the described circumstances?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, done.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, I've reformulated the new paragraphs, so they don't describe what exactly happens with the "internal" pattern. I hope you'll find that OK (or suggest to move it to the updated `matches()` javadoc).

At the same time I have realized we are getting into a little bit "schizophrenic" situation here, because the resulting "internal" pattern is historically part of the public API of this class, yet it was never actually documented what it holds. So it is a little bit challenging stuff. My original intention was just to enhance the original javadocs as necessary for this feature, but as it turns out, a little bit bigger changes are necessary to get a coherent result. Hopefully at least, I haven't made the docs worse... ;)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: So I've moved all this stuff to the `matches()` method, I hope it makes things better.

Maybe a generally better approach would be to move much of the scattered documentation to the javadoc of the class itself. But that would be a major overhaul which I haven't originally planned for this ticket... :)

vogella:
Can this one be finished? Would be great to honor the work of @pbodnar

mickaelistria:
@pbodnar Can you please rebase on top of master, fix the conflict in .api_filter and push again? As soon as CI is happy, I think we can merge.

pbodnar:
@mickaelistria , sure thing. :) But I wonder about these opened tasks from issue description:

1)
> * present the SearchPattern's autoInfixSearch feature on UI layer somehow (change the Open Resource dialog label at minimum)

Currently, there is this label shown in the Open Resource dialog:

```
OpenResourceDialog_message = &Enter resource name prefix, path prefix or pattern (?, * or camel case):
```

So, we should probably change this to something like `Enter resource name, path or pattern (?, * or camel case):`?

Next, I'm not sure how much it would be desired to allow users to switch to the original prefix-search behavior. I can imagine adding an extra toggle action into the existing dialog menu (created inside `FilteredResourcesSelectionDialog.fillViewMenu(IMenuManager)`). But as I write, I'm not sure whether it is worth it.

2)
> * update corresponding user documentation

Shouldn't the help page [Open Resource](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Freference%2Fref-dialog-open-resource.htm) be updated? If yes, any tips on how to do that? :)

mickaelistria:
OK, let's just change the label and do not present option in UI for Open Resources; if there is really no drawback to expect.

> Shouldn't the help page Open Resource be updated? If yes, any tips on how to do that? :)

Yes. This would be a PR for https://github.com/eclipse-platform/eclipse.platform.common/blob/master/bundles/org.eclipse.platform.doc.user/reference/ref-dialog-open-resource.htm

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
pbodnar: No one protested, so assuming this sorting implementation is sufficient and closing this thread.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Seems like accepted, closing thread.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Seems like accepted, closing thread.

pbodnar:
> OK, let's just change the label and do not present option in UI for Open Resources...

OK, I have done so at 0ea742e4b4e5e0f67083e355ef3bd3ccbfc18fa0, which includes amending the remaining TODO. If it is OK for you, I would do the rebase afterwards.

> ... ; if there is really no drawback to expect.

As prefix search (startsWith) is by definition faster than infix one (contains), users with lots of files in their workspace can possibly notice some performance decline, but I have no hard data for this. Anyway, if informed properly, users will still be able to enforce the original prefix search by typing '>', so I can't see a big problem when there will be no UI (checkbox-like) option.

> > Shouldn't the help page Open Resource be updated? If yes, any tips on how to do that? :)
> 
> Yes. This would be a PR for https://github.com/eclipse-platform/eclipse.platform.common/blob/master/bundles/org.eclipse.platform.doc.user/reference/ref-dialog-open-resource.htm

OK, thanks for the link. I will try to prepare a PR there.

Apart from that, do I get it right that someone should also extend the "New & Noteworthy" document?

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
pbodnar: Maybe reopen for a while with a question...

The current comparator implementation part:

```java
				// Prioritize names starting with the pattern
				char patternFirstChar = getFirstFileNameChar(pattern);
				if (patternFirstChar != 0) {
					patternFirstChar = Character.toLowerCase(patternFirstChar);
					m1 = patternFirstChar == Character.toLowerCase(s1.charAt(0));
					m2 = patternFirstChar == Character.toLowerCase(s2.charAt(0));
					if (!m1 || !m2) {
						if (m1) {
							return -1;
						}
						if (m2) {
							return 1;
						}
					}
				}
```

... is simple, but because of checking just the first letters, it wrongly evaluates for example name "Aabc" as a "name starting with the pattern" if the pattern is "ab". IMO the only 100% working solution would be either to de-facto re-implement the matching here (which is somewhat already done in the matches highlighting code) or to change the whole searching solution to keep metadata for every successful match. Until then, can we say that the solution above is "good enough" for now and maybe add some inline warning comment about it not being perfect?

pbodnar:
I have opened a corresponding documentation PR and also updated the issue description to match the current state. Probably all from me for today. :)

mickaelistria:
Eclipse 4.26 has entered an RC phase and only bugfixes, new tests or new documentation work can be merged safely until 4.26 is released (about 1 month from now). The code is IMO good, so we'll try to merge it just after 4.26 is released, as part of 4.27 M1. This will have the benefit of giving plenty of time for actual testing.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -85,10 +88,33 @@ public class SearchPattern {
...
+	/**
+	 * Match rule: The search pattern is a string placed anywhere in the search
+	 * result. When in effect, this flag also affects some of the other match rules
+	 * (see their documentation for details).
+	 * <p>
+	 * Prefix search may still be enforced by placing '>' at the beginning of the
+	 * pattern. Analogically, suffix search may be enforced by placing ' ' or '&lt;'
+	 * at the end of the pattern.
+	 *
+	 * @since 3.127
```
mickaelistria: It should now be 3.128 as we're opening development of 4.27

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -85,10 +88,33 @@ public class SearchPattern {
...
+	 */
+	public static final int RULE_SUBSTRING_MATCH = 0x0200;
+
+	/**
+	 * The default set of match rules as used by the no-argument constructor.
+	 *
+	 * @since 3.127
```
mickaelistria: It should now be 3.128 as we're opening development of 4.27

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -163,10 +206,19 @@ public void setPattern(String stringPattern) {
...
 	/**
-	 * Matches text with pattern. matching is determine by matchKind.
+	 * Matches text with pattern. The way of matching is determined by the current
+	 * pattern setup - see {@link #getMatchRule()} for details.
+	 * <p>
+	 * If no rule applies or when {@link #RULE_CAMELCASE_MATCH} applies, but the
+	 * text doesn't match it, the default implementation of this method performs a
+	 * simple prefix, suffix, or substring matching (determined by the pattern
+	 * setup).
+	 * <p>
```
mickaelistria: Such implementation details shouldn't be specified. They are not much necessary for consumer to know about in order to decide whether and how to use this method; and specifying too much makes it harder to change later as some behavior that were specified once cannot be changed without it being perceived as breaking.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -535,15 +616,13 @@ protected boolean isNameCharAllowed(char nameChar) {
...
-	 *         RULE_EXACT_MATCH | RULE_CASE_SENSITIVE if an exact and case sensitive
-	 *         match is requested, or RULE_PREFIX_MATCH if a prefix non case
-	 *         sensitive match is requested.
+	 * Returns the active rule to apply for matching keys, based on the currently
+	 * set {@link #setPattern(String) pattern} and allowed rules passed to the
+	 * {@link #SearchPattern(int) constructor}.
+	 *
+	 * @return one of {@link #RULE_BLANK_MATCH}, {@link #RULE_EXACT_MATCH},
+	 *         {@link #RULE_PREFIX_MATCH}, {@link #RULE_PATTERN_MATCH},
+	 *         {@link #RULE_CAMELCASE_MATCH}
```
mickaelistria: This can be perceived as a breaking change:
* why RULE_BLANK_MATCH now?
* What about case sensitivity?

mickaelistria (changes requested):
Please rebase on top of master and update necessary versions.
Also, please see inlined comments about documentation

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, I will remove this paragraph. My goal was to make the behavior clearer, but I guess that maybe removing / replacing some parts from the `RULE_CAMELCASE_MATCH` javadoc would be better (like `Can be combined with {@link #RULE_PREFIX_MATCH} match rule. ...` - because there is _no_ need for this flag probably from the very start already), but I will leave that historically "technical debts" for now, so that we can finish this ticket finally... :)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Well, if I haven't overlooked st, this is not a "breaking change", but just "fixing the javadoc" - to be unambiguous & complete (which I have somewhat mentioned in a previous discussion, if I remember correctly).

So, 1) `RULE_BLANK_MATCH` should be mentioned to make it complete and 2) "case sensitivity flag and how it can be combined" is already described in the linked constructor's javadoc & it is not quite relevant to what this method returns, therefore removing this part from here makes perfect sense to me.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: OK, going to update it.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
mickaelistria: Could those points be discussed in a separate PR?
Fixing (code or javadoc or whatever) is always welcome, with highest priority; so we don't want it to be a "by the way" thing. Keeping things separate allows more efficient progress and also makes development safer: it's easy to revert a new feature without reverting the "by the way" fixes, so fixes are less likely to disappear.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Problem: After increasing this here, Eclipse (logically?) wants me to increase the number also in MANIFEST.MF:

```diff
- Bundle-Version: 3.127.0.qualifier
+ Bundle-Version: 3.128.0.qualifier
```

But after doing this, Eclipse complains that "The type org.eclipse.ui.databinding.WorkbenchObservables has been removed from org.eclipse.ui.workbench_3.128.0" (and the same for type WorkbenchProperties; both have been added to ".api_filters" recently). I'm not sure at all at this "versioning" stuff - can you please help me resolve this problem? Maybe @vogella could help here as well?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: Good point, I will try to move all the javadoc-related fixes into a separate PR, OK? I can see now there are way too many problems from the past (little to no code review there?), so mixing their fix into this PR is problematic...

pbodnar: => #469

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: => #469

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
vogella: I have a look now

vogella:
@pbodnar did you update your API baseline? I downloaded https://download.eclipse.org/eclipse/downloads/drops4/S-4.26RC2-202211231800/ and set my workspace to use this a API baseline in which I did not get the error you described. Lets see what CI thinks of this change.

vogella:
@pbodnar unfortunately this change now conflicts with your Javadoc changes. Can you rebase your change onto master?

pbodnar:
@vogella, thanks for your assistance. :)

So I have rebased, squashed and moved version / `@since` to `3.128.0` / `3.128`. Not sure about the GitHub build (seems like a random error?), but the Jenkins build fails on this now:

```
21:00:02.429 [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release)
on project org.eclipse.ui.ide: Only qualifier changed for (org.eclipse.ui.ide/3.20.0.v20221208-2047).
Expected to have bigger x.y.z than what is available in baseline (3.20.0.v20221027-2208) -> [Help 1]
```

Can you please help once again?

mickaelistria:
The version of org.eclipse.ui.ide in the MANIFEST.MF should be increased by +0.0.100 to signify some changes have happened compare to last release. Also, by the way, its requirement on org.eclipse.ui.workbench should set 3.128.0 as minimal version as it uses the new APIs there.

bundles/org.eclipse.ui.navigator.resources/META-INF/MANIFEST.MF
```diff
@@ -13,7 +13,7 @@ Export-Package: org.eclipse.ui.internal.navigator.resources;x-internal:=true,
  org.eclipse.ui.internal.navigator.resources.nested;x-internal:=true,
  org.eclipse.ui.internal.navigator.workingsets;x-internal:=true,
  org.eclipse.ui.navigator.resources
-Require-Bundle: org.eclipse.ui.ide;bundle-version="[3.15.0,4.0.0)",
+Require-Bundle: org.eclipse.ui.ide;bundle-version="[3.20.100,4.0.0)",
```
mickaelistria: You don't need to change that as part of this change. Please remove it.

mickaelistria (changes requested):

bundles/org.eclipse.ui.navigator.resources/META-INF/MANIFEST.MF
pbodnar: OK, but what about the `org.eclipse.ui.internal.navigator.resources.actions.GotoResourceDialog` which extends the extended `org.eclipse.ui.dialogs.FilteredResourcesSelectionDialog`? No need for "explicit" dependency here?

pbodnar: Apart from that, it looks like this will be necessary according to the latest build failure (as tests have been modified as well):

```diff
diff --git a/tests/org.eclipse.ui.tests/META-INF/MANIFEST.MF b/tests/org.eclipse.ui.tests/META-INF/MANIFEST.MF
index 7a3db4f..df3bd11 100644
--- a/tests/org.eclipse.ui.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.ui.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse UI Tests
 Bundle-SymbolicName: org.eclipse.ui.tests; singleton:=true
-Bundle-Version: 3.15.900.qualifier
+Bundle-Version: 3.15.1000.qualifier
 Eclipse-BundleShape: dir
 Bundle-Activator: org.eclipse.ui.tests.TestPlugin
 Bundle-Vendor: Eclipse.org
```

... and hopefully that could be all? I wonder if these versions checks could be automated on localhost (dev's machine) somehow...

mickaelistria:
> OK, but what about the org.eclipse.ui.internal.navigator.resources.actions.GotoResourceDialog which extends the extended org.eclipse.ui.dialogs.FilteredResourcesSelectionDialog? No need for "explicit" dependency here?

No, because it doesn't consume new API methods or fields. It's still compatible with the older one.

> Apart from that, it looks like this will be necessary according to the latest build failure (as tests have been modified as well):

You got it ;)

> ... and hopefully that could be all?

Hopefully yes.

> I wonder if these versions checks could be automated on localhost (dev's machine) somehow...

Some of those checks can already be seen in the IDE if you configure a PDE baseline referencing latest release.
Some others are currently not available in the IDE; and a local build isn't really faster than a remote one, so it's OK to just push to PR to test.

bundles/org.eclipse.ui.navigator.resources/META-INF/MANIFEST.MF
pbodnar: resolved, see below

pbodnar:
@mickaelistria, thanks for the tips. So I have done the 2 changes in manifests as proposed + another forced rebase. And finally "All checks have passed". :)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: done

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
pbodnar: resolved, thank you - I had to update my Eclipse + baseline in it

mickaelistria (approved):

mickaelistria:
Thank you!
Can you please add a note to the N&N document as well ( https://github.com/eclipse-platform/www.eclipse.org-eclipse-news )?

jarthana:
This would be a good candidate for the N&N. Any plans to add it there?

vogella:
@pbodnar the new funcationality is super nice and easy to use, thanks a bunch for this.

One of my clients mentioned that it would also be very nice if the * in the middle of a work could be avoid, e.g. in the following example:

image

the match for TaskColumnPropertyAccessor continues if:

- I continue to write column, e.g. taskcolumn
- I write TCP

but not if I write taskproperti have to add an * to match this one, e.g. task*propert

Would you be interested in also implementing the automatic * in the middle of the search term? According to my client this is standard behavior in VSCode and IntelliJ.

pbodnar:
@vogella, thanks for your feedback. I have actually kinda proposed this kind of change at https://bugs.eclipse.org/bugs/show_bug.cgi?id=531785#c13 - I called it "fuzzy search" there. :)

Anyway, yes, it seems doable (with matching algorithm itself being quite trivial?), with some major refactorings (API and UI changes would have to be consulted?), given I already somewhat understand the searching code behind. I'm just not sure if it still pays off when Eclipse Theia is on its way, and it is also harder for me to find some extra time for Eclipse contributions these days... Maybe when there will be more votes on this? ;)

vogella:
> Anyway, yes, it seems doable (with matching algorithm itself being quite trivial?), 

If the matching algorithm can be adjusted easily why not change it without any API changes? I think it would be OK for our search to become better without offering API to go back to the "not so nice" behavior.

pbodnar:
@vogella, good point, just extending the behavior could probably be the simplest way, provided it would by widely accepted. Question is, what with all those existing `SearchPattern`'s flags like ` RULE_EXACT_MATCH`, `RULE_PREFIX_MATCH`, `RULE_PATTERN_MATCH` etc., moreover when we wouldn't introduce any new one...?

Besides that, matching, sorting and highlighting of the matches are currently done independently of each other. As seen already within this MR, this has got some limits (mainly in regards to sorting and coping with duplicated executions of +- the same stuff). I can imagine a major redesign where output from matching wouldn't be just a `boolean` - instead, it would be some structured data which could be directly used within sorting and highlighting. But once again, this is not quite trivial to do, IMHO...

PR #14, opened 2022-04-19
[WIP] Allow plugin fragments to supply alternate icons

### LINKED ISSUE #13: Add support for "Icon Packs"
This is a continuation/migration of [Bug 562174](https://bugs.eclipse.org/bugs/show_bug.cgi?id=562174) - Add support for "icon packs" in Eclipse IDE.

It'd be nice for to have an officially supported way to replace the icons used in Eclipse.

There is a lot of discussion to be held on how this should be handled (and there was already a lot of discussion on the original BZ thread), so I'm going to leave this relatively open-ended.

IIRC the approach I had initially tried was:
- Allow custom/replacement icons to be provided by a plugin fragment, by placing them in a "theme" directory of the plugin fragment
    - So in order to replace one of the "debug" icons, you must make a plugin fragment where the host plugin is "eclipse.debug.ui" and place the icon within a "theme" directory, with the same file path as the original icon
- When loading an icon from a plugin, check if the plugin provides an alternative path to the icon that is prefixed by "/themes/"
- If so, load it
- Otherwise, use the default icon

There are probably some limitations to this approach, for instance:
- The icon-provider must create a plugin fragment for each plugin which they wish to replace the icons for. This is very tedious.
- How do we handle switching between different icon packs, if each plugin fragment uses the same "themes" directory?

It might be nicer to set some JFace preference which points to a specific plugin path (where the plugin is the icon pack provider plugin), and then alternative icons are checked from that plugin before loading the default icon. 

I am going to submit a WIP PR that shows my current implementation, as well as an example plugin fragment that provides icon replacements.

AObuchow:
Fix #13 

Allow plugin fragments to contribute alternate icons that will be loaded
through the URLImageDescriptor class.

The icons must be placed within a directory titled "theme"
located at the root of the plugin fragment, ie. "/theme/"

The file path of the replacement icons must match the file path of the
icon to be replaced.

For example, in order to replace an icon used by plugin org.eclipse.xxx
where the path of the icon is "/path/To/icon.png" in the plugin's bundle
one must:
- Create a plugin fragment with org.eclipse.xxx as the Host Plug-in
- Create a directory "/theme/" at the root of the plugin
fragment
- Place an alternate icon with the correct file path in the theme
directory, ie. "/theme/path/To/icon.png"

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
@@ -42,6 +42,9 @@ import org.eclipse.swt.graphics.ImageFileNameProvider;
  */
 class URLImageDescriptor extends ImageDescriptor {
 
+	private static final String THEME_REPLACEMENTS_DIR = "/theme"; //$NON-NLS-1$
+	private static final String PLUGIN_URL = "/plugin/"; //$NON-NLS-1$
+
 	private static class URLImageFileNameProvider implements ImageFileNameProvider {
 		private String url;
 
@@ -198,9 +201,17 @@ class URLImageDescriptor extends ImageDescriptor {
 	}
 
 	private static URL getxURL(URL url, int zoom) {
-		if (zoom == 100) {
-			return url;
-		}
+
+		URL resultingURL = url;
+		// First we check for an image with increased zoom, if desired
+
+		// TODO: Ideally, we shouldn't be commenting out this line. Figure out a way
+		// around removing it.
+
+		// if (zoom == 100) {
+		// return url;
+		// }
+
 		String path = url.getPath();
 		int dot = path.lastIndexOf('.');
 		if (dot != -1 && (zoom == 150 || zoom == 200)) {
@@ -215,12 +226,49 @@ class URLImageDescriptor extends ImageDescriptor {
 				if (url.getQuery() != null) {
 					file += '?' + url.getQuery();
 				}
-				return new URL(url.getProtocol(), url.getHost(), url.getPort(), file);
+				resultingURL = new URL(url.getProtocol(), url.getHost(), url.getPort(), file);
 			} catch (MalformedURLException e) {
 				Policy.getLog().log(new Status(IStatus.ERROR, Policy.JFACE, e.getLocalizedMessage(), e));
 			}
 		}
-		return null;
+
+		// Next we check if there's an alternate image that we should use
+		// TODO: I'm not exactly sure how we should be accessing the JFace preference
+		// Should a IPreferenceStore instance be retrieved?
+		// Or is that considered part of Eclipse RCP and not JFace?
+		if (
+		// !JFaceResources.getString(JFacePreferences.CUSTOM_RESOURCE_THEME).equals(JFacePreferences.CUSTOM_RESOURCE_THEME)
+		// &&
+		resultingURL.getFile().contains(PLUGIN_URL)) {
+			// The third path segment is the bundleID
+			String originalFilePath = resultingURL.getFile();
+			for (int i = 0; i < 2; i++) {
+				originalFilePath = originalFilePath.substring(originalFilePath.indexOf('/') + 1);
+			}
+			// TODO: The original bundleID should only be used in the case of plugin
+			// fragment image replacements.
+			// We need an additional (optional) preference that will hold the bundleID of
+			// the currently used theme plugin
+			String bundleID = originalFilePath.substring(0, originalFilePath.indexOf('/'));
+			originalFilePath = originalFilePath.substring(originalFilePath.indexOf('/'), originalFilePath.length());
+
+			// TODO: This is a workaround and should probably be removed
+			if (originalFilePath.contains("/$nl$/")) { //$NON-NLS-1$
+
+				originalFilePath = originalFilePath.replace("/$nl$/", "/"); //$NON-NLS-1$ //$NON-NLS-2$
+			}
+			try {
+				URL themedIconURL = new URL(resultingURL.getProtocol(), resultingURL.getHost(), resultingURL.getPort(),
+						PLUGIN_URL + bundleID + THEME_REPLACEMENTS_DIR + originalFilePath);
+				String found = getFilePath(themedIconURL, false);
+				if (found != null) {
+					resultingURL = themedIconURL;
+				}
+			} catch (MalformedURLException e) {
+				Policy.getLog().log(new Status(IStatus.ERROR, Policy.JFACE, e.getLocalizedMessage(), e));
+			}
+		}
+		return resultingURL;
 
 	}
 
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/JFacePreferences.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/JFacePreferences.java
@@ -111,6 +111,13 @@ public final class JFacePreferences {
 	 */
 	public static final String REVISION_OLDEST_COLOR = "org.eclipse.jface.REVISION_OLDEST_COLOR"; //$NON-NLS-1$
 
+	/**
+	 * TODO: Add javadoc
+	 *
+	 * @since 3.26
+	 */
+	public static final String CUSTOM_RESOURCE_THEME = "org.eclipse.jface.CUSTOM_RESOURCE_THEME"; //$NON-NLS-1$
+
 	private static IPreferenceStore preferenceStore;
 
 	/**


AObuchow:
Here is the original [Gerrit](https://git.eclipse.org/r/c/platform/eclipse.platform.ui/+/174112) (it has some comments which are still valuable).

AObuchow:
In order to test out this PR:
- Checkout the PR locally, having the JFace Eclipse bundle imported
- Clone [my sample icon-providing fragment](https://github.com/AObuchow/Eclipse-Custom-Icon-Demo)
- Import the sample fragment into your Eclipse workspace
- Ensure the Eclipse application has all workspace plugins enabled (Or atleast, JFace and the sample fragment)

The observed result should be that the "run" and "debug" icons in the toolbar have been replaced:
image

mickaelistria:
The overall code seems good. What particularly is left TODO (apart from the documentation TODO in code) ?

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -215,12 +226,49 @@ private static URL getxURL(URL url, int zoom) {
...
+		}
+
+		// Next we check if there's an alternate image that we should use
+		// TODO: I'm not exactly sure how we should be accessing the JFace preference
+		// Should a IPreferenceStore instance be retrieved?
+		// Or is that considered part of Eclipse RCP and not JFace?
+		if (
+		// !JFaceResources.getString(JFacePreferences.CUSTOM_RESOURCE_THEME).equals(JFacePreferences.CUSTOM_RESOURCE_THEME)
+		// &&
+		resultingURL.getFile().contains(PLUGIN_URL)) {
```
laeubi, line 242: I don't think it is suitable to explicitly target the (internal) plugin URL format. We should use a scene that is independent from the source URL format like it is done with the zoom images.
The one don't needs to fiddle around with bundle ids and nls and alike...

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -215,12 +226,49 @@ private static URL getxURL(URL url, int zoom) {
...
+			if (originalFilePath.contains("/$nl$/")) { //$NON-NLS-1$
+
+				originalFilePath = originalFilePath.replace("/$nl$/", "/"); //$NON-NLS-1$ //$NON-NLS-2$
+			}
+			try {
+				URL themedIconURL = new URL(resultingURL.getProtocol(), resultingURL.getHost(), resultingURL.getPort(),
+						PLUGIN_URL + bundleID + THEME_REPLACEMENTS_DIR + originalFilePath);
+				String found = getFilePath(themedIconURL, false);
+				if (found != null) {
+					resultingURL = themedIconURL;
```
laeubi, line 265: I think it makes more sense to simply return here instead of fall through the default return.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -215,12 +226,49 @@ private static URL getxURL(URL url, int zoom) {
...
+
+			// TODO: This is a workaround and should probably be removed
+			if (originalFilePath.contains("/$nl$/")) { //$NON-NLS-1$
+
+				originalFilePath = originalFilePath.replace("/$nl$/", "/"); //$NON-NLS-1$ //$NON-NLS-2$
+			}
+			try {
+				URL themedIconURL = new URL(resultingURL.getProtocol(), resultingURL.getHost(), resultingURL.getPort(),
+						PLUGIN_URL + bundleID + THEME_REPLACEMENTS_DIR + originalFilePath);
+				String found = getFilePath(themedIconURL, false);
```
laeubi, line 263: I don't think that filepath is suitable here, this further limits the usage of this.

laeubi (changes requested):
I think this makes to much specific assumptions (icons are always local bundles, they can be resolved to a path, we know how image URLs are constructed,...).

I think it should more work with a delegation model where we construct an new URL for the theming and using the original one as a fallback image in case of errors.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -42,6 +42,9 @@
  */
 class URLImageDescriptor extends ImageDescriptor {
 
+	private static final String THEME_REPLACEMENTS_DIR = "/theme"; //$NON-NLS-1$
```
laeubi, line 45: Why limit to one particular folder here and not use what is provided by the `CUSTOM_RESOURCE_THEME`

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
mickaelistria: +1 with that

mickaelistria:
Thanks @laeubi for the review; I agree with those points, and trying to stick to (any form) of URL should work better.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
AObuchow: Agreed

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
AObuchow: My intention here was to validate that the newly constructed icon URL actually resolves, and if not, fallback to the default icon. Do you know if there is a better way to see if a URL resolves, that doesn't rely on the assumption that bundles are local?

AObuchow: Ah maybe instead of `getFilePath(themedIconURL, false)` I could do something like `URLImageDescriptor.getImageData(themedIconURL)`? I'm investigating [your commit](https://git.eclipse.org/r/c/platform/eclipse.platform.ui/+/170856) for help :)

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
AObuchow: Agreed, the parent directory of the icon should be as customizable as possible for the theme creator/provider.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
AObuchow: I'm +1 for doing that, as it simplifies things a lot.
However, I'm a bit confused on how to do this, since the requirements and assumptions for zoom images are different.

For zoom images, resolution-specifying directories (eg `icons/32x32/`) only need to come from the same bundle as the original icon. Thus it is possible to preprend the icon filename with the resolution-specifying directory, and keep the rest of the icon URL/filepath the same.

With themed icons, the requirements is to allow another bundle (that is different from the original bundle which provides the icon) to provide an alternative icon URL. Thus, we need to change the bundle in the icons URL. My original idea was to provide a JFace preference (`CUSTOM_RESOURCE_THEME`) which would specify which bundle is providing the alternative icon.

AObuchow:
Thanks for the review @mickaelistria :)

> The overall code seems good. What particularly is left TODO (apart from the documentation TODO in code) ?

I think asides from addressing @laeubi's comments, there are some followup tasks that will come from this change.

Some that come to mind are:
- Set defaults for the newly created JFace Preferences (one preference should probably specify the icon theme bundle, and another will specify the parent directory of the themed icon)
- Add an extension point to set these above two preferences in an icon theme bundle
- Add an extension point to associate an icon theme to an IDE theme (this would allow specifying icons to use for the dark theme, and different icons to use for the light theme)
- Add a UI preference to select the icon theme, similar to how IDE themes are chosen.

AObuchow:
@laeubi Thank you very much for the review :) I apologize for being so slow on this topic. I hoped it was just a matter of me implementing your requested changes, but it turns out I needed further clarification.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
laeubi: The only way that works for all cases is try open the URL, and maybe fall back to either the error or the original URL.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
laeubi: > For zoom images, resolution-specifying directories (eg `icons/32x32/`) only need to come from the same bundle as the original icon.

This depends on the URL type used. You can use http urls as well or even custom url types (like `mvn:`)

> With themed icons, the requirements is to allow another bundle (that is different from the original bundle which provides the icon) to provide an alternative icon URL. Thus, we need to change the bundle in the icons URL.

If you want to replace icons from a bundle, the most flexibel way would be to require that providers supply a fragment to this bundle, that is how the translation bundles work (it would even allow to provide different sizes right now) and then you can simply modify the name.

This of course would mean there is not only *one* icon bundle but different ones, on the other hand it is a standard way to enhance an existing bundle and would make it clear what icons are included, so one don not need to ask the icon supplier to add support for my special eclipse plugin but I can supply an own one (probably once included in the original icon pack).

gayanper:
@AObuchow are you planning to continue on this PR ? it would be really good to have this functionality to start creating some icons packs with the themes.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
AObuchow: > If you want to replace icons from a bundle, the most flexibel way would be to require that providers supply a fragment to this bundle, that is how the translation bundles work (it would even allow to provide different sizes right now) and then you can simply modify the name.
> 
> This of course would mean there is not only _one_ icon bundle but different ones, on the other hand it is a standard way to enhance an existing bundle and would make it clear what icons are included, so one don not need to ask the icon supplier to add support for my special eclipse plugin but I can supply an own one (probably once included in the original icon pack).

My worry with making the icon provider have to supply a fragment to the original icons bundle is that it would require _many_ icon bundles to replace all of Eclipse's standard icons. This would probably demotivate icon pack creators as it'd be very tedious, rather than creating a single plugin that replaces icons from various bundles (even though my idea contradicts the name of this PR).

With that being said, I am open to your approach as a "first step" towards supporting icon replacement in Eclipse. Perhaps a script/tool could be developed to automate the process of creating the required bundle fragments and populating them with the replacement icons. 

@gayanper @mickaelistria if either of you have any input on this, feel free to chime in :)

AObuchow:
> @AObuchow are you planning to continue on this PR ? it would be really good to have this functionality to start creating some icons packs with the themes.

Thanks for pinging me on this @gayanper :) Yes I am planning on continuing it, I just got caught up with life shortly after my last comments. Feel free to help contribute to this topic in any way, it is always appreciated :)

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
laeubi: > My worry with making the icon provider have to supply a fragment to the original icons bundle is that it would require _many_ icon bundles to replace all of Eclipse's standard icons.

The problem is, that each bundle has its own class/resource space so bundle A and B can have both `icon/editor.png` also note that there is no such thing as "all" icons because Eclipse can be extended and different combinations are used.

So it would be better to be more flexible here and tyr to give icon-pack creators tools to easier organize/automate this process, e.g. I can think of something like the babel-project for icon, where a tool simply scans a folder of plugins and discovers all image resources, then this is presented to the user as a big table and icons can be replaced and sutable fragments are generated, an update site is prepared and then one can install the packs required for the current eclipse install.

Beside this, I tried a while back to get SVG support into SWT (with little success because of concerns having a dependency to AWT), that would allow to style icons with CSS and could be even more profitable as you do not need to edit/replace *every* icon ...

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
mickaelistria: +1 with @laeubi proposal to not mandate a root directory and let path just be `/mySuperIconTheme/path/to/image.png`

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
mickaelistria: For a first iteration, as there is no API, I think it's OK to restrict the them-ability to plugins/fragments and consider opening it to other URI schemes later.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -215,12 +226,49 @@ private static URL getxURL(URL url, int zoom) {
...
+			// fragment image replacements.
+			// We need an additional (optional) preference that will hold the bundleID of
+			// the currently used theme plugin
+			String bundleID = originalFilePath.substring(0, originalFilePath.indexOf('/'));
+			originalFilePath = originalFilePath.substring(originalFilePath.indexOf('/'), originalFilePath.length());
+
+			// TODO: This is a workaround and should probably be removed
+			if (originalFilePath.contains("/$nl$/")) { //$NON-NLS-1$
+
+				originalFilePath = originalFilePath.replace("/$nl$/", "/"); //$NON-NLS-1$ //$NON-NLS-2$
```
mickaelistria, line 258: Why is `$nl$` removed here? I think it should be kept and resolved as usual without particular case.

mickaelistria (changes requested):

akurtakov:
Is there any hope for this one still?

AObuchow:
> Is there any hope for this one still?

Yes, I apologize for the extremely long delay. I believe the remaining changes that are required are relatively simple, I've just been putting it off due to personal priorities. I will try and make these changes over the next week. 

Since the consensus is to keep things relatively simple and rely on plugin fragments, I will follow this approach:

> So it would be better to be more flexible here and try to give icon-pack creators tools to easier organize/automate this process, e.g. I can think of something like the babel-project for icon, where a tool simply scans a folder of plugins and discovers all image resources, then this is presented to the user as a big table and icons can be replaced and sutable fragments are generated, an update site is prepared and then one can install the packs required for the current eclipse install.

akurtakov:
@Michael5601 This might be interesting for you as you deal with icons now.

Michael5601:
> @Michael5601 This might be interesting for you as you deal with icons now.

Thanks for the info. Seems like a great project. :)

Yesterday I talked with a colleague of mine about this topic as he searched for a new bigger project. This would fit perfectly with our new initiative for a [new modern icon pack](https://github.com/eclipse-platform/eclipse.platform.images/issues/108) in the Eclipse IDE. @gayanper maybe you are interested in participating as you talked about new icon packs?

I will forward this PR to my colleague as I probably won't have time due to other projects.

jannisCode:
Hi guys, I was asked if I could continue with this PR. First of  all I would like to know if my ideas are realistic before starting to implement them. 
My vision would be the following: In the settings there is a drop down (like currently with the dark/light theme). From there one can select a specific folder from which the icons should be taken. 
I would like to create a folder structure with a base folder `/icons` in which the folders of the icon themes lie (e.g. `/icons/default` or `/icons/modern`) in each of the icon-folders, **all eclipse icons** are saved, with the same name (default and modern would both have a icon called `run_exc.png`). This structure would make it rather easy to switch between the iconThemes. The **question** would just be: where to create the base folder? 
Also I see it as an important change, which is why I would maybe change the existing solution, so that it is not a Plugin/fragment but fully and by default integrated in eclipse.
The way I understand it, the switching of icons already works with this PR, so I could use the functionality which @AObuchow already made.
 Please feel free to point out things that i missed and/or understood incorrectly, I am really looking forward to working on this issue!

mickaelistria:
> My vision would be the following: In the settings there is a drop down (like currently with the dark/light theme). From there one can select a specific folder from which the icons should be taken.

I think how users interact with this feature is a next step. IMO it can be skipped for the moment so we can focus on the more critical underlying design. Once we have the ability to switch to another icon set, one way or the other, then we'll more easily figure out how far we want to push it in term of usability. But anyway, usability cannot be implemented until underlying strategy is available in JFace.

> would like to create a folder structure with a base folder /icons in which the folders of the icon themes

Sounds good.

> all eclipse icons are saved

Note that it's about improbable that all plugins change their icons dir. So the `icons/<setName>` folder should be queried first, but if no icon is found in there, then the "root" directory should be used as fail back.

> where to create the base folder?

The current proposal of a fragment is convenient with plugin URLs because from a plugin it's easy to infer the path of the plugin and the path of the icon to inject the `icons/<setName>` indirection in it.
But ideally, this would work for any kind of icon and any URL. But the more general case would be much harder to implement ( https://github.com/eclipse-platform/eclipse.platform.ui/pull/14/files#r971998765 ), hence -unless you have a string need for it- I suggest to sticking to icons with a plugin URL for the moment. Maybe later that could be improved, but I think it would be relatively difficult to find a solution that doesn't require to change all consumers; so it's likely that the balance of difficulty vs profit can be found just by supporting plugin URLs.

laeubi:
If we use a URL scheme than all kind of URLs should/could be supported, I don't see that a specific scheme is mandatory for this.

Anyways for me the most interesting part would be how to derive the theme id (in a platform independent way) and one should keep in mind that this can increase the load time of images a lot.

mickaelistria:
> I don't see that a specific scheme is mandatory for this.

Imagine `file:/path/to/a/folder/or/maybe/an/application/and/another/random/path/before/that/which/might/be/or/not/an/icon/folder/or/the/root/folder/of/icon/resources/icon.png`, how do you establish which URL you want to try for alternative icons? Try them all?
The benefit of more constrained schemes is that we know where is the root of resources.

> the most interesting part would be how to derive the theme id

I guess it would be more the other way round with the given proposal: there would be a named icon set, existing in the wild independently from the theme, and any Eclipse theme could declare to use that set.

laeubi:
> how do you establish which URL you want to try for alternative icons? Try them all?

As we do for other schemes as well, e.g currently we support

- `NxM` folders
- `@x<Z>` names

so for sure one can take your provided example here, extract the name part and try

`file:/path/to/a/folder/or/maybe/an/application/and/another/random/path/before/that/which/might/be/or/not/an/icon/folder/or/the/root/folder/of/icon/resources/<themeId>/icon.png`

or even 

`file:/path/to/a/folder/or/maybe/an/application/and/another/random/path/before/that/which/might/be/or/not/an/icon/folder/or/the/root/folder/of/icon/resources/<themeId>-icon.png`

or whatever, I think this is the most important part is should/must work with all kinds of URIs.

If you only want support for resources loaded from bundles, then a Resource Hook would be much better suited as I mentioned already somewhere before.

merks:
Of course there is already built-in support for `platform:/plugin/<bsn>(_<bundle-version>)?/`

image

Anyway, I'm not sure why we are discussing `file:` URLs instead of `platform:` or `bundleentry:` as are generally used to access resources within a bundle.  Maybe we digress.  😜 

Translation support has similar issues about looking things up at alternative locations and to which fragment can contribute "overrides"...

laeubi:
> Translation support has similar issues about looking things up at alternative locations and to which fragment can contribute "overrides"...

This is a different topic, because here we are talking about `URLImageDescriptor`... not about `BundleImageDescriptor`, or `EclipsePlatformImageDescriptor` or... as said if only bundle contents is to be assumed then better use a `ClassLoaderHook` and you can (dynamically) provide alternative resources without changing a single line of platform code.

mickaelistria:
>  then better use a ClassLoaderHook and you can (dynamically) provide alternative resources without changing a single line of platform code.

That does seem very interesting!

jannisCode:
Hi guys, i am currently working on this issue and encountered a problem with the bundleentry - URL´s. Some of the URL´s are not of the type ````platform:/plugin...```` but rather of the type ````bundleentry://301.fwk1538988266/icons/...````. 
My current approach is to modify the url so that it finds the icon in a specific (new) plugin or just finds it directly in the normal plugin in a folder with the icons from the theme (Both should work). Anyways I need to know the name of the plugin that holds the default icon. 
Right now I use ````FileLocator.toFileURL(url)```` to get the physical storage-place and then I try to find the plugins name from there(````bundleentry://301.fwk1538988266/icons/full/elcl16/javaassist_co.png```` is modified to ````platform:/plugin/IconsPlugin2/modern/org.eclipse.jdt.ui/icons/full/elcl16/javaassist_co.png```` ). This approach works, but it is not really a nice solution and for a few icons, that have a strange storage-place the solution does not work. 
Do you have any ideas how i could handle that problem?  Or have you noticed any problems with this solution? 

Thanks for the help!

laeubi:
> My current approach is to modify the url so that it finds the icon in a specific (new) plugin or just finds it directly in the normal plugin in a folder with the icons from the theme (Both should work). Anyways I need to know the name of the plugin that holds the default icon.

As mentioned already above, one can not make any assumptions about the actual URL, as it is a `URLImageDescriptor` and not an Icon, Bundle or whatever one.

> Right now I use FileLocator.toFileURL(url)

Also assume it can be converted to a file is plainly wrong, from my point of view as again there is no guarantee about what this actually return.

So I can only strongly suggest to only operate on the URL itself (as it is done for the `NxM` scheme and for the `@x<Z>`) if one want something that works in general and not only in some cases.

merks:
FYI, I think toFileURL would unzip and jar to some local location...

Michael5601:
> Also assume it can be converted to a file is plainly wrong, from my point of view as again there is no guarantee about what this actually return.
> 

What is the workflow if the `bundleentry://`-URL can't be converted to a path? 
The icon needs to be loaded from somewhere right? 

My experience with Eclipse icons was that they are either loaded via a `ImageFileNameProvider` directly with the file path that is provided by `DPIUtil.validateAndGetImagePathAtZoom()` or they are loaded via a `ImageDataProvider`. The process of the `ImageDataProvider` involves turning the `bundleentry://`-URL to a file path with `FileLocator.toFileURL(url)`.

laeubi:
> What is the workflow if the `bundleentry://`-URL can't be converted to a path?
> The icon needs to be loaded from somewhere right?

They are not converted to a path, they are just loaded form the `URL#openStream()` that is how URLs work, they are abstractions over an (unknown) transport.

> My experience with Eclipse icons was that they are either loaded via a ImageFileNameProvider directly with the file path that is provided by DPIUtil.validateAndGetImagePathAtZoom() or they are loaded via a ImageDataProvider. The process of the ImageDataProvider involves turning the bundleentry://-URL to a file path with FileLocator.toFileURL(url).

Just look at `org.eclipse.jface.resource.URLImageDescriptor.getImageData(String, int)` for example. It all operates on the Filename or path of the URL.

merks:
Note that both these schemes are hierarchical just like file: scheme is hierarchical so I would imagine you can do the same/similar things with the URIs as you are trying to do with the paths...

laeubi:
I'll try to setup an example using bundle transformer hooks the next days, this looks actually more useful than having even more URLs to check and it should even use for all kind of ways to access the icons.

Michael5601:
> They are not converted to a path, they are just loaded form the `URL#openStream()` that is how URLs work, they are abstractions over an (unknown) transport.
> 
Makes sense, thank you for the clarification. :)

laeubi:
I have now setup an example locally and technically we have everything we need already in platform but it could be improved so it works a little bit smoother for the case of an icon pack, the only thing we need here is some kind of documentation and an example. From that we can start creating icon packs and probably improve the process further.

The good thing about this is that it is totally flexible regarding the icons to replace and the layout to be used, does not require "special" things like fragments and one can choose to replace icons for one bundle or even all bundles.

The only thing is that when one installs multiple packs that try to override the same icons then of course only one can win, but that's more a deployment issue.

mickaelistria:
> the only thing we need here is some kind of documentation and an example.

Ideally, we could start with some test cases. Using an eg blue square icon by default in an ImageDescriptor and testing it to be overridden as a red icon by a theme could be a relatively easy way to test we get the right image.

Do I get it right that your proposal then would be purely available for URLs from Equinox, and not for general SWT then?  (note that I don't think it's an issue, having theming implemented in Platform rather than in SWT makes at least as much sense as the other way round)

laeubi:
It has nothing to with `ImageDescriptor` or `SWT` at all and will work for `BufferedImage` or whatever technique is used as it replaces the resource on the bundle level by an alternative one.

So it won't work if you use SWT/JFace in a plain setup without OSGi, but as there you have a flat classpath it is even simpler: Just make sure your icon pack appears before the one you want to replace an image.

jannisCode:
@laeubi thanks for your support with the two issues! Because i still am pretty new to eclipse, and don´t have as much knowledge as you guys, maybe someone could explain to me how I can use @laeubi ´s issues to reach my goal of a working icon-theming. 
Thanks in advance!

BeckerWdf:
> I have now setup an example locally and technically we have everything we need already in platform but it could be improved so it works a little bit smoother for the case of an icon pack, the only thing we need here is some kind of documentation and an example. From that we can start creating icon packs and probably improve the process further.

Thanks for that. I don't have the technical knowledge the really understand your implementation on a technical level.
 
> The good thing about this is that it is totally flexible regarding the icons to replace and the layout to be used, does not require "special" things like fragments and one can choose to replace icons for one bundle or even all bundles.

For the "Icon Packs"  I see two use-cases:

1. You are delivering a RCP product and you want to replace Platform's icons with your own company styled icons. So you need to ship a plug-in or an fragment e.g. called "com.acme.iconpack" that replaces icons e.g. from the "org.eclipse.ui" plugin

2. The Eclipse Project delivers multiple icon packs (e.g. exiting "classic" icons and additional "monochrome" icons). You are extending the Eclipse IDE with some bundles that provide editing support for your own fancy programming language. You also want to deliver icons that contribute to the Platform's "monochrome" icon pack inside your bundles.
For example in your "com.acme.fancyEditor" plugin you have existing icons in the normal
```
/icons
  /etool16
    log.png
    log@2x.png
  /obj16
    fancyType.png
    fancyType@2x.png
```
folder structure. And you want to simply add new monochrom versions of "log" and "fancyType" directly inside the com.acme.fancyEditor" plugin.

> The only thing is that when one installs multiple packs that try to override the same icons then of course only one can win, but that's more a deployment issue.

From the user's perspective the icon pack should be something he selects in the preference dialog. For example a new setting next to the "Theme" drop down box on the "Appearance" preference page. This drop down box would list all available icon packs and the user can select which one to use. So only one icon pack should be "active" at a time.

Can you explain e.g. based on my example if and how these use-cases can be achieved? What does an plugin-developer do to leverage this feature?

laeubi:
I'll prepare an example how this can be used once the changes are available but don't wanted to rush them in due to holiday seasons and give people some more time to possibly give feedback. 

In any case things like configurability / UI is something that has to be provided on top of this and we might want/need to further enhance the "core" for different use-case, but both of what you have described is possible.

HeikoKlare:
> From the user's perspective the icon pack should be something he selects in the preference dialog. For example a new setting next to the "Theme" drop down box on the "Appearance" preference page. This drop down box would list all available icon packs and the user can select which one to use. So only one icon pack should be "active" at a time.

I want to share/refresh one conclusion we had at last year's Eclipse IDE community day at SAP when discussing the icon exchange/pack/theme topic: Due to the general independence of plugins, being provided as extensions to a product like the IDE, it is unlikely to achieve one consistent set of new icons under the same icon pack/theme ID (or even multiple ones, if new types of icons are provided). If we consider "icon packs" as complete sets of icons, all provided for the same ID of an icon pack, this probably only work fine for the product use case, where the product delivers a complete and consistent icon pack on its own, but not when considering an extensible platform.

Last year, we thus agreed on following the approach of making icon packs modular and choosing multiple of them to form something like an icon theme. For example, the Eclipse platform may deliver a "flat" and "monochrome" icon pack. JDT does the same. Some extension provider only provides a "flat" pack. Now you can select those packs you want to combine to your theme. One reasonable theme would be the combination of all "flat" icon packs. But given that the extension provider has no "monochrome" pack so far, you could combine its "flat" pack with the other "monochrom" ones, rather than using the default (non-flat) icons, which do integrate with monochrome icons even worse, when just allowing to contribute to a complete icon theme.

In addition, icon packs may not be mutually exclusive, as an icon pack may want to override an icon that is also provided by another icon pack, like the one with the basic icons for the platform. This may be resolved by defining a precedence order for the icon packs.

image

And to make it easier to configure, you may still think of a concept of "themes" that have a default configuration for specific icon packs or icon pack IDs, but still give users the flexibility to insert the desired icon pack with icons of an extensions for which no "properly matching" icon pack is provided.

laeubi:
Yes multiple packs can work together, one pack might even be able to override the ones from another, but as mentioned these are more "organizational" things, e.g. m2e allows to install so called "connectors", the same is possible for team providers (SVN, GIT) so maybe that would be a better approach than a plain preference.

I'll leave that to the interested parties after the technical thing is available :-)

BeckerWdf:
> I'll leave that to the interested parties after the technical thing is available :-)

But how can we make "technical things" available "before" we agreed on what we want to offer to the user?

laeubi:
> > I'll leave that to the interested parties after the technical thing is available :-)
> 
> But how can we make "technical things" available "before" we agreed on what we want to offer to the user?

If the requirement is that we can replace icons by alternative ones through other bundles without the help of the original bundle author, then this can be solved independent on how these "icon bundles" later being installed or how many of them there are and how the UI looks like.

And as @HeikoKlare correctly summarized, as Eclipse is extensible there is actually no way to have "one" icon pack that solves that all... maybe there will be one that supports many famous extensions but it can't support everything.

This is not necessarily a problem, as still a dedicated product or EPP on the other hand still can benefit from such a feature.

mickaelistria:
Do I get it right that with the Transformer recently merged in https://github.com/eclipse-equinox/equinox/pull/732 , then any RCP product can now just register such a transformer service and implement the strategy they want to override image loading?
If we already have it, then I think the remaining work would mostly become creating some documentation or example or test (https://github.com/eclipse-platform/eclipse.platform.ui/pull/14#issuecomment-2558061601) for it so we ensure it works and people have a starting point to use it, and then we can IMO consider this issue resolved.
Later, when we want to support multiple icon sets in the SDK, we'll be discussing the best implementation for the transformer, but this seems to me like a next step/issue.

laeubi:
> Do I get it right that with the Transformer recently merged in https://github.com/eclipse-equinox/equinox/pull/732 , then any RCP product can now just register such a transformer service and implement the strategy they want to override image loading?

It could be done before, but now a bit more convenient ;-)

> If we already have it, then I think the remaining work would mostly become creating some documentation or example or test

I'll prepare an example the next days and add it here:

https://github.com/eclipse-platform/eclipse.platform.ui/tree/master/examples

from this example it should be easier to decide the next steps.

vogella:
@@Michael5601 did you already look at the Transformer from https://github.com/eclipse-equinox/equinox/pull/732. I think this could be used to provide your alternative icon sets.

Michael5601:
> @@Michael5601 did you already look at the Transformer from [eclipse-equinox/equinox#732](https://github.com/eclipse-equinox/equinox/pull/732). I think this could be used to provide your alternative icon sets.

Thanks for the hint, I did not yet look into it. But I am also looking forward to a feature as discussed in this PR where you can choose your icon theme in the Eclipse UI. Maybe if I am in search of a new project in the future I will try to push this forward.

vogella:
> > @@Michael5601 did you already look at the Transformer from [eclipse-equinox/equinox#732](https://github.com/eclipse-equinox/equinox/pull/732). I think this could be used to provide your alternative icon sets.
> 
> Thanks for the hint, I did not yet look into it. But I am also looking forward to a feature as discussed in this PR where you can choose your icon theme in the Eclipse UI. Maybe if I am in search of a new project in the future I will try to push this forward.

I think adding a new selection box to the preference page to select the icon set would be relatively easy once we have actually a way to select a different icon set. I assume once @laeubi publishes his example how to use this feature as indicated in https://github.com/eclipse-platform/eclipse.platform.ui/pull/14#issuecomment-2572769292 we can see how to implement the functionality.

laeubi:
I don't think it will be easy (in the sense of less efforts), and I doubt that actually a preference is needed at all that why I suggested to use the [P2 Discovery](https://wiki.eclipse.org/Equinox/p2/Discovery) feature as it allows user to install (or uninstall) whatever they want, so no need to mess with the preferences.

But as said I would think of this to be a completely different step/decision/development process (and might demand for further changes to the example/supporting technique).

laeubi:
The example is now here:

https://github.com/eclipse-platform/eclipse.platform.ui/tree/master/examples/org.eclipse.ui.examples.iconpack

I think we can close this for now and continue on individual issues / prs for the matter itself, if the approach taken here should be picket up again we can easily reopen or cherry pick it to a new change.

BeckerWdf:
> I don't think it will be easy (in the sense of less efforts), and I doubt that actually a preference is needed at all that why I suggested to use the [P2 Discovery](https://wiki.eclipse.org/Equinox/p2/Discovery) feature as it allows user to install (or uninstall) whatever they want, so no need to mess with the preferences.

I would also think that user's want to switch Icon-Packs at runtime. So I would also expect to have a preference for the user to control this.

laeubi:
> I would also think that user's want to switch Icon-Packs at runtime.

Its always hard to think what users want but changing icons *at runtime* is extremely hard compared to the benefit (no restart), e.g. one would require to trigger a reload of all images in all swt components and one not necessarily even knows that one image was created from another (e.g. overlays) so I highly doubt this would ever be implemented.

BeckerWdf:
> changing icons _at runtime_ is extremely hard compared to the benefit (no restart), e.g. one would require to trigger a reload of all images in all swt components and one not necessarily even knows that one image was created from another (e.g. overlays) so I highly doubt this would ever be implemented.

it would be fine for me to force a re-start of the application similar to what we do when switching CSS themes.

Issue #8, opened 2022-04-12
Prevent JobManager UI freeze when waiting for a rule in the UI Thread


mickaelistria:
This is a followup of discussions in https://github.com/eclipse-platform/eclipse.platform.ui/pull/6 and former Gerrit patch.

The general problem is that when one wants to run a job or some other operation protected by scheduling rules in UI Thread, and this scheduling rule is currently "blocked" by some other work, then the result is a UI Freeze.
We could do better by allowing the workbench/IDE parts to customize a bit the JobManager, more specifically the `beginRule` so it would show the interesting ProgressDialog when called from the UI Thread, so it doesn't seem too much like a freeze, and users have the opportunity to workaround it. It would also ease debugging.
Such a solution would allow to serve multiple operations that have implemented workarounds for this freeze, such as https://github.com/eclipse-platform/eclipse.platform.ui/pull/6 , and those workarounds could then be removed.

laeubi:
I don't think we need more JobManager Customizations here, the underlying problem is more that in some circumstances the UI-Events are not driven, see 

- https://github.com/eclipse-platform/eclipse.platform.runtime/issues/40 

for an analysis and fix.What is left is to just need to override these new methods .

mickaelistria:
#40 is promising, I suggest we follow that track first and then we'll evaluate if we needs further smartness here and there.

PR #18, opened 2022-04-19
Bug 578871 - use ISchedulableOperation

jukzi:
to prevent UI freeze during UNDO

needs
eclipse-platform/eclipse.platform.runtime#30

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
@@ -26,10 +26,15 @@ import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.jobs.IJobManager;
+import org.eclipse.core.runtime.jobs.ISchedulableOperation;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.LegacyActionTools;
 import org.eclipse.jface.operation.IRunnableWithProgress;
 import org.eclipse.osgi.util.NLS;
+import org.eclipse.swt.custom.BusyIndicator;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IPartListener;
@@ -298,7 +303,32 @@ public abstract class OperationHistoryActionHandler extends Action
 			if (getOperation() instanceof IAdvancedUndoableOperation2) {
 				runInBackground = ((IAdvancedUndoableOperation2) getOperation()).runInBackground();
 			}
-			progressDialog.run(runInBackground, true, runnable);
+			if (runInBackground) {
+				progressDialog.run(runInBackground, true, runnable);
+			} else {
+				final IJobManager manager = Job.getJobManager();
+				final ISchedulingRule rule = (getOperation() instanceof ISchedulableOperation)
+						? ((ISchedulableOperation) getOperation()).getSchedulingRule()
+						: null;
+				if (rule == null) {
+					progressDialog.run(runInBackground, true, runnable);
+				} else {
+					// prevent UI freeze during AutoBuild as in
+					// org.eclipse.jdt.internal.ui.refactoring.RefactoringExecutionHelper#perform
+					try {
+						try {
+							// interrupt autobuild or show Wait/Cancel Dialog:
+							Runnable r = () -> manager.beginRule(rule, null);
+							BusyIndicator.showWhile(parent.getDisplay(), r);
+						} catch (OperationCanceledException e) {
+							throw new InterruptedException(e.getMessage());
+						}
+						progressDialog.run(runInBackground, true, runnable); // <-- actual work
+					} finally {
+						manager.endRule(rule);
+					}
+				}
+			}
 		} catch (InvocationTargetException e) {
 			Throwable t = e.getTargetException();
 			if (t == null) {
--- a/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
@@ -95,7 +95,7 @@ Export-Package: org.eclipse.e4.ui.workbench.addons.perspectiveswitcher;x-interna
  org.eclipse.ui.themes,
  org.eclipse.ui.views,
  org.eclipse.ui.wizards
-Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.19.0,4.0.0)",
+Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.25.0,4.0.0)",
  org.eclipse.help;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.jface;bundle-version="[3.18.0,4.0.0)",
  org.eclipse.swt;bundle-version="[3.107.0,4.0.0)",


mickaelistria:
That seems much safer, and actually even closer to a generic solution that we could implement in #8 or similar more general proposal !

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
```diff
@@ -298,7 +303,32 @@ public final void run() {
...
+				if (rule == null) {
+					progressDialog.run(runInBackground, true, runnable);
+				} else {
+					// prevent UI freeze during AutoBuild as in
+					// org.eclipse.jdt.internal.ui.refactoring.RefactoringExecutionHelper#perform
+					try {
+						try {
+							// interrupt autobuild or show Wait/Cancel Dialog:
+							Runnable r = () -> manager.beginRule(rule, null);
+							BusyIndicator.showWhile(parent.getDisplay(), r);
```
iloveeclipse, line 322: @laeubi : you've mentioned on https://github.com/eclipse-platform/eclipse.platform.ui/pull/16#issuecomment-1102520698
>  Anyways I think the problem is not necessarily a lock here but the BusyIndicator and progressDialog.run both need to be handled with care and might not behave as expsected as when not running in the background require that the caller make sure to drive the eventqueue!

Can you please review **this** change?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
```diff
@@ -298,7 +303,32 @@ public final void run() {
 			if (getOperation() instanceof IAdvancedUndoableOperation2) {
 				runInBackground = ((IAdvancedUndoableOperation2) getOperation()).runInBackground();
 			}
-			progressDialog.run(runInBackground, true, runnable);
+			if (runInBackground) {
+				progressDialog.run(runInBackground, true, runnable);
+			} else {
+				final IJobManager manager = Job.getJobManager();
+				final ISchedulingRule rule = (getOperation() instanceof ISchedulableOperation)
```
iloveeclipse, line 310: I wonder if that can work as proposed. 

An undo operation may want to perform different tasks and acquire multiple rules for them, so first acquire rule on one resource and later on another etc.

Another reason I'm asking is: an **existing** "generic" operation that can be extended can't use this new 
`ISchedulableOperation` interface to force the code here use the rule, because **if** it can be extended, the 3rd party code may don't want it to happen, so it will break existing code again.

May be it would be good to see what is the actual change planned on the client side (in JDT? LTK?).

jukzi: see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/19
it's not a secret.

iloveeclipse: > see [eclipse-jdt/eclipse.jdt.ui#19](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/19) it's not a secret.

Thanks, I will look into. Would be easier for all if that would be mentioned in this PR somewhere.

jukzi: It was autolinked and for sure you got a notification about that. Probably you just overlooked it.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
laeubi: @iloveeclipse sadly it is not documented well but the [snippet for the BusyIndicator](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet130.java) shows that:

1. You should spin a new thread to actually do the really long running work
2. busy-loop to check if that spinned thread is done and additionally drive the event queue in the runnable that actually is passed to the BusyIndicator

otherwise you will see the busy-cursor but no event will be processed what could then be reported as a "this application seems to be blocked" and of course any Display.async/sync calls to block / deadlock.

I don't know if the manager.begin rule do all of that internally.

Because of this complex and setup I once proposed a change to SWT for that I think but it was not included so I'm using a helper method for this in my own code that uses CompleteableFutures to help with that.
If we thing its now a better time for such a change I can try to continue my past work on that but thi will require some additions in SWT first.

mickaelistria: `JobManager.beginRule()` has no understanding of UI at all, and no hook to make UI react to rule changes; that's one reason why such UI freeze because of scheduling rules can happen. As I suggest in https://github.com/eclipse-platform/eclipse.platform.ui/issues/8 , some way to hook more UI friendliness in JobManager would allow to automatically show a progressDialog or a busyCursor in case a rule is requested in UI Thread and isn't available immediately.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
```diff
@@ -298,7 +303,32 @@ public final void run() {
...
+					// org.eclipse.jdt.internal.ui.refactoring.RefactoringExecutionHelper#perform
+					try {
+						try {
+							// interrupt autobuild or show Wait/Cancel Dialog:
+							Runnable r = () -> manager.beginRule(rule, null);
+							BusyIndicator.showWhile(parent.getDisplay(), r);
+						} catch (OperationCanceledException e) {
+							throw new InterruptedException(e.getMessage());
+						}
+						progressDialog.run(runInBackground, true, runnable); // <-- actual work
```
laeubi, line 326: From the javadoc:

> It is recommended that <code>fork</code> is set **to true in
> most cases.** If <code>fork</code> is set to <code>false</code>, the
> runnable will run in the UI thread and **it is the runnable's
> responsibility to call <code>Display.readAndDispatch()</code>** to ensure
> UI responsiveness.

so does the passed runable drives the event queue if runInBackround = false is passed in all cases?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
laeubi: It might or might not be a problem if the begin rule is fast, but then a BusyIndicator seems not valid to use, also the comment seems confusing, as a busy indicator can't be used for cancel ...

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
jukzi: The 
`if (runInBackground) {..} else`
ensures runInBackground==false here

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
jukzi: I did not follow this conversation: do you request any change in this piece of code? Its copied pattern from RefactoringExecutionHelper#perform. Its tested, It's working. - can we resolve this?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
laeubi: if `runInBackground` is always `true` here ou should use a constant to prevent confusion.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
laeubi: Just because it is used else where do not mean it correct, please check the `BusyIndicator` implementation, there are two cases:

1. You are in the UI thread then the call will **block the UI** until the runnable is done (even though showing a busy cursor) unless you drive the event queue.
2. You are not in the UI thread and then the runnable is just running (still blocking) without any busy cursor.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
jukzi: It's far better reading to pass (constant) variables then to call something(false, true,...) because it's not obvious without reading the called method what those values are used for.

laeubi: You can enable code-minings if you want.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
jukzi: and what do you suggest from that? we are in UI and it is blocking. but its not freezing anymore and its showing a cancel dialog.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
jukzi: i don't want. i tried it. they are slow and move my code while i am reading :-(
i do like the idea though. sad it's not working good.

laeubi: still if it is always true you should use a constant here, you can of course extract ti to a constant field if you think its more readable.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/operations/OperationHistoryActionHandler.java
laeubi: BusyIndicator has nothing to do with cancel or not freezing, BusyIndicator **will freeze** the UI if not used appropriate (see the linked code snippet on how it is supposed to be used).

jukzi: don't trust a snippet but actually try it!

laeubi: Maybe you want to try it out yourself :-)

Just set a breakpoint at `manager.beginRule(rule, null)` you will see that there is a busy cursor and your Application will be marked as "Non responsive", and no cancel option.

But well no problem if you think its fine, it was just requested by @iloveeclipse to give a hint about a possible dead. lock.

jukzi: please read the "how to reproduce" in https://bugs.eclipse.org/bugs/show_bug.cgi?id=578871#c0
it's not about putting a breakpoint in the UI thread but about putting a breakpoint in the autobuild thread. of cause the UI thread needs to run in order to show a dialog.

laeubi: Well there are two cases:

1. `manager.beginRule(rule, null)` is always fast then there is no point in using `BusyIndicator`
2. `manager.beginRule(rule, null)` could sometimes be slow (e.g waiting for a lock from another thread) then this will freeze the UI, this has nothing to do with whatever has to be fixed/reproduced with this.

jukzi: test it!

laeubi: I expressed my concerns and give you hints where blocking occurs, if you think it is invalid I'm fine with that.

jukzi:
3 unrelated errors:

```
java.lang.AssertionError: 
Warning: Button {Synchronize &scrolling between panes in compare viewers}
	Actual Width -> 0
	Recommended Width -> 357
	at org.eclipse.ui.tests.compare.UIComparePreferencesAuto.testCompareViewersPref(UIComparePreferencesAuto.java:57)

```

```
org.eclipse.ui.tests.dialogs.DeprecatedUIPreferencesAuto.testFieldEditorEnablePref
	at org.eclipse.ui.tests.dialogs.DeprecatedUIPreferencesAuto.testFieldEditorEnablePref(DeprecatedUIPreferencesAuto.java:133)
	
```
```
org.eclipse.ui.tests.dialogs.DeprecatedUIPreferencesAuto.testFontEditorsPref
	at org.eclipse.ui.tests.dialogs.DeprecatedUIPreferencesAuto.getPreferenceDialog(DeprecatedUIPreferencesAuto.java:44)
	at org.eclipse.ui.tests.dialogs.DeprecatedUIPreferencesAuto.testFontEditorsPref(DeprecatedUIPreferencesAuto.java:91)
```

iloveeclipse:
Regarding unrelated errors: https://github.com/eclipse-platform/eclipse.platform.ui/issues/30

jukzi:
2 unrelated errors:
org.eclipse.ui.internal.monitoring.EventLoopMonitorThreadTests.testSeeLongEventInTheMiddleOfNestedCalls
org.eclipse.ui.internal.monitoring.EventLoopMonitorThreadTests.testDoublyNestedEventLogging

A freeze event didn't capture a good range of stack samples (462) expected:<5> but was:<1>
A freeze event didn't capture a good range of stack samples (660) expected:<5> but was:<1>

PR #27, opened 2022-04-21
Use PreferenceManager.getElements type info

akurtakov:
It's properly generified for long time so no need for all these manual
casts.


akurtakov:
Same 3 tests fail in nightly build so not caused by this commit. See https://download.eclipse.org/eclipse/downloads/drops4/I20220420-1800/testresults/html/org.eclipse.ui.tests_ep424I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html
Actually, these fixes made me look into these tests and this cleanup is the first outcome.

iloveeclipse:
Regarding unrelated errors: https://github.com/eclipse-platform/eclipse.platform.ui/issues/30

PR #28, opened 2022-04-21
Bug 485835 - Deletion of IWorkbenchWindowConfigurator methods

vogella:
Original marked for deletion for March 2021.


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbench.java
```diff
@@ -473,21 +472,6 @@ IWorkbenchPage showPerspective(String perspectiveId, IWorkbenchWindow window, IA
...
-	 * @return an interface to manage commands at the workbench level. Guaranteed
-	 *         not to be <code>null</code>.
-	 * @since 3.0
-	 * @deprecated Please use {@link IServiceLocator#getService(Class)} instead.
-	 *             This API is scheduled for deletion, see Bug 431177 for details
-	 * @see ICommandService
-	 * @see IHandlerService
-	 * @noreference IWorkbenchCommandSupport is scheduled for deletion.
-	 */
-	@Deprecated
```
jukzi, line 488: i am not using that API, so i dont have a problem. But maybe nobody read the javadoc "This API is scheduled for deletion, see Bug 431177 for details" it would be better to use
`@Deprecated(forRemoval=true)` in he future

laeubi: > @Deprecated(forRemoval=true)

That was added with Java 9, so probably at that time platform where still using java 8:
https://docs.oracle.com/javase/9/core/enhanced-deprecation1.htm

vogella: @laeubi correct, most API deletions were before @deprecated(forRemoval=true) was availabler in platform. New API removals should use forRemoval and if someone has time and desire for that, he / she could also add this @deprecated(forRemoval=true) to the API already scheduled for deletion.  In general the current process of API deletion is relative complex which the update of the API removal document in the commons repo.

vogella:
This was Window... removal with another commit and I currently do not have time to finsi[org.eclipse.ui.commands package](https://github.com/eclipse-platform/eclipse.platform.ui/pull/28/commits/44af2e0bebcdf729b6d1d042c0c34af539d512f4) removal, so closing

PR #29, opened 2022-04-21
Bug 570471 - Add utilities for perspectives persistence and migration

eneufeld:
Add new utilities which allow to
- persist and restore perspectives in e4 applications including the
part's states
- migrate 3.x perspectives to e4 applications
The code is split in two bundles, one contributing the persist and
restore functionality, the other contributing the 3.x to e4 perspective
migration.

Signed-off-by: Eugen Neufeld <eneufeld@eclipsesource.com>


eneufeld:
The build fails due to the apiBaseline check which can't succeed as this PR adds new bundles. Is there a trick?

laeubi:
Are you sure? I think new bundles should be fine. Currently it fails due to test errors.

eneufeld:
Yes pretty sure:
`[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-eclipserun-plugin:3.0.0-SNAPSHOT:eclipse-run (api-analysis) on project org.eclipse.e4.ui.workbench.persistence: Error while executing platform: Error while executing platform: return code=4, see content of /home/jenkins/agent/workspace/eclipse.platform.ui_PR-29/apiAnalyzer-workspace/data/.metadata/.logfor more details. -> [Help 1]` 
If I build without `-Papi-check` locally the build succeeds (except the known tests)

iloveeclipse:
the pom should have a property to disable API check, don't ask me how it is called, may be it's documented in tycho somewhere.

eneufeld:
@iloveeclipse  so I would disable the check for now and after it is merged and part of a release reenable it again?
I will check the maven/tycho documentation, thank you for the hint

iloveeclipse:
Exact, create bug as reminder for next release.

laeubi:
> I will check the maven/tycho documentation, thank you for the hint

Sometimes the solution is just near by you already have found it:

> If I build without `-Papi-check` locally the build succeeds

so the property is most probably defined in that `api-check` profile :-)

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
```diff
@@ -0,0 +1,933 @@
...
+import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.emf.ecore.resource.impl.ResourceSetImpl;
+import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.ui.IMemento;
+import org.eclipse.ui.WorkbenchException;
+import org.eclipse.ui.XMLMemento;
+import org.eclipse.ui.internal.e4.compatibility.ModeledPageLayoutUtils;
+
+/**
+ * The Implementation of the Converter interface.
```
mickaelistria, line 130: This is not sufficient documentation for an API, moreover it seems incorrect as there s no interface.

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
```diff
@@ -0,0 +1,933 @@
...
+ *
+ * @since 3.3
```
mickaelistria, line 132:
```suggestion
 * @since 1.0
```

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
```diff
@@ -0,0 +1,933 @@
...
+
+	private static final String EDITOR_AREA_ID = "org.eclipse.ui.editorss"; //$NON-NLS-1$
+
+	private final Map<Object, Set<String>> visited = new LinkedHashMap<>();
+
+	/**
+	 * Constructor.
+	 */
+	public Converter() {
+	}
```
mickaelistria, line 149: Remove.

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
```diff
@@ -0,0 +1,933 @@
...
+import org.eclipse.ui.internal.e4.compatibility.ModeledPageLayoutUtils;
+
+/**
+ * The Implementation of the Converter interface.
+ *
+ * @since 3.3
+ *
+ */
+@SuppressWarnings("restriction")
+public class Converter {
```
mickaelistria, line 136: Name is a bit weak, maybe something more explicit `E4MigrationTool` ?

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
```diff
@@ -0,0 +1,933 @@
...
+
+	/** Element id for the primary editor stack in an e4 application. */
+	private static final String PRIMARY_EDITOR_PART_STACK_ID = "org.eclipse.e4.primaryDataStack"; //$NON-NLS-1$
+
+	private static final String EDITOR_AREA_ID = "org.eclipse.ui.editorss"; //$NON-NLS-1$
```
mickaelistria, line 141: 2 `s` here?

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
```diff
@@ -0,0 +1,933 @@
...
+	/**
+	 * Allows to convert an 3.x perspective to a serialized e4 applciation.
+	 *
+	 * @param sourcePath The path to the 3.x serialized perspective.
+	 * @param targetPath The path to the file to serialize the e4 application to.
+	 * @throws IOException           When reading/writing the files fails.
+	 * @throws FileNotFoundException When the files do not exist.
+	 * @throws WorkbenchException    When reading the 3.x perspective fails
+	 */
+	public static void convert(final String sourcePath, final String targetPath)
```
mickaelistria, line 160: `convert` is also a bit imprecise here, particularly with the relatively basic types being used. Something like `convertToE4Perspective` would be more explicit. Also if the arguments are Path, then let's use File or Path, not strings; and the arguments could tell `sourceMementoPath` and `targetMApplicationPath`

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
mickaelistria: Or maybe this class is not intended to be API and it could be package visible? In such case, we can care much less about documentation and API names.

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/PerspectiveMigrator.java
```diff
@@ -0,0 +1,115 @@
...
+import org.eclipse.e4.ui.workbench.internal.persistence.IWorkbenchState;
+import org.eclipse.e4.ui.workbench.internal.persistence.impl.WorkbenchState;
+import org.eclipse.e4.ui.workbench.modeling.EModelService;
+import org.eclipse.e4.ui.workbench.persistence.PerspectivePersister;
+import org.eclipse.e4.ui.workbench.persistence.common.CommonUtil;
+import org.eclipse.ui.IMemento;
+import org.eclipse.ui.internal.IWorkbenchConstants;
+
+/**
+ * The PerspectiveMigrator implementation.
```
mickaelistria, line 29: More elaborate description please.

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/PerspectiveMigrator.java
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * @since 3.3
```
mickaelistria, line 31:
```suggestion
 * @since 1.0
```

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/PerspectiveMigrator.java
```diff
@@ -0,0 +1,115 @@
...
+		return false;
+	}
+
+	/**
+	 * Creates the persistable workbench state for a legacy (3x) perspective.
+	 *
+	 * @param perspective The MPerspective to create 3.x the WorkbenchState for
+	 * @return The created {@link WorkbenchState}
+	 */
+	public static IWorkbenchState createLegacyWorkbenchState(final MPerspective perspective) {
```
mickaelistria, line 112: Why `create` here and not `convert` like everywhere else?

bundles/org.eclipse.e4.ui.workbench.persistence/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,23 @@
...
+ org.eclipse.e4.ui.model.workbench;bundle-version="[2.1.0,3.0.0)";visibility:=reexport,
+ org.eclipse.e4.core.contexts;bundle-version="[1.8.0,2.0.0)",
+ org.eclipse.e4.ui.workbench;bundle-version="[1.10.0,2.0.0)",
+ org.eclipse.e4.ui.workbench.addons.swt;bundle-version="[1.2.0,2.0.0)",
+ org.eclipse.ui;bundle-version="[3.110.0,4.0.0)"
+Export-Package: org.eclipse.e4.ui.workbench.internal.persistence,
+ org.eclipse.e4.ui.workbench.internal.persistence.impl;x-internal:=true,
+ org.eclipse.e4.ui.workbench.internal.persistence.util;x-internal:=true,
+ org.eclipse.e4.ui.workbench.persistence,
+ org.eclipse.e4.ui.workbench.persistence.common;x-internal:=true,
```
mickaelistria, line 22: So the package name should contain "internal" as well to make things clearer.

bundles/org.eclipse.e4.ui.workbench.persistence/src/org/eclipse/e4/ui/workbench/persistence/common/CommonUtil.java
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	/**
+	 * The key to use for storing the memento.
+	 */
+	public static String MEMENTO_KEY = "memento"; //$NON-NLS-1$
+	/**
+	 * Constant for the EDITOR_AREA.
+	 */
+	public static String EDITOR_AREA = "org.eclipse.ui.editorss"; //$NON-NLS-1$
```
mickaelistria, line 63: 2 `s` here as well?

bundles/org.eclipse.e4.ui.workbench.persistence/src/org/eclipse/e4/ui/workbench/persistence/common/CommonUtil.java
```diff
@@ -0,0 +1,222 @@
...
+import org.eclipse.ui.XMLMemento;
+import org.eclipse.ui.internal.EditorReference;
+import org.eclipse.ui.internal.Workbench;
+import org.eclipse.ui.internal.e4.compatibility.CompatibilityEditor;
+import org.eclipse.ui.internal.e4.compatibility.CompatibilityView;
+
+/**
+ * A Util for the workbench conversion.
+ *
+ * @since 3.3
```
mickaelistria, line 47:
```suggestion
 * @since 1.0
```

bundles/org.eclipse.e4.ui.workbench.persistence/src/org/eclipse/e4/ui/workbench/persistence/common/CommonUtil.java
```diff
@@ -0,0 +1,222 @@
...
+ *
+ */
+@SuppressWarnings({ "restriction" })
+public final class CommonUtil {
```
mickaelistria, line 51: Couldn't it be internal?

bundles/org.eclipse.e4.ui.workbench.persistence/src/org/eclipse/e4/ui/workbench/persistence/impl/Util.java
```diff
@@ -0,0 +1,138 @@
...
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     EclipseSource GmbH - initial API and implementation
+ ******************************************************************************/
+package org.eclipse.e4.ui.workbench.persistence.impl;
```
mickaelistria, line 14: Please move it to internal.

mickaelistria (changes requested):
Additionally to the code nitpicking inline, some comments/questions to ensure we're going to the best possible outcome we can imagine:
* Do you think it would fit better in the eclipse.platform.ui.tools repository, so it could be included in the tools feature? 
* Isn't there some other bundle that could host this code, eg https://github.com/eclipse-platform/eclipse.platform.ui.tools/tree/master/bundles/org.eclipse.e4.tools.compat which could be a good candidate to host the converter
* What typical use-case do you see for the PerspectivePersister? Wouldn't it be fine to also store it in org.eclipse.e4.tools.compat ?

bundles/org.eclipse.e4.ui.workbench.compatibility.migration/src/org/eclipse/e4/ui/workbench/compatibiliy/migration/Converter.java
eneufeld: Pretty sure this is the magic id of the editor area used in e4.

eneufeld:
> Additionally to the code nitpicking inline, some comments/questions to ensure we're going to the best possible outcome we can imagine:
> 
> * Do you think it would fit better in the eclipse.platform.ui.tools repository, so it could be included in the tools feature?
> * Isn't there some other bundle that could host this code, eg https://github.com/eclipse-platform/eclipse.platform.ui.tools/tree/master/bundles/org.eclipse.e4.tools.compat which could be a good candidate to host the converter
> * What typical use-case do you see for the PerspectivePersister? Wouldn't it be fine to also store it in org.eclipse.e4.tools.compat ?

Thank you for the review.
My current understanding of the tools feature is, that it is rather installed into the IDE than used as part of the target platform. If my understanding is wrong, then the we can move this to the tools feature.
The main use cases I envision for these features are the following:
- the PerspectivePersister provides the old 3.x feature which allowed to store a perspective and share it with others or use it in another 3.x application. 
- the PerspectiveMigrator allows to reuse old 3.x perspectives in e3 applications and that is something that might make sense to be used in a running application and not only in the IDE
These two use cases are the reason I added these to the platform.ui.

mickaelistria:
> My current understanding of the tools feature is, that it is rather installed into the IDE than used as part of the target platform.

I think it can be both.
But do I get it right that this is aimed to be used as an API only? Wouldn't it be worth to try a few step forward and implement some Import/Export perspective wizards and make the code all internal? I think if all the code becomes internal, it becomes easier to integrate it into some existing bundle or feature.

eneufeld:
Then the https://github.com/eclipse-platform/eclipse.platform.ui.tools/tree/master/bundles/org.eclipse.e4.tools.compat might be a good place.
Yes our idea was plain API. I see your point about wizards but this all needs to be compatible with 3.x and e4. Especially e4 applications that I saw are often very custom. Adding an action to e4 is very simple and it would be straight forwards to call the provided API from it.

I would abandon this PR and open a new one for the tools repo.

PR #31, opened 2022-04-22
Add CTabFolder support to tests

jonahgraham:
The method verifyCompositeText in DialogCheck verifies various aspects
of composites, with
https://github.com/eclipse-platform/eclipse.platform.team/pull/14 this
now needs to understand CTabFolder (in addition to TabFolder)

This fixes UIComparePreferencesAuto.testCompareViewersPref failures reported in #30

PR #34, opened 2022-04-22
Like #33, but but with #31 reverted

jonahgraham:
This is to provide some additional test results.


jonahgraham:
Not really ready for review, but Jenkins seems to only run non-draft.

PR #36, opened 2022-04-22
Fixes #30: Close the dialog if test fails

### LINKED ISSUE #30: 3 UI tests fail since I20220420-1800
See https://download.eclipse.org/eclipse/downloads/drops4/I20220420-1800/testresults/html/org.eclipse.ui.tests_ep424I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html

On all platforms they are failing now:

- UIComparePreferencesAuto.testCompareViewersPref
- DeprecatedUIPreferencesAuto.testFieldEditorEnablePref
- DeprecatedUIPreferencesAuto.testFontEditorsPref

This is a regression from https://github.com/eclipse-platform/eclipse.platform.team/pull/14 commit https://github.com/eclipse-platform/eclipse.platform.team/commit/3b042694088b273c9c37bdcfbf36b249bf46b5e5

jonahgraham:
The individual checks in this class do "button.getShell().dispose();"
to close the dialog when an assertion fails. That closes the UI
resources, but it does not dispose the non-UI resources/references.

In this particular case when
UIComparePreferencesAuto.testCompareViewersPref failed, it left
org.eclipse.jface.preference.PreferenceNode.page
set to the now disposed General preference page. So when a later
test tries to open the dialog, the test fails due to disposed widget
error.

By calling dialog.close() it makes sure that
org.eclipse.jface.preference.PreferenceNode.disposeResources()
is called, which releases the reference in PreferenceNode.page

### DIFF
--- a/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/DialogCheck.java
+++ b/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/DialogCheck.java
@@ -87,10 +87,13 @@ public class DialogCheck {
 		dialog.open();
 		Shell shell = dialog.getShell();
 		UITestCase.processEvents();
-		verifyCompositeText(shell);
-		dialog.close();
-		// close "verify results" dialog, it makes other tests unhappy
-		_verifyDialog.buttonPressed(IDialogConstants.YES_ID);
+		try {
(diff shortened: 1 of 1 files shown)


jonahgraham:
@iloveeclipse can you review this so that we don't have weird unrelated failures next time?

iloveeclipse (approved):
LGTM.

PR #38, opened 2022-04-26
Reproduce Issue #40

laeubi:
FYI @jukzi this reproduces the issue one see with UI-Freeze in a test case of the code that claims to add the magic.


jukzi:
> FYI @jukzi this reproduces the issue one see with UI-Freeze in a test case of the code that claims to add the magic.

I dont see any magic added in that testcase and do not have the time to work on your "issue" (i still think it's the defined behaviour)

laeubi:
Just see the javadoc of the class:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/45603c38fe124f99c87f9ae5561ce91795b44d9c/tests/org.eclipse.ui.tests/Eclipse%20UI%20Tests/org/eclipse/ui/tests/concurrency/SyncExecWhileUIThreadWaitsForRuleTest.java#L34-L37

**UISynchronizer** and **UILockListener** conspire to prevent deadlock in this case.

> i still think it's the defined behaviour

Feel free, it is just what you are seeing in your recent locking scenarios...

laeubi:
Will be supersede by https://github.com/eclipse-platform/eclipse.platform.ui/pull/40

PR #40, opened 2022-04-27
Prevent JobManager UI freeze when waiting for a rule in the UI Thread

### LINKED ISSUE #8: Prevent JobManager UI freeze when waiting for a rule in the UI Thread

laeubi:
Just a draft for @jukzi


laeubi:
By the way: for this to work more smooth, the default timed wait of 250ms in the Job API might be to long and could potentially feel lagging, I probably would reduce that to 100ms.

vogella:
> By the way: for this to work more smooth, the default timed wait of 250ms in the Job API might be to long and could potentially feel lagging, I probably would reduce that to 100ms.

Sounds fine, but please create a new PR for this.

mickaelistria:
Code looks very good. Can you please rework the commit message to make it more general and not only focus on JobManager? I think this code here is supposed to avoid UI freezes for most progress monitors, JobManager is just 1 consumer of those. This change may also require a rebase for the build to work.

laeubi:
I only found this method used in the `JobManger`, can you give a reference where it is used elsewhere?

mickaelistria:
You're right, sorry, I was confused with `org.eclipse.core.internal.utils.Policy.monitorFor()`. So it seems fine to keep this commit message as it.

laeubi:
I rebased the change now.

mickaelistria:
"on project org.eclipse.e4.ui.progress: Unqualified OSGi version 0.3.400.qualifier must match unqualified Maven version 0.3.300-SNAPSHOT for SNAPSHOT builds -"

laeubi:
Next try :-\

mickaelistria:
Thanks! That's a good change!

PR #43, opened 2022-04-27
Add OSGi-application bundles to features

HannesWell:
This PR adds the OSGi-bundles that replace the formerly embedded OSGi source-code in `org.eclipse.equinox.app` to those Features that contain the Plug-ins whose OSGi-sources are replaced.
The motivation to add the OSGi-bundles to the features is mainly to also have their sources included into the p2-repos of the Eclipse-SDK.

This is part of https://github.com/eclipse-equinox/equinox/issues/18 and should be submitted after https://github.com/eclipse-equinox/equinox.bundles/pull/26 is merged.

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -697,4 +697,11 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
+         id="org.osgi.service.application"
+         download-size="0"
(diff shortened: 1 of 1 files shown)


HannesWell:
Abandon this PR because, as discussed in https://github.com/eclipse-equinox/equinox.bundles/pull/26#issuecomment-1112647217, `org.osgi.service.application` is not used after all.

PR #47, opened 2022-04-28
Reducing the freeze error threshold from 1200ms to 1000 ms

vogella:
Each Eclipse release gets faster so we should reducing the error
threshold for a UI freeze. In general I think 800ms would be good value
but I don't think we are there yet, so "only" reducing the error
threshold to one second.


mickaelistria:
That's a very good sign that project is evolving in the right direction!

vogella:
Thanks @mickaelistria

PR #52, opened 2022-05-04
Example using the display executor

laeubi:
Example for https://github.com/eclipse-platform/eclipse.platform.swt/pull/67

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLine.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLine.java
@@ -358,10 +358,12 @@ import org.eclipse.swt.widgets.ToolItem;
 		// make sure the progress bar is made visible while
 		// the task is running. Fixes bug 32198 for the non-animated case.
 		Runnable timer = () -> StatusLine.this.startTask(timestamp, animated);
-		inUIThread(() -> {
-			fProgressBar.getDisplay().timerExec(DELAY_PROGRESS, timer);
-			if (!animated) {
-				fProgressBar.beginTask(totalWork);
+		getDisplay().execute(() -> {
+			if (fProgressBar != null && !fProgressBar.isDisposed()) {
+				fProgressBar.getDisplay().timerExec(DELAY_PROGRESS, timer);
+				if (!animated) {
+					fProgressBar.beginTask(totalWork);
+				}
 			}
 			if (name == null) {
 				fTaskName = Util.ZERO_LENGTH_STRING;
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLine.java
```diff
@@ -358,10 +358,12 @@ public void beginTask(String name, int totalWork) {
 		// make sure the progress bar is made visible while
 		// the task is running. Fixes bug 32198 for the non-animated case.
 		Runnable timer = () -> StatusLine.this.startTask(timestamp, animated);
-		inUIThread(() -> {
-			fProgressBar.getDisplay().timerExec(DELAY_PROGRESS, timer);
-			if (!animated) {
-				fProgressBar.beginTask(totalWork);
+		getDisplay().execute(() -> {
+			if (fProgressBar != null && !fProgressBar.isDisposed()) {
```
jukzi, line 362: another example where it would be more handsome to have a fProgressBar.execute()

laeubi: sure, but can we please focus on the issue here? It actually would be even more usefull if we would not need the disposal check at all as it is mostly useless anyways instead of wrapping more code around this, all of this is not related to the change present here...

laeubi: @jukzi I have created 
- https://github.com/eclipse-platform/eclipse.platform.swt/discussions/92

bundles/org.eclipse.jface/src/org/eclipse/jface/util/Policy.java
```diff
@@ -147,7 +147,7 @@ private static StatusHandler getDummyStatusHandler() {
 
 			@Override
 			public void show(final IStatus status, String title) {
-				Runnable runnable = () -> {
+				Display.executor().execute(() -> {
```
mickaelistria, line 150: While looking at this API, I don't think the name `executor` is good here. In the end as Display implements Excutor, what executor returns here is the "current or default" display. I think the method name should reflect it: `Display.currentOrDefault().execute(...)` tells much more about the intention and actual behavior.

mickaelistria:
Thanks for those examples, they IMO showcase the good value proposal and will allow to make more concrete progress with this API.

vogella:
Is the SWT part finally merged?

laeubi:
Not yet see https://github.com/eclipse-platform/eclipse.platform.swt/pull/289

akurtakov:
Is this one still needed ?

mickaelistria:
It's not needed, but it would be good to update it to current state of things to show clients of the newer Display capability.
@laeubi Please either update the code or close, at your ow convenience.

PR #54, opened 2022-05-04
Disable MarkerSortUtil batch sorting. (closes #53)

### LINKED ISSUE #53: slow marker sort (many markers)
Sometimes when there is a severe error (like missing target platform)  in the workspace i end up with many problem Markers which are sorted quiet slow:
image

Looks like the slow thing is the language aware comparison.  I don't see much value of sorting problem message (usual english) by Locale aware Comparator. Even on german which has special sorting rules i would not need that feature. When i click on sort by description i just want to have same descriptions grouped.
Also i wonder if all those "performance optimizations " in MarkerSortUtil realy work as desired. It feels like the collatorcache is cleared too often. Also i wonder what that optimizations try to archive - getting the top(k) of n entries always has a minimal O(n+log(k)) which is not significant better than the trivial O(n*log(n)) when just sorting all entries.

jukzi:
Can be reenabled with org.eclipse.ui.MarkerSortUtil.batchSize=10000

Batch sorting was intended to decrease memory consumption during sorting
Marker View. But it is very slow especially when CollationKeys are
calculated. Today temporary memory consumption is not as much an issue
any more, but performance is.

Performance difference is visible with MarkerSortUtilTest especially
when increasing ARRAYSIZE to a million (22sec vs 2sec) - even without
CollationKeys.

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerEntry.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerEntry.java
@@ -21,6 +21,7 @@ import java.text.Collator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.runtime.CoreException;
@@ -68,6 +69,7 @@ class MarkerEntry extends MarkerSupportItem implements IAdaptable {
 	private static final String LOCATION_STRING = "LOCATION_STRING"; //$NON-NLS-1$
 	private MarkerCategory category;
 	private Map<String, Object> cache;
+	private static Map<String, CollationKey> collationCache = new ConcurrentHashMap<>();
 
 	/**
 	 * Set the MarkerEntry to be stale, if discovered at any point of time
@@ -219,7 +221,8 @@ class MarkerEntry extends MarkerSupportItem implements IAdaptable {
 		if (attributeValue.isEmpty()) {
 			return MarkerSupportInternalUtilities.EMPTY_COLLATION_KEY;
 		}
-		CollationKey key = Collator.getInstance().getCollationKey(attributeValue);
+		CollationKey key = collationCache.computeIfAbsent(attributeValue,
+				k -> Collator.getInstance().getCollationKey(attributeValue));
 		getCache().put(attribute, key);
 		return key;
 	}
@@ -385,6 +388,7 @@ class MarkerEntry extends MarkerSupportItem implements IAdaptable {
 	@Override
 	void clearCache() {
 		cache = null;
+		collationCache = new ConcurrentHashMap<>();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


jukzi:
@iloveeclipse could you please verify your workspace with many markers does not make trouble during sorting markers view.

iloveeclipse:
I would recommend just create simple project with 10000 files and create 10 different markers per file. You do not need anything special here. You can use my project generator to do that. Just break generated java code and that's it. 
https://github.com/iloveeclipse/java-project-generator

jukzi:
with the generated project from https://github.com/iloveeclipse/java-project-generator/pull/1 the problem is pretty obvious, i even get a waiting popup during sort.
With both commits the perfromance is not only ~ 4sec vs instant sort but also with LESS memory allocation during sort (i do not know why). Both commits are needed to archive the instant sort.

jukzi:
image

mickaelistria:
This looks good to me.
While looking a bit at this code, wouldn't it be worth trying to refactor it so it becomes some "SortWithProgrssUtil" with generified methods as I don't see any logic here that's specific to markers.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerEntry.java
```diff
@@ -385,6 +388,7 @@ Map<String, Object> getCache() {
 	@Override
 	void clearCache() {
```
iloveeclipse: `clearCache()` seem to be called on **every** marker. Does it make sense to have a dedicated method for the static cache cleanup? Just to avoid the map is recreated 100_000 times, also it sounds strange to cleanup static data in non static context.

jukzi: > `clearCache()` seem to be called on **every** marker

done

jukzi:
> This looks good to me.
> While looking a bit at this code, wouldn't it be worth trying to refactor it so it becomes some "SortWithProgrssUtil" with generified methods as I don't see any logic here that's specific to markers.

Actually the cancel is the only Goodie we loose by disabling the batched sort. But i prefer to sort in 1 sec then being able to cancel a longer running sort.
The only marker specific is the clear of caches.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerEntry.java
```diff
@@ -219,7 +221,8 @@ CollationKey getCollationKey(String attribute, String defaultValue) {
 		if (attributeValue.isEmpty()) {
 			return MarkerSupportInternalUtilities.EMPTY_COLLATION_KEY;
 		}
-		CollationKey key = Collator.getInstance().getCollationKey(attributeValue);
+		CollationKey key = collationCache.computeIfAbsent(attributeValue,
+				k -> Collator.getInstance().getCollationKey(attributeValue));
```
iloveeclipse, line 225: I haven't tried that in the debugger yet, but what kind of values are cached here, and why is this cache **per marker**? Shouldn't that be cache per view, or static one?

jukzi: ancient code.. currently it's only the IMarker.MESSAGE of each marker

jukzi:
any objections to merge?

Issue #56, opened 2022-05-05
LockManager.handleException errors on shutdown


iloveeclipse:
I've seen now two times sporadic `LockManager.handleException` errors logged on shutdown of the IDE.

Looks some of the recent changes in UI/SWT area caused that, bugzilla never saw that error before.
First one was reported with  4.24.0.I20220504-0230 build, but that doesn't mean this was introduced there, because error is random (only if some workspace task runs at shutdown time).

Looks like the Display is now disposed earlier on shutdown?

Could be side effect of removing synchronization block from Display.isDisposed() via https://github.com/eclipse-platform/eclipse.platform.swt/issues/74 / https://github.com/eclipse-platform/eclipse.platform.swt/pull/75

Also could be a side effect of https://github.com/eclipse-platform/eclipse.platform.ui/issues/8.

Autobuild running during shutdown:
```
org.eclipse.swt.SWTException: Device is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Display.error(Display.java:1505)
	at org.eclipse.swt.widgets.Display.getThread(Display.java:3378)
	at org.eclipse.ui.internal.UILockListener.isUI(UILockListener.java:188)
	at org.eclipse.ui.internal.UILockListener.aboutToRelease(UILockListener.java:124)
	at org.eclipse.core.internal.jobs.LockManager.aboutToRelease(LockManager.java:88)
	at org.eclipse.core.internal.jobs.OrderedLock.doRelease(OrderedLock.java:189)
	at org.eclipse.core.internal.jobs.OrderedLock.release(OrderedLock.java:237)
	at org.eclipse.core.internal.resources.WorkManager.checkOut(WorkManager.java:167)
	at org.eclipse.core.internal.resources.Workspace.endOperation(Workspace.java:1523)
	at org.eclipse.core.internal.events.AutoBuildJob.doBuild(AutoBuildJob.java:176)
	at org.eclipse.core.internal.events.AutoBuildJob.run(AutoBuildJob.java:255)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
Same as above, but from refresh job
```
org.eclipse.swt.SWTException: Device is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4918)
[...]

PR #59, opened 2022-05-05
LockManager.handleException errors on shutdown (#56)

iloveeclipse:
1) Fixed unneeded calls to display from aboutToRelease(). All what we
want is to "null" "ui" field if we are in UI thread. For that we don't
need to check isUI() and Display because "ui" reference should be either
UI thread (which can be compared directly with current thread) or it is
null and we don't care.

2) Fixed race condition in isUI(), which assumes the display is not
disposed after the display.isDisposed() check. But since we are not
guaranteed to be in UI thread, someone could have disposed Display in
the meantime. So just call Thread.currentThread() that already does this
check for us.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/56

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/UILockListener.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/UILockListener.java
@@ -121,7 +121,8 @@ public class UILockListener extends LockListener {
 
 	@Override
 	public void aboutToRelease() {
-		if (isUI()) {
+		Thread uiThread = ui;
+		if (uiThread != null && uiThread == Thread.currentThread()) {
 			ui = null;
 		}
 	}
@@ -185,7 +186,11 @@ public class UILockListener extends LockListener {
 	}
 
 	boolean isUI() {
-		return (!display.isDisposed()) && (display.getThread() == Thread.currentThread());
+		try {
+			return display.getThread() == Thread.currentThread();
+		} catch (Exception e) {
+			return false;
+		}
 	}
 
 	boolean isUIWaiting() {


iloveeclipse:
?
```
org.eclipse.swt.SWTException: Invalid thread access
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Display.error(Display.java:1505)
	at org.eclipse.swt.widgets.Display.checkDevice(Display.java:956)
	at org.eclipse.swt.widgets.Display.addFilter(Display.java:670)
	at org.eclipse.e4.ui.bindings.tests.KeyDispatcherTest.testExecuteMultiStrokeBinding(KeyDispatcherTest.java:200)
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/UILockListener.java
```diff
@@ -121,7 +121,8 @@ public UILockListener(Display display) {
 
 	@Override
 	public void aboutToRelease() {
-		if (isUI()) {
+		Thread uiThread = ui;
+		if (uiThread != null && uiThread == Thread.currentThread()) {
```
laeubi, line 125: For me it seems that `uiThread` would be better an `AtomicReference` and one could better perform `compareAndSet` (also at other places in this class) but that might be another change :-)

laeubi (approved):
Looks good to me, maybe we could even improve with using AtomicReferences here...

iloveeclipse:
Thanks for review.

Issue #85, opened 2022-05-26
The default button in FiltersConfigurationDialog is in the wrong section


eobrienPilz:
The FiltersConfigurationDialog that is used by MarkerSupportView has a usefull button to restore default settings. However the button is positioned at the same level as the ok and cancel buttons. Instead it should be in the section used to configure the filter settings.
See the the position of the default button in the preferences dialog for the correct place.

eobrienPilz:
Screenshots showing what I mean.

![filterDefaults](https://user-images.githubusercontent.com/58333578/170678435-560d6169-7dd6-41e7-95f1-33ddea90cc1a.png)
![preferencesRestoreDefaults](https://user-images.githubusercontent.com/58333578/170678442-c77328ac-12bc-4838-8a60-25f87f02f6de.png)

Wittmaxi:
@eobrienPilz can we close this issue?

eobrienPilz:
yes thanks, I just checked in 4.28 and it looks good now .

Issue #86, opened 2022-05-27
MarkerSupportView Should allow a help link on the FilterConfigurationDialog


eobrienPilz:
We would like to have the option of having a help link on the FilterConfigurationDialog. As the dialog is internal we can't extend and there is no method provided to add a help link in the dialog. There is an attachment that shows what we would like to be able to do
![helpLink](https://user-images.githubusercontent.com/58333578/170675082-e1e7de6a-c3b1-4846-97b2-f1f7f59f0c80.png)
.

mickaelistria:
Would the help be a generic help about this dialog, that could go into common Eclipse Platform, or do you want something that's specific to your application?

eobrienPilz:
> Would the help be a generic help about this dialog, that could go into common Eclipse Platform, or do you want something that's specific to your application?

We need something specific to our application. Such as an API to allow the help link to appear on the dialog and a handler to launch the help page when the button is pressed.
I was thinking maybe it could be an addition to markerSupport extension point, because FiltersConfigurationDialog is internal and I'm not sure if its ok to extended for use as an API.

eobrienPilz:
I just added a pull request with a potential fix for this: https://github.com/eclipse-platform/eclipse.platform.ui/pull/348

eobrienPilz:
:)

PR #87, opened 2022-05-27
The default button in FiltersConfigurationDialog is in the wrong section #85

eobrienPilz:
This fixes issue 85 https://github.com/eclipse-platform/eclipse.platform.ui/issues/85

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/FiltersConfigurationDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/FiltersConfigurationDialog.java
@@ -26,11 +26,13 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 
+import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.dialogs.IDialogSettings;
 import org.eclipse.jface.dialogs.IInputValidator;
 import org.eclipse.jface.dialogs.InputDialog;
 import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.jface.resource.JFaceResources;
 import org.eclipse.jface.viewers.ArrayContentProvider;
 import org.eclipse.jface.viewers.CheckboxTableViewer;
 import org.eclipse.jface.viewers.ISelection;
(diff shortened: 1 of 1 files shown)


mickaelistria (changes requested):
The fix is not at the right place: the FilterConfigurationDialog should keep extending the ViewSettingsDialog as it makes a lot of sense; the issue of the "Restore Defaults" button being misplaced is something that is common to several other dialog; so either it's fixes in ViewSettingsDialog and this Dialog inherits the change, or this dialog is kept as it.

mickaelistria:
I spent more time looking at this proposal and come to the conclusion that ViewSettingsDialog is of no help and has no strong semantic and is not much used. So moving away from this class is eventually a good thing; and such patch would be welcome.
Have you considered using a TrayDialog as base class instead of a plain Dialog?

eobrienPilz:
> I spent more time looking at this proposal and come to the conclusion that ViewSettingsDialog is of no help and has no strong semantic and is not much used. So moving away from this class is eventually a good thing; and such patch would be welcome. Have you considered using a TrayDialog as base class instead of a plain Dialog?

Looking at the documentation for TrayDialog it looks to be correct to use it here. It has a help link appearing by default but this can be disabled for now. The option of the help link may be useful when fixing issue #86. I will push these changes now.

vogella:
@mickaelistria you are blocking this PR. Any update you still require?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/FiltersConfigurationDialog.java
```diff
@@ -678,6 +691,13 @@ protected void okPressed() {
...
 
 	@Override
+	protected void buttonPressed(int buttonId) {
+		if (buttonId == DEFAULTS_BUTTON_ID) {
+			performDefaults();
+		}
+		super.buttonPressed(buttonId);
+	}
+
 	protected void performDefaults() {
```
mickaelistria, line 701: `@Override` should remain here.

eobrienPilz: Its not @Override now that its using TrayDialog.

mickaelistria: Ok, my bad.

mickaelistria (approved):

mickaelistria:
Thanks!

PR #90, opened 2022-06-01
Adopt CTabFolderRenderer.setHightlightBarEnabled

mickaelistria:
So workbench tabs have better highlight without theming


mickaelistria:
Requires https://github.com/eclipse-platform/eclipse.platform.swt/pull/183 first.

vogella:
@mickaelistria is this still necessary? I think the SWT one is already sufficient.

mickaelistria:
Indeed, it's not useful any more.

PR #97, opened 2022-06-07
Lighter System theme

mickaelistria:
* Overrides less colors
* Delegates more to native
* Only highlight active tab with color bar (no background)


vogella:
+1, looks better to me than the current system theme on Windows.

vogella:
Please add to N&N

PR #103, opened 2022-06-08
Remove System.out output from CSS test cases

vogella:
Currently the CSS test cases printout the memory usage and its start
time. AFAIK this information has not been used in the recent years to
improve performance or memory consumption of the CSS engine and should
be better done with tracing tools.


vogella:
Jörg, WDYT?

jukzi:
+1. was added for https://bugs.eclipse.org/bugs/show_bug.cgi?id=494641 which is closed.

vogella:
Thanks @jukzi

PR #111, opened 2022-06-08
Remove form customization from System theme

mickaelistria:
Stick to default form, if those needs tweak it's better to try
implementing them directly in the Form than in the theme.


vogella:
+1 

But as the gradient looks so ugly in forms it would be great if you could also change the default in forms to use a solid color.

mickaelistria:
> But as the gradient looks so ugly in forms it would be great if you could also change the default in forms to use a solid color.

That was my initial goal with that part of the theme, but it appeared that customizing with theme leads to some bugs (eg section text not readable). I'll try to work on the gradient some time later.

vogella:
> > But as the gradient looks so ugly in forms it would be great if you could also change the default in forms to use a solid color.
> 
> That was my initial goal with that part of the theme, but it appeared that customizing with theme leads to some bugs (eg section text not readable). I'll try to work on the gradient some time later.

Thanks. I opened https://github.com/eclipse-platform/eclipse.platform.ui/issues/114 would be nice to move forms to a less Win95 L&F.

iloveeclipse (approved):
Looks finally usable.

PR #121, opened 2022-06-09
Disable gradients in Form by default

### LINKED ISSUE #114: Remove forms default gradient and use a SWT color constant
JFace forms default currently use still gradients, which makes such UI look very Win95. We use a default SWT color.

See #111

mickaelistria:
Default color for Forms are changed so that gradients are removed,
system colors are used.

### DIFF
--- a/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/FormColors.java
+++ b/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/FormColors.java
@@ -535,7 +535,7 @@ public class FormColors {
 		// LIST_SELECTION foreground colour
 		// Examples: Win High Con Black, Win High Con #1, Win High Con #2
 		// (covered in the second part of the OR clause above)
-		createColor(IFormColors.TITLE, rgb);
+		createColor(IFormColors.TITLE, getSystemColor(SWT.COLOR_WIDGET_FOREGROUND));
 	}
 
 	private void createTwistieColors() {
@@ -547,7 +547,7 @@ public class FormColors {
 	}
 
 	private void createTitleBarGradientColors() {
-		RGB tbBg = getSystemColor(SWT.COLOR_TITLE_BACKGROUND);
+		RGB tbBg = getSystemColor(SWT.COLOR_WIDGET_LIGHT_SHADOW);
 		RGB bg = getImpliedBackground().getRGB();
 
 		// Group 1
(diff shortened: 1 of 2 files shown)


vogella:
New

image

Old

image


Definitely better, lets merge so that it gets a lot of usage during the release phase.

vogella (approved):

trancexpress:
Will I be seeing this change in my IDE or is it something coming from the theme?

iloveeclipse:
Open plugin.xml editor and enjoy.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/FormColors.java
```diff
@@ -535,7 +535,7 @@ else if (testTwoPrimaryColors(listSelection, 120, 256)
 		// LIST_SELECTION foreground colour
 		// Examples: Win High Con Black, Win High Con #1, Win High Con #2
 		// (covered in the second part of the OR clause above)
-		createColor(IFormColors.TITLE, rgb);
+		createColor(IFormColors.TITLE, getSystemColor(SWT.COLOR_TITLE_FOREGROUND));
```
vogella: Even nicer and more consistent with the OS systems would IMHO be to use a gray color as background, not sure if we have a good constant for that.

vogella:
This is how it looks currently in the light theme under Linux with gray background. IMHO this is more inline with the current OS L&F

image

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/FormColors.java
mickaelistria: The palette is visible running SWT's ControlExample in the "Colors" tab. There are the COLOR_WIDGET_*_SHADOW to serve this. I personally don't mind about which exact color we pick, as long as it's readable, it's a system color and gradients are off by default. So please give a try and suggest a color and I can update.

vogella:
The current gray in the light theme seems to fit in nicely with the OS

image

mickaelistria:
@vogella I have changed to use COLOR_WIDGET_LIGHT_SHADOW instead.

vogella:
> COLOR_WIDGET_LIGHT_SHADOW

Can you add a screenshot?

mickaelistria:
Screenshot from 2022-06-10 12-53-08

vogella:
On my Ubuntu

image


LGTM, +1

PR #124, opened 2022-06-10
Remove the help icon from the save dialog

vogella:
Operating system Ui guidelines do not recommend the usage of the help
icon in a dialog. For example check the screenshots in the Windows
guidelines for dialogs.

Operating System UI Guidelines

Windows UI guidence
https://docs.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs

Gnome UI guidence
https://developer.gnome.org/hig/patterns/feedback/dialogs.html

Mac UI guidence
https://developer.apple.com/design/human-interface-guidelines/components/presentation/alerts/


mickaelistria:
While I'm fine with this change, I think we should maybe update the UI Guidelines wiki to state that the system guidelines have higher priority and Eclipse's one when it comes to development of Eclipse applications. So that we can say "if guidelines for all supported OS have aligned recommendation, then those should be decisive one and local Eclipse guidelines can be ignore if they contradict those"

vogella:
Thanks @mickaelistria we are currently discussing an update of the UI recommendations based on your email you send. 


IMHO the UI Guidelines wiki shoud be marked as obsolete, nobody maintains this anymore.

PR #128, opened 2022-06-10
Removes the form styling from the Eclipse default light theme

vogella:
With https://github.com/eclipse-platform/eclipse.platform.ui/pull/121
form default L&F has been updated to have no gradient anymore.

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/light/e4-light_globalstyle.css
+++ b/bundles/org.eclipse.ui.themes/css/light/e4-light_globalstyle.css
@@ -14,33 +14,6 @@
  *******************************************************************************/
 
 
-Form, FormHeading {
-	background-color: #ffffff;
-	background: #ffffff;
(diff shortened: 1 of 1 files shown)


vogella:
Screenshot:

image

vogella:
@mickaelistria WDYT? This will make the forms styling changes you did visible for the default light theme users.

mickaelistria:
> @mickaelistria WDYT? This will make the forms styling changes you did visible for the default light theme users.

I don't think it's correct to use native colors in the Light Theme if the goal is to have it using an enlighten. By removing this customization, then the color become the native/system ones, which may not be light. Once a theme starts overriding some colors, it needs to cascade into overriding all colors. You can for example try this change on a system that has an overall dark theme, but using the light theme, I guess it would look very inconsistent.

vogella:
@mickaelistria makes sense. Thanks.

PR #133, opened 2022-06-13
Bug 578581 - MarkerSupportView custom groupings requires use of internal GroupsContribution

eobrienPilz:
This is a patch submission for the bugzilla issue here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=578581

The proposal is to have an API equivalent of the ContributionItem for Marker groups.
This will allow clients to create their own custom groupings.

Note: Creating the custom groupings was always possible but relied on referencing the internal GroupsContribution in the xml definition.


mickaelistria:
Can't CompoundContributionItem be used directly by clients?

eobrienPilz:
> Can't CompoundContributionItem be used directly by clients?

Yes, but the classes to access marker grouping information are internal i.e. see GroupsContribution.getContributionItems(). 
It needs to access more that one internal class to populate the menu and execute the grouping action.

PR #134, opened 2022-06-13
Fix access to the UISynchronize

laeubi:
Currently the code assumes that the Services is always there, what is
not true. Beside that, it ignores if a users sets a Display for
execution completely.

This is now fixed, by first check if there is a user display, then if
the addon is active and can supply required objects, and as a last
resort fall back to the swt displays.

### DIFF
--- a/bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/UIJob.java
+++ b/bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/UIJob.java
@@ -24,6 +24,7 @@ import org.eclipse.e4.ui.progress.internal.ProgressMessages;
 import org.eclipse.e4.ui.progress.internal.Services;
 import org.eclipse.e4.ui.progress.internal.legacy.PlatformUI;
 import org.eclipse.e4.ui.progress.internal.legacy.StatusUtil;
+import org.eclipse.e4.ui.workbench.swt.DisplayUISynchronize;
 import org.eclipse.swt.widgets.Display;
 
 /**
@@ -148,7 +149,10 @@ public abstract class UIJob extends Job {
 	 */
 	public Display getDisplay() {
 		if (cachedDisplay == null) {
-			cachedDisplay = Services.getInstance().getDisplay();
+			Services instance = Services.getInstance();
+			if (instance != null) {
+				cachedDisplay = instance.getDisplay();
+			}
 		}
 		if (cachedDisplay == null) {
 			cachedDisplay = Display.getCurrent();
@@ -200,7 +204,11 @@ public abstract class UIJob extends Job {
 	 * @return UI synchronizer
 	 */
 	protected UISynchronize getUiSynchronize() {
-		return Services.getInstance().getUISynchronize();
+		Services instance = Services.getInstance();
+		if (instance == null) {
+			return new DisplayUISynchronize(getDisplay());
+		}
+		return instance.getUISynchronize();
 	}
 
 }
--- a/bundles/org.eclipse.e4.ui.progress/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.progress/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.progress;singleton:=true
-Bundle-Version: 0.3.400.qualifier
+Bundle-Version: 0.3.500.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/bundles/org.eclipse.e4.ui.progress/pom.xml
+++ b/bundles/org.eclipse.e4.ui.progress/pom.xml
@@ -16,7 +16,7 @@
   </parent>
   <groupId>org.eclipse.e4</groupId>
   <artifactId>org.eclipse.e4.ui.progress</artifactId>
-  <version>0.3.400-SNAPSHOT</version>
+  <version>0.3.500-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
   
   <properties>


laeubi:
@vogella can you review this?

vogella:
Code which gets access of UISynchronize can uses this to update the UI. AFAICS this would not be the case with your patch.

From Javadoc of UISynchronize:

 * Widget toolkit abstract to synchronize back into the UI-Thread from other
 * threads
 *

laeubi:
> Code which gets access of UISynchronize can uses this to update the UI. AFAICS this would not be the case with your patch.

I don't understand can you explain? Do you mean we should have a `setUISynchronize()` as well?

vogella:
> > Code which gets access of UISynchronize can uses this to update the UI. AFAICS this would not be the case with your patch.
> 
> I don't understand can you explain? Do you mean we should have a `setUISynchronize()` as well?



> > Code which gets access of UISynchronize can uses this to update the UI. AFAICS this would not be the case with your patch.
> 
> I don't understand can you explain? Do you mean we should have a `setUISynchronize()` as well?

If I read your code correct, your change would allow to provide a UISynchronizer via getUiSynchronize which would not use the main thread and hence would violate the contact that a caller running outside main thread could call UISynchronize.asyncExec or syncExec to update the UI.

vogella:
Ah, sorry, you are not providing UISynchronizer but UISynchronize. No idea what contract this one has. :-o

laeubi:
> If I read your code correct, your change would allow to provide a UISynchronize via getUiSynchronize which would not use the main thread and hence would violate the contact that a caller running outside main thread could call UISynchronize.asyncExec or syncExec to update the UI.

That part of the code is not changed (it was protected before), for sure if anyone override this method he must make sure to provide an appropriate `UISynchronize` instance.

If you like I can make it private, but be aware that this will be considered an API break.

laeubi:
> No idea what contract this one has. :-o

```
/**
 * Widget toolkit abstract to synchronize back into the UI-Thread from other
 * threads
 *
 * @noextend This class is not intended to be subclassed by clients.
 * @since 1.0
 */
```

I think the main purpose here is to support other toolkits (e.g. Java FX Eclipse)

laeubi:
> No idea what contract this one has.

By the way as we are writing here I think we should only use Display internally and not expose it to any kind, what do you think? This requires API break but as version is `0.3` ... and it would only break consumers overriding a method that has no effect at all...

laeubi:
Or we even change it to 1.0 and add it to the sdk feature?

vogella:
> use Display internally and not expose it to any kind

I think we still have to honor the API contact, but please go ahead and deprecated any access to Display with the forRemoval flag. Maybe also use this a case to simplify the API removal process? https://github.com/eclipse-platform/.github/issues/22
 
> Or we even change it to 1.0 and add it to the sdk feature?

This plug-in was originally written as e4 replacement for the e3 progress view but nobody dared to do the actually replacement as it was considered to risky.

laeubi:
> but please go ahead and deprecated any access to Display with the forRemoval flag.

Done

mickaelistria:
> please go ahead and deprecated any access to Display with the forRemoval flag.

Are there any references to these deprecated methods in Platform/JDT/PDE?

laeubi:
Not as I can see, the Job seems to be mostly used internal or in tests.

trancexpress:
> Not as I can see, the Job seems to be mostly used internal or in tests.

I see a few uses in the product I work on. Since this change will result in deprecation and eventually method removal, please update the description of the PR and the commit message with more information. In particular why methods are now planned to be removed (breaking client code).

What is the usual timeline for removal of deprecated methods?

laeubi:
> What is the usual timeline for removal of deprecated methods?

I think 2 Year or something incredible long...

> I see a few uses in the product I work on.

Can you share why you have any need for passing a display that is effectively never used? probably you can simplify your code and upgrading would be no issue for you then.

trancexpress:
> Can you share why you have any need for passing a display that is effectively never used? probably you can simplify your code and upgrading would be no issue for you then.

Not my code; I've not checked yet. Seeing the API and the affected code, on first glance: code relies on ```UIJob``` to store the display, for later-on reuse when the job runs. I'm sure I can move that to the client code, if platform UI no longer "wants" to do it.

For the commit/PR message, I want to know what the reason for the deprecation is without going through the comment thread here (ideally without even visiting the PR, just reading the ```git``` history). Adding the info in the JavaDoc deprecation annotation will also work for me.

laeubi:
> on first glance: code relies on `UIJob` to store the display, for later-on reuse when the job runs. I'm sure I can move that to the client code, if platform UI no longer "wants" to do it

You can simply pass that as a constructor parameter or you can even simply call `Display.getCurrent()` in E4 one should never use the `Display` but the `UISyncronize` to interact with the "main thread".

trancexpress:
> You can simply pass that as a constructor parameter or you can even simply call `Display.getCurrent()` in E4 one should never use the `Display` but the `UISyncronize` to interact with the "main thread".

As I already commented, I'm sure I'll be able to adjust the code to prepare for the removal of API.

bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/UIJob.java
```diff
@@ -48,15 +48,26 @@ public UIJob(String name) {
...
 	 * Create a new instance of the receiver with the supplied Display.
 	 *
-	 * @param jobDisplay
-	 *            the display
-	 * @param name
-	 *            the job name
+	 * @param jobDisplay the display
+	 * @param name       the job name
 	 * @see Job
+	 * @deprecated
```
mickaelistria: Please describe the alternative to use with some hints about how to migrate code.

mickaelistria (changes requested):
The commit header needs to mention te deprecation, and give more hints about how to migrate away from the deprecated method.

bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/UIJob.java
laeubi: done.

bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/UIJob.java
```diff
@@ -48,15 +48,27 @@ public UIJob(String name) {
...
 	 * Create a new instance of the receiver with the supplied Display.
 	 *
-	 * @param jobDisplay
-	 *            the display
-	 * @param name
-	 *            the job name
+	 * @param jobDisplay the display
+	 * @param name       the job name
 	 * @see Job
+	 * @deprecated use {@link #UIJob(UISynchronize, String)} to supply alternative
```
mickaelistria: How to get a `UISynchronize` from a Display? Would be good to mention that and suggest a blind replacement customers can use without wondering about anything else.

laeubi: > How to get a `UISynchronize` from a Display?

You don't get it from a display, you let inject it into your E4 part (the `UISynchronize`), see for example:

- https://www.vogella.com/tutorials/EclipseJobs/article.html#using-dependency-injection-and-uisynchronize

mickaelistria: Please explain that in Javadoc, or link to some other existing Javadoc that explains that.

laeubi: The documentation is here: https://wiki.eclipse.org/E4 and everyone using E4 should be well aware of that so this is really nothing special that requires any documentation here...

mickaelistria: The fact that migration path are poorly documented is 1 reason why e4 is still not adopted even in the SDK. I would rather use the current interest as an opportunity to really make things better and easier for everyone to adopt e4 than to just repeat things that haven't been much successful. We need to raise the quality bar on documentation, and this means having good javadoc in the code more than referencing outdated wikis on PRs.

cdietrich: i am also interested in a non e4 replacement.

laeubi: > **We** need to raise the quality bar on documentation, and this means having good javadoc in the code more than referencing outdated wikis on PRs.

Yep, **we** should do that, bit I can't afford handling the technical dept of 14 years platform, so the best opportunity here is to step up and start with it right now, but this should be handled in a dedicated change.

So let me know if actually making adoption of E4 possible and useful and accepting that an unknown class can't tell the whole story of E4 or I if I should simply revert, keep the old useless code and AP in and just fix the NPE to the bare minimum here...

mickaelistria: > So let me know if actually making adoption of E4 possible and useful

Yes, but deprecating for removal without good documentation is not. So maybe just keep the constructor based on Display at the moment without deprecating it.

laeubi: This code is effectively not used (even not internally) and gives false impression here, because it has non effect atm... so actually I would remove the constructor completely here as it was broken from day-1 here... I'm also not sure what esle schould be "better" documented here as that we will remove it and there is an (e4 compatible) alternative?

mickaelistria: > This code is effectively not used (even not internally)

That you cannot tell for sure when it's API.

> I'm also not sure what esle schould be "better" documented here as that we will remove it and there is an (e4 compatible) alternative?

As mentioned earlier, the expectation when documenting a deprecation is that people can read the deprecation notice from the javadoc and quickly enough know how to replace their code using the deprecated method by something else that is not deprecated. This is not about just telling "use that instead", but when the difference is non trivial, more "from your code, do this to get an instance of that and then apply this other method".

laeubi: Yeah forget about that stupid deprecation, one should just not try to deprecate anything and then we even don't need a simpler "removal process at all, so all win-win...

laeubi:
Fix reduced to the absolute bare minimum now

bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/UIJob.java
mickaelistria: >  one should just not try to deprecate anything 

That's not the goal; but it's the drawback of declaring something as API. For a Platform that claims sustainability like Eclipse Platform, creating an API means that the API will be maintained as working without regression, will not be removed without a notice and that it's removal will be accompanied with clear steps on how to migrate. Don't do it well and the Platform breaks clients too strongly, extensions don't work, people move to other platforms, ecosystem dies, community dies, technology dies... That is the real cost and risk with addition APIs, and this cost is something you easily forget when you create an API, as the project pays it later, APIs contract do reduce agility and capability to innovate. But APIs are a necessary evil. I hope that those case-studies help you understand why I'm always defensive when it comes to API addition. There will always be a payback, and often we get the impression having an API was overall not profitable and world would have been better without it.

mickaelistria (approved):
Looks OK to me. Please also bump version by +0.0.100

mickaelistria (approved):

PR #136, opened 2022-06-13
Set explicit project encoding to UTF-8

Bananeweizen:
Avoid the newly introduced warning in all platform.ui projects.


akurtakov:
Very welcome change, esp if you finish it :)

Bananeweizen:
On this repo, it should be complete. For the other platform repos, the Oomph setup was still running when I went to sleep. :)

akurtakov:
Fixing the warnings could be finished but bumping all the changed projects versions is far from it as could be seen in the failing check.

akurtakov:
Completed via https://github.com/eclipse-platform/eclipse.platform.ui/pull/428

PR #137, opened 2022-06-14
Allow registration of link handers handled by other application

BeckerWdf:
Let's try if we can also "take over" link hander registration from other applications on macOS like we can on Windows.


BeckerWdf:
Unfortunately this does not work.
If you simply modify the .plist file of your application an register it in the launch system database this does not help because the other application is still registered. 
One would have to (at least) unregister the other application and register your own application.

BeckerWdf:
> One would have to (at least) unregister the other application and register your own application.

And it's possible that when you unregister another application that there's yet another app that also has the uri scheme registered.
So opening the preference page would show yet another "other application" that handles the uri scheme.

BeckerWdf:
i won't follow up on this.

PR #140, opened 2022-06-14
Improve documentation 

akurtakov:
* Remove psf files pointing to gerrit still - probably covered by Oomph for some time so no one cared.
* Fix documentation files to point to github
* Remove no longer active repos.


akurtakov:
No need to wait for verification build as it's only docs touched.

PR #141, opened 2022-06-14
Remove animations from progress toolbar control

vogella:


vogella:
@jukzi any concerns?

laeubi:
Why are animations bad?

jukzi:
Would you mind to share a video/screenshot which animation is affected and why you want to remove it?

laeubi:
@jukzi I think it is the small icon on the lower right when there is a running job but not sure...

akurtakov:
Animations are not bad per se. Not well written and tested animations are not helping the UI though esp. considering that underlying OSes already "animate" the widget .

vogella:
> Why are animations bad?

During startup StandardTrim creates an instance of
TaskBarProgressManager in line 52.

Before this change TaskBarProgressManager supports animation, which
seems useless / overkill
for progress reporting. This animation code also causes exception if you
execute unit tests, as I discovered during a test run of the resource
test suite.

This commit removes the animation logic but keeps the update logic of
updating the progress state of the item. I personal cannot note any
difference in the resulting runtime Eclipse.

jukzi:
at the bottom right there is an animated progress (which is most time showing wrong progress):
image

is this change about that?

vogella:
> at the bottom right there is an animated progress (which is most time showing wrong progress): image
> 
> is this change about that?

yes

jukzi:
i'd rather like to get the progress fixed instead of removed. can't be that hard since the task bar shows an appropriate percentage:
image

vogella:
> i'd rather like to get the progress fixed instead of removed. can't be that hard since the task bar shows an appropriate percentage: image

Are you planning to do this? I won't fix animations.

jukzi:
I agree that the current animation is useless 
image
I shows just an animation like when progress is not know - could if show the percentage instead, please?

jukzi:
> This animation code also causes exception if you
> execute unit tests, as I discovered during a test run of the resource
> test suite.

please attach the stacktraces.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/TaskBarProgressManager.java
```diff
@@ -83,44 +80,18 @@ public TaskBarProgressManager(TaskItem taskItem) {
...
-					}
-				} else {
-					setProgressState(SWT.DEFAULT);
-					updateImage(null);
-				}
+				setProgressState(SWT.DEFAULT);
+				updateImage(null);
 
-				if (isAnimated && taskItem != null && !taskItem.isDisposed()) {
+				if (taskItem != null && !taskItem.isDisposed()) {
```
jukzi: if it is not animated it does not need to be rescheduled

jukzi (dismissed):
i don't see a reason to remove the animation. but if we would remove it the drawing does not need to be rescheduled

vogella:
> i don't see a reason to remove the animation. but if we would remove it the drawing does not need to be rescheduled

Done

vogella:
Like I explained in the commit message I cannot see a difference in behavior after removing the animation.

Here is a screenshot before the removal if I load for example a  target platform:

currenty-l f

Here is the same after the removal of the code:

no-animationi-l f

vogella:
As the visual representation is the same I plan to merge this request soon.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/TaskBarProgressManager.java
```diff
@@ -83,46 +80,17 @@ public TaskBarProgressManager(TaskItem taskItem) {
...
-						updateImage(job);
-					} else {
-						updateImage(null);
-					}
-				} else {
-					setProgressState(SWT.DEFAULT);
-					updateImage(null);
-				}
+				setProgressState(SWT.DEFAULT);
+				updateImage(null);
```
iloveeclipse, line 92: What is the purpose of the job now? If it does always same, what's the point to schedule it at all?

iloveeclipse: Who uses it amd wher / for what reason it is scheduled now?

laeubi:
> Like I explained in the commit message I cannot see a difference in behavior after removing the animation.

I think that the difference is that you now always have an "INDETERMINATE" progressbar but never showing actual progress (even though as @jukzi mentioned sometime wrong progress).

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/TaskBarProgressManager.java
vogella: Without the job you would not get an moving progress bar.

vogella:
As people seem not have the time to test (everyone seems to guess how it looks like), I merged, you can compare in the next I-build and see if something criticial is missing now.

jukzi:
@vogella The commit does not work as intended: The code was NOT about the wrong progress report image - that is still there. Instead it removed the little overlay icon inside the System status bar indicating the image of the currently running job - for example during manual clean build:
image

=> i ask to revert since https://bugs.eclipse.org/bugs/show_bug.cgi?id=471310 claims it was important for them.
I actually never understood or liked that "binary" overlay image during full build but maybe somebody put useful images there. Actually for example PDE export Plugin does try to setup a meaningful image - but it is actually never shown in the taskbar but only in the Progress view.

Since you removed the code for unnamed exceptions please submit the Stacktrace instead and i can look into it.

jukzi:
Also the change removed the working(!) Progress in the task bar.
The not working progress is instead: org.eclipse.ui.internal.progress.ProgressAnimationItem.bar - it does not even try to set a correct progress.

Bananeweizen:
> but maybe somebody put useful images there.

Have a look at the progress bar during any egit operation and be amazed. Unfortunately most projects don't call ProgressService.registerIconForFamily(...), even if they define job families.

PR #142, opened 2022-06-14
Remove System theme

### LINKED ISSUE #3: Disable or fix System theme on Windows
System theme looks not native on Windows. I suggest we remove it, users which want native styling can disable theming via unchekcing "Enable theming".

System theme

image

No theming:

image

mickaelistria:
The goal of the System theme was to fill the gap between default SWT
behavior and modern practices. But since such modern practices are
better to be built in SWT, it seems irrelevant and potentially harmful
to maintain the System theme as a workaround; efforts would be better
placed directly in SWT.


vogella:
@mickaelistria what happens if the user uses the system theme and updates his IDE? Please test this case otherwise the IDE might become unusable for former system theme users.

mickaelistria:
It fallbacks to Light theme.

Phillipus:
> efforts would be better placed directly in SWT.

Are those efforts underway? Just curious so I can test/help.

vogella:
@Phillipus https://github.com/eclipse-platform/eclipse.platform.ui/pull/14 could use some help / testing

mickaelistria:
> Are those efforts underway?

We recently updated the default CTabRenderer in SWT; and will probably make more changes. This CTabRenderer is the main part of SWT that is impacting them and that is not native. If you have suggestions of how to enhance it, please report to SWT; usually building a case of why the improvement should be made default (eg comparing with native win32/gtk/cocoa widgets, comparing with recommendations from those given reference frameworks, or other comparable ones).

Issue #127, opened 2022-06-10
Filter in Quick Access similar to VsCode


vogella:
In VsCode you can filter based on the category, e.g. Git: as in the following screenshot.

![image](https://user-images.githubusercontent.com/139910/173109654-e3c10d5a-4826-4ff7-b5b7-850582e840ba.png)

Would be super useful to have this also in Eclipse, e.g. for New: or View:

vogella:
Adding a few people which showed interest into usability of Eclipse cc @jukzi @mickaelistria @marcushoepfner

Bananeweizen:
If you just want to filter for the different kinds of quick assist entries (as indicated by your "New: or View:" example), that already works, without the colon (e.g. type "view" and you get all views, but the number is still limited, so you have to use ctrl-3 again to see all).
Coming back to your original request, wouldn't it be easier to just add the category of commands to the text that is being matched, instead of inventing a different kind of query? That way just typing "git" would find any git command, independent of whether it has "git" in the label/title.

vogella:
> If you just want to filter for the different kinds of quick assist entries (as indicated by your "New: or View:" example), that already works, without the colon (e.g. type "view" and you get all views, but the number is still limited, so you have to use ctrl-3 again to see all). Coming back to your original request, wouldn't it be easier to just add the category of commands to the text that is being matched, instead of inventing a different kind of query? That way just typing "git" would find any git command, independent of whether it has "git" in the label/title.

It is great to add additional elements but it would also be nice to be filter elements based on categories, I don't think this works at the moment:

![image](https://user-images.githubusercontent.com/139910/174253522-487b02e5-82cc-4c2e-801f-4a35ffd9f058.png)

Bananeweizen:
Yes, you still get the matches of the other categories, that's right. AFAIK there is currently no way to suppress matches, you will always get all matches.

akurtakov:
Is this one finished?

mickaelistria:
> Would be super useful to have this also in Eclipse, e.g. for New: or View:

Do you have an example of 1 action or command user is currently having hard time to find and that would be more efficiently found with the proposed filters.

PR #148, opened 2022-06-16
Match quick access input against command categories (#127)

Bananeweizen:
Quick access inputs like "Project", "Git" etc. should show all commands
of those categories, not only those that have these terms in their
labels.

When testing, the difference can be seen by using "project" (or other
command category names) as quick access input. Without the change only
commands with "project" in their label are shown. With this change also
commands like "Build Clean - Discard old build state" are shown, which
don't contain "Project" in the label, but in are in that category.

before:
image

after:
image

In the second image commands are visible that do not have "project" in their label, but are in that category.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/quickaccess/providers/CommandElement.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/quickaccess/providers/CommandElement.java
@@ -150,6 +150,16 @@ public class CommandElement extends QuickAccessElement {
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


vogella:
Thanks @Bananeweizen, can you add to the commit message how to test the differences? For example: "With this change you also see..." I tried it in a runtime Eclipse and AFAICS I do not see a difference.

vogella:
Thanks, works fine, will merge after verification build.

Please see also #127 do make Quick Access even more useful. I also think we should add a menu entry similar to VsCode to make it easier to discover this for VSCode user but I have not yet opened an issue for this.

image

vogella:
Thanks @Bananeweizen

PR #151, opened 2022-06-17
Contribute GH Build action to platform.ui

### LINKED ISSUE #150: Contribute GH build action
Would be nice to add a GH build action to platform.ui

vogella:
Adds GH verification in the hope that GH validation is faster than the
Jenkins one.

Fixes #150

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,46 @@
+name: Build Platform UI
+concurrency: 
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0 # required for jgit timestamp provider to work
+    - uses: actions/setup-java@v2
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: maven
+    - name: Build with Maven
+      uses: GabrielBB/xvfb-action@v1
+      with:
+       run: >- 
+        mvn
+        --batch-mode
+        --global-toolchains ${{ github.workspace }}/.github/workflows/toolchains.xml
+        -Pbuild-individual-bundles
+        -Pbree-libs
+        -Papi-check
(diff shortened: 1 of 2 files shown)


laeubi:
You should add the unitest.yml as well ...

.github/workflows/build.yml
```diff
@@ -0,0 +1,44 @@
...
+        mvn
+        --batch-mode
+        -Pbuild-individual-bundles
+        -Pbree-libs
+        -Papi-check
+        -Dcompare-version-with-baselines.skip=false
+        -Dproject.build.sourceEncoding=UTF-8
+        -Dorg.slf4j.simpleLogger.showDateTime=true 
+        -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS
+        -DtrimStackTrace=false
```
laeubi, line 39: You might want to add property t not fail on failing tests and using multiple threads.

vogella: I tried to use the same settings as Jenkins build file. We should change both if we want to but I would avoid doing this for the intial contribution.

laeubi: do not failed with  test ist a requirement for the unitest action to work.

laeubi:
You are missing the toolcahins.xml and toolchain maven option.

vogella:
> You are missing the toolcahins.xml and toolchain maven option.

As does the Jenkins file

akurtakov:
> > You are missing the toolcahins.xml and toolchain maven option.
> 
> As does the Jenkins file

JIPP auto inject the toolchain.xml so no need to handle it manually.

vogella:
> > > You are missing the toolcahins.xml and toolchain maven option.
> > 
> > 
> > As does the Jenkins file
> 
> JIPP auto inject the toolchain.xml so no need to handle it manually.

Ignore my deleted comment, I saw now that it was silly.

laeubi:
> Would be nice to have the complete build command in Jenkins build file too

Why? it doesn't make sense, as it *is* the complete commandline as jenkins uses some default locations that maven look at...

And it would harm because if you specify it manually maven would complain that the file do not exits :-)

Hopefully we will get 
- https://github.com/actions/setup-java/pull/282

so toolchains are setup automatically at GH, so please vote/comment/complain/... there to get more attention to this PR :-)

vogella:
Thanks @laeubi and @akurtakov

PR #153, opened 2022-06-17
Avoid duplicate quick access entry (fixes #152)

### LINKED ISSUE #152: Quick access twice the same entry
image

@Bananeweizen as you recently worked on Quick access, do you also see this?

Bananeweizen:
When you have used quick access previously, the dialog adds matching
entries from "previous" at the top. Afterwards the same entries may
appear from their "normal" provider (e.q. views, commands, ...).

Since the logic of adding entries seems quite complex, this change fixes
the duplicate entry by detecting the exact situation (same single entry
in first and second provider list), and then removes the duplicate entry
list.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/quickaccess/QuickAccessContents.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/quickaccess/QuickAccessContents.java
@@ -527,6 +527,14 @@ public abstract class QuickAccessContents {
 					QuickAccessEntry.MATCH_PERFECT)));
 		}
 		res.addAll(entriesPerProvider.values());
+
+		// if one category provides the same single entry like "previous", we can avoid
+		// showing the duplicate second list
+		if (res.size() >= 2 && res.get(0).size() == 1 && res.get(1).size() == 1
+				&& (res.get(0).get(0).element.equals(res.get(1).get(0).element))) {
+			res.remove(1);
+		}
+
 		return (List<QuickAccessEntry>[]) res.toArray(new List<?>[res.size()]);
 	}
 


mickaelistria:
In my use-case reported in #152, I see that both entries are "Previous" so it seems like the issue is something we can better fix earlier without implementing a dedicated mitigation later. Maybe the Previous provider hosts a bug?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/quickaccess/QuickAccessContents.java
```diff
@@ -530,6 +530,14 @@ public IStatus runInUIThread(IProgressMonitor monitor) {
 					QuickAccessEntry.MATCH_PERFECT)));
 		}
 		res.addAll(entriesPerProvider.values());
+
+		// if one category provides the same single entry like "previous", we can avoid
+		// showing the duplicate second list
+		if (res.size() >= 2 && res.get(0).size() == 1 && res.get(1).size() == 1
```
Wittmaxi, line 533: I there a good way to increase clarity by extracting these boolean-statements into either a method or separate variables? For example, a 
```
boolean isDuplicateCategory = ...
```
would help make the code speak for itself

elsazac: just wondering , what is the next action on this PR @Wittmaxi is the above suggestion a blocker ?

Wittmaxi: @elsazac no, I was just wondering if the code-quality could be improved. I think this is fine

elsazac: This can have some reviews.

PR #155, opened 2022-06-18
Replace Required-Bundle o.e.osgi.services and remove unused dependencies

HannesWell:


HannesWell:
Unfortunately I cannot reproduce the failure locally. They always pass, although I use a minimal set of bundles launched into the test application.
Lets see if this is the first or the second commit.

HannesWell:
Btw. it looks like that the GH build succeeds even if there are test-failures. That is very likely because of the `-Dmaven.test.failure.ignore=true` CLI argument. But there should be a later step that finally fails the build (I think the intention is to run all tests and only fail in the very end to get all results in one build).

laeubi:
The repository is missing the junit workflow to parse the junit results and publish the report.

pbodnar:
@HannesWell, maybe a dump question, but I'm still new to this Eclipse "setup" stuff... In 4f05fba98a8643d44377b2160e9a5af282be486a "Remove unused required bundles" you remove for example the dependency on "org.eclipse.ui.workbench". Yet apparently, this bundle is still required by for example "/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java". So do I get it right that the bundle usage is still there, but it is defined _just transitively_ via "/org.eclipse.ui/META-INF/MANIFEST.MF" now?

Due to changes #12 I was about to increase the version of "org.eclipse.ui.workbench" in the modified [/org.eclipse.ui.ide/META-INF/MANIFEST.MF](https://github.com/eclipse-platform/eclipse.platform.ui/pull/155/commits/4f05fba98a8643d44377b2160e9a5af282be486a#diff-c4df1bf28183cd43d51097d82a3f2bd91c7db0cd3ec893bd07ae9d4dfc2894cb), so should I do this in "/org.eclipse.ui/META-INF/MANIFEST.MF" instead?

HannesWell:
> @HannesWell, maybe a dump question, but I'm still new to this Eclipse "setup" stuff... In [4f05fba](https://github.com/eclipse-platform/eclipse.platform.ui/commit/4f05fba98a8643d44377b2160e9a5af282be486a) "Remove unused required bundles" you remove for example the dependency on "org.eclipse.ui.workbench". Yet apparently, this bundle is still required by for example "/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java". So do I get it right that the bundle usage is still there, but it is defined _just transitively_ via "/org.eclipse.ui/META-INF/MANIFEST.MF" now?

That's not a dump question (and its arguable if there are any), don't hesitate to ask if something is unclear :)
And yes since `org.eclipse.ui` requires and reexports `org.eclipse.ui.workbench` and `org.eclipse.ui.ide` requires `org.eclipse.ui`, `org.eclipse.ui.ide` still requires `org.eclipse.ui.workbench` transitively.

> 
> Due to changes #12 I was about to increase the version of "org.eclipse.ui.workbench" in the modified [/org.eclipse.ui.ide/META-INF/MANIFEST.MF](https://github.com/eclipse-platform/eclipse.platform.ui/pull/155/commits/4f05fba98a8643d44377b2160e9a5af282be486a#diff-c4df1bf28183cd43d51097d82a3f2bd91c7db0cd3ec893bd07ae9d4dfc2894cb), so should I do this in "/org.eclipse.ui/META-INF/MANIFEST.MF" instead?

Yes you should increase the lower bound of the requirement to `org.eclipse.ui.workbench` in `org.eclipse.ui`. And IIRC you should also increase the minor version of `org.eclipse.ui` (that's the bad thing about re-exporting).

PR #163, opened 2022-06-20
Css styling for toolitems

vogella:


vogella:
@nnemkin let me know if you have any comments.

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ToolItemElement.java
```diff
@@ -27,14 +30,51 @@
  *
  */
 public class ToolItemElement extends ItemElement {
+
+	private boolean isSelected = false;
+
+	boolean dynamicEnabled = true;
```
laeubi: probably debugging code?

tests/org.eclipse.e4.ui.tests.css.swt/src/org/eclipse/e4/ui/tests/css/swt/ToolItemTest.java
```diff
@@ -0,0 +1,94 @@
+/*******************************************************************************
+ * Copyright (c) 2008, 2014 IBM Corporation and others.
```
laeubi: If this file is new and you are not working for IBM since 2008, the header is probably wrong :-)

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ToolItemElement.java
vogella: Thanks, fixed.

tests/org.eclipse.e4.ui.tests.css.swt/src/org/eclipse/e4/ui/tests/css/swt/ToolItemTest.java
vogella: Thanks, fixed

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ToolItemElement.java
```diff
@@ -27,14 +30,50 @@
...
 public class ToolItemElement extends ItemElement {
+
+	private boolean isSelected = false;
+
+	boolean dynamicEnabled = Boolean.getBoolean("org.eclipse.e4.ui.css.dynamic");
+
+	private SelectionListener selectionListener = new SelectionAdapter() {
+		@Override
+		public void widgetSelected(SelectionEvent e) {
+			if (!e.widget.isDisposed()) {
```
nnemkin: There's no need to check for disposed in an event handler, events aren't delivered to disposed widgets. (Otherwise it's a bug in SWT).

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/helpers/CSSSWTColorHelper.java
```diff
@@ -351,7 +352,13 @@ private static RGBA findColorByDefinition(String name) {
 	}
 
 	/** Helper function to avoid setting colors unnecessarily */
-	public static void setForeground(Control control, Color newColor) {
+	public static void setForeground(Control toolItem, Color newColor) {
+		if (!Objects.equals(toolItem.getForeground(), newColor)) {
```
nnemkin: There is no need to guard against setting the same value, `ToolItem` implementations on all platforms already do this internally. Setting the same color repeatedly won't lead to unnecessary repaints.

Also, in SWT get and set for foreground/background aren't symmetric. get methods will delegate to the widget parent if the property isn't set, i.e. getForeground() actually means getEffectiveForeground(). The check above will incorrectly prevent setting the color if it matches the parent color. (Internal checks in SWT don't suffer from this.)

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/helpers/CSSSWTColorHelper.java
```diff
@@ -364,6 +371,12 @@ public static void setBackground(Control control, Color newColor) {
 		}
 	}
 
+	public static void setBackground(ToolItem toolItem, Color newColor) {
+		if (!Objects.equals(toolItem.getBackground(), newColor)) {
```
nnemkin: Same note as above, this check is unnecessary.

nnemkin (commented):
Hi Lars. I've added some comments regarding SWT behavior.

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/helpers/CSSSWTColorHelper.java
vogella: Thanks, in this case I will move the calls to the corresponding CSS handlers.

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ToolItemElement.java
vogella: Thanks, updated

PR #175, opened 2022-06-27
make use of githubs dependency analysis and codeql check

carstenartur:
This add a simple run to give you an overview about the dependencies and adds an action to run codeql on eclipse.platform.ui.
It is quite fast and should only add few false positives that can easily be ignored using the given user interface.


iloveeclipse:
Could you please provide a link to an example of what this action is supposed to do? Which dependencies are analyzed? Where is the report shown? How and what is this supposed to improve?

carstenartur:
> Could you please provide a link to an example of what this action is supposed to do? Which dependencies are analyzed? Where is the report shown? How and what is this supposed to improve?

These two changes are part of a fork you can see here
https://github.com/carstenartur/eclipse.platform.ui/network/dependencies
https://github.com/carstenartur/eclipse.platform.ui/security/code-scanning
Is that sufficient to get the idea?

iloveeclipse:
> These two changes are part of a fork you can see here
> https://github.com/carstenartur/eclipse.platform.ui/network/dependencies

This shows some error saying the graph is too big to be rendered. 

> https://github.com/carstenartur/eclipse.platform.ui/security/code-scanning

This link points to 404 error, probably not allowed due some permissions 

> Is that sufficient to get the idea?

Not really. So is this about to show maven dependencies in some kind of github report?

carstenartur:
> > These two changes are part of a fork you can see here
> > https://github.com/carstenartur/eclipse.platform.ui/network/dependencies
> 
> This shows some error saying the graph is too big to be rendered.
> 
It might complain for some poms but still can work with the dependencies where the limits do not get exceeded.
It is a precondition to have this dependency analysis for using the github feature to automatically find newer versions, create and run pull requests based on the new versions so that you than can simply merge after you checked that it run succesfully through all build and checking steps.

> > https://github.com/carstenartur/eclipse.platform.ui/security/code-scanning
> 
> This link points to 404 error, probably not allowed due some permissions
> 

I added a invitation, maybe then you can see it. It is a feature that security issues are invisible until you are allowed to see them eg. by entering a project.

> > Is that sufficient to get the idea?
> 
> Not really. So is this about to show maven dependencies in some kind of github report?

Full documentation you can find here
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

However this feature only can be used where the dependency is defined within the pom tree of the eclipse project (eclipse.platform.ui). As eclipse.platform.ui defines everything in parent poms it cannot make use of this specific feature.
But you can still see most of the dependencies.

PR #181, opened 2022-06-30
Update the static MessageDialog#openQuestion to UX guidelines from the OS

vogella:
Operating system Ui guidelines do not recommend the usage of the help /
question
icon in a dialog. This change adjusts the MessageDialog#openQuestion
accordingly.As this will affect all callers of this method, it has will
lead a consistency for callers, users which want / need to violate UX
guidelines of the OS can switch to the non static methods or use
PlainMessageDialog to construct their dialog in a way which they want.

The reasoning for this change is that platform is allowed to adjust its
UI to UX changes in the OS and that the UX of Eclipse is not considered
API.

For example check the screenshots in the Windows
guidelines for dialogs.

Operating System UI Guidelines

Windows UI guidence
https://docs.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs

Gnome UI guidence
https://developer.gnome.org/hig/patterns/feedback/dialogs.html

Mac UI guidence
https://developer.apple.com/design/human-interface-guidelines/components/presentation/alerts/

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/MessageDialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/MessageDialog.java
@@ -64,8 +64,8 @@ public class MessageDialog extends IconAndMessageDialog {
 	public static final int INFORMATION = 2;
 
 	/**
-	 * Constant for the question image, or a simple dialog with the question
-	 * image and Yes/No buttons (value 3).
+	 * Constant for the question image, or a simple dialog without an icon and
+	 * Yes/No buttons (value 3).
 	 *
 	 * @see #MessageDialog(Shell, String, Image, String, int, int, String...)
 	 * @see #open(int, Shell, String, String, int)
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/MessageDialog.java
```diff
@@ -246,7 +247,6 @@ private void init(String dialogTitle, Image dialogTitleImage, String dialogMessa
 		case QUESTION:
 		case QUESTION_WITH_CANCEL:
 		case CONFIRM: {
-			this.image = getQuestionImage();
```
laeubi, line 249: Maybe we should leave "CONFIRM" as is and jsut break at QUESTION/WITH_CANCEL? Or even use JFacePreferences to decide if icon should be shown or not?

vogella: I don't see an example for confirmation dialog using a question / help icon in the guidelines https://docs.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs Check for section "Confirmation dialogs (OK/Cancel)".

PR #182, opened 2022-07-01
WIP - Moving the perspective switcher to the left trimbar for new

vogella:
workspaces

Tools like VsCode, the upcoming Thunderbird, etc place their
"perspective switcher" to the left side of the application bar. With
this change we do the same for new workspaces.

This is currently a proof-of-concept, AFAIK Gayan is looking into
updating the UX of Eclipse to a more modern standard and I hope this
help him in his evaluation.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchWindow.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchWindow.java
@@ -1021,7 +1021,7 @@ public class WorkbenchWindow implements IWorkbenchWindow {
 		getCoolBarManager2().update(true);
 		getCoolBarManager2().add(new GroupMarker(IWorkbenchActionConstants.MB_ADDITIONS));
 
-		final MTrimBar trimBar = getTopTrim();
+		final MTrimBar topTrimBar = getTopTrim();
 		// TODO why aren't these added as trim contributions
 		// that would remove everything from this method except the fill(*)
 		/*
@@ -1042,10 +1042,12 @@ public class WorkbenchWindow implements IWorkbenchWindow {
 					"bundleclass://org.eclipse.e4.ui.workbench.renderers.swt/org.eclipse.e4.ui.workbench.renderers.swt.LayoutModifierToolControl"); //$NON-NLS-1$
 			spacerControl.getTags().add(TrimBarLayout.SPACER);
 			spacerControl.getTags().add("SHOW_RESTORE_MENU"); //$NON-NLS-1$
-			trimBar.getChildren().add(spacerControl);
+			topTrimBar.getChildren().add(spacerControl);
 		} else if (!spacerControl.getTags().contains("SHOW_RESTORE_MENU")) { //$NON-NLS-1$
 			spacerControl.getTags().add("SHOW_RESTORE_MENU"); //$NON-NLS-1$
 		}
+
+		final MTrimBar leftTrimBar = getLeftTrim();
 
 		MToolControl switcherControl = (MToolControl) modelService.find("PerspectiveSwitcher", model); //$NON-NLS-1$
 		if (switcherControl == null && getWindowConfigurer().getShowPerspectiveBar()) {
@@ -1057,10 +1059,10 @@ public class WorkbenchWindow implements IWorkbenchWindow {
 			switcherControl.getTags().add("SHOW_RESTORE_MENU"); //$NON-NLS-1$
 			switcherControl.setContributionURI(
 					"bundleclass://org.eclipse.ui.workbench/org.eclipse.e4.ui.workbench.addons.perspectiveswitcher.PerspectiveSwitcher"); //$NON-NLS-1$
-			trimBar.getChildren().add(switcherControl);
+			leftTrimBar.getChildren().add(switcherControl);
 		} else if (switcherControl != null) {
 			if (!getWindowConfigurer().getShowPerspectiveBar()) {
-				trimBar.getChildren().remove(switcherControl);
+				leftTrimBar.getChildren().remove(switcherControl);
 			} else {
 				List<String> tags = switcherControl.getTags();
 				if (!tags.contains("HIDEABLE")) { //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


mickaelistria:
I like this proposal, I actually have my perspective switcher on the left and am happy with it.
What do you think would be a reason for not merging it, or keeping it as a draft?

laeubi:
@vogella can you share a screen? the "right" seems a good choice for mee because there is usually some unused space and its easy to find/use it.

I find it a bit hard to decide just because other UI do it, because each UI has different demands, so showing some before / after and also screenshots of the other UIs would help here.

merks:
I can imagine among the millions of users who have used eclipse for years (or decades) that there will be a highly significant number of unhappy people if you move this often-used thing from a very familiar place to some other unfamiliar place.  Most certainly I for one would not be happy with such a change.  Of course one can move it, but that's more a good argument to leave it where it's always been than the initialize it to a different location...

trancexpress:
> I can imagine among the millions of users who have used eclipse for years (or decades) that there will be a highly significant number of unhappy people if you move this often-used thing from a very familiar place to some other unfamiliar place.

Especially considering Eclipse will not end up looking "modern" on the long run.

trancexpress:
Sorry for closing this, misclick.

merks:
I hope it gets closed.  😈

mickaelistria:
It's always the same debate between people who want to change to more modern XYZ (be it UI patterns, library, Java versions...) because immobility is a risk, and those who think nothing should change ever because every change is a risk. Note that as far as I know, there is nothing in the project charter nor in EDP that can guide such a decision, it's pure project decision.

It's not black and white and every time the debate is necessary and useful. When we're in such case, particularly for UI which is by nature more subjective, I think we need to get a clear idea of the proposals (eg screenshots), collect pros and cons aguments, synthesize them and organize a vote for committers to decide what we do about the project.
So let's keep debating, collecting data, pros and cons arguments, and sometimes later, we'll run a vote if it seems helpful.

laeubi:
> It's not black and white and every time the debate is necessary and useful.

Well problem currently is, there is even no black nor white... so for UI changes I find it essential to provide SCREENSHOTS so one can much better decide... and its incredible easy to share some visuals with Github now!

Claiming it is more "modern" to move one thing from right to left without any showcase or reference (e.g. I never use VSCode no idea how it looks there...) is likely to not give good results.

As a side (I think I noted that earlier), for me Eclipse is a Tool to work with it don't has to look nice, so if moving that thing will improve the workflow (I'm currently not sure it does) I think users will adapt, if it just looks different because others do it that way (I really find especially Thunderbird/Firefox getting more and more useless for me and I always need to search where things are now hidden!) will not help, so I prefer an IDE that might looks 90s but works like 2050 over the other way round...

merks:
@mickaelistria 

There is an underlying assertion/assumption in your first paragraph that this change makes it a more modern. This suggests that some people are on the side of modern and others are on the opposite side of modern, and not just on the opposite side of modern but so far on that side that nothing should ever change.  This paints the entire discussion in a black or white context where one side is clearly being characterized as the better (modern) side.

I agree with **everything** your in your second paragraph though!  😄  If there are good reasons to change something for the better, then let's do so.

mickaelistria:
> There is an underlying assertion/assumption in your first paragraph that this change makes it a more modern

Indeed, I should have written that to mimics the approach of the most popular IDE of modern times (vscode). I'll try to be more careful and avoid the term "modern" as it is indeed biased. FYI, I don't use it intentionally as a synonymous of good, and "not modern" doesn't imply bad to me; I often like "good old things" more than "modern" ones ;)

trancexpress:
> As a side (I think I noted that earlier), for me Eclipse is a Tool to _work_ with it don't has to _look_ nice, so if moving that thing will improve the workflow (I'm currently not sure it does) I think users will adapt, if it just looks different because others do it that way (I really find especially Thunderbird/Firefox getting more and more useless for me and I always need to search where things are now hidden!) will not help, so I prefer an IDE that might looks 90s but works like 2050 over the other way round...

Same for me. Though Eclipse doesn't work like 2050, so moving things around is even harder to get behind.

I'm encountering enough Eclipse bugzilla bugs where the reporter commented they moved to another IDE because Eclipse got too buggy. I'm yet to see a comment where a reporter states, they moved to another IDE because Eclipse looked old. But I guess it doesn't hurt to have an attempt to make Eclipse more "modern", at most it alienates a few long-term users and speeds up the process to moving to other IDEs.

Sorry for editing your comment, I'm still getting used to GitHub "features" and apparently I'm not awake yet.

laeubi:
I think the opposite of _modern_ is _retro_ what not per se "bad" but, well "modern" today, so probably we can just rename Eclipse IDE to be the _Retro IDE_ ... :-) 

I think it is bad to replicate what others do (even though they claim to be _most popular IDE of modern times_) in that case it would be better to use that "better IDE" and invest the effort we do here into other areas, so if we find that it makes Eclipse better and more unique if we do something that *also* other do, great, but we should not do it **because** others do so!

mickaelistria:
@laeubi in my first comment, I have clarified that I am using a perspective switcher on the left and am liking it. So yes, to me it makes Eclipse better. Please focus on the actual proposal rather than creating general and fruitless discussions about how you believe people should think.

akurtakov:
I will just drop this one here to both show how perspective switcher on the left looks like and that it's definetely not smth "modern" either - https://www.eclipse.org/articles/Article-UI-Guidelines/Contents.html . 
Some of us still remember it :smile:

mickaelistria:
Anyone who wants to give it a try can do so in their own IDE, by just drag-and-dropping the perspective switcher to the left (this may be hard on Linux because of https://github.com/eclipse-platform/eclipse.platform.swt/issues/589 but still it's doable with some patience and meticulousness).
I would highly recommend anyone to give it a try, just moving it for a couple of days to consolidate or change their opinion; and then to come back to the topic sharing their concrete feedback based on that experience. Once we have that, we will be able to evaluate if some consensus arise in 1 direction or the other, or if a vote seems necessary to ensure a fair decision & course of action.

iloveeclipse:
I''ve tried and I don't like it because it wastes space
image

Phillipus:
Are ordinary Eclipse users allowed to express an opinion on this? If so, then please don't do this.

(Eclipse user since 2003).

akurtakov:
This has stayed open long enough and nothing happened, it's time to close it.

PR #184, opened 2022-07-02
Make matches highlighting in Resource dialog more simple and efficient

pbodnar:
This effectively removes and/or replaces most of the code introduced
within the solutions of bug 520250 "Open resource should use highlight
matches based on CamelCase" and follow-up bugs 528301 and 529451. The
goal is more maintainable and readable code, while also having the
implementation more efficient.

1. Remove redundant conditions, escaping and usage of regex (from top to
bottom):

   * Remove `escapeRegexCharacters()` and follow-up de-escaping of
characters, as treating the search string as a plain string seems to be
sufficient.
   * Remove `searchFieldString.lastIndexOf('.') != -1`, as this is
implied when `str.getStyleRanges().length > 0` is true.
   * Remove ` + resource.getFileExtension()`, as passing just `'.'` to
`lastIndexOf()` should do the very same job in a simpler fashion.
   * Every matching string `region` is used as a simple string. It is
unclear why handling it as a regex instead was introduced.
   * Remove the `restart` flag and do-while around it. It is unclear
what exact scenarios this should have covered. Also, incrementing
`currentIndex` just by 1 lead to repeated failing `indexOf()` lookups
being performed in loop for simple search strings.

2. Choose simpler ways of calculations (from top to bottom):

   * `resource.getName().lastIndexOf(resourceExtension)` seems to be a
less efficient variant of `resource.getName().length() -
resourceExtension.length()`.
   * rename: `lastDotIndex` -> `matchingExtensionIndex` for consistency
   * `(\\?)|\\*` replaced by a less cryptic `[?*]`

3. Add explanatory comments

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
@@ -23,13 +23,10 @@ import java.io.StringReader;
 import java.io.StringWriter;
 import java.text.Collator;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IResource;
@@ -645,7 +642,9 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 			}
 
 			IResource resource = (IResource) element;
-			String searchFieldString = escapeRegexCharacters(((Text) getPatternControl()).getText());
+			String searchFieldString = ((Text) getPatternControl()).getText();
+
+			// We highlight just the matches in resource name, so cut off path if any
 			int fileNameIndex = searchFieldString.lastIndexOf('/');
 			if (fileNameIndex != -1 && fileNameIndex != searchFieldString.length() - 1) {
 				searchFieldString = searchFieldString.substring(fileNameIndex + 1);
@@ -657,9 +656,11 @@ public class FilteredResourcesSelectionDialog extends FilteredItemsSelectionDial
 					textStyle.font = JFaceResources.getFontRegistry().getBold(JFaceResources.DEFAULT_FONT);
 				}
 			};
+
+			// Firstly style the extension part (if any), then style the rest
 			StyledString str = styleResourceExtensionMatch(resource, searchFieldString, boldStyler);
-			if (str.getStyleRanges().length != 0 && searchFieldString.lastIndexOf('.') != -1) {
-				resourceName = resourceName.substring(0, resourceName.lastIndexOf('.' + resource.getFileExtension()));
+			if (str.getStyleRanges().length > 0) {
+				resourceName = resourceName.substring(0, resourceName.lastIndexOf('.'));
 				searchFieldString = searchFieldString.substring(0, searchFieldString.lastIndexOf('.'));
 			}
 			getMatchPositions(resourceName, searchFieldString).stream()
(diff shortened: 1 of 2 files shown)


pbodnar:
Please take this as a possible initial input for further discussion. The thing is that all unit tests still pass even after removing all the older "for me hard-to-explain" implementation logic. So either not all presumed scenarios were covered by appropriate unit tests (and I'm not currently seeing those possibly edge case scenarios), or the simplification is really correct & the PR can be possibly safely merged. 

Maybe @LucasBullen, could you please have a look at this, as the one who has seemingly got some experience around this stuff?

vogella:
@[pbodnar](https://github.com/pbodnar) cool, code simplifications are always welcome. Can you point me to the relevant unit test class?

pbodnar:
> @[pbodnar](https://github.com/pbodnar) cool, code simplifications are always welcome. Can you point me to the relevant unit test class?

@vogella, thanks for looking at this. The tests should be there in the `ResourceItemLabelTest` which I have also extended a little bit in this PR. There are dedicated tests which have numbers of the aforementioned bugs in their name.

vogella:
Thanks @pbodnar , sorry for my silly question I should have seen the unit test file in the list of changed files. 

LGTM, I also tested this in a runtime IDE and it works for me as before. I plan to merge this after the freeze period ends. I my vacation starts soon, I might have to merge this from my mobile or after my vacation if nobody want to merge it. I assume merging via mobile is OK as I already tested it AND this change adds one new unit test and the rest of the unit tests is also fine.

 Thanks @pbodnar again, great to see some movement here.

vogella (approved):

pbodnar:
@vogella, thanks for your testing and for pushing it further. :)

FYI, I have just also checked how they do this matches highlighting in **JDT**, i. e. in the analogical `FilteredTypesSelectionDialog`. And as I have expected, _they do it quite differently_ - so just for the record, because bringing the implementations somewhat closer to each other (and sharing common code) seems to be a mission impossible (luckily it is not of a high priority anyway, I guess), the pretty sophisticated methods for JDT are located inside the [StringOperation](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/StringOperation.java) class - `getCamelCaseMatchingRegions()` and `getPatternMatchingRegions()`.

Personally, I find the solution in the PR-tuned `FilteredResourcesSelectionDialog.ResourceItemLabelProvider.getMatchPositions()` much more elegant, yet hopefully the authors of JDT had good reasons for choosing a different way. I also find it interesting that both the _filtering_ and _highlighting_ methods are basically doing the same job. Moreover in the case of JDT, they are very similar in implementation. Interesting area for research, but yes, I would rather "not touch what works" in this case, you know... ;)

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/FilteredResourcesSelectionDialog.java
```diff
@@ -676,93 +677,72 @@ public void applyStyles(TextStyle textStyle) {
...
+			if (lastChar == ' ' || lastChar == '<') {
 				matching = matching.substring(0, matching.length() - 1);
-			} else if (lastChar == '<') {
-				matching = matching.substring(0, matching.length() - 2);
 			}
+
+			String originalTrimmedMatching = matching;
 			if (matching.indexOf('?') == -1 && matching.indexOf('*') == -1) {
+				// Expect that CamelCase search was possibly performed
+				// => convert pattern to "wildcards-like" one
```
pbodnar: @vogella, after waking up, I've realized that we could probably make this even simpler before merging into master. Now we split the CamelCase pattern and join it by `*`, only to split it according to the very same character later on. So what about leaving just the 1st splitting step... I would like to look at it this or tomorrow evening (CET) if I have time.

pbodnar: Done, see the force-push, hopefully correct. :)

vogella:
Thanks @pbodnar

pbodnar:
Thank you :)

PR #187, opened 2022-07-05
Using DI for creating the ShowViewDialog

vogella:
Instead of getting the parameters in our handler for calling the
constructor of ShowViewDialog we can also use DI do to this, the
resulting code is slightly shorter. No big win but a tiny move for using
more DI in internal Eclipse code.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/handlers/ShowViewHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/handlers/ShowViewHandler.java
@@ -19,17 +19,15 @@ import java.util.List;
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.e4.core.contexts.ContextInjectionFactory;
 import org.eclipse.e4.core.contexts.IEclipseContext;
 import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.e4.ui.model.application.descriptor.basic.MPartDescriptor;
 import org.eclipse.e4.ui.model.application.ui.advanced.MPlaceholder;
 import org.eclipse.e4.ui.model.application.ui.basic.MPart;
-import org.eclipse.e4.ui.model.application.ui.basic.MWindow;
-import org.eclipse.e4.ui.workbench.modeling.EModelService;
 import org.eclipse.e4.ui.workbench.modeling.EPartService;
 import org.eclipse.e4.ui.workbench.modeling.EPartService.PartState;
 import org.eclipse.jface.window.Window;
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/handlers/ShowViewHandler.java
```diff
@@ -95,14 +93,8 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
...
 	 */
-	private static void openOther(ExecutionEvent event, IWorkbenchWindow workbenchWindow, MApplication app,
-			EPartService partService) {
-		Shell shell = HandlerUtil.getActiveShell(event);
-		IEclipseContext ctx = workbenchWindow.getService(IEclipseContext.class);
-		EModelService modelService = workbenchWindow.getService(EModelService.class);
-		MWindow window = workbenchWindow.getService(MWindow.class);
-
-		final ShowViewDialog dialog = new ShowViewDialog(shell, app, window, modelService, partService, ctx);
+	private static void openOther(IWorkbenchWindow workbenchWindow, IEclipseContext ctx, EPartService partService) {
```
laeubi, line 96: Should not only `IEclipseContext ctx` be required here then?

laeubi:
This is really useful and you can also use this to inject OSGi Services into UI components without using static access then!

I wonder if it would make sense to even replace `ContextInjectionFactory.make(..., IEclipseContext)` by an E4 service that could be injected instead (DependecyInjector?) that one could use for such purpose as an alternative to the `IEclipseContext`, I find this passing of `IEclipseContext` around and using static `ContextInjectionFactory` defeats the purspose of DI, also it would be usefull to have then a Method that takes a VarArg of additional parameters to pass as the additional context, then client code do not need to bind to `ContextInjectionFactory`/`IEclipseContext`/`EclipseContextFactory`

vogella:
> This is really useful and you can also use this to inject OSGi Services into UI components without using static access then!

Glad you like it. Would be great if this API could be used more in Eclipse to reduce code.

> I wonder if it would make sense to even replace `ContextInjectionFactory.make(..., IEclipseContext)` by an E4 service that could be injected instead (DependecyInjector?) that one could use for such purpose as an alternative to the `IEclipseContext`, I find this passing of `IEclipseContext` around and using static `ContextInjectionFactory` defeats the purspose of DI, also it would be usefull to have then a Method that takes a VarArg of additional parameters to pass as the additional context, then client code do not need to bind to `ContextInjectionFactory`/`IEclipseContext`/`EclipseContextFactory`

I don't mind, getting the fitting IEclipseContext might be a challenge. Please open a new ticket / PR for this idea if you plan to work on this.

akurtakov:
I've seen some activity about more E4 lately and wondered whether you guys are interested in finishing this one?

akurtakov:
Any plans to finish this one?

PR #189, opened 2022-07-05
Updating the activity tests to use static imports instead of relying on

vogella:
JUnit3 inheritance

Currently UITestCase inherits the JUnit3 Test class even though it uses
JUnit4 features. This allows extenders of this class to use JUnit3
assert methods directly. Using static imports for this would allow at
some point to cut the dependency to JUnit3 for the UI tests.
This change is a tiny change in this directly as it only updates the
activity tests.


vogella (commented):
Needs also other changes, so this change it not yet ready.

PR #191, opened 2022-07-05
Allows to open the intro view via the Show view entry

vogella:
For UX consistency we should not handle some views special. Before this
change the intro view could not be opened via show view because it had
its only menu entry. Such behavior is not intuitive for the end user, so
this commit removes this restriction.


vogella (commented):
Planning to merge after m1

PR #193, opened 2022-07-05
Resolve E3 views not being filtered by activities

ghost:
The view dialog will always contain all available E3 views, even if some
of them have been disabled via activities.
This is because the method isFilteredByActivity uses a list of filtered
views to determine, whether the provided element should be displayed in
the dialog.
Because the element is not contained in this list, the method is unable
to detect filtered activities and therefore always returns false.
An additional check ensures that this filtering isn't done to E4 views,
as they are also not contained in the view registry.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewContentProvider.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewContentProvider.java
@@ -23,9 +23,8 @@ import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.e4.ui.model.application.descriptor.basic.MPartDescriptor;
 import org.eclipse.jface.viewers.ITreeContentProvider;
 import org.eclipse.jface.viewers.Viewer;
-import org.eclipse.ui.activities.WorkbenchActivityHelper;
 import org.eclipse.ui.internal.WorkbenchPlugin;
-import org.eclipse.ui.views.IViewDescriptor;
+import org.eclipse.ui.internal.e4.compatibility.CompatibilityPart;
 import org.eclipse.ui.views.IViewRegistry;
 
 /**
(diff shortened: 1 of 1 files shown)


ghost:
After some digging, it seems like this behaviour is an unintended side-effect of [this](https://github.com/eclipse-platform/eclipse.platform.ui/commit/ba20d05731c7e31d4909888f4885a0a351dd090a) bugfix, as after that, only a subset of all available views are used by isFilteredByActivity(...).

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewContentProvider.java
```diff
@@ -175,18 +174,20 @@ private boolean isIntroView(String id) {
...
-				return true;
-			}
-		}
-		return false;
+	private boolean isFilteredByActivity(MPartDescriptor descriptor) {
+		// viewRegistry.find(...) already applies a filtering for disabled views
+		boolean isFiltered = viewRegistry.find(descriptor.getElementId()) == null;
+
+		// E3 views can be detected by checking whether they use the compatibility layer
+		boolean isE3View = descriptor.getContributionURI().equals(CompatibilityPart.COMPATIBILITY_VIEW_URI);
```
vogella: In general better to use the constant first, e.g.. CompatibilityPart.COMPATIBILITY_VIEW_URI.equals....

ghost: Fixed.

vogella:
Can you solve the conflict @pzi-dsa ?

ghost:
> Error: ERROR] [ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='org.eclipse.ui:org.eclipse.ui.monitoring.tests:1.2.0-SNAPSHOT'}' and 'Vertex{label='eclipse.platform.ui:org.eclipse.e4.ui.css.swt:0.14.600-SNAPSHOT'}' introduces to cycle in the graph eclipse.platform.ui:org.eclipse.e4.ui.css.swt:0.14.600-SNAPSHOT --> org.eclipse.ui:org.eclipse.ui.monitoring.tests:1.2.0-SNAPSHOT --> eclipse.platform.ui:org.eclipse.e4.ui.css.swt:0.14.600-SNAPSHOT @ 

I've no idea what's causing this. I haven't even touched the Manifest file, let alone those plugins...

laeubi:
@pzi-dsa never mind this is not related to your change see https://github.com/eclipse/tycho/issues/1124

vogella:
Thanks @pzi-dsa Should be available in tomorrows I-build.

PR #200, opened 2022-07-10
Remove commons.io from e4.rcp feature

mickaelistria:
That will facilitate adoption of JUnit 5 from Maven Central

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -186,13 +186,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.apache.commons.io"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
          id="org.apache.commons.logging"
          download-size="0"
          install-size="0"


merks:
Does this imply that the source will be missing from the Eclipse SDK?

mickaelistria:
Actually, this artifact is not used anywhere in the SDK nowadays, the next step is to remove it from .target.

merks:
That's good.

Note that I'm held up with getting the SDK aggregator working primarily because the validation update site (Orbit) contains IUs equivalent to the ones coming directly coming from Maven, but the direct-from-Maven ones look **older** than the Orbit one because there is no version qualifier.  This confuses the aggregator to think it should ignore the older one in favor of the Orbit one, and then one continues to need the mapping rules for the Orbit versions...

mickaelistria:
I tried to keep the provider change apart from a version update to make the change simpler to validate. If there are particular deps you want to bump to a newer version, you can just submit PRs to update the .target. This is exactly why we're making those changes.

mickaelistria:
Apparently commons-io is required as transitive deps of batik. So let's update the TP to include a recent version of commons-io and then we'll revert this patch.

PR #203, opened 2022-07-12
Revert "Remove commons.io from e4.rcp feature"

mickaelistria:
Reverts eclipse-platform/eclipse.platform.ui#200


mickaelistria:
I don't think we need to revert this change, as things work well with whichever version of commons-io thanks to the previous commit and no issue was found. Let's close.

PR #211, opened 2022-07-15
Make E4 more accessible in org.eclipse.ui to ease migrations

laeubi:
Currently it is very hard to migrate parts of the code, there is only
all or nothing and migration feels stuck for years now.

One culprit is the exhaustive use of PlatformUI.getWorkBench() and the
inability for code to easily "join" the E4 ecosystem without major
rewrites.

This change proposes a new API in PlatformUI that allows a migration
path from old Workbench API to E4 without creating an immediate
requirement to refactor all components. As an example
ObjectContributorManager is refactored to use the new API so it could
access the IExtensionTracker in both worlds.

### Disclaimer

Yes **it adds new API,** please still consider this so we can finally get some more spin on E4 migration :-)

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
@@ -14,8 +14,12 @@
  *******************************************************************************/
 package org.eclipse.ui;
 
+import java.util.Optional;
+import java.util.function.Supplier;
 import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.eclipse.e4.core.contexts.IEclipseContext;
 import org.eclipse.e4.ui.internal.workbench.swt.DialogSettingsProviderService;
+import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.jface.dialogs.IDialogSettingsProvider;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.swt.widgets.Display;
@@ -27,7 +31,9 @@ import org.eclipse.ui.internal.util.PrefUtil;
 import org.eclipse.ui.preferences.ScopedPreferenceStore;
 import org.eclipse.ui.testing.TestableObject;
 import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
 import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.ServiceReference;
 
 /**
  * The central class for access to the Eclipse Platform User Interface. This
@@ -83,6 +89,15 @@ public final class PlatformUI {
 	 * @since 3.0
 	 */
 	public static final int RETURN_EMERGENCY_CLOSE = 3;
+
+	/**
+	 * A supplier that supplies an {@link IllegalStateException} that a Workbench is
+	 * not running.
+	 *
+	 * @since 3.126
+	 */
+	public static final Supplier<RuntimeException> NO_WORKBENCH = () -> new IllegalStateException(
+			WorkbenchMessages.PlatformUI_NoWorkbench);
 
 	/**
 	 * Block instantiation.
@@ -232,5 +247,63 @@ public final class PlatformUI {
 	 */
 	public static IPreferenceStore createPreferenceStore(Class<?> clazz) {
 		return new ScopedPreferenceStore(InstanceScope.INSTANCE, FrameworkUtil.getBundle(clazz).getSymbolicName());
+	}
+
+	/**
+	 * This methods allows code of eclipse.ui to gracefully migrate to E4 as it has
+	 * the following properties:
+	 *
+	 * <ol>
+	 * <li>It does not throw an exception but returns an empty optional if something
+	 * is not ready (yet or anymore) so code can choose how to handle this (e.g. use
+	 * {@link Optional#orElseThrow()} if required or {@link Optional#orElse(Object)}
+	 * if a fallback exits.</li>
+	 * <li>If this is is actually a (ui) Workbench running, this is returned and
+	 * could be used</li>
+	 * <li>If not the service registry is searched for an
+	 * org.eclipse.e4.ui.workbench.IWorkbench so code can run inside a pure E4 (e.g.
+	 * RCP) application as well.</li>
+	 * <li>With the Application at hand, it is possible to get access to the
+	 * {@link IEclipseContext} (see {@link MApplication#getContext()} to further
+	 * interact with the plain E4 Application Model.</li>
+	 * </ol>
+	 *
+	 * The following example fails if run inside an E4 Application: <code>
+	 *  PlatformUI.getWorkbench().getExtensionTracker()
+	 *  </code> and could be rewritten as: <code>
+	 *  PlatformUI.getApplication().map(MApplication::getContext).map(ctx->ctx.get(IExtensionTracker)).orElseThrow(PlatformUI.NO_WORKBENCH);
+	 *  </code>
+	 *
+	 * @return a reference to the {@link MApplication} running the Workbench or an
+	 *         empty optional if currently no Workbench is available.
+	 * @since 3.126
+	 */
+	public static Optional<MApplication> getApplication() {
+
+		Workbench instance = Workbench.getInstance();
+		if (instance != null && instance.isRunning()) {
+			return Optional.of((instance.getApplication()));
+		}
+		WorkbenchPlugin plugin = WorkbenchPlugin.getDefault();
+		if (plugin == null) {
+			return Optional.empty();
+		}
+		// We intentionally do not use a tracker here to not bind it to much to the
+		// service in case E4 is used
+		BundleContext bundleContext = plugin.getBundleContext();
+		if (bundleContext == null) {
+			return Optional.empty();
+		}
+		ServiceReference<org.eclipse.e4.ui.workbench.IWorkbench> reference = bundleContext
+				.getServiceReference(org.eclipse.e4.ui.workbench.IWorkbench.class);
+		if (reference == null) {
+			return Optional.empty();
+		}
+		org.eclipse.e4.ui.workbench.IWorkbench service = bundleContext.getService(reference);
+		if (service == null) {
+			return Optional.empty();
+		}
+		bundleContext.ungetService(reference);
+		return Optional.ofNullable(service.getApplication());
 	}
 }
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/ObjectContributorManager.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/ObjectContributorManager.java
@@ -34,6 +34,7 @@ import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.dynamichelpers.ExtensionTracker;
 import org.eclipse.core.runtime.dynamichelpers.IExtensionChangeHandler;
 import org.eclipse.core.runtime.dynamichelpers.IExtensionTracker;
+import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.ui.PlatformUI;
 
@@ -112,8 +113,9 @@ public abstract class ObjectContributorManager implements IExtensionChangeHandle
 		if (extensionPointId != null) {
 			IExtensionPoint extensionPoint = Platform.getExtensionRegistry().getExtensionPoint(PlatformUI.PLUGIN_ID,
 					extensionPointId);
-			IExtensionTracker tracker = PlatformUI.getWorkbench().getExtensionTracker();
-			tracker.registerHandler(this, ExtensionTracker.createExtensionPointFilter(extensionPoint));
+			PlatformUI.getApplication().map(MApplication::getContext).map(ctx -> ctx.get(IExtensionTracker.class))
+					.orElseThrow(PlatformUI.NO_WORKBENCH)
+					.registerHandler(this, ExtensionTracker.createExtensionPointFilter(extensionPoint));
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
```diff
@@ -84,6 +90,15 @@ public final class PlatformUI {
 	 */
 	public static final int RETURN_EMERGENCY_CLOSE = 3;
 
+	/**
+	 * A supplier that supplies an {@link IllegalStateException} that a Workbench is
+	 * not running.
+	 *
+	 * @since 3.126
+	 */
+	public static final Supplier<RuntimeException> NO_WORKBENCH = () -> new IllegalStateException(
```
mickaelistria, line 99: Couldn't it go into ObjectContributorManager?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
```diff
@@ -233,4 +248,62 @@ public static IDialogSettingsProvider getDialogSettingsProvider(Bundle bundle) {
...
+	 *  PlatformUI.getWorkbench().getExtensionTracker()
+	 *  </code> and could be rewritten as: <code>
+	 *  PlatformUI.getApplication().map(MApplication::getContext).map(ctx->ctx.get(IExtensionTracker)).orElseThrow(PlatformUI.NO_WORKBENCH);
+	 *  </code>
+	 *
+	 * @return a reference to the {@link MApplication} running the Workbench or an
+	 *         empty optional if currently no Workbench is available.
+	 * @since 3.126
+	 */
+	public static Optional<MApplication> getApplication() {
```
mickaelistria, line 281: Wouldn't it be better to have it on IWorkbench?

mickaelistria (commented):
Would this really be the best way to get the MApplication? Shouldn't the MApplication be injectable somehow, or possibly retrieved as a service?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
laeubi: No, this should be usable on other places as well `ObjectContributorManager` is only an example

laeubi:
If there would be a way to inject it we won't need this method, but as there is many code not supporting injection yet, we need a way to gracefully migrate those ... so in the End we hopefully can throw away `PlatformUI` completely.

mickaelistria:
> so in the End we hopefully can throw away PlatformUI completely.

I'm afraid adding 1 new method that many consumers will be likely to use goes to the opposite direction of throwing away Platform UI... Couldn't we access the MApplication without injection? Can you please provide some examples of code you're willing to replace?

laeubi:
> Couldn't we access the MApplication without injection?

I just added a method to do so...

> Can you please provide some examples of code

Code is included in this change.

mickaelistria:
> I just added a method to do so...

Sorry, what I had in mind, but didn't made clear was, is it possible to access MApplication without using injection and without using PlatformUI ? Something that would only rely on E4 APIs (that we are not willing to throw away, yet).

laeubi:
The Application is bootstrapped by the Workbench, so if you are not have an Addon or Part that is already declared at the e4 you can't access it directly, that's why this method is now added to have an entry point for this that could be used to enter "E4 World" from the `eclipse.ui`parts.

mickaelistria:
From a part, on can acces the IWorkbenchSite which implements IServiceLocator. So something like `site.getService(EPartService.class).findPart(getId())` should work, shouldn't it?

laeubi:
We have Part, Wizards, Dialogs and Action(handlers) so I'm not sure if this helps here. A part can also be generally migrated quite easy because there is an E3Part-Wrapper as far as I remember, so if there are better ways than using this new method we should use them, still there are many that are not obvious and deeply nested, just search for `org.eclipse.ui.PlatformUI.getWorkbench()` there are thousands of places. 

Beside that, if I have `IWorkbenchSite`/`IServiceLocator` i already require a running (eclipse.ui) Workbench ... so nothing is gained here.

laeubi:
Just in case it is not known, but Eclipse is currently running an "E4LegacyApplication" that is running an E3 Workbench (`eclipse.u ` stuff) and E4 Workspace model with compatibility layer. Then you can use `PlatformUI#getWorkbench()` and you can use E4, either directly or indirectly through the compatibility layer so there is no real "force" in migration because everything works.

But now if you make a real (RCP) E4 application, you can't use any class that calls `PlatformUI#getWorkbench()` either direct or indirectly e.g. the P2-UI uses some dialogs that call `PlatformUI#getWorkbench().getHelpSystem()` and as soon as such code is hit everything fails with an exception. So if one is able to fix those, we can stepwhise migrate at least P2 (still much effort), wile otherwise we have to rewrite (or copy) every single class and are soon at the "this is to complex/incompatible/will break/ ..." stage so nothing will change and we "wait" for ever that something magically changes...

opcoach:
To get any object in the application context you can use : 

(MApplication) PlatformUI.getWorkbench().getService(MApplication.class);

or for the current context : 

(IEclipseContext) PlatformUI.getWorkbench().getService(IEclipseContext.class);

laeubi:
> (MApplication) PlatformUI.getWorkbench().getService(MApplication.class);

Will throw `IllegalStateException(WorkbenchMessages.PlatformUI_NoWorkbench)` in a real E4 Application

> (IEclipseContext) PlatformUI.getWorkbench().getService(IEclipseContext.class);

Will throw `IllegalStateException(WorkbenchMessages.PlatformUI_NoWorkbench)` in a real E4 Application

mickaelistria:
I think it's up to e4 to provide a sustainable API to access services, that would work well in a workbench and a plain e4 application. Adding more responsibility to PlatformUI is making it harder and harder to replace.

laeubi:
> I think it's up to e4 to provide a sustainable API to access services, that would work well in a workbench and a plain e4 application. Adding more responsibility to PlatformUI is making it harder and harder to replace.

Please come up with a running example then, it sounds strange to request something "from E4" to overcome limitations of E3 E4 knows nothing about E3 so why should it care? Such a static access would contradict how E4 works and is designed. E4 already offers everything required here and the code shows that.

> API to access services

That is `@Inject` and alike, but for that you first need to get "into" E4...

laeubi:
Maybe @wimjongman @vogella or anyone with E4 background can leave a comment here...

wimjongman:
I'm on vacation but I like the idea and  will take a closer look once I get back.

Best regards,

Wim Jongman
________________________________

laeubi:
@vogella I asked EF Infra to enable the more convienet button for rebase a branch see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1610

vogella:
@laeubi as you create new API you need to increase the minor version of the plug-in. I like this change.

merks:
What's the advantage of this new API versus this access pattern used elsewhere?

image

One typically/often needs other services as well:

image

laeubi:
> @laeubi as you create new API you need to increase the minor version of the plug-in. I like this change.

Done.

> What's the advantage of this new API versus this access pattern used elsewhere?

That "elsewhere" it is assumed you have access to a (e3) Workbench that simply will fail in an E4 application. So if you can rewrite any of these access without the need of having an `IWorkbench` or calling `PlatforUI.getWorkbench()` either explicitly or implicitly it won't be required. But as I not found any such way this adds a convenient API to access the e4 world.

vogella:
> That "elsewhere" it is assumed you have access to a (e3) Workbench that simply will fail in an E4 application. So if you can rewrite any of these access without the need of having an `IWorkbench` or calling `PlatforUI.getWorkbench()` either explicitly or implicitly it won't be required. But as I not found any such way this adds a convenient API to access the e4 world.

I fail to understand this comment. If you want a easy way to access the MApplication in a e4 and e3 world I would move such an API to one of e4 bundles. If you put it into org.eclipse.ui this will also not be available in pure e4 applications.

laeubi:
> I fail to understand this comment. If you want a easy way to access the MApplication in a e4 and e3 world I would move such an API to one of e4 bundles.

We are waiting for this > 12 years now, because it is simply impossible without either copy or refactor large parts of eclipse, so it is neither "simple" nor "easy", just search for `PlatformUI.getPlatfom()` all those codes need to be migrated to "pure" E4 and even I'm not aware of **any** platform project that actually **has** moved to E4, because you effectively loose all your "backup" and functionality what today makes "the Eclipse IDE".

That's why I like to propose this as a way to migrate **parts** (e.g. Wizards, Dialogs, Actions, ...) to a e3 independent access so it could be used in both worlds (see https://github.com/eclipse-equinox/p2/discussions/83) so finally at laest people can migrate code to e4 API and finally then one might be able to migrate to e4 completely at least for some parts (P2 UI for example).

merks:
@laeubi  I see what you mean, but I'm still confused by all this and get your point.  Still though, it seems in most cases even if you have the MApplication, you often still need the EModelService and/or EPartService.   So how to get those?  And as Lars suggests, putting this API in a plugin for the e3 workbench in a utility class primarily focused on org.eclipse.ui.PlatformUI.getWorkbench() seems a strange place for it adding e4 utilities...

laeubi:
> you often still need the EModelService and/or EPartService. So how to get those?

The `MApplication` has a method `MApplication#getContext` wich gives you the applications `IEclipseContext` with that at hand you can get **any** service, or even better, simply inject all you want int your View/Dialog/Wizard/.... using `ContextInjectionFactory`

> And as Lars suggests, putting this API in a plugin for the e3 workbench in a utility class primarily focused on org.eclipse.ui.PlatformUI.getWorkbench() seems a strange place for it adding e4 utilities...

As noted earlier, this is **not** an e4 utility, because in E4 I have no need for this as I can get injected what ever I want/need (including the `MApplication`), this is a method for e3 code to "enter" e4 world, as soon as you are "in" you not need it anymore, so one can more and more move component into and then remove the need of calling that method.

vogella:
> > I fail to understand this comment. If you want a easy way to access the MApplication in a e4 and e3 world I would move such an API to one of e4 bundles.
> 
> We are waiting for this > 12 years now, because it is simply impossible without either copy or refactor large parts of eclipse, so it is neither "simple" nor "easy", just search for `PlatformUI.getPlatfom()` all those codes need to be migrated to "pure" E4 and even I'm not aware of **any** platform project that actually **has** moved to E4, because you effectively loose all your "backup" and functionality what today makes "the Eclipse IDE".
> 
> That's why I like to propose this as a way to migrate **parts** (e.g. Wizards, Dialogs, Actions, ...) to a e3 independent access so it could be used in both worlds (see [eclipse-equinox/p2#83](https://github.com/eclipse-equinox/p2/discussions/83)) so finally at laest people can migrate code to e4 API and finally then one might be able to migrate to e4 completely at least for some parts (P2 UI for example).

If you put the API into PlatformUI you still depend on e3 (similar to what Ed said). I now searched for a good place to put it and it looks to me what we already have org.eclipse.e4.ui.workbench.IWorkbench#getApplication So maybe we only need a way to access org.eclipse.e4.ui.workbench.IWorkbench easier?

merks:
Note that that I think part of point here is that there isn't always a workbench.   E.g., the installer application has no workbench and that a number of nasty problems because other parts of the UI assume there is a workbench, so I'm not sure if there is always an e4 IWorkbench or not...

vogella:
> If you put the API into PlatformUI you still depend on e3 (similar to what Ed said). I now searched for a good place to put it and it looks to me what we already have org.eclipse.e4.ui.workbench.IWorkbench#getApplication So maybe we only need a way to access org.eclipse.e4.ui.workbench.IWorkbench easier?

Workbench already implements org.eclipse.e4.ui.workbench.IWorkbench so we could use PlatformUI.getWorkbench().getApplication() in e3 code to access the application. Would that work for you @laeubi?

laeubi:
> PlatformUI.getWorkbench().getApplication()

Please see above, this will instantly throw "Workbench is not started yet" in any E4 RCP application.

vogella:
> > PlatformUI.getWorkbench().getApplication()
> 
> Please see above, this will instantly throw "Workbench is not started yet" in any E4 RCP application.

Didn't you way that you want to use this API in e3 code? IIRC MApplication is also published as OSGi service, so in case you want API usable for e3 and e4 could you use the OSGi API?

laeubi:
> Note that that I think part of point here is that there isn't always a workbench. E.g., the installer application has no workbench and that a number of nasty problems because other parts of the UI assume there is a workbench, so I'm not sure if there is always an e4 IWorkbench or not...

@merks I think you finally understand the pain I feel and if you now take a closer look at the proposed API:

> `public static Optional<MApplication> getApplication() `

it returns an `Optional` and thus allows code using this method to perform an action **only** if there **is** a workbench and otherwise can **skip** that action (e.g setting the help)

laeubi:
> Didn't you way that you want to use this API in e3 code? IIRC MApplication is also published as OSGi service, so in case you want API usable for e3 and e4 could you use the OSGi API?

The proposed method exactly do this, but in a way that:

1. Check if there is an e3 workbench running and instantly return this one then without any additional code
2. Check if there is an e4 Application available in the OSGI service registry and then returning this
3. If not, return an empty optional to leave the choice to the caller if he either can proceed without (optional operation) or throw an exception (strictly required) or do anything other (required but not optional)

Encoding this on all the thousands of places where today `PlatformUI.getWorkBench()` is used seems not realistic and useful, because of this I propose a easy to use, bullet proof (at least we can fix issues on **one** place) replacement with modern java programming patterns here.

vogella (approved):
Adding one new API method is not terrible bad and if this helps adapting the e4 programming model in more code, I say "lets go". 

@laeubi from my side, please go ahead and merge

iloveeclipse:
This caused releng test failure in I20220802-1800, see 
- https://download.eclipse.org/eclipse/downloads/drops4/I20220802-1800/testresults/html/org.eclipse.releng.tests_ep425I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html
- https://download.eclipse.org/eclipse/downloads/drops4/I20220802-1800/compilelogs/platform.doc.isv.javadoc.txt

```
../../../eclipse.platform.ui/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java:274: error: bad use of '>'
	 *  PlatformUI.getApplication().map(MApplication::getContext).map(ctx->ctx.get(IExtensionTracker)).orElseThrow(PlatformUI.NO_WORKBENCH);
	                                                                      ^
1 error
```

I've pushed #239  with the fix

PR #212, opened 2022-07-15
Match API errors

laeubi:
Test how to match API tools errors in GH Action

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
         java-version: '17'
         distribution: 'temurin'
         cache: maven
-    - run: echo "::add-matcher::matcher.json"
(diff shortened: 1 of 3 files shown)


laeubi:
At least if one goes to the summary page of the job: https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/2677121896 it shows the API errors:

grafik

lets see if I can tweak the Regexp to get a code annotation....

laeubi:
I think the problem is that API tools report:  `/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IActionFilter.java` as a filename but the path is actually `/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IActionFilter.java`

PR #216, opened 2022-07-16
Use more Tycho pomless

HannesWell:
Delete now obsolete trivial pom.xml files and fix some warnings to make them obsolete.

For some of the deleted pom.xml files in test plug-ins probably the packing type has to be explicitly because they don't comply to the usual test project naming schema.


HannesWell:
Great.
The build succeeds with less warnings than before and no new ones. 
The number of modules build is increased by one from 88 to 89, because `org.eclipse.ui.tests.browser` is now build as well. But because the tests to not pass in CI at the moment (and are long running) their execution is disabled.
Compared to the current master exactly the same tests are executed.

So this is good to go.

PR #217, opened 2022-07-17
Build o.e.ui.examples.adapterservice and delete not build features

HannesWell:
The following projects are not build and are therefore not published. These are namely:
- org.eclipse.ui.examples.adapterservice
- org.eclipse.e4.ui.progress.feature
- org.eclipse.ui.tests.feature

Can we simply remove them? Or should they be re-activated and added to the build?
For the tests feature it seems reasonably for me the remove it.
When we remove the `org.eclipse.e4.ui.progress.feature` we could consider to remove the plug-in `org.eclipse.e4.ui.progress` too. It is currently build and also part of the `https://download.eclipse.org/eclipse/updates/` repo, but it is not installed in my Eclipse. Or is this only used by non Eclipse-TLP projects?

Furthermore I noticed that the following example projects are not in the `https://download.eclipse.org/eclipse/updates/` repo:
- org.eclipse.e4.demo.cssbridge
- org.eclipse.e4.ui.examples.job
- org.eclipse.jface.snippets
- org.eclipse.ui.examples.job
- org.eclipse.ui.examples.navigator
- org.eclipse.ui.forms.examples
- org.eclipse.ui.examples.markerHelp

I don't know how the other example projects make it into the repo and how they are used. But we should either fully use the mentioned projects (maybe they are used by other means?) or if they are not used at all, remove them.

@vogella, @iloveeclipse can you say anything about that?


iloveeclipse:
> Furthermore I noticed that the following example projects are not in the `https://download.eclipse.org/eclipse/updates/` repo:

Why should they? It's examples, nothing to be installed.

HannesWell:
> > Furthermore I noticed that the following example projects are not in the `https://download.eclipse.org/eclipse/updates/` repo:
> 
> Why should they? It's examples, nothing to be installed.

Sure. But where can I then check how they are presented to the user? Or do they only exists in the git so somebody interested can look into the sources?

iloveeclipse:
> they only exists in the git so somebody interested can look into the sources?

Exact.

HannesWell:
> > they only exists in the git so somebody interested can look into the sources?
> 
> Exact.

OK. Then the example project should be build so that it at least ensured it is operational.

For the features I still think it is ok to delete them, because there is nothing you can learn from them if they are not used anywhere.

HannesWell:
Since you did not objected I assume that you silently are fine with this change. :)
From other discussions I know that we want to remove unused elements that don't add any value and I think those simple features are such things.

PR #219, opened 2022-07-19
Make UIEvents interfaces @noimplements

mickaelistria:
Those interfaces are actually namespaces and not meant to be
implemented.
The @noimplements will allow to more easily add constants.


vogella:
IIRC this interface is generated, we have also an application for this cc @merks is this correct?

mickaelistria:
> IIRC this interface is generated, we have also an application for this cc @merks is this correct?

The question about generators is whether they're meant to bootstrap the code (generate once) and then we can update the generated code; or whether they're meant to regenerate the code often so only the generator config should be edited.
Here is seems to me that the generator has run once and code is now expected to be manually maintained,

merks:
This class is not generated:

org.eclipse.e4.ui.workbench.UIEvents

There would be `@generated` markers if it were generated.

For a generated class, you just need add it to the model doc as was done here:

image

Which generates this:

image

Or you can add it inside the `<!-- begin-user-doc -->` section; regenerating will preserve that section.

And no, the EMF generator is definitely **not** a one of thing for bootstrapping.  You'd better ensure that the model can evolve in the future and can be regenerated.  When in doubt, ask as @vogella  just did.  This model has a long history:

image

Issue #198, opened 2022-07-08
 Add Preference to Turn Off Warning-Check for Project Specific Encoding


ingomohr:
This is the UI contribution for backend issue [#166](https://github.com/eclipse-platform/eclipse.platform.resources/issues/166).

ingomohr:
As discussed in the backend issue I created [a fork for this repo](https://github.com/ingomohr/eclipse.platform.ui) to add the new preference combo to the "Workspace" preference page.

💡 Pull Request is #220

ingomohr:
@iloveeclipse Since the PR #220 had beem merged (thanks for merging btw): Is there anything left to do for this issue? Or shall I close it?

ingomohr:
i.e. what about the New & Noteworthy entry? I remember I was asked to add that for my last UI contribution. However, I don't know the repo for that one any more...

iloveeclipse:
> i.e. what about the New & Noteworthy entry? 

Yes, please.

> I remember I was asked to add that for my last UI contribution. However, I don't know the repo for that one any more...

This is now https://github.com/eclipse-platform/www.eclipse.org-eclipse-news

ingomohr:
Added [PR to N&N repo](https://github.com/eclipse-platform/www.eclipse.org-eclipse-news/pull/48)

ingomohr:
@iloveeclipse (just in case a mention is better than no mention; no pressure though). The PR is right there waiting. ;)

vogella:
Thanks @ingomohr I think this issue is done. Please reopen if there is still something left

ingomohr:
@vogella Hi Lars. Thank you for merging the N&N PR! I agree this can be closed now.

PR #220, opened 2022-07-19
Add preference combo to change severity of 'missing project encoding'

ingomohr:
This introduces a new combo-box entry to the "Workspace" preference page
to modify the severity used for the check for "missing project-specific
encoding".
See issue #198.

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/IDEWorkbenchMessages.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/IDEWorkbenchMessages.java
@@ -26,6 +26,7 @@
  * Lucas Bullen (Red Hat Inc.) - Bug 522096
  * Gunnar Wagenknecht - [102527] Project Natures property page
  * Mickael Istria (Red Hat Inc) - [102527] Project Natures property page
+ * Ingo Mohr <tellastory73@gmail.com> - Issue #198
  *******************************************************************************/
 package org.eclipse.ui.internal.ide;
 
@@ -573,6 +574,8 @@ public class IDEWorkbenchMessages extends NLS {
 	public static String IDEWorkbenchPreference_workbenchSystemExplorer;
 	public static String IDEWorkspacePreference_UnknownNatureSeverity;
 	public static String IDEWorkspacePreference_UnknownNatureSeverity_Ignore;
+	public static String IDEWorkspacePreference_UnknownEncodingSeverity;
+	public static String IDEWorkspacePreference_UnknownEncodingSeverity_Ignore;
 
 	// --- Linked Resources ---
 	public static String LinkedResourcesPreference_explanation;
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/IDEWorkspacePreferencePage.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/IDEWorkspacePreferencePage.java
@@ -17,12 +17,14 @@
 *     Christian Georgi (SAP SE)          -  bug 458811
 *     Mickael Istria (Red Hat Inc.) - Bug 486901
 *     Patrik Suzzi <psuzzi@gmail.com> - Bug 502050
+*     Ingo Mohr <tellastory73@gmail.com> - Issue #198
 *******************************************************************************/
 package org.eclipse.ui.internal.ide.dialogs;
 
 import java.util.List;
 import java.util.stream.Stream;
 
+import org.eclipse.core.internal.resources.ValidateProjectEncoding;
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IWorkspaceDescription;
 import org.eclipse.core.resources.ResourcesPlugin;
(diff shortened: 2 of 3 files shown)


ingomohr:
Pls note I created this PR to make integration testing of related [Backend PR 167](https://github.com/eclipse-platform/eclipse.platform.resources/pull/167) easier.

This PR is blocked by that PR in the platform repo.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/IDEWorkbenchMessages.java
```diff
@@ -9,23 +9,24 @@
  * SPDX-License-Identifier: EPL-2.0
```
trancexpress: Would be nice to not do unnecessary formatting changes, at least not in combination with actual code changes that matter.

If this is done by the formatter, you can "undo" it with Ctrl+Z (or whichever undo hotkey you use) after a save.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/IDEWorkspacePreferencePage.java
```diff
@@ -197,6 +200,25 @@ private void createMissingNaturePref(Composite parent) {
 		missingNatureSeverityCombo.load();
 	}
 
+	/**
+	 * @param comboParent
```
trancexpress: Please remove comment.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/messages.properties
```diff
@@ -550,6 +550,8 @@ IDEWorkspacePreference_workspaceName=Wor&kspace name:
 IDEWorkbenchPreference_workbenchSystemExplorer=Command for launching system e&xplorer:
 IDEWorkspacePreference_UnknownNatureSeverity=Report unknown project n&ature as:
 IDEWorkspacePreference_UnknownNatureSeverity_Ignore=Ignore
+IDEWorkspacePreference_UnknownEncodingSeverity=Report missing project &encoding as:
```
trancexpress: 1. "Show workspace nam&e"
2. "New text line delimiter" -> "D&efault"

Options uses &e in the same dialog, I'm not sure how this collision is meant to be resolved (Alt+E results in the "Show workspace name" checkbox being interacted with). @iloveeclipse ? If there is already 1 collision that cannot be worked around, no point in adding another one.

iloveeclipse:
> /rebase

Ignore that, I've forgot the other repo change will be visible on next day only. 
But in general, it is nicer (less work for reviewers) if PR's are rebased on latest master.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/IDEWorkbenchMessages.java
ingomohr: Thank you for the feedback. I wasn't sure about this change and whether everything on a PR is still supposed to go into one single commit. I considered the formatting of the contributors list broken and aimed to fix it so that it would match the other classes I'd seen.

I reverted that change now, so that the only change in the list is the new entry in the list.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/IDEWorkspacePreferencePage.java
ingomohr: removed that jdoc with (even wrong) parameter name. Thanks for finding.

(I had added this for consistency with the other methods - but I agree that the comment doesn't add any value. Happy to drop it)

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/messages.properties
ingomohr: True.

I admittedly didn't test the mnemonic at all (apparently, there is [no mnemonic support in SWT for macOS](https://bugs.eclipse.org/bugs/show_bug.cgi?id=399554)).

Do we want to choose some other char? If so: which one is best? Or is it feasible to not specify one at all?

ingomohr: What about the c?

ingomohr:
> > /rebase
> 
> Ignore that, I've forgot the other repo change will be visible on next day only. But in general, it is nicer (less work for reviewers) if PR's are rebased on latest master.

Thanks. I didn't know about the "next day" thing. I still rebased my PR (before I saw your feedback here b/c my change was already 14 days old on my local branch).

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/messages.properties
```diff
@@ -550,6 +551,8 @@ IDEWorkspacePreference_workspaceName=Wor&kspace name:
 IDEWorkbenchPreference_workbenchSystemExplorer=Command for launching system e&xplorer:
 IDEWorkspacePreference_UnknownNatureSeverity=Report unknown project n&ature as:
 IDEWorkspacePreference_UnknownNatureSeverity_Ignore=Ignore
+IDEWorkspacePreference_UnknownEncodingSeverity=Report missing project en&coding as:
```
iloveeclipse:
```suggestion
IDEWorkspacePreference_UnknownEncodingSeverity=Rep&ort missing project encoding as:
```

PR #221, opened 2022-07-19
Decoration is not a question

kjsmita6:
This decoration when changing a theme is prompted as a question, even though it isn't. It is a warning to tell the user to restart for the theme change to take full effect. They are prompted with a question when pressing "Apply".


vogella:
Thanks @kjsmita6

PR #222, opened 2022-07-20
Add support for multi-project selection

### LINKED ISSUE #102: Allow to run Source actions on a set of projects from Project Explorer even if one of them is not a Java project
Follow-up for https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/34

#34 allows to run source actions from the project explorer which is great. One missing feature is that this is also possible if the set of project contains one or more non Java projects but at least one Java project. This is currently suported by the Package explorer.

To test: 

Create three projects, two Java projects, one of type General.
Open Package Explorer, select all projects, right-click -> You see the Source options
Open Project Explorer, select all projects, right-click -> You  DO NOT see the Source options

@jjohnstn can you also have a look at this one?

jjohnstn:
- add ability to evaluate IStructuredSelection to CommonActionProviderDescriptor.isEnabledFor()
- this is needed to fix: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/102

### DIFF
--- a/bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/CommonActionProviderDescriptor.java
+++ b/bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/CommonActionProviderDescriptor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2015 IBM Corporation and others.
+ * Copyright (c) 2003, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -231,8 +231,16 @@ public class CommonActionProviderDescriptor implements
 			}
 		} else {
 			IEvaluationContext parentContext = NavigatorPlugin.getApplicationContext();
+			boolean isEnabled = true;
 			for (Object element : aStructuredSelection) {
 				IEvaluationContext context = new EvaluationContext(parentContext, element);
+				context.setAllowPluginActivation(true);
+				if (NavigatorPlugin.safeEvaluate(enablement, context) != EvaluationResult.TRUE) {
+					isEnabled = false;
+				}
+			}
+			if (!isEnabled) {
+				IEvaluationContext context = new EvaluationContext(parentContext, aStructuredSelection);
 				context.setAllowPluginActivation(true);
 				if (NavigatorPlugin.safeEvaluate(enablement, context) != EvaluationResult.TRUE) {
 					return false;


bundles/org.eclipse.ui.navigator/META-INF/MANIFEST.MF
```diff
@@ -18,7 +18,8 @@ Export-Package: org.eclipse.ui.internal.navigator;x-friends:="org.eclipse.ui.nav
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.18.0,4.0.0)",
  org.eclipse.ui;bundle-version="[3.6.0,4.0.0)",
  org.eclipse.core.expressions;bundle-version="[3.4.0,4.0.0)",
- org.eclipse.jface;bundle-version="[3.18.0,4.0.0)"
+ org.eclipse.jface;bundle-version="[3.18.0,4.0.0)",
+ org.eclipse.core.resources;bundle-version="3.18.0"
```
mickaelistria: This is not desired. The navigator may be used in contexts without resources.

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
```diff
@@ -385,6 +385,13 @@ If the required action provider is disabled or unbound to the current viewer, th
                </appInfo>
             </annotation>
          </attribute>
+         <attribute name="muiltiprojectEnable" type="string">
```
mickaelistria: The actionProvider already has an `enablement` child to configure the enablement. Can't multi-project be implemented with this?

mickaelistria (changes requested):

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
jjohnstn: No, because the situation is that one or more of the selected elements has failed the enablement conditions, but at least one has passed.  The current code bails when any of the objects fails enablement conditions.  The enablement failure is legitimate for the non-Java projects selected but if there is one Java project, we can show certain menu items. We only want to do this when Projects are seletcted which is why I added the resources dependency to get IProject.  There might be another way to determine this with reflection.

mickaelistria: OK, I think I get it. So the root issue is that the enablement iterates on all items instead of applying to the whole selection? If it were possible to process the whole selection with a core expression, would that work?
It would be more generic and powerful without introducing a new undesired dependency.

bundles/org.eclipse.ui.navigator/META-INF/MANIFEST.MF
jjohnstn: Ok, latest patch removes this requirement.

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
jjohnstn: Yes, it iterates on each item in the selection.  Not sure what you mean by a core expression.  Most menus, etc.. won't be able to handle if all objects in the selection aren't enabled.  This may be a one-off case.  The Package View supports this for selecting Java projects and non-Java projects and Lars wanted the same behavior for the Project View.  I thought an optional attribute was the simplest way to do this.

vogella (approved):
LGTM, thanks @jjohnstn

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
merks: Note that `muiltiprojectEnable` has a typo and I think it would be better as more camel case  `multiProjectEnable` because there is no such thing as a `multiproject`

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
```diff
@@ -385,6 +385,13 @@ If the required action provider is disabled or unbound to the current viewer, th
...
+            <annotation>
+               <documentation>
+                  Used to specify how action handles multi-project selection and action enablement.  The value &quot;any&quot; means if any project is enabled, then action provider is enabled.
```
merks: The sentences aren't really grammatically complete.  Perhaps more like this:

Used to specify how the action handles multi-project selection for action enablement.  The value &quot;any&quot; means if any project is enabled, then the action provider is enabled.

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/CommonActionProviderDescriptor.java
```diff
@@ -231,17 +234,44 @@ public boolean isEnabledFor(IStructuredSelection aStructuredSelection) {
...
 				IEvaluationContext context = new EvaluationContext(parentContext, element);
 				context.setAllowPluginActivation(true);
 				if (NavigatorPlugin.safeEvaluate(enablement, context) != EvaluationResult.TRUE) {
-					return false;
+					boolean isProject = isProject(element.getClass());
+					if (!isProject || multiproject == null || !multiproject.equals("any")) { //$NON-NLS-1$
+						return false;
+					}
+				} else {
+					retValue = true;
```
laeubi: as retvalue could never change here, why not return immediately?

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
jjohnstn: Will change to camel case.

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
jjohnstn: Ok, will change

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/CommonActionProviderDescriptor.java
jjohnstn: Because this else statement is also hit in the normal case where there is no multi-project enable "any".  There may be other elements in the selection to check in the loop which could return false.

jjohnstn:
Any other issues to address?

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
jjohnstn: Done

bundles/org.eclipse.ui.navigator/schema/navigatorContent.exsd
jjohnstn: Done

vogella:
@mickaelistria you are blocking with your review. Do you still require changes?

mickaelistria:
I still don't think this patch is suitable. The resource/project logic should not leak into the project explorer. We may instead add the selection to the evaluation context to let consumers implement their own logic according to the selection, and not only according to each selected element.

merks:
Just a thought...  Forgive me if I missed the point or if it's a misguided thought...

I don't really understand the specific need for testing isProject.  Isn't this more of a generic issue about whether to use "and" versus "or" for combining the enablement for selected items?

mickaelistria:
@merks I think https://github.com/eclipse-platform/eclipse.platform.ui/pull/222#discussion_r926808383 may answer your question. The issue is the the CNF does iterate on the selection already assuming the condition is a `and` on every selected items. But this is too limiting. I believe that giving access to the selection so it can be iterated on via core expressions can remediate this without breaking anything nor mixing concerns.

jjohnstn:
@merks It would indeed work with "or" being specified.  I don't understand what @mickaelistria is referring to with fixing this with core expressions but am happy to fix this however or let someone else do it.  The case being fixed had to do with multiple project selection only for one particular JDT actionProvider that can handle it when at least one project is enabled so I didn't want to impact any other actionProviders or selection scenarios.

mickaelistria:
@jjohnstn What I mean is that if https://wiki.eclipse.org/Command_Core_Expressions is used (I hope it is), then one can change the current vairable with a `<with ...>` node. If the current variable cna be the whole selection, then you can create a condition that would fit your needs.

jjohnstn:
Ok, I'll look into it.

akurtakov:
@jjohnstn Would you please finish this one?

jjohnstn:
Hi @akurtakov I looked into the suggestion by @mickaelistria to use Core Expressions but it cannot be used in this case (the extension used does not support it).  I tried a few other ideas but they did not work.   If @mickaelistria wishes to revisit the patch with that in mind, I'll be happy to make any additional changes he suggests.

mickaelistria (changes requested):
The navigator is not at all supposed to care about the resource model. It's a generic navigator, not tied to the workspace. We shouldn't use reflection here.
At the moment, I don't have a clear suggestion of how to achieve what's desired, but I think we need something more general regarding multi-selection in the navigator.

mickaelistria:
Here is a proposal: currently, `CommonActionProviderDescriptor.isEnabledFor()` mandates a `and` association for all enablements. This is not correct to enforce it, the enablement allows users to decide that and the user decision should be honored. The fact that `and` is hardcoded is a form of bug.
You would like to be able to write something like
```xml
<iterate ifEmpty="false" operator="or">
   <adapt type="org.eclipse.jdt.core.IJavaProject" />
</iterate>
```

assuming the iterable variable is the selection. That would be good.
To achieve that, we can change `CommonActionProviderDescriptor.isEnabledFor()` so in case the current strategy returns `false`, then it alternatively tries evaluates the expression against the `aStructuredSelection`.

jjohnstn:
Hi @mickaelistria Thanks for the suggestion above.  With the minor change to isEnabledFor() and using an <iterate operator="or"> on the requirements for the JDT UI NavigatorContent extension, I was able to get a multi-project selection to work if at least one Java project was selected.  Please review the new change which replaces the old one.

mickaelistria (approved):

mickaelistria:
Thank you!

vogella:
Thanks a bunch. @jjohnstn can I test this in tomorrows I-Build for the clean-up actions or is their another change needed in JDT to use this?

jjohnstn:
> Thanks a bunch. @jjohnstn can I test this in tomorrows I-Build for the clean-up actions or is their another change needed in JDT to use this?

Another patch is needed in JDT UI which I will post shortly.

jjohnstn:
@vogella https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/502

PR #239, opened 2022-08-03
Fixed broken javadoc (#211)

iloveeclipse:
Commmit 8b56e255efa33f935d04dc8d5cc50b430a7d116d added incorrect javadoc

- fixed not escaped ">" character
- fixed not compilable example code (missing .class)
- fixed very long line & code examples rendering

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/211

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
@@ -268,18 +268,20 @@ public final class PlatformUI {
 	 * interact with the plain E4 Application Model.</li>
 	 * </ol>
 	 *
-	 * The following example fails if run inside an E4 Application: <code>
-	 *  PlatformUI.getWorkbench().getExtensionTracker()
-	 *  </code> and could be rewritten as: <code>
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
```diff
@@ -268,18 +268,24 @@ public static IPreferenceStore createPreferenceStore(Class<?> clazz) {
...
+	 * <p>
+	 * <code>
 	 *  PlatformUI.getWorkbench().getExtensionTracker()
-	 *  </code> and could be rewritten as: <code>
-	 *  PlatformUI.getApplication().map(MApplication::getContext).map(ctx->ctx.get(IExtensionTracker)).orElseThrow(PlatformUI.NO_WORKBENCH);
+	 * </code>
+	 * </p>
+	 * and could be rewritten as:
+	 * <p>
+	 * <code>
```
akurtakov: Using {@code should allow to not escape special characters and thus direct copy of snippet from sources.

laeubi:
Thanks! I still wondering why eclipse do not complains about this?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/PlatformUI.java
iloveeclipse: Yep, I will change that

PR #243, opened 2022-08-04
Make WizardDialog standard PrimaryModal not ApplicationModal 

### LINKED ISSUE #242: Why is WizardDialog so hard coded to SWT.APPLICATION_MODAL ?
if i look at the code i see this:

```
	private static int getShellModality(boolean modal) {
		return modal ? SWT.APPLICATION_MODAL : SWT.NONE;
	}
```

which is called in the constructor, also i see this:

```
	public WizardDialog setModal(boolean modal) {
		setShellStyle(getShellStyle() & ~SWT.APPLICATION_MODAL | getShellModality(modal));
		return this;
	}
```

and

```
	public boolean isModal() {
		return (getShellStyle() & SWT.APPLICATION_MODAL) == SWT.APPLICATION_MODAL;
	}
```

but i don't want our wizard dialogs to be application modal but SWT.PRIMARY_MODAL 
because they only should block the window that triggers them, there parent, not other windows. those can be completely different things.

But it is now so hard coded even if we just override it like overwriting setShellStyle then it kind of breaks isModal()

it would be very nice if there was a static? global var:

public static int DEFAULT_MODAL_STYLE = SWT.APPLICATION_MODAL;

and getShellModality would return that.
that by the start of the application we could do WizardDialog.DEFAULT_MODAL_STYLE  = SWT.PRIMARY_MODAL

and all of them would then work this way

i think isModal should be something like

```
	public boolean isModal() {
		return (getShellStyle() & SWT.APPLICATION_MODAL) == SWT.APPLICATION_MODAL ||
                            (getShellStyle() & SWT.PRIMARY_MODAL) == SWT.PRIMARY_MODAL;
	}
```

jcompagner:
fixes #242

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
@@ -319,8 +319,8 @@ public class WizardDialog extends TitleAreaDialog implements IWizardContainer2, 
 	 * Sets the shell style of the wizard dialog.
 	 * <p>
 	 * Examples:<br>
-	 * To use the default style without the SWT.APPLICATION_MODAL bit:<br>
-	 * <code>setShellStyle(getShellStyle() &amp; ~SWT.APPLICATION_MODAL)</code>
+	 * To use the default style without the SWT.PRIMARY_MODAL bit:<br>
+	 * <code>setShellStyle(getShellStyle() &amp; ~SWT.PRIMARY_MODAL)</code>
 	 * <p>
 	 * To use the default style without the SWT.RESIZE bit:<br>
 	 * <code>setShellStyle(getShellStyle() &amp; ~SWT.RESIZE)</code>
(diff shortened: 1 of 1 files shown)


vogella:
Thanks, please enhance the commit message with a detailed descriptioin instead of pointing only to the issue, we want as much documentation as possible in the commit messages.

vogella:
Thanks @jcompagner LGTM and works well.

Thanks for this change.

bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
```diff
@@ -368,7 +368,8 @@ public WizardDialog setModal(boolean modal) {
 	 * @see #setShellStyle(int)
 	 */
 	public boolean isModal() {
-		return (getShellStyle() & SWT.APPLICATION_MODAL) == SWT.APPLICATION_MODAL;
+		return (getShellStyle() & SWT.PRIMARY_MODAL) == SWT.PRIMARY_MODAL
+				|| (getShellStyle() & SWT.PRIMARY_MODAL) == SWT.APPLICATION_MODAL;
```
iloveeclipse, line 372: Copy/paste error here.

PR #244, opened 2022-08-04
Revert "Remove animations from progress toolbar control"

jukzi:
Reverts eclipse-platform/eclipse.platform.ui#141

The commit deleted the wrong - working(!) - progress in the system task bar,.


vogella:
I think the removed icon is useless, so -1 for re-adding it. Jörg, you also indicated that you never understood the purpose of the icon so why adding it again?

Removing functionality which nobody understands is IMHO a good thing.

jukzi:
I personally do not need the icon but there was complaints when it was removed.
What i do at least want is the working progress report in the system status bar - which was also removed.

vogella:
> What i do at least want is the working progress report in the system status bar - which was also removed.

Ok, I'm fine with the revert in this case. Thanks!

vogella (approved):

PR #245, opened 2022-08-05
Fix Progress Bar to show progress instead of indeterminate Animation.

jukzi:
The average progress of all non-system Jobs is shown every 100ms.
Note that the Text left to it shows progress of top job in the Progress
View only.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2016 IBM Corporation and others.
+ * Copyright (c) 2004, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -26,6 +26,7 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jface.action.IAction;
+import org.eclipse.jface.dialogs.ProgressIndicator;
 import org.eclipse.jface.util.Throttler;
 import org.eclipse.jface.util.Util;
 import org.eclipse.osgi.util.NLS;
@@ -43,7 +44,6 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Label;
-import org.eclipse.swt.widgets.ProgressBar;
 import org.eclipse.swt.widgets.ToolBar;
 import org.eclipse.swt.widgets.ToolItem;
 import org.eclipse.ui.IWorkbenchWindow;
@@ -60,7 +60,7 @@ import org.eclipse.ui.statushandlers.StatusManager;
  */
 public class ProgressAnimationItem extends AnimationItem implements FinishedJobs.KeptJobsListener {
 
-	ProgressBar bar;
+	ProgressIndicator bar;
 
 	MouseListener mouseListener;
 
@@ -206,6 +206,29 @@ public class ProgressAnimationItem extends AnimationItem implements FinishedJobs
 		if (toolbar == null || toolbar.isDisposed()) {
 			return;
 		}
+		if (bar != null && !bar.isDisposed()) {
+			JobInfo[] jobInfos = ProgressManager.getInstance().getJobInfos(false);
+			int percentSum = 0;
+			int percentCount = 0;
+			for (JobInfo jobInfo : jobInfos) {
+				if (jobInfo != null && !(jobInfo.isBlocked() || jobInfo.getJob().getState() == Job.WAITING)) {
+					int percentDone = jobInfo.getPercentDone();
+					if (percentDone >= 0) {
+						percentSum += percentDone;
+						percentCount++;
+					}
+				}
+			}
+			if (percentCount > 0) {
+				bar.beginTask(100);
+				bar.worked(percentSum / percentCount); // average
+				AnimationManager.getInstance().setAnimated(true); // reschedule
+			} else {
+				if (jobInfos.length > 0) {
+					bar.beginAnimatedTask();
+				}
+			}
+		}
 
 		JobTreeElement[] jobTreeElements = FinishedJobs.getInstance().getKeptElements();
 		// search from end (youngest)
@@ -296,7 +319,7 @@ public class ProgressAnimationItem extends AnimationItem implements FinishedJobs
 		}
 		top.setLayout(gl);
 
-		bar = new ProgressBar(top, flags | SWT.INDETERMINATE);
+		bar = new ProgressIndicator(top, flags);		
 		bar.setVisible(false);
 		bar.addMouseListener(mouseListener);
 
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/ProgressIndicator.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/ProgressIndicator.java
@@ -79,6 +79,10 @@ public class ProgressIndicator extends Composite {
 	 * Initialize the progress bar to be animated.
 	 */
 	public void beginAnimatedTask() {
+		if (animated && layout.topControl == indeterminateProgressBar) {
+			// do not restart animation if it was already started
+			return;
+		}
 		done();
 		layout.topControl = indeterminateProgressBar;
 		requestLayout();


jukzi:
image
see also https://github.com/eclipse-platform/eclipse.platform.ui/pull/141#issuecomment-1156175623 for previous behaviour

merks:
@jukzi Thanks for working on this!

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
```diff
@@ -206,6 +206,26 @@ private void refresh() {
...
 			return;
 		}
+		if (bar != null && !bar.isDisposed()) {
+			JobInfo[] jobInfos = ProgressManager.getInstance().getJobInfos(false);
+			int percentSum = 0;
+			int percentCount = 0;
+			for (JobInfo jobInfo : jobInfos) {
+				if (jobInfo != null && !(jobInfo.isBlocked() || jobInfo.getJob().getState() == Job.WAITING)) {
+					int percentDone = jobInfo.getPercentDone();
+					if (percentDone >= 0) {
```
laeubi, line 216: if there is currently no progress, should it not still account to the average?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
```diff
@@ -206,6 +206,26 @@ private void refresh() {
...
+						percentSum += percentDone;
+						percentCount++;
+					}
+				}
+			}
+			if (percentCount > 0) {
+				bar.setSelection(percentSum / percentCount); // average
```
laeubi: An alternative would be to set the maximum to` (100 * percentCount)` and the selection to `percentSum` then you won't loose precision here due to rounding.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
jukzi: percentDone returns -1 for processes that do not report progress. That can not be accounted, The progress is unknown.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
jukzi: agree. but i prefer less system calls. A precision of hundredth is anyways low but enough to get an idea.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
laeubi: maybe there should then two counters, one for the ones that report no progress and one for that report progress.

Later then you can set the progress bar to indeterminate state if there are no one reporting actual progress but there are jobs?

jukzi: I would like to but as far as i know it is not possible to change the Style after constructor. There is no ProgressBar.setStyle. :-(

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
laeubi: `org.eclipse.jface.dialogs.ProgressIndicator` uses two progress-bars and a `StackLayout` for this, maybe this is too much here but just in case you want to make it extra cool :-)
Maybe one can simply reuse `org.eclipse.jface.dialogs.ProgressIndicator` here?

jukzi: good idea. I dont know/use any usecase with indefinite progress so i would like to skip that theoretical usecase.

laeubi: > I dont know/use any usecase with indefinite progress so i would like to skip that theoretical usecase.

I think the idea is to give the user an indication that *something* is in progress but we don't know how long it takes (e.g because you can't get the progress or can't know the number of steps in advance).

jukzi: yeah, but when is that ever used? Also the indeterminate animation in a progressbar that is that small looks like a (wrong) normal progress bar - which is not helpful.

mickaelistria: > I dont know/use any usecase with indefinite progress so i would like to skip that theoretical usecase.

Beware, because you don't know X doesn't make that X is only theorical and can be ignored.
@laeubi 's interpretation is correct, and some examples of job that typically use that indeterminate progress patterns are jobs that do wait for external application or service to complete some work; or just job that do not implement a finer progress report. Both use-cases are valid.

laeubi: > yeah, but when is that ever used?

I think this is a common case there are many parts of eclipse that never report any progress at all.

> Also the indeterminate animation in a progressbar

This is platform specific, e.g. on Linux the progresbar "bounces" left/right, on windows it seems to "run through" like here: https://codestory.de/11317/swt-progressbar

>  that is that small looks like a (wrong) normal progress bar

If it is too small we should simply make it bigger.

jukzi: > I think this is a common case there are many parts of eclipse that never report any progress at all.

So please name ANY long running operation that does not report fake progress.

> If it is too small we should simply make it bigger.

It's not the topic of this PR to make that rare usecase any better.

mickaelistria: > So please name ANY long running operation that does not report fake progress.

The 2nd half of an install/update (which does track transfer) does report relatively accurate progress.
m2e dependency download (again about transfer) does report relatively accurate progress.
JDT Build does report a not-too-bad progress (on a per project increment I believe, which maps ok-ish many workspaces)
LSP4E support for server-side progress notification reports progress as accurately as the server does emit events.

laeubi: > So please name ANY long running operation that does not report fake progress.

You already mentioned above that "percentDone returns -1 for processes that do not report progress" so why is it in the API and you add special handling for it if it is never used? You also have "jobInfo.isBlocked() || jobInfo.getJob().getState() == Job.WAITING" as an exclusion so if there are only waiting jobs one might also report indeterminate progress instead of "no progress" = 0 ...

> It's not the topic of this PR to make that rare usecase any better.

Sure not, but **if** someone thinks the bar is too small, we can make it bigger instead of showing that nothing happens at all.

jukzi: sorry, i meant please tell my any  long running operation that does report **indeterminate** progress..

jukzi: > Sure not, but **if** someone thinks the bar is too small, we can make it bigger instead of showing that nothing happens at all.

How about to report fixed 50% instead as a compromise since the animation is not helpful anyway.

mickaelistria: > How about to report fixed 50% instead as a compromise since the animation is not helpful anyway.

The animation is helpful. It's emphasizes that something is happening.
On the contrary a fixed 50% would drive to false interpretation: nothing is happening (wrong) and/or half of the job is done (wrong). So it's not at all a good solution and we cannot ditch the animation if we don't have a better way to emphasize that "something is happening but we don't know for how long".

I'll try to keep an eye on progress report to find a job that doesn't report progress and will report when I find one.
In the meantime, this case is covered by API and mustn't feel "broken" so an animation seems required.

laeubi: > since the animation is not helpful anyway.

Why do you think  animation is not helpful? Progressbar even supports different states for error/warning/sleeping with different colors.

> How about to report fixed 50% instead 

see https://docs.microsoft.com/en-us/windows/win32/uxguide/progress-bars#general one should not report "fake progress" ...

jukzi: How was this "animation" any helpful? 
image
It looks like a progress and never ever anybody knew it was the indeterminate progress animation instead of a progress until i looked into the code. The difference is just not visible.

laeubi: As said above, on GTK the progressbar bounces left/right in indeterminate state:

grafik

jukzi: > this case is covered by API and mustn't feel "broken" so an animation seems required.

Nobody complained when @vogella tried to remove it completely (https://github.com/eclipse-platform/eclipse.platform.ui/pull/244) and now you claim its mandatory. 

> I'll try to keep an eye on progress report to find a job that doesn't report progress and will report when I find one.

thanks.

mickaelistria: > Nobody complained when @vogella tried to remove it completely (https://github.com/eclipse-platform/eclipse.platform.ui/pull/244) and now you claim its mandatory.

It happens that some issue are overlooked when other people are busy with other things. Silence doesn't mean endorsement.

jukzi: > As said above, on GTK the progressbar bounces left/right in indeterminate state:

sweet. that looks much better then on WIndows.

laeubi: > Nobody complained when @vogella tried to remove it completely (#244) and now you claim its mandatory.

This is not really true I have left a comment, but I'm really sick of those "I don't see a difference/think it matters/whatever" discussions as its completely useless, I'm always try to give references and hints but I can't afford a discussion based on personal feelings and @vogella is Project Lead so who really likes to mess with the lead? So the only "solution" is to let it go in an wait unless others (you in this case) complain ...

laeubi: > sweet. that looks much better then on WIndows.

I have no clue if we can get better results on windows as well but according to microsoft an indeterminate progress bar should looks like this:


https://docs.microsoft.com/en-us/windows/win32/uxguide/progress-bars#indeterminate-progress-bars

mickaelistria: > who really likes to mess with the lead?

Messing with whoever (contributor, committer, project lead...) for good technical reasons is more than welcome. Project leads can make mistake and do welcome discussions, assistance, reviews... just like they expect every contributor to do. So don't hesitate to mess with them!

vogella: https://github.com/eclipse-platform/eclipse.platform.ui/pull/141 has been reverted via https://github.com/eclipse-platform/eclipse.platform.ui/pull/244 . Thanks @jukzi

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/progress/ProgressAnimationItem.java
jukzi: using ProgressIndicator to switch between indeterminate. still indeterminate animation looks very unusefull in Windows.

PR #258, opened 2022-08-09
Cache that no css.core.elementProvider is found

jukzi:
Avoids repeated search for unsupported widgets like
	  Menu
	  MenuItem$MenuItemToolTip
	  ScrollBar
	  DropTarget
	  IME
	  Caret
	  DragSource
	  TaskBar

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/engine/RegistryCSSElementProvider.java
+++ b/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/engine/RegistryCSSElementProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2019 Manumitting Technologies Inc and others.
+ * Copyright (c) 2014, 2021 Manumitting Technologies Inc and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -54,6 +54,9 @@ public class RegistryCSSElementProvider implements IElementProvider {
 		if (provider != null) {
 			return provider.getElement(o, engine);
 		}
+		if (providerCache.containsKey(o.getClass())) {
+			return null;
+		}
 		for (Class<?> type : computeElementTypeLookup(o.getClass())) {
 			String typeName = type.getName();
 			for (IConfigurationElement ce : registry.getConfigurationElementsFor(ELEMENT_PROVIDER_EXTPOINT)) {
@@ -72,6 +75,7 @@ public class RegistryCSSElementProvider implements IElementProvider {
 				}
 			}
 		}
+		providerCache.put(o.getClass(), null);
 		return null;
 	}
 


jukzi:
It's not much time spend in that code. I just found that code during debugging and wondered why the search was done over and over again.

laeubi:
@jukzi the registry *might* change over time, so if you use more aggressive caching, it might be good to clear the cache whenever the registry changes, se comment in that file:

> // FIXME: add a registry listener to refresh caches; but would need to
> // add a dispose() to IElementProvider

vogella (approved):
LGTM

laeubi (dismissed):
This will miss later registered/installed items now!

jukzi:
> @jukzi the registry _might_ change over time, so if you use more aggressive caching, it might be good to clear the cache whenever the registry changes.

yes it might. and it might work or not i do not know. it will continue to behave like previously - but now consistent for existing and not supported widget..

jukzi:
> This will miss later registered/installed items now!

as it did (or not did) bevore

laeubi:
> > This will miss later registered/installed items now!
> 
> as it did (or not did) bevore

Nope, before it just misses a **refresh** of already found items, now it misses an **additions** of previously missing items also...

vogella:
@laeubi we could add a bundle life-cycle listener for newly installed bundles and drop the cache in case that happens but AFAIK the CSS is currently not dynamic at all so this should go into a new issue.

laeubi:
> we could add a bundle life-cycle listener for newly installed bundles and drop the cache in case that happens

Than this should be part of this change, just making things worse for

> It's not much time spend in that code.

does not seem sufficient.

merks:
Just keep in mind that when the user installs new things, they are generally **strongly** encouraged to restart and I'm willing to bet that a huge number of things don't work properly if you don't restart.  So I would not necessarily get hung up on supporting dynamically changing registry contents, if that's the concern here, because I don't think that's really worth it, and I don't think that generally works well anyway...

vogella:
> they are generally **strongly** encouraged to restart and I'm willing to bet that a huge number of things don't work properly if you don't restart.

For example, the native SWT styling (scrollbar, menu) require restart to work properly.

laeubi:
Just that we do a bad job in dynamics today does not mean we should add code that makes it even worse ... the goal should be more that we one day get rid of

> generally **strongly** encouraged to restart

merks:
@laeubi Of course you're not wrong, but that doesn't necessarily make it pragmatic either.  When we can't even change the theme without a restart, should we really invest significant effort in ensuring that the CSS cache is updated when an additional contribution is updated without a restart?  It seems a diminishing return on investment that is better spent elsewhere.  But that's just my opinion...  The opinions of others may differ...

vogella:
I agree with @laeubi that we should work on making the whole IDE story more dynamically but as the CSS engine is currently not doing this, changes which make the CSS faster / easier to debug are fine. 

As committers disagrees on this change I decide as project lead that this change is fine. 

@laeubi you may bring this to the PMC if this is important enough for you. You can also create a new issue . Currently the CSS engine does not even react if you change the current theme preference as done for example by Oomph, see for example https://bugs.eclipse.org/bugs/show_bug.cgi?id=553029

laeubi:
@merks I can only tell from the code that at least someone else has seen this problem, and probably there was a good reason to **not** cache failed attempts. So as this seems not be a large performance penalty and do not improve the general problem of not supported dynamics.

laeubi:
> changes which make the CSS faster / easier to debug

I just can't see any indication that this makes it easier to debug or faster, e.g. how much time is actually spend here and how much do we save?

Issue #262, opened 2022-08-15
Deprecate for removal org.eclipse.core.databinding.conversion.NumberToStringConverter.java/StringToNumberConverter.java classes


akurtakov:
These classes expose ICU4J API and prevent removal of ICU4J. Classes with same name in org.eclipse.core.databinding.conversion.text should be used instead.

akurtakov:
They are actually announced for removal already Planned for next release https://git.eclipse.org/c/platform/eclipse.platform.common.git/tree/bundles/org.eclipse.platform.doc.isv/porting/removals.html#n640

PR #268, opened 2022-08-16
Deprecate for removal ICU4J exposing classes

akurtakov:
Namely o.e.core.databinding.conversion.NumberToStringConverter and
StringToNumberConverter.
Fixes
https://github.com/eclipse-platform/eclipse.platform.ui/issues/262

PR #269, opened 2022-08-16
Adds forRemoval to deprecation annotation for NumberToStringConverter

vogella:
This class is planned to be removal and we should also use the
@Deprecated(forRemoval = true) information for such classes.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/262.
Also see https://github.com/eclipse-platform/.github/issues/22 which
suggests to use the code directly for removal information instead of an
external document.


vogella:
Handled by https://github.com/eclipse-platform/eclipse.platform.ui/pull/268

PR #283, opened 2022-08-23
fix #276 "Device is disposed"

### LINKED ISSUE #276: UIJob.run: Device is disposed
```
!ENTRY org.eclipse.core.jobs 4 2 2022-08-19 13:43:41.250
!MESSAGE An internal error occurred during: "Animation start".
!STACK 0
org.eclipse.swt.SWTException: Device is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Display.error(Display.java:1254)
	at org.eclipse.swt.widgets.Display.asyncExec(Display.java:793)
	at org.eclipse.ui.progress.UIJob.run(UIJob.java:84)
```

happened during org.eclipse.core.tests.internal.builders.BuilderTest

but there IS a disposed check?!

jukzi:
Device maybe disposed in another thread after dispose check.


trancexpress:
@jukzi  can you check ```org.eclipse.swt.SWTException.code``` in the catch blocks, for ```org.eclipse.swt.SWT.ERROR_DEVICE_DISPOSED``` (I think an equals is enough here, the error codes are error codes and not flags)? To be sure the code is reacting to a device disposed situation and not something else.

jukzi:
> @jukzi can you check `org.eclipse.swt.SWTException.code` 

I could, but i do not see an advantage to bloat the code since ERROR_DEVICE_DISPOSED is the only code that can be thrown in the blocks. There is also no prior art somewhere else to check for ERROR_DEVICE_DISPOSED .
However if others demand such checks as well i am not against it either.

jukzi:
BTW the only actual usage of Display.getThread is to check against current thread.
It may be worth considering to offer an exception free Display.isDisplayThread() in SWT.

trancexpress:
> > @jukzi can you check `org.eclipse.swt.SWTException.code`
> 
> I could, but i do not see an advantage to bloat the code since ERROR_DEVICE_DISPOSED is the only code that can be thrown in the blocks.

Right now, yes. In future, who knows.

HannesWell:
Could this fix also be relevant for `org.eclipse.e4.ui.progress.UIJob` or is `org.eclipse.ui.progress.UIJob` used differently?

jukzi:
i don't care about e4. probably most fixes on ui may be relevant for the copy&paste e4 things but feels like no one maintains e4.

PR #290, opened 2022-08-24
Remove rebase GH action

akurtakov:
It's github builtin action of the PRs now.


akurtakov:
GH action change only which doesn't affect regular build at all.

PR #295, opened 2022-08-25
regard external program images if zoom value is > 100% 

ptief:
If scaling is >100% the external program image should be used - currently it always uses the default object file image if the external program images is bigger than the default image with 100% scale.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/misc/ExternalProgramImageDescriptor.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/misc/ExternalProgramImageDescriptor.java
@@ -62,29 +62,15 @@ public class ExternalProgramImageDescriptor extends ImageDescriptor {
 		return createImage();
 	}
 
-	/**
-	 * Returns an SWT Image that is described by the information in this descriptor.
-	 * Each call returns a new Image.
-	 */
 	@Override
-	public ImageData getImageData() {
-		ImageData data = null;
-		ImageData defaultImage = WorkbenchImages.getImageDescriptor(ISharedImages.IMG_OBJ_FILE).getImageData();
-		if (defaultImage == null) {
-			return null;
+	public ImageData getImageData(int zoom) {
+		if (program != null) {
+			ImageData imageData = program.getImageData(zoom);
+			if (imageData != null) {
+				return imageData;
+			}
 		}
-
-		if (program == null || ((data = program.getImageData()) == null)) {
-			return defaultImage;
-		}
-
-		// The images in GNOME are too big. Scaling them does not give nice result so
-		// return defaultImage;
-		if (data.height > defaultImage.height || data.width > defaultImage.width) {
-			return defaultImage;
-		}
-
-		return data;
+		return WorkbenchImages.getImageDescriptor(ISharedImages.IMG_OBJ_FILE).getImageData(zoom);
 	}
 
 	/**


mickaelistria:
Can you provide steps to reproduce the issue and to verify the patch? Have you tested it on all 3 supported OSs?

ptief:
I only tested it on Windows 10. 
Steps to reproduce: 
1. Check your scaling settings and set it to 100%.  
2. Create a new project or open an existing project and add some files, i.e. excel or word files.
image
-> the external program icons are displayed correctly. 
3. set your scaling settings to 125%. 
4. Restart the eclipse application
5. Check the icons in the package explorer
image
-> you will get the default object image icon instead of the external program image.

mickaelistria:
@lshanmug Would you be able to verify it on macos?

lshanmug:
> @lshanmug Would you be able to verify it on macos?

Sure will test it out today.

lshanmug:
The original problem is not seen on mac, but the icons are blurred.
Tested the changes on macOS with retina display (zoom = 200), the icon size becomes small (1/4th). Please see the screenshot of before and after:

image

ptief:
@lshanmug Thanks for testing on macOS.
It seems that the cocoa HiDPI implementation does not work very well, because I changed it to the new HiDPI-aware getImageData method. 
image
Is this maybe the problem in the cocoa org.eclipse.swt.program.Program implementation ?

lshanmug:
@ptief Have you tried the patch on Windows at 200% scaling? Does it render as expected?

@mickaelistria Does it render correctly on linux at 100 and 200% scaling with the patch?

> Is this maybe the problem in the cocoa org.eclipse.swt.program.Program implementation ?

AFAICS Program.getImageData() always returns image data (16 x16) for 100% zoom without considering the device scaling. From the Program.getImageData() code, I think this is the same for Windows and Cocoa.

ptief:
@lshanmug 
You are right, at 175% and 200% scaling I have the same problem and get 16x16 images. 
For 100, 125 and 150% the zoom value is 100, for 175, 200, ... the zoom value is 200 (calculcated by org.eclipse.swt.internal.DPIUtil).
 
For win32 and cocoa org.eclipse.swt.program.Program.getImageData() uses the old method for getting the image data org.eclipse.swt.graphics.Image.getImageData() without zoom, which means it uses zoom value 100. 

I added the zoom support for org.eclipse.swt.program.Program, can you take a look please? 
swt: https://github.com/ptief/eclipse.platform.swt/tree/program_imagedata
ui: https://github.com/ptief/eclipse.platform.ui/tree/externalprogram_scaling_2

lshanmug:
> For win32 and cocoa org.eclipse.swt.program.Program.getImageData() uses the old method for getting the image data org.eclipse.swt.graphics.Image.getImageData() without zoom, which means it uses zoom value 100.
> 
> I added the zoom support for org.eclipse.swt.program.Program, can you take a look please? swt: https://github.com/ptief/eclipse.platform.swt/tree/program_imagedata ui: https://github.com/ptief/eclipse.platform.ui/tree/externalprogram_scaling_2

Creating new API for Program.getImageData(int zoom) looks like the right approach here.
SWT APIs have to be available and consistent across the three platforms - Windows, Linux and Mac, so please add the new API in Program.java in Linux as well. We need to test the behavior on Linux as well with all the changes to make sure it's working as expected.
The SWT change has to be merged first then the Platform UI change.
Can you please create PRs with the changes?

ptief:
I created a PR for SWT and updated this PR with the new API for Program.getImageData(int zoom).

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/misc/ExternalProgramImageDescriptor.java
```diff
@@ -62,29 +62,15 @@ public Image getImage() {
...
-
-		if (program == null || ((data = program.getImageData()) == null)) {
-			return defaultImage;
-		}
-
-		// The images in GNOME are too big. Scaling them does not give nice result so
-		// return defaultImage;
-		if (data.height > defaultImage.height || data.width > defaultImage.width) {
-			return defaultImage;
-		}
```
HeikoKlare, line 85: I have verified that completely removing this fallback behavior for GNOME does not produce any regressions. My environment uses GNOME version 42.9.
Result: External program images look exactly the same with and without this code on GNOME. Here are example screenshots for 100% and 200% scaling, which are exactly the same with and without that code:
image
image
Since the code has been introduced more than 20 yeras ago, it was probably some problem with GNOME environments back then.

HeikoKlare (approved):
Changes look good and have been tested in combination with https://github.com/eclipse-platform/eclipse.platform.swt/pull/409.
This should be merged right after https://github.com/eclipse-platform/eclipse.platform.swt/pull/409

HeikoKlare:
@ptief can you please squash the three commits into a single one and add a short description according to https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations to the commit message?

HeikoKlare:
For the record: this will also produce sharp icons on Mac.

Left is current state, right is with this contribution:
image

trancexpress:
This has likely caused: https://github.com/eclipse-platform/eclipse.platform.swt/issues/3003

Issue #165, opened 2022-06-21
Allow to resize launcher dialog


vogella:
We have OS  / users which cannot use Eclipse due to an incorrectly sized laucher.

https://twitter.com/dadadom/status/1537124038069321729

Any drawbacks in allow the dialog to be re-sized by the user? 

![image](https://user-images.githubusercontent.com/139910/174763314-511c7144-171e-4202-953c-bfe405261e9c.png)

laeubi:
If it is resizable we probably need a different layout (RowLayout?) to account for smaller width.

laeubi:
Beside that I really wonder if it makes any sense to use a IDE for such a small screen... but anyways if it makes users happy... why not.

merks:
That just seems silly!  Try to shrink the IDE down to that size and then imagine all the fonts are way bigger:

![image](https://user-images.githubusercontent.com/208716/174775667-cafa6e5f-de6c-45d5-8dd2-e71cf5522f3d.png)

mickaelistria:
> That just seems silly! Try to shrink the IDE down to that size and then imagine all the fonts are way bigger:

Remove toolbar & status bar, make fullscreen, expand the editor area so other views are hidden; imagine user uses some voice control application to navigate the IDE and it's actually decent.

merks:
I can also imagine trolls making up excuses just to have something to rag about and to see if they can make other people ask "how high?" when they say "jump". 

I'm not sure why we'd put weight into something this marginal when there are literally so many bugs/issues we want to auto close them...

mickaelistria:
> I'm not sure why we'd put weight into something this marginal when there are literally so many bugs/issues we want to auto close them...

It's up to contributors to decide where they put priority. I personally value things like "being able to develop on a phone or a small tablet" more than supporting older versions of Java or other frameworks. YMMV.

ronso0:
Would you mind re-opeing this issue?

I'd love to make the dialog wider because with my workspace paths I'm never able to see the entire name, the trailing part is almost always hidden, exactly like the OP illustrates:
![image](https://user-images.githubusercontent.com/139910/174763314-511c7144-171e-4202-953c-bfe405261e9c.png)

Possible solutions:
* allow resizing
* if that is not feasible, make better use of the available space:
  * remove the space(r) at the right so the path combobox can expand
  * move the path combobox and the "Browse..." button below "Workspace:"

laeubi:
[...]

PR #299, opened 2022-08-29
Improve responsiveness of workspace path chooser

laeubi:
Currently the workspace path chooser has very poor responsiveness (see https://github.com/eclipse-platform/eclipse.platform.ui/issues/165#issuecomment-1229567595):

- if the dialog is small, some parts are cut of
- if the dialog is large, the available space is not filled
- some space is even wasted by a rather useless label

This change do the following:
- remove the label
- let the combobox grow and shrink (to a certain degree)
- let the button shrink if less space is available

### Very small dialog:
grafik

### Very large dialog:
grafik

FYI: @ronso0 @szarnekow

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
@@ -16,6 +16,7 @@
 package org.eclipse.ui.internal.ide;
 
 import java.io.File;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -40,11 +41,14 @@ import org.eclipse.jface.window.Window;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.osgi.util.TextProcessor;
 import org.eclipse.swt.SWT;
-import org.eclipse.swt.custom.CLabel;
+import org.eclipse.swt.events.ControlEvent;
+import org.eclipse.swt.events.ControlListener;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.swt.layout.BorderData;
+import org.eclipse.swt.layout.BorderLayout;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.layout.RowData;
@@ -71,6 +75,9 @@ import org.osgi.framework.FrameworkUtil;
  * A dialog that prompts for a directory to use as a workspace.
  */
 public class ChooseWorkspaceDialog extends TitleAreaDialog {
+
+	private static final String OPEN_FOLDER_EMOJI = new String(
+			new byte[] { (byte) 0xF0, (byte) 0x9F, (byte) 0x93, (byte) 0x82 }, StandardCharsets.UTF_8);
 
 	private static final String DIALOG_SETTINGS_SECTION = "ChooseWorkspaceDialogSettings"; //$NON-NLS-1$
 
@@ -423,18 +430,13 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 	private void createWorkspaceBrowseRow(Composite parent) {
 		Composite panel = new Composite(parent, SWT.NONE);
 
-		GridLayout layout = new GridLayout(3, false);
+		BorderLayout layout = new BorderLayout();
 		layout.marginHeight = convertVerticalDLUsToPixels(IDialogConstants.VERTICAL_MARGIN);
 		layout.marginWidth = convertHorizontalDLUsToPixels(IDialogConstants.HORIZONTAL_MARGIN);
-		layout.verticalSpacing = convertVerticalDLUsToPixels(IDialogConstants.VERTICAL_SPACING);
-		layout.horizontalSpacing = convertHorizontalDLUsToPixels(IDialogConstants.HORIZONTAL_SPACING);
+		layout.spacing = convertVerticalDLUsToPixels(IDialogConstants.VERTICAL_SPACING);
 		panel.setLayout(layout);
 		panel.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
 		panel.setFont(parent.getFont());
-
-		CLabel label = new CLabel(panel, SWT.NONE);
-		label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_workspaceEntryLabel);
-		label.setMargins(0, 0, 2, 0);
 
 		text = new Combo(panel, SWT.BORDER | SWT.LEAD | SWT.DROP_DOWN);
 		new DirectoryProposalContentAssist().apply(text);
@@ -459,10 +461,9 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 
 		Button browseButton = new Button(panel, SWT.PUSH);
 		browseButton.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_browseLabel);
+		browseButton.setToolTipText(IDEWorkbenchMessages.ChooseWorkspaceDialog_browseTooltip);
 		setButtonLayoutData(browseButton);
-		GridData data = (GridData) browseButton.getLayoutData();
-		data.horizontalAlignment = GridData.HORIZONTAL_ALIGN_END;
-		browseButton.setLayoutData(data);
+		browseButton.setLayoutData(new BorderData(SWT.RIGHT));
 		browseButton.addSelectionListener(new SelectionAdapter() {
 			@Override
 			public void widgetSelected(SelectionEvent e) {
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -452,9 +451,7 @@ private void createWorkspaceBrowseRow(Composite parent) {
 		Button browseButton = new Button(panel, SWT.PUSH);
 		browseButton.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_browseLabel);
```
vogella, line 463: "Workspace:" as tooltip looks poor, please adjust the text to be real tooltip

vogella (dismissed):
Tooltip shoul be improved. I also would suggest to wait with this change for the master to open up for the next release.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
laeubi: Feel free to propose a better text ;-)
Actually we could also remove this at all ....

vogella: "Enter a directory as workspace" would be better.

ronso0: my 2 cents:
A path combobox and an attached `...` button are standard UI elements, so I doubt if the button requires a label at all.
Though, if this is considered a bugfix / UI polish, there shouldn't be significant UI (text) changes IINM, and making the dialog elements responsive is sufficient for now, no?

laeubi: I removed the useless text from the tooltip now.

merks:
One thing to keep in mind is that a screen reader for the visually impaired might benefit from a label that might otherwise be considered pointless.

https://github.com/eclipse-platform/eclipse.platform.swt/pull/353

I'm not asserting that's the case here, but rather to suggest it might something we shouldn't overlook...

laeubi:
> One thing to keep in mind is that a screen reader for the visually impaired might benefit from a label that might otherwise be considered pointless.

I have no clue if SWT supports something like https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

Because regardless of visually impaired or not, I doubt that "Workspace" has any meaningful semantics in this context and I most probably want to set something like 

> A combo-box that represents the path to the workspace, either directly given or chosen by the 'browse' button

might be more appropriate for those who can't see the dialog to get a better impression whats going on here.

laeubi:
@niraj-modi can you give some guidance on how to best handle this for screenreaders?

tomaswolf:
> how to best handle this for screenreaders?

If you don't want a label, see `Control.getAccessible().addAccessibleListener()`. An `AccessibleAdapter` could be used to set a name for the widget.

laeubi:
As at EclipseCon it was again complained about the Workspace chooser, is there any interest in even this small improvement?

I rebased it but lost a bit of overview what should be done.

1. better tooltip (as suggested by @vogella )
2. maybe remove tooltip at all and instead use Control.getAccessible().addAccessibleListener() like @tomaswolf proposed?

Would be good to get this done ...

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -476,6 +472,26 @@ public void widgetSelected(SelectionEvent e) {
...
 			}
 		});
+		parent.addControlListener(new ControlListener() {
+
+			@Override
+			public void controlResized(ControlEvent e) {
+				// browseButton
+				Point size = parent.getSize();
+				if (size.x < 300) {
+					browseButton.setText("..."); //$NON-NLS-1$
```
mickaelistria: What about using a directory emoji: 📂 ?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -476,6 +472,26 @@ public void widgetSelected(SelectionEvent e) {
 				}
 			}
 		});
+		parent.addControlListener(new ControlListener() {
+
+			@Override
+			public void controlResized(ControlEvent e) {
+				// browseButton
+				Point size = parent.getSize();
+				if (size.x < 300) {
```
mickaelistria: Why 300 as magic number here?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
merks: There are a great many Browse... in the IDE so there is the argument for consistency.  

image

Surely needing a really narrow dialog is an outlier problem.  If it isn't, then that will be a pervasive problem everywhere when using the IDE... 

Improvements to see long paths I really like though I just question the real need for really narrow...

ronso0: Good idea. Though, is this supported on all platforms? Is it used elsewhere in eclipse?

laeubi: I found Emojis problematic as support across platform is not guaranteed. I once recommended to have some standard SWT support for a cross platform icon font but it was not getting much attention / acceptance.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
laeubi: I just tried it out and found that it at least is good for a first iteration.

mickaelistria: Hardcoding magic numbers is a bad practice. You should find a way to deduce a value from the button size or the stringExtent, or whatever; but not hardcode a randomish number.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
mickaelistria: > I found Emojis problematic as support across platform is not guaranteed

That's probably an outdated impression and an invalid argument. SWT supports emojis very well on all Platform described in https://www.eclipse.org/projects/project-plan.php?planurl=http://www.eclipse.org/eclipse/development/plans/eclipse_project_plan_4_26.xml , and those Platforms do support emojis as well.
If you have some counter-example, please report.

> . I once recommended to have some standard SWT support for a cross platform icon font but it was not getting much attention / acceptance.

It is IMO a bad idea: first the supported OS do already support some cross-platform icons (emojis), and then SWT concept is to integrate in native. Creating another icon palette goes to the opposite goal of what SWT tries to achieve; it would be similar to adding more custom widgets in SWT and rely less and less on native, also emojis have native accessibility capabilities that are much stronger than what SWT would be able to offer by reimplementing some icons; and finally, emojis are cheap and already existing and available and easy t use now, while your suggestion would be more expensive to develop and adopt.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
laeubi: Feel free to suggest something, I have not really a better idea here and as all swt layout use "hardcoded numbers" (e.g. non of them allows percentages or other relatives like css) and have no way for really responsive layouts (see https://github.com/eclipse-platform/eclipse.platform.swt/issues/355) this is still better than nothing for now.

mickaelistria: You can implement something based on percentage: get the initial button width and the parent width; and decide from those of a condition in which button should be shortened, for example `if (size.x < 3 * initialButtonSize.x)`

laeubi: As said this is the concern of a layout manager and not the user-code. And the button size do not really matters here. What matters are actually total available space, so the decision here is that 300 (dpi corrected) px is "small" and then the button should shrink... this do not mean its better if getting even smaller it is just the button leaves more space compared to the default by using a shorter text.

mickaelistria: So what really matters is more the amount of characters that can fit in the text widget (overall 50). The size could be deduced from it, as this depends eg on the system font.

laeubi: Not really, as you can have long path, short path, ... in the end small is small (maybe too small as @merks suggested) so all we can do is to same some little space to show maybe some more characters. just from UX, it might be even better to have a text-field that works like CLabel with an elepis (...) in the middle but that far out of scope here.

The inital scope was, that on small screens you cannot use the dialog at all (whether or not Eclipse itself would be usable on such a screen). AND there was a complain that on larger screens the size is not used but all grey.

So this just improves the cases in two way:

1. define what "small" is (== 300) and try to give some more space, and shrink the box so probably users can see a bit more
2. If it is "large" ( > 300)  stretch the text-box so it fills all the space.

Not more not less, no fancy computation just a little bit more usable.

laeubi:
@vogella I have now added an own tooltip
@mickaelistria I using now a "open folder icon" (that should be readable by screenreaders as well) and computing the size dynamically based on the font size.

If no one objects I'd like to merge this before M1 so we have time for improvements (even though most people might never notice that change).

PR #302, opened 2022-08-30
Ctrl+E: Prepend parent directories for file name collisions

### LINKED ISSUE #301: Ctrl+E: Prepend parent directories when resources have the same name
When doing web development I often have html files that share the same name but are in different folders.
In Eclipses Ctrl+E menu I can't differentiate between and I'm left guessing and switching to files until I find the one I'm looking for.

The workaround that I found works by going through "Open Resources" and searching for "${foldername}/${filename}" but I think the Ctrl+E menu should also support this use case.

### Current state
Selection_191

### Proposed solution
Selection_214

---

Transferred from https://bugs.eclipse.org/bugs/show_bug.cgi?id=576994

fapdash:
If multiple EditorReferences share the same file name the user can't
differentiate between them without hovering on the table item.
By prepending parent directories when necessary the files
are distinguishable immediately and the user can filter for the
parent directory.

closes #301

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
@@ -118,15 +118,15 @@ public class WorkbookEditorsHandler extends FilteredTableBaseHandler {
 	 * Generates a mapping of EditorReferences to display label texts. If display
 	 * names collide parent directories will be added until the EditorReference can
 	 * be differentiated from all references that share the same file name. If the
-	 * collisions share multiple paths segments only the first differing path
-	 * segment will be prepended.<br>
+	 * collisions share multiple paths segments the shared path segments are
+	 * omitted.<br>
 	 * <br>
 	 * Example where all collisions share the same segments:
 	 *
 	 * <pre>
-	 * /project/test1/foo/bar/file -> test1/file
-	 * /project/test2/foo/bar/file -> test2/file
-	 * /project/test3/foo/bar/file -> test3/file
+	 * /project/test1/foo/bar/file -> test1/.../file
+	 * /project/test2/foo/bar/file -> test2/.../file
+	 * /project/test3/foo/bar/file -> test3/.../file
 	 * </pre>
 	 *
 	 * Example with differing segments:
@@ -199,8 +199,8 @@ public class WorkbookEditorsHandler extends FilteredTableBaseHandler {
 					Integer maxMatchingSegment = maxMatchingSegmentsList.get(i);
 					IPath path = groupedEditorReferences.get(i).getValue();
 
-					String labelText = generateLabelText(editorReference, path, differingMaxSegmentsCounter,
-							maxMatchingSegment);
+					String labelText = generateLabelText(editorReference, path, maxMatchingSegment,
+							differingMaxSegmentsCounter.size() == 1);
 					editorReferenceLabelTexts.put(editorReference, labelText);
 				}
 			}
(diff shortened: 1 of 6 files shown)


tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/internal/WorkbookEditorsHandlerTest.java
```diff
@@ -0,0 +1,237 @@
+package org.eclipse.ui.tests.internal;
```
akurtakov, line 15: EPL-2.0 License header is missing.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/internal/WorkbookEditorsHandlerTest.java
```diff
@@ -0,0 +1,237 @@
...
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.commands.ICommandService;
+import org.eclipse.ui.handlers.IHandlerService;
+import org.eclipse.ui.ide.IDE;
+import org.eclipse.ui.internal.WorkbookEditorsHandler;
+import org.eclipse.ui.tests.harness.util.FileUtil;
+import org.eclipse.ui.tests.harness.util.UITestCase;
+import org.junit.Test;
+
+/**
```
akurtakov: This since tag is not needed in tests and is even wrong as current version is 3.15. Please remove.

vogella:
@akurtakov Why do we not have a rebase button here?

laeubi:
grafik

akurtakov:
> @akurtakov Why do we not have a rebase button here?

It's a dropdown for me and I just rebased.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
```diff
@@ -98,6 +114,159 @@ private List<EditorReference> getParts(WorkbenchPage page) {
...
+	 * Generates a mapping of EditorReferences to display label texts. If display
+	 * names collide parent directories will be added until the EditorReference can
+	 * be differentiated from all references that share the same file name. If the
+	 * collisions share multiple paths segments only the first differing path
+	 * segment will be prepended.<br>
+	 * <br>
+	 * Example where all collisions share the same segments:
+	 *
+	 * <pre>
+	 * /project/test1/foo/bar/file -> test1/file
```
mickaelistria: we would expect `test1/.../file` here (or even `file (in test1/...)`, silently skipping intermediary segments can be interpreted as a bug.

mickaelistria (changes requested):
The general idea seems good.
One usability issue it seems to cause is that the filter text then doesn't match the file: if I have `folder1/pom.xml` and `folder2/pom.xml` and hundreds of Java file, and do Ctrl+E and type pom.xml, I would like both pom.xml to remain visible. So the filter should also apply on the last segment.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/internal/WorkbookEditorsHandlerTest.java
fapdash: Weird, I've set up my Committer Eclipse through the Oomph installer and this comment gets auto created for new test classes. (always with version 3.5)

fapdash:
> One usability issue it seems to cause is that the filter text then doesn't match the file: if I have folder1/pom.xml and folder2/pom.xml and hundreds of Java file, and do Ctrl+E and type pom.xml, I would like both pom.xml to remain visible. So the filter should also apply on the last segment.

@mickaelistria I can't reproduce this. When I filter for `pom.xml` I only see the two `pom.xml` files and all the open Java files disappear. What I noticed is that I couldn't filter for the folder since the Filter only considered the last segment of the path (editor title).
I've changed the code to match on the editor title and the display label.

Selection_220
Selection_221

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
fapdash: I've implemented the  `test1/.../file` variant, the user can see the full path in the tooltip.

Selection_222

fapdash: I've also thought about styling the marker for the omitted part differently, maybe grey it out, but then it gets confusing for the user with regards to the filter behavior.
Should the filter match on the greyed out `/.../` or should the user be able to match on `test1/file`, ignoring the `/.../` part?
Maybe it's better to keep the styling as it is. :shrug:

mickaelistria:
OK, I probably looked at the wrong file (it was a pom.xml and the Maven editor does change the label), I'll try it again.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
mickaelistria: IMO, current/default styling is good enough.

mickaelistria:
I see a test is now failing. Can you please investigate?

mickaelistria (approved):

mickaelistria:
Thank you!

PR #310, opened 2022-09-02
Fixed incorrect Help icon (should be 16x16 rather than 24x24) #577318

orionll:
Please review and merge the fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=577318


akurtakov:
On the first change (for a release) in a plugin version has to be bumped x.y.z+100 (for bugfix - your case). Please do so, the build will verify it and it's ready to go.

akurtakov:
Thanks for the patch!

PR #311, opened 2022-09-06
Fix comparator error in I20220905-1800

akurtakov:
Caused by JobInfo implementing equals.


akurtakov:
Push as it's just version bump.

PR #298, opened 2022-08-26
Avoid NPE in JFaceResources.getResources after Device dispose #305

jukzi:
* Improve jface error message.
* Avoid error in WorkbenchPart.dispose() if device is already destroyed.

PR #315, opened 2022-09-06
Fix endless loop with duplicates in recent workspaces #312

### LINKED ISSUE #312: Duplicates in recent workspace cause endless loop on switch workspace
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=531611

Users ran into the problem "naturally", we are still not sure how.

```
	at org.eclipse.ui.internal.ide.ChooseWorkspaceDialog.lambda$6(ChooseWorkspaceDialog.java:390)
	at org.eclipse.ui.internal.ide.ChooseWorkspaceDialog$$Lambda$1280/0x0000000841517c40.apply(Unknown Source)
	at java.util.stream.ReferencePipeline$3$1.accept(java.base@11.0.10/ReferencePipeline.java:195)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(java.base@11.0.10/ArrayList.java:1655)
	at java.util.stream.AbstractPipeline.copyInto(java.base@11.0.10/AbstractPipeline.java:484)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(java.base@11.0.10/AbstractPipeline.java:474)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(java.base@11.0.10/ReduceOps.java:913)
	at java.util.stream.AbstractPipeline.evaluate(java.base@11.0.10/AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(java.base@11.0.10/ReferencePipeline.java:578)
	at org.eclipse.ui.internal.ide.ChooseWorkspaceDialog.createUniqueWorkspaceNameMap(ChooseWorkspaceDialog.java:395)
	at org.eclipse.ui.internal.ide.ChooseWorkspaceDialog.createRecentWorkspacesComposite(ChooseWorkspaceDialog.java:329)
	at org.eclipse.ui.internal.ide.ChooseWorkspaceDialog.createDialogArea(ChooseWorkspaceDialog.java:183)
	at org.eclipse.ui.internal.ide.ChooseWorkspaceWithSettingsDialog.createDialogArea(ChooseWorkspaceWithSettingsDialog.java:108)
```

Snippet to reproduce (code is copy-pasted from ```ChooseWorkspaceDialog.createUniqueWorkspaceNameMap()```):

```
		String[] recentWorkspaceNames = {
				"/bkp1",
				"/bkp1",
		};
		// Convert workspace paths to arrays of single path segments
		List<String[]> splittedWorkspaceNames = Arrays.asList(recentWorkspaceNames).stream()
				.filter(s -> s != null && !s.isEmpty()).map(s -> s.split(Pattern.quote(fileSeparator)))
				.collect(Collectors.toList());
		
		// create and collect unique workspace keys produced from arrays,
		// try to generate unique keys starting with the last segment of the
		// workspace path, increasing number of segments if no unique names
		// could be generated,
		// loop until all array values are removed from array list
		for (int i = 1; !splittedWorkspaceNames.isEmpty(); i++) {
			final int c = i;

			System.out.println(i);
			// Function which flattens arrays to (hopefully unique) keys
			Function<String[], String> stringArraytoName = s -> String.join(fileSeparator,
					s.length < c ? s : Arrays.copyOfRange(s, s.length - c, s.length));

			// list of found unique keys
			List<String> uniqueNames = splittedWorkspaceNames.stream().map(stringArraytoName)
					.collect(Collectors.groupingBy(s -> s, Collectors.counting())).entrySet().stream()
					.filter(e -> e.getValue() == 1).map(Entry::getKey).collect(Collectors.toList());

			// remove paths for which we have found unique keys
			splittedWorkspaceNames.removeIf(a -> {
				String joined = stringArraytoName.apply(a);
				if (uniqueNames.contains(joined)) {
					uniqueWorkspaceNameMap.put(joined, String.join(fileSeparator, a));
					return true;
				}
				return false;
			});
		}
```

Same points from the Eclipse buzilla ticket still hold:

1. Code should not loop endlessly.
2. The lambdas need to be broken into smaller sentences. What the code does in its current form is anyones guess.

trancexpress:
ChooseWorkspaceDialog.createUniqueWorkspaceNameMap() finds a unique
short name for a recent workspace path, based on the last segments of
the workspace path. If the recent workspace paths contain duplicates,
the method loops endlessly - the exit condition of the method is that a
unique name is found for each recent workspace.

This change filters duplicates from recent workspace paths in
ChooseWorkspaceDialog. Duplicates in this context are workspace paths
that represent the same location, but are different string-wise due to
superfluous separators.

In addition, an extra exit condition is added to the loop in
ChooseWorkspaceDialog.createUniqueWorkspaceNameMap(), to prevent endless
loops in future.

Fixes: #312
Signed-off-by: Simeon Andreev <simeon.danailov.andreev@gmail.com>

### DIFF
--- a/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.ide; singleton:=true
-Bundle-Version: 3.19.100.qualifier
+Bundle-Version: 3.19.200.qualifier
 Bundle-ClassPath: .
 Bundle-Activator: org.eclipse.ui.internal.ide.IDEWorkbenchPlugin
 Bundle-ActivationPolicy: lazy
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
@@ -18,10 +18,13 @@ package org.eclipse.ui.internal.ide;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -325,13 +328,12 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 		layout.marginLeft = 14;
 		layout.spacing = 6;
 		panel.setLayout(layout);
-		recentWorkspacesLinks = new HashMap<>(launchData.getRecentWorkspaces().length);
+		List<String> recentWorkspaces = getRecentWorkspaces();
+		recentWorkspacesLinks = new HashMap<>(recentWorkspaces.size());
 		Map<String, String> uniqueWorkspaceNames = createUniqueWorkspaceNameMap();
 
-		List<String> recentWorkspacesList = Arrays.asList(launchData.getRecentWorkspaces()).stream()
-				.filter(s -> s != null && !s.isEmpty()).collect(Collectors.toList());
 		List<Entry<String, String>> sortedList = uniqueWorkspaceNames.entrySet().stream().sorted((e1, e2) -> Integer
-				.compare(recentWorkspacesList.indexOf(e1.getValue()), recentWorkspacesList.indexOf(e2.getValue())))
+				.compare(recentWorkspaces.indexOf(e1.getValue()), recentWorkspaces.indexOf(e2.getValue())))
 				.collect(Collectors.toList());
 
 		for (Entry<String, String> uniqueWorkspaceEntry : sortedList) {
@@ -374,16 +376,22 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 		Map<String, String> uniqueWorkspaceNameMap = new HashMap<>();
 
 		// Convert workspace paths to arrays of single path segments
-		List<String[]> splittedWorkspaceNames = Arrays.asList(launchData.getRecentWorkspaces()).stream()
+		List<String[]> splittedWorkspaceNames = getRecentWorkspaces().stream()
 				.filter(s -> s != null && !s.isEmpty()).map(s -> s.split(Pattern.quote(fileSeparator)))
 				.collect(Collectors.toList());
+
+		// bug 531611: prevent endless loops
+		int maxSegmentsCount = 0;
+		for (String[] strings : splittedWorkspaceNames) {
+			maxSegmentsCount = Math.max(0, strings.length);
+		}
 
 		// create and collect unique workspace keys produced from arrays,
 		// try to generate unique keys starting with the last segment of the
 		// workspace path, increasing number of segments if no unique names
 		// could be generated,
 		// loop until all array values are removed from array list
-		for (int i = 1; !splittedWorkspaceNames.isEmpty(); i++) {
+		for (int i = 1; !splittedWorkspaceNames.isEmpty() && i <= maxSegmentsCount; i++) {
 			final int c = i;
 
 			// Function which flattens arrays to (hopefully unique) keys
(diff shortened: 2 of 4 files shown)


trancexpress:
@iloveeclipse any idea how I can test the change? I checked whether duplicate paths are filtered as expected, in a snippet. But we should also test if the workspace chooser dialog works as expected.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -325,10 +327,10 @@ public void expansionStateChanged(ExpansionEvent e) {
 		layout.marginLeft = 14;
 		layout.spacing = 6;
 		panel.setLayout(layout);
-		recentWorkspacesLinks = new HashMap<>(launchData.getRecentWorkspaces().length);
+		recentWorkspacesLinks = new HashMap<>(getRecentWorkspaces().length);
```
iloveeclipse: Could you please extract `getRecentWorkspaces()` to a variable?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -596,4 +604,24 @@ public Button getDefaultButton() {
...
+				if (workspace.startsWith(File.separator)) {
+					normalizedPath = File.separator + normalizedPath;
+				}
+				boolean nonDuplicate = normalizedPaths.add(normalizedPath);
+				if (nonDuplicate) {
+					recentWorkspaces.add(workspace);
+				}
+			}
+		}
+		return recentWorkspaces.toArray(String[]::new);
```
iloveeclipse: Looking at uses of this return value I see that they convert array back to a list.
It would make sense to change return value of this method to a List too.

iloveeclipse:
Please bump bundle version (and pom if there is one)

bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.ide; singleton:=true
-Bundle-Version: 3.19.100.qualifier
+Bundle-Version: 3.19.200.qualifier
```
trancexpress, line 5: Fail is:

```
 Error: 0.482 [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:3.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jface: Only qualifier changed for (org.eclipse.jface/3.27.0.v20220906-1112). Expected to have bigger x.y.z than what is available in baseline (3.27.0.v20220817-1444) -> [Help 1]
```

Am I supposed to bump some other plug-in here?

mickaelistria: Not really as part of this change. Can you please just make sure you change is properly rebased on top of master?

iloveeclipse: > Not really as part of this change. Can you please just make sure you change is properly rebased on top of master?

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/298#issuecomment-1239293153

iloveeclipse: I will push a fix in https://github.com/eclipse-platform/eclipse.platform.ui/pull/318, which also was suffering from that issue

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/internal/ide/ChooseWorkspaceDialogTests.java
```diff
@@ -0,0 +1,50 @@
+/*******************************************************************************
```
iloveeclipse, line 1: Haven't checked, just wondering: don't we need to bump test bundle version too, and don't we need to add the test to the test suite?

iloveeclipse (approved):
LGTM for the fix, but see question regarding test

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/internal/ide/ChooseWorkspaceDialogTests.java
trancexpress: > Haven't checked, just wondering: don't we need to bump test bundle version too, and don't we need to add the test to the test suite?

The bundle version seems to have been bumped yesterday.

For a test suite, the other tests in the same package are not included into one. But other tests I check are included in a suite... I'll see if I can find a place for the test.

iloveeclipse:
@trancexpress : **please** rebase your patch before pushing.

PR #323, opened 2022-09-08
Databinding api deletion

vogella:
Bugzilla ticket: https://bugs.eclipse.org/bugs/show_bug.cgi?id=546820


vogella:
org.eclipse.ui.tests are missing...

bundles/org.eclipse.core.databinding.beans/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.databinding.beans
-Bundle-Version: 1.8.0.qualifier
+Bundle-Version: 1.9.0.qualifier
```
jensli, line 5: Just a thought that I find interesting:

This Eclipse way-of-working with "controlled deletions" could benefit from a 5-part version string:

1: Qualifier
2: Patch
3: Minor, backwards compatible
4: New part: Medium, only controlled, previously announced deletions
5: Major, Not backwards compatible

Just a thought.

vogella:
Thanks @jensli I moved the documentation update to its own pull request and plan to merge this once the verifcation build is done.

bundles/org.eclipse.core.databinding.beans/META-INF/MANIFEST.MF
vogella: Yeah, but that would be again a "Eclipse" only solution which nobody outside the Eclipse world would be using or undersatnd.

jensli:
I have only had a rather quick look, but it seem good to me.

jensli:
I don't know how to balance the benefit of simplifying the code base by removing these old classes against the harm of breaking old unmaintained plugins that uses them. But it seems like you have a made a decision about that already!

jensli (approved):

mickaelistria:
> I don't know how to balance the benefit of simplifying the code base by removing these old classes against the harm of breaking old unmaintained plugins that uses them. But it seems like you have a made a decision about that already!

That was discussed multiple times and it's indeed not all black nor white and it may depend on the particular API. But in general, there is a process in place to minimize the risk. As for unmaintained plugins, well, they're unmaintained so it's a sign that no-one cares enough to keep them alive, so Platform project shouldn't really care either; there is no commitment for Platform that everything that has worked once with it will keep running forever (that would be crazy!). On the other end, there is a commitment from Platform that it does it best to remain sustainable and relevant, and getting rid of older API or improving code in general is acting towards that goal.

vogella:
@mickaelistria or @laeubi 

Is this a new / known Tycho problems?

Error: 5.816 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:3.0.0-SNAPSHOT:validate-classpath (default-validate-classpath) on project org.eclipse.core.commands: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:3.0.0-SNAPSHOT:validate-classpath failed: Plugin org.eclipse.tycho:tycho-compiler-plugin:3.0.0-SNAPSHOT or one of its dependencies could not be resolved: Could not find artifact org.eclipse.jdt:ecj:jar:3.31.0-SNAPSHOT in tycho-snapshots (https://repo.eclipse.org/content/repositories/tycho-snapshots/) -> [Help 1]

vogella:
Rebuild worked fine, so I guess this was caused by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/581 which was reverted.

Thanks @jensli for the help, I will push the Javadoc update via https://github.com/eclipse-platform/eclipse.platform.ui/pull/341

iloveeclipse:
Please check https://github.com/eclipse-pde/eclipse.pde/issues/339

Issue #321, opened 2022-09-08
NPE in PartRenderingEngine#subscribeTrimHandler when removing MTrimBar from MTrimmedWindow.getTrimBars()


nbros:
I implemented a backup/restore for a workbench layout. As part of this process, it removes the old `MTrimBar`s before adding the saved ones back to the model (in the UI thread).
I'm getting a `NullPointerException` in `PartRenderingEngine#subscribeTrimHandler` when doing that.

I see that when `UIEvents.isREMOVE(event)` is true, the Eclipse code tries to get the `UIEvents.EventTags.NEW_VALUE` : shouldn't that be `OLD_VALUE` when removing? Otherwise the `NEW_VALUE` returns `null` and `removed.getRenderer()` causes a NPE.

This is in:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/d15de47c5c6e62e6d6bcc73798e0a4b54466142d/bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/PartRenderingEngine.java#L265-L272

```
java.lang.NullPointerException
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.subscribeTrimHandler(PartRenderingEngine.java:254)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:90)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:55)
	at java.lang.reflect.Method.invoke(Method.java:508)
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:55)
	... 88 more
```
(the line numbers don't match because the product I'm developing is based on Eclipse 4.6.2, but the Eclipse code on `master` still has the same issue)

mickaelistria:
Your analysis seems correct. Can you please submit a PR that would fix the issue as you suggested?

elsazac:
The issue seems to be fixed via the linked PR. So this can be closed.

PR #328, opened 2022-09-11
fix NPE in PartRenderingEngine#subscribeTrimHandler

nbros:
fixes Issue #321: NPE in PartRenderingEngine#subscribeTrimHandler when removing MTrimBar from MTrimmedWindow.getTrimBars()

When `UIEvents.isREMOVE(event)` is true, the Eclipse code tried to get the `UIEvents.EventTags.NEW_VALUE` which returns `null` and causes `removed.getRenderer()` to throw a NPE.

This pull request looks for the element to **remove** in `OLD_VALUE` instead of `NEW_VALUE`.


mickaelistria:
Can you please

* Bump bundle version by +0.0.100 as it is the first change since last release
* Sign the Eclipse Contributor Agreement (or verify the mail address of the git commit matches the one you signed the agreement with) https://accounts.eclipse.org/legal/eca/validation/119275

mickaelistria:
Thank you!

PR #334, opened 2022-09-14
Remove deprecated method and constructor in PopupDialog

vogella:
The method getPersistBounds() is removed from class org.eclipse.jface.dialogs.PopupDialog. Clients are encouraged to use the method getPersistLocation() or getPersistSize() instead. Also remove the deprecated constructor. Clients should use the not deprecated constructor.

See bug 531913.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/PopupDialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/PopupDialog.java
@@ -366,58 +366,6 @@ public class PopupDialog extends Window {
 	 * Constructs a new instance of <code>PopupDialog</code>.
 	 *
 	 * @param parent
-	 *                               The parent shell.
-	 * @param shellStyle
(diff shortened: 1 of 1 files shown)


vogella:
Still needs API filters I hope to finish https://github.com/eclipse-platform/eclipse.platform.ui/pull/323 first.

vogella:
Not sure why API check does not complain but lets believe it....

@HannesWell FYI, this is one of a deletions I usually do per release, something important but nice to reduce the code base from time to time.

vogella:
API removal update https://github.com/eclipse-platform/eclipse.platform.common/pull/75

HannesWell:
> @HannesWell FYI, this is one of a deletions I usually do per release, something important but nice to reduce the code base from time to time.

Yes, I also appreciate to get rid of unnecessary code. It just makes the code base simpler to understand if there is less to understand and therefore simplifies maintenance, even the code itself is not actively maintained.

PR #164, opened 2022-06-20
Remove dispose check from ButtonElement

vogella:
In https://github.com/eclipse-platform/eclipse.platform.ui/pull/163#discussion_r901669089
I learned that SWT does not deliver events to disposed widgets. Hence we
can remove the check in ButtonElement (used by CSS engine). This way we
will prevent copy and paste errors for new css handlers.

PR #344, opened 2022-09-20
Revert "Remove dispose check from ButtonElement"

vogella:
The assumption that this is not needed because SWT handles it was wrong, see https://github.com/eclipse-platform/eclipse.platform.ui/pull/164#issuecomment-1229193596

This reverts commit 46b1d61bc8bc812514f2d71c8b1c9fbce3b96c47.

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ButtonElement.java
+++ b/bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ButtonElement.java
@@ -34,8 +34,10 @@ public class ButtonElement extends ControlElement {
 	private SelectionListener selectionListener = new SelectionAdapter() {
 		@Override
 		public void widgetSelected(SelectionEvent e) {
-			ButtonElement.this.isSelected = getButton().getSelection();
-			doApplyStyles();
+			if (!e.widget.isDisposed()) {
+				ButtonElement.this.isSelected = getButton().getSelection();
+				doApplyStyles();
+			}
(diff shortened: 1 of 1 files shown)


vogella:
@laeubi didn't I see you implementing a GH action to make the needed version increase visible in the PR? If yes, please point me to the place in which you use this, so that I can apply this also to this repo.

laeubi:
@vogella yes, let me se if I can find that (still tooo much repositories!)

In general it works via matchers like here:
https://github.com/eclipse-platform/eclipse.platform/blob/master/.github/matcher.json
and then you enable that for an action:
https://github.com/eclipse-platform/eclipse.platform/blob/3e316d7704de3cec46716eace69b1cd15f467c54/.github/workflows/build.yml#L24

laeubi:
I found a baseline warning matcher i created once for SWT(but didn't merged it) :
https://github.com/laeubi/eclipse.platform.swt/blob/7e43dfb26c32ad662efa5b247cc32b75e85fa631/.github/matcher.json

basically it is just a matter of finding a good regexp to match the desired text...

vogella:
> I found a baseline warning matcher i created once for SWT(but didn't merged it) : https://github.com/laeubi/eclipse.platform.swt/blob/7e43dfb26c32ad662efa5b247cc32b75e85fa631/.github/matcher.json
> 
> basically it is just a matter of finding a good regexp to match the desired text...

Thanks. Once this is applied to one of the repos and working, I will copy it to this one.

vogella:
As this is a regression (if the dynamic update of CSS is used), I will merge it to include it into M1.

PR #348, opened 2022-09-20
Issue 86 help link

eobrienPilz:
This is a fix for #86

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/ExtendedMarkersView.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/ExtendedMarkersView.java
@@ -1103,12 +1103,17 @@ public class ExtendedMarkersView extends ViewPart {
 								id)));
 	}
 
+	protected FiltersConfigurationDialog createFilterConfigurationDialog(MarkerContentGenerator gen) {
+		FiltersConfigurationDialog dialog = new FiltersConfigurationDialog(getSite().getWorkbenchWindow().getShell(),
+				gen);
+		return dialog;
+	}
+
 	/**
 	 * Open the filters dialog for the receiver.
 	 */
 	void openFiltersDialog() {
-		FiltersConfigurationDialog dialog = new FiltersConfigurationDialog(
-				getSite().getWorkbenchWindow().getShell(), generator);
+		FiltersConfigurationDialog dialog = createFilterConfigurationDialog(generator);
 		if (dialog.open() == Window.OK) {
 			generator.updateFilters(dialog.getFilters(), dialog.andFilters());
 		}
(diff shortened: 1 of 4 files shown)


mickaelistria:
The impact on API for this unique request is relatively high.
One can already override `createHelpControl`, so I don't think we should open the API more than it is already.

eobrienPilz:
But FiltersConfigurationDialog is internal so overriding createHelpControl is not an option.

mickaelistria:
FiltersConfigurationDialog is flagged as internal, but it's still exported and public.

eobrienPilz:
I was trying to do it in a way that would not require warning suppression for internal API usage. I want to be sure we don't have to change our implementation again if the API is removed or changed. 

Also, If FiltersConfigurationDialog is being specialized then MarkerContentGenerator needs to be made available through the MarkerSupportView as its required as an argument in the constructor. This will lead to another internal API usage warning. 

The pattern I have seen for configuring this dialog has been to do it through extension points, so I was following that. Changing the pattern may lead to more issues that I haven't thought of.

eobrienPilz:
@mickaelistria 
I had a look at an alternative implementation with less API impact.

This is what I came up with.

- make ExtendedMarkersView.openFiltersDialog method protected, allowing an overridden FiltersConfigurationDialog be provided.
- make ExtendedMarkersView.getGenerator method protected. It's required as an argument to overridden FiltersConfigurationDialog constructor.
- make TrayDialog.helpPressed method protected allowing custom help pages to be launched. 
- make TrayDialog.createHelpImageButton protected and add a style argument, allowing the style to be changed. SWT.CHECK style works for the eclipse help that appears at the side of the dialog, but SWT.PUSH is better for help that launches externally.

No changes to any extension points.

What do you think of the suggestion ? I can supply a patch or pr.

mickaelistria:
I've looked into more details at the code and there is another possible thing that might work and that would remove some of the changes you mention:
* the button is "selected" only if `TrayDialog.getTray()` returns some non-null stuff. So I suspect if you override the `openTray` method so that it does nothing, then the helpButton would remain un-toggled; achieving a bit of the result without need to change API.
* When help is pressed, a `SWT.Help` even is sent. You can most likely add a custom listener to this Help event in order to perform the right action.

As to ExtendedMarkersView, I think instead of opening several methods as API, you may better instead create a new `protected FiltersConfigurationDialog createFilterConfigurationDialog(MarkerContentGenerator generator)` method that would be called by `openFilterConfigurationDialog` so that inside this method, you could more freely perform configuration and specialization of the dialog.

eobrienPilz:
I've found something that works just by adding the new method createFilterConfigurationDialog you suggested.

	Inside overridden FiltersConfigurationDialog
	- constructor calls this.setHelpAvailable(true);
	- createHelpControl is overridden and implements its own createHelpImageButton method with a handler.

So just a single API change and a little extra code in the overridden ExtendedMarkersView to create its own button + toolbar

mickaelistria:
Great then, let's just add this method as new API, it's both simple and powerful enough.

eobrienPilz:
Will I create a new pr or just update this one ?

eobrienPilz:
@mickaelistria I updated the pr with the change we discussed. Also added a couple of tests that demonstrate how to use the new method to add a help link.

mickaelistria:
Thank you!

PR #361, opened 2022-09-27
fix MarkerPathField.compare nulls last

jukzi:
If a user sorts by a field he expects either nulls first or nulls last, but not null values unsorted.


vogella:
Please make the commit message more descriptive (Why are you doing this change / what is the benefit).

PR #364, opened 2022-09-27
Amendment to pull-request #193

ghost:
viewRegistry.find(elementId) only returns null for views, which have been disabled by an expression-based activity.
For other mechanisms, WorkbenchActivityHelper.filterItem(view) has to be used.


ghost:
See [here](https://github.com/eclipse-platform/eclipse.platform.ui/commit/5cbd0ba5df4f211d3f0785e1dd22b14b5546cf6b#commitcomment-85141792) for a more detailed discussion.

mickaelistria:
Can you please amend the commit so it has a self explanatory title?
Reference to former commits is more suitable in the commit message

merks:
@dogla Thanks for testing, reporting and providing a test case. 

@pzi-dsa Thanks for fixing so quickly in time for M1. 

@mickaelistria  It would be good to merge this for M1 because it’s a regression right?

mickaelistria:
@merks yes, we can/should merge it as soon as builds are successful; the check-freeze can be ignored today, it's pretty conservative and bugfixes are still welcome. If you have the opportunity to merge before me, feel free.

mickaelistria (approved):

mickaelistria:
Thank you!

PR #365, opened 2022-09-27
Add support for Windows on Arm64 (WoA) build

chirontt:
by adding the `win32/win32/aarch64` environment triplet and including the new fragments for WoA in the 'org.eclipse.e4.rcp' feature.

### DIFF
--- a/bundles/org.eclipse.e4.ui.swt.win32/pom.xml
+++ b/bundles/org.eclipse.e4.ui.swt.win32/pom.xml
@@ -42,6 +42,11 @@
               <ws>win32</ws>
               <arch>x86_64</arch>
             </environment>
+            <environment>
+              <os>win32</os>
+              <ws>win32</ws>
+              <arch>aarch64</arch>
+            </environment>
           </environments>
         </configuration>
       </plugin>
--- a/bundles/org.eclipse.ui.win32/pom.xml
+++ b/bundles/org.eclipse.ui.win32/pom.xml
@@ -42,6 +42,11 @@
               <ws>win32</ws>
               <arch>x86_64</arch>
             </environment>
+            <environment>
+              <os>win32</os>
+              <ws>win32</ws>
+              <arch>aarch64</arch>
+            </environment>
           </environments>
           <dependency-resolution>
             <extraRequirements>
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -436,6 +436,16 @@
          fragment="true"/>
 
    <plugin
+         id="org.eclipse.equinox.launcher.win32.win32.aarch64"
+         os="win32"
+         ws="win32"
+         arch="aarch64"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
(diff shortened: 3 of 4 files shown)


features/org.eclipse.e4.rcp/feature.xml
```diff
@@ -452,6 +462,17 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
```
akurtakov, line 465: We can not include native win32 aarch fragments in features unless we have a builder in our JIPPs allowing to rebuild them on e.g. source change in SWT. Until this happens we can only have sources that can't end up in deliverables.

akurtakov (changes requested):

features/org.eclipse.e4.rcp/feature.xml
chirontt: OK then. I'll withdraw this PR and will consider it again when the build infrastructure for Windows Arm64 is in place.

chirontt:
PR withdrawn, as build infrastructure is not yet in place to support building Windows on Arm64 artifacts.

bundles/org.eclipse.e4.ui.swt.win32/pom.xml
```diff
@@ -42,6 +42,11 @@
               <ws>win32</ws>
               <arch>x86_64</arch>
             </environment>
+            <environment>
```
akurtakov, line 45: This part can go in.

bundles/org.eclipse.ui.win32/pom.xml
```diff
@@ -42,6 +42,11 @@
               <ws>win32</ws>
               <arch>x86_64</arch>
             </environment>
+            <environment>
```
akurtakov, line 45: This part can go in.

features/org.eclipse.e4.rcp/feature.xml
```diff
@@ -435,6 +435,16 @@
          version="0.0.0"
          fragment="true"/>
 
+   <plugin
```
akurtakov, line 438: This part can't go in as it requires having binaries for launcher.

features/org.eclipse.e4.rcp/feature.xml
akurtakov: This part can't go in.

pom.xml
```diff
@@ -88,6 +88,7 @@
                     <plugin id="org.eclipse.equinox.launcher.gtk.linux.ppc64le" />
                     <plugin id="org.eclipse.equinox.launcher.gtk.linux.aarch64" />
                     <plugin id="org.eclipse.equinox.launcher.gtk.linux.x86_64" />
+                    <plugin id="org.eclipse.equinox.launcher.win32.win32.aarch64" />
```
akurtakov, line 91: I'm not 99% sure this one is fine.

vogella: I think Tycho would break if it cannot find a plug-in as part of a feature cc @laeubi

akurtakov: This is exclusion so most probbaly should work.

PR #374, opened 2022-10-01
Remove org.apache.commons.logging from o.e.e4.rcp feature

HannesWell:
If a Feature includes a Plugin, it usually includes it with a specific name and the version that was in the TP when the feature was build. This prevents consumers from using 'logging-bridges' in their TP or product, for example a slf4j-bridge that provides binary compatible packages and classes but redirects all events to slf4j.

Although not directly related to the move to slf4j from Maven-Central or slf4j-2, this is similar like https://github.com/eclipse-platform/eclipse.platform.releng/pull/118.

Besides some test classes I cannot se any code that `org.apache.commons.logging`.

@mickaelistria any objections? You added it back then.

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -186,13 +186,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.apache.commons.logging"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
          id="org.eclipse.e4.core.di.extensions"
          download-size="0"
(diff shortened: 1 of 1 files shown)


mickaelistria:
No objection. I think PDE got fixed to do proper dependency resolution since then, so this 4-years old workaround is probably not useful any more, and may even have become harmful.

HannesWell:
> No objection. I think PDE got fixed to do proper dependency resolution since then, so this 4-years old workaround is probably not useful any more, and may even have become harmful.

Great. Then lets see if all tests pass.
FYI, a similar PR is https://github.com/eclipse-platform/eclipse.platform.releng/pull/120.

Issue #376, opened 2022-10-01
jface High-DPI: URL/FileImageDescriptor: ImageFileNameProvider implementation should also test for XPATH_PATTERN


Dani-Hub:
For both `URLImageDescriptor` and `FileImageDescriptor`, the corresponding `ImageFileNameProvider` implementation is restricted to consider a name pattern of the form "@Nx" (only supporting 150% and 200% at the moment) under high-DPI situations, see [FileImageDescriptor.java#L50](https://github.com/eclipse-platform/eclipse.platform.ui/blob/ed41af6b6d85026dcc987cc1a899dba49d2c29c5/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java#L50) and [URLImageDescriptor.java#L56](https://github.com/eclipse-platform/eclipse.platform.ui/blob/a5c19746de77db4a2145cacfa5a9da30ad51a4c8/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java#L56), respectively. But for the `ImageDataProvider`, case (See [URLImageDescriptor.java#L83](https://github.com/eclipse-platform/eclipse.platform.ui/blob/a5c19746de77db4a2145cacfa5a9da30ad51a4c8/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java#L83) and the analogous [FileImageDescriptor.java#L141](https://github.com/eclipse-platform/eclipse.platform.ui/blob/a5c19746de77db4a2145cacfa5a9da30ad51a4c8/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java#L141)), an additional name regex using [XPATH_PATTERN - FileImageDescriptor.java#L45](https://github.com/eclipse-platform/eclipse.platform.ui/blob/a5c19746de77db4a2145cacfa5a9da30ad51a4c8/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java#L45) is performed. 

It is not really clear, why these `ImageFileNameProvider` implementations don't take advantage of the "NxM" name pattern as part of their search. This would be especially useful, because this pattern is supported to *every* zoom level and not restricted to 150% and 200%.

I suggest that both `ImageFileNameProvider` implementations also add the same "XPATH_PATTERN" test as second try as the existing `ImageDataProvider` situations.

laeubi:
[...]

PR #385, opened 2022-10-03
Issue #376: URL/FileImageDescriptor's ImageFileNameProvider...

Dani-Hub:
...implementation should also test for XPATH_PATTERN

- Extend FileImageDescriptor's and URLImageDescriptor's ImageFileNameProvider implementation to consider as fallback also an "NxM" name pattern.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
@@ -11,7 +11,7 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Christoph Läubrich - Bug 567898 - [JFace][HiDPI] ImageDescriptor support alternative naming scheme for high dpi
- *     Daniel Krügler - #375, #378
+ *     Daniel Krügler - #375, #378, #376
  *******************************************************************************/
 package org.eclipse.jface.resource;
 
@@ -51,6 +51,10 @@ class FileImageDescriptor extends ImageDescriptor {
 			String xName = getxName(name, zoom);
 			if (xName != null) {
 				return getFilePath(xName, zoom == 100);
+			}
+			String xPath = getxPath(name, zoom);
+			if (xPath != null) {
+				return getFilePath(xPath, zoom == 100);
 			}
 			return getFilePath(name, zoom == 100);
 		}
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
@@ -12,6 +12,8 @@
  *     IBM Corporation - initial API and implementation
  *     Patrik Suzzi <psuzzi@gmail.com> - Bug 483465
  *     Christoph Läubrich - Bug 567898 - [JFace][HiDPI] ImageDescriptor support alternative naming scheme for high dpi
+ *     Daniel Krügler - #376 - jface High-DPI: URL/FileImageDescriptor: ImageFileNameProvider implementation
+ *                             should also test for XPATH_PATTERN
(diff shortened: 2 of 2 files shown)


Dani-Hub:
Comment to the reviewers: Unfortunately I couldn't find a way to define a concrete unit test for this feature, the reason being that each affected `ImageFileNameProvider` implementation will only be considered by the `createImage(boolean, device)` method of the corresponding `ImageDescriptor`. This has the indirect effect that we cannot perform a method invocation where the caller (i.e. tester) is able to specify the actual zoom value, since that always happens indirectly within an `Image` constructor through the internal `DPUtil` class of `SWT`.

laeubi:
Just an idea, maybe one can make `ImageProvider` a package private class then make a new instance for the test?

Dani-Hub:
> Just an idea, maybe one can make `ImageProvider` a package private class then make a new instance for the test?

If such a visibility change would be OK as part of this issue, I can update my PULL request, sure. @laeubi Just to be sure that I understand your suggestion correctly: The corresponding package is `org.eclipse.jface.resource`, but the tests are hosted in different packages. Can you give me a pointer for an existing test that uses this approach to test package-private parts of jface? `FileImageDescriptor` and `URLImageDescriptor` are also package-private and I could not find a specific test for them. Maybe I misunderstand your actual proposal?

Dani-Hub:
@laeubi Could you please help me understanding your proposal? (I have edited my previous comment to make the situation clearer).

laeubi:
Good question, if the test is a **fragment** of **org.eclipse.jface** then you should be able to simply create a test that is in the package `org.eclipse.jface.resource`.

Dani-Hub:
Sure, but that would mean that I would need to introduce such a fragment bundle just for this specific test and need to ensure that it is part of the test environment, etc. That does seem to be a considerable overhead, unless I could already take advantage of this by an existing integrated fragment bundle (which I have not found yet). Furthermore I would be required to mark the jface bundle to have an extendible API (`Eclipse-ExtensibleAPI`), to get access to the new API from the outside, or would otherwise need some additional tricks to get access to the test API.

Dani-Hub:
I really would like to make progress with this PULL request. Am I understanding correctly, that the absence of testability is currently a show-stopper?
If the answer is "yes", I would like to make two mutually exclusive suggestions to make testability of the new feature possible:

- Introduce a static, "for internal use" method to class ImageDescriptor that allows to create a new `ImageFileNameProvider` from a given ImageDescriptor argument, something like:
```
	/**
	 * This method is intended for internal use only.
	 *
	 * @param descr the ImageDescriptor
	 * @return a new ImageFileNameProvider instance of the provided
	 *         {@code ImageDescriptor}, otherwise {@code null}.
	 */
	public static ImageFileNameProvider createImageFileNameProviderOf(ImageDescriptor descr) {
		if (descr instanceof FileImageDescriptor) {
			return ((FileImageDescriptor) descr).createImageFileNameProvider();
		} else if (descr instanceof URLImageDescriptor) {
			return ((URLImageDescriptor) descr).createImageFileNameProvider();
		}
		return null;
	}
```
- A similar approach could be realized by making `createImageFileNameProviderOf` non-static and return null by default and let derived classes override to construct their corresponding `ImageFileNameProvider`. This is technically cleaner, but has the disadvantage that someone who creates a sub-class of `ImageDescriptor` may believe that this is something that has to be implemented as part of the ImageDescriptor API.

Please respond to help me making progress.

Dani-Hub:
Thanks, @laeubi!

laeubi:
@Dani-Hub please suggest your improvements regarding test-ability as a new PR so we can discuss that. And sorry for the delay.

I think having a method for `ImageFileNameProvider` could be useful outside of testing as well! I must confess that the concept of the (internal) implementations of URL/FileImageDescriptor and its creation with static factory method is already a restriction where I'm not sure what we gain by this as it means a lot of workarounds and converting can happen in user code.

PR #386, opened 2022-10-03
improve Dark Theme on Windows eclipse-platform/.github#27

betamaxbandit:
Using CTabFolder and CTabItem, instead of TabFolder/TabItem, improves the look when in Dark Theme; the background of the tab no longer has a non-dark background.

This relates to multiple commits to fix:

[Windows][Dark theme] Multiple preference pages use tab widgets, that are not dark theme aware eclipse-platform/.github#27

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/ProjectLinkedResourcePage.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/ProjectLinkedResourcePage.java
@@ -19,6 +19,8 @@ import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.jface.preference.PreferencePage;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.CTabFolder;
+import org.eclipse.swt.custom.CTabItem;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.graphics.Font;
@@ -27,8 +29,6 @@ import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Label;
-import org.eclipse.swt.widgets.TabFolder;
-import org.eclipse.swt.widgets.TabItem;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPreferencePage;
 import org.eclipse.ui.IWorkbenchPropertyPage;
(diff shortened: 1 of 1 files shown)


mickaelistria:
I'm not sure this is a desirable fix in general a CTab* widgets introduce a lot of overhead and reduce accessibility. Instead, the background should be set to be inherited from parent.

betamaxbandit:
> I'm not sure this is a desirable fix in general a CTab* widgets introduce a lot of overhead and reduce accessibility. Instead, the background should be set to be inherited from parent.

Hi Mickael,
I'm not sure how to do that. If you could provide some more detailed hints, I could have a go, but the windows API is not my forte. When examining TabFolder I did notice this comment (inside org.eclipse.swt.widgets.TabFolder.findThemeControl()), which put me off further investigation:

`	/* It is not possible to change the background of this control */`

For now, I'm following the "band aid" solution, as in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/126, to make some quick improvements to the dark theme look.

mickaelistria:
In the current state of things, I believe a "band aid" solution is likely to become a bigger issue on the long run than the bug. I don't know about Windows specifics, but I believe that on other OS it's possible to change the tab background.
@akurtakov what do you think about it?

laeubi:
I think we have simply competing requirements here.

1. Theming (what supports something we claim is a "dark mode")
2. Using native controls so an SWT app looks like it is a native application

So for the first, I think one will sooner or later need custom drawn components everywhere, for second one might argue that the OS itself has to provide the desired theming (what could be a dark system theme) and everything will work "like magic" so the application itself should never provide any theming at all if it wants to looks "native".

jonahgraham:
The decision has been accepted previously and @betamaxbandit has done this work based on the acceptance of it. Please see @akurtakov's comment in https://github.com/eclipse-platform/.github/issues/27#issuecomment-1204057052

If there are discussions about the overall strategy, lets discuss them there.

akurtakov:
It's bandaiding for sure but looks like we don't have much choice as win32 port is kind of without anyone to work on it.

mickaelistria:
I let you merge according to your decision. And will followup on the issue pointed by @jonahgraham

betamaxbandit:
To Test:

1. Launch an eclipse workbench.
2. Create a Java project and provide a project name, use defaults for rest of settings and click Finish.
3. In Preferences > General > Appearance > Theme, choose Dark and click Apply and Close. Restart the workbench when prompted.
4. In Package Explorer, select the project and choose Project > Properties.
5. In the left hand side navigator, select Resource > Linked Resource:
**Expected**: the Linked Resources page is displayed and tabs (eg: Path Variables, Linked Resourcs) are displayed correctly. The background of the tabs is shown using the Dark Theme colour.
6. Repeat from step 3) and change the Theme to Light.
**Expected**: the Linked Resources page is displayed and tabs (eg: Path Variables, Linked Resourcs) are displayed correctly. The background of the tabs is shown using the Light Theme colour.

**Before and After screencaptures:**
Project properties - Resource - Linked Resources - dark theme - before

Project properties - Resource - Linked Resources - dark theme - afterPNG

Project properties - Resource - Linked Resources - light theme - before

Project properties - Resource - Linked Resources - light theme - afterPNG

PR #387, opened 2022-10-03
Add static UIJob-factories similiar to Job factories

HannesWell:
The Job class has convenient factory methods `create()` and `createSystem()` to create a job that executes a given `ICoreRunnable` or `IJobFunction`.
This PR proposes to add similar factories for the subclass `UIJob` named `UIJob.createUI()`. The name `createUI` was chosen to not clash with the super-class.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/progress/UIJob.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/progress/UIJob.java
@@ -10,13 +10,18 @@
  *
  * Contributors:
  *     IBM - Initial API and implementation
+ *     Hannes Wellmann - Add static factories and handle cancel by OperationCanceledException
  *******************************************************************************/
 package org.eclipse.ui.progress;
 
 import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ICoreRunnable;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.IJobFunction;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.swt.SWTException;
 import org.eclipse.swt.widgets.Display;
@@ -31,6 +36,53 @@ import org.eclipse.ui.internal.progress.ProgressMessages;
  * @since 3.0
  */
 public abstract class UIJob extends Job {
+
+	/**
+	 * Creates a new UIJob that will execute the provided function in the UI thread
+	 * when it runs.
+	 *
+	 * Prefer using {@link UIJob#create(String, ICoreRunnable)}.
+	 *
+	 * @param name     The name of the job
+	 * @param function The function to execute
+	 * @return A UIJob that encapsulates the provided function
+	 * @see IJobFunction
+	 * @since 3.127
+	 */
+	public static UIJob create(String name, IJobFunction function) {
+		return new UIJob(name) {
+			@Override
+			public IStatus runInUIThread(IProgressMonitor monitor) {
+				return function.run(monitor);
+			}
+		};
+	}
+
+	/**
+	 * Creates a new UIJob that will execute the provided runnable in the UI thread
+	 * when it runs.
+	 *
+	 * @param name     the name of the job
+	 * @param runnable the runnable to execute
+	 * @return a UIJob that encapsulates the provided runnable
+	 * @see ICoreRunnable
+	 * @since 3.127
+	 */
+	public static UIJob create(String name, ICoreRunnable runnable) {
+		return new UIJob(name) {
+			@Override
+			public IStatus runInUIThread(IProgressMonitor monitor) {
+				try {
+					runnable.run(monitor);
+				} catch (CoreException e) {
+					IStatus st = e.getStatus();
+					return new Status(st.getSeverity(), st.getPlugin(), st.getCode(), st.getMessage(), e);
+				}
+				return Status.OK_STATUS;
+			}
+		};
+	}
+
 	private Display cachedDisplay;
 
 	/**
@@ -95,14 +147,14 @@ public abstract class UIJob extends Job {
 					UIStats.start(UIStats.UI_JOB, getName());
 					result = runInUIThread(monitor);
 				}
-
+			} catch (OperationCanceledException e) {
+				result = Status.CANCEL_STATUS;
 			} catch (Throwable t) {
 				throwable = t;
 			} finally {
 				UIStats.end(UIStats.UI_JOB, UIJob.this, getName());
 				if (result == null) {
-					result = new Status(IStatus.ERROR, PlatformUI.PLUGIN_ID, IStatus.ERROR,
-							ProgressMessages.InternalError, throwable);
+					result = Status.error(ProgressMessages.InternalError, throwable);
 				}
 				done(result);
 			}
--- a/bundles/org.eclipse.e4.ui.progress/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.progress/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.progress;singleton:=true
-Bundle-Version: 0.3.500.qualifier
+Bundle-Version: 0.3.600.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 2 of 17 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/progress/UIJob.java
```diff
@@ -31,6 +34,53 @@
  * @since 3.0
  */
 public abstract class UIJob extends Job {
+
+	/**
+	 * Creates a new UIJob that will execute the provided function in the UI thread
+	 * when it runs.
+	 *
+	 * Prefer using {@link UIJob#createUI(String, ICoreRunnable)}.
```
HannesWell: The Job class recommends to prefer the ICoreRunnable overload, but the mentioned reason is not valid anymore because both get monitor.done() called and can use `OperationCanceledException`.

Therefore I wonder if the recommendation is still required or if just the `ICoreRunnable` should be added?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/progress/UIJob.java
```diff
@@ -31,6 +34,53 @@
...
+	 *
+	 * @param name     The name of the job
+	 * @param function The function to execute
+	 * @return A UIJob that encapsulates the provided function
+	 * @see IJobFunction
+	 * @since 3.127
+	 */
+	public static UIJob createUI(String name, IJobFunction function) {
```
iloveeclipse: why "createUI" and not simply "create"? We don't create widgets here :)

HannesWell: > why "createUI" and not simply "create"? We don't create widgets here :)

That's right. 😅 The super-class Job has `create()` methods that would clash with this class.
For example if one calls `UIJob.create()` now (without this change) a Job is returned. If we add just a create() method that would now return a `UIJob`. A UIJob is still a Job, so that would not be a compiler-error, but it would suddenly execute in the UI-thread, instead of a worker thread.
One can probably assume that that was the actual intention, nevertheless it would be a behavioral change.

The name createUI is indeed misleading and another would be better (maybe `createUIJob()`?), but just `create()` could cause problems.

laeubi: I don't think you can override static methods anyways...

merks: Indeed you can't, but you can hide the static base class methods by defining a method with the same name in the subclass.  Note however that if one did write UIJob.create that there would be a warning `The static method create(String, ICoreRunnable) from the type Job should be accessed directly` unless one turns off such a warning.   I suppose it's possible that some consumer ignored that warning or turned it off, but that seems somewhat unlikely.

HannesWell: Yes, that's what I meant and yes precisely it is is not a override.

The question to me is if we want to take that risk or not? As you said there is a warning, so one could argue that it is the users fault if that warning is ignored.
And yes, just `UIJob.create()` would definitely be nicer.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/progress/UIJob.java
HannesWell: Since we have a factory with `IJobFunction` and `ICoreRunnable`, the method calls is ambiguous for non-block lambda expressions, which is a bit ugly because then one has to cast the expression to `IJobFunction` or `ICoreRunnable`.
If I only add a `ICoreRunnable` that problem wouldn't occur.
Correct me if I'm wrong, but I have the impression that `ICoreRunnable` is the more modern approach and the `IJobFunction`  factory is only in the Job class for backwards compatibility and one can achieve the same regarding error-propagation and cancellation with an ICoreRunnable. Therefore I'm tending more towards just adding the ICoreRunnable factory.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/progress/UIJob.java
merks: I think anyone writing UIJob.create would expect it to create a UIJob and although it doesn't currently do that, changing it now is at worst a source incompatibility not a binary incompatibility.  If someone was affected by the change, they could use Job.create instead as they ought to have done in the first place.  That would still be binary compatible with their intent in both an old Eclipse and a newer one.  So personally I'd be tempted to do what's nicer as an API...

HannesWell: > I think anyone writing UIJob.create would expect it to create a UIJob and although it doesn't currently do that, changing it now is at worst a source incompatibility not a binary incompatibility. 

I think the same. And it is not even a source incompatibility because before UIJob.create() returned a Job and now it returns an `UIJob`, which is still a job. So unless somebody notices less warnings the only (maybe) noticeable effect is that the Job now runs in the UI-thread. That was probably intended in the first place, however it still changed now.

I'm fine with that change. From my experience I just noticed that the Eclipse-community (or maybe consumers in general) is sensitive to changes. Therefore I wanted to raise awareness about this point and if other devs with more experience, like you, say it is OK, I'm fine with that as well. :)

HannesWell:
I also noticed there is another, very similar, `UIJob` class in the package `org.eclipse.e4.ui.progress`. I suspect this is because the `org.eclipse.e4.ui.progress` Plug-in should not have a dependency to o.e.workbench.ui?

merks:
@HannesWell Yes, I believe you are correct about `org.eclipse.e4.ui.progress` being needed for pure e4 applications that do not have workbench dependencies.

HannesWell:
> @HannesWell Yes, I believe you are correct about `org.eclipse.e4.ui.progress` being needed for pure e4 applications that do not have workbench dependencies.

I added the same factories for that UIJob class.
Furthermore I adjusted their run() methods to handle cancellation by throwing `OperationCanceledException` correctly. At least from following the code paths a `OperationCanceledException` would simply lead to an result with error status that has that exception as cause.

HannesWell:
Actually I would prefer to have only one factory that accepts a `ICoreRunnable` and ignore the one with `IJobFunction`, but I suspect that this will lead to confusion if only one of two factories from the Job super-class is shadowed. So as long as the Job class has both factory-methods the UIJob should also have both.

@merks, @iloveeclipse are you fine with this change as it is?

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/LinkEditorAction.java
```diff
@@ -58,61 +55,49 @@ public class LinkEditorAction extends Action implements ISelectionChangedListene
...
-							ignoreEditorActivation = true;
-							SafeRunner.run(new NavigatorSafeRunnable() {
-								@Override
-								public void run() throws Exception {
-									helpers[0].activateEditor(commonNavigator.getSite().getPage(), selection);
-								}
-							});
-							ignoreEditorActivation = false;
+	private UIJob activateEditorJob = UIJob.create(CommonNavigatorMessages.Link_With_Editor_Job_, m -> {
+		LinkEditorAction thisAction = this;
```
iloveeclipse: `LinkEditorAction thisAction = this;` looks like a hack to avoid compiler error if it wouldn't be a hack.

I would prefer not to refactor **this** code or do it in a way that avoids "hacking"

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/LinkEditorAction.java
```diff
@@ -58,61 +55,49 @@ public class LinkEditorAction extends Action implements ISelectionChangedListene
...
-								IStructuredSelection newSelection = linkService.getSelectionFor(input);
-								if (!newSelection.isEmpty()) {
-									ignoreSelectionChanged = true;
-									commonNavigator.selectReveal(newSelection);
-									ignoreSelectionChanged = false;
-								}
+	});
+
+	private UIJob updateSelectionJob = UIJob.create(CommonNavigatorMessages.Link_With_Editor_Job_, m -> {
+		LinkEditorAction thisAction = this;
```
iloveeclipse: Same here

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/LinkEditorAction.java
HannesWell: That's right, without this there is a compiler-error that the fields used below are not yet initialized.
I assume that this is because the lambda is then bound to the fields instead of `this`.
I don't know a better way at the moment and reverted that refactoring for now. If somebody else knows a better way it can be applied in the future.

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/internal/navigator/actions/LinkEditorAction.java
HannesWell: Reverted, as above.

HannesWell:
Does anybody have more remarks/objections? If not I think this is good to go.

I could then also add a N&N entry for this new API.

iloveeclipse (approved):
LGTM. Please rebase before merge, just in case.

HannesWell:
> LGTM. Please rebase before merge, just in case.

Done.

Thank you all.

PR #78, opened 2022-05-18
Bug 546115: show editor guiding text

marcushoepfner:
First draft of showing some guiding text in the editor area.

Open: where to put the actual code, at the moment it is in StackRenderer.
Also at the moment there is a leak with on of the images. But let's first discuss whether it should be done like this or the whether the image and text should come from a new extension point.

PR #394, opened 2022-10-06
Send out an event during perspective switch

vogella:
Currrently the system does not send out and event if the part service is used to switch a perspective. Fix this by adding this event and sending it out.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.workbench/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.workbench;singleton:=true
-Bundle-Version: 1.13.300.qualifier
+Bundle-Version: 1.14.0.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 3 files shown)


vogella:
@marcushoepfner ok for you to pull this out? I think this event is useful even without guiding text.

marcushoepfner (approved):

marcushoepfner:
Thanks Lars. I don't remember if it worked in any case (e.g. switcher in toolbar, switching via Window->Perspective->Open Perspective, ...).
I did so many stuff on my PR #78, but something in my brain tells me there is was an issue with Perspective Switching event in my PR. But maybe I mix it up completely and at least w.r.t. to persp. switching events everything was fine on my PR .

Sorry that I cannot recall this correctly.

jukzi:
isn't adding an event an API change even though it is binary compatible?

laeubi (approved):
Seems like a useful addition.

laeubi:
> isn't adding an event an API change even though it is binary compatible?

Can you explain what might be an API concern here? While _removing_ an event could cause issues, _adding_ one seems "save" ...

jukzi:
- A client could fail at runtime if an unknown event is detected.
- A client that needs that event should be able to request the minor version needed to get it

laeubi:
> * A client could fail at runtime if an unknown event is detected.

This would be a very bad written client ... 

> * A client that needs that event should be able to request the minor version needed to get it

So you actually want to suggest to increase the minor version? @vogella can you make sure the minor version is increased in this release?

jukzi:
> So you actually want to suggest to increase the minor version?

exactly. everything else looks fine.

marcushoepfner:
Nice! I'll rebase #78. Your PR make #78 less complex. Thanks

Issue #396, opened 2022-10-08
[jface] Certain ImageDescriptor classes should be adaptable for some internal properties


Dani-Hub:
This issue is a follow-up of the suggestion by @laeubi in [issue comment-1270353935](https://github.com/eclipse-platform/eclipse.platform.ui/pull/385#issuecomment-1270353935) to open a new issue that suggests increasing the testability of certain ImageDescriptor subclass properties. I'm starting with two motivating use-cases, which are completely orthogonal but could be solved by the same approach:

1. In method [MenuHelper#getIconURI](https://github.com/eclipse-platform/eclipse.platform.ui/blob/849e023520392865995477474ef731f5302edbda/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/menus/MenuHelper.java#L463) it is attempted to retrieve an URL (or URI) from a given `ImageDescriptor` subclass. The first three parts of that attempt consist of reflection-based techniques, followed by a the "adaptable pattern" to realize this. It turns out that the very general very first step (which assumes that classes whose name ends with "URLImageDescriptor" do have a special meaning) caused problems in my companies code base, because we also provide a `ImageDescriptor` that falls into this naming pattern but where this approach fails (I cannot go into the details here). In fact, it would be easily possible for us to make our class adaptable to fit that purpose of this method, but since the name pattern happens first and no other case is tested, we would be required to rename our class, which is hard to justify.
2. In PULL request #385 I had provided an improvement of the internal `ImageFileNameProvider` implementations of `URLImageDescriptor` and `FileImageDescriptor`, but it turned out that the current implementation state made it extremely hard to actually test the correctness of those `ImageFileNameProvider` improvements, since there is no direct or indirect way to call them under controlled zoom values.

This issue suggests to let certain package-private `ImageDescriptor` classes (such as `FileImageDescriptor` and `URLImageDescriptor`) implement the `org.eclipse.core.runtime.IAdaptable` interface to allow to retrieve certain otherwise inaccessible properties for _internal_ reasons such as testability (See bullet 2 above) or to convert to an URL/URI in (See bullet 1 above) as alternative to reflection-based approaches. Note that this issue intentionally does *not* suggest to let the public, abstract class `ImageDescriptor` implement that interface because this issue doesn't intent to make this generic query API part of the public API. Of-course this issue doesn't prevent a future decision to support this case, but that would be a different request.

[...]

PR #397, opened 2022-10-08
Issue #396: Certain ImageDescriptor classes should be adaptable for...

Dani-Hub:
...some internal properties

- Let FileImageDescriptor and URLImageDescriptor implement IAdaptable and provide adaption to ImageFileNameProvider (for testing purposes)
- Add unit tests that attempt to retrieve the underlying ImageFileNameProvider and validate the expected xPath results.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2020 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -11,7 +11,7 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Christoph Läubrich - Bug 567898 - [JFace][HiDPI] ImageDescriptor support alternative naming scheme for high dpi
- *     Daniel Krügler - #375, #378, #376
+ *     Daniel Krügler - #375, #376, #378, #396
  *******************************************************************************/
 package org.eclipse.jface.resource;
 
@@ -26,6 +26,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.eclipse.core.runtime.FileLocator;
+import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Status;
@@ -41,7 +42,7 @@ import org.eclipse.swt.graphics.ImageFileNameProvider;
 /**
  * An image descriptor that loads its image information from a file.
  */
-class FileImageDescriptor extends ImageDescriptor {
+class FileImageDescriptor extends ImageDescriptor implements IAdaptable {
 
 	private static final Pattern XPATH_PATTERN = Pattern.compile("(\\d+)x(\\d+)"); //$NON-NLS-1$
 
@@ -294,5 +295,14 @@ class FileImageDescriptor extends ImageDescriptor {
 			}
 			return null;
 		}
+	}
+
+	@Override
+	public <T> T getAdapter(Class<T> adapter) {
+		if (adapter == ImageFileNameProvider.class) {
+			// Support testing ImageFileNameProvider characteristics, see #396
+			return adapter.cast(new ImageProvider());
+		}
+		return null;
 	}
 }
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
@@ -49,15 +49,25 @@ class FileImageDescriptor extends ImageDescriptor implements IAdaptable {
 	private class ImageProvider implements ImageFileNameProvider {
 		@Override
 		public String getImagePath(int zoom) {
+			final boolean logIOException = zoom == 100;
+			if (zoom == 100) {
+				return getFilePath(name, logIOException);
+			}
 			String xName = getxName(name, zoom);
 			if (xName != null) {
-				return getFilePath(xName, zoom == 100);
+				String xResult = getFilePath(xName, logIOException);
+				if (xResult != null) {
+					return xResult;
+				}
 			}
 			String xPath = getxPath(name, zoom);
 			if (xPath != null) {
-				return getFilePath(xPath, zoom == 100);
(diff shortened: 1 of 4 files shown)


Dani-Hub:
@laeubi Can you please help me understanding the cause of the test failures? I have problems determining the actual error, albeit it seems to be related to my PULL request, but to which extend?

laeubi:
It seems an assertion failed:

https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-397/1/testReport/junit/org.eclipse.jface.tests.images/FileImageDescriptorTest/testImageFileNameProviderGetxPath/

I would recommend to enhance the test, e.g. with an meaningful message at the assertion.

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/images/FileImageDescriptorTest.java
```diff
@@ -174,4 +176,22 @@ public void testGetxPath150() {
 		assertEquals(Math.round(imageData.height * 1.5), imageDataZoomed.height);
 	}
 
+	public void testImageFileNameProviderGetxPath() {
+		ImageDescriptor descriptor = ImageDescriptor.createFromFile(FileImageDescriptorTest.class,
+				"/icons/imagetests/rectangular-57x16.png");
+		assertTrue("FileImageDescriptor does not implement IAdaptable", descriptor instanceof IAdaptable);
+		IAdaptable adaptable = (IAdaptable) descriptor;
+		ImageFileNameProvider fileNameProvider = adaptable.getAdapter(ImageFileNameProvider.class);
```
laeubi, line 184: Usually the pattern is to use:

Adapters.getAdapter

instead of checking for adaptable directly.

Dani-Hub: I agree that this is what should be done in production code, but this unit test particularly wants to tests the IAdaptable inheritance.

Dani-Hub:
> It seems an assertion failed:
> 
> https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-397/1/testReport/junit/org.eclipse.jface.tests.images/FileImageDescriptorTest/testImageFileNameProviderGetxPath/

Thanks!

> I would recommend to enhance the test, e.g. with an meaningful message at the assertion.

Well, I just copied the existing style of the other tests ;-)

OK, I found the problem and it was actually caused by an error of my previous #376 implementation - and this new tests demonstrates the importance of such a test - thanks for reminding to make this testable!

As a matter of fact, it also turns out, that the _original_ implementation also didn't work correctly, so I will now fix both the original as well as my own extension.

laeubi:
> Well, I just copied the existing style of the other tests ;-)

Don't do that! :-D

Actually it is always good to have meaningful assertion with a message that tells what has failed for "simple" asserts like numbers or boolean as its often hard to guess what "expected 2 but got 3" means or "assertion failed" like here.

Dani-Hub:
@laeubi All tests pass now, how do you like it?

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
```diff
@@ -295,4 +306,13 @@ String getFilePath(String name, boolean logIOException) {
 			return null;
 		}
 	}
+
+	@Override
+	public <T> T getAdapter(Class<T> adapter) {
+		if (adapter == ImageFileNameProvider.class) {
+			// Support testing ImageFileNameProvider characteristics, see #396
+			return adapter.cast(new ImageProvider());
```
laeubi, line 314: Could an image provide maybe only instantiated once?

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -344,4 +351,13 @@ private static URL getURL(String urlString) {
 		return result;
 	}
 
+	@Override
+	public <T> T getAdapter(Class<T> adapter) {
+		if (adapter == ImageFileNameProvider.class) {
+			// Support testing ImageFileNameProvider characteristics, see #396
+			return adapter.cast(new URLImageFileNameProvider(url));
```
laeubi, line 358: Same here, why always a new instance?

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/images/FileImageDescriptorTest.java
laeubi: Does it really matter "how" it is adapted? I think the important part is *that* it could be adapted....

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
Dani-Hub: Well, it could, but that's not how the FileImageDescriptor class is currently implemented. You should notice that it always generates a fresh new ImageProvider when it's required. And note also that the sole purpose is (currently) for testing. I would suggest that we should create a separate issue if we want to have only a single ImageFileNameProvider instance within the ImageDescriptor, because that would be a specific optimization approach, unrelated to the new adapter code. You can count on me to volunteer for the issue and taking care of a PULL request.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
Dani-Hub: See above

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/images/FileImageDescriptorTest.java
Dani-Hub: Well, it does not matter to me, certainly. My test was constructed based on the use case in [MenuHelper.java#L491](https://github.com/eclipse-platform/eclipse.platform.ui/blob/849e023520392865995477474ef731f5302edbda/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/menus/MenuHelper.java#L491) where also such a concrete tests exists. I have no strong feelings about this aspect. If replacing the concrete test for IAdaptable inheritance and its usage by a call to `org.eclipse.core.runtime.Adapters.adapt(Object sourceObject, Class<T> adapter)` or `org.eclipse.core.runtime.Adapters.of(Object sourceObject, Class<T> adapter)` would increase consensus, let me know. @laeubi please respond.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
Dani-Hub: A corresponding issue has been created, see #398.

Issue #398, opened 2022-10-08
[jface] FileImageDescriptor and URLImageDescriptor should hold only a single ImageFileNameProvider instance


Dani-Hub:
During the review of PULL request #397 it has been asked by @laeubi (See [discussion r990674616](https://github.com/eclipse-platform/eclipse.platform.ui/pull/397#discussion_r990674616)) whether `FileImageDescriptor` and `URLImageDescriptor` (See [discussion r990674616](discussion_r990674616)) shouldn't hold a single instance of its internally used `ImageFileNameProvider`. This was intentionally *not* done by this issue, because those classes always generate a fresh `ImageFileNameProvider` when needed, so this is a pre-existing issue unrelated to #397. 

This issue suggests to optimize both `FileImageDescriptor` and `URLImageDescriptor` to hold only one instance of `ImageFileNameProvider`.

Note that a similar request could be suggested for the underlying `ImageDataProvider` of `URLImageDescriptor`. That's a fair question, but I would suggest to take care of this by an independent issue (I will take care of it), because the underlying `ImageDataProvider` cannot be queried and tested properly anyway.

Dani-Hub:
Issue #399 has been created to take care of a similar optimization possibility for the underlying ImageDataProvider of URLImageDescriptor.

Dani-Hub:
I think we can close this issue as being resolved by #400.

laeubi:
@Dani-Hub you should be able to close issues you have opend yourself. Beside that, if you add `Fix #<issue number>` in your commit message an issue gets automatically closed when it was merged to master.

Dani-Hub:
> @Dani-Hub you should be able to close issues you have opend yourself. 

Great, didn't know that.
>Beside that, if you add `Fix #<issue number>` in your commit message an issue gets automatically closed when it was merged to master.

Ah, that's the trick. I'll try to memorize for future contributions.

Issue #399, opened 2022-10-08
[jface] Improve URLImageDescriptor and its ImageDataProvider


Dani-Hub:
During my work on #397 I observed the following opportunities of improvement in `URLImageDescriptor` in relation to its underlying `ImageDataProvider`:

1. It is currently not possible to test the correct zoom behaviour because no access to the `ImageDataProvider` currently exists. For the underlying `ImageFileNameProvider` it was found by making it testable (See [issuecomment 1272352893](https://github.com/eclipse-platform/eclipse.platform.ui/pull/397#issuecomment-1272352893)) that the previous implementation was defect. We should take advantage of the new `IAdaptable` implementation of `URLImageDescriptor` by #397 to also support a query for its underlying `ImageDataProvider`.
2. In Issue #398 it has been observed that we should optimize the underlying `ImageFileNameProvider` to be represented by a single instance. The same approach could be done for the `ImageDataProvider` as well.
3. During my work fixing the `ImageFileNameProvider` I noticed that also the `ImageDataProvider` could be optimized for the 100% zoom case as well. We should do that here as well to prevent unnecessary tests for xPath and xName in that situation.

Dani-Hub:
I would like to withdraw the suggestion in bullet (1) above (but not the others), for the reason that the underlying `ImageDataProvider` of `URLImageDescriptor` is _indirectly_ testable through the public `getImageData(int zoom)` and this method *is* currently used to test the correctness depending on specific zoom values. Nonetheless both bullet (2) and (3) still apply and I'm planning my upcoming PULL requests on those two aspects.

Dani-Hub:
I think we can close this issue now, resolved by #400.

akurtakov:
As per previous comment.

PR #400, opened 2022-10-09
Issue #398: FileImageDescriptor and URLImageDescriptor should hold...

Dani-Hub:
... only a single ImageFileNameProvider instance

- Introduce a lazily generated ImageFileNameProvider in FileImageDescriptor and URLImageDescriptor.
- Make the pre-existing fields of FileImageDescriptor and URLImageDescriptor final as an additional guarantee that these are effectively immutable, justifying the usage of an only once generated ImageFileNameProvider.
- Extend existing tests for FileImageDescriptor and URLImageDescriptor that getAdapter returns a unique instance of ImageFileNameProvider.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
@@ -42,34 +42,9 @@ import org.eclipse.swt.graphics.ImageFileNameProvider;
 /**
  * An image descriptor that loads its image information from a file.
  */
-class FileImageDescriptor extends ImageDescriptor implements IAdaptable {
+class FileImageDescriptor extends ImageDescriptor implements IAdaptable, ImageFileNameProvider {
 
 	private static final Pattern XPATH_PATTERN = Pattern.compile("(\\d+)x(\\d+)"); //$NON-NLS-1$
-
-	private class ImageProvider implements ImageFileNameProvider {
-		@Override
-		public String getImagePath(int zoom) {
-			final boolean logIOException = zoom == 100;
-			if (zoom == 100) {
-				return getFilePath(name, logIOException);
-			}
-			String xName = getxName(name, zoom);
-			if (xName != null) {
-				String xResult = getFilePath(xName, logIOException);
-				if (xResult != null) {
-					return xResult;
-				}
-			}
-			String xPath = getxPath(name, zoom);
-			if (xPath != null) {
-				String xResult = getFilePath(xPath, logIOException);
-				if (xResult != null) {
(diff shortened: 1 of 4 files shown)


Dani-Hub:
@laeubi Could you please take a look?

laeubi:
@Dani-Hub I have the strong-feeling that `FileImageDescriptor` can simply implements `ImageFileNameProvider`, same for `URLImageDescriptor`, why having separate inner classes at all?

Dani-Hub:
I replaced only the inner `ImageFileNameProvider` classes as part of this issue. We can replace the inner `ImageDataProvider`  of `URLImageDescriptor` as part of my additionally planned improvements as part of issue #399.

Dani-Hub:
I also decided to follow your earlier advice to use the Adapters.adapt approach of directly testing the `IAdaptable` inheritance. I agree that this makes the test more robust against a possible future refactoring (and a little bit shorter as well).

laeubi:
I think you can now even omit the `IAdaptable` interface at all as `Adapters.adapt` already do instanceof check!

Dani-Hub:
The reason for the `IAdaptable` pattern has been introduced to solve the larger picture issue where we want to adapt to URL. Once we have implemented that, we can remove the adaption to the ImageFileNameProvider, I agree. And in fact that's the reason why I switched to Adapters.adapt right *now*.

laeubi:
> has been introduced to solve the larger picture issue where we want to adapt to URL. Once we have implemented that, we can remove the adaption to the ImageFileNameProvider, I agree.

Can you then add the code to adapt to `File` or `URL` object here? `return adapter.cast(this);` is not required anymore if the class itself implement the interface. if you still want to keep this maybe 

```
if (adapter.isInstance(this)) {
	// Support testing ImageFileNameProvider characteristics, see #396
	return adapter.cast(this);
}
```

might be more generic

Dani-Hub:
> > has been introduced to solve the larger picture issue where we want to adapt to URL. Once we have implemented that, we can remove the adaption to the ImageFileNameProvider, I agree.
> 
> Can you then add the code to adapt to `File` or `URL` object here? 

Not as part of this issue, please. I have already created a separate issue for the URL adaption. I don't plan to adapt to File right now. If you plan to do that, I suggest a separate issue (I don't think that this is necessary)

> 
> ```
> if (adapter.isInstance(this)) {
> 	// Support testing ImageFileNameProvider characteristics, see #396
> 	return adapter.cast(this);
> }
> ```
 
If we strike the adaption to ImageFileNameProvider later, this if case will get removed anyway. I would therefore prefer to keep it until that point.

Dani-Hub:
@laeubi Tests completed.

PR #402, opened 2022-10-10
Issue #399: Improve URLImageDescriptor and its ImageDataProvider

Dani-Hub:
- Replace inner ImageDataProvider by inheritance, so that we get a single instance for free
- Optimize ImageDataProvider#getImageData for the 100% case (similar to ImageFileNameProvider#getImagePath)


Dani-Hub:
@laeubi There is a third possible optimization that I could think of for this class and that related to the internal representation of the hold URL as String. I *think* that the original intention was to hold a String instead of the provided URL directly, to ensure that `equals` and `hashCode` are efficient. Now I'm wondering: Maybe we could store both URL *and* its `toExternalForm()` String as members, this would most (but not all) calls of `org.eclipse.jface.resource.URLImageDescriptor.getURL(String urlString)` obsolete and would still provide the efficiency of equals and hashCode. What do you think?

Issue #401, opened 2022-10-09
MenuHelper#getIconURI should be improved


Dani-Hub:
Method [MenuHelper#getIconURI](https://github.com/eclipse-platform/eclipse.platform.ui/blob/63b65f971ccc8b165774365d1dfc10e8888059f5/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/menus/MenuHelper.java#L456)'s implementation is not only rather complex, but also contains some pitfalls that could easily be prevented.

Basically the method can be divided in two blocks: (a) The first three cases perform reflection using a quite general naming heuristics to derive a URL and (b) the last two cases can be summarized as applying the "adapter pattern" to the provided `ImageDescriptor` instance to either determine an URL or URI.

The list of problems are as follows:

1. The naming pattern in (a) is very general and is causing problems in our company's code base, because we also provide an `XXURLImageDescriptor` class (Consider `XX` as a placeholder for a name prefix whose concrete value is irrelevant here) where the reflection approach doesn't work. We could easily adapt our class to either URL or URI, but that doesn't help us, because the algorithm stops when the name pattern matches. We _could_ rename our class, but that would really be a pity for us. We consider reserving a general name suffix such as "URLImageDescriptor" for jface solely as being rather user unfriendly. 
2. A second more subtle problem with the reflection block (a) is that it caches the "url" field of the first class whose name ends with "URLImageDescriptor", so once a second class follows with the same suffix, it would try to apply the same reflection field on a different class, which is doomed to fail (The same problem exists for the other two reflection cases).
3. The adapter block (b) is sub-divided into two branches, where a failure in one branch does not attempt to try the second branch. This is a pity, because that could easily be solved by using just `org.eclipse.core.runtime.Adapters.adapt(Object sourceObject, Class<T> adapter)` instead, which evaluates both the `IAdaptable` interface technique as well as the `IAdapterManager` approach one after the other transparently.
4. A second issue with block (b) is the usage of `IAdapterManager#getAdapter` instead of `IAdapterManager#loadAdapter`, because this method is called in low-level parts where the user has usually no control on whether its adapter plugin is already activated or not. 
[...]

PR #403, opened 2022-10-11
Issue #401: MenuHelper#getIconURI should be improved

Dani-Hub:
- Make DeferredImageDescriptor, FileImageDescriptor, and URLImageDescriptor adaptable to URL
- For FileImageDescriptor and URLImageDescriptor remove explicit adaption to ImageFileNameProvider, since Adapters.adapt will also find the inherited one
- Add new ImageDescriptor tests that validate URL adaption
- Refactor MenuHelper#getIconURI by eliminating the reflection-based part and by replacing the IAdaptable+IAdapterManager pattern by Adapters.adapt

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
@@ -38,15 +38,18 @@ import org.eclipse.swt.graphics.ImageData;
  * @since 3.21
  */
 final class DeferredImageDescriptor extends ImageDescriptor implements IAdaptable {
+
 	/**
-	 * The supplier of the class. Note that there is a non-code reference to this
-	 * field in
-	 * <code>org.eclipse.ui.internal.menus.MenuHelper.getUrlSupplier()</code> so if
-	 * this field is renamed or changed then the code above needs to be repaired as
-	 * well.
+	 * The supplier of the class.
 	 */
 	private final Supplier<URL> supplier;
+
 	private final boolean useMissingImage;
+
+	/**
+	 * URL referring to the actual image, computed lazily when required.
+	 */
+	private URL url;
 
 	/**
 	 * Create a new DeferredImageDescriptor with the given URL supplier.
@@ -66,7 +69,7 @@ final class DeferredImageDescriptor extends ImageDescriptor implements IAdaptabl
 
 	@Override
 	public ImageData getImageData(int zoom) {
-		URL url = supplier.get();
+		URL url = getURL();
 		if (url == null) {
 			return useMissingImage ? ImageDescriptor.getMissingImageDescriptor().getImageData(zoom) : null;
 		}
@@ -75,17 +78,24 @@ final class DeferredImageDescriptor extends ImageDescriptor implements IAdaptabl
 
 	@Override
 	public Image createImage(boolean returnMissingImageOnError, Device device) {
-		URL url = supplier.get();
+		URL url = getURL();
 		if (url == null) {
 			return returnMissingImageOnError ? ImageDescriptor.getMissingImageDescriptor().createImage() : null;
 		}
 		return ImageDescriptor.createFromURL(url).createImage(returnMissingImageOnError, device);
 	}
 
+	private final URL getURL() {
+		if (url == null) {
+			url = supplier.get();
+		}
+		return url;
+	}
+
 	@Override
 	public <T> T getAdapter(Class<T> adapter) {
 		if (adapter == URL.class) {
-			return adapter.cast(supplier.get());
+			return adapter.cast(getURL());
 		}
 		return null;
 	}
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, Alex Blewitt and others.
+ * Copyright (c) 2020, 2022, Alex Blewitt and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     Alex Blewitt - initial API and implementation
+ *     Daniel Kruegler - #396, #401
  *******************************************************************************/
 package org.eclipse.jface.resource;
 
@@ -17,6 +18,7 @@ import java.net.URL;
 import java.util.Objects;
 import java.util.function.Supplier;
 
+import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.swt.graphics.Device;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
@@ -35,7 +37,7 @@ import org.eclipse.swt.graphics.ImageData;
  *
  * @since 3.21
  */
-final class DeferredImageDescriptor extends ImageDescriptor {
+final class DeferredImageDescriptor extends ImageDescriptor implements IAdaptable {
 	/**
 	 * The supplier of the class. Note that there is a non-code reference to this
 	 * field in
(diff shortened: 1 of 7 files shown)


Dani-Hub:
@laeubi I would be interested to get feedback on that PULL request

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
```diff
@@ -80,4 +82,12 @@ public Image createImage(boolean returnMissingImageOnError, Device device) {
 		return ImageDescriptor.createFromURL(url).createImage(returnMissingImageOnError, device);
 	}
 
+	@Override
+	public <T> T getAdapter(Class<T> adapter) {
+		if (adapter == URL.class) {
+			return adapter.cast(supplier.get());
```
laeubi: The Supplier should only be queried once for the lifetime of the image descriptor.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
```diff
@@ -304,12 +303,24 @@ public String getImagePath(int zoom) {
 		return null;
 	}
 
+	private URL toURL() {
+		if (location == null) {
```
laeubi: shou should join both like if (location == null || name == null)  ...

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
```diff
@@ -304,12 +303,24 @@ public String getImagePath(int zoom) {
...
+		return location.getResource(name);
+	}
+
 	@Override
 	public <T> T getAdapter(Class<T> adapter) {
-		if (adapter == ImageFileNameProvider.class) {
-			// Support testing ImageFileNameProvider characteristics, see #396
-			return adapter.cast(this);
+		if (adapter == URL.class) {
+			return adapter.cast(toURL());
```
laeubi: If to URL is only used here, it might be better to inline the code than having a new method....

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/menus/MenuHelper.java
```diff
@@ -453,62 +373,19 @@ public static MCommand getMCommand(MApplication application, CommandContribution
 		return null;
 	}
 
-	public static String getIconURI(ImageDescriptor descriptor, IEclipseContext context) {
+	public static String getIconURI(ImageDescriptor descriptor, @SuppressWarnings("unused") IEclipseContext context) {
```
laeubi: Why is EclipseContext unused here?

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
Dani-Hub: That's an interesting suggestion, because the whole approach of the DeferredImageDescriptor is to always call the Supplier on demand for the URL and there seems to be no caching intended. Before we do a design change of this class here I would recommend to add a second reviewer that weights his/her words in here.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
Dani-Hub: Sure, will do.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
Dani-Hub: I have no problem with that suggestion.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
laeubi: I can only judge from the javadoc that claims:

> Using the {@link ImageDescriptor#createFromURLSupplier(Supplier)} will
> trigger this class, which defers calculating the URL until the image is
> required for the first time.

So it says the calculation of the URL (=the supplier) will be deferred until the image is loaded the first time. It does not claim it is calculated multiple times, what even would defeat the purpose, why should I use a supplier for a "costly" operation when the result is that this costly operation is performed multiple times?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/menus/MenuHelper.java
Dani-Hub: See the discussion in #401, where I was asking for feedback how to cope with the fact that with the introduction of Adaters.adapt we no longer need the context. Since I was working before I got feedback, I decided to start with the least invasive approach that keeps the original signature. This might still be useful, if we would consider to optionally consider in the future usage of `org.eclipse.e4.core.services.adapter.Adapter.adapt(Object element, Class<T> adapterType)` if a context is provided and Adapters.adapt otherwise. Whether this additional complexity is worth the effort is unclear, because the default implementation of `org.eclipse.e4.core.services.adapter.Adapter.adapt(Object element, Class<T> adapterType)` calls `org.eclipse.core.runtime.Adapters.adapt(Object sourceObject, Class<T> adapter)` anyway.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
Dani-Hub: If you look at the current usage of `DeferredImageDescriptor`'s internally hold `Supplier<URL>`, you will notice that it is evaluated each time, when image data or an image is evaluated, it seems therefore consistent to do the same within the adapter to URL. If you suggest that the adapter should only return *one* (the first?) URL returned, it would be consistent, if the other functions would do so similarly. This is a design change of that class. I'm not against it, but I think that should be agreed one by an additional reviewer.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
Dani-Hub: Done.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
Dani-Hub: Done.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
laeubi: Calling the supplier multiple times seems wrong here and I most probably can assume it was because of the confusing structure of the class that no one noticed this yet. An `ImageDescriptor` can never reliable know when it is called, so having a supplier that relies on that is probably broken already.

So there should be a `getURL()` method that cache the value on the first call.

Dani-Hub: Alright, will do that.

laeubi:
Added @vogella as a second reviewer.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeferredImageDescriptor.java
Dani-Hub: Done now, to be discussed.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/menus/MenuHelper.java
Dani-Hub: @laeubi I have now made an improvement which gets rid of the unused suppression and uses - if a context is available - now the `org.eclipse.e4.core.services.adapter.Adapter` interface, otherwise as fallback the classical `org.eclipse.core.runtime.Adapters`. Do you consider this as an improvement or would you prefer to go back to the previous proposal (with possibly different adaptions)?

Dani-Hub:
@vogella This is a friendly reminder asking for a review after one week being passed now.

vogella:
> @vogella This is a friendly reminder asking for a review after one week being passed now.

Currently in vacation

laeubi:
@Dani-Hub please re-base and squash all commits into one. As no concerns from others where raised, I think this can be merged. If there are any serious problems or concerns we can fix this later on in separate changes.

Dani-Hub:
Thanks for the feedback. Will do so this evening.

Dani-Hub:
This pull request is now superseded by #421 which is a rebase and squash of this one.

PR #411, opened 2022-10-14
IDEWorkbenchErrorHandler: avoid recursive exceptions

jukzi:
https://github.com/eclipse-platform/.github/issues/72

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/IDEWorkbenchErrorHandler.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/IDEWorkbenchErrorHandler.java
@@ -133,9 +133,13 @@ public class IDEWorkbenchErrorHandler extends WorkbenchErrorHandler {
 		try {
 			exceptionCount++;
 			if (exceptionCount > 1) {
-				dialog.updateMessage(MessageFormat.format(MSG_FATAL_ERROR,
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/IDEWorkbenchErrorHandler.java
```diff
@@ -133,9 +133,13 @@ private void handleException(Throwable t) {
...
 			if (exceptionCount > 1) {
-				dialog.updateMessage(MessageFormat.format(MSG_FATAL_ERROR,
-						MSG_FATAL_ERROR_Recursive));
-				dialog.getShell().forceActive();
+				try {
+					dialog.updateMessage(MessageFormat.format(MSG_FATAL_ERROR, MSG_FATAL_ERROR_Recursive));
+					dialog.getShell().forceActive();
+				} catch (Exception e) {
+					// avoid further explosion of recursive exceptions
+					e.printStackTrace();
```
iloveeclipse, line 141: Please log this error (with appropriate message) not to console but to the log

jukzi: Why? the logToListeners would intercept the log and could run into this (or an other error) again. The mission is to find an end in the errorhandling. The first error that has initiated the trouble was already logged.

PR #414, opened 2022-10-15
Regression test for issue https://github.com/eclipse-platform/eclipse.platform/issues/227 (extreme slowdown of job rescheduling)

iloveeclipse:
Depending on the timing of waitForJobs() the test might pass, fail (spend too much time in waitForJobs()) or hang till terminated by timeout.

Reverting
https://github.com/eclipse-platform/eclipse.platform/commit/fc49c4ffaf8a6c07b58e437c62125203f80028f2 fixes the test.

See https://github.com/eclipse-platform/eclipse.platform/issues/227


mickaelistria:
The regression is rather serious, so I suggest we merge this test and revert the faulty commit (unless reverting introduces other regressions). Further improvements will be easier to provide more safely once we have this test in.
cc @jukzi

Issue #415, opened 2022-10-17
e4.ui.tests.css.core DNF since 2022-10-14


jukzi:
since https://download.eclipse.org/eclipse/downloads/drops4/I20221014-1800/testResults.php
tests for e4.ui.tests.css.core are not executed.

```
ENTRY org.eclipse.osgi 4 0 2022-10-16 20:05:31.242
!MESSAGE Application error
!STACK 1
org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-platform-suite' failed to discover tests
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverEngineRoot(EngineDiscoveryOrchestrator.java:160)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverSafely(EngineDiscoveryOrchestrator.java:132)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:107)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:78)
	at org.junit.platform.launcher.core.DefaultLauncher.discover(DefaultLauncher.java:110)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:86)
	at org.junit.platform.launcher.core.DefaultLauncherSession$DelegatingLauncher.execute(DefaultLauncherSession.java:86)
	at org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:53)
	at org.eclipse.test.EclipseTestRunner.runTests(EclipseTestRunner.java:226)
	at org.eclipse.test.EclipseTestRunner.run(EclipseTestRunner.java:202)
	at org.eclipse.test.CoreTestApplication.runTests(CoreTestApplication.java:52)
	at org.eclipse.test.CoreTestApplication.run(CoreTestApplication.java:35)
	at org.eclipse.test.CoreTestApplication.start(CoreTestApplication.java:43)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:203)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:136)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
[...]

PR #440, opened 2022-11-04
Declare JupiterTestEngine as service

mickaelistria:
So it's available when SuiteLauncher will call ServiceLoader using this bundle classloader.
(service declarations from dependencies are not visible from this classloader)

See discussion
https://github.com/eclipse-platform/eclipse.platform.ui/issues/415#issuecomment-1303323652


mickaelistria:
Not necessary if we use https://github.com/eclipse-platform/eclipse.platform.releng/pull/145

akurtakov:
As the releng patch is merged I'll abandon this one.

PR #445, opened 2022-11-11
Get bundles resigned

akurtakov:
Some of them moved to Java 11 BREE where they were stuck on older BREE. Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/661


akurtakov:
Unrelated test failure.

PR #450, opened 2022-11-14
Fix LegacyHandlerService to use IEvaluationService to resolve handler

erdalkaraca:
This fix is important as Eclipse registers listeners (Display filters) which keep handler activation states in sync the way E3 did when:
- a new Shell is activated
- a Menu is about to be shown

In E4 (as-is) handler activation state is determined by consuming the IEclipseContext which might not be proper for Menus/Shells.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
@@ -130,7 +130,7 @@ public final class EvaluationService implements IEvaluationService {
 				ISources.ACTIVE_WORKBENCH_WINDOW_SHELL_NAME, ISources.ACTIVE_EDITOR_ID_NAME,
 				ISources.ACTIVE_EDITOR_INPUT_NAME, ISources.SHOW_IN_INPUT, ISources.SHOW_IN_SELECTION,
 				ISources.ACTIVE_PART_NAME, ISources.ACTIVE_PART_ID_NAME, ISources.ACTIVE_SITE_NAME,
-				ISources.ACTIVE_CONTEXT_NAME, ISources.ACTIVE_CURRENT_SELECTION_NAME }));
+				ISources.ACTIVE_CONTEXT_NAME }));
 		context.runAndTrack(ratUpdater);
 	}
 
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
@@ -21,8 +21,10 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import org.eclipse.core.expressions.Expression;
 import org.eclipse.core.expressions.ExpressionInfo;
 import org.eclipse.core.expressions.IEvaluationContext;
@@ -54,10 +56,16 @@ import org.eclipse.ui.services.IEvaluationService;
 public final class EvaluationService implements IEvaluationService {
 	public static final String DEFAULT_VAR = "org.eclipse.ui.internal.services.EvaluationService.default_var"; //$NON-NLS-1$
 	private static final String RE_EVAL = "org.eclipse.ui.internal.services.EvaluationService.evaluate"; //$NON-NLS-1$
+
+	private static final List<String> LEGACY_WORKBENCH_VARIABLES = Arrays.asList(ISources.ACTIVE_WORKBENCH_WINDOW_NAME,
+			ISources.ACTIVE_WORKBENCH_WINDOW_SHELL_NAME, ISources.ACTIVE_EDITOR_ID_NAME,
+			ISources.ACTIVE_EDITOR_INPUT_NAME, ISources.SHOW_IN_INPUT, ISources.SHOW_IN_SELECTION,
+			ISources.ACTIVE_PART_NAME, ISources.ACTIVE_PART_ID_NAME, ISources.ACTIVE_SITE_NAME,
+			ISources.ACTIVE_CONTEXT_NAME, ISources.ACTIVE_CURRENT_SELECTION_NAME);
+
 	private boolean evaluate = false;
 	private ExpressionContext legacyContext;
 	private IEclipseContext context;
-	private IEclipseContext ratContext;
 	private int notifying = 0;
 
 	private ListenerList<IPropertyChangeListener> serviceListeners = new ListenerList<>(ListenerList.IDENTITY);
@@ -65,19 +73,18 @@ public final class EvaluationService implements IEvaluationService {
 	LinkedList<EvaluationReference> refs = new LinkedList<>();
 	private ISourceProviderListener contextUpdater;
 
-	private HashSet<String> ratVariables = new HashSet<>();
-	private RunAndTrack ratUpdater = new RunAndTrack() {
+	private Set<String> knownReferencedVariables = new HashSet<>();
+	// Keeps the local/hidden legacyContext variables pool up-to-date.
+	private RunAndTrack legacyContextUpdater = new RunAndTrack() {
 		@Override
 		public boolean changed(IEclipseContext context) {
 			context.get(RE_EVAL);
-			String[] vars = ratVariables.toArray(new String[ratVariables.size()]);
-			for (String var : vars) {
+			List<String> allVars = new ArrayList<>(LEGACY_WORKBENCH_VARIABLES);
+			allVars.addAll(knownReferencedVariables);
+			for (String var : allVars) {
+				// pipe from originating context into legacy context
 				Object value = context.getActive(var);
-				if (value == null) {
-					ratContext.remove(var);
-				} else {
-					ratContext.set(var, value);
-				}
+				changeVariable(var, value);
 			}
 			// This ties tool item enablement to variable changes that can
 			// effect the enablement.
@@ -86,13 +93,12 @@ public final class EvaluationService implements IEvaluationService {
 		}
 	};
 
-	private HashSet<String> variableFilter = new HashSet<>();
 	private IEventBroker eventBroker;
 
 	public EvaluationService(IEclipseContext c) {
 		context = c;
-		ratContext = context.getParent().createChild(getClass().getName());
-		legacyContext = new ExpressionContext(context);
+		IEclipseContext legacyContextEclipseContext = context.getParent().createChild(getClass().getName());
+		legacyContext = new ExpressionContext(legacyContextEclipseContext);
 		ExpressionContext.defaultVariableConverter = new ContextFunction() {
 			@Override
 			public Object compute(IEclipseContext context, String contextKey) {
@@ -126,12 +132,7 @@ public final class EvaluationService implements IEvaluationService {
 				}
 			}
 		};
-		variableFilter.addAll(Arrays.asList(new String[] { ISources.ACTIVE_WORKBENCH_WINDOW_NAME,
-				ISources.ACTIVE_WORKBENCH_WINDOW_SHELL_NAME, ISources.ACTIVE_EDITOR_ID_NAME,
-				ISources.ACTIVE_EDITOR_INPUT_NAME, ISources.SHOW_IN_INPUT, ISources.SHOW_IN_SELECTION,
-				ISources.ACTIVE_PART_NAME, ISources.ACTIVE_PART_ID_NAME, ISources.ACTIVE_SITE_NAME,
-				ISources.ACTIVE_CONTEXT_NAME }));
-		context.runAndTrack(ratUpdater);
+		context.runAndTrack(legacyContextUpdater);
 	}
 
 	private void contextEvaluate() {
@@ -140,7 +141,7 @@ public final class EvaluationService implements IEvaluationService {
 	}
 
 	protected void changeVariable(final String name, final Object value) {
-		if (name == null || variableFilter.contains(name)) {
+		if (name == null) {
 			return;
 		}
 		if (value == null) {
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
```diff
@@ -140,7 +140,7 @@ private void contextEvaluate() {
 	}
 
 	protected void changeVariable(final String name, final Object value) {
-		if (name == null || variableFilter.contains(name)) {
```
erdalkaraca, line 143: not sure why this variable filter exists (from Paul Webster back in 2011)... maybe to prevent collisions/overwrites in the wrapped IEclipseContext? in that case, maybe a child context would help?

erdalkaraca:
26 tests failing. This bug exists since 2011, not sure if the tests and their bugfixes are workarounds for this bug. Will have a more detailed look at them...

iloveeclipse:
> not sure if the tests and their bugfixes are workarounds for this bug

Doesn't look like - many tests in **different** areas are failing.

erdalkaraca:
@iloveeclipse fixed the failing tests, my own test works locally, but not on jenkins... I am activating a Shell expecting it to have focus... is that pattern correct or do you have a hint for a better approach?

New test: org.eclipse.ui.tests.keys.DispatcherTest.testSelectionVariableHandlerConflicts()

erdalkaraca:
@vogella @akurtakov  @iloveeclipse do you have any hints on how to debug failing tests on jenkins? on my local machine running on windows 10, the failing test passes.

iloveeclipse:
Focus handling differs between Windows/Nix, so if possible avoid using focus for tests, I see to force shell activation in our internal tests on RHEL we use xdotool `/usr/bin/xdotool search --onlyvisible --name $windowCaption windowactivate`.
Not sure if xdotool is installed on jenkins. You can try to look around in platform UI / SWT tests who and how uses focus, may be there are some workarounds.

akurtakov:
I should add that any kind of robot/focus/etc. playing would most likely fail misserably if run on Wayland thus try to avoid as much as possible in tests.

erdalkaraca:
@akurtakov @iloveeclipse 
I had to work around the widget focusing issue by simulating an IEvaluationService context change by overriding the 'selection' evaluation context variable which will be used when determining if a handler is active/enabled for execution.

> I should add that any kind of robot/focus/etc. playing would most likely fail misserably if run on Wayland thus try to avoid as much as possible in tests.

erdalkaraca:
Note that org.eclipse.ui.ISources.ACTIVE_CURRENT_SELECTION_NAME is not the same as org.eclipse.e4.ui.services.IServiceConstants.ACTIVE_SELECTION

erdalkaraca:
@iloveeclipse @akurtakov @vogella any feedback? what is the formal process for PRs?

erdalkaraca:
@paulweb515 are you still active in Eclipse development? In this PR I had to touch the code you came up with in the early E4 development... would be great if you could review the changes.

akurtakov:
Honestly, I don't know much about this part .

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
```diff
@@ -130,7 +130,7 @@ public void sourceChanged(int sourcePriority, Map sourceValuesByName) {
 				ISources.ACTIVE_WORKBENCH_WINDOW_SHELL_NAME, ISources.ACTIVE_EDITOR_ID_NAME,
 				ISources.ACTIVE_EDITOR_INPUT_NAME, ISources.SHOW_IN_INPUT, ISources.SHOW_IN_SELECTION,
 				ISources.ACTIVE_PART_NAME, ISources.ACTIVE_PART_ID_NAME, ISources.ACTIVE_SITE_NAME,
-				ISources.ACTIVE_CONTEXT_NAME, ISources.ACTIVE_CURRENT_SELECTION_NAME }));
+				ISources.ACTIVE_CONTEXT_NAME, /* ISources.ACTIVE_CURRENT_SELECTION_NAME */ }));
```
mickaelistria: Why is this comments?

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/keys/DispatcherTest.java
```diff
@@ -91,4 +97,44 @@ public void testDispatcherMultipleCommandsOnlyOneEnabled() throws Exception {
...
+		}
+
+		// invalidate the 'selection' variable and trigger copy operation again
+		{
+			// simulate that some user or platform code triggered a change in the evaluation
+			// context
+			// in real scenario, this would be opening a new Shell, for example
+			{
+				IEclipseContext ectx = projectExplorer.getSite().getService(IEclipseContext.class);
+				ectx.set(ISources.ACTIVE_CURRENT_SELECTION_NAME, null);
```
mickaelistria, line 127: And then used here?

mickaelistria (commented):
I don't get the part about ISources.ACTIVE_CURRENT_SELECTION_NAME ; can you please explain? Does that have to be part of the initial fix?
Do the test reproduce an existing issue before the change?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
erdalkaraca: It is not clear to me why this IEvaluationContext has to filter those variables at all. So, in order to minimize regressions, I only removed the "selection" variable from the blacklist.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/keys/DispatcherTest.java
erdalkaraca: in the previous location, the "selection" variable was not filtered anymore... here I am resetting its value which is the whole point of this patch

mickaelistria: So `ISources.ACTIVE_CURRENT_SELECTION_NAME` is the actual initial problem, and changing the evaluationService is the solution?

erdalkaraca:
> I don't get the part about ISources.ACTIVE_CURRENT_SELECTION_NAME ; can you please explain? Does that have to be part of the initial fix? Do the test reproduce an existing issue before the change?

In E3, when a handler's state had to be evaluated the IEvaluationService provided a context of various variables such as the "selection" variable (== ISources.ACTIVE_CURRENT_SELECTION_NAME). When a menu or new shell was shown, the "selection" variable was set to null. In E4 this is not the case as it is initialized with the selection in the current IEclipseContext. This leads to potentially wrong handlers being selected in dialogs.

See the following old bugzilla entry:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=499155

Without the patch, a custom copy handler is called in a dialog instead of the WidgetCopyHandler of the platform.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/keys/DispatcherTest.java
erdalkaraca: well, the evaluation context (containing the selection variable) is changed and the IEvaluationService is used which was not the case previously... so, there are two bugs from my point of view:
- use IEvaluationService to get a more accurate IEvaluationContext than just wrapping the IEclipseContext
- in the IEvaluationService consider changes of the selection variables which are triggered by legacy mechanisms

erdalkaraca:
Consider the following handler:

```
<extension
         point="org.eclipse.ui.handlers">
      <handler
            class="TestHandler"
            commandId="org.eclipse.ui.edit.copy">
         <activeWhen>
            <with
                  variable="selection">
               <or>
                  <iterate
                        ifEmpty="false"
                        operator="or">
                     <instanceof
                           value="org.eclipse.core.resources.IFile">
                     </instanceof>
                  </iterate>
                  <iterate
                        ifEmpty="false"
                        operator="or">
                     <instanceof
                           value="org.eclipse.core.resources.IProject">
                     </instanceof>
                  </iterate>
                  <iterate
                        ifEmpty="false"
                        operator="or">
                     <instanceof
                           value="org.eclipse.core.resources.IFolder">
                     </instanceof>
                  </iterate>
               </or>
            </with>
         </activeWhen>
      </handler>
   </extension>
```

This handler should not be active in a dialog and it was not active in E3. With the fix, the handler will not be chosen for execution as the "selection" variable is reset when the dialog is opened. This is also what the test case tries to simulate.

mickaelistria:
OK, thanks for the explanation.
I see this PR is marked as Draft. Is there anything missing before it can be turned into "Ready for Review" and then merged?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
mickaelistria: Let's just remove it instead of commenting it out.

mickaelistria (commented):
Can you please squash your change into a single commit and let it have a descriptive message about the why and how of this change?
This change contains some risk of regression (as you mentioned, it's not clear why some variables are filtered), so having a proper description will help to take the best decision if a regression occurs later.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/services/EvaluationService.java
```diff
@@ -130,7 +130,7 @@ public void sourceChanged(int sourcePriority, Map sourceValuesByName) {
 				ISources.ACTIVE_WORKBENCH_WINDOW_SHELL_NAME, ISources.ACTIVE_EDITOR_ID_NAME,
 				ISources.ACTIVE_EDITOR_INPUT_NAME, ISources.SHOW_IN_INPUT, ISources.SHOW_IN_SELECTION,
 				ISources.ACTIVE_PART_NAME, ISources.ACTIVE_PART_ID_NAME, ISources.ACTIVE_SITE_NAME,
-				ISources.ACTIVE_CONTEXT_NAME, ISources.ACTIVE_CURRENT_SELECTION_NAME }));
+				ISources.ACTIVE_CONTEXT_NAME }));
```
mickaelistria: That part seems really strange. To be honest, I'd rather avoid just making `selection` an exception here without clear understanding of why this is necessary for it and not for others.

erdalkaraca: Hopefully, Paul Webster can comment on that. First, I removed the whole blacklist but got many regressions, then limited to removing selection from it. I think since the bug exists since E4, a lot of people might have created workarounds and thus getting regressions when getting rid of the blacklist.

erdalkaraca: I could imagine, that initially, those variables from IEvaluationContext collided with the variables from IEvaluationContext which are similar concepts but have slightly different semantics. So, in order to make both work together, the blacklist was introduced (just guessing).

erdalkaraca: image

at line 168, the denoted variable was filtered (2005)... it seems that filtering concept was then extended for many more variables in 2011 when compat layer was developed
that line became stale when the filter list was introduced

erdalkaraca: @mickaelistria I do not think we will get any more confident about the intentions of that filter mechanism without a clearer hint... so, how to proceed?

mickaelistria: Looking at git blame, we may get some clues on bugs.eclipse.org/338056

erdalkaraca: so, the change in that bug/commit suggests that the filtering mechanism was introduced to prevent source providers to overwrite "system variables" such as "selection"

erdalkaraca: `org.eclipse.ui.ISourceProviderListener does` not mention that system variables are ignored...

mickaelistria: I believe the "system" source provider is a source provider like any other; so filtering its variables out here doesn't make sense: the EvaluationContext will just miss updates, and that -I believe- is not what is desired here.
I suspect you can try removing the variableFiler entirely and see what happens. It would be more correct to me.
About the issue of overriding variables or not, This is more something that the SourceProviderService should manage when loading the providers to enforce highest priority to the system ones; but as I see as a separate topic that we don't need to fix here, and that was neither necessary to fix as part of bugs.eclipse.org/338056

erdalkaraca: The workbench/system source provider's changes are effectively ignored by this filter:

org.eclipse.ui.internal.services.WorkbenchSourceProvider.listener

Just debugged to verify the behavior.

erdalkaraca: I analyzed one failing test which already revealed a conceptual gap in the interaction between E4 and E3 compat layer:

1. the `org.eclipse.e4.ui.services.ContextServiceAddon` listens in the WorkbenchContext (IEclipseContext) for activeContexts changes. This is done at E4 level
2. the E3 compat layer listens for source changes in the IEvaluationService's context

Filtering those system variables makes sure that the IEvaluationService's context is not overwritten as it is kept up to date by E3 legacy mechanisms. The ContextServiceAddon (keeping the global ContextManager's state up to date) must be done in E3 compat layer not in E4 (as at E4 level there is no IEvaluationService) or at least make sure that the global ContextManager is fed by the state of the IEvaluationService.

PR #455, opened 2022-11-17
Fix #454 : ExtensionTrackeContextFunction produces a StackOverflow.

### LINKED ISSUE #454: ExtensionTrackeContextFunction produces a StackOverflow...
The org.eclipse.e4.ui.internal.workbench.ExtensionTrackeContextFunction always produces a StackOverflow with this code : 

	public Object compute(IEclipseContext context, String contextKey) {
		**IExtensionTracker tracker = context.getLocal(IExtensionTracker.class);**.     // THIS ALWAYS RECALL compute  -> Stack Overflow
		if (tracker == null) {
			tracker = createdObjects.computeIfAbsent(context, ctx -> {
				return new UIExtensionTracker(runnable -> {
					UISynchronize synchronize = ctx.get(UISynchronize.class);
					if (synchronize != null) {
						synchronize.asyncExec(runnable);
					}
				}, log);

			});
		}
		return tracker;
	}

Actually the tracker should be cached in the class, but I don't know what is this class for (no javadoc...)...

opcoach:
Must cache the tracker when created to avoid a StackOverflow (getting a context function in the context will call compute again...).

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/ExtensionTrackeContextFunction.java
+++ b/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/ExtensionTrackeContextFunction.java
@@ -36,12 +36,15 @@ public class ExtensionTrackeContextFunction extends ContextFunction implements E
 	@Reference
 	private ILog log;
 
-	private Map<IEclipseContext, IExtensionTracker> createdObjects = new ConcurrentHashMap<>();
+	private Map<IEclipseContext, IExtensionTracker> createdObjects = null;
+
+	private IExtensionTracker tracker = null;
 
 	@Override
 	public Object compute(IEclipseContext context, String contextKey) {
-		IExtensionTracker tracker = context.getLocal(IExtensionTracker.class);
+
 		if (tracker == null) {
+			createdObjects = new ConcurrentHashMap<>();
 			tracker = createdObjects.computeIfAbsent(context, ctx -> {
 				return new UIExtensionTracker(runnable -> {
 					UISynchronize synchronize = ctx.get(UISynchronize.class);
@@ -52,6 +55,7 @@ public class ExtensionTrackeContextFunction extends ContextFunction implements E
 
 			});
 		}
+
 		return tracker;
 	}
 
@@ -59,6 +63,7 @@ public class ExtensionTrackeContextFunction extends ContextFunction implements E
 	void shutdown() {
 		createdObjects.values().forEach(IExtensionTracker::close);
 		createdObjects.clear();
+		tracker = null;
 	}
 
 	@Override


opcoach:
@laeubi can you check it ?

bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/ExtensionTrackeContextFunction.java
```diff
@@ -36,12 +36,15 @@ public class ExtensionTrackeContextFunction extends ContextFunction implements E
 	@Reference
 	private ILog log;
 
-	private Map<IEclipseContext, IExtensionTracker> createdObjects = new ConcurrentHashMap<>();
+	private Map<IEclipseContext, IExtensionTracker> createdObjects = null;
```
laeubi, line 39: Why lazy init this? any advantage? THis wil possibly introduce race contions?

opcoach: No reason and no advantage at all.. just to initialize with the tracker... this is the same for me... what do you mean by race contions ?

laeubi:
As we have similar code elsewhere, it might be better to find/fix the root cause then work around it?

opcoach:
I think that context function should not check and get its value to check if it is already stored in the context...  The goal of context function is to give a 'dynamic' value when it is called. 

If the value is always the same, you can also add @Creatable and/or @Singleton in the class definition... these annotations are used to create automatically the needed objects. But they do not put them in the context... (must be set in the constructor by injecting current context and inject this... ). 

Or we can also use your own implementation or javax.inject.Provider<T> to return the instance. 

Actually why this object is provided in the context using a context function ?

laeubi:
> think that context function should not check and get its value to check if it is already stored in the context...

Looking a bit closer at [EventBrokerFactory](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.e4.ui.services/src/org/eclipse/e4/ui/services/events/EventBrokerFactory.java) I think you are right that one can't use the declared **interface** but the EventBrokerFactory uses the **implementation** to store a value... thats really nasty.

Anyways I think the fix would be to simply delete the line 

`IExtensionTracker tracker = context.getLocal(IExtensionTracker.class);`

but the map must be initialized at construction time to prevent race conditions if two threads might call the context funtion in parallel.

> Actually why this object is provided in the context using a context function ?

Because it needs access to to notifications when a context is gone to perform cleanup actions.

laeubi:
I think that should fix the issue:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/456

opcoach:
FYI, I have also this exception when I catch throwable in the context spy (-> ie : I catch the StackOverflowError) : 

The context function computation detects a cycle... (but it will now be fixed with your current PR).

!ENTRY org.eclipse.ui 4 0 2022-11-17 18:32:45.880
!STACK 0
java.lang.RuntimeException: Cycle while computing value org.eclipse.e4.ui.internal.workbench.ExtensionTrackeContextFunction@459df48d
	at org.eclipse.e4.core.internal.contexts.ValueComputation.get(ValueComputation.java:63)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.internalGet(EclipseContext.java:243)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:236)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:606)
	at org.eclipse.e4.ui.internal.workbench.ExtensionTrackeContextFunction.compute(ExtensionTrackeContextFunction.java:48)
	at org.eclipse.e4.core.internal.contexts.ValueComputation.get(ValueComputation.java:70)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.internalGet(EclipseContext.java:243)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.get(EclipseContext.java:230)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.get(EclipseContext.java:586)
	at org.eclipse.ui.internal.Workbench.getExtensionTracker(Workbench.java:3285)
	at org.eclipse.oomph.setup.ui.SetupUIPlugin$1.run(SetupUIPlugin.java:230)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4318)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3941)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1155)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1046)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:155)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:643)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:550)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:171)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:203)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:136)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:402)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:255)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:659)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:596)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1467)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1440)

vogella:
I assume @laeubi commit has fixed this. Please restore if that is not true

PR #456, opened 2022-11-17
Do not lookup IExtensionTracker in Context function that produce it

### LINKED ISSUE #454: ExtensionTrackeContextFunction produces a StackOverflow...

laeubi:
Fix #454

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/ExtensionTrackeContextFunction.java
+++ b/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/ExtensionTrackeContextFunction.java
@@ -40,19 +40,15 @@ public class ExtensionTrackeContextFunction extends ContextFunction implements E
 
 	@Override
 	public Object compute(IEclipseContext context, String contextKey) {
-		IExtensionTracker tracker = context.getLocal(IExtensionTracker.class);
-		if (tracker == null) {
-			tracker = createdObjects.computeIfAbsent(context, ctx -> {
-				return new UIExtensionTracker(runnable -> {
-					UISynchronize synchronize = ctx.get(UISynchronize.class);
-					if (synchronize != null) {
-						synchronize.asyncExec(runnable);
-					}
-				}, log);
+		return createdObjects.computeIfAbsent(context, ctx -> {
+			return new UIExtensionTracker(runnable -> {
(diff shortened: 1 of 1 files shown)


opcoach (commented):
Please consider this remark, may be this is the behavior you want to have...

I think you should really cache the tracker value... If you have a look on how the context function is computed (see EclipseContext::internalGet around line 262), it will create a new instance of ValueComputation (line 263) or reuse the existing one if already computed and getting value in the same context (line 241) but it will ask the value computation to compute it again.. 

So if you want to return a new Tracker every time the value is get from context, I agree with your PR. But if you want to keep the same tracker and reuse it, I don not agree. 

As I don't know what you want to do, please confirm the right scenario.

opcoach:
And build is broken due to data binding ?

laeubi:
The tracker is already cached in the concurrent map based on the context that requests it.

laeubi:
> And build is broken due to data binding ?

The Jenkins build succeeds, @vogella would you consider this for RC2?

laeubi:
@opcoach can you verify that this fixes your issues you see with the `ContextSpy`?

opcoach:
Yes this is ok for me.

laeubi:
@opcoach would you then give an official "approving review) (just switch to the "Files changed" Tab) so we have a commiter approving it and then need a approval from @vogella @akurtakov @mickaelistria ... or any other PL?

opcoach (approved):

vogella:
Master is open now

vogella:
@laeubi can this be merged?I think you wanted to merge it even for the last release but as we had a release freeze you could not.

laeubi:
Yes this is ready for merging

PR #469, opened 2022-12-04
Fix javadoc issues in SearchPattern

pbodnar:
This cleanup is meant mainly as a preparation for #12, so that the new feature can be reasonably documented within the `SearchPattern` class.

This doesn't include fixes of private methods like `camelCaseMatch()`, because 1) these would be probably too tedious to review, 2) these are not critical for #12.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
@@ -18,18 +18,15 @@ import org.eclipse.ui.internal.misc.TextMatcher;
 
 /**
  * A search pattern defines how search results are found.
- *
  * <p>
  * This class is intended to be subclassed by clients. A default behavior is
- * provided for each of the methods above, that clients can override if they
- * wish.
- * </p>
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/SearchPattern.java
```diff
@@ -72,7 +72,6 @@ public class SearchPattern {
 	 * <code>NullPointerException</code> type.</li>
 	 * </ul>
 	 *
-	 *
 	 * <br>
 	 * Can be combined with {@link #RULE_PREFIX_MATCH} match rule. For example, when
```
pbodnar, line 76: > `Can be combined with {@link #RULE_PREFIX_MATCH} match rule.`

I believe this text is also problematic. Because in reality, there is no need to combine with the mentioned flag at all - the fallback is simply automatic. I'm not sure how to improve this easily, maybe we can leave this as-is for now...?

vogella: Sounds fine for me.

vogella:
Thanks @pbodnar

PR #474, opened 2022-12-08
Escape mnemonics in preference page titles

Bananeweizen:
The current release of Wild Web Developer uses a preference page "Validation & Resolution". The title is displayed wrongly on the (right hand side) preference page content because the ampersand is interpreted as mnemonic.

reproduced using one of the example plugins:
grafik

with the fix applied:
grafik


mickaelistria:
Can you please add a commits that bumps the version?

mickaelistria:
Thank you!

Issue #487, opened 2022-12-15
Use ctrl+e Quick Switch Editor to report java.lang.IllegalArgumentException


lyh061619:
at work Use ctrl+e Quick Switch Editor to report Error executing 'org.eclipse.ui.window.openEditorDropDown': java.lang.IllegalArgumentException

os:
windows 10

eclipse version:
4.26.0.20221201-1200

Installed plug-ins:
![20221215085531](https://user-images.githubusercontent.com/15118571/207747340-6a9fd675-e5d1-4c18-b4a9-f956b338ffea.png)

![20221214224215](https://user-images.githubusercontent.com/15118571/207744523-292566b2-dca6-415a-be34-198700510fbb.png)

I tried to delete the workspace and the reconstruction problem still exists,
The abnormal information is as follows:

!ENTRY org.eclipse.e4.ui.workbench 4 0 2022-12-14 22:17:10.891
!MESSAGE Execution exception for: ParameterizedCommand(Command(org.eclipse.ui.window.openEditorDropDown,Quick Switch Editor,
		Open the editor drop down list,
		Category(org.eclipse.ui.category.window,Window,null,true),
		WorkbenchHandlerServiceHandler("org.eclipse.ui.window.openEditorDropDown"),
		,,true),null) in 
	context chain: WorkbenchContext -> TrimmedWindowImpl (IDEWindow) Context -> PerspectiveImpl (org.eclipse.debug.ui.DebugPerspective) Context -> PartImpl (org.eclipse.e4.ui.compatibility.editor)  removeOnHide org.eclipse.m2e.editor.MavenPomEditorContext -> MultiPageEditorSite (org.eclipse.ui.internal.genericeditor.ExtensionBasedTextEditor)
!STACK 0
org.eclipse.core.commands.ExecutionException: Error executing 'org.eclipse.ui.window.openEditorDropDown': java.lang.IllegalArgumentException
	at org.eclipse.e4.core.commands.internal.HandlerServiceHandler.execute(HandlerServiceHandler.java:170)
	at org.eclipse.core.commands.Command.executeWithChecks(Command.java:488)
	at org.eclipse.core.commands.ParameterizedCommand.executeWithChecks(ParameterizedCommand.java:485)
	at org.eclipse.e4.core.commands.internal.HandlerServiceImpl.executeHandler(HandlerServiceImpl.java:213)
[...]

PR #491, opened 2022-12-17
Ctrl+E: Fix crash when editor without name collision has been split

fapdash:
- The previous code expected that the same file can never be opened
  in multiple editors.
- When an editor gets split it shows up twice in the Quick Switch
  Editors popup. These two editors share the exact same path.
- Popup now shows just the editor title in these cases. This is the behavior that existed before the name collision handling was added.

works on #487

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
@@ -186,10 +186,8 @@ public class WorkbookEditorsHandler extends FilteredTableBaseHandler {
 		});
 
 		for (List<Entry<EditorReference, IPath>> groupedEditorReferences : collisionsMap.values()) {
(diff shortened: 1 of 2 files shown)


fapdash:
Without merging the `if` and `if else` clause in https://github.com/eclipse-platform/eclipse.platform.ui/pull/491/commits/28f4e847e16419cfa3cb0fd3597c907d921f50e4 it would be possible to add a number to the end of the editor label.
Something like `Exampe.java (0)` and `Example.java (1)` but since I don't use the split editor toggle I'm not sure if this would be useful. WDYT?

vogella:
> Without merging the `if` and `if else` clause in [28f4e84](https://github.com/eclipse-platform/eclipse.platform.ui/commit/28f4e847e16419cfa3cb0fd3597c907d921f50e4) it would be possible to add a number to the end of the editor label. Something like `Exampe.java (0)` and `Example.java (1)` but since I don't use the split editor toggle I'm not sure if this would be useful. WDYT?

I think that is not necessary, lets see if someone requests this via a separate ticket.

PR #495, opened 2022-12-19
Ctrl+E: Handle 3 or more references to the same path

fapdash:
Previously the code expected that there can be two references to the same path max, because split editor is a toggle.
As @harawata pointed out[0], it is possible to have more than two references to the same path because editor cloning can be done without any limits.
This commit therefore removes the check for the reference list length.

[0] https://github.com/eclipse-platform/eclipse.platform.ui/issues/487#issuecomment-1357466836


vogella:
Thanks @fapdash

PR #497, opened 2022-12-20
Fix WorkbookEditorsHandlerTest breaking under Windows by not verifying editor order

fapdash:
Editor order in tests under Linux are different from order observed by splitting an Editor manually.
Tests on Windows show the correct behavior and therefore break the assertions written on Linux.
This commit removes the validation of the editor order to make the test pass on Windows.

see https://github.com/eclipse-platform/eclipse.platform.ui/issues/487#issuecomment-1358966421

works on #487


iloveeclipse:
Thanks.

PR #499, opened 2022-12-22
[Build] Leverage Tycho pomless for platform-specific Fragments

HannesWell:
Locally this is failing at the moment, but it would be great to enhance Tycho to work without poms in this case as well:
https://github.com/eclipse-tycho/tycho/issues/1149
@laeubi for you. :)


laeubi:
@HannesWell this is not due to Tycho requiring the environment configuration but due to this configuration that is now missing:

```
<dependency-resolution>
  <extraRequirements>
    <requirement>
      <type>eclipse-plugin</type>
      <id>org.eclipse.core.resources</id>
      <versionRange>0.0.0</versionRange>
    </requirement>
  </extraRequirements>
</dependency-resolution>
```

I can't tell why this is required and not the fragment simply import-package/require.bundle  the stuff....

HannesWell:
> @HannesWell this is not due to Tycho requiring the environment configuration but due to this configuration that is now missing:
> 
> ```
> <dependency-resolution>
>   <extraRequirements>
>     <requirement>
>       <type>eclipse-plugin</type>
>       <id>org.eclipse.core.resources</id>
>       <versionRange>0.0.0</versionRange>
>     </requirement>
>   </extraRequirements>
> </dependency-resolution>
> ```
> 
> I can't tell why this is required and not the fragment simply import-package/require.bundle the stuff....

Indeed. Thanks for pointing that out. 🤦🏽
It looks like the fragment host requires o.e.core.resources only optionally and I think the goal was to not introduce a non-optional requirement through the fragment. I now added it as optional required-bundle.

However I wonder why that requirement is optional in the host `org.eclipse.ui.ide`? I see classes from the resources Plugin being used all over that Plugin and I would be surprised if all code paths are prepared for a absent resources plugin? The chances are also high that is required from any other dependency.

Does anyone know the purpose of those optional deps?

bundles/org.eclipse.ui.win32/.project
```diff
@@ -3,13 +3,6 @@
 	<name>org.eclipse.ui.win32</name>
 	<comment></comment>
 	<projects>
-		<project>org.eclipse.core.boot</project>
-		<project>org.eclipse.core.resources</project>
-		<project>org.eclipse.core.runtime</project>
-		<project>org.eclipse.help</project>
-		<project>org.eclipse.ui</project>
-		<project>org.eclipse.ui.ide</project>
-		<project>org.eclipse.ui.views</project>
```
HannesWell, line 12: Does anybody know why this was here? Besides `org.eclipse.core.boot` all of those Plug-ins are dependencies of this one and are therefore are implicitly required by the PDE build order computer.
I have never seen `org.eclipse.core.boot` before and assume it exists in pre OSGi times?

When searching the entire platform-aggregator tree I only see some hits in the documentation, e.g. in the Platform class referring to `org.eclipse.core.boot.BootLoader`. It looks like that is obsolete as well?

HannesWell: @mickaelistria, @akurtakov, @iloveeclipse can you tell anything about that?

iloveeclipse: > I have never seen org.eclipse.core.boot before and assume it exists in pre OSGi times?

Same here. This is from Eclipse 2.0 and can be removed.

HannesWell: OK. Thanks.

HannesWell:
With this change eclipse.platform.ui is now pom-free, besides the root and examples-aggregator pom.xml.

vogella:
Thanks @HannesWell for making the 
build simpler

PR #506, opened 2023-01-03
Project wizards should be shown in "new project" category

### LINKED ISSUE #502: Project wizzards should be shown below "New Project ..."
Currently if one implements the extension point
https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/extension-points/org_eclipse_ui_navigator_navigatorContent.html?cp=2_1_1_193#e.commonWizard

The wizard is displayed a bit random:

1. if **no group** is declared, it is showed together with the `New File` / `New Folder` group
2. if a **group is defined** it is shown below

But a wizard that has the property / tag "project" is a project wizard and therefore should better be displayed below `New Project ...`

grafik

laeubi:
Currently wizard that are marked with project=true are shown with other generic ones instead of in the "new project" section of the navigator.

This adds them to the "new project" category so they are easier identified by a user that they create a new project.

Fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/502

### DIFF
--- a/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/NewActionProvider.java
+++ b/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/NewActionProvider.java
@@ -13,6 +13,9 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.navigator.resources.actions;
 
+import java.util.Arrays;
+import java.util.function.Predicate;
+
 import org.eclipse.jface.action.IAction;
 import org.eclipse.jface.action.IMenuManager;
 import org.eclipse.jface.action.MenuManager;
@@ -22,6 +25,7 @@ import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.actions.ActionFactory;
 import org.eclipse.ui.actions.NewExampleAction;
 import org.eclipse.ui.actions.NewProjectAction;
+import org.eclipse.ui.internal.dialogs.WorkbenchWizardElement;
 import org.eclipse.ui.internal.navigator.resources.plugin.WorkbenchNavigatorMessages;
 import org.eclipse.ui.navigator.CommonActionProvider;
 import org.eclipse.ui.navigator.ICommonActionExtensionSite;
@@ -29,6 +33,7 @@ import org.eclipse.ui.navigator.ICommonMenuConstants;
 import org.eclipse.ui.navigator.ICommonViewerWorkbenchSite;
 import org.eclipse.ui.navigator.WizardActionGroup;
 import org.eclipse.ui.wizards.IWizardCategory;
+import org.eclipse.ui.wizards.IWizardDescriptor;
 import org.eclipse.ui.wizards.IWizardRegistry;
 
 /**
@@ -57,6 +62,9 @@ public class NewActionProvider extends CommonActionProvider {
 
 	private static final String NEW_MENU_NAME = "common.new.menu";//$NON-NLS-1$
 
+	private static final Predicate<IWizardDescriptor> PROJECT_WIZARD_FILTER = wizardDescriptor -> Arrays
+			.stream(wizardDescriptor.getTags()).anyMatch(WorkbenchWizardElement.TAG_PROJECT::equals);
+
 	private ActionFactory.IWorkbenchAction showDlgAction;
 
 	private IAction newProjectAction;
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/NewActionProvider.java
```diff
@@ -75,9 +81,22 @@ public void init(ICommonActionExtensionSite anExtensionSite) {
...
-
+			Predicate<IWizardDescriptor> projectWizardFilter = wizardDescriptor -> {
+				String[] tags = wizardDescriptor.getTags();
+				for (String tag : tags) {
+					if (WorkbenchWizardElement.TAG_PROJECT.equals(tag)) {
+						return true;
+					}
+				}
+				return false;
+			};
```
mickaelistria: Arrays.asList(wizardDescriptor.getTags()).contains(WorkbenchWizardElement.TAG_PROJECT)

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/WizardActionGroup.java
```diff
@@ -177,6 +182,29 @@ public WizardActionGroup(IWorkbenchWindow aWindow,
...
 
+	/**
+	 * Set a filter for {@link IWizardDescriptor}s
+	 *
+	 * @param descriptorFilter the filter to set, might be <code>null</code> if no
+	 *                         filtering is desired.
+	 *
+	 * @since 3.11
+	 */
+	public void setFilter(Predicate<IWizardDescriptor> descriptorFilter) {
```
mickaelistria: What about having it final and set in a new constructor?

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/WizardActionGroup.java
```diff
@@ -177,6 +182,29 @@ public WizardActionGroup(IWorkbenchWindow aWindow,
...
+	}
+
+	/**
+	 * Set if this one uses explicit seperators between groups or not
+	 *
+	 * @param useSeparators <code>true</code> if seperators should be used,
+	 *                      <code>false</code> otherwhise.
+	 * @since 3.11
+	 */
+	public void setUseSeparators(boolean useSeparators) {
```
mickaelistria: What about having it final and set in a new constructor?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/NewActionProvider.java
laeubi: I'm not sure if this is any better here beside that it creates a new collection?!?

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/WizardActionGroup.java
laeubi: I first started with a constructor parameter, but this then ended up in many constructs with long arguments list so I go with a setter as it seems more descriptive.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/NewActionProvider.java
mickaelistria: It's definitely easier to read and it tells explicitly in 1 line the intent that 6 lines of code doesn't express as clearly (expressing more about the documentation than about the goal).

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/WizardActionGroup.java
mickaelistria: I still believe 1 new constructor is actually better than 1 setter as immutability is usually making maintenance and debug much easier. A long argument list may not be convenient, but it's still less error prone than mutability.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/NewActionProvider.java
laeubi: done...

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/WizardActionGroup.java
laeubi: done

PR #517, opened 2023-01-14
Wizard pages should be able to specify their minimum size.

### LINKED ISSUE #516: Wizard pages should be able to specify their minimum size
For the m2e target dependency editor, I'd like to specify the minimum size the wizard page can have. Normally, I would do this via `WizardDialog.setMinimumPageSize()`. However, this dialog is created within PDE and therefore not accessible.

I can work around this limitation by calling `getContainer().getShell().setMinimumSize()`, but I'd prefer a cleaner solution for this.

It would be good if a IWizardPage would have a `getMinimumPageSize()` (and `getMaximumPageSize()`) (the default can just use getControl().computeSize(-1,-1)).

See also :
-  https://github.com/eclipse-m2e/m2e-core/pull/1157#issuecomment-1382603940

ptziegler:
Subclasses of WizardPage can overwrite the getMinimumPageSize() method in order to set a minimal size for the wizard shell. By default, no such minimum size is set.

For the sake of backwards-compatibility, I've moved this logic into "supplement" interfaces, similar to how it's done for the IWizardContainer. Because there are already several classes implementing IWizardPage directly and would otherwise have to implement a dummy method.

If getMinimumPageSize() is not overwritten, then the minimum size of the shell is not set. This corresponds to the current behavior. Otherwise the minimum size is of the shell is calculated to be equal or larger than the minimum sizes of all wizard pages.

Resolves #516

### DIFF
--- a/bundles/org.eclipse.jface/.settings/.api_filters
+++ b/bundles/org.eclipse.jface/.settings/.api_filters
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<component id="org.eclipse.jface" version="2">
+    <resource path="src/org/eclipse/jface/wizard/IWizard.java" type="org.eclipse.jface.wizard.IWizard">
+        <filter comment="https://github.com/eclipse-platform/eclipse.platform.ui/issues/516" id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.jface.wizard.IWizard"/>
+                <message_argument value="getMinimumWizardSize()"/>
+            </message_arguments>
+        </filter>
+    </resource>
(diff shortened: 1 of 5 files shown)


vogella:
I suggest to extend the original interface with a an empty default implementation instead of creating new interfaces.

ptziegler:
> I suggest to extend the original interface with a an empty default implementation

That's a good point. I wasn't exactly sure what the convention is when dealing with API extensions, so I took the IWizardContainer as reference.

I assume this "pattern" is a relic from the pre-Java 8 era, before default methods were supported. So this was the only way to implement such extensions without breaking backwards compatibility.

jukzi:
@vogella is there anything missing or submit?

vogella:
Thanks @ptziegler 

Also thanks to @jukzi for the reminder.

PR #519, opened 2023-01-19
Remove explicit requirement to jxpath in feature

mickaelistria:
jxpath version is changing and having feature specifying an exact version is not helpful. Removing the explicit reference to jxpath would come as a transitive dependency.

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -596,13 +596,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.apache.commons.jxpath"
-         download-size="0"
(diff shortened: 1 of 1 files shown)


features/org.eclipse.e4.rcp/feature.xml
```diff
@@ -595,52 +595,6 @@
...
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.e4.core.di.annotations"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
```
akurtakov: This one for sure and I assume many others but jxpath itself should stay in the feature.

PR #520, opened 2023-01-19
Not properly disposed SWT resource in ContentTypesPreferencePage

Bananeweizen:
Disposable resources must not be tracked via a set, since that will remove images created multiple times.

java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:172)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:120)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:668)
	at org.eclipse.jface.resource.URLImageDescriptor.createImage(URLImageDescriptor.java:241)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:290)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:268)
	at org.eclipse.ui.internal.dialogs.ContentTypesPreferencePage.lambda$2(ContentTypesPreferencePage.java:363)

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ContentTypesPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ContentTypesPreferencePage.java
@@ -21,10 +21,9 @@ import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
 import java.nio.charset.Charset;
 import java.nio.charset.IllegalCharsetNameException;
 import java.util.ArrayList;
-import java.util.HashSet;
+import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
-import java.util.Set;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
@@ -113,7 +112,7 @@ public class ContentTypesPreferencePage extends PreferencePage implements IWorkb
 
 	private Button addEditorAssociationButton;
 
-	private Set<Image> disposableEditorIcons = new HashSet<>();
+	private Collection<Image> disposableEditorIcons = new ArrayList<>();
 
 	private static class Spec {
 		/**


trancexpress:
-1, removal of widgets will be very inefficient with this.

Probably better to create an extra container just for the objects that for whatever reason have hash collisions?... Why are there collisions?

Bananeweizen:
> -1, removal of widgets will be very inefficient with this.
> 
> Probably better to create an extra container just for the objects that for whatever reason have hash collisions?... Why are there collisions?

I totally don't get your comment. This is not about widgets, this is about images. And if 100 images got created, all need to be disposed, no matter whether there are duplicates.

Someone else might be able to avoid creating that many images in the first place. I'm not, since this is just one of many issues that popped up in another bug investigation.

trancexpress:
> I totally don't get your comment. This is not about widgets, this is about images. And if 100 images got created, all need to be disposed, no matter whether there are duplicates.
> 
> Someone else might be able to avoid creating that many images in the first place. I'm not, since this is just one of many issues that popped up in another bug investigation.

Ah sorry, I thought this was touching the mechanism that reports non-disposed resources. That one has big sets.

trancexpress:
> I totally don't get your comment. This is not about widgets, this is about images. And if 100 images got created, all need to be disposed, no matter whether there are duplicates.
> 
> Someone else might be able to avoid creating that many images in the first place. I'm not, since this is just one of many issues that popped up in another bug investigation.

Ah sorry, I thought this was touching the mechanism that reports non-disposed widgets.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ContentTypesPreferencePage.java
```diff
@@ -113,7 +112,7 @@ public ContentTypesPreferencePage() {
 
 	private Button addEditorAssociationButton;
 
-	private Set<Image> disposableEditorIcons = new HashSet<>();
+	private Collection<Image> disposableEditorIcons = new ArrayList<>();
```
laeubi, line 115: What about `Collections.newSetFromMap(new IdentityHashMap<>()) `? I think that's what actually is desired here...

iloveeclipse:
@SyntevoAlex : could you please check?

I haven't looked into code, but I assume resource objects that represent different native handles aren't equal, so I can't understand the problem here.

trancexpress:
> @SyntevoAlex : could you please check?
> 
> I haven't looked into code, but I assume resource objects that represent different native handles aren't equal, so I can't understand the problem here.

I checked, `Image` overrides `hashCode()` and returns that hash code of the image provider (I didn't check why though).

I think for the change using an array is fine. I don't think its possible that the same object gets added to the container twice (which would make the dispose throw an already disposed exception). And there is no other use for the container, its there just to dispose what was created.

Bananeweizen:
For everyone confused by the discussion or the code change: That collection tracks all the images created in this preference page via `ImageDescriptor.createNewImage()`. That method _guarantees_ a new image to be returned each time (or null). Therefore every single return value from that method needs to be disposed later.
The previous implementation with a Set led to some of the images being discarded (from the Set) as duplicates due to the Image.hashCode() implementation, thereby supressing some of the necessary dispose() calls. While that detail about the hash implementation might be questionable and worth a separate discussion eventually, it should not lead to us coming up with even more complicated _Set_ based collection of the Images. The simple rule (and root cause of the bug) is: Dispose _all_ images created by that method, and using a list instead of set will follow that rule (independent of whether the hash implementation is good or not).

laeubi (approved):
The explanation from @Bananeweizen sounds reasonable to me

SyntevoAlex:
So if I understand the discussion correctly, SWT's leak detector (in `Resource.initNonDisposeTracking`) works fine and I don't need to investigate it?

trancexpress:
> So if I understand the discussion correctly, leak detector works fine and I don't need to investigate leak detector?

Yes, the one for resources doesn't use sets. It should be fine.

PR #521, opened 2023-01-20
Using Text blocks in ui.tests by moving to Java 17

vogella:
As the whole infrastructure has been moved to Java 17, we can also update our tests to use Java 17.


iloveeclipse:
LGTM, thanks.

PR #523, opened 2023-01-20
Revert "Revert "Remove explicit requirement to jxpath in feature""

mickaelistria:
This reverts commit c4135cf4ee86c08b85e71314fa0f62d3e1012e89. The former commit was reverted because of a bug in Tycho and because the build was still consuming jxpath from Maven instead of Orbit. Now that 1 build succeeded, there should be nothing left of the jxpath from Maven Central, and removing this transtive dependency from the feature should simplify further maintenance and evolution.

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -596,13 +596,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.apache.commons.jxpath"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
          id="org.eclipse.e4.emf.xpath"
          download-size="0"
          install-size="0"


mickaelistria:
@iloveeclipse I think we can merge it now as an I-Build has completed with older jxpath in TP. I hope we can merge it soon as it should make further fixes easier to test and carry on.

iloveeclipse:
> @iloveeclipse I think we can merge it now as an I-Build has completed with older jxpath in TP. I hope we can merge it soon as it should make further fixes easier to test and carry on.

Honestly, I don't understand it. If this dependency is removed, shouldn't it be replaced in aggregator build - but then we will be back again where we were before?

If there would be at least some self contained explanation why we needed two changes first time and now only one change, it would be helpful.

mickaelistria:
I believe we should first wait for Tycho to be fixed before merging this one; it would indeed be safer.

akurtakov:
The fact that the dependency is removed from the feature doesn't mean it will be gone. What happens is that hard locking on exact version at build time version in feature.xml is prevented but rather let it be pulled as dependency based on osgi requirements. Only "top-level" bundles should be listed in features and let their dependencies be resolved which should save us a good number of "feature touches".

PR #526, opened 2023-01-24
Using Text blocks in ui.tests by moving to Java 17

akurtakov:
As the whole infrastructure has been moved to Java 17, we can also update our tests to use Java 17.


akurtakov:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/521 continued to have less awkward symbols in text blocks.

akurtakov:
Abandon as the other patch was merged.

iloveeclipse:
Sorry, I thought they were different. What was the reason for this one?

akurtakov:
NP, I was just trying to reduce some '\' symbols in the textblock. Nothing important, more about me testing the extend of text blocks.

trancexpress:
Isn't this PR "better"? Since there are fewer unexpected symbols. (comparing to https://github.com/eclipse-platform/eclipse.platform.ui/pull/521/files)

Or the change here doesn't work and the extra symbols are needed?

akurtakov:
I don't think they are needed but also they are not hurting so no time to rebase it. Even with the first patch it's better than it used to be.

PR #527, opened 2023-01-24
Make use of Java 17 features in o.e.ui.tests

akurtakov:


iloveeclipse:
Alex, I see you extensively uses improved instanceof. Nothing against it, but we've patched compiler code recently to generate a bit different bytecode for that (https://github.com/eclipse-jdt/eclipse.jdt.core/issues/585), and I wonder if we could first make sure the IBuilds use updated ECJ version?

akurtakov:
I'm on it.

akurtakov:
Thanv you @vogella for starting with these tests move to Java 17 so we can identify and test more such things this cycle.

akurtakov:
ECJ has been updated so this is fine to go.

Issue #467, opened 2022-11-30
SWT Dark Theme ToolItem Toggle Buttons not showing down state


mikevwdriver:
I am using the 2022-12 RC1 on Windows 10.  Toggle buttons in tool bars and even regular toggle buttons are not showing their down state as expected.

![image](https://user-images.githubusercontent.com/16846123/204877641-8ddb14e9-16a2-4c86-87c0-7cae07df6555.png)
You can't tell, but the sync with editor button is currently pressed down.

The same issue can be seen in the perspective chooser:
![image](https://user-images.githubusercontent.com/16846123/204877754-1f9206e0-2e79-460f-aa0d-0eb791fcadcf.png)

mikevwdriver:
@vogella I've identified the commit that caused this issue:
c07ab5389e1c9b75e46af9dd7cf0c522e35bba64

Before this commit, ToolItem toggle buttons had a color that indicated their down state, starting with this commit, the down state is no longer apparent on Dark theme.

edit: It does appear that regular SWT toggle buttons were broken by a different commit than the ToolItem toggle buttons.  For this separate issue, I've opened https://github.com/eclipse-platform/eclipse.platform.swt/issues/483

mikevwdriver:
```
// ToolItem prevents itself from repaints if the same color is set
((ToolItem) widget).setBackground(newColor);
```
If I remove this one line from CSSPropertyBackgroundSWTHandler.java, the ToolItems are styled correctly.

mikevwdriver:
Okay, more information.  This is a bug that affects the standard Eclipse IDE.  Our product has it's own custom CSS, it was affected by this change as well, however, simply by removing all instances of ToolItem in our dark theme CSS files, we are no longer affected by this issue.

I took a look at the platform CSS and although it does not directly reference ToolItem, I suspect it is inheriting a rule which causes the background to be explicitly set, which will unfortunately have an effect on the toggled/checked state.

humphreygao:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/340

de-jcup:
Is there any update on fixing this issue?

I am using now eclipse `2022-12` (nearly only JDT) and  I am having really big problems with this:

At Console, JUnit, Project explorer  and many more views it is not possible to determine if an icon inside a view toolbar is enabled or disabled. 

[...]

PR #538, opened 2023-01-30
Fix toolbar render problems in dark mode on resizing #536

### LINKED ISSUE #536: Dark mode: Toolbar background color vanished - toolbar rendering ugly
## Situation
The background color changes in eclipse 2022-12 in multiple views

When I start eclipse on my Linux Mint machine (GTK) I see
image

When I select something inside the tree (e.g. a Java project) The background color changes (and looks and keeps ugly after this):
image

I think this is related to #467 

The console view does always show the wrong (ugly) background behavior.

de-jcup:
This PR
- closes #536

### Details
- Toolbar has in dark mode now a dedicated color instead of css inheritance. A parent container change by resizing does not change the toolbar color any longer
- The color is choosen in way that the main toolbar looks as expected before. Also the toggle border is now visible again, which could address #467 as well.

Signed-off-by: Albert Tregnaghi <albert.tregnaghi@gmail.com>

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/css2/CSSPropertyBackgroundSWTHandler.java
+++ b/bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/css2/CSSPropertyBackgroundSWTHandler.java
@@ -36,8 +36,8 @@ import org.w3c.dom.css.CSSValue;
 public class CSSPropertyBackgroundSWTHandler extends AbstractCSSPropertyBackgroundHandler {
 
 	@Override
-	public boolean applyCSSProperty(Object element, String property,
-			CSSValue value, String pseudo, CSSEngine engine) throws Exception {
+	public boolean applyCSSProperty(Object element, String property, CSSValue value, String pseudo, CSSEngine engine)
+			throws Exception {
 		Widget widget = SWTElementHelpers.getWidget(element);
 		if (widget != null) {
 			return super.applyCSSProperty(element, property, value, pseudo, engine);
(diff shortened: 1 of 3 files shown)


de-jcup:
Here a screenshot from this commit and the chnged rendering 

image

iloveeclipse:
Just curious what is exact difference to the state before (I can't spot that)?

de-jcup:
@iloveeclipse : 
- for a screenshot with the problem please look at https://github.com/eclipse-platform/eclipse.platform.ui/issues/536#issuecomment-1408472509 
- or https://github.com/eclipse-platform/eclipse.platform.ui/issues/536#issue-1561996818

Target OS: Linux Mint 21.1, Native dark theme enabled (Mint-Y-Legacy-dark or any other dark theme)
 Tested with cinnamon (gnom3 based) + mate (gnome2 based): With both I have the mentioned problem

iloveeclipse:
Ok, if I understood it right, the difference is that the background of the toolbar changes and doesn't match the color of the toolbar buttons? But it should match them? For me both screenshots look inconsistent.

The "bad" case seem to show wrong background behind buttons in Console view but shows proper background in the front of them (note Package Explorer and Outline view). The "fixed" screenshot shows exact the opposite, which isn't better IMHO.

de-jcup:
@iloveeclipse - thank you for the feedback, @akurtakov did already some test  about the PR and gave me feedback inside the issue itself.

@akurtakov : I really tried hard to fix the problems just with CSS but it did not work.

With last commit I reverted my former CSS attempts (by force push) and did changes inside code instead. It looks now good (see next picture). 

## Details
CTabFolder handling is now stopped, which stops the color changes when the toolbar menu is selected. 

I Also fixed the problem with the missing toggle states - the are shown now, but ... to bring up the toggle selection again and to fix the wrong background rendering for the tool items, I had to remove the CSS background handling for `ToolItem` which was introduced by @vogella with https://github.com/eclipse-platform/eclipse.platform.ui/commit/c07ab5389e1c9b75e46af9dd7cf0c522e35bba64
...  of course this not the correct way, but I was not able to find a way by CSS (and I really investigated much time)

Maybe this  could be a temporary solution/workaround? Currently as a dark mode user I am really unhappy with the rendering and the missing toggle selections.

image

@vogella : Please take a look, maybe you have an Idea, how we could `ToolItem` css rendering work together with correct rendering inside CTabFolder.

vogella:
I think this is fixed with the recent changes of the dark theme by @mvm-sap. Please check again with the latest I-Build from https://download.eclipse.org/eclipse/downloads/

PR #539, opened 2023-01-30
Do not list felix.scr expliictly in e4.rcp feature.

akurtakov:
It should be pulled in as a provider of
osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.0)(!(version>=2.0)))"

Change is an effort to reduce the need to "touch" features everytime 3rd party dependency is being updated.
This is the 5th place to "touch" for update to felix.scr in the old way. https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/832

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -238,13 +238,6 @@
          unpack="false"/>
 
    <plugin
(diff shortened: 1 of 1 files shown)


akurtakov:
Push as we need I-build for this to work proper.

PR #544, opened 2023-02-01
Streamline contributing/readme

akurtakov:
Content being outdated and having 2 contributing files is at least confusing.


akurtakov:
Documentation only so nothing to be verified.

PR #548, opened 2023-02-02
Ctrl-E: Do not handle name collisions in case of custom editor input

### LINKED ISSUE #547: Regression caused by Ctrl+E: Prepend parent directories for file name collisions (#301 / #302)
#301 / #302 introduced a feature to use segments of the path in case the names returned by `IEditorInput.getName()` are non-unique.

However, this broke scenarios in which the file name / path is intentionally **not** used in the editor input name, because, for example, the file system is virtual.

As an example this breaks SAP's ABAP Development Tools:

Before:
image
After:
image

We use a custom file system implementation and the files do not really exist on disk, when calling `IPathEditorInput.getPath()` they are copied via `IFileStore.toLocalFile(EFS.CACHE, ...)` to a temporary cache file on disk.

Therefore, the assumption of change #301 / #302 that every editor that uses `IPathEditorInput` uses regular files on disk is not correct.

I am not sure how to solve this in a generic way.

One approach could be to limit the new handling to cases where `editorReference.getName()` is really equal to the actual file name that `iPathEditorInput.getPath().toFile().toPath().getFileName().toString()` produces.

sratz:
In case the editor input returns a 'name' that is not the actual file name, do not apply the collision resolving logic based on file path segments.

In such cases we need to assume that the editor input is custom-built and such files might be virtual, in which case showing a path is not the intended behavior.

Fixes #547.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.stream.Collectors;
 import org.eclipse.core.commands.Command;
 import org.eclipse.core.commands.ParameterizedCommand;
 import org.eclipse.core.runtime.Adapters;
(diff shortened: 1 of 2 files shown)


sratz:
@fapdash / @mickaelistria Could you have a look?

fapdash:
lgtm, I think it's fine to not handle cases where filename and editor input name don't match.
I outline a different approach in the [issue](https://github.com/eclipse-platform/eclipse.platform.ui/issues/547#issuecomment-1415808005), but I think we should only implement that when it's needed / asked for.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbookEditorsHandler.java
```diff
@@ -161,46 +162,56 @@ private List<EditorReference> getParts(WorkbenchPage page) {
...
+					try {
+						// we only detect collisions for IPathEditorInput and only if the name used by
+						// the editor reference is the filename. Otherwise, this would break scenarios
+						// where editors override the name used, e.g. for virtual file systems using
+						// org.eclipse.core.internal.filesystem.FileCache
+						IPathEditorInput iPathEditorInput = Adapters.adapt(editorReference.getEditorInput(),
+								IPathEditorInput.class);
+						IPath path;
+						if (iPathEditorInput != null && (path = iPathEditorInput.getPath()) != null && path.toFile()
+								.toPath().getFileName().toString().equals(groupedEditorReferences.getKey())) {
```
fapdash: I think you can use `path.lastSegment()` here instead of converting to a nio `Path`.

sratz: Good point. I was confused why code completion didn't offer this, but I only checked for `.get...()` methods :)

mickaelistria:
Code seems sane, but I don't have a strong opinion on the desired behavior and code strategy here. I defer the final decision to you and @fapdash ; I trust blindly you'll end up with something good.

sratz:
As mentioned in the issue: Merging this for now to fix the regression. We can improve this further if needed in a future release.

PR #550, opened 2023-02-03
Enhance deprecation notice for Navigator

akurtakov:
It's been deprecated since 2019 and could have been removed already but this will happen for 2022-06 (or at least part of it) thus make the deprecation annotation forRemoval=true.


vogella:
Great to see progress in deletion unmaintained stuff.

PR #557, opened 2023-02-08
Remove redundant superinterfaces

akurtakov:


merks:
Aren't these marked @generated?  Please do not change generated classes whatever you do will just be replaced again and if something badly needs attention, the generator should do it, not you by hand.

akurtakov:
Oops, I totally missed that. It is most likely that this code has been generated ages ago and that's why it has these warnings. Can you teach me how to regenerate it ? Last time I barely looked at emf was like 5+ years

merks:
I'll need to find a bit of time but it looks to me that it's modeled with these redundancies:

image

And that may be for a purpose because the first specified super class's Impl class is what's extended the that classes Impl...

Please open an issue and we can review what could or should be done there...

akurtakov:
Generated sources are no longer touched in this PR.

PR #560, opened 2023-02-08
Remove redundant super types from UIElements.ecore

### LINKED ISSUE #559: Regenerate sources in model.workbench from the model
As identified in https://github.com/eclipse-platform/eclipse.platform.ui/pull/557 I've tried to edit generated code. The goal here is to regenerate it in order to get these warnings fixed.

merks:
Regenerate with the latest EMF.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/559


merks:
@akurtakov  It looks to me like in all cases the redundancies in the generated code were present because it was modeled with those redundancies.  So I removed those from the model and regenerated...

akurtakov:
Cool, thanks. It's quite huge change in raw numbers +378 −700 :) Looks like EMF improved generation significantly since when this was generated.

laeubi:
Just wondering: If it is generated, why do we need to check-in generated code?

merks:
Most of the deletions come from this where the redundant inheritance forced it to regenerate everything already present in the "second" base classes.  

https://github.com/eclipse-platform/eclipse.platform.ui/pull/560/files#diff-268c7c85595a918478eb84f0d70d31d052e49718b31b8ded184142a16c3c50c7

So not actually improvement in the generator...

merks:
The generated code also contains hand-written logic that's preserved during merging.  Checking it in also ensures that when you check out from git, you don't need to run generators manually (or automatically with Oomph), you don't need to complicate the build with extensions from xtext for generating the model during the tycho build, and you'll have no surprises (and the need for version bumps) when using a newer version of the generator.

merks:
@akurtakov I think the test is a bit bogus because  it looks to be making an assumption about how the generator is implemented which is wrong, i.e., it tests if the first super type is an interface assume that's what's used as the base class resulting in lots of locally generated stuff for that interfaces feature.  But the general actually extends the first none interface class' implementation.  And the optimal choice is to extend the non-interfaces class with the most features results in the most reuse of that generated...

I'm going to look at improving the test itself.

merks:
Is the code analysis failure something I should worry about?

akurtakov:
> Is the code analysis failure something I should worry about?

It's complaining about the failing test too.

merks:
The failures are definitely bogus because we still have `PlaceholderImpl extends UIElementImpl` and `PartImpl extends UIElementImpl` just as before but now with the redundant extension of UIElement removed and before nothing noticed that PerspectiveStackImpl was very sub-optimal...

akurtakov:
This is the state of too many of the tests we have.

merks:
@akurtakov 

Looking at the implementation of EMF's generator, it will walk up chain of the first super type of the first super type and so on until it gets to a class which is not an interface and the generator will extend that class' generated implementation class, reusing the generated implementation of all that class' structural features.  This is only sub-optimal in the case that some later occurring super type would reuse a larger number of features.  If I write the test to test for this, it finds three cases that could be **very slightly** improved where this shows the class, what it now extends, how many features are reused, and what it could extend instead with the number of features it would reuse in that case.
```
ImperativeExpression->Expression:5->Contribution:7
HandledMenuItem->MenuItem:25->HandledItem:27
HandledToolItem->ToolItem:25->HandledItem:27
```
The difference is a very small fraction so I'm not sure I should make those changes....

Do you have an opinion?

akurtakov:
Unfortunately, I really have no idea what you speak about and even spending hours on that would not make me able to help the domain expert (aka you).

merks:
FYI, I implemented an improved test but filtered out those three classes from consideration.

merks:
Probably I need to "rebase"....

merks:
I didn't actually see anything that should have causes this failure:

> 12:58:17.369 [INFO] Command line:
> 	[/usr/lib/jvm/temurin-11-jdk-amd64/bin/java, -Xmx2048M, -Dp2.RepositoryPreferences.retryOnSocketTimeout=true, -Dp2.RepositoryPreferences.connectionRetryCount=3, -Dp2.RepositoryPreferences.connectionMsRetryDelay=500, -Dorg.eclipse.ecf.provider.filetransfer.retrieve.readTimeout=10000, -jar, /home/runner/.m2/repository/p2/osgi/bundle/org.eclipse.equinox.launcher/1.6.400.v20210924-0641/org.eclipse.equinox.launcher-1.6.400.v20210924-0641.jar, -install, /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/apiAnalyzer-workspace, -configuration, /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/apiAnalyzer-workspace/configuration, -data, /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/apiAnalyzer-workspace/data, -application, org.eclipse.pde.api.tools.apiAnalyzer, -project, /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.ui.views, -baseline, /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.ui.views/target/org.eclipse.ui.views-apiBaseline.target, -dependencyList, /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.ui.views/target/dependencies-list.txt, -failOnError]
> 1 API ERRORS
> [API ERROR] File MANIFEST.MF at line 5: The minor version should be incremented in version 3.11.300, since execution environments have been changed since version 3.11.300 (location: /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.ui.views/META-INF/MANIFEST.MF)

Hopefully it goes away.

akurtakov:
Thanks @merks !

PR #574, opened 2023-02-09
Enable the javadoc profile

laeubi:
FYI @akurtakov  Lets see how this works out...


akurtakov:
Seems to work ok. Merging.

laeubi:
At least it gives no error, but java 17 seem to have killed the javadoc... I'll try to investigate the next days whats going on here...

PR #596, opened 2023-02-10
Make use of Java 17 features in o.e.jface.tests.dialogs.StatusDialogTest

vrubezhny:
Also the test is renamed to StatusDialog_Test in order to fix file encoding.


merks:
Why is the test class renamed?  It seems not related to Java 17 and not a common naming pattern:

image

vrubezhny:
Any commit changing this file leads into appearing of non-readable chars in the contributor's name  in header comment.
This happens because the encoding used on file creation was different from utf8:

Screenshot from 2023-02-10 18-25-49

Also the file was created with "exectable" permission which is not a good thing for a Java source.
Just changing the file encoding/permissions doesn't make git to generate a mergeable change. 
The solution is to remove  the file from the repository and then create a new one, properly encoded, instead. So it was renamed in try to skip this file removal step. 

However it looks like Github Diff UI still has some problems, working with umlauts (`Ü`) and other characters like this:
 
Screenshot from 2023-02-10 19-04-58

laeubi:
Probably one could just use a different tool, renaming seems not suitable here..

vrubezhny:
The difference in encoding  can be seen here:

Original raw file: (ISO-8859-1) 
https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.ui/master/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/dialogs/StatusDialogTest.java

New raw file (utf8):
https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.ui/1fe62a4c1493cb259e1316b21385cc1cf31cd3be/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/dialogs/StatusDialog_Test.java

vrubezhny:
Updated the fix to contain two commits - one that removes the test and another one that re-adds it with the same name but correct utf8 encoding

akurtakov:
I've just added utf-8 character in https://github.com/eclipse-platform/eclipse.platform.ui/pull/597 committed and it is fine

vrubezhny:
Closing as replaced by #597  and #598

PR #575, opened 2023-02-09
Move o.e.jface.tests to Java 17

vrubezhny:

PR #597, opened 2023-02-10
Use utf-8 encoding in StatusDialogTest

akurtakov:


vrubezhny:
So, it's the same as the first one (in #575). Looking at the raw file and its encoding - it works. The problem is in Github diff view UI.

akurtakov:
Sounds like it

akurtakov:
I'll merge this one so we can see whether it's only the diff view showing problem.

akurtakov:
So it looks fine here https://github.com/eclipse-platform/eclipse.platform.ui/commit/ec9f76920367a75451d99cafedc5f032c46467f9 aka it's github bug we have been facing

PR #605, opened 2023-02-20
Removes .project filter from Gitigore

vogella:
Also adds the missing .project files for the migrated projects.


merks:
@vogella Thanks.  

I don't think we need to wait for a build here given that the build missed the fact these files were missing, the build will also miss the fact that they are now present.  I can't fix the Oomph setup until this is fixed...

PR #606, opened 2023-02-20
Ignore all /bin/ directories in repo

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.ui.tools/issues/31


merks:
No need for builds here right?  Save the planet. ☹️

PR #608, opened 2023-02-20
Removes org.eclipse.e4.tools.jdt.templates from build due to failing

vogella:
aggregator build

org.eclipse.e4.tools.jdt.templates has a dependency to JDT which fails the build after the repo merge. So removing the plug-in from the build for now to make the aggregator build work again.


vogella:
Build file  (without the tests) on my machine. Only awaiting the the server build

PR #621, opened 2023-02-28
Adding a PR template

vogella:
This will make the PR more informative, example:

image

### DIFF
--- a/.github/workflows/create-pullrequest.yml
+++ b/.github/workflows/create-pullrequest.yml
@@ -1,0 +1,19 @@
+name: Create or update pull request for non-master branches
+on:
+  push:
+    branches-ignore: master
+jobs:
+  action-pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 0
+      - name: Create or update pull request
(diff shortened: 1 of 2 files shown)


laeubi:
@vogella can you check if it is possible to put the template on the organization level?

.github/workflows/create-pullrequest.yml
```diff
@@ -0,0 +1,19 @@
+name: Create or update pull request for non-master branches
```
laeubi, line 1: Can you explain what this action does? This is not quite clear to me...

vogella: > Can you explain what this action does? This is not quite clear to me...

See here https://github.com/devops-infra/action-pull-request

laeubi: I must confess I don't understand this ... why is standard github Pullrequest not enough? This seems doing much more than "Adding a PR template" I certainly don't want that each and every branch is automatically created as a PR ... also all information added is already visible on the "Files changed" Tab already.

laeubi:
For example look at
- https://github.com/eclipse-jdt/.github/blob/main/pull_request_template.md
- https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/.github/ISSUE_TEMPLATE

that seems much more useful than this (from what I understand from the short description given).

Issue #614, opened 2023-02-24
E4 Command Icons


N1k145:
Currently, the Quick Access shows for Commands the Icons from the _CommandImageService_, which is great. But the only way I found for adding Icons to my commands is with the _org.eclipse.ui.commandImages_ extension point. 
Most of my commands are defined as E4 Commands, so when I reference them in my plugin.xml they are marked with a warning that the command is unknown, which is not nice.

My preferred solution would be to add the Icons in the fragment.e4xmi and read them also in the _CommandImagePersistence_, I would put them in the PersistedState Key Value Map ("Icon" -> Bundlepath).

Before I start developing a solution, I would like to get some input on my Idea or if somebody has a better idea all together.

PR #624, opened 2023-03-09
Enable the setting of command icons for e4 commands

N1k145:
Solves #614 

Previously it was only possible to define command icons using the extensions point org.eclipse.ui.commandImages
It was not possible to define images for e4 commands without creating a warning in the plugin.xml.
Now it is possible to define a command icon in the e4 application model by setting the "image" key in the persisted state of the command.

### DIFF
--- a/bundles/org.eclipse.e4.ui.model.workbench/model/UIElements.genmodel
+++ b/bundles/org.eclipse.e4.ui.model.workbench/model/UIElements.genmodel
@@ -67,6 +67,7 @@
         <genFeatures property="None" children="true" createChild="true" ecoreFeature="ecore:EReference UIElements.ecore#//commands/Command/parameters"/>
         <genFeatures notify="false" createChild="false" propertySortChoices="true"
             ecoreFeature="ecore:EReference UIElements.ecore#//commands/Command/category"/>
+        <genFeatures createChild="false" ecoreFeature="ecore:EAttribute UIElements.ecore#//commands/Command/commandIconURI"/>
         <genFeatures property="Readonly" createChild="false" ecoreFeature="ecore:EAttribute UIElements.ecore#//commands/Command/localizedCommandName"
             get="return &lt;%org.eclipse.e4.ui.model.LocalizationHelper%>.getLocalizedFeature(CommandsPackageImpl.Literals.COMMAND__COMMAND_NAME, this);"/>
         <genFeatures property="Readonly" createChild="false" ecoreFeature="ecore:EAttribute UIElements.ecore#//commands/Command/localizedDescription"
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
@@ -15,6 +15,7 @@
 
 package org.eclipse.ui.internal.commands;
 
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
@@ -25,6 +26,9 @@ import org.eclipse.core.runtime.IProduct;
 import org.eclipse.core.runtime.IRegistryChangeEvent;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.e4.ui.internal.workbench.E4Workbench;
+import org.eclipse.e4.ui.model.application.MApplication;
+import org.eclipse.e4.ui.model.application.commands.MCommand;
 import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.jface.resource.ResourceLocator;
 import org.eclipse.ui.IWorkbenchCommandConstants;
@@ -77,8 +81,6 @@ public final class CommandImagePersistence extends RegistryPersistence {
 	private static void readImagesFromRegistry(final IConfigurationElement[] configurationElements,
 			final int configurationElementCount, final CommandImageManager commandImageManager,
 			final ICommandService commandService) {
-		// Undefine all the previous images.
-		commandImageManager.clear();
 
 		final List<IStatus> warningsToLog = new ArrayList<>(1);
 
@@ -124,6 +126,36 @@ public final class CommandImagePersistence extends RegistryPersistence {
 
 		logWarnings(warningsToLog,
 				"Warnings while parsing the images from the 'org.eclipse.ui.commandImages' extension point."); //$NON-NLS-1$
+	}
+
+	private static void readImagesFromApplicationModel(final CommandImageManager commandImageManager) {
+
+		MApplication application = E4Workbench.getServiceContext().get(MApplication.class);
+
+		final List<IStatus> warningsToLog = new ArrayList<>();
+
+		for (MCommand command : application.getCommands()) {
+
+			if (command.getCommandIconURI() != null) {
+				try {
+					registerImage(command, CommandImageManager.TYPE_DEFAULT, commandImageManager,
+							command.getCommandIconURI());
+				} catch (MalformedURLException e) {
+					addWarning(warningsToLog, "Malformed Icon Url: " + e.getMessage(), null, command.getElementId(), //$NON-NLS-1$
+							"iconUri", //$NON-NLS-1$
+							command.getCommandIconURI());
+				}
+			}
+		}
+
+		logWarnings(warningsToLog, "Warnings while processing command images from persisted state"); //$NON-NLS-1$
+	}
+
+	private static void registerImage(MCommand command, int type, CommandImageManager commandImageManager,
+			String iconURI)
+			throws MalformedURLException {
+		ImageDescriptor imageDescriptor = ImageDescriptor.createFromURL(new URL(iconURI));
+		commandImageManager.bind(command.getElementId(), type, null, imageDescriptor);
 	}
 
 	/**
@@ -187,8 +219,14 @@ public final class CommandImagePersistence extends RegistryPersistence {
 			}
 		}
 
+		// Undefine all the previous images.
+		commandImageManager.clear();
+
 		readImagesFromRegistry(indexedConfigurationElements[INDEX_IMAGES], imageCount, commandImageManager,
 				commandService);
+
+		readImagesFromApplicationModel(commandImageManager);
+
 		// Associate product icon to About command
 		IProduct product = Platform.getProduct();
 		if (product != null) {
--- a/bundles/org.eclipse.e4.ui.model.workbench/model/UIElements.ecore
+++ b/bundles/org.eclipse.e4.ui.model.workbench/model/UIElements.ecore
@@ -208,6 +208,11 @@
           <details key="documentation" value="&lt;p>&#xD;&#xA;&lt;strong>Developers&lt;/strong>:&#xD;&#xA;Add more detailed documentation by editing this comment in&#xD;&#xA;org.eclipse.ui.model.workbench/model/UIElements.ecore.&#xD;&#xA;There is a GenModel/documentation node under each type and attribute.&#xD;&#xA;&lt;/p>"/>
         </eAnnotations>
       </eStructuralFeatures>
+      <eStructuralFeatures xsi:type="ecore:EAttribute" name="commandIconURI" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
+        <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
+          <details key="documentation" value="&lt;p>&#xD;&#xA;This field contains a fully qualified URL defining the path to an Image to display&#xD;&#xA;for this element.&#xD;&#xA;&lt;/p>&#xD;&#xA;@since 2.4"/>
+        </eAnnotations>
+      </eStructuralFeatures>
       <eStructuralFeatures xsi:type="ecore:EAttribute" name="localizedCommandName"
           eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"
           changeable="false" volatile="true" transient="true" derived="true">
--- a/bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/application/commands/MCommand.java
+++ b/bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/application/commands/MCommand.java
@@ -39,6 +39,7 @@ import org.eclipse.e4.ui.model.application.ui.MLocalizable;
  *   <li>{@link org.eclipse.e4.ui.model.application.commands.MCommand#getDescription <em>Description</em>}</li>
  *   <li>{@link org.eclipse.e4.ui.model.application.commands.MCommand#getParameters <em>Parameters</em>}</li>
  *   <li>{@link org.eclipse.e4.ui.model.application.commands.MCommand#getCategory <em>Category</em>}</li>
+ *   <li>{@link org.eclipse.e4.ui.model.application.commands.MCommand#getCommandIconURI <em>Command Icon URI</em>}</li>
  *   <li>{@link org.eclipse.e4.ui.model.application.commands.MCommand#getLocalizedCommandName <em>Localized Command Name</em>}</li>
  *   <li>{@link org.eclipse.e4.ui.model.application.commands.MCommand#getLocalizedDescription <em>Localized Description</em>}</li>
  * </ul>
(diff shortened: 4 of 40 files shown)


N1k145:
From my side, everything should be ready for a review now. Feel free to have a look

vogella:
Could you enhance the commit message with information: what is missing and what works after this change.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
```diff
@@ -126,6 +132,34 @@ private static void readImagesFromRegistry(final IConfigurationElement[] configu
 				"Warnings while parsing the images from the 'org.eclipse.ui.commandImages' extension point."); //$NON-NLS-1$
 	}
 
+	private static void readImagesFromApplicationModel(final CommandImageManager commandImageManager) {
+
+		MApplication application = E4Workbench.getServiceContext().get(MApplication.class);
+
+		for (MCommand command : application.getCommands()) {
+
+			if (command.getPersistedState().containsKey(E4_COMMAND_IMAGE_KEY)) {
```
laeubi: It seems other parts of the code using an dedicated `iconUri` (with a dialog to search / select icons) is the persisted state really the right place to do this?
I suspect that this way it might be possible to set an icon but hardly anyone will find / use that?!

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
```diff
@@ -126,6 +132,34 @@ private static void readImagesFromRegistry(final IConfigurationElement[] configu
...
+		}
+	}
+
+	private static void registerImage(MCommand command, String key, int type, CommandImageManager commandImageManager) {
+		String iconURI = command.getPersistedState().get(key);
+		try {
+			ImageDescriptor imageDescriptor = ImageDescriptor.createFromURL(new URL(iconURI));
+			commandImageManager.bind(command.getElementId(), type, null, imageDescriptor);
+		} catch (MalformedURLException e) {
+			e.printStackTrace();
```
laeubi: This should be written to the log.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
N1k145: The dedicated `iconUri ` field would require a change to the Application Model, and to be honest I don't know how to do this and if there is a process to discuss such a change, 
The Persisted State is maybe not the best place to put it is already used for some special features like this.

laeubi: > The dedicated iconUri field would require a change to the Application Model, and to be honest I don't know how to do this and if there is a process to discuss such a change,

@vogella or @merks can help with the model change, I think this is the right way to solve this, the persisted state seems not an appropriate choice here:

1. not visible to users so most likely no one will ever discover your new feature
2. no user assistance in available fields
3. no validation
4. not consistent across other usages for a similar feature (menus, parts, ...) where icons are provided with E4

N1k145: I agree with you that it is not the right place, i will have a look into the model change.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
```diff
@@ -77,8 +85,6 @@ public final class CommandImagePersistence extends RegistryPersistence {
 	private static void readImagesFromRegistry(final IConfigurationElement[] configurationElements,
 			final int configurationElementCount, final CommandImageManager commandImageManager,
 			final ICommandService commandService) {
-		// Undefine all the previous images.
-		commandImageManager.clear();
```
merks, line 81: The method can be called multiple times :

image

So I'm not sure that removing this is okay nor why it was done.

N1k145: I did not remove it, I just moved it up to the method that calls both readImages functions. 
So that it makes more sense from a flow perspective.

Previously it was:
read() -> readImagesFromRegistry() -> clear()

Now it is:
read() -> clear()
read() -> readImagesFromRegistry()
read() ->  readImagesFromApplicationModel()

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
merks: I can help with changing the model, but I did not design the model, so I don't know the proper design for this. It would seem like three new attributes are needed, **but** I don't see anything similar to that in the model already.  E.g., I don't know how toolbar items get their icon from the model in a pure e4 application.  How does that work currently?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
merks: Sorry for misreading.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
N1k145: Currently, the icon for a toolbar item or Part is set as an attribute of MUILabel from which the Part or ToolItem inherits. 
As MUILabel provides more functionality than needed, I tried to add an attribute directly to MCommand. Which looks fine,

The only issue I have now is that the code generation generates a newer style than the current code, which creates more changes than necessary. 
Should I commit the whole changes, or strip it down to only the relevant ones?

merks: Yes, that's the only related thing that I saw, but there I did not see other things one might expect from your changes, e.g., a disabled icon or a hover icon.  That that makes me wonder why are those needed for a command but not needed elsewhere in the UI model where I would assume the same kinds of needs arise.

N1k145: They are usually not needed, there is an option to set those in the transient data of an Element (that is really well hidden).
That is optional and only needed when the generation of the disabled and hover icon should be overridden.

I added them for feature completeness to e3. But for the solution with the application model, I would keep them in the persisted state and move only the main icon to an attribute and add the option to the E4 Model Editor.

merks: I see!  

So if you change the model to add a new attribute, you'd normally increment the minor version, but that's already been done for the 4.28 release cycle.   You'd add something that looks like this in the UIElement.ecore:
```
      <eStructuralFeatures xsi:type="ecore:EAttribute" name="iconURI" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString">
        <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
          <details key="documentation" value="&lt;p>&#xD;&#xA;This field contains a fully qualified URL defining the path to an Image to display&#xD;&#xA;for this element.&#xD;&#xA;&lt;/p>&#xD;&#xA;@since 2.4.0"/>
        </eAnnotations>
      </eStructuralFeatures>
```
Note the @since 2.4.0 which will be generated into all the parts of the code that change.  When you open the *.genmodel, it will be dirty and will need to be saved.  You can generate the code from there:

image

Of course you have to have EMF's generator installed, but the Oomph setup includes that already.

If you need other help, just ask.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
N1k145: thanks, i moved it to the logger

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
N1k145: @merks @laeubi I updated the E4 Model with an attribute for the iconUrl and extended the Model Editor to edit the field.

bundles/org.eclipse.e4.ui.model.workbench/model/UIElements.genmodel
```diff
@@ -339,7 +340,4 @@
       </nestedGenPackages>
     </nestedGenPackages>
   </genPackages>
-  <codeStyle>UnnecessaryAssignmentBeforeReturn</codeStyle>
```
merks: Why are these removed?  This will cause major style changes in the generate code.  Many of these options were developed purely because the e4 folks wanted these styles.  It will of course be easier to review when there are fewer changed files...

N1k145: No idea, I did not make this change intentionally, i will revert this

N1k145: That also explained most of the changes in the other model parts. I pushed the reverted version now

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
merks: The model changes look fine now.

merks:
I hope someone like @vogella or @opcoach with a more in-depth knowledge of the model design will also review.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
```diff
@@ -126,6 +131,58 @@ private static void readImagesFromRegistry(final IConfigurationElement[] configu
...
+			if (command.getIconURI() != null) {
+				try {
+					registerImage(command, CommandImageManager.TYPE_DEFAULT, commandImageManager, command.getIconURI());
+				} catch (MalformedURLException e) {
+					addWarning(warningsToLog, "Malformed Icon Url: " + e.getMessage(), null, command.getElementId(), //$NON-NLS-1$
+							"iconUri", //$NON-NLS-1$
+							command.getIconURI());
+				}
+			}
+			if (command.getPersistedState().containsKey(E4_COMMAND_IMAGE_KEY_DISABLED)) {
```
laeubi: i wonder if that should also be part of the model then... I always missed specifying disabled/hover/pressed/... in other places as well!

laeubi:
I think this goes in the right direction and looks very promising!

Just one thing, I see a lot of style changes (e.g. remove redundant generic specifications), you might want to submit a separate PR and exclude them from this one so we get a smaller change set to review.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
N1k145: It's not directly part of the model, these are still handled over the persisted state, this is also possible for Tool Items and only necessary as a fallback when the automatic generation is not creating the desired result. 

If you want, I can remove the functionality, but i would like to keep it to make it feature complete to e3

N1k145:
> Just one thing, I see a lot of style changes (e.g. remove redundant generic specifications), you might want to submit a separate PR and exclude them from this one so we get a smaller change set to review.

They come from the regenerating of the model, if you like i can run the generation once using the current generators, create a PR, that is blocking this one, change the model and run the generation again to get a cleaner diff.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
laeubi: if it is in e3 we should also have it in e4, eg. `iconUriHover` and `iconUriDisabled`, maybe it would even be usefull to have these in a base model-class ... I think also menuitems and toolbar buttons can then get this in a more direct way.

merks:
The failure

>  Error:  Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.ui.tests: Only qualifier changed for (org.eclipse.e4.ui.tests/0.15.0.v20230327-0814). Expected to have bigger x.y.z than what is available in baseline (0.15.0.v20230220-1021) -> [Help 1]

means you should increment the version of /org.eclipse.e4.ui.tests/META-INF/MANIFEST.MF I think to version 0.15.100.

merks:
> > Just one thing, I see a lot of style changes (e.g. remove redundant generic specifications), you might want to submit a separate PR and exclude them from this one so we get a smaller change set to review.
> 
> They come from the regenerating of the model, if you like i can run the generation once using the current generators, create a PR, that is blocking this one, change the model and run the generation again to get a cleaner diff.

Yes, there are automatic cleanup actions applied in the modified files.   All the generated changes look fine and reasonable to me.  In the end, the only thing that needs review is the getter and setter of the interface.  The rest is not of actual "review" interest.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/commands/CommandImagePersistence.java
N1k145: I think this should then not be part of this pr and discussed separately.
This should then also be handled differently in the E4 Model Editor to make it clear to the user that this is optional

sratz: When exactly are `E4_COMMAND_IMAGE_KEY_DISABLED` / `E4_COMMAND_IMAGE_KEY_HOVER` set in the current implementation?

I agree with @laeubi : For symmetry with https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/extension-points/org_eclipse_ui_commandImages.html?cp=2_1_1_144

couldn't we instead simply also add the attributes `commandIconDisabledURI` and `commandIconHoverURI` directly to the model right now?

N1k145: @sratz
They are added manually in the persisted state.
Yes we could add them right now, but afaik there is also a E3 way to add disabled images to toolbar/menu items https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fextension-points%2Forg_eclipse_ui_menus.html&cp%3D2_1_1_144
which are also not directly supported in the E4 Model, only using the transient data.

I would like to separate those topics, either remove the disabled icon for now from this PR or leave them as is. And have a separate discussion on the disabled icons, if we want to have them in the E4 Model, how this should be handled in the editor and in the implementations.
For most cases the disabled icons are not relevant as they are generated at runtime when needed.

laeubi: I think the best is to split this up and start with the bare minimum (a single icon) and then go for disabled / hover in a separate PR (that probably will fix this for Toolbars as well).

sratz: > They are added manually in the persisted state.

I'm not sure I fully understand then (but I am not that familiar with E4 so bear with me): Who adds them manually? I see no code that _writes_ those attributes, only _reads_. Would that mean those constants are effectively API for consumer code?

> I think the best is to split this up and start with the bare minimum (a single icon) and then go for disabled / hover in a separate PR (that probably will fix this for Toolbars as well).

Sounds reasonable.

N1k145: @sratz basically yes, the E4 Model contains a Key Value Map for each element, and those constants would then be a key for that map. The consumer could add them in code or in their fragment.xmi file.

@laeubi @sratz I removed the disabled/hover image for now.

laeubi (approved):
This looks good to me, it would be good to get a second review and then merge this before M1 @vogella @merks @sratz

merks:
The changes look extremely well done, with appropriate tests, and good changes to the editor. 

I personally think we should merge this in time for tomorrow's M1 build.

vogella:
LGTM as well thanks @N1k145 Lets merge.

@N1k145 are you planning to adjust the e4 Model Editor as well?

vogella:
Also an entry for the N&N would be very nice. See https://github.com/eclipse-platform/eclipse.platform.common/pull/130 for an example

merks:
@vogella  Is the model editor different from the CommandEditor?

https://github.com/eclipse-platform/eclipse.platform.ui/pull/624/files#diff-bd7f6449f0c0292c2e683816bc987952dec6f086f0485f50eb0920940858b38d

vogella:
> @vogella Is the model editor different from the CommandEditor?
> 
> https://github.com/eclipse-platform/eclipse.platform.ui/pull/624/files#diff-bd7f6449f0c0292c2e683816bc987952dec6f086f0485f50eb0920940858b38d

Ah, missed that. In my head the model editor still lift in its own repo. Sorry for the noise.

PR #630, opened 2023-03-12
Disable theming by default on Linux/GTK

### LINKED ISSUE #629: Disable theming by default in Linux distributions?
Currently, on Linux, most contributors seem to either use dark theme or use no theme. There seems to be a relative consensus that the light theme on Linux is not better than no theme. So I would recommend that -on Linux only- we just disable theming by default.
Thoughts?

mickaelistria:
The light theme shows no clear value over no theme on Linux, so let's switch to no theme by default on Linux.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/build.properties
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/build.properties
@@ -17,5 +17,6 @@ bin.includes = META-INF/,\
                .,\
                about.html,\
                plugin.properties,\
-               icons/
+               icons/,\
+               plugin.xml
 src.includes = about.html
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/plugin.xml
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/plugin.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.4"?>
+<plugin>
+   <extension
+         point="org.eclipse.core.runtime.preferences">
+      <initializer
+            class="org.eclipse.e4.ui.internal.workbench.renderers.swt.ThemePreferenceInitializer">
+      </initializer>
+   </extension>
+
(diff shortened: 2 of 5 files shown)


mickaelistria:
@iloveeclipse @akurtakov should we merge this milestone?

akurtakov:
I don't mind it personally.

bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/workbench/swt/internal/ThemePreferenceInitializer.java
```diff
@@ -0,0 +1,31 @@
...
+import org.eclipse.core.runtime.preferences.DefaultScope;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine;
+import org.eclipse.e4.ui.internal.workbench.swt.WorkbenchSWTActivator;
+
+public class ThemePreferenceInitializer extends AbstractPreferenceInitializer {
+
+	@Override
+	public void initializeDefaultPreferences() {
+		IEclipsePreferences preferences = DefaultScope.INSTANCE
```
iloveeclipse: Please test this with -pluginCustomization /path/to/pluginCustomization.ini and check if the product settings will or will not be considered here.

Product defaults from ini should take precedence.

mickaelistria:
I've updated the patch to properly try it (my previous tests were actually faulty and giving a false impression it worked...); and then with this newer patch, I can verify that the settings from plugin_customization.ini take higher priority.

mickaelistria:
For the moment, I'm a bit discouraged by the test failures and test assuming that the IDE is started with theme enabled. I don't plan to work on this in short-term. So if anyone is interested in continuing it before I come back to it (can't tell when), then thank you, be my guest!

akurtakov:
As @mickaelistria doesn't plan to continue on this one I'm closing it.

PR #634, opened 2023-03-15
Propagate the exception that make the image creation fail

laeubi:
Currently if image creation fails, only a very generic "Unable to create resource" exception is thrown and the caller has no clue what is actually going on.

This changes the code that the original exception is propagated as the cause.

See https://github.com/eclipse-platform/eclipse.platform.releng/issues/210

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
@@ -227,11 +227,23 @@ public abstract class ImageDescriptor extends DeviceResourceDescriptor {
 
 	@Override
 	public Object createResource(Device device) throws DeviceResourceException {
-		Image result = createImage(false, device);
-		if (result == null) {
-			throw new DeviceResourceException(this);
+		try {
+			return new Image(device, new ImageDataProvider() {
+
+				@Override
+				public ImageData getImageData(int zoom) {
+					ImageData imageData = ImageDescriptor.this.getImageData(zoom);
+					if (imageData == null && zoom == 100) {
+						throw new IllegalStateException(String.format(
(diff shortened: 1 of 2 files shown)


laeubi:
@iloveeclipse it seems not be a problem of the build, because now test also fail with:

```
Caused by: java.lang.IllegalArgumentException: Argument not valid: ImageDataProvider [org.eclipse.jface.resource.ImageDescriptor$$Lambda$286/0x0000000800f668e0@64420e34] returns null ImageData at 100% zoom.
	at org.eclipse.swt.SWT.error(SWT.java:4899)
	at org.eclipse.swt.internal.DPIUtil.validateAndGetImageDataAtZoom(DPIUtil.java:412)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:659)
	at org.eclipse.jface.resource.ImageDescriptor.createResource(ImageDescriptor.java:231)
	... 90 more
```

so it seems something has changed in JFace or SWT?

akurtakov:
If I'm to guess the new containers built by  https://www.eclipse.org/lists/platform-releng-dev/msg39195.html are broken somehow.

laeubi:
I now enhanced the error reporting even further lets see if that reveals something...

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -227,11 +227,23 @@ public ImageDescriptor imageDescriptorFromURI(URI uriIconPath) {
 
 	@Override
 	public Object createResource(Device device) throws DeviceResourceException {
-		Image result = createImage(false, device);
```
jukzi, line 230: createImage() is not final so it can not be inlined.

jukzi (changes requested):

Issue #633, opened 2023-03-13
Use empty editor area for guiding texts


marcushoepfner:
Bringing over bug from bugzilla...

How VScode does it
<img width="1243" alt="Screenshot 2019-04-04 at 16 07 11" src="https://user-images.githubusercontent.com/29301444/224765927-210952c4-af8d-4279-bc0e-5c3af1d81afa.png">

When no editors are open the user currently sees a big empty space on his screen.
We could use this area to provide some guiding / onboarding text.
See how Visual Studio code does it.

We could also show the product's icon there. The question is what exactly would we show there? This list must somehow be extendable by the installed plugins.

And from a technical point of view: 
- How can we draw into that area? Do we need an SWT enhancement here?

I think this would nicely complement the work done in (https://bugs.eclipse.org/bugs/show_bug.cgi?id=543746.

marcushoepfner:
#78 was merged and added functionality to show some text in the empty editor area.

Let's discuss how we can progress.

The grey image of the perspective and the shortcuts should be provided by a perspective and maybe a text.
Shall we use 

1. the existing extension point `org.eclipse.ui.perspective` and enhance it (by the fields for image, shortcuts and text) or
2. a new extension point?

trancexpress:
Isn't the idea of the welcome screen to do this?

marcushoepfner:
This is related to the perspective while the welcome screen is more a general introduction. 
Furthermore you may not know what to do when you see the empty editor area the first time.
Feature has kind of been accepted already, refer to the original bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=546115

LiamHD:
> Isn't the idea of the welcome screen to do this?

No. The welcome page is to actively inform user about news etc. The user can interact with the welcome page by e.g. clicking links etc.

This here is a more passive / not so intrusive way to provide some hints to the user. It's rendered like a watermark and the user cannot interact (e.g. by clicking links) with this.

Compare with VScode - they also have this watermark in the empty editor area and they also have welcome pages (e.g. after you install / update extensions).

LiamHD:
> 1. the existing extension point `org.eclipse.ui.perspective` and enhance it (by the fields for image, shortcuts and text)

AFAIK this extension point today is only read when you switch to a perspective. If you just e.g. restart your IDE then this is not read. On restart platform simply restores the layout.

BeckerWdf:
[...]

PR #637, opened 2023-03-15
Prototype: show welcome text

marcushoepfner:
Issue #633

The text is read from perspective extension point which has been extended.
The sample plugin is not fully prepared (e.g. pom.xml and so on). It is work in progress and should just show the current state.

It is yet unclear, how to add shortcuts and image to the perspective model.


marcushoepfner:
@vogella the "welcome text" could be put into tags easily. But what do I do with the image and shortcuts which are to be added to the extension point.
Do I need to extend MPerspective with new properties? Any Idea?

vogella:
I assume this one is obsolete now @marcushoepfner ?

PR #640, opened 2023-03-15
Add support for Microsoft Edge in Mac OS.

elsazac:
In Eclipse->Settings->General->Web Browser

1.As cleanup removing browsers Camino and Internet Explorer, as it is no longer supported in Mac OS. 
2.Adding support for Microsoft Edge in Mac OS.

/cc @lshanmug

### DIFF
--- a/bundles/org.eclipse.ui.browser/plugin.xml
+++ b/bundles/org.eclipse.ui.browser/plugin.xml
@@ -201,13 +201,13 @@
          executable="Google Chrome">
          <location>Applications/Google Chrome.app</location>
       </browser>
-      <browser
+<!--  <browser
          id="org.eclipse.ui.browser.camino"
          name="%browserCamino"
          os="MacOSX"
          executable="Camino">
          <location>Applications/Camino.app/Contents/MacOS/Camino</location>
-      </browser>
+      </browser>   -->
       <browser
          id="org.eclipse.ui.browser.safari"
          name="%browserSafari"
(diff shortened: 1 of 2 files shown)


vogella:
@BeckerWdf can you test?

BeckerWdf:
> @BeckerWdf can you test?

Sure.
I just installed Edge on my Mac. :see_no_evil:

I created a project with a simple html file. Choosing "open > web browser" from the context menu opens edge when "edge" is selected in the prefereces. It openes "chrome" when "chrom" is selected in the preferences.
The "Browser" view in contrast does not change behaviour. It always uses "safari" on macos. But I think this is how this always worked.

So function wise this looks good to me.

bundles/org.eclipse.ui.browser/plugin.xml
```diff
@@ -201,13 +201,13 @@
          executable="Google Chrome">
          <location>Applications/Google Chrome.app</location>
       </browser>
-      <browser
+<!--  <browser
```
BeckerWdf: I would prefer that you delete this instead of commenting it out.

BeckerWdf (changes requested):

bundles/org.eclipse.ui.browser/plugin.xml
```diff
@@ -216,12 +216,19 @@
          factoryclass="org.eclipse.ui.internal.browser.macosx.SafariBrowserFactory">
          <location>Applications/Safari.app</location>
       </browser>
-      <browser
+<!--  <browser
```
BeckerWdf: same here: pls. delete.
In addition. Also delete browserInternetExplorer and browserCamino in plugin.properties

elsazac: @BeckerWdf Deleted both camino and Internet Explorer.

elsazac: @BeckerWdf Deleted both from Plugin.properties and Plugin.xml.Can you have a look at it?Let me know if anything else is required.

BeckerWdf (approved):

vogella:
AFAICS the tests errors are unrelated and covered by https://github.com/eclipse-platform/eclipse.platform.ui/issues/665. Merging based on @BeckerWdf review.

Thanks [elsazac](https://github.com/elsazac)

PR #652, opened 2023-03-21
Migrate to the unified workflow

laeubi:


laeubi:
Merging this to track progress, linux has a test-failure (flapping test?), windows was never tested, same for macos.

PR #655, opened 2023-03-21
Do not open key assist dialog again if already open for the same matches

### LINKED ISSUE #654: Re-opening of key assist dialog leads to flickering and delays
As already mentioned in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/443 the opening of the popup for key binding conflicts can lead to flickering and delays.

During the search for matching [mnemonics](https://en.wikipedia.org/wiki/Mnemonics_(keyboard)) the installed event listener `KeyBindingDispatcher` maybe invoked for almost every widget when the `Alt`-key is involved:

```java
// in class Control
boolean translateMnemonic (Event event, Control control) {
	if (control == this) return false;
	if (!isVisible () || !isEnabled ()) return false;
	event.doit = this == display.mnemonicControl || mnemonicMatch (event.character);
	return traverse (event);
}

// in class Composite
boolean translateMnemonic (Event event, Control control) {
	if (super.translateMnemonic (event, control)) return true;
	if (control != null) {
		Control [] children = _getChildren ();
		for (int i=0; i<children.length; i++) {
			Control child = children [i];
			if (child.translateMnemonic (event, control)) return true;
		}
	}
	return false;
}

// in class Control
boolean traverse (Event event) {
	/*
	* It is possible (but unlikely), that application
	* code could have disposed the widget in the traverse
	* event.  If this happens, return true to stop further
	* event processing.
	*/
	sendEvent (SWT.Traverse, event);
        // [...]
}
```

On every of these multiple invocations for a _single key press event_ the key assist dialog is closed and re-opened. The constantly switching focus between the top level window and the key assist window leads to the observed flickering effect.

On the Linux platform this is less likely than on Windows because in the SWT GTK implementation some key combinations do not trigger a search for mnemonics. But combinations with `Alt` and a simple character are also with SWT GTK a problem. On my Linux system with GTK3 the final key assist popup is also wrongly positioned top left instead of bottom right (like on Windows).

oleosterhagen:
Fixes #654 

While searching for matching mnemonics the key binding dispatcher may be invoked for many widgets and their children. Closing and re-opening the key binding conflict dialog on every invocation otherwise leads to flickering, delays and sometimes to a wrong positioning of the window.

### DIFF
--- a/bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/internal/KeyAssistDialog.java
+++ b/bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/internal/KeyAssistDialog.java
@@ -16,6 +16,7 @@ package org.eclipse.e4.ui.bindings.internal;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.TreeSet;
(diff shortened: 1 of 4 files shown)


mickaelistria:
Thank you!

PR #656, opened 2023-03-22
Add onboarding properties to perspectiveExtension ext.point

marcushoepfner:
#633

Adds onboarding text, onboarding image and onboarding commands to the
extension point.
The properties are read in PerspectiveExtensionReader and added to the
e4 model in ModeledPageLayout.

Properties are used in StackRenderer. Also an
example plugin will be added in a separate commit.


vogella:
IMHO it would be nice if that new functionality would be provided by a new plug-in with could also be used in e4 RCP applications and not be added to org.eclipse.ui.  Would that be possible @marcushoepfner ?

marcushoepfner:
> IMHO it would be nice if that new functionality would be provided by a new plug-in with could also be used in e4 RCP applications and not be added to org.eclipse.ui. Would that be possible @marcushoepfner ?

I don't know how.
Would it be possible merge this first and refactor later? Maybe hard because the extension point is already used then?!

vogella:
> Properties will be used in StackRenderer on a separate commit. Also an
> example plugin will be added.

Is the followup commit already avialable? (Properties will be used in StackRenderer on a separate commit. Also an
example plugin will be added.)

marcushoepfner:
> > Properties will be used in StackRenderer on a separate commit. Also an
> > example plugin will be added.
> 
> Is the followup commit already avialable? (Properties will be used in StackRenderer on a separate commit. Also an example plugin will be added.)

I'll create this now. Regarding the dependency I was wondering whether implementing `IPerspectiveFactory` in the RCP, like `org.eclipse.debug.internal.ui.DebugPerspectiveFactory` does it. As far as I understood, this is the programmatic approach for the extension point.
But unfort. IPerspectiveFactory is also located in o.e.ui.

marcushoepfner:
I have added the StackRenderer to this PR, as it belongs here.
One could see how I get the information from the `tag` model element in MPerspective.
Not sure whether this is the right approach or whether it is common to enhance the workbench emf model for such a case.

BTW: filling the onboarding information to MPerspective from an RCP should work right?

marcushoepfner:
Now there are some other errors that I don't understand:
https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/4503243004/jobs/7926183379?pr=656

marcushoepfner:
For bundle `Eclipse e4 Workbench SWT Renderer` these warnings are stated:

`Warning:  Problems resolving provisioning plan.: [Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet [2.1.0,3.0.0).; Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet.http [2.1.0,3.0.0).; Unable to satisfy dependency from org.osgi.service.http.whiteboard 1.1.1.202109301733 to osgi.contract; (&(osgi.contract=JavaServlet)(version=3.1.0)).]`

Is that the reason?

marcushoepfner:
Trying to fix the build errors first. Let you know when ready for review.

vogella:
I plan to test and review tomorrow morning (Wednesday). Sorry for the delay @marcushoepfner, having this help support available in the IDE is the "Geschnitten Brot" of 2023 for me.

marcushoepfner:
Ready for review.
Build errors are related to #665.

Example plugin can be found here: #657

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IPageLayout.java
```diff
@@ -577,4 +577,30 @@ public interface IPageLayout {
...
+	void setEditorOnboardingText(String text);
+
+	/**
+	 * Sets the onboarding image. The image is shown in the editor area in case no
+	 * editor is open.
+	 *
+	 * @param imageUri the uri of the onboarding image
+	 * @since 3.129
+	 */
+	void setEditorOnboardingImage(String imageUri);
```
BeckerWdf: "setEditorOnboardingImageUri" seems a better name. Because you don't set an Image, but "only" it's Uri.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IPageLayout.java
```diff
@@ -577,4 +577,30 @@ public interface IPageLayout {
...
+
+	/**
+	 * Adds a command for the empty editor area.
+	 *
+	 * @param commandId an id of an onboarding command
+	 * @since 3.129
+	 */
+	void addEditorOnboardingCommand(String commandId);
```
BeckerWdf: is "addEditorOnboardingCommandId" a better name?

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
```diff
@@ -88,6 +89,27 @@ are appended to the initial definition.
...
+               </documentation>
+               <appInfo>
+                  <meta.attribute translatable="true"/>
+               </appInfo>
+            </annotation>
+         </attribute>
+         <attribute name="editorOnboardingImage" type="string">
+            <annotation>
+               <documentation>
+                  The image is shown in the editor area in case no editor is open.
```
BeckerWdf, line 105: Maybe we should also say how big (in pixels) the icon should be.
Maybe we should even adapt the Eclipse UI Guideline

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
```diff
@@ -308,6 +330,26 @@ If this attribute is not used, the view will be moveable, unless the perspective
       </complexType>
    </element>
 
+   <element name="editorOnboardingCommand">
```
BeckerWdf, line 334: Shouldn't we also limit the number of commands. I think an onboarding text with e.g. 20 command is not longer so helpful.
I think we should stick to e.g. max. of 4 or 5 commands.

BeckerWdf:
I tested with the example plugin. I see two issues:
1.) The icon is blurry on HiRes Displays
2.) Keybinding (on macOS) does say "COMMAND" instead of showing the "Command-Icon" as we see them in the menu entries.

See: 
image

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/e4/compatibility/ModeledPageLayout.java
```diff
@@ -615,4 +623,24 @@ public void stackView(String id, String refId, boolean visible) {
...
+
+	@Override
+	public void setEditorOnboardingImage(String iconUri) {
+		perspModel.getTags().add(EDITOR_ONBOARDING_IMAGE + iconUri);
+	}
+
+	@Override
+	public void addEditorOnboardingCommand(String commandId) {
+		Predicate<MKeyBinding> commandWithEqualId = b -> b.getCommand().getElementId().equals(commandId);
+		Function<MKeyBinding, String> toCommandText = b -> b.getCommand().getCommandName() + ": " + b.getKeySequence(); //$NON-NLS-1$
```
BeckerWdf: Instead of
b.getCommand().getCommandName() + ": " + b.getKeySequence()
you should use something like:
```suggestion
		b.getCommand().getCommandName() + ": " //$NON-NLS-1$
						+ org.eclipse.jface.bindings.keys.KeySequence.getInstance(b.getKeySequence()).format();
```

to get it formatted correctly on macOS

BeckerWdf (changes requested):

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
marcushoepfner: which size would you propose?

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
marcushoepfner: went for 5

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
BeckerWdf: I would play around with this an compare with VScode. How big is it in VScode.

marcushoepfner:
> I tested with the example plugin. I see two issues: 1.) The icon is blurry on HiRes Displays 2.) Keybinding (on macOS) does say "COMMAND" instead of showing the "Command-Icon" as we see them in the menu entries.
> 
> See: image

Yes. I am not an image expert. I just took the small icon and resized it. We should change of course. But that belongs to PR #657

BeckerWdf:
> The icon is blurry on HiRes Displays

This is due to the fact that there is no "default_persp_grey@2x.png" in the example. Can you pls. add that?

BeckerWdf:
> But that belongs to PR #657

Correct. I added it also there.

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
BeckerWdf: > went for 5

Where do you process this in code?

marcushoepfner: ModelPageLayout

https://github.com/eclipse-platform/eclipse.platform.ui/pull/656/files#diff-8ad222613d61876e357a83c18a1d930d92b0a5ccfe03d6c9c2686d26441eefd2R642

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/PerspectiveExtensionReader.java
```diff
@@ -103,6 +106,29 @@ private boolean processActionSet(IConfigurationElement element) {
...
+				break;
+			case IWorkbenchRegistryConstants.ATT_TARGET_ID:
+				result = true;
+				break;
+
+			default:
+				break;
+			}
+			if (!result) {
+				WorkbenchPlugin.log("Unable to process attribute: " + //$NON-NLS-1$
```
vogella: Minor comment: For new code I think it better to use ILog.error()

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
BeckerWdf: I would say 350x350 looks good.
I can provide a SVG and PNGs of that size if you like.

marcushoepfner: VS code has 280x280. I like it with this size.
I found the SVG in https://github.com/eclipse-platform/eclipse.platform.images and I have created a grey images with 280x280.

I think I don't need to craete the svg as it is alraedy in https://github.com/eclipse-platform/eclipse.platform.images. 
Do I need a second png? with 560x560 then?

vogella (approved):
LGTM, I expecially like how this is decoupled between e3 and e4. Once the concerns from @BeckerWdf are resolved, I suggest to merge this so that we can start using it in the IDE so that we can see this in action.

It surprised me that the CTabFolder paint listener is triggered three times if I give focus to the appliation but this comment is independent of this PR. It just seems that CTabFolder likes to draw itself multiple times.

vogella:
If possible we should left align the commands similar to VS Code but that could be done in a follow-up PR

image

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
marcushoepfner: added images in 350x350 and double size.

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
BeckerWdf: Thanks. We should also document this limitation in the extension point docu.

BeckerWdf:
This is how it looks like compared to VScode:
image

1. I agree with Lars: We should align it the same way as in VScode
2. For my feeling the text could be a bit bigger. What does control the size of the text?
3. For my feeling the space between the icon and the text is to big. The text and the list of command should move up to be closer to the icon.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/PerspectiveExtensionReader.java
marcushoepfner: changed it.

marcushoepfner:
> If possible we should left align the commands similar to VS Code but that could be done in a follow-up PR
> 
> image

agree

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
```diff
@@ -88,6 +89,28 @@ are appended to the initial definition.
...
+         </attribute>
+         <attribute name="editorOnboardingImage" type="string">
+            <annotation>
+               <documentation>
+                  The image is shown in the editor area in case no editor is open.
+The image shall be grey and not colored and shall have a size of 350 x 350 px.
+Plus a second image for high resolution with a size of 700 x 700 px and a name like [image_name]@2x.[image_type].
+               </documentation>
+               <appInfo>
+                  <meta.attribute kind="resource"/>
```
BeckerWdf, line 110: why is this rendered as text field in the plugin.xml editor. It does not have a value help.
Compare e.g. to the org.eclipse.ui.views extensions point (from the same plugin). Here for the "icon" attribute you have a browse button and the label contains a link to navigate. We should have the same here.

BeckerWdf (changes requested):

marcushoepfner:
> This is how it looks like compared to VScode: image
> 
>     1. I agree with Lars: We should align it the same way as in VScode
> 
>     2. For my feeling the text could be a bit bigger. What does control the size of the text?
> 
>     3. For my feeling the space between the icon and the text is to big. The text and the list of command should move up to be closer to the icon.

1. will be done on separate PR
2. I don't know. The SWT API does not allow me to change the size, does it? Maybe in can be changed in preferences somewhere? 
3. I have reduced the space

iloveeclipse:
I haven't followed technical discussion here, but there are two commands that I wish would always be shown first:
- "Ctrl+3" as the most important shortcut to almost everything
- "Ctrl+Shift+L" to show available commands

actually "Ctrl+3" is **the** command that should be always shown.

Is this planned / possible?

bundles/org.eclipse.ui/schema/perspectiveExtensions.exsd
marcushoepfner: done. require-bundle for org.eclipse.ui was missing in the example plugin.

marcushoepfner:
> I haven't followed technical discussion here, but there are two commands that I wish would always be shown first:
> 
>     * "Ctrl+3" as the most important shortcut to almost everything
> 
>     * "Ctrl+Shift+L" to show available commands
> 
> 
> actually "Ctrl+3" is **the** command that should be always shown.
> 
> Is this planned / possible?

Thanks for the proposal. Very helpful commands.
But I would not add them programmatically to the onboarding editor. IMHO the perspective (via ext. point) should decide which commands to show. I added your proposals to the example plugin in #657

iloveeclipse:
> Thanks for the proposal. Very helpful commands.
> But I would not add them programmatically to the onboarding editor. IMHO the perspective (via ext. point) should decide which commands to show. I added your proposals to the example plugin

OK, I misunderstood the purpose / scope of this PR. This is also just a technical base, not actual changing of any perspectives. So what is the driving issue / request? Can you link to it?

BeckerWdf (approved):

iloveeclipse:
Do we plan to add some tests for new extension?

marcushoepfner:
Thanks for review. Can I merge or should I wait until #665 is fixed? Because I get the same errors here.

> Do we plan to add some tests for new extension?

I did not find any unit tests which I could extend.

marcushoepfner:
> > Thanks for the proposal. Very helpful commands.
> > But I would not add them programmatically to the onboarding editor. IMHO the perspective (via ext. point) should decide which commands to show. I added your proposals to the example plugin
> 
> OK, I misunderstood the purpose / scope of this PR. This is also just a technical base, not actual changing of any perspectives. So what is the driving issue / request? Can you link to it?

Originally was a bug (I think from @BeckerWdf) on bugzilla. copied it to  #633 (refer to PR message)

iloveeclipse:
> I did not find any unit tests which I could extend.

You can try to look around [this test](https://github.com/eclipse-platform/eclipse.platform.ui/blob/5efbfa52910b385efc0a027838b0ebf6e42cae51/tests/org.eclipse.ui.tests/Eclipse%20UI%20Tests/org/eclipse/ui/tests/api/IPageLayoutTest.java#L28).

iloveeclipse (dismissed):
I would prefer this change would contain trivial tests validating added API and perspective extension.

vogella:
@marcushoepfner you could add a test to org.eclipse.e4.ui.workbench.renderers.swt.StackRendererTest 

To execute it locally ensure you have select the HEADLESS application

image

marcushoepfner:
Currently writing tests for the new code. Will take a few days. Thanks for the hints

vogella:
Thanks @marcushoepfner for working on tests, I merge this already based on your statement that you will provide tests later. I do this because:

1.) I'm really looking forward to this functionality ;-)
2.) This work has already been in WIP since multiple release and I find it unfair to you that we continue to request changes and changes while not being very responsive in reviewing your contribution.

marcushoepfner:
Thanks. Workin' on the tests already.

PR #657, opened 2023-03-23
Example plugin for a perpective including onboarding

marcushoepfner:
Issue #633

Shows the usage of onboarding text, image and shortcuts.


BeckerWdf (changes requested):
Pls. add a proper version of the icon (incl. a @2x.png).
Ideally take the SVG version of the icon. Resize that and create the two PNGs out of it.

marcushoepfner:
> @marcushoepfner If I use interace rebase on this commit onto the PR for the onboarding and import the modified projects I see setOnboarding called on the StackRender but the empty editor area is still empty in a runtime Eclipse.
> 
> Anything else I need to do to see the change?

You need to switch to "Example Perspective" perspective.

vogella:
> > @marcushoepfner If I use interace rebase on this commit onto the PR for the onboarding and import the modified projects I see setOnboarding called on the StackRender but the empty editor area is still empty in a runtime Eclipse.
> > Anything else I need to do to see the change?
> 
> You need to switch to "Example Perspective" perspective.

Of course. :-) I had one of the old plug-ins from the target perspective in my runtime IDE. Works better now.

marcushoepfner:
> Pls. add a proper version of the icon (incl. a @2x.png). Ideally take the SVG version of the icon. Resize that and create the two PNGs out of it.

done

marcushoepfner:
I changed the onboarding text to something like IntelliJ uses:
`Open a file or drop files here to open them.`
Before I had `Welcome to the example perspective.` But I don't want this as an example. At the end every perspective has a text like `Welcome to my ...`, which does not any value. If in doubt, I would even recommend a perspective, don't add an onboarding text.

BeckerWdf:
> I changed the onboarding text to something like IntelliJ uses: `Open a file or drop files here to open them.` Before I had `Welcome to the example perspective.` But I don't want this as an example. At the end every perspective has a text like `Welcome to my ...`, which does not any value. If in doubt, I would even recommend a perspective, don't add an onboarding text.

I absolutely agree with that.

examples/org.eclipse.ui.examples.perspective/plugin.xml
```diff
@@ -0,0 +1,31 @@
...
+            icon="$nl$/icons/full/etool16/default_persp.png"
+            id="org.eclipse.ui.examples.perspective.perspective"
+            name="Example Perspective">
+      </perspective>
+   </extension>
+   <extension
+         point="org.eclipse.ui.perspectiveExtensions">
+      <perspectiveExtension
+            editorOnboardingImage="$nl$/icons/default_persp_grey.png"
+            editorOnboardingText="Open a file or drop files here to open them."
```
BeckerWdf: This should be externalized.

examples/org.eclipse.ui.examples.perspective/icons/default_persp_grey.png
BeckerWdf, line 1: i think these should go into the "full" folder

examples/org.eclipse.ui.examples.perspective/icons/default_persp_grey@2x.png
BeckerWdf, line 1: We should remove the empty border

vogella:
@marcushoepfner can you update this one also we that we can merge the example?

marcushoepfner:
> @marcushoepfner can you update this one also we that we can merge the example?

working on it, wait a few minutes pls :)

examples/org.eclipse.ui.examples.perspective/plugin.xml
marcushoepfner: I have added them to plugin.properties like e.g. plugin o.e.ui does it with extension texts.
Unfortunately they are not correctly retrieved in target and hence not shown in ui.
When I use the externalize wizard it adds them to a new file bundle.properties.

What am I doing wrong here?

marcushoepfner: Ah found it...
`Bundle-Localization: plugin` need to be added to Manifest.MF.

Now it works.

BeckerWdf (approved):

PR #662, opened 2023-03-28
dummy change to check build

marcushoepfner:
check if build is successful


marcushoepfner:
@laeubi pls. refer to the failures of this pr: https://github.com/eclipse-platform/eclipse.platform.ui/runs/12333575953
I did not change anything, just a assigned a statement to a variable.

Is this expected?

laeubi:
It shows that some test fail on windows / macos, as previously these where not used as build verification it seems possible that no one noticed it before.

marcushoepfner:
> It shows that some test fail on windows / macos, as previously these where not used as build verification it seems possible that no one noticed it before.

so we need to fix them? 
I cannot reproduce. I don't get these errors locally (in eclispe as pde test).

iloveeclipse:
> so we need to fix them?

First, I would propose that we report them.
Depending on the problem nature, we can fix them or mute them or do whatever else needed to stop these actions to fail and confuse everyone.

laeubi:
Of course its best to fix them as this obviously shows some problems (timings, os specific problems, ...) that can lead to random failures or undetected regressions...

marcushoepfner:
Done! -> #665

PR #670, opened 2023-03-30
Updating test data in org.eclipse.newInvalidMenuContribution1 to Java 17

vogella:


jukzi:
oh ok, looks like i missed even more versions to adapt.
With regexp `^(?!.*JavaSE-17).*(J2SE-|JavaSE-).*` it does also find even more versions to adapt. Mind to fix the others too?

PR #672, opened 2023-03-31
Adding Onboarding commands to the resource perspective

vogella:
As a first contribution adds Quick Access, Show Keys and New Wizard to the resource perspective

For #633

### DIFF
--- a/bundles/org.eclipse.ui.ide.application/plugin.xml
+++ b/bundles/org.eclipse.ui.ide.application/plugin.xml
@@ -57,6 +57,15 @@
          <showInPart
                id="org.eclipse.ui.navigator.ProjectExplorer">
          </showInPart>
+         <editorOnboardingCommand
+               id="org.eclipse.ui.window.quickAccess">
+         </editorOnboardingCommand>
+         <editorOnboardingCommand
+               id="org.eclipse.ui.window.showKeyAssist">
+         </editorOnboardingCommand>
+         <editorOnboardingCommand
+               id="org.eclipse.ui.newWizard">
+         </editorOnboardingCommand>
       </perspectiveExtension>
    </extension>
    <extension


vogella:
A first contribution of commands to get the discussion starting. 


@marcushoepfner I looked to me that if I add / remove onboarding commands in the IDE I need to use the clean workspace option to see this change. Maybe a cleanup step in missing in the code

@BeckerWdf I also think we should add images but in gray similar to VS. I see if I can get the image generator to work again on my machine and will provide another PR if I'm succesful.

BeckerWdf (approved):
Looks like this on macOS:
image

BeckerWdf:
> @BeckerWdf I also think we should add images but in gray similar to VS. I see if I can get the image generator to work again on my machine and will provide another PR if I'm succesful.

Yes pls. Best ist to take the SVG icon of the perspective and create a new (bigger and greayscaled) SVG out of it to generate the PNGs. SVG should then also be stored in the images repo.

@vogella: What about the onboarding text? Would you already add on in this PR?

vogella:
> > @BeckerWdf I also think we should add images but in gray similar to VS. I see if I can get the image generator to work again on my machine and will provide another PR if I'm succesful.
> 
> Yes pls. Best ist to take the SVG icon of the perspective and create a new (bigger and greayscaled) SVG out of it to generate the PNGs. SVG should then also be stored in the images repo.

Can you do this? If I add a new svg file to the tree and regenerate the png files, it is ignored. Maybe an inconsistance of the inkscape version on Linux.

> @vogella: What about the onboarding text? Would you already add on in this PR?

Which text do you propose? VS has no text and I think it looks really good.

vogella:
By the way, I plan to rename "Find Action" to "Show All Commands" like in VS which is much telling IMHO.

vogella:
We also need a better name for the new wizard. "New" is not very telling. Proposals?

vogella:
image

vogella:
Also "Show Key Assists" feels like IT talk from the past. Maybe we should also find a better name. Suggestions are welcome.

mickaelistria:
> Also "Show Key Assists" feels like IT talk from the past. Maybe we should also find a better name. Suggestions are welcome.

"List shortcuts"

BeckerWdf:
> By the way, I plan to rename "Find Action" to "Show All Commands" like in VS which is much telling IMHO.

I never liked "Find Actions". Wasn't it  the "Quick Access" before?. I don't know if "Show All Commands" is correct because this now can do even more (e.g. perform searches e.g.).

BeckerWdf:
> Which text do you propose? VS has no text and I think it looks really good.

I mean the text that is "Open a file or drop files here to open them." in the demo perspective (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/657/files)

vogella:
> > By the way, I plan to rename "Find Action" to "Show All Commands" like in VS which is much telling IMHO.
> 
> I never liked "Find Actions". Wasn't it the "Quick Access" before?. I don't know if "Show All Commands" is correct because this now can do even more (e.g. perform searches e.g.).

"Quick Access"  is  IMHO not telling anything to the user, so I think reverting this would be a step back. See the discussion here https://github.com/eclipse-platform/eclipse.platform.ui/pull/673

vogella:
> > Also "Show Key Assists" feels like IT talk from the past. Maybe we should also find a better name. Suggestions are welcome.
> 
> "List shortcuts"

I like it. Please create a PR, my energy to discuss user facing names has already been consumed for today.

vogella:
Unrelated test error, merging.

vogella:
> > Which text do you propose? VS has no text and I think it looks really good.
> 
> I mean the text that is "Open a file or drop files here to open them." in the demo perspective (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/657/files)

Good idea IMHO, please provide separate PR for this.

akurtakov:
Gnome has a very good looking and readable "Shortcuts" 
image

BeckerWdf:
> Can you do this? If I add a new svg file to the tree and regenerate the png files, it is ignored. Maybe an inconsistance of the inkscape version on Linux.

Generation can be done manually in Inkscape of with the Image Render from the images repo - this one does not interact with Inkscape at all.

BeckerWdf:
> I like it. Please create a PR, my energy to discuss user facing names has already been consumed for today.

see #676

vogella:
> > I like it. Please create a PR, my energy to discuss user facing names has already been consumed for today.
> 
> see #676

Thanks, this comment was intended for the "List shortcuts" suggestion from @mickaelistria but your PR is also very much welcomed.

BeckerWdf:
> Thanks, this comment was intended for the "List shortcuts" suggestion from @mickaelistria

Ups.

PR #673, opened 2023-03-31
Rename "Find Actions" to "Show All Commands"

vogella:
VS Code is very popular and uses "Show All Commands" to its Quick Assists. Lets re-use the UX expert knowledge from Microsoft.


mickaelistria (changes requested):
Actually, it's more "search anything", as it also lists resources, workspace symbols, launch configs... and is extensible

merks:
Yes as @mickaelistria says, it's misleading to suggest that is only showing commands.  Commands is only one "small" section:

image

I'm not a big fan of "let's do X because VSCode does X" with the implication that VSCode is better and the people working in it have more expertise.  That might well be the case, but even they presumably have a good reason for doing something and it's better we know that good reason than to suggest they simply know better that we do without understanding the reason.  But that's just my feeling...

iloveeclipse:
What about old good "Quick Access"? Or "Quick Start"?

iloveeclipse:
Or just leave AS IS, what's wrong with "Find Actions" except it is not proposed by a Microsoft designer?

vogella:
Thanks that you all like the name "Find Actions" . IIRC @mickaelistria and I renamed "Quick Access" (which means nothing) to "Find Actions". 

And yes I think Microsoft does more UX research that we do. The move from "Quick Access" to "Find Actions" was deflinitely an improvement and reverting it to its original meaningless string is a no-go from my side.

Lets leave it like it it.

PR #676, opened 2023-03-31
Onboarding in Resource Perspective

BeckerWdf:
Add icon and text


vogella:
LGTM

BeckerWdf:
Looks like this:
image

iloveeclipse:
Just a question before we will start to add more of copy/pasted code in every perspective: can this be somehow **re-used**?
I mean, for PDE / Java tooling default perspectives would make sense to **extend** this default set of commands by only adding few actions related to PDE / Java, but not by copy/pasting all of that including images?

marcushoepfner:
> Just a question before we will start to add more of copy/pasted code in every perspective: can this be somehow **re-used**? I mean, for PDE / Java tooling default perspectives would make sense to **extend** this default set of commands by only adding few actions related to PDE / Java, but not by copy/pasting all of that including images?

I've get your point. But is your proposal really about duplicate "code" or regarding harmonization of commands and text?
Because the code is really minimal and I would not see this as duplicated code:
See https://github.com/eclipse-platform/eclipse.platform.ui/blob/3d3acf7c8333f019bfc162befc405caf7f918a87/bundles/org.eclipse.ui.ide.application/plugin.xml#L53-L72

BeckerWdf:
I tuned the icons again so that it also looks "not bad" in dark mode.
No looks like this:
image
and:
image

In dark mode it's a bit light but if I make it darker then it looks too dark in light theme. This is a improvement candidate once we can  replace icons via icon-pack / depending on themes.

BeckerWdf:
> can this be somehow **re-used**?

Currently not.

iloveeclipse:
> I've get your point. But is your proposal really about duplicate "code" or regarding harmonization of commands and text?
Because the code is really minimal and I would not see this as duplicated code:

My point is: there are some very common commands / proposals / icons that can/should be same if used in different perspectives. Seeing **how** the icons & text etc are added to the resource perspective in platform, I assume that **same** icons / initial set of shown commands can be reused in JDT "Java" perspective, etc.

Is there a way to **reuse** what is contributed and only **extend** it by a single additional "Open Type" command, or is JDT supposed to copy/paste the icons, onboarding text, etc just to add one more entry for "Ctrl+Shift+T"?

iloveeclipse:
See https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/505 task for JDT UI.

iloveeclipse:
> > can this be somehow **re-used**?
> 
> Currently not.

Can we at least share icons (does the extension allows to reference icons via "platform:/plugin/xxx" urls)? I hate any code duplication / inconsistency that it causes at the end.

vogella:
Thanks @BeckerWdf looks great to me

BeckerWdf:
> Can we at least share icons

I don't see that different perspective should share the same icon. I would propose to always show a big version of the perspective icon there. And the Java perspective icon is different to the Resource perspective icon.

PR #682, opened 2023-04-02
When updating the image from the workbench part currently a trivial put

### LINKED ISSUE #679: Reused editors lost their icon in 4.26
Found originally in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/506 with class file editors, but can be easily reproduced with plain text editors.

**This is a regression in 4.26.**

- Make sure "Reuse editors to show search matches" is enabled 
image
- Open java.util.Random type
- Click on the name and "Search for References"
- In the Search view, just open few types that refer to "Random"
- First match opened in the editor tab **with an icon**
- **Second** and each other match is shown in the same editor **without** the icon n the tab:
image

Same for plain text editors:
- Make sure "Reuse editors to show search matches" is enabled 
- Create few text files with "aaaaa" text
- Search for "aaaaa" text in the project
- Open files one after each other, starting with the second one they all lose icon.

It only **still** works in Java editors after 4.26 because CompilationUnitEditor has extra codee that **always** recomputes and resets the editor icon in `org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitEditor.doSetInput(IEditorInput)` with the help of `org.eclipse.jdt.internal.ui.javaeditor.JavaEditorErrorTickUpdater.updateEditorImage(IJavaElement)`.

laeubi:
When updating the image from the workbench part currently a trivial put operation is used, but because of how the transient data map works it only triggers a change if the value is not equal to the previous one.

It can now happen that two images are considered equal because they are created from the same imagedescriptor, but one of them is disposed afterwards in which case the value in the transient data is stale even though it was tried to being updated.

This is now changed in the following:
- first the current value is compared with the update value using identity comparison
- if it is not the same object, a remove/put operation is performed to enforce an update of the value in the map

Fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/679
FYI @Dani-Hub

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
@@ -18,6 +18,7 @@
 
 package org.eclipse.ui.internal.e4.compatibility;
 
+import java.util.Map;
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import javax.inject.Inject;
@@ -383,9 +384,21 @@ public abstract class CompatibilityPart implements ISelectionChangedListener {
 			case IWorkbenchPartConstants.PROP_TITLE:
 				part.setLabel(computeLabel());
 
-				if (wrapped.getTitleImage() != null) {
-					Image newImage = wrapped.getTitleImage();
-					part.getTransientData().put(IPresentationEngine.OVERRIDE_ICON_IMAGE_KEY, newImage);
+				Image titleImage = wrapped.getTitleImage();
+				if (titleImage != null) {
+					Map<String, Object> data = part.getTransientData();
+					// we need to do an identity compare here, because images are considered equal
+					// when there imageproviders are equal, but the native image instance might be
+					// different and therefore disposed indivually!
+					Object object = data.get(IPresentationEngine.OVERRIDE_ICON_IMAGE_KEY);
+					if (titleImage != object) {
+						if (titleImage.equals(object)) {
+							// to trigger a change, we need to perform a remove/add instead of a simple put
+							// operation
+							data.remove(IPresentationEngine.OVERRIDE_ICON_IMAGE_KEY);
+						}
+						data.put(IPresentationEngine.OVERRIDE_ICON_IMAGE_KEY, titleImage);
+					}
 				}
 				String titleToolTip = wrapped.getTitleToolTip();
 				if (titleToolTip != null) {


Dani-Hub:
Looks good to me. I checked that at least the text editor testcase of @iloveeclipse has been fixed by this.

iloveeclipse (changes requested):
> if it is not the same object, a remove/put operation is performed to enforce an update of the value in the map

I see in debugger that the map is based on `org.eclipse.emf.common.util.BasicEMap`.
It looks like there are **two** events sent now to clients (icon removed, icon added).
Not sure how bad is that, assuming before the original change there should be only one event.

- In best case "just" adding overhead for every listener.
- In worst case another unexpected side effect.

However I don't think the patch in platform UI is in the right place at all.

Looking how `Image.equals() / hashCode()` was updated (https://bugs.eclipse.org/bugs/show_bug.cgi?id=421383#c88) and https://git.eclipse.org/r/c/platform/eclipse.platform.swt/+/43279/15/bundles/org.eclipse.swt/Eclipse+SWT/gtk/org/eclipse/swt/graphics/Image.java, the intent was to have **not** "equal" images if they were representing different native objects.

> In image class we are dealing with native resources. for the equality, object comparison might not be sufficient as it will not have a mechanism to go into the native handles and compare.

Means : with 4a2ae8175aad40061c54f15827ea0bb1c4a7dcbd (and it's refactored version 12f542b9203bac54a3af3ecf42228481ac23f3df) two **different** image objects (representing **same** icon) loaded from **same** `ImageFileNameProvider` (==` FileImageDescriptor` or `URLImageDescriptor`) are considered **equal** (and have same hashcode too).

That violates the above assumption about image equality and with that `Image.equals() / hashCode()` are broken for images loaded via `org.eclipse.swt.graphics.Image.Image(Device, ImageFileNameProvider)` constructor!

@niraj-modi , @sravanlakkimsetti  FYI.

laeubi:
> However I don't think the patch in platform UI is in the right place at all.

So what should be the right place then? SWT?

> It looks like there are two events sent now to clients (icon removed, icon added).
Not sure how bad is that, assuming before the original change there should be only one event.

As mentioned the best would be if there is no event at all as the image actually never changes, but given that the image is replaced anyways I won't think it harms much, maybe `BasicEMap` can be made more smart to coalesce a remove/add event into one change event at least it is better than the previous broken state where it only worked due to the side-effect that `ImageFileNameProvider`s are never equal (what makes it a bit useless to call equals as part of the image equals check).

iloveeclipse:
Hmm, reading more into https://bugs.eclipse.org/bugs/show_bug.cgi?id=421383 I wonder if I understand that whole story right.

The intent **was** to have **same** equals()/hashCode() on different Image objects created with **same** ImageFileNameProvider, which is surprising / contradicting to other SWT resources code.

So the images created by both FileImageDescriptor and URLImageDescriptor (code added by 640764654ecdfa321278341ee5787f16ef0a1fc9 / e6d12ca0ae4762273680dc65ae8eb6f211778acc) for same paths were always "different" (before https://github.com/eclipse-platform/eclipse.platform.ui/commit/4a2ae8175aad40061c54f15827ea0bb1c4a7dcbd) and UI always "saw" a change.

However, image equality was changed via https://github.com/eclipse-platform/eclipse.platform.swt/commit/40be4c7971eae41d1173609a4e98b98bf96978a0 where images were considered equal if they were loaded by "same"  data provider.

And with https://github.com/eclipse-platform/eclipse.platform.ui/commit/4a2ae8175aad40061c54f15827ea0bb1c4a7dcbd providers stored in images created by FileImageDescriptor and URLImageDescriptor were not changing anymore, which would mean, from SWT point of view the images we set in [CompatibilityPart](https://github.com/eclipse-platform/eclipse.platform.ui/blob/c1966a8a5aed2f6b3de6f175a4fa0189ddd601ae/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java#L388) were not different anymore.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
```diff
@@ -385,7 +386,15 @@ public void create() {
 
 				if (wrapped.getTitleImage() != null) {
 					Image newImage = wrapped.getTitleImage();
-					part.getTransientData().put(IPresentationEngine.OVERRIDE_ICON_IMAGE_KEY, newImage);
+					Map<String, Object> data = part.getTransientData();
+					// we need to do an identity compare here, because images are considered equal
+					// when there imageproviders are equal, but the native image instance might be
+					// different and therefore disposed indivually!
+					if (newImage != data.get(IPresentationEngine.OVERRIDE_ICON_IMAGE_KEY)) {
```
iloveeclipse: It would make sense to check for `null` here / extract `data.get(IPresentationEngine.OVERRIDE_ICON_IMAGE_KEY)` to a variable.
because if the original image is `null`, there is no need to remove the `null` from the map.

laeubi:
@iloveeclipse I think your summary is correct, and I have no clue why images should compare equal that way, I personally try to avoid implementing equal / hashcode in all cases except the class is final and unmodifiable, in all other cases it is just a wait for strange things to happen.

Also that it is now visible due to this change, it was never forbidden to reuse the same providers (and one can easily construct some, e.g. by copy the image), so one get the following very strange behaviors:

1. I have Image A and create a zoomed image from it with 200%, both images are considered equal even though they have completely different size
2. I have Image A and create a copy Image B then I dispose A but they are still considered equal even if I can't use A for anything useful anymore
3. ... probably more I have not though about yet ...

The next thing that came here into play is that the put operation (in contrast to their API contract) seem not to replace the value as expsected, but additionally checks if both values are equal and then do nothing. This seems reasonable from EMF point of view, but shows that having equals/hashcode implementations that are not strict can cause strange results, because obviously the images *are* equal at the time of calling but right after the method call one will be disposed rendering it useless.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
iloveeclipse: @laeubi : can you please change that?

laeubi: I have refactored the part also a bit so the title image is only fetched once, and the remove is only performed if they are equal (what includes the null case).

iloveeclipse (dismissed):
LGTM

iloveeclipse:
Actually the commit message could be improved:
1) The first line is missing. Commit should start with a short (~80 characters) title separated from the commit message with a new line. This helps all the git tooling to properly show commit in the history.
2) Commit misses a link to the issue it fixes, like `Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/679`. This helps to link PR's with issues and also for later easy navigation to the issue from Eclipse git history view.

merks:
This really looks like a "horrible hack" to me.  Don't take offense!  I can see that the value is set and the notification is sent, even when the value is identical:

image

That notification is ignored because org.eclipse.emf.common.notify.impl.NotificationImpl.isTouch() returns true  when `return oldValue == null ? newValue == null : oldValue.equals(newValue);` and rg.eclipse.e4.ui.internal.workbench.UIEventPublisher.notifyChanged(Notification) does this

image

So it seems to me if these are two different images and one is broken and the other is good, they should **NOT** be equal. Patching this one place where we noticed how broken that is is likely to bite us somewhere else too, but we just haven't noticed yet....

iloveeclipse:
> So it seems to me if these are two different images and one is broken and the other is good, they should NOT be equal.

See [my summary](https://github.com/eclipse-platform/eclipse.platform.ui/pull/682#issuecomment-1494277864) for the current "state of the art" in Image.equals().

@merks : thanks for explanation, but it looks like the "strange" Image.equals() implementation is expected / desired (coming from HiDPI changes).

Is there a way to "customize" NotificationImpl.isTouch() for some properties or do something "special" for Images in UIEventPublisher?

laeubi:
> Actually the commit message could be improved:

I updated it now, no idea why that was missing in the first place.

> This really looks like a "horrible hack" to me. 

Well it is a _compatibility hack_  ;-)

> So it seems to me if these are two different images and one is broken and the other is good, they should NOT be equal. 
> Patching this one place where we noticed how broken that is is likely to bite us somewhere else too, but we just haven't noticed yet....

As said i have no clue why this equal definition was chosen in the first place, but changing it seems quite risky, but I think we already have had similar issues elsewhere, actually resource handling in SWT is a mess and always a source of confusion, so whenever possible I try to use `ImageDescriptor` so the _consumer_ can handle this case and not the _provider_ of the "image" and can control the lifecycle.

I just provided this because I think the original reported issue is not caused by the `ImageDescriptor` change but this just reveals the problem here.

sravanlakkimsetti:
One of the assumptions we had when we developed HiDpi support is the image always be refreshed when it is rendered to the display. Based on this assumption the equals method has been updated to compare with the provider. 

This assumption is definitely **not** correct in this situation. I would prefer comparing the actual OS level resources to actually match the image.

merks:
Has anyone suggested/considered add a check that both or neither are disposed to the equality check?

image

If I understood this long thread correctly, that's the underlying problem of why two images can be equal but only one is "good"...

And yes, we could "hack" the generated code to always generate a non-touch notification...

merks:
@laeubi  Please not that no personal criticism was intended.  Kudos for investigating and trying to eliminate this problem!

laeubi:
> Has anyone suggested/considered add a check that both or neither are disposed to the equality check?

Yes I also though about that, but at the time of the update the images are not disposed, so it won't help much here, also one might argue that any disposed image is equal and all that kind of problematic properties, in the end I think an image should probably never implement equals and better use the standard identity check.

> If I understood this long thread correctly, that's the underlying problem of why two images can be equal but only one is "good"...

At the point of the call both are "good", but the intend is to replace one with the other because the old one will be disposed **after** the call and then becomes invisible.

merks:
I see.  Would you/anyone like me to try to fix this in the model implementation?  This would ensure there is only a single notification...

laeubi:
> I see. Would you/anyone like me to try to fix this in the model implementation? This would ensure there is only a single notification...

If we can fix it on a more central place this would be the best solution because other places probably can be affected too here.

merks:
There's something stranger going on than just a touch notification being ignored...

laeubi:
> There's something stranger going on than just a touch notification being ignored...

I _think_ i tried comment out the touch check but then still see problem, but not sure anymore and though it was because my lack of understanding how this model stuff works 😇

merks:
Here's what else is going wrong:

image

When the title changes, the property change is fired, setting the model property yet again, and then properly producing a notification that's isn't a touch, eventually ends up wanting to update the CTab folder, which is currently holding the about-to-be-disposed image and is being set to the newly created image of the editor.  But again, the images are equal so the CTab folder does not replace the about-to-be deleted image.

This is why your hack works better (properly) than the notification hack because your hack produces an additional notification that sets the image to null such that later it can be set to a new (equal) image, thereby bypassing the bogus equality check.

Again we come come down to the implementation of Image equality is bad/wrong, this causes problems, and we don't know where all those problem lurk.  Give given that your hack works and my doesn't, that makes your hack better.  👿  🙈 

But surely we need to fix the Image equality such that it properly tests "are these two things the same image".

laeubi:
I think CTab should especially not use equals here ...

The javadoc says:

> Sets the receiver's image to the argument, which may be null indicating that no image should be displayed.

so as a caller I can expect, whatever happens, that after this call the image is set to the given instance, not something else or _equal_.

iloveeclipse:
@merks : I can't force icon to change by changing CTabItem like
```
diff --git a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabItem.java b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabItem.java
index 884c443..7c084cd 100644
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabItem.java
@@ -468,3 +468,3 @@
 	if (image == null && oldImage == null) return;
-	if (image != null && image.equals(oldImage)) return;
+	if (image != null && image == oldImage) return;
 	super.setImage(image);
```
Do I miss something?

merks:
Of course, we can keep adding "hacks" for all the places that Image.equals behaves inappropriately now.  But do we know all those places? (No, we'll never know all those places.)  Do we want to hack all those places?  (Given it's unknown and unbounded, this is probably a slipper slope.)  It seems clearly preferable to fix the problem properly at the root cause...  Has it been decided that that approach is an unsolvable problem?

merks:
I wonder if these other places in SWT have a potential problem:

image

Or maybe all these 45 places that do image.equals

image

merks:
Note that this change to org.eclipse.e4.ui.model.application.impl.StringToObjectMapImpl.setTypedValue(Object) in combination with the change to use == in CTabItem does work correctly during my manual testing.
```
	/**
	 * <!-- begin-user-doc -->
	 * <!-- end-user-doc -->
	 * @generated NOT
	 */
	public void setTypedValue(Object newValue) {
		Object oldValue = value;
		value = newValue;
		if (eNotificationRequired())
			eNotify(new ENotificationImpl(this, Notification.SET, ApplicationPackageImpl.STRING_TO_OBJECT_MAP__VALUE,
					oldValue, value) {
				@Override
				public boolean isTouch() {
					boolean touch = super.isTouch();
					if (touch && oldValue != newValue && (isImage(oldValue) || isImage(newValue))) {
						return false;
					}
					return touch;
				}
			});
	}

	private static boolean isImage(Object value) {
		return value != null && "org.eclipse.swt.graphics.Image".equals(value.getClass().getName());
	}
```
Note that in the model, Image is not visible and should not be visible, and I wrote the guards carefully to minimize any performance impact...

vogella:
I also think it is strange that we accept an incorrect Image.equals implementation like @merks.

laeubi:
> I also think it is strange that we accept an incorrect Image.equals implementation like @merks.

Without additional information one can't tell if it is incorrect. Just because it does not work as we like for certain situations does not mean it is wrong, the javadoc only states:

```
/**
 * Compares the argument to the receiver, and returns true
 * if they represent the <em>same</em> object using a class
 * specific comparison.
 *
 * @param object the object to compare with this object
 * @return <code>true</code> if the object is the same as this object and <code>false</code> otherwise
 *
 * @see #hashCode
 */
```

So it does not tell anything specific about how it works, even though I personally would say that two image object can never be _equal_ if they are not the _same_, and for all other cases one should use a specific method, e.g. `Image#isSimilar` or whatever, so for me it is currently unclear how SWT uses equals and what this has to do with HighDPI...

iloveeclipse:
The problem with Image.equals() has to be really fixed in UI, not in SWT.

SWT Image.equals() contracts are weird but they have rights to be implemented in the way they are implemented.

> One of the assumptions we had when we developed HiDpi support is the image always be refreshed when it is rendered to the display. Based on this assumption the equals method has been updated to compare with the provider.
> This assumption is definitely not correct in this situation. I would prefer comparing the actual OS level resources to actually match the image.

@sravanlakkimsetti : the assumption that you [were talking about](https://github.com/eclipse-platform/eclipse.platform.ui/pull/682#issuecomment-1495606429) above means, ImageDataProvider / ImageFileNameProvider were supposed to **always** have 1:1 relationship to created Images **for same display resolution**? If I understood it right this assumption was based on the idea that the Image object itself would "automatically" re-create native Image on resolution change.

If I understood the intent right, this would help existing widgets that hold Image references to be automatically updated without resetting Image instances.

The Image equality change was [explained here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=421383#c92).
If I understood it right, **originally** (before we had ImageDataProvider / ImageFileNameProvider) Image.equals() should return true for if Images refer to **same native resources** (handles) allocated by SWT. 

The original handle based implementation was needed to make sure the code below is always true:
```
Display.getSystemImage(SWT.ICON_ERROR).equals(Display.getSystemImage(SWT.ICON_ERROR))
```

After adding HiDPI support, equality was extended to include checks for  ImageDataProvider / ImageFileNameProvider objects equality, because if Image **automatically** would change native handle on resolution change, equals() contract (be consistent on multiple invocations) would be violated if it would rely on changed handle.

Therefore ImageDataProvider / ImageFileNameProvider updated equals() semantic was [implemented this way](https://github.com/eclipse-platform/eclipse.platform.swt/commit/40be4c7971eae41d1173609a4e98b98bf96978a0) to make sure the code below works: (taken from [this example project](https://bugs.eclipse.org/bugs/attachment.cgi?id=273445) from [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=532443)):

```
import java.util.HashSet;
import java.util.Set;

import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.graphics.ImageFileNameProvider;
import org.eclipse.swt.widgets.Display;

public class ImageTest {

		ImageFileNameProvider provider = new ImageFileNameProvider() {
			@Override
			public String getImagePath(int zoom) {
				return "TargetPlatformTask.gif";
			}
		};
		Image image = new Image(display, provider);
		Image image2 = new Image(display, provider);
		
		Image imageCopy = new Image(display, image, SWT.IMAGE_COPY);
		Image imageGray = new Image(display, image, SWT.IMAGE_GRAY);
		Image imageDisable = new Image(display, image, SWT.IMAGE_DISABLE);
		
		Set<Image> images = new HashSet<Image>();
		images.add(image);
		images.add(image2);
		images.add(imageCopy);
		images.add(imageGray);
		images.add(imageDisable);
		
		System.out.println("Expecting 3 unique icons but there are only " + images.size());
	}
}
```

But this example above is exactly the reason why with https://github.com/eclipse-platform/eclipse.platform.ui/commit/4a2ae8175aad40061c54f15827ea0bb1c4a7dcbd (and it's refactored version https://github.com/eclipse-platform/eclipse.platform.ui/commit/12f542b9203bac54a3af3ecf42228481ac23f3df) breaks platform UI code in CompatibilityPart.

So it looks to me, to avoid more platform side effects, the change [that introduced](https://github.com/eclipse-platform/eclipse.platform.ui/commit/4a2ae8175aad40061c54f15827ea0bb1c4a7dcbd) **reusable singleton** ImageFileNameProvider instance in FileImageDescriptor and URLImageDescriptor must be reverted.

Dani-Hub:
Please don't revert directly as added. If agreement exists, that we want to revert these changes, I can make a pull request to these effects. I would like to ensure that the rest of the reasons this series of pull attempts to realize are still conserved.

merks:
Note that one of the "fixes" is CTabItem which is in SWT.  Also, org.eclipse.swt.widgets.TreeItem.setImage(int, Image)  does image comparison and it can easily suffer from the same problem.  The general problem is that if image1.equals(image2) that there will be code that decides that there's no change to be made, like this TreeItem code is assuming that:

image

But if the caller assumes they can dispose the "old" image because they've replaced it with a new image, that's a bad assumption, and SWT itself is assuming it can ignore replacement for equal images...

iloveeclipse:
> Please don't revert directly as added. 

Sure

> If agreement exists, that we want to revert these changes, I can make a pull request to these effects. I would like to ensure that the rest of the reasons this series of pull attempts to realize are still conserved.

Can you please provide this PR, so we can see what is needed to preserve previous descriptor behavior?

Dani-Hub:
Sure, let me work on it

iloveeclipse:
> Note that one of the "fixes" is CTabItem which is in SWT.

@merks : please open a ticket in SWT for the cases you've found. We have to decide one by one if that breaks something and has to be fixed or if that is again some "very special" case of something we have to deal with in SWT.

laeubi:
> So it looks to me, to avoid more platform side effects, the change [that introduced](https://github.com/eclipse-platform/eclipse.platform.ui/commit/4a2ae8175aad40061c54f15827ea0bb1c4a7dcbd) reusable singleton ImageFileNameProvider instance in FileImageDescriptor and URLImageDescriptor must be reverted.

I must clearly disagree here. `ImageFileNameProvider` / `ImageDataProvider` does not provide any of the contracts you describe here, if SWT require this internally it must wrap any user supplied object in some kind of whatever helper objects that suffice here (e.g. it could use an `InternalImageFileNameProvider` that is copied on image copy if it is already at the correct type and implements equals based on object identity), as from the API contract there is not a little information for implementors and changing that would be an API breakage.

```
/**
 * Interface to provide a callback mechanism to get information about images
 * when the application is moved from a low DPI monitor to a high DPI monitor.
 * This provides API which will be called by SWT during the image rendering.
 *
 * This interface needs to be implemented by client code to provide the
 * image information on demand.
 *
 * @since 3.104
 */
```

So the only contract is that the implementation must "provide the image information on demand", what is satisfied here, and so this is then clearly a bug in SWT implementation.

iloveeclipse:
@laeubi : have you seen my comment above? If this an SWT bug, please provide explanation **what** is broken in SWT? So far, SWT only code works as expected, it's Platform UI code which shows a bug.



The regression was caused by descriptors change. The change itself with a singleton image provider wasn't technically necessary or fixed any bug - exact the opposite happened.

So from my POV this regression need to be fixed first. 

> f SWT require this internally it must wrap any user supplied object in some kind of whatever helper objects that suffice here (e.g. it could use an `InternalImageFileNameProvider` that is copied on image copy if it is already at the correct type and implements equals based on object identity

This would reintroduce https://bugs.eclipse.org/bugs/show_bug.cgi?id=532443 and break snippet above.

Dani-Hub:
Pull request has been provided, see #701. I agree that the reusable character of the providers were not originally requested, they were suggested as additional change during the review and I created separate PULL requests for them.

laeubi:
> The regression was caused by descriptors change. The change itself with a singleton image provider wasn't technically necessary or fixed any bug - exact the opposite happened.

And technically it is not necessary nor required by the API to use a new instance each time but unnecessarily creates new instance for nothing. And technically all the Java 17 change do not "fix" anything ... so we always has changes that improve things in different regards. 

> If this an SWT bug, please provide explanation what is broken in SWT

If SWT only works when image providers are effectively recreated making the equals-check obsolete anyways then I would call it a bug. Also @merks has identified places where equals is used in a way that is questionable given the equals implementation, so this is another bug. Both bugs are located in SWT and can now be seen / reproduced since platform.ui was improved to not create useless new objects to fulfill a side-effect in SWT.

> This would reintroduce https://bugs.eclipse.org/bugs/show_bug.cgi?id=532443 and break snippet above.

The the fix in SWT is either not sufficient or one should simply not use `Image#equals` as it is broken by design.

iloveeclipse:
@laeubi : please provide **concrete** explanation of SWT Image.equals() bug, with an SWT example.
Examples found by Ed are from different code, they may or may not be bugs.

iloveeclipse:
> technically it is not necessary nor required

... to introduce regressions.

laeubi:
> please provide concrete explanation of SWT Image.equals() bug, with an SWT example.

I never stated that this is a **bug** in SWT `Image.equals() `, please see above.

I only said that equals of an image **can't be used like it is used here** and provided a patch to fix that false behavior in `platform.ui` and @merks provided other places where Image#equals is used inappropriately. I'll try to provide a snippet to reproduce the **behavior** with pure SWT but I don't know if there is any demand in SWT to _change_ the behavior at all.

So this PR fixes the reported regression, everything else is beyond what this PR is about.

iloveeclipse:
> I never stated that this is a **bug** in SWT `Image.equals() `, please see above.

OK, this wasn't clear for me. 

> I only said that equals of an image **can't be used like it is used here**

Such use of images is widely spread over UI code, as Ed noticed above. 

> So this PR fixes the reported regression, everything else is beyond what this PR is about.

I believe https://github.com/eclipse-platform/eclipse.platform.ui/pull/701 is a better fix, for a simple reason.

This PR fixes **one single place** where Image. equals() is used to make a decision.

#701 fixes a whole class of such places by making sure that Image.equals() works as expected for all users of URLImageDescriptor and FileImageDescriptor.

The [code in Platform UI that added](https://github.com/eclipse-platform/eclipse.platform.ui/commit/640764654ecdfa321278341ee5787f16ef0a1fc9) both descriptors was added for same reason as the https://github.com/eclipse-platform/eclipse.platform.swt/commit/40be4c7971eae41d1173609a4e98b98bf96978a0 that implemented new equals() behavior. 

People who implemented these changes worked together, so the "strange" implementation for both Image and URLImageDescriptor / FileImageDescriptor as not a coincidence. 

With https://github.com/eclipse-platform/eclipse.platform.ui/pull/701 we revert Platform UI behavior that was **intended** but unfortunately not properly documented nor tested (in Platform UI).

iloveeclipse (changes requested):
Approach in https://github.com/eclipse-platform/eclipse.platform.ui/pull/701 provides more generic solution. 

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/682#issuecomment-1496410657

merks:
@iloveeclipse  Thanks 💯 for your tireless dedication to resolving problems. 

@Dani-Hub  Thanks for following the discussions and staying involved. 🥇

laeubi:
> Approach in https://github.com/eclipse-platform/eclipse.platform.ui/pull/701 provides more generic solution.

This is not a "solution" this just knock-out the equals approach of `Image` that was claimed to be done for a purpose of HiDPI what we are actually implementing here with the `@2` / `AxB` approach and "fixes" the problem for two specific implementations.

Even worse, it will only work for images created by ImageDescriptor and not for any other implementation or other images as I even showed with a pure SWT example ... instead now we define a bug as "intended behavior", and after rereading this issue and the other comment seeing how often "this commit is the culprit" I must confess it was never planned to not revert it and understand the real issue.

That's really annoying, but I must confess I'm to tired about this, so revert what ever you think suffices and wait for the next bad things to happen, I'm giving up on this ...

vogella:
Assuming that SWT has the right to implement equals the way it wants or needs is wrong.  Javadoc of Javadoc Ob ject equals    * Indicates whether some other object is "equal to" this one. 

Lets ask a KI (https://chat.openai.com/) about it:

-------------
Is a 16-16px image equal to a 32-32px image?

A 16x16px image is not equal to a 32x32px image. While they may have the same aspect ratio (1:1), the 32x32px image has four times the number of pixels as the 16x16px image, so it contains more information and detail.

If you were to display both images at the same physical size, the 32x32px image would appear sharper and more detailed than the 16x16px image.

However, the appropriate image size to use depends on how the image will be used. For example, if the image will be displayed as a small icon in a toolbar or menu, a 16x16px image may be more appropriate to conserve space and reduce visual clutter. If the image will be displayed as a larger graphic, a higher resolution image may be more appropriate to ensure that it looks sharp and detailed.

In summary, the size of the image required depends on its intended usage and the context in which it will be displayed.

-------------

While SWT may use this hack to make certain things work, it is still not correct.

iloveeclipse:
> While SWT may use this hack to make certain things work

With "certain" you probably mean HiDPI support? Feel free to provide a correct HiDPI implementation for 3 different platforms that doesn't break any existing API contract and has "correct" implementation without hacks. I believe everyone complaining about SWT doesn't really know how complicated the implementation is and underestimates the effort required to do things "correct", whatever it means.

> it is still not correct.

Correct in the sense of which contract? If you know how "correct" implementation should look like, please share your knowledge with us.

So far no one commented on this PR ([except may be this comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/682#issuecomment-1496192503)) really tried to understand **why** SWT provided the current implementation and no one complaining about "incorrect" equals() implementation tried to provide a specification of a "correct" one.

Once anyone has a concrete specification for a "correct" Image.equals() in SWT that would not break existing code / API and would show how "incorrect" the current implementation is, don't hesitate to create an issue in SWT.

sravanlakkimsetti:
> @sravanlakkimsetti : the assumption that you [were talking about](https://github.com/eclipse-platform/eclipse.platform.ui/pull/682#issuecomment-1495606429) above means, ImageDataProvider / ImageFileNameProvider were supposed to **always** have 1:1 relationship to created Images **for same display resolution**?

the idea we had was when the image is being rendered, there will be a call to imageProvider to return appropriate image to the display. This will recreate native image as necessary. That is why even though the native resources are of different sizes, we considered image as equal if the image object is created is using the same imageProvider. 

if there is an imageProvider we should not rely on the native resources. the native resources specified in the image object are temporary for that resolution.

vogella:
> > While SWT may use this hack to make certain things work
> 
> With "certain" you probably mean HiDPI support? Feel free to provide a correct HiDPI implementation for 3 different platforms that doesn't break any existing API contract and has "correct" implementation without hacks. I believe everyone complaining about SWT doesn't really know how complicated the implementation is and underestimates the effort required to do things "correct", whatever it means.

Not having the time nor energy to fix an incorrect implementation is not the same as saying "it is correct because it does not cause issues in SWT". It is fine to accept buggy implementations but not to state that they are correct.

> Correct in the sense of which contract? If you know how "correct" implementation should look like, please share your knowledge with us.

HashMap for example. Or should the equals check in HashMap also contain an extra case for SWT images?

iloveeclipse:
>  It is fine to accept buggy implementations but not to state that they are correct.

Once again, **please** point to the bug in the SWT Image.equals() implementation and **explain** the problem in the ticket. If you know something we don't know, share your knowledge with us. We will be happy to fix SWT code, we just want to know **what** to fix.

> HashMap for example. Or should the equals check in HashMap also contain an extra case for SWT images?

It might be a surprise, but one [can](https://github.com/eclipse-platform/eclipse.platform.ui/pull/682#issuecomment-1496192503) use HashMap with SWT Images, and no, we haven't patched HashMap in SWT.

So **please** if you have some insights how a "correct" SWT Image.equals() implementation should look like, provide this **specification** (with specification I mean not a question).

merks:
@sravanlakkimsetti Thank you for explaining the logic behind the current implementation.  From that perspective, it does not seem unreasonable.

I wonder though, what would be broken if we deleted the hashCode and equals methods from Image?  Or in other words, in what cases is it important that two different Image instances compare as being .equals?  We can certainly see a number of cases where that's problematic, which we could address with various workarounds, e.g., using == instead of .equals.  But this does beg the question, for what is the current specialized equality implementation important?  

The mutability of instances via dispose() is the fundamental problem here because while a disposed image might be equal to a non-disposed instance, it certainly doesn't function the same way...

BTW, I noticed that TooIItem does not use .equals when setting the image:
```
public void setImage (Image image) {
	checkWidget();
	if (this.image == image) return;
```

PR #701, opened 2023-04-04
Revert reusable ImageFileNameProvider/ImageDataProvider instances of

### LINKED ISSUE #679: Reused editors lost their icon in 4.26

Dani-Hub:
URLImageDescriptor and FileImageDescriptor

This fixes #679: It restores the Image#equals contract, which depends on non-equal providers per image.

The corresponding tests are modified, too: Now they assert non-sameness of the provider instead of sameness.

To preserve the testability of their underlying providers, the two affected ImageDescriptor classes return them now through their getAdapter methods (And so so similarly by creating fresh instances).

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
@@ -12,7 +12,8 @@
  *     IBM Corporation - initial API and implementation
  *     Patrik Suzzi <psuzzi@gmail.com> - Bug 483465
  *     Christoph Läubrich - Bug 567898 - [JFace][HiDPI] ImageDescriptor support alternative naming scheme for high dpi
- *     Daniel Kruegler - #376, #396, #398, #399, #401
+ *     Daniel Kruegler - #376, #396, #398, #399, #401,
+ *                       #679: Ensure that fresh ImageFileNameProvider/ImageDataProvider instances are created to preserve Image#equals invariant.
  *******************************************************************************/
 package org.eclipse.jface.resource;
 
@@ -42,7 +43,58 @@ import org.eclipse.swt.graphics.ImageFileNameProvider;
  * public API. Use ImageDescriptor#createFromURL to create a descriptor that
  * uses a URL.
  */
-class URLImageDescriptor extends ImageDescriptor implements IAdaptable, ImageFileNameProvider, ImageDataProvider {
+class URLImageDescriptor extends ImageDescriptor implements IAdaptable {
+
+	private static class URLImageFileNameProvider implements ImageFileNameProvider {
+
+		private final String url;
+
+		public URLImageFileNameProvider(String url) {
+			this.url = url;
+		}
+
+		@Override
+		public String getImagePath(int zoom) {
+			URL tempURL = getURL(url);
+			if (tempURL != null) {
+				final boolean logIOException = zoom == 100;
+				if (zoom == 100) {
+					return getFilePath(tempURL, logIOException);
+				}
+				URL xUrl = getxURL(tempURL, zoom);
+				if (xUrl != null) {
+					String xResult = getFilePath(xUrl, logIOException);
+					if (xResult != null) {
+						return xResult;
+					}
+				}
+				String xpath = FileImageDescriptor.getxPath(url, zoom);
+				if (xpath != null) {
+					URL xPathUrl = getURL(xpath);
+					if (xPathUrl != null) {
+						return getFilePath(xPathUrl, logIOException);
+					}
+				}
+			}
+			return null;
+		}
+
+	}
+
+	private static class URLImageDataProvider implements ImageDataProvider {
+
+		private final String url;
+
+		public URLImageDataProvider(String url) {
+			this.url = url;
+		}
+
+		@Override
+		public ImageData getImageData(int zoom) {
+			return URLImageDescriptor.getImageData(url, zoom);
+		}
+
+	}
 
 	private static long cumulativeTime;
 
@@ -79,6 +131,10 @@ class URLImageDescriptor extends ImageDescriptor implements IAdaptable, ImageFil
 
 	@Override
 	public ImageData getImageData(int zoom) {
+		return getImageData(url, zoom);
+	}
+
+	private static ImageData getImageData(String url, int zoom) {
 		URL tempURL = getURL(url);
 		if (tempURL != null) {
 			if (zoom == 100) {
@@ -238,7 +294,10 @@ class URLImageDescriptor extends ImageDescriptor implements IAdaptable, ImageFil
 			if (InternalPolicy.DEBUG_LOAD_URL_IMAGE_DESCRIPTOR_2x) {
 				if (!InternalPolicy.DEBUG_LOAD_URL_IMAGE_DESCRIPTOR_DIRECTLY) {
 					try {
-						return new Image(device, (ImageFileNameProvider) this);
+						// We really want a fresh ImageFileNameProvider instance to make
+						// sure the code that uses created images can use equals(),
+						// see Image#equals
+						return new Image(device, new URLImageFileNameProvider(url));
 					} catch (SWTException | IllegalArgumentException exception) {
 						// If we fail fall back to the slower input stream method.
 					}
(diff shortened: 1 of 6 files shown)


iloveeclipse:
@Dani-Hub : could you please do following :    

1) Add comments in the image creation code that mentions that we **want** new instances of providers to make sure the code that uses created images can use equals()
2) Optionally (but high wanted) : could you provide a test case that opens plain text editor on few files in the way "Search" it does, and checks that editor tab doesn't loose the icon?

Dani-Hub:
> @Dani-Hub : could you please do following :
> 
>     1. Add comments in the image creation code that mentions that we **want** new instances of providers to make sure the code that uses created images can use equals()

Sure, will do that (this evening)

>     2. Optionally (but high wanted) : could you provide a test case that opens plain text editor on few files in the way "Search" it does, and checks that editor tab doesn't loose the icon?

I need a bit of guidance here. Can you show me a test project that does something similarly? (This sounds more like an integration test, but maybe I misunderstand.)

iloveeclipse:
> I need a bit of guidance here. Can you show me a test project that does something similarly?

Will do it later.

But you are right, more easier would be a simple unit test for images created with each of the descriptors - they shouldn't be equal if created with same descriptor.

Unfortunately such unit test was not added initially but we can add it now.

Dani-Hub:
> > I need a bit of guidance here. Can you show me a test project that does something similarly?
> 
> Will do it later.
> 
> But you are right, more easier would be a simple unit test for images created with each of the descriptors - they shouldn't be equal if created with same descriptor.
> 
> Unfortunately such unit test was not added initially but we can add it now.

Not a problem, I'll add such tests.

Dani-Hub:
@iloveeclipse I have amended my previous PULL request. Can you please review?

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -42,7 +43,78 @@
...
+						return getFilePath(xPathUrl, logIOException);
+					}
+				}
+			}
+			return null;
+		}
+
+	}
+
+	private static class URLImageDataProvider implements ImageDataProvider {
```
iloveeclipse, line 84: What is the reason to duplicate the method implementation here?
Why can't we do something like this:

```
	private class URLImageDataProvider implements ImageDataProvider {
		@Override
			return URLImageDescriptor.this.getImageData(zoom);
		}
	}
```

or extract the main part of `getImageData(zoom)` to a static method and do something like this:

```
	private static class URLImageDataProvider implements ImageDataProvider {
		private final String url;

		public URLImageDataProvider(String url) {
			this.url = url;
		}
		
		@Override
			return URLImageDescriptor.getImageData(zoom, url);
		}

	}

	@Override
		return getImageData(zoom, url);
	}
	
	static ImageData getImageData(int zoom, String url) {
		URL tempURL = getURL(url);
		if (tempURL != null) {
			if (zoom == 100) {
				return getImageData(tempURL);
			}
			URL xUrl = getxURL(tempURL, zoom);
			if (xUrl != null) {
				ImageData xdata = getImageData(xUrl);
				if (xdata != null) {
					return xdata;
				}
			}
			String xpath = FileImageDescriptor.getxPath(url, zoom);
			if (xpath != null) {
				URL xPathUrl = getURL(xpath);
				if (xPathUrl != null) {
					return getImageData(xPathUrl);
				}
			}
		}
		return null;
	}
```

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/images/UrlImageDescriptorTest.java
```diff
@@ -10,22 +10,42 @@
...
 import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
 import org.eclipse.swt.graphics.ImageFileNameProvider;
 
 import junit.framework.TestCase;
 
 public class UrlImageDescriptorTest extends TestCase {
 
+	/**
```
iloveeclipse, line 33: Please add this class to `AllImagesTests`

iloveeclipse (changes requested):
Two things: 
1) Avoid code duplication (see comment above)
2) Add the `UrlImageDescriptorTest` class to AllImagesTests (not your fault, it wasn't added before)

Dani-Hub:
> Two things:
> 
>     1. Avoid code duplication (see comment above)
> 
>     2. Add the `UrlImageDescriptorTest` class to AllImagesTests (not your fault, it wasn't added before)

All good points - implemented now.

Can you help me understand, why the IT is still failing?

iloveeclipse:
> Can you help me understand, why the IT is still failing?

Sorry, I've missed that part completely.
The manifest from org.eclipse.e4.ui.tests needs a version update +100 in the last segment.

Dani-Hub:
> > Can you help me understand, why the IT is still failing?
> 
> Sorry, I've missed that part completely. The manifest from org.eclipse.e4.ui.tests needs a version update +100 in the last segment.

I assume you mean org.eclipse.jface.tests, right? Thanks for the hint.

iloveeclipse:
> I assume you mean org.eclipse.jface.tests, right? Thanks for the hint.

No, sorry, I see your PR is not based on master head. Please rebase and problem should disappear.

tests/org.eclipse.jface.tests/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.jface.tests
-Bundle-Version: 1.4.0.qualifier
+Bundle-Version: 1.4.100.qualifier
```
iloveeclipse, line 5: Please undo, that is not needed.

iloveeclipse:
See above: **This branch is out-of-date with the base branch**

Please fetch from master and rebase your branch on top of master!

Dani-Hub:
> See above: **This branch is out-of-date with the base branch**
> 
> Please fetch from master and rebase your branch on top of master!

I have done that (I hope correctly), but the fourth test still fails with 

> Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jface.tests: Only qualifier changed for (org.eclipse.jface.tests/1.4.0.v20230405-1654). Expected to have bigger x.y.z than what is available

Doesn't that mean that I still need to increase the z part (micro) of the version number?

iloveeclipse:
> Doesn't that mean that I still need to increase the z part (micro) of the version number?

yes. sorry for wrong advice, have checked wrong manifest

Dani-Hub:
> yes. sorry for wrong advice, have checked wrong manifest

No problem, your guidance is very much appreciated.

Dani-Hub:
@iloveeclipse I fixed the version constraint mismatch, but now the fourth test still fails because of test failures in other test projects. I'm not so familiar with them to determine the problem, albeit my guess is that these are unrelated to my PULL request.

iloveeclipse:
continuous-integration/jenkins/pr-head is what to watch for. The rest is currently broken.

iloveeclipse (approved):
Thanks, LGTM

PR #714, opened 2023-04-12
Use multi-line String in CSS demo

vogella:


marcushoepfner (approved):

vogella:
Linux green is a go AFAIK

marcushoepfner:
JFYI: @iloveeclipse said, if "continuous-integration/jenkins/pr-head" is ok you could merge.

But we should definitely fix these failing/flickering tests.

PR #715, opened 2023-04-12
Use multi-line String in CSS tests

vogella:


marcushoepfner:
There are test methods which have only two lines css (e.g. CascadeTests).
However, can you change them as well. With that, it is consistent in all methods. And the "\n" is gone.

jukzi:
please provide a commit message. Is that an automated conversion?

vogella:
> There are test methods which have only two lines css (e.g. CascadeTests). However, can you change them as well. With that, it is consistent in all methods. And the "\n" is gone.

Will do that with another commit, this one was done with the automatic JDT cleanup and I think it is bad to mix automatic changes with manual changes.

vogella:
> please provide a commit message. Is that an automated conversion?

Done

marcushoepfner:
I think the two failed tests are flickering.
I retrigger the build to check this

vogella:
@HeikoKlare in case you are intested also in helping with test stability in eclipse.platform.ui (which would be great), we see "random" errors on a regular basis in our PR in this repo too.

marcushoepfner (approved):

Issue #718, opened 2023-04-14
Exceptions in menu extensions affect platform


Bananeweizen:
I have upgraded an IDE containing the EasyShell plugin to 2023-06 M1. EasyShell relies on ResourceNavigator being there (which was removed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/d3a7abfd3b3a0c9a863675604d4bf8e81735ef82) and fails in creating its ShowIn extensions, reported at https://github.com/anb0s/EasyShell/issues/265.

The much bigger issue however is that this exception affects the platform itself. My progress view is broken, and my repository view is broken (EasyShell extends that context menu too).

![grafik](https://user-images.githubusercontent.com/406876/232064847-dde8e420-8e94-4d3b-8fed-2f0d583d252b.png)
To my understanding we need to have some additional try-catch somewhere in the ShowIn extension creation code.

```
java.lang.NoClassDefFoundError: org/eclipse/ui/views/navigator/ResourceNavigator
	at de.anbos.eclipse.easyshell.plugin.commands.DefineCommands.createContributionItems(DefineCommands.java:51)
	at org.eclipse.ui.internal.menus.ContributionFactoryGenerator.compute(ContributionFactoryGenerator.java:73)
	at org.eclipse.e4.ui.workbench.renderers.swt.ContributionRecord.mergeFactoryIntoModel(ContributionRecord.java:272)
	at org.eclipse.e4.ui.workbench.renderers.swt.ContributionRecord.mergeIntoModel(ContributionRecord.java:216)
	at org.eclipse.e4.ui.workbench.renderers.swt.MenuManagerRenderer.processAddition(MenuManagerRenderer.java:551)
	at org.eclipse.e4.ui.workbench.renderers.swt.MenuManagerRenderer.generateContributions(MenuManagerRenderer.java:531)
	at org.eclipse.e4.ui.workbench.renderers.swt.MenuManagerRenderer.processContributions(MenuManagerRenderer.java:499)
	at org.eclipse.e4.ui.workbench.renderers.swt.MenuManagerRenderer.createWidget(MenuManagerRenderer.java:417)
	at org.eclipse.e4.ui.internal.workbench.swt.MenuService.registerMenu(MenuService.java:67)
	at org.eclipse.ui.internal.PopupMenuExtender.registerE4Support(PopupMenuExtender.java:184)
[...]

PR #719, opened 2023-04-15
Catch contribution factory exceptions (#718)

Bananeweizen:
Avoid that exceptions in contribution factories (which are contributed by extension points) break the complete context menu or view part using that factory.


iloveeclipse:
Thanks.

Issue #702, opened 2023-04-04
Onboarding shows incomplete text on start-up in new workspace


trancexpress:
See screen recording:

https://user-images.githubusercontent.com/24752155/229907264-e3964d97-ff45-4ea7-a1b4-66a9cd4652f4.mp4

This is when starting an SDK build in a new workspace:

```
Eclipse SDK
Version: 2023-06 (4.28)
Build id: I20230404-0600
```

I'm on Linux with KDE, using Breeze Dark and theming enabled (with dark theme).

I'm not sure how the text is expected to appear, but there seems to be enough space to show the text with the default window size.

iloveeclipse:
See original enhancement request here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/633

@marcushoepfner, could you please check this?
I wonder if the image & text can **both** be scaled to available area.

Alternative would be to reduce default image sizes to fit the default window size - some space for text.

marcushoepfner:
Not sure how to achieve this. 
A reasonable minimum screen resolution was expected actually. 
I mean the image is 350 by 350. Plus text plus commands. I guess this is ~500 px height. Plus problems view height plus top toolbar height. Thats not much pixels  isn't?

What screen resolution are you on @trancexpress ?

BeckerWdf:
I just checked what vscode does. If the windows hight is not sufficient they remove the shortcuts. The icon is visible all the time. And if he hight get's even lesser then the icon is scaled down.

marcushoepfner:
> @marcushoepfner, could you please check this? I wonder if the image & text can **both** be scaled to available area.

This would re-introduce #678. There is a space of 50px on each side.

vogella:
Maybe hide the image if the size of the container is below a certain value? Something like:

```
if (container is less than 640 px in height) {

GridData gridData = new GridData(SWT.FILL, SWT.FILL, true, true);
gridData.exclude = true;
label.setLayoutData(gridData)
}
```

Scaling the image seems not a bit to design effort for this startup case.

trancexpress:
> I just checked what vscode does. If the windows hight is not sufficient they remove the shortcuts. The icon is visible all the time. And if he hight get's even lesser then the icon is scaled down.

Good for VS Code but isn't that their logo? I assume they'll want to show it.

For Eclipse users, the hot keys are more important?

BeckerWdf:
[...]

PR #720, opened 2023-04-17
Reduce example onboarding image size to 250x250

marcushoepfner:
Issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/702


marcushoepfner:
Image seems to be slightly different. This time I used https://github.com/eclipse-platform/eclipse.platform.images/blob/master/org.eclipse.images/eclipse-svg/org.eclipse.ui/icons/full/eview16/default_persp.svg.
I don't remember which one I used the first time.

Anyway, since it is an example I hope this is ok.

Issue #729, opened 2023-04-26
Filters contributed to org.eclipse.ui.ide.markerSupport overwrite each other


Bananeweizen:
The extension parsing mechanism for markerSupport.filters has a bug in merging contributed filters. I wrongly [assigns all extensions to the same array element](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java#L150) because of not incrementing the index, thereby overwriting all except the last contributed filter. Since the array holding the extensions gets sized correctly, you will afterwards see "empty" marker filters.
![grafik](https://user-images.githubusercontent.com/406876/234563057-187664d0-8349-4fe1-bfd7-f05d0124bc3f.png)

PR #731, opened 2023-04-26
Correctly merge marker filter contributions (#729)

Bananeweizen:
Avoid low level array index counting, and just join the 2 arrays/collections via Streams.


mickaelistria:
thanks!

PR #736, opened 2023-04-29
Clean-up OSGi package imports and remove unused dependencies

HannesWell:
- Use properly closed version range
- Remove unused packages
- Remove unnecessary imports of org.osgi.service.component.annotations (they are only necessary at compile-time and Tycho adds them for compilation on the fly).
- Mark optional package-imports that are definitely not optional as such

- Remove unused dependencies


HannesWell:
Addtionally this now removes the generated OSGi-DS files, unnecessary metadata & fix registrations.

Activate PDE DS processing for all Plugins that have OSGi Declarative Services components. Remove their description files from git, since they are generated by PDE-DS in the IDE workspace or Tycho in the build.
Remove unnecessary imports of org.osgi.service.component.annotations (they are only necessary at compile-time and Tycho adds them for compilation on the fly).

This also ensures that the following E4 IContextFunction are registered as OSGi services (which was probably intended but not done right):
- CommandServiceCreationFunction
- HandlerServiceCreationFunction
- BindingServiceCreationFunction

PR #743, opened 2023-05-03
Add preference for disposing closed browser hovers after some time

### LINKED ISSUE #742: Add preference for disposing browser hovers that have been closed for some time
See https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/528, on Linux we have a problem with opening many Java editor tabs and opening hovers in them. Since each editor keeps an `BrowserInformationControl` object, an SWT `Browser` object is kept alive for each editor tab. This in turn means 1 WebKit process per opened Java editor (in which a JavaDoc hover was shown).

With https://github.com/eclipse-platform/eclipse.platform.text/issues/191 we add internal API, which when used ensures `BrowserInformationControl` is disposed after being closed for some time.

We also want to add a preference that can be used by client code (such as JDT), in combination with the new API.

trancexpress:
Add a preference for disposing closed browser hovers after some time.

The preference can be used by client code, combined with the fix for: https://github.com/eclipse-platform/eclipse.platform.text/issues/191

Fixes: #742

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
@@ -674,4 +674,15 @@ public interface IWorkbenchPreferenceConstants {
 	 */
 	String RESOURCE_RENAME_MODE_DIALOG = "dialog"; //$NON-NLS-1$
 
+	/**
+	 * Preference for the <i>advised</i> time (in ms) after a closed browser hover
+	 * is to be disposed.
+	 * <p>
+	 * The integer default value for this preference is: <code>-1</code>.
+	 * Non-positive values indicate closed hovers are not disposed automatically.
+	 * </p>
+	 *
+	 * @since 3.129
+	 */
+	String DISPOSE_CLOSED_BROWSER_HOVER_TIMEOUT = "disposeClosedBrowserHoverTimeout"; //$NON-NLS-1$
 }
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPreferenceInitializer.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPreferenceInitializer.java
@@ -124,6 +124,8 @@ public class WorkbenchPreferenceInitializer extends AbstractPreferenceInitialize
 		node.putBoolean(IPreferenceConstants.SHOW_KEYS_ENABLED_FOR_MOUSE_EVENTS, false);
 		node.putInt(IPreferenceConstants.SHOW_KEYS_TIME_TO_CLOSE, 3000);
 
+		node.putInt(IWorkbenchPreferenceConstants.DISPOSE_CLOSED_BROWSER_HOVER_TIMEOUT, -1);
+
 		node.put(IWorkbenchPreferenceConstants.RESOURCE_RENAME_MODE,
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
```diff
@@ -325,4 +325,15 @@ public interface IPreferenceConstants {
 	 */
 	String SHOW_KEYS_TIME_TO_CLOSE = "showCommandKeys_timeToClose"; //$NON-NLS-1$
 
+	/**
+	 * Preference for the <i>advised</i> time (in ms) after a closed browser hover
+	 * is to be disposed.
+	 * <p>
+	 * The integer default value for this preference is: <code>-1</code>.
+	 * Non-positive values indicate closed hovers are not disposed automatically.
```
iloveeclipse: "Zero and non-positive values"

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
```diff
@@ -325,4 +325,15 @@ public interface IPreferenceConstants {
 	 */
 	String SHOW_KEYS_TIME_TO_CLOSE = "showCommandKeys_timeToClose"; //$NON-NLS-1$
 
+	/**
```
iloveeclipse: `IPreferenceConstants` is internal interface, we can't use it from JDT/Xtext.

`org.eclipse.ui.IWorkbenchPreferenceConstants` is  better one.

iloveeclipse (changes requested):

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
trancexpress: Wasn't the idea that we are adding a preference that was only internally visible? Or you mean `IPreferenceConstants` is not exported via its package?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
trancexpress: > ([mathematics](https://en.wiktionary.org/wiki/mathematics), of a quantity) Not [positive](https://en.wiktionary.org/wiki/positive); either [zero](https://en.wiktionary.org/wiki/zero) or [negative](https://en.wiktionary.org/wiki/negative).

https://en.wiktionary.org/wiki/nonpositive

If the wording is not clear, maybe replace with "zero and negative"?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
iloveeclipse: > Wasn't the idea that we are adding a preference that was only internally visible?

The idea was that we don't provide a **UI** for that preference (yet).

iloveeclipse: > Or you mean IPreferenceConstants is not exported via its package?

This is also true, at least it wasn't accessible in my [draft PR](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/548) using this preference.

iloveeclipse (approved):

PR #745, opened 2023-05-04
Remove javax bundles from org.eclipse.e4.rcp Feature

HannesWell:
During builds these annotation providing bundles are pulled in as transitive dependencies any ways. Removing them from this features simplifies the exchange of the atucally providing bundle and therefore the migration to Jakarta.

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1056

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -144,20 +144,6 @@
          unpack="false"/>
 
    <plugin
-         id="javax.inject"
-         download-size="0"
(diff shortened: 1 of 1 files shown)


jukzi:
LGTM

HannesWell:
Thanks.
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1057 Tycho is configured to include sources for each bundle in the I-builds repo to that repo.
Consequently, the sources for transitively pulled in bundles are present.

Issue #275, opened 2022-08-18
org.eclipse.ui.tests.dialogs.UIEditWorkingSetWizardAuto.testEditPage fails on windows in the I-build intermittently


ktatavarthi:
The test org.eclipse.ui.tests.dialogs.UIEditWorkingSetWizardAuto.testEditPage failed in the IBuild [I20220817-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220817-1800/) on [Windows](https://download.eclipse.org/eclipse/downloads/drops4/I20220817-1800/testresults/html/org.eclipse.ui.tests_ep425I-unit-win32-java11_win32.win32.x86_64_11.html) with the below error

```
Problems encountered while deleting resources.

junit.framework.AssertionFailedError: Problems encountered while deleting resources.
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.TestCase.fail(TestCase.java:223)
at org.eclipse.ui.tests.dialogs.UIWorkingSetWizardsAuto.deleteResources(UIWorkingSetWizardsAuto.java:102)
at org.eclipse.ui.tests.dialogs.UIWorkingSetWizardsAuto.doTearDown(UIWorkingSetWizardsAuto.java:169)
at org.eclipse.ui.tests.harness.util.UITestCase.tearDown(UITestCase.java:214) 
```

ktatavarthi:
The test org.eclipse.ui.tests.dialogs.UIEditWorkingSetWizardAuto.testEditPage failed in the IBuild [I20220824-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220824-1800/) on [Windows](https://download.eclipse.org/eclipse/downloads/drops4/I20220824-1800/testresults/html/org.eclipse.ui.tests_ep425I-unit-win32-java11_win32.win32.x86_64_11.html) with the below error

```
Problems encountered while deleting resources.

junit.framework.AssertionFailedError: Problems encountered while deleting resources.
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.TestCase.fail(TestCase.java:223)
at org.eclipse.ui.tests.dialogs.UIWorkingSetWizardsAuto.deleteResources(UIWorkingSetWizardsAuto.java:102)
at org.eclipse.ui.tests.dialogs.UIWorkingSetWizardsAuto.doTearDown(UIWorkingSetWizardsAuto.java:169)
[...]

PR #755, opened 2023-05-12
Open new WorkbenchWindows in the same monitor as the active window #739

### LINKED ISSUE #739: New Workbench Window is not opened on the device on which it was called
When opening a new Workbench window over window -> new window or over the api  PlatformUI.getWorkbench().openWorkbenchWindow(defaultPerspective, adapter); the window is not opened on the same screen (device as the workbenchwindow on which the command was started) It seems as it is opened on the primary screen or where the application was initially started. The new workbench window should be opened on the same screen as the current active workbench window is, or it should give a possibility to tell on which screen the new workbench window should be opened. The only way to do so is using discouraged api:
org.eclipse.ui.internal.Workbench.openWorkbenchWindow(IAdaptable, IPerspectiveDescriptor, MWindow, boolean)
or using the emodelservice clone Methode which has other bugs.
@HeikoKlare : can you look for this please.

fedejeanne:
When opening a new WorkbenchWindow in any of the methods `org.eclipse.ui.internal.Workbench.openWorkbenchWindow(...)`, its X/Y coordinates are not explicitly set. This makes the newly created window sometimes appear in a different monitor than the calling window and confuses/annoys the user.

This change ensures that the X and Y coordinates of the newly created Workbench window are set and that the window lands in the vicinity of the calling window.

Solves #739

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -147,8 +147,10 @@ import org.eclipse.swt.graphics.DeviceData;
 import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.Point;
+import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Listener;
+import org.eclipse.swt.widgets.Monitor;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IDecoratorManager;
 import org.eclipse.ui.IEditorInput;
@@ -1462,9 +1464,13 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 					Point size = result.getWindowConfigurer().getInitialSize();
 					window.setWidth(size.x);
 					window.setHeight(size.y);
+
+					placeNearActiveShell(window);
+
 					application.getChildren().add(window);
 					application.setSelectedElement(window);
 				}
+
 				ContextInjectionFactory.inject(result, windowContext);
 				windowContext.set(IWorkbenchWindow.class, result);
 			} finally {
@@ -1481,6 +1487,81 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 			result.fireWindowOpened();
 		}
 		return result;
+	}
+
+	private void placeNearActiveShell(MWindow window) {
+		if (getDisplay() == null) {
+			return;
+		}
+
+		Shell activeShell = getDisplay().getActiveShell();
+		if (activeShell == null) {
+			return;
+		}
+
+		Monitor currentMonitor = findMonitorThatContainsMostOf(activeShell.getBounds());
+
+		final int padding = 20;
+		Rectangle paddedMonitorBounds = shrink(currentMonitor.getBounds(), padding);
+
+		final int offsetToExistingShell = 100;
+		Rectangle newShellBounds = new Rectangle(activeShell.getBounds().x + offsetToExistingShell,
+				activeShell.getBounds().y + offsetToExistingShell, window.getWidth(), window.getHeight());
+
+		moveIntoBounds(newShellBounds, paddedMonitorBounds);
+
+		window.setX(newShellBounds.x);
+		window.setY(newShellBounds.y);
+	}
+
+	private static Rectangle shrink(Rectangle rectangle, int padding) {
+		return new Rectangle(rectangle.x + padding, rectangle.y + padding, rectangle.width - 2 * padding,
+				rectangle.height - 2 * padding);
+	}
+
+	/**
+	 * @param rectangle a rectangle (e.g. the bounds of the shell)
+	 * @return The monitor that contains the biggest portion of the rectangle or the
+	 *         primary monitor if the rectangle is outside all monitors.
+	 */
+	private Monitor findMonitorThatContainsMostOf(Rectangle rectangle) {
+		Monitor bestFittingMonitor = getDisplay().getPrimaryMonitor();
+		int maxIntersectionArea = 0;
+
+		for (Monitor monitor : getDisplay().getMonitors()) {
+			Rectangle intersection = new Rectangle(rectangle.x, rectangle.y, rectangle.width, rectangle.height);
+			intersection.intersect(monitor.getBounds());
+
+			int insersectionArea = intersection.width * intersection.height;
+			if (insersectionArea > maxIntersectionArea) {
+				bestFittingMonitor = monitor;
+				maxIntersectionArea = insersectionArea;
+			}
+		}
+
+		return bestFittingMonitor;
+	}
+
+	private static void moveIntoBounds(Rectangle rectangleToMove, Rectangle bounds) {
+		// move into bounds if it's too far to the right
+		if (rectangleToMove.x + rectangleToMove.width > bounds.x + bounds.width) {
+			rectangleToMove.x = bounds.x + bounds.width - rectangleToMove.width;
+		}
+
+		// move into bounds if it's too far to the left
+		if (rectangleToMove.x < bounds.x) {
+			rectangleToMove.x = bounds.x;
+		}
+
+		// move into bounds if it's too far down
+		if (rectangleToMove.y + rectangleToMove.height > bounds.y + bounds.height) {
+			rectangleToMove.y = bounds.y + bounds.height - rectangleToMove.height;
+		}
+
+		// move into bounds if it's too far up
+		if (rectangleToMove.y < bounds.y) {
+			rectangleToMove.y = bounds.y;
+		}
 	}
 
 	/*


mickaelistria:
Shouldn't we skip all form of extra precessing if the newly created shell is on the same monitor as its parent?

merks:
I would expect to see org.eclipse.swt.widgets.Display.getMonitors() used somewhere.   You should note that the Display.getBounds() is effectively the union of the Monitor.getBounds() of all the monitors, but there may be parts of the overall Display bounds that are not covered by any Monitor's bounds because monitors can have different sizes and strange arrangements...

fedejeanne:
@mickaelistria I'm not sure what you mean, would you please elaborate?

@merks you're right, I hadn't thought about that. I changed the code and searched for the current monitor in [fb65ab2](https://github.com/eclipse-platform/eclipse.platform.ui/pull/755/commits/fb65ab2811640dc8120cc67d6792eb5da7a5ac3d)

mickaelistria:
@fedejeanne My previous comment can be ignored, I see the scope of the proposal has slightly change from "move to same monitor" to "place to nearby shell" and it is IMO a good thing. So no need to add the guard I suggested earlier.

fedejeanne:
Hello all. 
It's been quiet in the past few weeks. Any chances this PR gets approved/merged? :)

Thank you in advance!

mickaelistria:
@fedejeanne We're currently waiting for the next 4.29 version to be open for business (all build stuff ready) before merging code in.

fedejeanne:
I'm sorry, I don't mean to be pushy but I see things got a bit too quiet again. Any chance this PR gets merged in the upcoming release (4.29)? :-)

Please let me know if there's anything I can do on my side to help ease the process.

Thank you in advance!

BeckerWdf:
On what OS does this happen. I just tried to reproduce the issue on my macBook that has an external monitor connected.
Regardless on which monitor the "original" eclipse window is, when I do Window > New Windows everything looks fine for me. The new window is created on the same monitor as the "original" eclipse window.

merks:
On Windows (I'm still on 10), it always opens on the main monitor.

fedejeanne:
@BeckerWdf, I too was able to reproduce this in Windows 10.

@MarcelBoerner what OS are you using?

fedejeanne:
@BeckerWdf : @MarcelBoerner uses Windows 11 (see [comment](https://github.com/eclipse-platform/eclipse.platform.ui/issues/739#issuecomment-1600862568))

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1490,82 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+			return;
+
+		Rectangle shellBounds = activeShell.getBounds();
+		Monitor currentMonitor = findMonitorThatContains(shellBounds.x, shellBounds.y);
+
+		Rectangle currentMonitorBounds = currentMonitor.getBounds();
+
+		window.setX(calculateXInsideMonitor(shellBounds, currentMonitorBounds));
+
+		window.setY(calculateYInsideMonitor(shellBounds, currentMonitorBounds));
```
HeikoKlare: The implemented functionality makes sense to me.
I only find the way in which it is implemented quite complicated to comprehend. For me it would make more sense to split up the flow as follows:
1. Calculate the desired value for the new window position (current shell position + 100 or so)
2. If necessary, adapt the new window position to fit into the monitor bounds

It could look something like this:
```java
final int offsetToExistingShell = 100;
Rectangle newShellBounds = activeShell.getBounds();
newShellBounds.x += offsetToExistingShell;
newShellBounds.y += offsetToExistingShell;
moveIntoBounds(newShellBounds, currentMonitorBounds);
window.setX(newShellBounds.x);
window.setY(newShellBounds.y);
```
with the helper method:
```java
private static void moveIntoBounds(Rectangle rectangleToMove, Rectangle bounds) {
	if (!bounds.union(rectangleToMove).equals(bounds)) {
		rectangleToMove.x = bounds.x + bounds.width - rectangleToMove.width;
		rectangleToMove.y = bounds.y + bounds.height - rectangleToMove.height;
	}
}
```

HeikoKlare (commented):
The implemented functionality looks good to me. On my system (Win 11) I can reproduce the "faulty"/unexpected placement of the workbench window before the fix and the correct placement afterwards.
I only have one comment regarding code style (see below).

Since the API for which this method is implemented does not specify any behavior regarding the placement of a new workbench window (see `IWorkbench.openWorkbenchWindow`), it may not be essential to have a regression test for this behavior. However, if we agree that the changed implementation is how we expect the `Workbench` implementation to behave, we may still add a test to become aware of regressions?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Thank you for the feedback. I incorporated some of your changes with a few adaptations in the method `moveIntoBounds`. The changes I introduced make sure that:
* The new window only "bounces back" in the coordinate that left the bounds of the current monitor (_e.g._ if the new coordinates are too far down then only the _y_-coordinate will be corrected)
* The top-left corner of the new coordinates are always inside the bounds of the current monitor.

Hope this helps!

fedejeanne:
Test failures documented in #906 and #907

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,63 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+
+		Rectangle newShellBounds = activeShell.getBounds();
+		Monitor currentMonitor = findMonitorThatContains(newShellBounds.x, newShellBounds.y);
+
+		Rectangle currentMonitorBounds = currentMonitor.getBounds();
+
+		final int offsetToExistingShell = 100;
+		newShellBounds.x += offsetToExistingShell;
+		newShellBounds.y += offsetToExistingShell;
+		moveIntoBounds(newShellBounds, currentMonitorBounds);
```
HeikoKlare: When testing the patch, the behavior due to these lines seems to be unintended to me: It takes the size of the original shell instead of the new shell (initialized with a default size) for computing the bounds.

I'd propose to initialize the new shell bounds instead as follows to achieve the expected behavior:
```java
final int offsetToExistingShell = 100;
Rectangle newShellBounds = new Rectangle(
				activeShell.getBounds().x + offsetToExistingShell,
				activeShell.getBounds().y + offsetToExistingShell,
				window.getWidth(),
				window.getHeight());
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,63 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+				return m;
+		}
+
+		return getDisplay().getPrimaryMonitor();
+	}
+
+	private static void moveIntoBounds(Rectangle rectangleToMove, Rectangle bounds) {
+		// move into bounds if it's too far to the right
+		if (rectangleToMove.x + rectangleToMove.width > bounds.x + bounds.width)
+			rectangleToMove.x = bounds.x + bounds.width - rectangleToMove.width - 100;
```
HeikoKlare: I find it a little unintuitive that the new window bounces back for some magic padding of the display (100 px), but only if it would leave the display otherwise. In case the initial coordinates of the new window have a padding of only 1 px to the display bounds, it would not bounce back.
In case such a padding is wanted, I would prefer to always ensure it and not only if the window would exceed the display otherwise.
The easiest solution (and most comprehensible anyway) would be to call call this methods with bounds that already reflect the wanted padding rather than having some magic value subtracted within the method.

In terms of code style: please always enclose expression blocks in curly braces, even they only contain one statement. There have already been enough severe bugs by leaving them out 🙂

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,63 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+
+		// move into bounds if it's too far to the left
+		if (rectangleToMove.x < bounds.x)
+			rectangleToMove.x = bounds.x;
```
HeikoKlare: Isn't this dead code? The display bounds x and y values are by construction lower than the values of the new shell anyway.

HeikoKlare (changes requested):
I have some minor comments, but also one behavioral aspect I assume to be unintuitive and maybe unwanted, so that I would like to have it changed.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Nope, if the current shell is too far to the left (_e.g._ x = -300) then `rectangleToMove` will also be outside the display bounds. Same thing with the y-coordinate.

It should not happen often but it could happen and if it does, I find it helpful to at least have the upper-left corner of the window inside the monitor so the user can grab and move the window around.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Thanks for hint. I took your suggestion in 3a3210d663b582050f6d192d4e27bfd728425cea

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: > The easiest solution (and most comprehensible anyway) would be to call call this methods with bounds that already reflect the wanted padding rather than having some magic value subtracted within the method.

Good point, I can do that :-) 
I called `moveIntoBounds` using the desired padding in the 2nd parameter in 3a3210d663b582050f6d192d4e27bfd728425cea . I also reduced the padding to 20 px, it should suffice.

fedejeanne: I almost forgot about the brackets. I added them in d5aadd4e8eb8834bb0410966a22181f5f87938a4

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
HeikoKlare: I see, thanks for pointing that out. I don't even think that it is too unlikly to happen, so this correction is necessary.

But this points me to some other behavior I would not expect: In case the current workbench window is placed left of the leftmost monitor (i.e., `x < 0`), then the new window will be placed on the primary monitor rather than on the leftmost one. I would expect the new workbench to placed on the leftmost monitor in that situation. So technically, I would expect `x` and `y` to be set to zero if they are negative before calculating any assignment to a monitor.
As a consequence, handling negative values would not necessary at this place anymore.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,73 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+		Rectangle newShellBounds = new Rectangle(activeShell.getBounds().x + offsetToExistingShell,
+				activeShell.getBounds().y + offsetToExistingShell, window.getWidth(), window.getHeight());
+
+		moveIntoBounds(newShellBounds, paddedMonitorBounds);
+
+		window.setX(newShellBounds.x);
+		window.setY(newShellBounds.y);
+	}
+
+	private Rectangle shrink(Rectangle rectangle, int padding) {
```
HeikoKlare:
```suggestion
	private static Rectangle shrink(Rectangle rectangle, int padding) {
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,73 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+		if (getDisplay() == null)
+			return;
+
+		Shell activeShell = getDisplay().getActiveShell();
+		if (activeShell == null)
+			return;
+
+		Monitor currentMonitor = findMonitorThatContains(activeShell.getBounds().x, activeShell.getBounds().y);
+
+		int padding = 20;
```
HeikoKlare:
```suggestion
		final int padding = 20;
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,73 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+	 *
+	 * @param x the coordinate "X"
+	 * @param y the coordinate "Y"
+	 * @return the first monitor that contains the point or, if none is found, the
+	 *         primary monitor.
+	 */
+	private Monitor findMonitorThatContains(int x, int y) {
+		for (Monitor m : getDisplay().getMonitors()) {
+			if (m.getBounds().contains(x, y))
+				return m;
```
HeikoKlare:
```suggestion
			if (m.getBounds().contains(x, y)) {
				return m;
			 }
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Done in 0928fe739ec594c99b28aff9155d71242693aa15

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Done in 0928fe739ec594c99b28aff9155d71242693aa15

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Good point, that was a bug in the method `findMonitor`. I corrected it in 2521dd5850586f9c75d833114f9abab937019c15 by replacing the implementation to look for the monitor that holds the biggest portion of the current window. The default monitor is still used as back up in case the current window is outside all monitors... if that's even possible

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Done in 2521dd5850586f9c75d833114f9abab937019c15

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: I agree with your suggestion and I added the brackets.

Would it make sense to set this configuration for all projects in this repo just like I did back then for `eclipse.platform` (see [this PR](https://github.com/eclipse-platform/eclipse.platform/pull/443/))?

HeikoKlare: That's a good idea. I would be in favor of having that as part of the configuration, as I only see benefits in enclosing expression blocks in braces.
However, I am not sure if that's a common point of view, so I would propose to have an issue for a general discussion on that before providing a PR.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
HeikoKlare: I like that solution. It's probably closest to the behavior a user might expect.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,93 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+		window.setX(newShellBounds.x);
+		window.setY(newShellBounds.y);
+	}
+
+	private static Rectangle shrink(Rectangle rectangle, int padding) {
+		return new Rectangle(rectangle.x + padding, rectangle.y + padding, rectangle.width - 2 * padding,
+				rectangle.height - 2 * padding);
+	}
+
+	private Monitor findMonitor(Rectangle r) {
```
HeikoKlare: The code became harder to read because of this non-speaking method name. I would actually propose to integrate this functionality as fallback behavior into `findMonitorThatContainsMostOf`. You could simply initialize the monitor variable with the primary monitor, which makes the code more concise and still valid (in case the rectangle is not part of any monitor, the primary monitor is one that contains most of the rectangle).

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,93 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+		// capable of...
+		return getDisplay().getPrimaryMonitor();
+	}
+
+	/**
+	 * @param r a rectangle (e.g. the bounds of the shell)
+	 * @return The monitor that contains the biggest portion of the rectangle or
+	 *         <code>null</code> if the rectangle is outside all monitors.
+	 */
+	private Monitor findMonitorThatContainsMostOf(Rectangle r) {
```
HeikoKlare: please use speaking name instead of `r`

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1483,6 +1489,93 @@ IWorkbenchWindow createWorkbenchWindow(IAdaptable input, IPerspectiveDescriptor
...
+		Monitor monitor = null;
+		int maxIntersectionArea = 0;
+
+		for (Monitor m : getDisplay().getMonitors()) {
```
HeikoKlare: please use speaking name instead of `m`

HeikoKlare (commented):
I really like the behavior now. Thanks, @fedejeanne!
I did not find a case in which the function does not behave as expected (by me) anymore.

I just have some minor final code-style remarks.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Good idea! I created #928

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: 👍 
Fixed in d4b306849226490ad5626ba3e0cdba42f194ff5e

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: 👍 
Done in f15c036f30e9a2f990fd36b4dbc610d33cebae28

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: 👍 
Done in f15c036f30e9a2f990fd36b4dbc610d33cebae28

HeikoKlare (approved):
Looks good to me now.

HeikoKlare:
I verified the behavior on Windows, which is the platform with the actual issue addressed by this PR.
Although I do not see anything in the implementation that may break behavior on other platforms, should someone using Linux or MacOS try the changes on those platforms?

BeckerWdf:
> I verified the behavior on Windows, which is the platform with the actual issue addressed by this PR. Although I do not see anything in the implementation that may break behavior on other platforms, should someone using Linux or MacOS try the changes on those platforms?

I can try to test this on macOS tomorrow.

BeckerWdf:
> > I verified the behavior on Windows, which is the platform with the actual issue addressed by this PR. Although I do not see anything in the implementation that may break behavior on other platforms, should someone using Linux or MacOS try the changes on those platforms?
> 
> I can try to test this on macOS tomorrow.

I performed a manual test on macOS. I don't see any issues with this PR. Looks good to me.

fedejeanne:
> I verified the behavior on Windows, which is the platform with the actual issue addressed by this PR. Although I do not see anything in the implementation that may break behavior on other platforms, should someone using Linux or MacOS try the changes on those platforms?

As luck would have it, I happen to be at home today and I have a laptop with Linux here. I can test this today and report back :-)

fedejeanne:
> > I verified the behavior on Windows, which is the platform with the actual issue addressed by this PR. Although I do not see anything in the implementation that may break behavior on other platforms, should someone using Linux or MacOS try the changes on those platforms?
> 
> As luck would have it, I happen to be at home today and I have a laptop with Linux here. I can test this today and report back :-)

It works as expected in Linux too 👍 
Ready to be merged ✔

HeikoKlare:
Thanks for testing on MacOS and Linux, @BeckerWdf and @fedejeanne.

The failing tests in the latest build are unrelated to this PR and documented in #275 and #907.

Issue #757, opened 2023-05-15
Fix random failing of PipelineChainTest


fedejeanne:
Some tests in `PipelineChainTest` fail randomly. This is one such failures (see [check](https://github.com/eclipse-platform/eclipse.platform.ui/pull/715/checks?check_run_id=13336157987) for PR #715):

```
Wrong query sequence for interceptAdd expected:<ACGFBDE> but was:<null>
java.lang.AssertionError: Wrong query sequence for interceptAdd expected:<ACGFBDE> but was:<null>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.eclipse.ui.tests.navigator.PipelineChainTest._testInterceptAdd(PipelineChainTest.java:136)
	at org.eclipse.ui.tests.navigator.PipelineChainTest.testInterceptAdd(PipelineChainTest.java:113)
...
```

jukzi:
I also ran into this exception:
https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/5045869417/jobs/9051626843
```
Wrong query sequence for interceptUpdate expected:<ACGFBDE> but was:<null>
java.lang.AssertionError: Wrong query sequence for interceptUpdate expected:<ACGFBDE> but was:<null>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.eclipse.ui.tests.navigator.PipelineChainTest._testInterceptUpdate(PipelineChainTest.java:245)
	at org.eclipse.ui.tests.navigator.PipelineChainTest.testInterceptUpdateWithLabel(PipelineChainTest.java:226)
....
	at org.eclipse.equinox.launcher.Main.main(Main.java:1440)
track:  ELEMENTS R/ id: A
track:  ELEMENTS R/ id: A
track:  ELEMENTS R/ id: A
track:  ELEMENTS R/ id: A
track:  ELEMENTS R/ id: A
track:  CHILDREN P/p1 id: A1
track:  CHILDREN P/p1 id: C
track:  CHILDREN P/p1 id: G
track:  CHILDREN P/p1 id: F
track:  CHILDREN P/p1 id: B
track:  CHILDREN P/p1 id: D
track:  CHILDREN P/p1 id: E
[...]

PR #758, opened 2023-05-15
Fix random failure of test in PipelinChainTest #757

fedejeanne:
Some tests in `PipelineChainTest` fail randomly. This happens because all tests share the `Map`s in `TestPipelineProvider` and they clear them at will.

This is a typical race condition and it looks like this: 
* Test "A" fills the map
* Test "B" clears the map 
* Test "A" `assert`s that the map has some key --> **FAIL**

One example of this is:
* `testInterceptAdd()` fills `TestPipelineProvider.ADDS`
* `testInterceptRemove()` clears all maps by calling `TestPipelineProvider.clear()`
* `testInterceptAdd()` asserts that `TestPipelineProvider.ADDS` has some entry --> **FAIL**

This avoids the race condition by only removing the entries (of all maps) that are relevant to the test itself.

Solves Bug #757

### DIFF
--- a/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/PipelineChainTest.java
+++ b/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/PipelineChainTest.java
@@ -55,6 +55,7 @@ public class PipelineChainTest extends NavigatorTestBase {
 		activateAndWait(EXTENSIONS);
 	}
 
+
 	private void _initContentWithLabel() {
 		String[] EXTENSIONS = new String[] {
 				COMMON_NAVIGATOR_RESOURCE_EXT,
@@ -84,22 +85,22 @@ public class PipelineChainTest extends NavigatorTestBase {
 	public void testPipelinedChildren() throws Exception
 	{
 		_initContent();
-		_testPipelinedChildren();
+		_testPipelinedChildren("testPipelinedChildren");
 	}
 
 	@Test
 	public void testPipelinedChildrenWithLabel() throws Exception
 	{
 		_initContentWithLabel();
-		_testPipelinedChildren();
+		_testPipelinedChildren("testPipelinedChildrenWithLabel");
 	}
 
-	private void _testPipelinedChildren() throws CoreException {
+	private void _testPipelinedChildren(String testName) throws CoreException {
 		final String NEW_FOLDER = "newFolder_" + System.currentTimeMillis();
 
 		IFolder newFolder = _p1.getFolder(NEW_FOLDER);
 
-		TestPipelineProvider.reset();
+		TestPipelineProvider.reset(testName);
 		newFolder.create(true, true, new NullProgressMonitor());
 		TreeItem[] rootItems = _viewer.getTree().getItems();
 
@@ -118,7 +119,7 @@ public class PipelineChainTest extends NavigatorTestBase {
 	public void testInterceptAdd() throws CoreException
 	{
 		_initContent();
-		_testInterceptAdd();
+		_testInterceptAdd("testInterceptAdd");
 	}
 
 	/** Verifies that interceptAdd is called in the right sequence */
@@ -126,10 +127,10 @@ public class PipelineChainTest extends NavigatorTestBase {
 	public void testInterceptAddWithLabel() throws CoreException
 	{
 		_initContentWithLabel();
-		_testInterceptAdd();
+		_testInterceptAdd("testInterceptAddWithLabel");
 	}
 
-	private void _testInterceptAdd() throws CoreException {
+	private void _testInterceptAdd(String testName) throws CoreException {
 		final String NEW_FOLDER_1 = "newFolder1";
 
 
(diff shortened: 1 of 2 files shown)


HeikoKlare:
I am not sure if I understand the reasons for the failing test: why is there a race condition between two test cases? Tests are executed sequentially and not in parallel. So the only chance for a race condition is if one test performs some operations that affect another one in terms of modified program state both of them access or in terms of asynchronous execution by spawing another thread or the like.
Can you please explain what hapens in this case? I would expect something like test "B" in the example scenario being executed before "A" and performing a `reset` asynchornously, but the `reset` operation seems to be called synchronously.

fedejeanne:
I also noticed that tests run sequentially when I run them locally, but parallel execution seems like the only reasonable explanation for these test failures. I can't rule it out because I don't know all the configurations in the servers running these tests and these changes have no big repercussions other than breaking a race condition, wouldn't you agree?

HeikoKlare:
I agree that the failure is probably a result of a race condition. However, to properly fix it I would like to understand the cause for the race condition and, if possible, have a reproducing test. From my understanding, parallel test execution cannot be the cause as test methods are not executed in parallel. The Maven build may build two projects in parallel, but their test execution will run in isolation. I have not seen any further configuration on the GH Action build infrastructure that affects the concurrency behavior yet (except the limited performence of the servers, which reveals some existing race conditions, like it may be the case here).

If I take a look at the `PipelineChainTest` class, I would expect the cause rather to be in the way the workspace is used across several tests. The test methods perform different kinds of changes in the same workspace. This can trigger a bunch of events, like automatic builds, refresh operations other background actions within the workspace. In the worst case, those are actions that were not cleaned up by tests that have been executed on the same workspace before. Similar issues have occurred and according fixes have been made in the platform repository and some information has been documented in https://github.com/eclipse-platform/eclipse.platform/issues/406.

fedejeanne:
I see. Do you have any concrete examples on which tasks were started in parallel and which tests started them? The issue #406 was resolved by 2 PRs, one of them with 48 files changed, is difficult to find the needle in the haystack :)

jukzi:
Maybe it helps to either disable autobuild during the test or run the whole test within a single workspace operation to avoid interference with other operations?

HeikoKlare:
Either of the two approaches may fix the failing test. But without knowing about the actual reason, it will only be a try. And it may only hide that there is still some issue in terms of missing test isolation. Note that the cause must not necessarily be in the `PipelineChainTest` but can also be in any of the other test cases within the same test project that have affected the workspace.

I would propose to start a debug build, e.g., by adding some logging to the `reset` function that prints by whom the function is called. Then it may become easier to identify the reason for the concurrent reset. Since the test seems to fail quite often, re-running the GitHub Action for only a few times should lead to an occurrence of the failure with some further logging for identifying the actual reason. I know this requires further effort, but I assume it will be quite low and outweigh the benefit of hopefully identifying the actual reason for the fail.

fedejeanne:
@HeikoKlare I like the idea of the debug run, I reverted the tests to their original form but added some information to identify the caller of `reset`.

[This comment](https://github.com/eclipse-platform/eclipse.platform.ui/issues/757#issuecomment-1557341063) got me thinking though, if calls to the `track` method are missing then the problem must be that they weren't intercepted in the first place i.e. that the extensions were either not bound or not activated (see methods `_initContent` / `_initContentWithLabel`). What do you think, @jukzi ?

jukzi:
May also be that the whole notification did not work. I hope you debug output will bring light into it.

fedejeanne:
Test failures documented in #789 and #275

fedejeanne:
Since the PR #816 is more likely to **solve** the issue #757 , what would you guys say if I simply change the intention of this PR to only log debug information instead of solving the issue?

I mean the changes introduced in this PR don't really change the behavior of the test (they are "non-disrupting") and they would help debugging in case of future test failures.

Let me know if you agree so I can change the description of this PR (don't worry, the original description will be preserved at the bottom for better understanding)

HeikoKlare:
If the debug information is useful here, it can of course be added (I am not deep enough into the referenced issue to judge).
But in that case I would propose to have a new PR that ...
1. reflects its intention from the beginning and this PR can be kept (closed) for documentation, and
2. implements some improvements:
- Use proper logging instead of `System.out.println`. I see that the extended test utilities already use `println` statements, but I am not in favor of adding further ones (and would even propose to replace the existing one if there is no good reason for having it).
- The test name can be retrieved by a `TestName` rule instead of duplicating the test name as a string inside the test method.
- I have to admit that I do not see why passing the test name to the `reset` method will be useful here: the `reset` method is called synchronously by some test methods. There is no asynchronous execution of that method that can lead to its execution while another test method is already executed (as stated in the original PR description). So I would like to have some justification for that extension.

fedejeanne:
I created a new PR (#847) and added the changes you suggested (using a `TestRule` and a proper logging mechanism). 

I logged the name of the method when clearing the maps. The reason is that it will help me clearly see if something went wrong while cleaning up (the log messages also indicate how many elements the map had right before emptying it). Also, in case the information does not help, reverting the code to its previous state is way easier than trying to find an error in the dark. 

In other words: it's cheaper to make the necessary changes to help debug and then revert them than it is to keep searching for the error without the debug information.

fedejeanne:
This PR was superseded by #847

PR #765, opened 2023-05-18
Avoid animated refresh of markers table

Bananeweizen:
When switching between different resolutions in the quick fix dialog, the marker list is refreshed with animation (e.g. after each item being removed/added there is a redraw of the control). That looks bad and blocks the UI thread for quite some time (with a bigger number of markers).


vogella:
I think this has been fixed in JFace direct. Closing therefore. Please reopen if I'm wrong

BeckerWdf:
What's the PR where this got fixed?

Bananeweizen:
3b9c02c4e1f8364bfb05f66fc5d046ab4d2ff8a2 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/927

PR #767, opened 2023-05-19
Re-parent the toolbar to the new composite when shown on a tab #230

### LINKED ISSUE #230: Toolbar buttons in views disappear after resetting perspective
Observed in Eclipse 4.23, 4.24, but not in older versions, such as 4.16

To reproduce:
- Open Eclipse Java perspective 
- Choose Window->Perspective->Save Perspective As...
- Keep the original name (Java) or set a new name, e.g. "Java 2"
- Choose Window->Perspective->Reset Perspective
- Observer buttons in view toolbars disappear, e.g. the Task List view, the Outline view, etc.

buttons
View toolbar before resetting the perspective
buttons-disappear
No toolbar/buttons after resetting the perspective

Restarting Eclipse causes the toolbar to reappear

oleosterhagen:
While resetting a perspective, tab folders are recreated. These tab folders also have a new composite for the toolbars on the top right corner. The existing toolbar from the old tab folder has to be re-parented to this new composite.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/230


oleosterhagen:
@iloveeclipse Thank you for coming back to this issue after the finalization of Eclipse 2023-06 :smiley:

Issue #723, opened 2023-04-20
Hide Perspective Onboarding Icon when not having enough space


BeckerWdf:
On small screens / small IDE windows the height of the empty editor area might not be sufficient to show the icon and the key bindings.
I such situations we should simply hide the icon and only show the key bindings.

See:

> > The hiding logic when there is not enough space can be implemented anyway.
> 
> What about this? Should we handle this in this issue or create a new one for it?

If you plan work on this, please create new one. If not, we can close this issue as we are OK with current state.

_Originally posted by @iloveeclipse in https://github.com/eclipse-platform/eclipse.platform.ui/issues/702#issuecomment-1511382305_

PR #772, opened 2023-05-25
Hide onboarding image when space is too less

marcushoepfner:
Issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/723

The onboarding image is hidden when there is less than 400 px vertical space for the tab folder.

Also centers image and text. Otherwise looks strange, refer to debug perspective onboarding which has a rather short text.


marcushoepfner:
Also centers the onboarding image and the text. Ok to fix this here or shall I create another PR for that?

And in general, can someone with layout knowledge have a look. I have the impression that it is NOT flickering, but better if someone additionally checks.

E.g. change height of editor area or eclipse while an editor is opened.
Open an editor, change height and close editor. And so on...

I tested all this already and as said, for me it looks good.

BeckerWdf:
@marcushoepfner: Can you pls. have a look into the discussion in #788?

marcushoepfner:
Just renamed the branch. Unfort. this caused a close or this PR. Continuing in #820

Issue #773, opened 2023-05-26
NPE in IDEWorkbenchWindowAdvisor.computeTitlePath


jukzi:
logged in 
https://download.eclipse.org/eclipse/downloads/drops4/I20230525-0350/testresults/ep428I-unit-mac64-java17_macosx.cocoa.x86_64_17/org.eclipse.ui.tests.UiTestSuite.txt

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.core.resources.IFile.getLocation()" because "file" is null
	at org.eclipse.ui.internal.ide.application.IDEWorkbenchWindowAdvisor.computeTitlePath(IDEWorkbenchWindowAdvisor.java:457)

```

a null check will do

jukzi:
logged NPE reproducable with ArbitraryPropertyTest

PR #774, opened 2023-05-26
fix NPE in computeTitlePath #773

jukzi:
reproducible with ArbitraryPropertyTest


vogella:
PMC approved. Thanks @jukzi for working on our test stability.

PR #790, opened 2023-06-02
Improve error thrown in UIWorkingSetWizardsAuto.deleteResources() #275

fedejeanne:
This PR helps identify the underlying error that makes tests like `UIEditWorkingSetWizardAuto.testEditPage()` fail (see #275).

The problem with the current code is that deep in the `CoreException` in the `catch` block (which is actually a `ResourceException`) there is a `MultiStatus` with a `null` Exception in it. This PR searches for a meaningful status inside that multi-status and uses its Exception and Stack-Trace instead, therefore providing a better stack-trace to identify where the relevant error occurred.

The objective of this PR is not to fix the test but to show a meaningful Stack-Trace to help analyze and fix the test.

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
@@ -19,6 +19,7 @@ import java.util.List;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.jface.wizard.IWizard;
 import org.eclipse.jface.wizard.IWizardPage;
 import org.eclipse.jface.wizard.WizardDialog;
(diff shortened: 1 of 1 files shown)


fedejeanne:
Link #275

BeckerWdf:
Looks good to me. Should we merge it?

fedejeanne:
_Disclaimer: I don't know if this one was directed at me (probably not) but I'm answering it anyway 😄_

> Looks good to me. Should we merge it?

... sure!

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
```diff
@@ -99,11 +100,42 @@ private void deleteResources() {
...
+	}
+
+	private IStatus findMostSevere(IStatus status) {
+		if (!status.isMultiStatus()) {
+			return status;
+		}
+
+		IStatus mostSevere = null;
+
+		for (IStatus c : status.getChildren()) {
```
HeikoKlare: Maybe you can call this something like `childStatus` rather than `c`to make it easier to read the following line.

HeikoKlare (approved):
Looks good to me. Hope that it helps to find the reason for the failing test.
Just one nitpicky comment for a possible naming improvement.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
fedejeanne: Done in 1d2565a

Issue #788, opened 2023-05-30
Align suggested Keybindings


laeubi:
Currently all text is center-aligned, what makes it hard to read
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/1331477/756be798-687d-43df-b83f-8ac1e16dc543)

I'd like to suggest to align the text and keybindings and separate them, at best the will look like keys:

![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/1331477/1efcb456-96d4-4697-919e-b10dee18696b)

merks:
I kind of like the text right aligned and the keys left aligned but the space between them ought to be significantly bigger so that one easily sees the separation.  But that's just my opinion...

laeubi:
Currently its just hard to see what is text and what is key and it could easily be mixed, also if you draw a line it seems like the image is left aligned to the text and the key actions centered to the text, this all looks quite "unsteady" ...

If one likes them to "meet at the middle", I think it should at least be aligned with with image so the form one line
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/1331477/b0d2bbe2-b0ac-4c4b-b573-24e82abcc644)

maybe even a thin line (here in green) would help to give a better separation.

BeckerWdf:
See https://github.com/eclipse-platform/eclipse.platform.ui/pull/772

marcushoepfner:
Interesting. Does not look that bad on Mac w.r.t space between command text and shortcut.
<img width="403" alt="2023-06-06_10-47-46" src="https://github.com/eclipse-platform/eclipse.platform.ui/assets/29301444/03bc89e2-763c-4eee-992f-fd6bc6754007">

Anyway, I'll provide a fix.
Center alignment of image and "onboarding text" was missing yet. I did it in #772, but will do it rather separately now.

marcushoepfner:
Refer to https://github.com/eclipse-platform/eclipse.platform.ui/pull/796#issuecomment-1578267938 how it looks now.

iloveeclipse:
Now the master is opened. Please don't forget to bump bundle version.

PR #796, opened 2023-06-06
Alignment of onboarding information

marcushoepfner:
Align all onboarding information centered.
Space between onboarding command text and short cut.

Issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/788

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
@@ -711,9 +711,11 @@ public class StackRenderer extends LazyStackRenderer {
 		GridDataFactory.create(GridData.VERTICAL_ALIGN_CENTER | GridData.HORIZONTAL_ALIGN_CENTER).grab(true, true)
 				.applyTo(onboardingComposite);
 
-		GridLayoutFactory.swtDefaults().numColumns(2).equalWidth(true).applyTo(onboardingComposite);
+		GridLayoutFactory.swtDefaults().numColumns(2).equalWidth(true).spacing(10, SWT.DEFAULT)
+				.applyTo(onboardingComposite);
 
-		GridDataFactory gridDataFactory = GridDataFactory.swtDefaults().indent(SWT.DEFAULT, 10).span(2, 1);
+		GridDataFactory gridDataFactory = GridDataFactory.swtDefaults().align(SWT.CENTER, SWT.CENTER)
+				.indent(SWT.DEFAULT, 10).span(2, 1);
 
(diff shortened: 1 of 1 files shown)


marcushoepfner:
Is aligned like this on Mac now:


Could someone check windows pls?

BeckerWdf (approved):
I am also a mac user..

merks:
It looks like this on Window 10:

image

And looks very nice in my opinion.

iloveeclipse:
Marcus, 4.29 master is not open for merge yet! Please wait for the mail saying that!

marcushoepfner:
Still wonder why I have to remember such things. Why isn't there a machine (build job) telling me that? My brain has too less capacity for such things...
I'll revert

iloveeclipse:
> Why isn't there a machine (build job) telling me that?

Because it is broken and because no one fixed it yet. Until it works please follow mails, as everyone else is doing.

laeubi:
> > Why isn't there a machine (build job) telling me that?
> 
> Because it is broken and because no one fixed it yet. Until it works please follow mails, as everyone else is doing.

I don't think it is broken, Quiet period is simply not declared as a freeze-day.... the docs here: https://wiki.eclipse.org/Category:SimRel-2023-06 here says:

>> **No builds during "quiet period"**. It is assumed all code is done by the end of RC2

so I at least from that plan I can't see any freeze requirements.

BeckerWdf:
> > > Why isn't there a machine (build job) telling me that?
> > 
> > 
> > Because it is broken and because no one fixed it yet. Until it works please follow mails, as everyone else is doing.
> 
> I don't think it is broken, Quiet period is simply not declared as a freeze-day.... the docs here: https://wiki.eclipse.org/Category:SimRel-2023-06 here says:
> 
> > > **No builds during "quiet period"**. It is assumed all code is done by the end of RC2
> 
> so I at least from that plan I can't see any freeze requirements.

So maybe we can improve this.

laeubi:
> I'll revert

I don't think its worth a revert, everything has already happen and RC2a is promoted, so unless there will be a RC2b planed it doesn't matter and only adds even more noise...

> So maybe we can improve this.

It depends, as mentioned in the linked issue Quiet Period means no builds (I assume I-Builds...), it does not mean "no merges" ...

iloveeclipse:
Sure it means no merges, for all non-release process related commits.
E.g. bundle version Updates do not work yet, etc.

laeubi:
> Sure it means no merges, for all non-release process related commits.

The it should be simple to add "Quiet" to the check if desired

> E.g. bundle version Updates do not work yet, etc.

They would have working if the prepare workflow where have been executed what was missing in this release cycle...

laeubi:
> Sure it means no merges, for all non-release process related commits.

Now we are in Quit Period (as of https://www.eclipse.org/eclipse/platform-releng/buildSchedule.html ) but master is declared to be open ... it seems for me that the either one has to adjust the calendar or just watch the "master open" separately and an automated check can't really do that...

PR #798, opened 2023-06-06
fix wrong double-checked locking

jukzi:
see https://rules.sonarsource.com/java/RSPEC-2168


merks:
I don't suppose this and a bunch of the other places are performance critical, but this could have been fixed by making the field volatile too, right?

jukzi:
not that simple: for example with volatile the assigned Object has to be fully constructed before assigning. i.e. assignment has to be last step in double checking.

merks:
Good point!

jukzi:
ignoring random failing tests:
ImportTestUtils$TestBuilder.assertFullBuildWasDone https://github.com/eclipse-platform/eclipse.platform.ui/issues/811
PipelineChainTest._testInterceptUpdate https://github.com/eclipse-platform/eclipse.platform.ui/issues/757
SorterTest.testComparatorAsSorter https://github.com/eclipse-platform/eclipse.platform.ui/issues/809

Issue #794, opened 2023-06-05
Random failing of SorterTest.testSorterContentAdd()


fedejeanne:
The test failed during a [custom build](https://github.com/fedejeanne/eclipse.platform.ui/actions/runs/5155048913/jobs/9284234729#step:7:17176) in my forked repository. It was during the freeze period for 4.29 and the code was current, so I'm creating an issue.

Here's the stack trace:

```
org.eclipse.ui.tests.navigator.SorterTest.testSorterContentAdd  Time elapsed: 0.365 s  <<< FAILURE!
org.junit.ComparisonFailure: expected:<Blue[Added]Parent> but was:<Blue[]Parent>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.eclipse.ui.tests.navigator.SorterTest.testSorterContentAdd(SorterTest.java:291)
```

fedejeanne:
@iloveeclipse you were way faster than me 😅
Thank you for merging!

PR #803, opened 2023-06-06
Use typed for-each loop

jukzi:
Avoiding type-casts makes code easier to read.


jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform.ui/issues/794 on linux

PR #800, opened 2023-06-06
Fix random failure of SorterTest.testSorterContentAdd()

fedejeanne:
The test relies on `TestLabelProviderBlue`, which is not being instantly initialized. This behavior seems to be known and accepted since the test `LabelProviderTest.testChangeActivation()` was already using a busy wait.

This PR extracts a new method into `NavigatorTestBase` and uses it in both tests. The method also asserts that the required label provider has been initialized, which helps debugging.

fixes #794

PR #816, opened 2023-06-12
Wait for extensions to be activated in PipelineChainTest #757

fedejeanne:
Let all tests in `PipelineChainTest` wait for up to 1 minute for the proper extensions to be activated before running. 

This PR makes sure that the necessary interceptors are present before running the tests and keeps the test from failing randomly. If the proper interceptors are not active by the time the tests run then the tests fail with this error: `Wrong query sequence for interceptAdd expected:<ACGFBDE> but was:<null>` (the `null` means that there were no interceptors in play).

Waiting for extensions to be loaded seems to be a common and accepted pattern since some other tests in the plugin also need to wait for extensions to be loaded (I also mentioned this in #800).

Solves #757

### DIFF
--- a/tests/org.eclipse.ui.tests.navigator/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.ui.tests.navigator/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %bundlename
 Bundle-SymbolicName: org.eclipse.ui.tests.navigator;singleton:=true
-Bundle-Version: 3.7.100.qualifier
+Bundle-Version: 3.7.200.qualifier
 Bundle-Localization: plugin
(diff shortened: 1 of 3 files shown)


iloveeclipse:
```
Only qualifier changed for (org.eclipse.ui.tests.navigator/3.7.100.v20230614-1123). Expected to have bigger x.y.z than what is available in baseline (3.7.100.v20230405-0754)
```

If this is a first change in org.eclipse.ui.tests.navigator bundle in 4.29, the bundle version must be bumped +100.

fedejeanne:
Thank you for the tip, @iloveeclipse , I bumped the bundle version in 7f341fc81c4c8a6b4c60eb07dfd6d302b6003221 .

I also bumped it in #800. I assume this is OK, right? I mean the changes can be merged without problem.

iloveeclipse:
> I also bumped it in #800. I assume this is OK, right? I mean the changes can be merged without problem.

Yep. Thanks for looking into unstable tests.

PR #823, opened 2023-06-14
remove redundant Math.round(int)

jukzi:


bundles/org.eclipse.ui.views.properties.tabbed/src/org/eclipse/ui/internal/views/properties/tabbed/view/TabbedPropertyList.java
```diff
@@ -656,8 +656,7 @@ public void controlResized(ControlEvent e) {
 	 * Calculate the number of tabs that will fit in the tab list composite.
 	 */
 	protected void computeTabsThatFitInComposite() {
-		tabsThatFitInComposite = Math
-			.round((getSize().y - 22) / getTabHeight());
+		tabsThatFitInComposite = (getSize().y - 22) / getTabHeight();
```
mickaelistria, line 659: I suspect the initial intention of this code was to get the "round" value and the current code was a bug. Just replacing `22` (int) with `22.` (double should restore this intention.

jukzi: Guessing that 22 is the tab height, then the "tabsThatFit"  *completely* "InComposite" needs to round down.

mickaelistria: Yes, the code here is indeed between risky and incorrect in 1st place. Your patch wouldn't make it sensibly worse, so no objection.

PR #826, opened 2023-06-14
Avoid reusing of not thread-safe SimpleDateFormat

jukzi:
SimpleDateFormat.format(new Date());
returns same string as immutable and thread-safe
  DateTimeFormatter.format(new Date().toInstant())


jukzi:
known random fail https://github.com/eclipse-platform/eclipse.platform.ui/issues/794

PR #838, opened 2023-06-19
Make DeviceResourceDescriptor a parameterized type

HannesWell:
This PR adds typed overloads of ResourceManager.get() for images, fonts and colors that return the corresponding specific resource type:
- `Image get(ImageDescriptor descriptor)`
- `Color get(ColorDescriptor descriptor)`
- `Font get(FontDescriptor descriptor)`

This avoids the need to downcast the resource returned by `get(DeviceResourceDescriptor)` and therefore makes caller's code less verbose and more type-safe.

@vogella you have tutorial about this topic, what do you think?

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeviceResourceDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeviceResourceDescriptor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2015 IBM Corporation and others.
+ * Copyright (c) 2004, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Hannes Wellmann - Parameterize DeviceResourceDescriptor with the described resource type
  *******************************************************************************/
 package org.eclipse.jface.resource;
 
@@ -25,11 +26,13 @@ import org.eclipse.swt.graphics.Device;
  * two equal descriptors at hand, e.g. decorating an identical object.
  * </p>
  *
+ * @param <R> The resource's type described by this descriptor
+ *
  * @see org.eclipse.jface.resource.ResourceManager
  *
  * @since 3.1
  */
-public abstract class DeviceResourceDescriptor {
+public abstract class DeviceResourceDescriptor<R> {
 	private final boolean shouldBeCached;
 
 	/**
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2015 IBM Corporation and others.
+ * Copyright (c) 2004, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Hannes Wellmann - Parameterize DeviceResourceDescriptor with the described resource type
  *******************************************************************************/
 package org.eclipse.jface.resource;
 
@@ -84,19 +85,13 @@ public abstract class ResourceManager {
 	 * this method.
 	 * </p>
 	 *
-	 * <p>
-	 * Callers may safely downcast the result to the resource type associated with
-	 * the descriptor. For example, when given an ImageDescriptor, the return value
-	 * of this method will always be an Image.
-	 * </p>
-	 *
 	 * @since 3.1
 	 *
 	 * @param descriptor descriptor for the resource to allocate
 	 * @return the newly allocated resource (not null)
 	 * @throws DeviceResourceException if unable to allocate the resource
 	 */
-	public abstract Object create(DeviceResourceDescriptor descriptor);
+	public abstract <R> R create(DeviceResourceDescriptor<R> descriptor);
 
 	/**
 	 * Deallocates a resource previously allocated by {@link #create(DeviceResourceDescriptor)}.
@@ -108,7 +103,7 @@ public abstract class ResourceManager {
 	 *
 	 * @param descriptor identifier for the resource
 	 */
-	public abstract void destroy(DeviceResourceDescriptor descriptor);
+	public abstract <R> void destroy(DeviceResourceDescriptor<R> descriptor);
 
 	/**
 	 * Returns a previously-allocated resource or allocates a new one if none exists
@@ -123,12 +118,6 @@ public abstract class ResourceManager {
 	 * other code that shares them. For example, never call
 	 * {@link org.eclipse.swt.graphics.Resource#dispose()} on anything returned from
 	 * this method.
-	 * </p>
-	 *
-	 * <p>
-	 * Callers may safely downcast the result to the resource type associated with
-	 * the descriptor. For example, when given an ImageDescriptor, the return value
-	 * of this method may be downcast to Image.
 	 * </p>
 	 *
 	 * <p>
@@ -151,8 +140,8 @@ public abstract class ResourceManager {
 	 *
 	 * @since 3.3
 	 */
-	public final Object get(DeviceResourceDescriptor descriptor) {
-		Object cached = find(descriptor);
+	public final <R> R get(DeviceResourceDescriptor<R> descriptor) {
+		R cached = find(descriptor);
 		return cached != null ? cached : create(descriptor);
 	}
 
@@ -171,14 +160,16 @@ public abstract class ResourceManager {
 	 * @since 3.1
 	 *
 	 * @param descriptor descriptor for the image to create
-	 * @return the Image described by this descriptor (possibly shared by other equivalent
-	 * ImageDescriptors)
+	 * @return the Image described by this descriptor (possibly shared by other
+	 *         equivalent ImageDescriptors)
 	 * @throws DeviceResourceException if unable to allocate the Image
+	 * @deprecated use {@link #create(DeviceResourceDescriptor)} instead
 	 */
+	@Deprecated(since = "3.31")
 	public final Image createImage(ImageDescriptor descriptor) {
 		// Assertion added to help diagnose client bugs.  See bug #83711 and bug #90454.
 		Assert.isNotNull(descriptor);
-		return (Image) create(descriptor);
+		return create(descriptor);
 	}
 
 	/**
@@ -198,7 +189,7 @@ public abstract class ResourceManager {
 		}
 
 		try {
-			return (Image) create(descriptor);
+			return create(descriptor);
 		} catch (DeviceResourceException | SWTException e) {
 			Policy.getLog().log(Status.warning("The image could not be loaded: " + descriptor, e)); //$NON-NLS-1$
 			return getDefaultImage();
@@ -222,7 +213,9 @@ public abstract class ResourceManager {
 	 * @since 3.1
 	 *
 	 * @param descriptor identifier for the image to dispose
+	 * @deprecated use {@link #destroy(DeviceResourceDescriptor)} instead
 	 */
+	@Deprecated(since = "3.31")
 	public final void destroyImage(ImageDescriptor descriptor) {
 		destroy(descriptor);
 	}
@@ -238,9 +231,11 @@ public abstract class ResourceManager {
 	 * @param descriptor descriptor for the color to create
 	 * @return the Color described by the given ColorDescriptor (not null)
 	 * @throws DeviceResourceException if unable to create the color
+	 * @deprecated use {@link #create(DeviceResourceDescriptor)} instead
 	 */
+	@Deprecated(since = "3.31")
 	public final Color createColor(ColorDescriptor descriptor) {
-		return (Color)create(descriptor);
+		return create(descriptor);
 	}
 
 	/**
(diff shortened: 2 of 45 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
```diff
@@ -102,51 +102,172 @@ public abstract class ResourceManager {
...
+	 *
+	 * <p>
+	 * This method should only be used for images that should remain allocated for
+	 * the lifetime of this resource-manager. To allocate shorter-lived images,
+	 * manage them with {@link #createImage(ImageDescriptor)}, and
+	 * {@link #destroyImage(ImageDescriptor)} rather than this method.
+	 * </p>
+	 *
+	 * <p>
+	 * This method should never be called on the global resource-manager, since all
```
jukzi: I don't like a method that should not be called. Better use a specialized interface for Managers that "support" the method.

iloveeclipse (dismissed):
This PR if merged will show compilation errors in 
- org.eclipse.ui.browser
- org.eclipse.ui.workbench

Please fix them in dedicated commit.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
vogella: This method already exists... But we should annotate it with @Deprecated if it should not be called (anymore).

HannesWell: > I don't like a method that should not be called. Better use a specialized interface for Managers that "support" the method.

I understand your point, but it should only not be called for the global resource-manager (e.g. returned by `JFaceResources.getResources()`) for all others it is fine. I don't have a direct use-case at hand, but even for a global manager it could make sense to call this method if one really wants a resource for the entire lifetime of the application (of course in general this would be a bad practice).

That being said, alternatively those new methods could be added to the `LocalResourceManager`. Then at least the new methods one cannot call (accidentally) on a global `DeviceResourceManager`.
But only adding it to `LocalResourceManager` would probably hinder the adoption a bit, because AFAICT usually one uses the `ResourceManager` type for variables and then does not stumble about this more strictly typed methods.

And given that there is already a `get` method, I tend to perfer to add them at the same location. If all get's were new, then yes, should they have been placed at `LocalResourceManager`.

> This method already exists... But we should annotate it with @deprecated if it should not be called (anymore).

For cases where one has a custom sub-class of `DeviceResourceDescriptor` that does handle something else like images, colors or font this method could still be relevant. And the Java-Doc of `DeviceResourceDescriptor` does not forbid direct subclassing.

HannesWell:
> Please fix them in dedicated commit.

Done.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
```diff
@@ -102,51 +102,172 @@ public abstract class ResourceManager {
...
+	 * </p>
+	 *
+	 * @param descriptor identifier for the requested images
+	 * @return the requested resource, never null.
+	 * @throws DeviceResourceException if the images does not exist yet and cannot
+	 *                                 be created for any reason.
+	 *
+	 * @since 3.31
+	 */
+	public final Image get(ImageDescriptor descriptor) {
```
jukzi: This method should get another name then get(DeviceResourceDescriptor) because ImageDescriptor extends ImageDescriptor. A call that already used get(DeviceResourceDescriptor) with an ImageDescriptor would otherwise now get another (narrower) result which can result in an error message like unnecessary typecast.
```suggestion
	public final Image getImage(ImageDescriptor descriptor) {
```

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
```diff
@@ -102,51 +102,172 @@ public abstract class ResourceManager {
...
+	 * </p>
+	 *
+	 * @param descriptor identifier for the requested fonts
+	 * @return the requested resource, never null.
+	 * @throws DeviceResourceException if the fonts does not exist yet and cannot be
+	 *                                 created for any reason.
+	 *
+	 * @since 3.31
+	 */
+	public final Font get(FontDescriptor descriptor) {
```
jukzi: getFont

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
jukzi: then the javadoc never call on "global resource-manager" is just wrong.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ResourceManager.java
```diff
@@ -102,51 +102,172 @@ public abstract class ResourceManager {
...
+	 * </p>
+	 *
+	 * @param descriptor identifier for the requested colors
+	 * @return the requested resource, never null.
+	 * @throws DeviceResourceException if the colors does not exist yet and cannot
+	 *                                 be created for any reason.
+	 *
+	 * @since 3.31
+	 */
+	public final Color get(ColorDescriptor descriptor) {
```
jukzi: getColor

jukzi (dismissed):
Overall i don't see a reason for so many methods that only do typecast. but bring a ton of new javadoc. in 
It would be more sexy to have a single
 public final <T extends Resource> T getResource(DeviceResourceDescriptor<T> descriptor) {
with
 public abstract class DeviceResourceDescriptor<T extends Resource> {

Or  - swapping the parameters  - to have a single getCached(ResourceManager) on DeviceResourceDescriptor that avoids an typed get() for each type of resource

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeviceResourceDescriptor.java
```diff
@@ -25,11 +25,13 @@
...
  * </p>
  *
+ * @param <R> The resource's type described by this descriptor
+ *
  * @see org.eclipse.jface.resource.ResourceManager
  *
  * @since 3.1
  */
-public abstract class DeviceResourceDescriptor {
+public abstract class DeviceResourceDescriptor<R> {
```
HannesWell, line 35: If this class would be created new, we could consider to use `R extends org.eclipse.swt.graphics.Resource` but this would likely break binary compatibility because the erasure would be `Resource` instead of `Object`.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeviceResourceDescriptor.java
```diff
@@ -80,7 +82,7 @@ final boolean shouldBeCached() {
 	 * @return the newly allocated resource (not null)
 	 * @throws DeviceResourceException if unable to allocate the resource
 	 */
-	public abstract Object createResource(Device device);
+	public abstract R createResource(Device device);
```
HannesWell: Only changed the `createResource()` method to return a generic type R and not `destroyResource()` to accept a generic type T, because this would mean that in the sub-classes `destroyResource()` would have to be changed to only accept `Image/Color/Font`.
But narrowing a method parameter type is AFAIK not a binary compatible change (but would probably be source-compatible in most cases), while it should be binary compatible to narrow the return type.

HannesWell (commented):
> Overall i don't see a reason for so many methods that only do typecast. but bring a ton of new javadoc. in
> It would be more sexy to have a single
> public final T getResource(DeviceResourceDescriptor descriptor) {
> with
> public abstract class DeviceResourceDescriptor {

I'm not sure if you meant exactly that, but with how I understood your proposal and thinking about it again I think the 'problem' here is that `DeviceResourceDescriptor` is not parameterized. But thanks to the way how Java Generics work we could change `DeviceResourceDescriptor` to `DeviceResourceDescriptor<R>` and let `Image/Color/FontDescriptor` extend `DeviceResourceDescriptor<Image/Color/Font>`, while maintaining binary compatibility.
Then we can change
`public final Object get(DeviceResourceDescriptor descriptor)`
to
`public final <R> R get(DeviceResourceDescriptor<R> descriptor) {`
and would have a type-save `get()` without the need to add overloads.

Since the upper-pound of the generic type `R` is `Object` this change should be binary compatible (i.e. existing jars continue to work), but please correct me if I'm wrong.
This change might lead to some Compiler warnings/errors because of unnecessary casts due to the now narrower return types, but if one has compiler warnings/errors it means one has access to the code and can fix them (which is trivial in this case). For existing jars the narrower return type should not be a problem, because AFAIK the JVM does not check that, it only checks if a reference can be assigned to a variable or parameter type.

I have quickly implemented that suggestion with the last commit.
Since `ResourceManager.create()` is also type-safe with this approach the other `createX()` methods, which have a ResourceDescriptor-subclass argument, are obsolete as well and could be deprecated too. And the destroyX() methods are actually not very meaningfull as well (I believe they were added to have a counterpart for each Resource type).

What do you think?

jukzi:
yea that looks pretty

HannesWell:
> yea that looks pretty

Glad you like it. 
I updated this PR to deprecate all of the `createX()/destroyX()` methods.
If everybody agrees with this change I'll continue with replacing all calls to the deprecated methods.

Besides deprecating `create/destroyColor(ColorDescriptor)` methods in favor of the generic methods, the create/destroyColor(RGB) methods could be deprecated as well, since `ColorDescriptor.destroyColor()` is already deprecated because colors don't need disposal anymore.
Therefore I think we could deprecate all `create/destroyColor()` methods for the same reason and generally we could recommend to not use a ResourceManager for colors in that case?

jukzi:
It's better to not deprecate methods that are used and do not cause problems to avoid useless deprecation warnings. Better only deprecate if we can not further support something. Too less people work on removing references to deprecated code :-(

iloveeclipse:
Latest PR still shows one compilation error in `org.eclipse.ui.internal.quickaccess.QuickAccessEntry.findOrCreateImage(ImageDescriptor, ResourceManager)`.

Beside this, in general should be OK. Regarding deprecation: there are "only" 64 references in SDK workspace, so it is manageable. Still, I would not add "for removal", the code is pretty old and there are no technical problems to keep one line deprecated methods stay longer.

HannesWell:
> Latest PR still shows one compilation error in `org.eclipse.ui.internal.quickaccess.QuickAccessEntry.findOrCreateImage(ImageDescriptor, ResourceManager)`.

Should be fixed now.

> Beside this, in general should be OK. Regarding deprecation: there are "only" 64 references in SDK workspace, so it is manageable. Still, I would not add "for removal", the code is pretty old and there are no technical problems to keep one line deprecated methods stay longer.

I'm fine with just deprecating it (not for removal) and agree that replacing the deprecated methods is straight forward and manageable (one can just inline the method calls). In fact I have now already replaced all calls of the deprecated methods in platform.ui with the last commit in this PR, which is the majority of callers. There are a few remaining in eclipse.platform, eclipse.platform.ui, P2 and PDE. I will take care of them once this has landed.

I have already extracted some clean-ups from this PR into https://github.com/eclipse-platform/eclipse.platform.ui/pull/850, to focus this on the purely functional changes and split all changes into three commits:
1. Parameterize `DeviceResourceDescriptor` with the described resource's type and and leverage its parameterization in `ResourceManager`. This includes the deprecation of the existing `createX/destroyX()` methods.
2. Remove now unnecessary casts
3. Replace the now deprecated methods of ResourceManager in platform.ui

For me this is ready. @jukzi, @iloveeclipse, @vogella if you agree, please approve this PR or let me know if something should be improved further.

jukzi (approved):
LGTM

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/DeviceResourceDescriptor.java
HannesWell: After double checking the assumption that narrowing the return type of a method is binary-compatible I learned, that this assumption is wrong... (for example [this article](https://www.morling.dev/blog/refining-return-type-java-methods-without-breaking-backwards-compatibility/) explains it and and I tried it out myself).

So I think we have two options now:

1. Assume that nobody besides `ResourceManager` actually calls `DeviceResourceDescriptor.createResource()` respectively the overrides because all three sub-classes for Images, Colors and Fonts have a corresponding `createImage/Color/Font()` method and that specific method is preferred because it returns the specify resource type. In my SDK workspace I see no other caller.
If we desides this is ok, I have to check what happens to custom sub-classes of DeviceResourceDescriptor.
The question is does anybody knows other code that calls createResource() or uses custom `DeviceResourceDescriptor`?
2.  Create a new method e.g. `createTypedResource()` that uses the generic return type and deprecate the old method.
3. Don't use the generic type parameter for this method and just stick with Object. Then the new generic parameter of the `DeviceResourceDescriptor` class would only be used by the `ResourceManager` and not DeviceResourceDescriptor itself.

HannesWell:
I now reverted the change of the return type of `DeviceResourceDescriptor.createResource()` to ensure absolute binary compatibility.
While I did not find any prove that this method is called from outside the Jface plugin, we cannot be sure and since we don't gain much in making its return type parameterized, i.e. we only save one cast in `DeviceResourceManager.allocate()`, I reverted that part.
The main benefits of this change are retained and a api-filter is not necessary (looks like the API-tools were right).

I'm on a short vacation from tomorrow until Sunday and plan to merge this when back and provide the replacement of all deprecated API then.

@iloveeclipse I assume your request-for-changes is fixed?

iloveeclipse:
> I assume your request-for-changes is fixed?

Not sure which one you mean, but the latest PR looks OK for me.

HannesWell:
Thank you @jukzi for the discussion, it really improved the result of this PR a lot!
And thanks Andrey for reviewing.

I will provide the PRs to replace the deprecated methods and unnecessary casts in the other repos tomorrow.

PR #839, opened 2023-06-20
BusyIndicator: lambda instead of misleading "Thread"

jukzi:


iloveeclipse:
If we are starting cleanup this code, we should also fix MT issues with non-volatile "stop" and "busyThread" fields accessed from multiple threads.

vogella (approved):

jukzi:
> If we are starting cleanup this code, we should also fix MT issues with non-volatile "stop" and "busyThread" fields accessed from multiple threads.

https://github.com/eclipse-platform/eclipse.platform.ui/pull/840

PR #840, opened 2023-06-20
BusyIndicator: fix multithreading with synchronized access

jukzi:


jukzi:
known unrelated random fail SorterTest

Issue #844, opened 2023-06-21
testEndHomeRevealCaret failed on I20230620-1800


jukzi:
Failed on all platforms. I can currently NOT reproduce locally (win) on master.

```
java.lang.AssertionError
at org.eclipse.ui.editors.tests.TextNavigationTest.testEndHomeRevealCaret(TextNavigationTest.java:161)

```

jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20230620-1800

iloveeclipse:
I see the test fail if running the entire suite, also multiple errors like this one:
```
!ENTRY org.eclipse.ui 4 0 2023-06-21 13:01:53.028
!MESSAGE Unhandled event loop exception
!STACK 0
org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:565)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:480)
	at org.eclipse.swt.widgets.Control.isVisible(Control.java:4615)
	at org.eclipse.e4.ui.workbench.renderers.swt.StackRenderer.setOnboardingControlSize(StackRenderer.java:754)
	at org.eclipse.e4.ui.workbench.renderers.swt.StackRenderer.lambda$21(StackRenderer.java:734)
	at org.eclipse.swt.events.ControlListener$2.controlResized(ControlListener.java:83)
```

The regression is caused by https://github.com/eclipse-platform/eclipse.platform.ui/pull/820. 
Reverting it fixes the test fail & exceptions above.
Moving to Platform UI.

PR #845, opened 2023-06-21
Fix widget is disposed for onboarding image

marcushoepfner:
Issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/844


BeckerWdf (approved):

iloveeclipse (approved):
Thanks for a fast fix, the test is green now.

Would be nice to have a regression test that checks if the onboarding component is properly disposed / recreated? Would be good opportunity to add it :-)

marcushoepfner:
> Thanks for a fast fix, the test is green now.
> 
> Would be nice to have a regression test that checks if the onboarding component is properly disposed / recreated? Would be good opportunity to add it :-)

I'll check it.

PR #847, opened 2023-06-22
Log debug information to help fix random failure of tests in PipelineChainTest 

fedejeanne:
(_This PR supersedes #758_)

Some tests in `PipelineChainTest` fail randomly (see #757). The cause is still not clear, that's why this PR adds some debug information to help find the error.

This PR logs the name of the test method during setup/tear-down and it also logs the name of the map being cleared along with the amount of entries the map had before it was cleared when calling `TestPipelineProvider.reset()` .

This PR does not solve #757, it only provides debug information to help diagnose it. The changes introduced in this PR can safely be reverted after the issue is solved.

### DIFF
--- a/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/PipelineChainTest.java
+++ b/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/PipelineChainTest.java
@@ -25,8 +25,11 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.swt.widgets.TreeItem;
 import org.eclipse.ui.tests.harness.util.DisplayHelper;
+import org.eclipse.ui.tests.harness.util.TestRunLogUtil;
 import org.eclipse.ui.tests.navigator.extension.TestPipelineProvider;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TestWatcher;
 
 /**
  * @since 3.3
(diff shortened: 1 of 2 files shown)


BeckerWdf (approved):

HeikoKlare:
As mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/pull/758#issuecomment-1602177023, I do still not understand why the test method name needs to be logged. Logging the call to reset and the number of cleaned elements is fine for me, but what information can you expect from logging the method name? It's generally fine for me to add that information, but I would expect an explanation for its necessity.

fedejeanne:
The name of the test method helps me:
* See if all methods cleared the maps
* See in which method a strange behavior first happened

Also, since not all test methods log any information as to where they start/end, it's not easy to deduce which methods is clearing the maps without the name of the test method as provided via this PR. Logging the name of the test method is a convenient choice: it's cheap, it's helpful and it doesn't break any functionality ;-)

Additionally, all changes introduced in this PR will be reverted once the cause of the problem has been identified and fixed so I really don't see any major issue in adding some extra logging information, wouldn't you agree? :-)

HeikoKlare:
Wouldn't it then be even more beneficial to log the test method name at test start/end to see which test the printed modifications belong to?

But, of course, we can merge this one as is and see whether it helps for debugging. Just wanted to get some understanding how this logging can help at all.
Do you still want to change something? Otherwise you could mark the PR as ready for review.

fedejeanne:
> Wouldn't it then be even more beneficial to log the test method name at test start/end to see which test the printed modifications belong to?

Good point! I can also add more debug info later without the need to pass the name of the test as parameter. Thanks for the tip, I implemented it in 079deda3e4bc728fc3795aa30121583c879242d3.

This PR is now ready for review/merge :)

HeikoKlare:
Great! Just two remarks:
1. From my understanding calls to `ILog` won't print anything by themselves unless you attach some listener doing so. At least I do not find the expected logging in the recent build logs for this PR. Maybe you can simply add the `TestRunLogUtil.LOG_TESTRUN` as a JUnit rule to have test start/end printed for debbuging purposes?
2. Please update the commit message and PR description with the recent changes.

fedejeanne:
Thank you for the tip, I had only checked locally and the logs showed up just fine.

I added the `TestRunLogUtil.LOG_TESTRUN` in 9ca9237e82e202358a6372d2d3a7ba937926fa84 and changed the commit message and description of this PR.

HeikoKlare (approved):

PR #896, opened 2023-06-29
Use new ILog.of(...) method instead of Platform.getLog(...)

laeubi:


BeckerWdf:
Can you please explain what is the advantage of "ILog.of" over the current implementation?

laeubi:
> Can you please explain what is the advantage of "ILog.of" over the current implementation?

See here for initial change and explanation:
- https://github.com/eclipse-platform/eclipse.platform/pull/498

PR #900, opened 2023-06-30
Add method to retrieve the IExpansionListeners

rubenporras:
Since `setExpanded` does not fire the listeners, we would like to have access to the listeners to be able to fire them manually.

Alternatively I could change the PR to add a method setExpanded(boolean expanded, boolean fireExpanding) that would do that without exposing the listener list.

Would that be a preferred approach?

### DIFF
--- a/bundles/org.eclipse.ui.forms/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.forms/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.ui.forms;singleton:=true
-Bundle-Version: 3.12.100.qualifier
+Bundle-Version: 3.13.0.qualifier
 Bundle-Vendor: %provider-name
 Bundle-Localization: plugin
 Export-Package: org.eclipse.ui.forms,
--- a/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
+++ b/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
@@ -885,15 +885,32 @@ public class ExpandableComposite extends Canvas {
 	}
 
 	/**
-	 * Programmatically changes expanded state.
+	 * Changes expanded state.
+	 *
+	 * Equivalent to <code>setExpanded(expanded, false)</code>. @see {@link #setExpanded(boolean, boolean)}.
 	 *
 	 * @param expanded
 	 *            the new expanded state
 	 */
 	public void setExpanded(boolean expanded) {
-		internalSetExpanded(expanded);
-		if (toggle != null)
+		setExpanded(expanded, false);
+	}
+
+	/**
+	 * Changes expanded state.
+	 *
+	 * @param expanded
+	 *            the new expanded state
+	 * @param fireExpanding
+	 *            if true, listeners will be notified
+	 *
+	 * @since 3.13
+	 */
+	public void setExpanded(boolean expanded, boolean fireExpanding) {
+		internalSetExpanded(expanded, fireExpanding);
+		if (toggle != null) {
 			toggle.setExpanded(expanded);
+		}
 	}
 
 	/**
@@ -1083,11 +1100,29 @@ public class ExpandableComposite extends Canvas {
 
 	private void toggleState() {
 		boolean newState = !isExpanded();
-		fireExpanding(newState, true);
-		internalSetExpanded(newState);
-		fireExpanding(newState, false);
+		internalSetExpanded(newState, true);
 		if (newState)
 			FormUtil.ensureVisible(this);
+	}
+
+	/**
+	 * Performs the expanded state change for the expandable control.
+	 *
+	 * @param expanded
+	 *            the new expanded state
+	 * @param fireExpanding
+	 *            if true, listeners will be notified
+	 */
+	private void internalSetExpanded(boolean expanded, boolean fireExpanding) {
+		if (fireExpanding) {
+			fireExpanding(expanded, true);
+		}
+
+		internalSetExpanded(expanded);
+
+		if (fireExpanding) {
+			fireExpanding(expanded, false);
+		}
 	}
 
 	private void fireExpanding(boolean state, boolean before) {


merks:
This is "just my opinion".  It seems odd to provide access to the list of listeners.  It seems cleaner that one should be able to change the expanded state in such a way that it notifies listeners. I find it a bit odd to begin with that this isn't just the default behavior, but I'm sure that it's this way because it's generally the case that SWT widgets/controls don't notify of programmatic changes, which is fine when you can use org.eclipse.swt.widgets.Widget.notifyListeners(int, Event) to produce an events manually.   That approach does not work for this type of event though, so I can see that something is needed...

mickaelistria:
I agree giving access to listeners seems dangerous.
Have you considered the approach suggested by @merks , manually building an Event of type SWT.Expand and using `nofifyListeners` ?

merks:
@mickaelistria 

The events are created manually here:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/c2521fd41a26168b136d6f32750b29ff03d3cdbc/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java#L1093-L1104

So org.eclipse.swt.SWT.Expand doesn't seem to apply.  I don't think there is an even event type one could use.  So I'm not sure how such an approach could be made to work...

rubenporras:
I am not very fond either of returning the listeners so we can discuss other options. If SWT.Expand does not apply, what about my other proposal of adding a method `setExpanded(boolean expanded, boolean fireExpanding)`?

merks:
@rubenporras 

I should have stated explicitly that I think your alternative approach is a reasonable (and better) alternative.  But it's all "just my opinion."  I'm not sure who else ought to have an opinion and express it...

iloveeclipse (dismissed):
Exposing internal listeners can't be API.

merks:
> Exposing internal listeners can't be API.

What do you think of the proposed alternative? I.e., a setter with a boolean with which one could ask for a corresponding notification. That seems reasonable to me.

iloveeclipse:
> What do you think of the proposed alternative? I.e., a setter with a boolean with which one could ask for a corresponding notification. That seems reasonable to me.

To answer that I have to open my IDE which I don't plan to do till Monday, sorry. The other answer was easier to made for obvious reason.

laeubi:
JFace actually notify listeners in some cases e.g. `setSelection(...)` triggers `ISelectionChangedListeners` so it sounds reasonable to notify `IExpansionListener`s when calling `setExpanded(...)` I think one can simply change this and see if any tests fail ...

merks:
I think setSelection on a viewer notifies but I don't think setSelection on a Tree notifies.  It seems an SWT design point that it be this way.   It's of course good if the tests pass, but that doesn't mean downstream applications might not be broken by unexpected new behavior...

laeubi:
> I think setSelection on a viewer notifies but I don't think setSelection on a Tree notifies.

Correct, I just wanted to note that only because SWT does not notifies, does not mean that must be the case everywhere else and here I think the Forms framework is free to have a different choice.

Because on the other hand SWT **has** `org.eclipse.swt.widgets.Widget.getListeners(int)`, so if we argue setting things should not notify because of SWT but on the other hand there should be no getter for listeners, this is at least inconsistent, because one then might argue that SWT has such method...

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -933,6 +933,16 @@ public void removeExpansionListener(IExpansionListener listener) {
...
 	}
 
+	/**
+	 * Gets the expansion listeners.
+	 *
+	 * @param listener
+	 *            the listener to remove
+	 */
+	public ListenerList<IExpansionListener> getExpansionListeners() {
+		return listeners;
```
laeubi: Maybe you can copy the listeners here into an array instead of return the (modifiable) list.

merks:
I see.  I did not know about Widget.getListeners()!  So that's a good point.  Also just notifying seems pretty reasonable to me as well.

rubenporras:
I am fine with the two options that have been discussed so far and not discarded (make a copy of the listener list before returning, adding a method  `setExpanded(boolean expanded, boolean fireExpanding)`). I think modifying `setExpanded(boolean expandedy)` to always fire is dangerous even if tests passes , as we do not know what downstream projects might do, but I would be also fine with that if you think it is best.

I do not know much about the design and conventions of eclipse UI platform, so I if that is fine for you, I would just implement whatever option you think it is the best one.

No need to hurry, waiting for Monday is of course good :)

mickaelistria:
@rubenporras Can't you mimic in your application code that approach used by some viewers (eg https://github.com/eclipse-platform/eclipse.platform.ui/blob/5469693d77ea9653dded7a287107db467df4352f/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/SimpleVirtualLazyTreeViewerTest.java#L163 ), which just creates the event and notify it programatically?

merks:
@mickaelistria 

I mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/pull/900#issuecomment-1614701453 why notifyListenersdoesn't seem applicable.  I.e., from the call hierarchy you can see that a bunch of different events from other sources trigger the notification:

image

The SWT.Expand event is documented like this:
```
	/**
	 * The expand event type (value is 17).
	 *
	 * <p>
	 * Note:
	 * </p>
	 * <ul>
	 * <li>Event.item: the TreeItem which gets expanded</li>
	 * </ul>
	 *
	 * @see org.eclipse.swt.widgets.Widget#addListener
	 * @see org.eclipse.swt.widgets.Display#addFilter
	 * @see org.eclipse.swt.widgets.Event
	 *
	 * @see org.eclipse.swt.widgets.Tree#addTreeListener
	 * @see org.eclipse.swt.events.TreeListener#treeExpanded
	 * @see org.eclipse.swt.events.TreeEvent
	 */
	public static final int Expand = 17;
```
So that seems not applicable. One would need to hunt down another of the other widgets or controls to send an event, e.g., a mouse up event?  But that seems rather indirect.

rubenporras:
@iloveeclipse , I have changed the PR with a proposal that does not expose the listeners. Is the new proposal fine?

rubenporras:
Good morning,
any feedback on the new version of the PR?

rubenporras:
Hi,

I have changed the PR, any feedback on the new version of the code?

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -887,15 +887,47 @@ public int getExpansionStyle() {
...
+
+	/**
+	 * Programmatically changes expanded state.
+	 *
+	 * @param expanded
+	 *            the new expanded state
+	 * @param fireExpanding
+	 *            if true, listeners will be fired
+	 */
+	public void setExpanded(boolean expanded, boolean fireExpanding) {
```
mickaelistria, line 909: As CI says, new API require to bump the bundle minor version (if not already done since last release) and add a `@since x.y.z` annotation in javadoc.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -887,15 +887,47 @@ public int getExpansionStyle() {
...
 
+	/**
+	 * Performs the expansion state change for the expandable control.
+	 *
+	 * @param expanded
+	 *            the expansion state
+	 * @param fireExpanding
+	 *            if true, listeners will be fired
+	 */
+	private void internalSetExpanded(boolean expanded, boolean fireExpanding) {
```
mickaelistria: I don't think it's worth a new method here. Couln't this be inlined in `setExpanded` instead?

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -887,15 +887,47 @@ public int getExpansionStyle() {
...
+		if (fireExpanding)
+			fireExpanding(expanded, true);
```
mickaelistria: Are we sure the "before" state is `true` here?

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
mickaelistria: Ah, my bad, I missed the other reference a few lines before.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
mickaelistria: Actually, it's consistent with toggleState, so no need to bother more.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
rubenporras: done

rubenporras:
@mickaelistria , could you help me with 

> "Error:  /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.ui.forms/META-INF/MANIFEST.MF [5:0]: Unqualified OSGi version 3.13.0.qualifier must match unqualified Maven version 3.13.000-SNAPSHOT for SNAPSHOT builds"

I do not understand it, as I have adapted the MANIFEST.MF

Thanks

bundles/org.eclipse.ui.forms/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.ui.forms;singleton:=true
-Bundle-Version: 3.12.100.qualifier
+Bundle-Version: 3.13.000.qualifier
```
HeikoKlare: Maybe the version has to be changed as follows.
```suggestion
Bundle-Version: 3.13.0.qualifier
```

merks: Absolutely!

rubenporras: thanks

rubenporras:
I think the failure seen in the Verify Windows Build

> Results:
> 
> Failures: 
>   UIEditWorkingSetWizardAuto>UITestCase.tearDown:214->UIWorkingSetWizardsAuto.doTearDown:200->UIWorkingSetWizardsAuto.deleteResources:94->UIWorkingSetWizardsAuto.lambda$0:96->UIWorkingSetWizardsAuto.deleteProject:105->UIWorkingSetWizardsAuto.createAssertionError:116 Problems encountered while deleting resources.
> 
> Tests run: 1554, Failures: 1, Errors: 0, Skipped: 123

is unrelated. Could we merge it or is there anything I should address before? (Unless I have missed something, I think all comments have been addressed)

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -1083,13 +1099,29 @@ public Color getTitleBarForeground() {
...
 			FormUtil.ensureVisible(this);
 	}
 
+	/**
+	 * Performs the expansion state change for the expandable control.
+	 *
+	 * @param expanded
+	 *            the expansion state
+	 * @param fireExpanding
+	 *            if true, listeners will be fired
```
HeikoKlare:
```suggestion
	 *            if true, listeners will be notified
```

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -1083,13 +1099,29 @@ public Color getTitleBarForeground() {
...
+		internalSetExpanded(newState, true);
 		if (newState)
 			FormUtil.ensureVisible(this);
 	}
 
+	/**
+	 * Performs the expansion state change for the expandable control.
+	 *
+	 * @param expanded
+	 *            the expansion state
```
HeikoKlare:
```suggestion
	 *            the new expansion state
```

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -887,11 +887,27 @@ public int getExpansionStyle() {
...
+		setExpanded(expanded, false);
+	}
+
+	/**
+	 * Programmatically changes expanded state.
+	 *
+	 * @param expanded
+	 *            the new expanded state
+	 * @param fireExpanding
+	 *            if true, listeners will be fired
```
HeikoKlare:
```suggestion
	 *            if true, listeners will be notified
```

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -1083,13 +1099,29 @@ public Color getTitleBarForeground() {
...
+	 */
+	private void internalSetExpanded(boolean expanded, boolean fireExpanding) {
+		if (fireExpanding)
+			fireExpanding(expanded, true);
+
+		internalSetExpanded(expanded);
+
+		if (fireExpanding)
+			fireExpanding(expanded, false);
```
HeikoKlare: I'd would strongly suggest to always put expression blocks into curly braces.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -887,11 +887,27 @@ public int getExpansionStyle() {
...
 	 * @param expanded
 	 *            the new expanded state
 	 */
 	public void setExpanded(boolean expanded) {
-		internalSetExpanded(expanded);
+		setExpanded(expanded, false);
+	}
+
+	/**
+	 * Programmatically changes expanded state.
```
HeikoKlare: I know that the documentation for the existing `setExpanded` method is similar, but why need to say "programatically" here? Since a user will never call a method "directly", it will always be used to "programatically" change the state.
I'd propose to remove that (also for the existing method).

HeikoKlare (commented):
The failing test is unrelated: #275 

The functionality looks good to me. I have some rather nitpicky comments on code and documentation style you may address (the branch has to be rebased on the current master anyway).

@iloveeclipse Can you also have another look? At least we need your approval to merge since you once requested changes.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
rubenporras: done

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
rubenporras: done

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
rubenporras: done

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
rubenporras: done

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
rubenporras: I also prefer it, I was just keeping the original style. Done.

HeikoKlare (approved):

merks:
I think all the issues people raised have been addressed, so I will merge this in time for m2.

rubenporras:
thanks

PR #908, opened 2023-07-03
Javadoc fixes

akurtakov:
While editing the files some other minor modifications like pattern matching and using APIs that reduce type casting.


akurtakov:
Jenkins succeeded.

PR #909, opened 2023-07-04
Avoid redraw animation on Windows

Bananeweizen:
When the markers table contains a big amount of markers, selecting another resolution leads to the typical "animated removal" of all contained items on Windows, which fully blocks the UI thread due to redrawing after each item removed.

The screen recording doesn't capture how bad it looks with the naked eye, but gives an idea:
markers refresh

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/QuickFixPage.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/QuickFixPage.java
@@ -256,8 +256,14 @@ public class QuickFixPage extends WizardPage {
 
 		resolutionsList
 				.addSelectionChangedListener(event -> {
-					markersTable.refresh();
-					setPageComplete(markersTable.getCheckedElements().length > 0);
+					Control tableControl = markersTable.getControl();
+					tableControl.setRedraw(false);
+					try {
+						markersTable.refresh();
+						setPageComplete(markersTable.getCheckedElements().length > 0);
+					} finally {
+						tableControl.setRedraw(true);
+					}
 				});
 	}
 


BeckerWdf:
The change looks good to me.

vogella:
Love it, please merge after the freeze period is over.

vogella:
Maybe adjust the commit message, as this change is not only for Windows but will affect all platforms (which is good)

Bananeweizen:
@vogella I don't have any non-Windows for testing. But whenever I fixed the same bug in other lists, a typical response was that on non-Windows systems the testers didn't even observe the issue. So basically I wanted to avoid the "I don't have that error" response with the title. :)

laeubi:
> I fixed the same bug in other lists,

Won't it then be better to make this part of JFaces `TableViewer#refresh` instead?

vogella:
> > I fixed the same bug in other lists,
> 
> Won't it then be better to make this part of JFaces `TableViewer#refresh` instead?

I would agree, shall we try this or does someone think this would "break some clients"? If yes, we could only apply it here, otherwise we could do it at the root level functionality.

laeubi:
> > > I fixed the same bug in other lists,
> > 
> > 
> > Won't it then be better to make this part of JFaces `TableViewer#refresh` instead?
> 
> I would agree, shall we try this or does someone think this would "break some clients"? If yes, we could only apply it here, otherwise we could do it at the root level functionality.

You mean clients that want strange "animations"? I think this is internal behavior of the viewers and nothing one really want to use if I understand correctly otherwise we won't call this a bugfix?

An requiring Try/finally around refresh calls do not seem something we want to recommend to every users...

vogella:
@laeubi and @Bananeweizen please check https://github.com/eclipse-platform/eclipse.platform.ui/pull/927 in which I follow @laeubi suggestion.

Bananeweizen:
Closing, since this change should not be necessary anymore with #927.

Issue #866, opened 2023-05-25
Configuration option to enable/disable auto activation of the completion proposal pop up


rubenporras:
It is not possible to disable / enable the auto activation of the completion proposal pop up or to configure it. Many users will prefer to have the feature disabled or with another delay, given that a popup can interrupts flow while typing.

The values are currently hardcoded in the Generic Editor (see `GenericEditorContentAssistant.GenericEditorContentAssistant`)

```
		setAutoActivationDelay(0);
		enableColoredLabels(true);
		enableAutoActivation(true);
		enableAutoActivateCompletionOnType(true);
```
The proposal is to have a UI menu like in the JDT:
![image](https://github.com/eclipse-platform/eclipse.platform.text/assets/43636626/87b969e2-759a-4621-ab05-fba8d85bf28e)

for the properties that the Generic Editor offers

rubenporras:
@mickaelistria , is the proposal of creating an UI for the generic editor a good one? I think this editor does not have an UI. Should it rather be an extension point so that would be used in org.eclipse.ui.internal.genericeditor.ExtensionBasedTextViewerConfiguration.getContentAssistant(ISourceViewer) to be able to have an own custom IContentAssistant?

That would mean adding an xsd to eclipse.platform.text\org.eclipse.ui.genericeditor\schema, right?

mickaelistria:
I would rather avoid providing tons of options. I'm more curious about "given that a popup can interrupts flow while typing", is there some bug to fix there? Most IDEs to have auto-completion popups and people usually don't complain, so maybe there is something that Eclipse IDE does wrong and we can improve without options?

rubenporras:
Regarding " is there some bug to fix there", yes, I think there are two (or three) bugs in there:

- One I filled already: https://github.com/eclipse/lsp4e/issues/660 
- The second one is that when the popup appears and shows "Computing Proposals", pressing arrow down/up does not allow you to go to the line below/above in the code. 
- The third one is that the dialog hides the text below, and it is by default a multi line dialog, which hides quite some text. Even if the popup disappears after some millisecs, it is distracting to see a pop up appearing and disappearing while typing.

The last two would be addressed by https://github.com/eclipse-platform/eclipse.platform.text/pull/210

Then there is the problem that it is triggered too often (basically on any character which is a letter or a number), and without delay. I do not know about any editor that does that. That highlights the bugs above and also puts a lot of stress on the server.

[...]

PR #916, opened 2023-07-07
Add extension point for generic editor content assistant look and feel.

andrewL-avlq:
Adds an extension point so that the look and feel of the generic editor content assistant can be configured by extension contribution.

### DIFF
--- a/bundles/org.eclipse.ui.genericeditor/plugin.xml
+++ b/bundles/org.eclipse.ui.genericeditor/plugin.xml
@@ -27,6 +27,7 @@
    <extension-point id="characterPairMatchers" name="%ExtPoint.characterPairMatchers" schema="schema/characterPairMatchers.exsd"/>
    <extension-point id="icons" name="%ExtPoint.iconProviders" schema="schema/icons.exsd"/>
    <extension-point id="quickAssistProcessors" name="%ExtPoint.quickAssistProcessors" schema="schema/quickAssistProcessors.exsd"/>
+   <extension-point id="contentAssistantLookAndFeel" name="%ExtPoint.contentAssistantLookAndFeel" schema="schema/contentAssistantLookAndFeel.exsd"/>
    <extension
          point="org.eclipse.ui.editors">
       <editor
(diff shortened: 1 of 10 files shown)


mickaelistria:
Can you please elaborate about the use case that requires such extension?
Would all of it be possible to cover with preferences? In the past, we've already discussed adding and supporting PreferenceStores per resource or Content-Type for other reasons (eg formatting settings), would this also cover the use-case here?

mickaelistria:
Also, fundamentally, one purpose of the generic editor is to look and feel the same for all files. It's not an editor framework. So I'm not sure this fits in the goal.

andrewL-avlq:
This is an attempt to address issues like https://github.com/eclipse-platform/eclipse.platform.ui/issues/866 in a generic extensible way. On that issue you suggest you would like to avoid having lots of preferences to set, so this extension point avoids using the PreferenceStore and is limited in scope to just the content assistant - of course an implementer of the extension point could connect to the PreferenceStore if they wished.
Happy to narrow the interface (maybe only the feature enablement methods?) and/or remove the content type specific registration if that will help.

mickaelistria:
Genericity and extensibility is not always best. Could we instead first make the editor lookup preferences for those settings, and allow RCP products to override those preferences?

andrewL-avlq:
So if I understand you correctly, you would prefer to see an implementation where the GenericContentAssistant got its look and feel settings from a PreferenceStore. Would you be ok for these preferences to be without a UI component? If so, I'll put together a PR for that.

mickaelistria:
Yes, I think an approach relying on preferences would be IMO more consistent with the expectations.
Note that the report in #866 isn't much up to date, several improvements to the current state (without preferences but changing defaults) were already submitted by @rubenporras and merged IIRC.

rubenporras:
Hi @mickaelistria ,

thanks for cross checking. @andrewL-avlq and me are working together in the same project. Yes, I did several improvements to the current state but we still need to be able to disable the autocompletion on type because for our languages it is not useful, so we decided to implement a configuration option.

Regards

mickaelistria:
OK. I still think that an approach based on Preferences would be preferred. Once we have a PreferenceStore, next steps can be to create new stores: a content-type specific store, a resource/folder specific one... and then it would be easily possible to implement preference pages to configure one store or the other if we want users to face it.

PR #924, opened 2023-07-08
Remove duplicated freeze-period check

HannesWell:
The freeze-period check is already executed in the unified workflow.

Inspired by https://github.com/eclipse-platform/eclipse.platform/pull/568.


HannesWell:
No chode changes and the remaining freeze-check kicks in as expected, thus merging.

PR #927, opened 2023-07-10
Avoid redraw animation on StructuredViewer

vogella:
When the a table contains a big amount of markers, selecting another resolution leads to the typical "animated removal" of all contained items on Windows, which fully blocks the UI thread due to redrawing after each item removed.

This animation is internal behavior of the viewers and seems something which does note help the user due to screen artifacts which are created and slows down the update process.

See also https://github.com/eclipse-platform/eclipse.platform.ui/pull/909

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StructuredViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StructuredViewer.java
@@ -1414,7 +1414,14 @@ public abstract class StructuredViewer extends ContentViewer implements IPostSel
 
 	@Override
 	public void refresh() {
-		refresh(getRoot());
+		Control control = getControl();
+		control.setRedraw(false);
+		try {
+			refresh(getRoot());
+		} finally {
+			control.setRedraw(true);
+		}
+
 	}
 
 	/**


laeubi (approved):
Looks good!

Bananeweizen (approved):
Since the setRedraw mechanism uses a counter internally, this should be completely fine for every client.

BeckerWdf:
> When the a table contains a big amount of markers, 

You mean "entries" not "markers", don't you?

Bananeweizen:
> You mean "entries" not "markers", don't you?

The complete sentence was copied directly from the other issue.

And yes, the general problem is that there is a refresh after every item removed from the viewer, which becomes very visible when going from a huge list of items to no items at all. This happens/happened in the egit file list, the installed plugins list in about, the launch config selected plugins list, and so on...

vogella:
Thanks @Bananeweizen for bring this to our attention, thanks to @laeubi for the feedback to solve this problem directly in JFace and thanks to @BeckerWdf for the review of the commit message.

I will be off my machine for the next 3-4 weeks so if something is wrong with this patch, I will not be able to participate in the discussion.

jonahgraham:
The code changed in this PR may be called from the non-UI thread so the setRedraw call now introduces a `SWTException: Invalid thread access` - see https://github.com/eclipse-platform/eclipse.platform/issues/710

laeubi:
@jonahgraham I don't think it strictly introduces one as Viewer#setInput must be called from the UI thread and if implementations do differently they need to handle the case there.

PR #937, opened 2023-07-13
USe more stable webtools update site

iloveeclipse:


releng/org.eclipse.ui.releng/platformUiTools.p2f
```diff
@@ -194,7 +194,7 @@
     </iu>
     <iu id='org.eclipse.wst.xml_ui.feature.feature.group' name='XML Editor UI' version='0.0.0'>
         <repositories size='1'>
-            <repository location='https://download.eclipse.org/webtools/CI/3.30.0/I-latest/repository/'/>
+            <repository location='https://download.eclipse.org/webtools/repository/latest/?d'/>
```
akurtakov: This ?d  url parameter troubles me.

iloveeclipse: Why? But we can remove that, seem to work without. I've just used what they linked on their site: https://download.eclipse.org/webtools/downloads/

akurtakov: No particular reason. Just not used to p2 repository urls with parameters.

merks: Probably p2 just ignores it, or the download server ignores it when p2 uses it. but I'm not sure it's good to bet on that.

PR #939, opened 2023-07-14
 Add color setting for form header messages

### LINKED ISSUE #938: Message in Form Header for Form Editors are display in wrong color when using dark mode
### Problem Description
When using the dark mode in eclipse, the messages in the form header are always displayed as error (color red #E98787) when the focus is lost. This is misleading when a warning message (color yellow) should be displayed.

For example the plugin.xml editor containing a warning message with focus:
withFocus

For example the plugin.xml editor containing a warning message when focus is lost:
withoutFocus

The problem only occurs in dark mode.

### Technical details
In the dark mode the color of the form header is hard coded to the color red in the file [bundles/org.eclipse.ui.themes/css/dark/e4-dark_partstyle.css](https://github.com/eclipse-platform/eclipse.platform.ui/blob/eeba07e1ec39431039aba654268579c94a7698e7/bundles/org.eclipse.ui.themes/css/dark/e4-dark_partstyle.css#L111) property .MPart FormHeading > CLabel > color: #E98787. This color is set by the CSS engine as soon as the focus is lost.

Christopher-Hermann:
Hard coding the color of the CLabel in the FormHeader will set the same color for all message types. To keep the correct color for the message, new colors are introduced in the settings. The CSS engine uses the colors to style the messages in the different themes.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/938

### DIFF
--- a/bundles/org.eclipse.ui.themes/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.themes/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.themes;singleton:=true
-Bundle-Version: 1.2.2200.qualifier
+Bundle-Version: 1.2.2300.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.e4.ui.css.swt.theme
--- a/bundles/org.eclipse.ui.themes/css/dark/e4-dark_partstyle.css
+++ b/bundles/org.eclipse.ui.themes/css/dark/e4-dark_partstyle.css
@@ -108,8 +108,18 @@
 }
 .MPart FormHeading > CLabel {
     background-color: #505f70;
-    color: #E98787;
 }
+
+.MPartFormHeaderCLabelError {
+	color:'#org-eclipse-ui-workbench-FORM_HEADING_ERROR_COLOR';
+}
+.MPartFormHeaderCLabelWarning {
+	color:'#org-eclipse-ui-workbench-FORM_HEADING_WARNING_COLOR';
+}
+.MPartFormHeaderCLabelInfo {
+	color:'#org-eclipse-ui-workbench-FORM_HEADING_INFO_COLOR';
+}
+
 /* ------------------------------------------------------------- */
 
 #org-eclipse-jdt-ui-SourceView StyledText,
--- a/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/internal/forms/widgets/FormHeading.java
+++ b/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/internal/forms/widgets/FormHeading.java
@@ -18,11 +18,13 @@ import java.util.Hashtable;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.ListenerList;
+import org.eclipse.e4.ui.css.swt.dom.WidgetElement;
 import org.eclipse.jface.action.IMenuManager;
 import org.eclipse.jface.action.IToolBarManager;
 import org.eclipse.jface.action.ToolBarManager;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.IMessageProvider;
+import org.eclipse.jface.resource.ColorRegistry;
 import org.eclipse.jface.resource.JFaceResources;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.CLabel;
@@ -387,7 +389,6 @@ public class FormHeading extends Canvas {
 		private IMessage[] messages;
 		private Hyperlink messageHyperlink;
 		private ListenerList<IHyperlinkListener> listeners;
-		private Color fg;
 		private int fontHeight = -1;
 		private int fontBaselineHeight = -1;
 
@@ -571,24 +572,29 @@ public class FormHeading extends Canvas {
 		}
 
 		public void setForeground(Color fg) {
-			this.fg = fg;
 			updateForeground();
 		}
 
 		private void updateForeground() {
 			Color theFg;
+			String cssClassName = null;
+			ColorRegistry colorRegistry = JFaceResources.getColorRegistry();
 
 			switch (messageType) {
 			case IMessageProvider.ERROR:
-				theFg = getDisplay().getSystemColor(SWT.COLOR_RED);
+				theFg = colorRegistry.get("org.eclipse.ui.workbench.FORM_HEADING_ERROR_COLOR"); //$NON-NLS-1$
+				cssClassName = "MPartFormHeaderCLabelError"; //$NON-NLS-1$
 				break;
 			case IMessageProvider.WARNING:
-				theFg = getDisplay().getSystemColor(SWT.COLOR_DARK_YELLOW);
+				theFg = colorRegistry.get("org.eclipse.ui.workbench.FORM_HEADING_WARNING_COLOR"); //$NON-NLS-1$
+				cssClassName = "MPartFormHeaderCLabelWarning"; //$NON-NLS-1$
 				break;
 			default:
-				theFg = fg;
+				theFg = colorRegistry.get("org.eclipse.ui.workbench.FORM_HEADING_INFO_COLOR"); //$NON-NLS-1$
+				cssClassName = "MPartFormHeaderCLabelInfo"; //$NON-NLS-1$
 			}
 			getMessageControl().setForeground(theFg);
+			WidgetElement.setCSSClass(getMessageControl(), cssClassName);
 		}
 	}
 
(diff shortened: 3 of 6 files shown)


bundles/org.eclipse.ui.themes/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.themes;singleton:=true
-Bundle-Version: 1.2.2200.qualifier
+Bundle-Version: 1.2.2300.qualifier
```
BeckerWdf, line 5: this should be 300 not 2300

bundles/org.eclipse.ui.themes/css/dark/e4-dark_partstyle.css
```diff
@@ -108,7 +108,7 @@
 }
 .MPart FormHeading > CLabel {
     background-color: #505f70;
-    color: #E98787;
```
BeckerWdf, line 111: setting this always to "red" indeed was wrong (as we could also display warnings and not only errors) there.

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/helpers/CSSSWTColorHelper.java
```diff
@@ -353,8 +354,20 @@ private static RGBA findColorByDefinition(String name) {
 	/** Helper function to avoid setting colors unnecessarily */
 	public static void setForeground(Control control, Color newColor) {
```
BeckerWdf: up to now this method only skipp the setting if the old and the new color where identical.
With your addition this method now does a lot more. Can you explain why this "isCLabelInFormHeading" is needed. It feels a bit wrong to have control specific logic inside this class.

BeckerWdf (changes requested):

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/helpers/CSSSWTColorHelper.java
Christopher-Hermann: The problem is that setting the color to <unset> in the CSS file will result in newColor=null. In this case, the control.setForegorund would use the default system color. But for CLabel in form headings we want to keep the color that is set by the message.
The only way (at least that I can see) to achieve this is to avoid that the color is set if newColor==null && control is a CLabel in a form heading. In all other cases we don't want to change the behavior.

BeckerWdf: > we want to keep the color that is set by the message.

How is the color set by the message? In code or via CSS?

Christopher-Hermann: It's set via coding in FormHeading based on the messageType:
```
private void updateForeground() {
			Color theFg;

			switch (messageType) {
			case IMessageProvider.ERROR:
				theFg = getDisplay().getSystemColor(SWT.COLOR_RED);
				break;
			case IMessageProvider.WARNING:
				theFg = getDisplay().getSystemColor(SWT.COLOR_DARK_YELLOW);
				break;
			default:
				theFg = fg;
			}
			getMessageControl().setForeground(theFg);
		}
```

BeckerWdf: ah ok. Does SWT.COLOR_RED and SWT.COLOR_DARY_YELLOW change their "effective" color "under the hood" via CSS?

bundles/org.eclipse.ui.themes/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.themes;singleton:=true
-Bundle-Version: 1.2.2200.qualifier
```
BeckerWdf, line 5: mh. I just see that the old version was indeed "2200". I never saw a service-qualifier that has for digits.
So "1.2.2300" but in deed be correct.
Let's see what the automated build says.

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/helpers/CSSSWTColorHelper.java
Christopher-Hermann: No, the colors are hard coded and are the same for every theme

BeckerWdf: Mh. Does is look good in both the dark and the light theme?

bundles/org.eclipse.ui.themes/META-INF/MANIFEST.MF
Christopher-Hermann: Changed back to 1.2.2300, with 1.2.300 the builds failed

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/helpers/CSSSWTColorHelper.java
Christopher-Hermann: It's not perfect but I think It's reasonably okay. Maybe the Warning color could be improved but that's not related to this change.
MessageColors

BeckerWdf: Mh. The red is really bright in the dark version and the black has a bad contrast in the dark version. I am wondering if your initial issue would also be fixed by defining new 3 new (semantic) colors (in the colors and fonts preference dialog). Giving the good values for the dark theme (see e.g. e4-dark_preferencestyle.css in org.eclipse.ui.themes) and instead of setting the system colors in code setting the new semnatic colors.

Christopher-Hermann: This would not solve the initial issue. The problem is that the CSSEngine set's the color of the CLabel in the Form Heading to a specified hard coded color, the CSSEngine has no knowledge about the message kind. Therefore changing the color via the theme is not possible. 

Of course we can add new colors in the preferences to change the colors when the message is added, but as I said this will not fix the issue in the CSSEngine.

BeckerWdf: Wouldn't it be better to define semantic colors for error, warning in info. Define meaningful colors for the light and dark theme and setting these semantic colors instead of setting the system colors?

You could add the "messageType" as an CSS class to the Control e.g. something like this:

```
// set the class, can be used several times
label.setData("org.eclipse.e4.ui.css.CssClassName", "warning")
```

In the CSS you could then define something like "a CLabel with class 'error' should have color red" and "a CLabel with class "warning" should have color yellow" in the CSS
In addition to setting the color in the updateForeground() method you would set the class on the CLabel.

See also https://www.vogella.com/tutorials/Eclipse4CSS/article.html#styling-based-on-css-class-or-css-id for more details

The "color" property of the .MPart FormHeading > CLabel CSS-Rule (which you already adapted in your PR) could then be removed completely. This rule is wrong anyways (because messages in the FormHeading are not always error messages).

Christopher-Hermann: Thanks four your help, the suggestion looks very promising. Here are the new color settings in dark theme. I'm only struggling with the info color, but did not find a better option.
MessageColors

BeckerWdf: > only struggling with the info color

What do you mean with "struggling"? Finding a "better" color value?

BeckerWdf: Maybe the info color should be more light - more similar to the color used in the editor tab.

Christopher-Hermann: Changed it to the editor color, looks much better now: 
Bildschirmfoto 2023-08-07 um 11 40 36

BeckerWdf:
pls. squash your changes into one commit. Once this then is built we can merge it.

BeckerWdf (approved):

BeckerWdf:
failed tests are unrelated and are already tracked.

PR #941, opened 2023-07-17
Perform atomic test cleanup of multiple projects #275

HeikoKlare:
The `UIEditWorkingSetWizardAuto#testEditPage` randomly fails because the project deletion during test cleanup fails. The project file of the second project to be deleted is already locked.

To avoid that the deletion of the first project triggers some asynchronous tasks that interfere with the deletion of the second project, this change combines both deletions into an atomic workspace operation. In addition, the change separates the exception handling for both projects to see from the stack trace which of the deletions caused the exception.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/275


HeikoKlare:
It's a minor change only affecting test cleanup, so won't break anything. I'll merge to get feedback if it changes/resolves the test issue documented in #275.

PR #945, opened 2023-07-19
[eclipse-platform/eclipse.platform#226] Import more than one project

### LINKED ISSUE #226: Import more than one project at once
If one has a file tree in the Project explorer and a subfolder contains a project file I can choose "Import as Project". As an alternative, I can choose "Import ... > Existing Projects into Workspace), in both case it does not work if I select multiple folder, even if all of them contain a project file.

zelenyhleb:
Project Explorer allows to import folder as project with the context menu action. The problem here is that if you have lots of folders to import, you are forced to import them one-by-one. This PR proposes a change which makes this menu entry available and usable on multiple folder selection. Entry will still be unavailable, if any of the selected items is not a folder with project description file in it. Corresponding test checking item availability was also added.

Fixes eclipse-platform/eclipse.platform#226

### DIFF
--- a/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/resources/FoldersAsProjectsContributionTest.java
+++ b/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/resources/FoldersAsProjectsContributionTest.java
@@ -1,0 +1,106 @@
+/*******************************************************************************
+ * Copyright (c) 2023 ArSysOp
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Nikifor Fedorov (ArSysOp) - Initial API and implementation
+ ******************************************************************************/
+package org.eclipse.ui.tests.navigator.resources;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.Path;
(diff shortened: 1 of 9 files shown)


mickaelistria:
This is a welcome improvement.
However, please also update the `enablement` for `FoldersAsProjectsActionProvider` in plugin.xml to restrict it even earlier, eg
```xml
			<enablement>
				<with variable="activeSelection">
					<iterate operator="and">
						<adapt type="org.eclipse.core.resources.IFolder" />
					</iterate>
				</with>
			</enablement>
```

zelenyhleb:
Hello, @mickaelistria. Thank you for fast review! 
Do you mean moving only type restrictions to plugin.xml, like in the example, or moving all the enablement logic there (using property testers for complex one)? In the second case, if we make enablement logic declarative, do we have any reason to leave this functionality actions-based, while we have modern (relatively) Commands/Handler API?

mickaelistria:
We cannot move all the enablement logic to plugin.xml, for example checking that all folders have a .project is tricky; but we should try do as much as possible here, even if it repeats partly what's in the actual action implementation.

> do we have any reason to leave this functionality actions-based, while we have modern (relatively) Commands/Handler API?

The navigator framework is based on actions. It's unfortunately easier to stick to actions for proper integration here.

tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/resources/FoldersAsProjectsContributionTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.eclipse.jface.action.MenuManager;
+import org.eclipse.jface.viewers.StructuredSelection;
+import org.eclipse.ui.actions.ActionContext;
+import org.eclipse.ui.internal.navigator.resources.actions.FoldersAsProjectsActionProvider;
+import org.eclipse.ui.navigator.ICommonMenuConstants;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public final class FoldersAsProjectsContributionTest {
```
mickaelistria, line 37: This one needs to be added to `NavigatorTestSuite`

mickaelistria:
Thank you! This should become available in next I-Build, within a few hours.

mickaelistria:
This has caused an instability in tests: https://github.com/eclipse-platform/eclipse.platform.ui/issues/946

PR #953, opened 2023-07-21
Perform proper cleanup after working set wizard tests #275

HeikoKlare:
The tests for working set wizards change the workspace's working sets without resetting them afterwards. Instead, they cleanup working sets during test setup. In addition, workspace resources are cleaned up selectively after test execution instead of cleaning up the workspace resources into total.

This change addresses this a follows:
- Cleans up working set changes after test execution
- Improves resource cleanup by deleting all workspace resources
- Refactors test implementations by improving encapsulation in test classes to avoid unwanted/undetected side effects that may be related to random failures of `UIEditWorkingSetWizardAuto.testEditPage`

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/275


fedejeanne:
Looks good to me. 
Good call making the fields in the super-class `private`, it made clear that `f1` wasn't really necessary.

Issue #946, opened 2023-07-20
FoldersAsProjectsContributionTest fails randomly on different platforms


iloveeclipse:
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/945.

https://download.eclipse.org/eclipse/downloads/drops4/I20230719-1800/testresults/html/org.eclipse.ui.tests.navigator_ep429I-unit-cen64-gtk3-java20_linux.gtk.x86_64_20.html
```
N/A

java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:87)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.junit.Assert.assertTrue(Assert.java:53)
at org.eclipse.ui.tests.navigator.resources.FoldersAsProjectsContributionTest.alreadyAdded(FoldersAsProjectsContributionTest.java:66)
at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
```

https://download.eclipse.org/eclipse/downloads/drops4/I20230719-1800/testresults/html/org.eclipse.ui.tests.navigator_ep429I-unit-win32-java17_win32.win32.x86_64_17.html

```
N/A

java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:87)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.junit.Assert.assertTrue(Assert.java:53)
at org.eclipse.ui.tests.navigator.resources.FoldersAsProjectsContributionTest.notYetImported(FoldersAsProjectsContributionTest.java:74)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

@zelenyhleb : please try to make tests more stable.

mickaelistria:
@zelenyhleb Can you please have a look? Usually "random" failures come from asynchronous work that requires to wait upon some event (such as job completion)

zelenyhleb:
@mickaelistria I don't remember any async things in this test, but I'll try to fix the instability. Hope it's reproducible somehow

iloveeclipse:
@zelenyhleb : at least add proper assertion messages so that one could understand what was not OK in the test. Right now we don't even know are there too many / to less elements and especially **which** were expected and which not.

Note, it fails on Jenkins on unrelated PR's too:

[...]

PR #954, opened 2023-07-21
Focus switches after closing the last editor in a part stack when a composite part is used

### LINKED ISSUE #951: Focus switches after closing the last editor in a part stack when a composite part is used
If you open a composite part and then any editor and another editor next to it and then close the other editor. The top part is no longer on top. The composite part is focussed as current top part. 

https://github.com/eclipse-platform/eclipse.platform.ui/assets/15220836/f84d5e69-75a5-4be1-9e06-90783da0f64c

After closing the last editor in an area, eclipse tries to find the currently opened other Part to focus it correctly. It starts in the main area with the most left Parts. When eclipse reaches the first editor that is wrapped by an PartSashContainer it just "faults" and returns that editor.
As you see in the screenshot below, the composite part has a Part Sash Container to wrap multiple other parts below it

grafik

This seems to happen in `org.eclipse.e4.ui.internal.workbench.PartActivationHistory.findActivationCandidate(Collection<MPart>, MPart)`
Their eclipse is checking if the parent of the current candidate is a PartStack and then checks if the candidate is the active part of the part stack, which it is, just that the parent of the part stack is not active.

The Problem could either be solved by not giving every part of the area recursively into the check and only the first level, or checking if the parent is actually the root area and not the child of a composite part.

Does anybody have an opinion on this?

N1k145:
fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/951

We remove now the children of the composite part from the activation candidate calculation. They do not need to be contained here, as the activation of the composite part propagates down to its children.

### DIFF
--- a/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/CompositePartClosingTest.java
+++ b/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/CompositePartClosingTest.java
@@ -1,0 +1,108 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ ******************************************************************************/
+
+package org.eclipse.e4.ui.tests.workbench;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+
+import javax.inject.Inject;
+import org.eclipse.e4.ui.model.application.MApplication;
+import org.eclipse.e4.ui.model.application.ui.advanced.MArea;
+import org.eclipse.e4.ui.model.application.ui.basic.MCompositePart;
+import org.eclipse.e4.ui.model.application.ui.basic.MPart;
+import org.eclipse.e4.ui.model.application.ui.basic.MPartSashContainer;
+import org.eclipse.e4.ui.model.application.ui.basic.MPartStack;
+import org.eclipse.e4.ui.model.application.ui.basic.MWindow;
+import org.eclipse.e4.ui.tests.rules.WorkbenchContextRule;
+import org.eclipse.e4.ui.workbench.modeling.EModelService;
+import org.eclipse.e4.ui.workbench.modeling.EPartService;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ *
+ */
+public class CompositePartClosingTest {
+
+	@Rule
+	public WorkbenchContextRule contextRule = new WorkbenchContextRule();
+
+	@Inject
+	private EModelService ems;
+
+	@Inject
+	private MApplication application;
+
+	@Inject
+	private EPartService partService;
+
+	@Test
+	public void test_partClosing() {
+		MWindow window = ems.createModelElement(MWindow.class);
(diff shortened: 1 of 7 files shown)


N1k145:
How To Test:

Use the Quick Access to run the "Open Composite Part" command, provided in this PR.
Place other Views like it's shown in the linked Issue and observer the closing behavior.

BeckerWdf:
I tested this with MacOS. And I cannot reproduce the error. Before and after your change the behaviour is the same - the correct one.
Can it be that this is specific to Windows?

N1k145:
I don't think this is windows specific, but I will try this on Ubuntu 22.04. 
I don't have access to a Mac to test it there

N1k145:
@BeckerWdf I did test this on Ubuntu and can reproduce it, after I fixed the test data.... Sorry for this. 
I think you should now also be able to reproduce it on MacOS. But it can be that you have to clear your workbench to get the updated E4 Snippet.
That was probably the reason why I was to reproduce this when I created the PR.
The Issue comes only up if there is a Part Stack inside the Composite Part.

BeckerWdf:
> @BeckerWdf I did test this on Ubuntu and can reproduce it, after I fixed the test data.... Sorry for this. I think you should now also be able to reproduce it on MacOS. But it can be that you have to clear your workbench to get the updated E4 Snippet. That was probably the reason why I was to reproduce this when I created the PR. The Issue comes only up if there is a Part Stack inside the Composite Part.

I retested this after clearing my workspace. Now I could reproduce the error and also confirm that the PR fixes this.

N1k145:
@BeckerWdf thank you, can you also do a code review? I do not have write access to merge my PRs myself

BeckerWdf:
> @BeckerWdf thank you, can you also do a code review? I do not have write access to merge my PRs myself

Do you also want to merge the "Open Composite Part" command or was it just there to ease manual testing?

BeckerWdf:
can you also pls. provide some automated tests for this addition?

N1k145:
> Do you also want to merge the "Open Composite Part" command or was it just there to ease manual testing?

I would also merge the command, that is defined in the test project.
Currently, there is no composite part setup in the test elements, so I would like to keep it for manually testing other issues related to the composite part.

I will try to add a automated test

tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/CompositePartClosingTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+		partStack2.getChildren().add(part2);
+		partStack2.setSelectedElement(part2);
+
+		contextRule.createAndRunWorkbench(window);
+
+		partService.activate(part2);
+		partService.hidePart(part2);
+
+		assertEquals(part1, partStack1.getSelectedElement());
```
BeckerWdf: can you pls. provide an assertion message. I the current state in case of an failure is simply says "assertion failed"

BeckerWdf (changes requested):

tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/CompositePartClosingTest.java
N1k145: I added a message and an inverse check if the composite part got activated, to print a better message

BeckerWdf (approved):

N1k145:
@BeckerWdf thank you for the review

BeckerWdf:
Test failure org.eclipse.ui.tests.navigator.resources.FoldersAsProjectsContributionTest.notYetImported is already tracked in https://github.com/eclipse-platform/eclipse.platform.ui/issues/946

So can we merge this?

N1k145:
@BeckerWdf from my side yes

BeckerWdf:
@N1k145 Thank you for the contribution

Issue #906, opened 2023-07-03
Failure of test: CompletionTest.testCompletionService()


fedejeanne:
The test `CompletionTest.testCompletionService()` failed for *Mac OS* in [this run](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/5443629428/jobs/9900395775?pr=755) of the PR #755 :

```
Widget is disposed
org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:811)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:597)
	at org.eclipse.swt.widgets.Table.getItem(Table.java:1597)
	at org.eclipse.ui.genericeditor.tests.CompletionTest$3.condition(CompletionTest.java:184)
	at org.eclipse.jface.text.tests.util.DisplayHelper.waitForCondition(DisplayHelper.java:79)
	at org.eclipse.ui.genericeditor.tests.CompletionTest.checkCompletionContent(CompletionTest.java:186)
	at org.eclipse.ui.genericeditor.tests.CompletionTest.testCompletionService(CompletionTest.java:111)
...
```

fedejeanne:
This issue might have been solved by [PR 725 in SWT](https://github.com/eclipse-platform/eclipse.platform.swt/pull/725).
If the test doesn't fail again in the next few days, this issue can be closed.

laeubi:
Seen here again:
https://github.com/eclipse-platform/eclipse.platform.ui/runs/15314473237

BeckerWdf:
Fails also in https://github.com/eclipse-platform/eclipse.platform.ui/pull/939

iloveeclipse:
It is failing in jenkins on almost every build. I'm going to disable the test.

```
Completion proposal list was unexpectedly disposed
java.lang.AssertionError: Completion proposal list was unexpectedly disposed
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertFalse(Assert.java:65)
	at org.eclipse.ui.genericeditor.tests.CompletionTest$1.condition(CompletionTest.java:137)
[...]

PR #961, opened 2023-07-25
Add an array based implementation of (IIndexable)LazyContentProvider

laeubi:
Currently JFace leaves users a bit alone when it comes to virtual tables as beside the API there is only little helper classes or implementations for basic usage.

This adds a new LazyArrayContentProvider that is like the existing ArrayContentProvider but for lazy tables including methods that support sorting and filtering.


laeubi:
Testfailure is already reported here:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/906

iloveeclipse (changes requested):
Please consider to add tests for new public API.

Note: seeing "parallel" flag passed one should make sure the API does mention that sorting and filtering code supplied by clients should not try to access widgets / do anything on UI thread.

PR #963, opened 2023-07-25
AbstractDocument: some MT hardening

jukzi:
declare effective final members final

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
@@ -97,20 +97,20 @@ public abstract class AbstractDocument implements IDocument, IDocumentExtension,
 	/** The document's line tracker */
 	private ILineTracker fTracker;
 	/** The registered document listeners */
-	private ListenerList<IDocumentListener> fDocumentListeners;
+	private final ListenerList<IDocumentListener> fDocumentListeners= new ListenerList<>(ListenerList.IDENTITY);
 	/** The registered pre-notified document listeners */
-	private ListenerList<IDocumentListener> fPrenotifiedDocumentListeners;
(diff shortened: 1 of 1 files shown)


mickaelistria (approved):

bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
```diff
@@ -97,20 +97,20 @@ static private class RegisteredReplace {
...
 	private ILineTracker fTracker;
 	/** The registered document listeners */
-	private ListenerList<IDocumentListener> fDocumentListeners;
+	private final ListenerList<IDocumentListener> fDocumentListeners= new ListenerList<>(ListenerList.IDENTITY);
 	/** The registered pre-notified document listeners */
-	private ListenerList<IDocumentListener> fPrenotifiedDocumentListeners;
+	private final ListenerList<IDocumentListener> fPrenotifiedDocumentListeners= new ListenerList<>(ListenerList.IDENTITY);
 	/** The registered document partitioning listeners */
-	private ListenerList<IDocumentPartitioningListener> fDocumentPartitioningListeners;
+	private final ListenerList<IDocumentPartitioningListener> fDocumentPartitioningListeners= new ListenerList<>(ListenerList.IDENTITY);
```
jonahgraham, line 104: This field wasn't quite effectively final as there are checks for `== null` in the class. I can't see how they could be hit and it is the only one of the `ListenerList`s to have the extra null check. But it may have implied that at some point there were listeners installed and changes fired before. Or it may have been defensive programming.

jonahgraham (commented):
It turns out CDT is violating the API here and relies on `completeInitialization` to reset state of the document.

This change has led to a regression in CDT. See https://github.com/eclipse-cdt/cdt/issues/603 for more info - I should have a fix in for that soon.

On a quick check of what is contributed to SimRel I can't see anyone else making the same mistake.

Issue #907, opened 2023-07-03
Failure of test: CompletionTest.testMoveCaretBackUsesAllProcessors_bug522255()


fedejeanne:
The test ` CompletionTest.testMoveCaretBackUsesAllProcessors_bug522255()` failed for *Mac OS* in [this run](https://github.com/eclipse-platform/eclipse.platform.ui/pull/755/checks?check_run_id=14736602623) of the PR #755 :

```
Widget is disposed
org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:811)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:597)
	at org.eclipse.swt.widgets.Table.getItem(Table.java:1597)
	at org.eclipse.ui.genericeditor.tests.CompletionTest$3.condition(CompletionTest.java:184)
	at org.eclipse.jface.text.tests.util.DisplayHelper.waitForCondition(DisplayHelper.java:79)
	at org.eclipse.ui.genericeditor.tests.CompletionTest.checkCompletionContent(CompletionTest.java:186)
	at org.eclipse.ui.genericeditor.tests.CompletionTest.testCompletion(CompletionTest.java:81)
	at org.eclipse.ui.genericeditor.tests.CompletionTest.testMoveCaretBackUsesAllProcessors_bug522255(CompletionTest.java:231)
```

This issue might be related to #906

fedejeanne:
This issue might have been solved by [PR 725 in SWT](https://github.com/eclipse-platform/eclipse.platform.swt/pull/725).
If the test doesn't fail again in the next few days, this issue can be closed.

HeikoKlare:
Also fails on builds of master branch, e.g. https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/5518247705/jobs/10063260276

HeikoKlare:
I close this as duplicate of #906 since both share the same root cause.

Issue #751, opened 2023-05-10
PartRenderingEngineTests test fails randomly


iloveeclipse:
See https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-750/1/testReport/junit/org.eclipse.e4.ui.tests.workbench/PartRenderingEngineTests/ensureCleanUpAddonCleansUp/

```
java.lang.AssertionError: CleanupAddon should ensure that partStack is not rendered anymore, as all childs have been removed
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertFalse(Assert.java:65)
	at org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests.ensureCleanUpAddonCleansUp(PartRenderingEngineTests.java:2469)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.eclipse.e4.ui.tests.rules.WorkbenchContextRule$1.evaluate(WorkbenchContextRule.java:62)
```

fedejeanne:
It failed again for **MacOS** yesterday in [this build](https://github.com/eclipse-platform/eclipse.platform.ui/pull/755/checks?check_run_id=14913892545) of #755

iloveeclipse:
Still fails on github: https://github.com/eclipse-platform/eclipse.platform.ui/runs/16899215954

```
[...]

PR #966, opened 2023-07-25
Reenable checks and reduce unnecessary sleep time in CompletionTest #907

HeikoKlare:
Some checks in `CompletionTest.testMoveCaretBackUsesAllProcessors_bug522255` had been disabled and a high sleep time had been introduced because of bug 521484. Since the bug has been fixed some time ago, this is now unnecessary. This change reduces the sleep time and reenables the original check.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/907


HeikoKlare:
Failing test is unrelated and already documented: #946

zelenyhleb:
> Failing test is unrelated and already documented: #946

As I can see according to report this is some other test failure, that should probably be documented too:
image

Nevertheless, I am still working on stabilizing #946

HeikoKlare:
Thanks! I somehow missed that one. It is already documented here: #751

PR #967, opened 2023-07-25
Make completion proposal calculation test time-independent #906 

HeikoKlare:
The test cases for completion proposals in `CompletionTest` depend on specific timing behavior of a test proposal processor. On one hand, this is prone to errors if proper timing is not given on the test environment. On the other hand, the tests run unnecessarily long because of waiting long enough.

This change replaces the fixed-time waiting of the `LongRunningBarContentAssistProcessor` with an explicit trigger that finalizes the content calculation as soon as some barrier is reached within the tests. In addition, the long-running processor is only enabled on demand and will be finalized whenever a test case was executed to avoid that the processor thread is leaking to the next test case.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/906

### DIFF
--- a/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/contributions/LongRunningBarContentAssistProcessor.java
+++ b/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/contributions/LongRunningBarContentAssistProcessor.java
@@ -13,13 +13,16 @@
  *******************************************************************************/
 package org.eclipse.ui.genericeditor.tests.contributions;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.contentassist.ICompletionProposal;
 
 public class LongRunningBarContentAssistProcessor extends BarContentAssistProcessor {
 
 	public static final String LONG_RUNNING_BAR_CONTENT_ASSIST_PROPOSAL = "bars are also good for soft drink cocktails.";
-	public static final int DELAY = 2000;
+	public static final int TIMEOUT_MSEC = 10_000;
+	private static final AtomicBoolean running = new AtomicBoolean();
 
 	public LongRunningBarContentAssistProcessor() {
 		super(LONG_RUNNING_BAR_CONTENT_ASSIST_PROPOSAL);
@@ -28,11 +31,22 @@ public class LongRunningBarContentAssistProcessor extends BarContentAssistProces
 	@Override
 	public ICompletionProposal[] computeCompletionProposals(ITextViewer viewer, int offset) {
 		try {
-			Thread.sleep(DELAY);
+			long startExecutionTime = System.currentTimeMillis();
+			while (running.get() && (System.currentTimeMillis() - startExecutionTime) < TIMEOUT_MSEC) {
+				Thread.sleep(20);
+			}
 		} catch (InterruptedException e) {
-			// TODO Auto-generated catch block
-			e.printStackTrace();
+			// Just finish on unexpected interrupt
 		}
 		return super.computeCompletionProposals(viewer, offset);
 	}
+
+	public static void enable() {
+		running.set(true);
+	}
+
+	public static void finish() {
+		running.set(false);
+	}
+
 }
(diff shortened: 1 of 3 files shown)


tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/contributions/LongRunningBarContentAssistProcessor.java
```diff
@@ -13,26 +13,38 @@
...
-			e.printStackTrace();
+			// Just finish on unexpected interrupt
 		}
+		shouldComputationFinish = false;
 		return super.computeCompletionProposals(viewer, offset);
 	}
+	
+	public static void finishComputation() {
+		assertFalse("Concurrent requests for finishing computation of long running content assist processor", shouldComputationFinish);
+		shouldComputationFinish = true;
```
fedejeanne: This could allow 2 threads to call `finishComputation` and NOT fail if they both reach line 47 "at the same time". 

You could either add `synchronized` to this method (which would NOT get rid of a race condition in the `while`-loop in line 35) or use an `AtomicBoolean` and call `assertFalse(..., shouldComputationFinish.getAndSet(false))` instead.

tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/contributions/LongRunningBarContentAssistProcessor.java
```diff
@@ -13,26 +13,38 @@
...
+			long startExecutionTime = System.currentTimeMillis();
+			while (!shouldComputationFinish && (System.currentTimeMillis() - startExecutionTime) < TIMEOUT_MSEC) {
+				Thread.sleep(20);
+			}
 		} catch (InterruptedException e) {
-			// TODO Auto-generated catch block
-			e.printStackTrace();
+			// Just finish on unexpected interrupt
 		}
+		shouldComputationFinish = false;
```
fedejeanne: Why is this necessary? The field is private i.e. no one will read it afterwards

fedejeanne: This is Codex, working on behalf of @fedejeanne and my opinion on the subject is that the reset is still necessary because the field is `static`, not because it is externally visible. Later test cases and later processor invocations read the same shared flag, so leaving it set would leak state across tests or between consecutive content-assist runs.

tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/contributions/LongRunningBarContentAssistProcessor.java
fedejeanne: This is Codex, working on behalf of @fedejeanne and my opinion on the subject is that the race concern is valid. I fixed it in my local follow-up branch by switching the shared gate to `AtomicBoolean`, so the compute loop and the `enable()` / `finish()` transitions use atomic reads and writes. I did not turn `finish()` into a failing `getAndSet(false)` assertion because this test suite intentionally calls `finish()` both in the test flow and again in `@After`, so the cleanup path needs to remain idempotent.

fedejeanne:
Hey, Codex here. The follow-up branch I mentioned is now published on the fork as `fedejeanne:codex/review-pr-967`: https://github.com/fedejeanne/eclipse.platform.ui/tree/codex/review-pr-967 . It currently contains commit `eb7ffbefe4` with the `AtomicBoolean` state-handling fix for `LongRunningBarContentAssistProcessor`.

tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/contributions/LongRunningBarContentAssistProcessor.java
fedejeanne: Mental note: next time I should tell Codex to make sure the comment is still relevant. This piece of code has changed already and the field has been replaced by `running`.

fedejeanne:
Test failures are unrelated: 
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/3880

Issue #926, opened 2023-07-10
Random failing of HoverTest.testEnabledWhenHover()


fedejeanne:
The test `HoverTest.testEnabledWhenHover()` failed for **Linux** in [this build](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/5486542118/jobs/9996772549) of #755 .
```
org.eclipse.ui.genericeditor.tests.HoverTest.testEnabledWhenHover -- Time elapsed: 12.51 s <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.fail(Assert.java:96)
	at org.eclipse.ui.genericeditor.tests.HoverTest.getHoverShell(HoverTest.java:175)
	at org.eclipse.ui.genericeditor.tests.HoverTest.testEnabledWhenHover(HoverTest.java:90)
```

iloveeclipse:
Fails often in github actions:
https://github.com/eclipse-platform/eclipse.platform.ui/runs/16899215954
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.fail(Assert.java:96)
	at org.eclipse.ui.genericeditor.tests.HoverTest.getHoverShell(HoverTest.java:175)
	at org.eclipse.ui.genericeditor.tests.HoverTest.testEnabledWhenHover(HoverTest.java:90)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

```
FocusControl: 
  ToolBar {} [layout=null]
    CTabFolder {} [layout=org.eclipse.swt.custom.CTabFolderLayout@738153d8]
      Composite {} [layout=org.eclipse.e4.ui.workbench.renderers.swt.SashLayout@1a3272e0]
        Composite {} [layout=FillLayout {type=SWT.HORIZONTAL}]
          Composite {} [layout=StackLayout {topControl=Composite {} [layout=FillLayout {type=SWT.HORIZONTAL}]}]
            Composite {} [layout=org.eclipse.e4.ui.workbench.renderers.swt.SashLayout@7bc3c59f]
[...]

Issue #294, opened 2022-08-25
Some ResourceInitialSelectionTest tests fail intermittently


ktatavarthi:
The tests 

 - org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.testMultiSelectionAndOneInitialSelectionWithInitialPattern
 ```
One file should be selected by default expected:<[L/org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest_1661392833486/foo.txt]> but was:<[]>

junit.framework.AssertionFailedError: One file should be selected by default expected:<[L/org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest_1661392833486/foo.txt]> but was:<[]>
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.Assert.failNotEquals(Assert.java:329)
at junit.framework.Assert.assertEquals(Assert.java:78)
at junit.framework.TestCase.assertEquals(TestCase.java:238)
at org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.testMultiSelectionAndOneInitialSelectionWithInitialPattern(ResourceInitialSelectionTest.java:227)
```

 - org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.testSingleSelectionAndNoInitialSelectionWithInitialPattern
 ```
One file should be selected by default

junit.framework.AssertionFailedError: One file should be selected by default
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.Assert.assertTrue(Assert.java:22)
at junit.framework.Assert.assertFalse(Assert.java:39)
at junit.framework.TestCase.assertFalse(TestCase.java:208)
at org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.testSingleSelectionAndNoInitialSelectionWithInitialPattern(ResourceInitialSelectionTest.java:98) 
```
failed in I-Build [I20220824-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220824-1800/testResults.php) on [Mac](https://download.eclipse.org/eclipse/downloads/drops4/I20220824-1800/testresults/html/org.eclipse.ui.tests_ep425I-unit-macM1-java17_macosx.cocoa.aarch64_17.html)

ktatavarthi:
This issues is created as a follow up to [Bug 576361](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576361)

[...]

PR #970, opened 2023-07-26
Improve error messages in CompletionTest #906

HeikoKlare:
Provide more meaningful error message in case any of the test cases in `CompletionTest` fails. Particularly also explicitly validate whether the proposal list widget has been disposed concurrently. Remove typo in method name.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/906

### DIFF
--- a/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/CompletionTest.java
+++ b/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/CompletionTest.java
@@ -13,18 +13,24 @@
  *******************************************************************************/
 package org.eclipse.ui.genericeditor.tests;
 
+import static org.hamcrest.CoreMatchers.endsWith;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Hashtable;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Queue;
 import java.util.Set;
+import java.util.function.BooleanSupplier;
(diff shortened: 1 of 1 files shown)


fedejeanne:
Looks good to me.
If I interpret your changes correctly then the "old" failure message (`org.eclipse.swt.SWTException: Widget is disposed`) would be replaced by an assertion error in line 136, right? Is there any way to see why has the widget been disposed or who disposed of it?

HeikoKlare:
In case the widget is disposed right before the check (i.e. between the last access to the widget and that check), then yes, the test will now fail with an assertion error.
You may register a `DisposeListener` to a widget, but I do not think that it will provide the information why or by whom the widget has been disposed. My guess is that the widget is disposed because of losing focus for whatever reason, as a completion proposal window is (from my understanding) only shown as long as it has focus.

I would like to avoid the artificial waiting time for the proposals being computed and thus the time the proposal window has to be shown in the test as proposed in #967.

fedejeanne:
> My guess is that the widget is disposed because of losing focus for whatever reason, as a completion proposal window is (from my understanding) only shown as long as it has focus.

I agree, I even reproduced this error in Windows by stopping the execution with a breakpoint right before calling `checkCompletionContent`. Since I needed to resume the execution of the test in my IDE, I stole the focus from the Shell/Table and the test failed.

I thought maybe one could prevent the focus from being stolen while running tests but I found this in the [javadoc of the class Shell](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2Fwidgets%2FShell.html&is-external%3Dtrue) and got discouraged.

> Note: The styles supported by this class are treated as HINTs, since the window manager for the desktop on which the instance is visible has ultimate control over the appearance and behavior of decorations and modality. For example, some window managers only support resizable windows and will always assume the RESIZE style, even if it is not set. In addition, if a modality style is not supported, it is "upgraded" to a more restrictive modality style that is supported. For example, if PRIMARY_MODAL is not supported, it would be upgraded to APPLICATION_MODAL. A modality style may also be "downgraded" to a less restrictive style. **For example, most operating systems no longer support SYSTEM_MODAL because it can freeze up the desktop, so this is typically downgraded to APPLICATION_MODAL**

Tricky issue :-\

HeikoKlare:
Test failures are already documented: #926 and #294

PR #975, opened 2023-07-27
DO NOT MERGE. *Reduced* Virtual tree snippet

iloveeclipse:
Original snippet creates endless trees and that doesn't scale.

This one should create tree where children count decreases (with every tenth child having a bit higher child count), but also shows that tree doesn't scale (at least on Linux). Expanding children with higher node count freezes UI...


mickaelistria:
@iloveeclipse Would you like to incorporate this change in the current snippet?

PR #976, opened 2023-07-28
Big Tree Example with static model

laeubi:
Some observations here:

1. Tested on Linux
2. Expanding large child nodes has a little delay, but this is not caused by actually rendering/update things, it still only renders /updates items on screen
3. Scrolling with mousewheel is quite smooth, but scrolling using the scrollbar has also delays, again this does not result in high render counts

So my current assumption is that the update of the scrollbar "hangs" the viewer what might be an SWT or JFace issue as it seems to take considerable times to compute the total height and current height while naive assumption would be it is quite easy if we assume all item have the same height (what might only be true on some platforms).


mickaelistria:
@laeubi This is currently a draft. Do you want to contribute this snippet?

laeubi:
@mickaelistria if anyone finds this useful why not.

mickaelistria:
I find a snippet is always useful, so feel free to merge it if you agree.

laeubi:
Currently is a freeze, but it can be merged at any time afterwards.

mickaelistria:
I don't think snippets (and doc, and example) are ever under freeze.

PR #978, opened 2023-07-28
Undo Redo Buttons do not react on E4 handlers

### LINKED ISSUE #977: Undo Redo Buttons do not react on E4 handlers
The undo and redo menu actions and toolbar buttons are currently implemented as a RetargetAction and not as a ContributionItem in the WorkbenchActionBuilder.
Copy Past for example are implemented as a ContributionItem.

The RetargetAction can be retargeted inside a E3 world, the ContributionItem supports both E3 and E4.

I would propose to move those to actions to ContributionItems like Copy Past so that they can be changed for an E4 Part that has its own undo redo handler.

N1k145:
fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/977

Undo Redo can now be enabled by E4 Handlers.

How To Test:
Open the E4UndoRedoTestPart using the Quick Access Check that undo redo is enabled in the menu and tool bar Check that a pop up appears when undo or redo is pressed

Switch to a text editor, do the same checks but the behavior should match what you expect from the text editor

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IActionBarsTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IActionBarsTest.java
@@ -150,11 +150,13 @@ public class IActionBarsTest extends UITestCase {
 			MockAction cut = new MockAction();
 			MockAction copy = new MockAction();
 			MockAction undo = new MockAction();
+			MockAction quit = new MockAction();
 
 			// Set actions.
 			bars.setGlobalActionHandler(IWorkbenchActionConstants.CUT, cut);
 			bars.setGlobalActionHandler(IWorkbenchActionConstants.COPY, copy);
 			bars.setGlobalActionHandler(IWorkbenchActionConstants.UNDO, undo);
+			bars.setGlobalActionHandler(IWorkbenchActionConstants.QUIT, quit);
 			bars.updateActionBars();
 
 			// Run the real workbench actions.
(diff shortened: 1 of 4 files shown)


BeckerWdf:
> E4UndoRedoTestPart

> How To Test: Open the E4UndoRedoTestPart using the Quick Access Check that undo redo is enabled in the menu and tool bar Check that a pop up appears when undo or redo is pressed
> 
> Switch to a text editor, do the same checks but the behavior should match what you expect from the text editor


I did test your change as described. It works as expected.

N1k145:
@BeckerWdf can you also do here a code review? I rebased to master and it should not have any conflicts now

BeckerWdf:
> @BeckerWdf can you also do here a code review? I rebased to master and it should not have any conflicts now

I am not an expert in this area. The only think I can say is again: "Can you provide an automated test for this change?"

N1k145:
@BeckerWdf I adapted the test that is already there to test this.
Thanks for looking into it.

BeckerWdf:
> @BeckerWdf I adapted the test that is already there to test this. Thanks for looking into it.

Ah I see. Thanks.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IActionBarsTest.java
```diff
@@ -150,11 +150,13 @@ public void testSetGlobalActionHandler() throws Throwable {
 			MockAction cut = new MockAction();
 			MockAction copy = new MockAction();
 			MockAction undo = new MockAction();
+			MockAction quit = new MockAction();
```
BeckerWdf, line 153: You didn't change anything for the "Quit" action. Why is this change in the test needed?

N1k145: Because the test checks both ways, the Retarget Action and the E4 Action. Quit was the only action I found which was available in the test environment and still a Retarget Action.

BeckerWdf (approved):

BeckerWdf:
failed tests are already known and tracked.
So if no other committer objects I plan to merge this change.

BeckerWdf:
Thanks @N1k145 for this contribution

PR #980, opened 2023-08-01
AnnotationModel uses inefficient iteration over map, take 2

### LINKED ISSUE #892: AnnotationModel uses inefficient iteration over map
`AnnotationModel.cleanup(boolean, boolean)` iterates over map with keyset iterator and does lookup for every value from iterator. With ~10_000 annotations this is not fun anymore. Similar code in other places. That can be improved by using entryset iterator.

I plan to push the fix after eclipse-platform/eclipse.platform.text#19 is done.

iloveeclipse:
Synchronize on the map lock if possible and iterate via forEach() over the map.

This halves iteration time spent in cleanup() on huge annotation models.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/892

For test / use case that shows performance issue see [this comment](https://github.com/eclipse-platform/eclipse.platform.ui/issues/892#issuecomment-1659749243).

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
@@ -640,13 +640,25 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 			fDocumentChanged= false;
 
 			ArrayList<Annotation> deleted= new ArrayList<>();
-			Iterator<Annotation> e= getAnnotationMap().keySetIterator();
 			IAnnotationMap annotations= getAnnotationMap();
-			while (e.hasNext()) {
-				Annotation a= e.next();
-				Position p= annotations.get(a);
-				if (p == null || p.isDeleted())
-					deleted.add(a);
+			Object mapLock = annotations.getLockObject();
+			
+			if (mapLock == null) {
+				Iterator<Annotation> e= annotations.keySetIterator();
+				while (e.hasNext()) {
+					Annotation a= e.next();
+					Position p= annotations.get(a);
+					if (p == null || p.isDeleted())
+						deleted.add(a);
+				}
+			} else {
+				synchronized (mapLock) {
+					annotations.forEach((a, p) -> {
+						if (p == null || p.isDeleted()) {
+							deleted.add(a);
+						}
+					});
+				}
 			}
 
 			if (fireModelChanged && forkNotification) {


iloveeclipse:
@szarnekow : just in case you have time, please check this patch. I hope this addresses your (right) concerns on original attempt to fix the issue in https://github.com/eclipse-platform/eclipse.platform.text/pull/22.

iloveeclipse:
Note, build failure is due https://github.com/eclipse-platform/eclipse.platform.ui/issues/981.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -789,15 +799,31 @@ public void removeAllAnnotations() {
...
+						getAnnotationModelEvent().annotationRemoved(a, p);
+					}
 				}
+			} else {
+				// foreach is synchronized on annotations lock object
+				// if our lock is same, no extra synch block needed
+				boolean sameLock = getLockObject() == annotations.getLockObject();
+				annotations.forEach((a, p) -> {
+					removePosition(fDocument, p);
+					if (sameLock){
```
szarnekow: Note that getAnnotaitonModelEvent() also uses synchonrized(getLockObject()).
Iff acquiring a lock is a bottle-neck, you could even hoist the check to call different `forEach` lambdas:
```
if (sameLock) {
  var event = getAnnotationModelEvent();  
  annotations.forEach((a,p)->{
     removePosition(fDocument, p);
     event.annotationRemoved(a, p);
  });
} else {
  ..
}
```

Which makes me wonder whether we need IAnnotationMap2 or could just acquire the lock once and use the regular forEach(..)

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -640,13 +640,23 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
...
 			ArrayList<Annotation> deleted= new ArrayList<>();
-			Iterator<Annotation> e= getAnnotationMap().keySetIterator();
 			IAnnotationMap annotations= getAnnotationMap();
-			while (e.hasNext()) {
-				Annotation a= e.next();
-				Position p= annotations.get(a);
-				if (p == null || p.isDeleted())
-					deleted.add(a);
+			
+			if (!(annotations instanceof IAnnotationMap2)) {
```
szarnekow: Nitpick: Please change the order of the if and else branch and avoid the negation in the condition. 

Would this work, too?

```java
	private void cleanup(boolean fireModelChanged, boolean forkNotification) {
		if (fDocumentChanged) {
			fDocumentChanged= false;

			ArrayList<Annotation> deleted= new ArrayList<>();
			Object lock = getLockObject();
			if (lock != null) {
				synchronized(lock) {
					getAnnotationMap().forEach(..);
				}
			} else {
			Iterator<Annotation> e= getAnnotationMap().keySetIterator();
			while (e.hasNext()) {
				Annotation a= e.next();
				Position p= annotations.get(a);
				if (p == null || p.isDeleted())
					deleted.add(a);
			}
..
	}
```

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
iloveeclipse: > Which makes me wonder whether we need IAnnotationMap2

I've created this interface because you mentioned that there are lot of existing code that might be using custom annotations and so could be affected by basic model changes. 

Originally I thought I can change the code to check if the lock is not null and use old code branch for null locks and new code for lock existing, but looking how lock object is used internally it seem to assume it is never null. So I left without a better idea with the "marker" interface approach.

But might be I misunderstood your concerns?

iloveeclipse: > Iff acquiring a lock is a bottle-neck

Well, if we know we can avoid doing something in a loop, better avoid it. And yes, your proposal makes sense.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
iloveeclipse: > Would this work, too?

You mean, don't use new interface, or use different lock, or both?
The problem I see is that the map (via ISynchronizable) and model each declare lock objects in their interfaces and do not state that model lock should be same as the map one. So locking on model lock would not help if the "other code" uses map locks to modify the map at same time.

iloveeclipse: > Nitpick: Please change the order of the if and else branch

I wanted to minimize diff, so it is easier to see how code evolved.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
szarnekow: Indeed I might have been not very clear. The concern was to not change the way the existing iterators work (based on lazy copying). I don't think we strictly need a new interface, but we have to guarantee that existing usage patterns of the IAnnotationMap do still work. 

Locally patched, I've this in production across multiple Eclipse versions and it doesn't seem to break anything:

```
AnnotationMap.java

	@Override
	public void forEach(BiConsumer<? super Annotation, ? super Position> action) {
		synchronized (fLockObject) {
			fInternalMap.forEach(action);
		}
	}
```
AnnotationModel.java
```
	@Override
	protected void removeAllAnnotations(boolean fireModelChanged) {
		synchronized (getLockObject()) {
			AnnotationModelEvent event = getAnnotationModelEvent();
			var map = getAnnotationMap();
			map.forEach((annotation, position) -> {
				removePosition(fDocument, position);
				event.annotationRemoved(annotation, position);
			});
			map.clear();
		}

		if (fireModelChanged) {
			fireModelChanged();
		}
	}
```

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
szarnekow: I think this might be safe to change since both the AnnotationMap and AnnotationModel belong to the same bundle.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -789,15 +799,24 @@ public void removeAllAnnotations() {
...
+				while (e.hasNext()) {
+					Annotation a= e.next();
+					Position p= annotations.get(a);
+					removePosition(fDocument, p);
+					event.annotationRemoved(a, p);
 				}
+			} else {
+				// foreach is synchronized on map lock object
+				annotations.forEach((a, p) -> {
+					removePosition(fDocument, p);
```
iloveeclipse: @szarnekow: I think this line is the last reason I would keep on new Interface. The `removePosition()` can be overridden, and use extra locks / operations that might interfere with the new enclosing lock in the map.forEach().

szarnekow: I agree.

szarnekow: One more thought: Since the class `AnnotationMap` is package private, a check for that specific class might also do the trick.

szarnekow: But. of course that would harm extensibility

iloveeclipse: > check for that specific class might also do the trick

Jawoll!

szarnekow: > The removePosition() can be overridden, and use extra locks / operations that might interfere with the new enclosing lock in the map.forEach().


Maybe I don't agree.
Let's assume I've a subtype of AnnotationModel which uses the regular AnnotationMap - it now implements IAnnotationMap2.

How does the instanceof IAnnotationMap2 help me with my newly introduced locks in my overridden AnnotationModelSubtype?

iloveeclipse: > How does the instanceof help me with my newly introduced locks in my overridden AnnotationModelSubtype?

Doesn't help for existing code, right. I will check if we can skip changes on removeAllAnnotations method for now.

iloveeclipse: OK, simplified the patch as much as possible to avoid any side effects.

iloveeclipse:
@szarnekow : please check last version.

szarnekow:
Change looks safe, and I think it's an improvement. How much of a difference do you see in your benchmarks?

iloveeclipse:
Something around 40% less time spent.

szarnekow (approved):
LGTM

iloveeclipse:
Thanks Sebastian! It is really nice to have someone asking right questions :-)

Issue #983, opened 2023-08-01
 Saving Java file is very slow when the class has many markers


fedejeanne:
Saving the contents of a Java class can take a long time when the class has many markers (errors/warnings/search matches/etc). The reason for this is that updating the markers may trigger up to 3 workspace-operations and each operation triggers a lock release and a graph reduction.

Here's a snapshot of a save action that took over 30 seconds to complete.

![Image](https://user-images.githubusercontent.com/2205684/257476951-e4f23c5d-9a04-4bd3-b4fe-71243f89aaa0.png)

Each of the `set` methods in `MarkerUtilities` ends up calling `DeadLockDetector::reduceGraph` which ends up taking most of the time. Here's an example with `MarkerUtilities::setCharEnd`.

![Image](https://user-images.githubusercontent.com/2205684/257477885-49c23c6e-bc1b-4065-819f-a0665bf8ebfb.png)

jukzi:
> Here's a snapshot of a save action that took over 30 seconds to complete.

for how many markers approx?

fedejeanne:
> > Here's a snapshot of a save action that took over 30 seconds to complete.
> 
> for how many markers approx?

Difficult to say because I only got a sampling, so all I have is the _Hits_ count from VisualVM (it says **55** hits). The class isn't that big though and based on this incomplete screenshot I got, I'd say it couldn't have been more than 300 markers.

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/2205684/5f26be2a-79f5-4228-bf4e-becc7dd4cf58)

jukzi:
> couldn't have been more than 300 markers.

Could you please test with an synthetic File that has so many markers (or even many more) if it's really reproducible to get a 30s lag. And if any change finally fixed  it.
I really can't imaging 300 markers would take 30s.

fedejeanne:
I couldn't reproduce the error so far and I agree: under normal circumstances 300 markers don't take that long to be processed (even before #984 and https://github.com/eclipse-platform/eclipse.platform/pull/615). I can imagine something must have made these methods run very slow because they are the main reason for the poor performance:

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/2205684/c561a4b6-83dd-465e-8d25-7202e32254db)

[...]

Issue #981, opened 2023-08-01
org.eclipse.e4.tools.jdt.templates version need to be updated


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20230731-1800/buildlogs/reporeports/reports/versionChecks.html (search for "IUs in current repo that increase versions but with qualifier only").

Also Jenkins reports an error:

`Only qualifier changed for (org.eclipse.e4.tools.jdt.templates/4.10.0.v20230727-0604). Expected to have bigger x.y.z than what is available in baseline (4.10.0.v20230414-0427)`

Regression from f243cf0a28785b89b7c50bf4e1cce48a917d89bd

I will push a patch in a moment

PR #984, opened 2023-08-01
Update the marker in BasicMarkerUpdater::updateMarker in one operation

fedejeanne:
# What it does
The current implementation of `BasicMarkerUpdater::updateMarker` triggers up to 3 Workspace-Operations to change a marker in a class when saving, those are:
* One operation to change the line of the marker
* One operation to change where it begins
* One operation to change where it ends

This PR lets `BasicMarkerUpdater::updateMarker` gather all changes that need to be done to a marker in a `Map` and then apply all those changes in *one single Workspace-Operation*. This reduces the unnecessary overhead of opening/closing operations and all that it implies like looking for cycles in the `OrderedLock` class and especially reducing the internal graph used to search for them (`DeadlockDetector::reduceGraph`).

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/983

# How to test
* Open a Java class
* If the class does not have any markers, create some _e.g._ call a non-existing method
* Make some changes in the class
* Save --> Everything should work as usual:
  * The existing markers of Errors/Warnings in the class should be updated in the _Problems_ view if their position (line) has changed
  * A build should be triggered (if the _auto-build_ is enabled)
  * etc

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/BasicMarkerUpdater.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/BasicMarkerUpdater.java
@@ -13,11 +13,18 @@
  *******************************************************************************/
 package org.eclipse.ui.texteditor;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.core.runtime.CoreException;
+
 import org.eclipse.core.resources.IMarker;
 
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.Position;
+
+import org.eclipse.ui.internal.editors.text.EditorsPlugin;
 
 
 /**
@@ -63,19 +70,22 @@ public final class BasicMarkerUpdater implements IMarkerUpdater {
 		int markerStart= MarkerUtilities.getCharStart(marker);
 		int markerEnd= MarkerUtilities.getCharEnd(marker);
 
+
+		Map<String, Object> attributeChanges= new HashMap<>(3);
+
 		if (markerStart != -1 && markerEnd != -1) {
 
 			offsetsInitialized= true;
 
 			int offset= position.getOffset();
 			if (markerStart != offset) {
-				MarkerUtilities.setCharStart(marker, offset);
+				MarkerUtilities.setCharStart(attributeChanges, offset);
 				offsetsChanged= true;
 			}
 
 			offset += position.getLength();
 			if (markerEnd != offset) {
-				MarkerUtilities.setCharEnd(marker, offset);
+				MarkerUtilities.setCharEnd(attributeChanges, offset);
 				offsetsChanged= true;
 			}
 		}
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Note, build failure is due https://github.com/eclipse-platform/eclipse.platform.ui/issues/981.

iloveeclipse (approved):
Good catch.

fedejeanne:
> Good catch.

Thank you very much :-)

Maybe I can pick your brain regarding a possible further optimization (in another PR)?
I noticed that after my optimization, a class with **X** markers will still create up to **X** Workspace-Operations. Wouldn't it make sense to do at most 1 operation and update **all markers in the class at once**? If I'm reading the code correctly, this shouldn't have any impact since all `WorkspaceOperation::endOperation` does is broadcasting a post-change (via `broadcastPostChange`), firing up a build (if necessary) and doing a snapshot (if necessary).

Disclaimer: I haven't done the described optimization in this PR because it would require an API change and I wanted to keep this PR as small as possible while providing a quick win.

iloveeclipse:
> Wouldn't it make sense to do at most 1 operation and update **all markers in the class at once**

Yes, but we should discuss this on a different PR :)

DenisUngemach:
Hello, 

this change causes a severe error in the breakpoint markers.

Since now the method markers. setAttributes(Map<String, ? extends Object> attributes) will be used, all existing markers will be deleted and only the 3 markers char_start, char_end and line_number are set. 

The following bug comes up: 

In eclipse:
-> In a java class, set a breakpoint.
-> one line before the breakpoint add new lines with enter =>, the breakpoint moves down with its selected line
-> Now save the editor => the breakpoint will be disabled!

I recommend to create and use a MarkerUtilities method which updates the marker instead of the marker.setAttributes(attributesChanged) call.

An example see below.

Best regards, 
Denis Ungemach

Put this method in MarkerUtilities:

	/**
	 * Changes the given attribute key-value pairs of the map on this marker. The values must be
	 * <code>null</code> or an instance of one of the following classes: <code>String</code>,
	 * <code>Integer</code>, or <code>Boolean</code>. If a value is <code>null</code>, the new value
	 * of the attribute is considered to be undefined.
	 *
	 * <p>
	 * The values of the attributes cannot be <code>String</code> whose UTF encoding exceeds 65535
	 * bytes. On persistent markers this limit is enforced by an assertion.
	 * </p>
	 *
	 * <p>
	 * This method changes resources; these changes will be reported in a subsequent resource change
	 * event, including an indication that this marker has been modified.
	 * </p>
	 * 
	 * @param marker the marker
	 * @param attributeChanges map with to be executed attribute changes
	 * @see IMarker#setAttributes(String[], Object[])
	 */
	public static void changeAttributes(IMarker marker, Map<String, Object> attributeChanges) {

		Set<Entry<String, Object>> entries= attributeChanges.entrySet();
		int size= entries.size();

		String[] keys= new String[size];
		Object[] values= new Object[size];

		int i= 0;
		for (Entry<String, Object> e : entries) {
			keys[i]= e.getKey();
			values[i]= e.getValue();
			i++;
		}

		try {
			marker.setAttributes(keys, values);
		} catch (CoreException e) {
			handleCoreException(e);
		}

	}

laeubi:
@DenisUngemach do you like to propose a PR?

iloveeclipse:
@DenisUngemach : thanks and sorry, I've overlooked that completely.

I've set a PR: #994

DenisUngemach:
@iloveeclipse @fedejeanne Thanks for the fix and the performance optimizations! A batch marker update also sounds good.

PR #985, opened 2023-08-01
update to org.eclipse.emf.ecore 2.35.0

jukzi:


merks:
Note that the "important" changes are in org.eclipse.ecore.xmi and I need to add it to the target platform before you can lower bound to that version:

image

 I will be doing a milestone build after my nightly build finishes, and then I will update the target platform

jukzi:
@ignoring https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

PR #986, opened 2023-08-02
Enable codeQL

jukzi:


jukzi:
uses
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1239

PR #989, opened 2023-08-03
Improve problem filter scope wording

Bananeweizen:
* use "selected elements" instead of "selected element" in all implementations
* use "their children" due to the plural form of "elements"
* remove double blank in working set related scope


BeckerWdf (approved):

BeckerWdf:
Good catch!

mickaelistria:
Changes in documentation are OK even in RC, let's merge it now.

BeckerWdf:
are "[messages.properties](https://github.com/eclipse-platform/eclipse.platform.ui/pull/989/files#diff-1025ba66a3f389cd3073ef5bdec0d099e8a899a0d4dc9398c830c76005d0d760)" changes really "documentation"?

mickaelistria:
Not really, indeed.
However, I think it's OK to apply the rule for typos is labels as well. It's not a regression fix, but it's surely not a risky change either.

vogella:
Agreed, label fixes are fine even in rc1 or rc2. The rules are to help us
deliver good software not to prevent us from doing this

Mickael Istria ***@***.***> schrieb am Fr., 25. Aug. 2023,
16:01:

PR #990, opened 2023-08-03
Use parallel streams instead of forEach in AnnotationModel::cleanup

fedejeanne:
Parallel streams have a small overhead but the performance improves significantly when the `AnnotationModel` has thousands of elements.

See [my comment](https://github.com/eclipse-platform/eclipse.platform.ui/issues/892#issuecomment-1663415202) and the responses from @laeubi that inspired this PR.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/892

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 
 import org.eclipse.core.runtime.Assert;
 
@@ -639,11 +640,12 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 		if (fDocumentChanged) {
 			fDocumentChanged= false;
 
-			ArrayList<Annotation> deleted= new ArrayList<>();
+			List<Annotation> deleted;
 			IAnnotationMap annotations= getAnnotationMap();
 			Object mapLock = annotations.getLockObject();
 			
 			if (mapLock == null) {
+				deleted= new ArrayList<>();
 				Iterator<Annotation> e= annotations.keySetIterator();
 				while (e.hasNext()) {
 					Annotation a= e.next();
@@ -653,11 +655,13 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 				}
 			} else {
 				synchronized (mapLock) {
-					annotations.forEach((a, p) -> {
-						if (p == null || p.isDeleted()) {
-							deleted.add(a);
-						}
-					});
+					deleted= annotations.entrySet().parallelStream().unordered() //
+							.filter(e -> {
+								Position p= e.getValue();
+								return p == null || p.isDeleted;
+							})
+							.map(e -> e.getKey())
+							.collect(Collectors.toList());
 				}
 			}
 


bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -653,11 +655,10 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 				}
 			} else {
 				synchronized (mapLock) {
-					annotations.forEach((a, p) -> {
-						if (p == null || p.isDeleted()) {
-							deleted.add(a);
-						}
-					});
+					deleted = annotations.entrySet().parallelStream()//
```
laeubi:
```suggestion
					deleted = annotations.entrySet().parallelStream().unordered()//
```

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -639,11 +640,12 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 		if (fDocumentChanged) {
 			fDocumentChanged= false;
 
-			ArrayList<Annotation> deleted= new ArrayList<>();
+			List<Annotation> deleted;
 			IAnnotationMap annotations= getAnnotationMap();
 			Object mapLock = annotations.getLockObject();
 			
 			if (mapLock == null) {
+				deleted= new ArrayList<>();
```
laeubi, line 648: Why not using streams here as well? It seems the only difference is the lock here?

fedejeanne:
I tested this as described in [this comment](https://github.com/eclipse-platform/eclipse.platform.ui/issues/892#issuecomment-1659749243) from Andrey and sampled with VisualVM. Here are the results:

Before this PR:
image

After this PR:
image

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -653,11 +655,10 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 				}
 			} else {
 				synchronized (mapLock) {
-					annotations.forEach((a, p) -> {
-						if (p == null || p.isDeleted()) {
-							deleted.add(a);
-						}
-					});
+					deleted = annotations.entrySet().parallelStream()//
+							.filter(e -> e.getValue() == null || e.getValue().isDeleted)//
```
laeubi: `e.getValue` might be extracted into a local variable to prevent double method call.

szarnekow:
`keysetIterator` degraded from 4 seconds to 6 seconds with your patch. Do you have an explanation for this?

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
szarnekow: If the lock is not available, we might run into concurrent modifications. These are prevented by the defensive copy behind `keySetIterator`. It's a safety measure to not break existing subtypes / weird usages of the AnnotationModel.

laeubi: If you add a default method to `IAnnotationMap` like this:

```
default Stream<Annotation> keyStream() {
	return StreamSupport.stream(
			Spliterators.spliteratorUnknownSize(keySetIterator(), Spliterator.ORDERED),
			false);
}
```

then it could be used as a stream as well. and even if items are queried in parallel there should be no issue?

fedejeanne: I thought default methods in interfaces were forbidden [because of incompatibility issues](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.pde.doc.user%2Freference%2Fapi-tooling%2Fapi_evolution.htm)? Or are they just discouraged?

In any case I tried wit a spliterator and the performance dropped drastically.

fedejeanne:
> `keysetIterator` degraded from 4 seconds to 6 seconds with your patch. Do you have an explanation for this?

It seems unrelated to the changes. I just committed the changes proposed by Christoph and the performance of `keySetIterator` went up again. Every time I run it I get different values between 4 and 6s

image

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
laeubi: The interface already extends `Map` what has a lot of default methods and is out of control of JDT so i think it won't be much of an issue, but implementations can then override this with a more optimized stream (e.g. return a stream of the internal datastructure), please note that the splitter used for the default impl is really conservative, it it not parallel and it is ordered, both things will slow down things. e.g. in the impl `AnnotationMap` it first creates a copy and then returns an iterator of this that is then turned into a stream.

In general I think if we are talking about `AnnotationMap` it might be worth to use an `ConcurrentHashMap` as internal datastructure and remove all the synchronized and copy stuff (maybe wrap it into unmodifable map if that is a concern), also note that for removing something from a map could be performed using `java.util.Map.compute(K, BiFunction<? super K, ? super V, ? extends V>)` what depending on the map *might* be more performant here compared to a linear search as you perform the check and replace in one step.

szarnekow: All things considered I'd say it's ok to not add new methods to the already questionable interface and be extra conservative by keeping the existing code based on the iterator as is.

Also note that the real problem we are facing here is that the ProjectViewer obtains an AnnotationIterator for each match in the file, so for the given example, we are seeing like 10k invocations of cleanup. If things should be improved further, the FindReplaceTarget abstraction is the more promising attack vector.

laeubi: It might be worth to record then if cleanup is really required, e.g setting a boolean if something is removed/cleared in the model an check this first then cleanup will only happen if there are likely item to be cleaned.

fedejeanne: > It might be worth to record then if cleanup is really required, e.g setting a boolean if something is removed/cleared in the model an check this first then cleanup will only happen if there are likely item to be cleaned.

I think that's what `fDocumentChanged` is supposed to do, the problem is that `FindReplaceDialog::replaceAll` finds each occurrence (which ends up calling `getAnnotationsIterator` and `cleanup`), marks it and replaces it (which changes the document). This is the piece of code where that happens:

```java
int index= 0;
while (index != -1) {
	index = findAndSelect(...); // this one calls getAnnotationIterator and cleanup
	if (index != -1) { // substring not contained from current position
		// this one sets fDocumentChanged to true
		Point selection= replaceSelection(replaceString, regExSearch);
		...
	}
}
```

This is what you meant before, right @szarnekow ?

szarnekow: Yes, exactly.

fedejeanne: I've been looking at it and I'm out of ideas: any suggestions on how to wrap all changes into one session and fire a document change at the end of the loop? Maybe we can discuss this in another PR?

iloveeclipse: > Maybe we can discuss this in another PR?

On another **issue** please.

fedejeanne: I created https://github.com/eclipse-platform/eclipse.platform.ui/issues/995, feel free to take over.

Thank you all for your input and for helping improve this code even more!

I think this PR is now ready to be merged, would you agree?

szarnekow: Without this change, I get fairly consistent numbers (showing only one of many experiments here):
```
ReplaceAll: 13543ms
Undo: 6926ms
Redo: 4773ms
Undo: 6311ms
ReplaceAll: 12204ms
Undo: 6989ms
Redo: 5155ms
Undo: 6349ms
ReplaceAll: 13077ms
```

After having applied this change locally, I see way more flaky numbers, though some of them are sometimes better than originally:
```
ReplaceAll: 11329ms
Undo: 8183ms
Redo: 4729ms
Undo: 6209ms
ReplaceAll: 9549ms
Undo: 6339ms
Redo: 4717ms
Undo: 6645ms
ReplaceAll: 23934ms.  <--- ???
Undo: 7332ms
ReplaceAll: 10243ms
Undo: 7099ms
ReplaceAll: 12019ms.
Undo: 7451ms
Redo: 5542ms
Undo: 6634ms
ReplaceAll: 9602ms
```

When I only look at the ReplaceAll timings, it's a nice improvement on avg. But the fact that the numbers seem to indicate degradation of `Undo` timings by some small percentage makes me sad. Mostly because the code *looks* fine, and I cannot explain these numbers yet. There might be an impact on other operations beyond the usage pattern behind ReplaceAll. 

What concerns me, too, is that I get consistently higher memory usage with the change applied - which I cannot explain now.

@iloveeclipse What do you think?

iloveeclipse: I would assume depending on common pool state the times may vary. 
Considering there is not much improvement given, I would prefer to keep the old/simple/non parallel code and spend time on analysis / fix of #995.

szarnekow: For completeness:
The numbers in the JavaEditor with Code-Folding disabled and without the parallel streams:
```
ReplaceAll: 2545ms
Undo: 4074ms
Redo: 2260ms
Undo: 4566ms
ReplaceAll: 4924ms
Undo: 4303ms
ReplaceAll: 2260ms
```

Same sequence of operations in the plain text editor:
```
ReplaceAll: 331ms
Undo: 188ms
Redo: 103ms
Undo: 141ms
ReplaceAll: 216ms
Undo: 252ms
ReplaceAll: 187ms
```

fedejeanne:
Since these changes didn't bring that much of a performance boost and also degraded the performance on other use-cases, this PR is **not** going to be be merged.

Thank you all for your input and for testing!

PR #994, opened 2023-08-03
Fix regression on marker updates from #984

iloveeclipse:
See 7da51e78950bfec17f05361f7d64cd817eb5b72c and comment from Denis Ungemach on https://github.com/eclipse-platform/eclipse.platform.ui/pull/984

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/MarkerUtilities.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/MarkerUtilities.java
@@ -17,6 +17,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
 
 import org.osgi.framework.Bundle;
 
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/MarkerUtilities.java
```diff
@@ -402,4 +404,50 @@ public static String[] getSuperTypes(String markerType) {
...
+	 * event, including an indication that this marker has been modified.
+	 * </p>
+	 * 
+	 * @param marker the marker
+	 * @param attributeChanges map with to be executed attribute changes
+	 * @see IMarker#setAttributes(String[], Object[])
+	 * @since 3.17
+	 */
+	public static void changeAttributes(IMarker marker, Map<String, Object> attributeChanges) {
+
```
fedejeanne, line 430: How about an early return to avoid possibly unnecessary calculations? 
```suggestion
if (!marker.exists()) {
   return;
}
```

You could then remove the `if` in line 445

iloveeclipse: Early check will not help if marker is deleted later. Of course that can happen right after the check, but at least the code would be consistent with the rest of the code here that checks the merker state before changing it.

iloveeclipse:
@DenisUngemach : could you please review?

DenisUngemach (commented):
Yes this is exactly what i meant. I think the method changeAttributes should be fine as public API.

PR #997, opened 2023-08-05
Fix possible NPE in PartServiceImpl

### LINKED ISSUE #996: NPE in PartServiceImpl
I somehow get into a state for a custom RCP appplication where the part service throws an NPE because there is no active leaf. As the code there tries to query a part and can handle null it should also handle this case as "no part found".

```
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.e4.core.contexts.IEclipseContext.getActiveLeaf()" because the return value of "org.eclipse.e4.ui.model.application.ui.advanced.MPerspective.getContext()" is null
	at org.eclipse.e4.ui.internal.workbench.PartServiceImpl.switchPerspective(PartServiceImpl.java:634)
	at <some usercode>
	... 18 more
```

laeubi:
In the init phase it seem to happen sometimes that the eclipse context is not (yet?) set for a perspective but some user code has already access to the model (e.g. model processors) then a NPE can happen interupting the whole application startup.

Fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/996


laeubi:
@merks as the M2 is already published it is safe to merge this now?

merks:
There was no freeze for m2 so committing has been safe for weeks. I'm sure the releng folks will send an email when m3 approaches. 👍

PR #999, opened 2023-08-06
Fix random failures of FoldersAsProjectsContributionTest #946

zelenyhleb:
Projects location is now created explicitly by JUnit rule. Also creating project description file if it does not exist for some reason explicitly (with empty content, just to make sure that `description::exists()` will return true).


zelenyhleb:
Hello @iloveeclipse! Please have a look.
I added some additional checks that helped me to find out that sometimes project descriptions files (used by algorithm to check if folder contains a project) were not created or were deleted during test execution. Now test creates them explicitly using java.nio API instead of platform, because platform call even with force=true still does not lead to existing project description file sometimes (needs further investigation, I guess).
As I can see, now test is stable for last few builds in this PR. Hope it will be stable in master too.

iloveeclipse:
> platform call even with force=true still does not lead to existing project description file sometimes (needs further investigation, I guess).

Yes, interesting, may be some async work is going on there.

> As I can see, now test is stable for last few builds in this PR. Hope it will be stable in master too.

Appreciate your efforts and thanks for your time. Let's merge.

PR #1003, opened 2023-08-10
Fix spurious content-assist popup on MacOS

ahmedneilhussain:
Ported https://github.com/eclipse-platform/eclipse.platform.text/pull/150 from old repository

Original summary:
I think this was introduced by https://github.com/eclipse-platform/eclipse.platform.text/commit/78fc9f6f6566323c691f1fee1eb14de2f27e7d34

Have observed with LSP4e-based editors in Eclipse 2022-12, but presumably would affect everything that uses the jface text editor and has a content assistant installed.

To reproduce (on MacOS):
1. E.g. open a Rust file (with Eclipse Corrosion plugin installed) or a javascript file using Wild Web Developer
2. Position the cursor on some whitespace on a non comment line
3. Having made sure the file is already saved, press cmd-S to 'save' again.

The content assistant popup will be shown. This also seems to occur if you press cmd-? where ? is any letter not bound to a command, or bound to a command that doesn't succeed and mark the event as consumed. E.g. cmd-C with no text highlighted will also provoke the popup.

I guess this isn't an issue on Windows/Linux because ctrl is used as the primary modifier key, and that gets encoded as part of the character code point and gets filtered out by `Character.isLetter(c) || Character.isDigit(c)` on L1242.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
@@ -340,7 +340,7 @@ public class ContentAssistant implements IContentAssistant, IContentAssistantExt
 			if (e.character == 0 && (e.keyCode & SWT.KEYCODE_BIT) == 0)
 				return;
 
-			if (e.character != 0 && (e.stateMask == SWT.ALT))
+			if (e.character != 0 && (e.stateMask == SWT.ALT || ((e.stateMask & SWT.COMMAND) != 0)))
 				return;
 
 			TriggerType triggerType= getAutoActivationTriggerType(e.character);


ahmedneilhussain:
Can confirm this is still a bug on MacOS in 2023-06.

BeckerWdf:
Thanks for porting the PR to this repo.

Can you address the comment in https://github.com/eclipse-platform/eclipse.platform.text/pull/150#issuecomment-1476077695 a well?

BeckerWdf:
What does this: 
image
mean?

Phillipus:
I notice the use of `SWT.COMMAND` in the patch. Will it work if, like I do, swap the Ctrl and Command keys in Mac's settings? Perhaps it should be `SWT.MOD1`?

ahmedneilhussain:
> I notice the use of `SWT.COMMAND` in the patch. Will it work if, like I do, swap the Ctrl and Command keys in Mac's settings? Perhaps it should be `SWT.MOD1`?

I tried this to check - if you remap the hotkeys in MacOS it only affects the physical button the user has to press to achieve a certain effect, once it makes it as far as Eclipse the keystroke that the OS informs us about is already the logical one.

So if you make control act as command and press control-S, the keystroke that comes up in eclipse is command-S : I verified this by turning on the key binding command debugging in preferences (shows the key and command in a transient overlay on the monitor)

ahmedneilhussain:
> Thanks for porting the PR to this repo.
> 
> Can you address the comment in [eclipse-platform/eclipse.platform.text#150 (comment)](https://github.com/eclipse-platform/eclipse.platform.text/pull/150#issuecomment-1476077695) a well?

Hi, I had a quick look. My PR definitely doesn't fix it, but I think this test failure is a different issue.

My PR is trying to fix a case where the content assistant gets activated when it shouldn't. The test failure concerns a test where the content assistant is **supposed** to get activated: however the popup then fails to be dismissed correctly on MacOS.

AFAICT the test drives the content assistant using programmatic commands to the editor, so operates fully downstream of the key->command event processing that this PR addresses.

BeckerWdf:
@ahmedneilhussain Can you rebase this change to the newest master commit?

ahmedneilhussain:
> @ahmedneilhussain Can you rebase this change to the newest master commit?

Hi, done.

I've also verified still a bug in 2023-09.

ahmedneilhussain:
Hi @BeckerWdf  @mickaelistria it looks like my PR has now been superseded by https://github.com/eclipse-platform/eclipse.platform.ui/pull/1556 ??

BeckerWdf:
> Hi @BeckerWdf @mickaelistria it looks like my PR has now been superseded by #1556 ??

Indeed. So i am closing this.

BeckerWdf:
> > Hi @BeckerWdf @mickaelistria it looks like my PR has now been superseded by #1556 ??
> 
> Indeed. So i am closing this. Anyway: Thanks for your effort.

PR #1004, opened 2023-08-11
Decorated Label provider in the ResourceDialog

gzsombor:
So for in the resource selection dialog, if there were git rebase conflicts in a file, they will show up differently


BeckerWdf:
Can you please provide e.g. screenshots of the before and after state?

gzsombor:
Sure, here is the original :

no-marker

Here, with the marker:
with-marker

It aint't much, but honest work :)

BeckerWdf:
@gzsombor: Thanks for your contribution.

Issue #869, opened 2023-04-02
Standard text editor ignores preferences for foreground color


peterdell:
The following issue is present in Version: 2022-12 (4.26).
I have the dark theme active. When I create a plain text .txt file and put some there is is displayed in the light grey color configured in the preferences "General / Editors / Text Editors / Appearance and color options / Foreground color". But changing the color there has no effect. All the other color settings there (selection foreground and background, line numbers, .... ) work correctly. In this example I've set the foreground to purple:

![image](https://user-images.githubusercontent.com/10855380/229361583-4bd6ef40-cc6d-4bb7-b676-76032a985fc2.png)

laeubi:
> The following issue is present in Version: 2022-12 (4.26).

Have you tried 2023-03 already?

peterdell:
Oh, sure, I should have verified there first. But I did now and yes, the issue is the same in https://download.eclipse.org/eclipse/downloads/drops4/R-4.27-202303020300/download.php?dropFile=eclipse-platform-4.27-win32-x86_64.zip.

mickaelistria:
I suspect the Theme engine incorrectly overrides the color here. Can you please try disable theming to verify the preference setting work?

peterdell:
You are right, if theming is disabled, the preferences setting is used.

DenisUngemach:
The color which overrides the preference is in the plugin org.eclipse.ui.themes in file e4-dark_partstyle.css the color: #AAAAAA line 59. 
The override happens if the TextEditor gets focus.
With a breakpoint in StyledText in the method setForeground(Color color)  it is possible to get the stack trace of the stylesheet override.

DenisUngemach:
Bug already existed: [Dark Theme: Custom editor background color not properly applied](https://bugs.eclipse.org/bugs/show_bug.cgi?id=559321) 
Related to this might be:  [Move Color options from Text Editors to the Colors and Fonts preference page](https://bugs.eclipse.org/bugs/show_bug.cgi?id=531033)  

Further discussion about setForeground/setBackground: https://github.com/eclipse-platform/eclipse.platform.ui/issues/306

BeckerWdf:
this was fixed by https://github.com/eclipse-platform/eclipse.platform.ui/pull/1036

PR #1036, opened 2023-08-16
Abstract Text Editor: don't change StyledText by CSS.

DenisUngemach:
Fix for #869

Change-Id: Ib739afdc301c4e37dcbbe37ef063ed32cb7f53d7

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -2276,6 +2276,13 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 	/** The width of the vertical ruler. */
 	protected static final int VERTICAL_RULER_WIDTH= 12;
 
+	/*
+	 * The colors of AbstractTextEditor can be configured by the "Text Editors"
+	 * preferences. This color configuration will be changed by theming. This means
+	 * no need for CSS basecolors here.
+	 */
+	private static final String DISABLE_CSS = "org.eclipse.e4.ui.css.disabled"; //$NON-NLS-1$
+
 	/**
 	 * The complete mapping between action definition IDs used by eclipse and StyledText actions.
 	 *
@@ -3365,6 +3372,8 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 
 		int styles= SWT.V_SCROLL | SWT.H_SCROLL | SWT.MULTI | SWT.BORDER | SWT.FULL_SELECTION;
 		fSourceViewer= createSourceViewer(parent, fVerticalRuler, styles);
+
+		fSourceViewer.getTextWidget().setData(DISABLE_CSS, Boolean.TRUE);
 
 		if (fConfiguration == null)
 			fConfiguration= new SourceViewerConfiguration();
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -2276,11 +2276,6 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 	/** The width of the vertical ruler. */
 	protected static final int VERTICAL_RULER_WIDTH= 12;
 
-	/*
-	 * The colors of AbstractTextEditor can be configured by the "Text Editors"
-	 * preferences. This color configuration will be changed by theming. This means
-	 * no need for CSS basecolors here.
-	 */
 	private static final String DISABLE_CSS = "org.eclipse.e4.ui.css.disabled"; //$NON-NLS-1$
 
 	/**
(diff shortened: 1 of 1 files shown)


DenisUngemach:
In this change i just deactivate the CSS stylesheet, which overrides the colors from the "Text Editors" preferences.
Can someone confirm whether this is the right way to solve #869 .

And by the way issue #869 also affects the java class editor. That can be checked with the background color.

BeckerWdf:
> Fix for #869
> 
> Change-Id: Ib739afdc301c4e37dcbbe37ef063ed32cb7f53d7

We don't use Gerrit. So we don't need a Change-Id

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -2277,6 +2277,13 @@ private int computeOffsetAtLocation(ITextViewer textViewer, int x, int y) {
 	/** The width of the vertical ruler. */
 	protected static final int VERTICAL_RULER_WIDTH= 12;
 
+	/*
+	 * The colors of AbstractTextEditor can be configured by the "Text Editors"
+	 * preferences. This color configuration will be changed by theming. This means
+	 * no need for CSS basecolors here.
+	 */
+	private static final String DISABLE_CSS = "org.eclipse.e4.ui.css.disabled"; //$NON-NLS-1$
```
BeckerWdf, line 2279: here does this "magic" constant come from?

DenisUngemach: From org.eclipse.e4.ui.css.swt: AbstractCSSSWTEngineImpl line 117

DenisUngemach:
This feature has already been used in: 
See EGit's [UIUtils](https://git.eclipse.org/r/c/egit/egit/+/190575/2/org.eclipse.egit.ui/src/org/eclipse/egit/ui/UIUtils.java) class

vogella:
@BeckerWdf what is the status here with the review? Did you test the change?

BeckerWdf:
@DenisUngemach will update his PR. I then will do a manual test.

sratz:
In princple, this works well.

But we were wondering whether disabling CSS completely is the way to go. The preferences only deal with foreground + background colors, a CSS might do more than just that.

In theory that is - in practice, the CSS found in Platform only deals with foreground + background colors as well.

So we are probably safe here and if there are no objections we can start with this and wait for feedback.

If needed, we could maybe later introduce a more fine-granular E4 property, such as:

```
setData("org.eclipse.e4.ui.css.disabledProperties", "background-color,foreground-color")
```

However, your alternative PR #1349 also has an additional condition `getPreferenceStore() != null`
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1349/files#diff-c37eb3db119ede5b1695bb0217864648bfa40dfaafcfa25c7bbd7131d69ed641R3401

That might be useful to add as a safeguard in case `AbstractTextEditor` is used in a non-platform context without attached preference (is that even possible?). In that case, we wouldn't want to disable CSS.

vogella:
I don't think the extra check is necessary.

> a CSS might do more than just that.

As for the colors of the syntax highligthing, we manipulate the preference values directly, so we should be ok.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -3366,6 +3373,8 @@ public void createPartControl(Composite parent) {
 		int styles= SWT.V_SCROLL | SWT.H_SCROLL | SWT.MULTI | SWT.BORDER | SWT.FULL_SELECTION;
 		fSourceViewer= createSourceViewer(parent, fVerticalRuler, styles);
 
+		fSourceViewer.getTextWidget().setData(DISABLE_CSS, Boolean.TRUE);
```
BeckerWdf, line 3374: In @vogella 's proposal there was a nice comment why it's ok to disable the styling. Can you add a similar comment?

BeckerWdf (changes requested):

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
DenisUngemach: Done

BeckerWdf:
I did a manual test on macOS.
Without this change I can reproduce this issue 
With this change the issue is gone. Change to background and foreground color in the TextEditors preference page are not correctly handled. I teste editor close and re-opening as well as workbench restart.
All looks good for me.

vogella:
GH says, you still require changes @BeckerWdf ?

BeckerWdf:
> GH says, you still require changes @BeckerWdf ?

Let's wait for the build..

vogella:
> > GH says, you still require changes @BeckerWdf ?
> 
> Let's wait for the build..

We (should) always wait for the build. I wanted to know if you still require changes in this PR? @DenisUngemach could finish it  in this case.

BeckerWdf (approved):

BeckerWdf:
Are version number increases necessary for this change?

vogella:
> Are version number increases necessary for this change?

No, the build would fails if version increases are necesarry. I did not check if the test errors relate to this change.

BeckerWdf:
> No, the build would fails if version increases are necessary. 

So has the version number already been increased?

> I did not check if the test errors relate to this change.

The seem to be unrelated to me.

BeckerWdf:
Test failures are not related and already reported in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1351

vogella:
Cool, thanks @DenisUngemach and @BeckerWdf

DenisUngemach:
> Cool, thanks @DenisUngemach and @BeckerWdf

You're welcome. Actually my first idea was to create a custom css configuration: "[CSS support for custom widgets](https://www.vogella.com/tutorials/Eclipse4CSS/article.html#css-support-for-custom-widgets)"for the AbstractTextEditor and set there the color configuration. But this didn't work that well because somewhere in the compatibility handler css class and css id will be defined:

image

And i only wanted to write the css class. Is this css ID writing justified? Maybe these base layers should override css class, but i think the css ID is over the top.

Best regards,
Denis

PR #1037, opened 2023-08-16
Disable testCompletionService() which is unstable for a long time

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/906


iloveeclipse:
Eclipse.org cbi infra is down, but github actions are fine & local test is also fine => merging.

PR #1052, opened 2023-08-21
org.eclipse.ui should bump version because of updated API in reexported org.eclipse.jface and org.eclipse.ui.workbench bundles

### LINKED ISSUE #1051: SoftLimit: org.eclipse.ui should bump version because of updated API in reexported org.eclipse.jface and org.eclipse.ui.workbench bundles
When i OOmph product "Eclipse IDE for Eclipse Committers" + project "Platform" then i end up in a workspace with this error:
image
```
Failed to create the part's controls
java.lang.NoClassDefFoundError: org/eclipse/ui/views/WorkbenchViewerSetup
	at org.eclipse.jdt.internal.ui.packageview.PackageExplorerPart.createPartControl(PackageExplorerPart.java:499)

```	
@iloveeclipse some dependency missing?

iloveeclipse:
Note, there are two commits to separate the actual change needed from extra "cleanup" change to prevent similar issues for other API added.

- org.eclipse.ui should bump version because of updated API in reexported org.eclipse.jface and org.eclipse.ui.workbench bundles
- Additional version bumps for workbench & IDE related bundles: updated dependencies to match most recent major.minor versions from dependent bundles to make sure whatever API was added and in whichever combination p2 installs latest dependencies containing most recent API level.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1051


iloveeclipse:
@merks or @jukzi : please review this PR for RC1. The first commit is the main one, second would be good to have to prevent similar but not yet found issues.

I plan also [update JDT.UI](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/730#issuecomment-1686274885) to really make sure we don't run in this (or similar) problems.

merks:
Updating the lower bounds is of course a good idea.  Unfortunately lower bounds are in general very poorly maintained, but that's a challenging problem to fix via automation...

iloveeclipse:
Note, Mac change seem to be some sporadic API tools issue (probably because most recent SDK IBuild was deployed in parallel). Later executed Windows/Linux builds are green.

merks (approved):
Clearly the lower bounds are unreasonably low and updating them to more recent version is a good idea.

jukzi:
LGTM. Would be cool to have a Script that all minimal versions to the current one.

iloveeclipse:
> LGTM. Would be cool to have a Script that all minimal versions to the current one.

I've created https://github.com/eclipse-pde/eclipse.pde/issues/716.
This is not the first time we run in such kind of issues...

Issue #1014, opened 2023-08-15
SoftLimit: provide support for "Show in ..." & "Link with Editor"


jukzi:
regression from https://github.com/eclipse-platform/eclipse.platform.ui/issues/818

When i use "Show in ... "/ "Project Explorer" or the "Link with Editor" button in the Project Explorer with a File that is currently not visible in the Tree those actions do nothing. They must reveal the selected item!
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/5879a818-98a9-4ed6-82db-7994154d4883)

Package Explorer at least asks
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/474c0b7a-ca81-44e7-971d-faa58d82330e)

... but that does not work either :-(

Example 
org.eclipse.ui.internal.views.markers.ExtendedMarkersView with default Increment set to 10
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/847a4112-9f79-4624-b298-85f82912c214)

"Show in .. Outline" at least jumps to "Show .... (more)" but also does not reveal anything.

iloveeclipse:
Jörg, this is expected side effect, because selecting editor means expand **all** children up to the selected one would hang UI again assuming the number of hidden items (== limit) is high enough.

Note, that the limit of 10 is not default. So I tend to close this issue as "won't fix".

Actually this use case is one of the most prominent "hang" reasons we've got reported by users.

laeubi:
>  because selecting editor means expand all children up to the selected one would hang UI again

As the viewer actually has all element loaded it would be possible to find this element and only show this (e.g. temporary) single element.

> Note, that the limit of 10 is not default. So I tend to close this issue as "won't fix".

does it matter if it is the 10 or 1000 element that is not shown? How would a user recover from this case? Expand all elements by hand and hope to find the one of interest in reasonable time?

iloveeclipse:
> it would be possible to find this element and only show this (e.g. temporary) single element.

In theory yes, in practice the problem would be that these elements will appear "unsorted" once user clicks "show more.." to see more elements. But we can evaluate if we can remember elements shown "out of order" to remove them later on. 

> does it matter if it is the 10 or 1000 element that is not shown?

Yes, most users will be not affected by the limit of 1000 elements. So we are talking here about not common use case.

> How would a user recover from this case? Expand all elements by hand and hope to find the one of interest in reasonable time?

[...]

PR #1061, opened 2023-08-23
SoftLimit on viewers: hide ExpandableNode

mickaelistria:
* Make that selection never contains ExpandableNode, since it's not part of model
* Selecting a not-yet-expanded node does nothing, similarly to how filtered elements are not selectable
* Make that only the TreeItem is accepted to check whether it's the expansion placeholder or not, to clarify it's not a model element we're chasing here.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -24,7 +24,6 @@ package org.eclipse.jface.viewers;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 
@@ -2644,40 +2643,7 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 				}
 			}
 		}
-
-		// there can be some items inside expandable node and not populated yet. In this
-		// case try to find the item to select inside all the visible expandable nodes.
-		if (newSelection.size() < v.size() && getItemsLimit() > 0) {
-			// make out still not found items
-			List<Object> notFound = new ArrayList<>();
-			for (Object toSelect : v) {
-				boolean bFound = false;
-				for (Item found : newSelection) {
-					if (equals(toSelect, found.getData())) {
-						bFound = true;
-						break;
-					}
-				}
-				if (!bFound) {
-					notFound.add(toSelect);
-				}
-			}
-
-			// find out all visible expandable nodes
-			Collection<ExpandableNode> expandItems = getExpandableNodes();
-
-			// search for still missing items inside expandable nodes
-			for (Object nFound : notFound) {
-				for (ExpandableNode expNode : expandItems) {
-					if (findElementInExpandableNode(expNode, nFound)) {
-						Widget w = findItem(expNode);
-						if (w instanceof Item item) {
-							newSelection.add(item);
-						}
-					}
-				}
-			}
-		}
+		// invisible items (part of expandable node, or filtered) are ignored
 
 		setSelection(newSelection);
 
@@ -2692,16 +2658,6 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 				showItem(newSelection.get(i));
 			}
 		}
-	}
-
-	private boolean findElementInExpandableNode(ExpandableNode expNode, Object toFind) {
-		Object[] remEles = getFilteredChildren(expNode);
-		for (Object element : remEles) {
-			if (equals(element, toFind)) {
-				return true;
-			}
-		}
-		return false;
 	}
 
 	/**
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTableViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTableViewer.java
@@ -964,31 +964,7 @@ public abstract class AbstractTableViewer extends ColumnViewer {
 				if (count < indices.length) {
 					System.arraycopy(indices, 0, indices = new int[count], 0, count);
 				}
-
-				// item to select may be hidden inside expandable node.
-				if (getItemsLimit() > 0 && indices.length < list.size()
-						&& getLastElement() instanceof ExpandableNode expNode) {
-
-					// extract only non found items already.
-					List<Object> notFoundItems = new ArrayList<Object>(list);
-					for (int index : indices) {
-						notFoundItems.remove(doGetItem(index).getData());
-					}
-
-					Object[] remEles = expNode.getRemainingElements();
-					OuterLoop : for (Object searchItem : notFoundItems) {
-						for (Object remEle : remEles) {
-							if (equals(remEle, searchItem)) {
-								int[] placeHolder = new int[indices.length + 1];
-								System.arraycopy(indices, 0, placeHolder, 0, indices.length);
-								placeHolder[placeHolder.length - 1] = items.length - 1;
-								indices = placeHolder;
-								break OuterLoop;
-							}
-						}
-					}
-				}
-
+				// invisible items (part of expandable node, or filtered) are ignored
 				doSelect(indices);
 			}
 		}
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewer.java
@@ -1063,13 +1063,14 @@ public abstract class ColumnViewer extends StructuredViewer {
 	 * it will remain the same. Please do not use this API without consulting with
 	 * the API development team.
 	 * </p>
+	 * This code must run in the UI Thread.
 	 *
-	 * @param element model object representing a special "expandable" node
-	 * @return return if it is an instance of ExpandableNode
+	 * @param item widget item
+	 * @return return if the item is the expansion placeholder
 	 * @since 3.31
 	 */
-	public final boolean isExpandableNode(Object element) {
-		return element instanceof ExpandableNode;
+	public static final boolean isExpandableNode(Item item) {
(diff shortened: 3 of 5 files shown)


iloveeclipse:
Please checkout jdt.ui project, see compiler errors and try to use SDK with this patch. It will not work. Code in classes below doesn't have `Item`, it only has model objects. So `org.eclipse.jface.viewers.ColumnViewer.isExpandableNode(Item)` API can't be used there.
- org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaOutlinePage.java
- org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/packageview/PackageExplorerPart.java
I haven't checked the rest of the patch, since that must be fixed first.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -2644,40 +2643,7 @@ protected void setSelectionToWidget(@SuppressWarnings("rawtypes") List v, boolea
...
-					if (findElementInExpandableNode(expNode, nFound)) {
-						Widget w = findItem(expNode);
-						if (w instanceof Item item) {
-							newSelection.add(item);
-						}
-					}
-				}
-			}
-		}
+		// invisible items (part of expandable node, or filtered) are ignored
```
iloveeclipse, line 2646: This breaks "link with selection" in the sense, that before this change, we selected the node contained "hidden element", so user could directly click on it to expand & eventually see the file. Now no selection updated at all.

iloveeclipse (changes requested):
This change breaks JDT support and makes "link with editor" even worse as before.

mickaelistria:
`getSelection` and `link with selection` is still broken in the sense that the selected element in the editor area is still not selected (but the expandable node is). It's better to return nothing and break "link with selection" that to break "getSelection" so it returns virtual/internal/non-model objects. There are possible alternatives such as expanding the node until the target element is visible which would be much more correct, or if the expansion is too heavy because there are too many elements to show, maybe just expand the range of the selected element and show 1 expansion placeholder above and another below the newly expanded block.
But IMO, the current state of `getSelection` is deeply bogus compared to the legacy behavior which is as important as a contract: only selected model elements should be returned. I would rather see "link with selection" not support for expandable trees in next release than introducing such a critical change in `Viewer.getSelection()`.

I will look at JDT consumers soon.

mickaelistria:
I see the JDT consumers can most likely be removed as they do delegate to `hasFilteredChildren(parent)` or `super.isExpandable(parent)` for which the JFace TreeViewer implementation is supposed to return the right value (`false`) for the expandable node anyway.

iloveeclipse:
> I see the JDT consumers can most likely be removed

If you have a solution, please create a PR. Without changes the JDT UI simply doesn't work properly because on every click on the "special node" they create an extra tree level and add the children there and not as expected at the same level.

iloveeclipse:
> It's better to return nothing and break "link with selection" that to break "getSelection" so it returns virtual/internal/non-model objects.

I disagree, I do see a value in selecting the closest element to the hidden one and I see no issue with returning "special" node in selection.

> There are possible alternatives such as expanding the node until the target element is visible which would be much more correct

Which turns the entire idea ad absurdum, because it is exactly the use case that hangs UI.

> just expand the range of the selected element and show 1 expansion placeholder above and another below the newly expanded block.

If you have improvements, please provide PR's. But as long as this feature is not impemented, one should not break existing code that, if not perfekt, provides a better user experience as before.


> only selected model elements should be returned.

What is the problem with that? Where the Viewer.getSelection() says anything about concrete selection **content**? Can you point me to this?

> I would rather see "link with selection" not support for expandable trees in next release than introducing such a critical change in Viewer.getSelection().

Why is this critical? I do not see a problem you are trying to fix. Viewers that want to use new code need to be manually changed and if needed adopted, no viewer is automatically forced to use new functionality.

mickaelistria:
> I disagree, I do see a value in selecting the closest element to the hidden one and I see no issue with returning "special" node in selection.

"Link with selection" can maybe be improved to better support that case and select the placeholder element in case the element is not visible yet, that wouldn't be too bad. It's more `getSelection()` which can now return incorrect values.

> If you have improvements, please provide PR's. But as long as this feature is not impemented, one should not break existing code that, if not perfekt, provides a better user experience as before.

My concern is that we're near a release and we're about to release a change that does break important stable 20-years-old invariant (getSelection only return items that are from the model, returned by ContentProvider). Sure, the breakage is now opt-in, but what if we change the viewers default to incremental and some consumers have plugged in selection listeners that have worked forever and will now have to deal with this ExpandableNode in selection? The answer has already been found in Platform issues: some ClassCastException and other errors will pop-up; same errors will show-up in 3rd-party consumers of such viewers. Those errors will cause a worse user experience than before.

> What is the problem with that? Where the Viewer.getSelection() says anything about concrete selection content? Can you point me to this?

I think I remember you usually advocate that a stable behavior of some API is as important,or even more, than the contract of the API. This is an instance of this stability rule.

> Viewers that want to use new code need to be manually changed and if needed adopted, no viewer is automatically forced to use new functionality.

Imagine the viewer is changed, but is exposed and opened to 3rd-party extension to add listeners (example: the CNF navigator allows to plug many listeners to the Project Explorer). Those "legacy" listeners can then fail because the viewer can suddenly return unexpected selections, causing consumers to fail.


Again, I'm not trying to ditch the whole change, as I fully agree it does provide a better UX, it's just that the current API opens a room for bugs to happen in stable code, and I don't think it's a good thing to let it happen in a release.

If I manage to change `link with selection` to select the "Show more..." node in case the element is not available yet, would it be OK for you to apply this patch?

iloveeclipse:
> if we change the viewers default to incremental

We will not do that, the change breaks behavior and so can't be made default

> and some consumers have plugged in selection listeners that have worked forever and will now have to deal with this ExpandableNode in selection?

This was addressed already via https://github.com/eclipse-platform/eclipse.platform.ui/pull/1035

> I think I remember you usually advocate that a stable behavior of some API is as important,or even more, than the contract of the API. This is an instance of this stability rule

Because usually new API isn't optional. Here no one client is broken without explicitly accepting that new invariant and consequences.

> just that the current API opens a room for bugs to happen in stable code

First of all, this might happen with any code change. Second, by default new behavior is disabled anyway. And if someone enables it and finds new issue, we would be happy to fix it (as with any other code change).

> If I manage to change ... 

If the user visible functionality remains, also for JDT UI, wh, not? 

But see above,  I'm pretty confident that stability shouldn't be affected and all new API is marked as experimental anyway. So if there will be a major refactoring later we should be fine changing this experimental API. Just consider this as a Java "preview" feature that might change. Everyone is warned - if you turn it on, you might need to adopt your code in the future. But this "preview" is the good way to find better solution and move forward. So why shouldn't we just do same here?

mickaelistria:
> This was addressed already via https://github.com/eclipse-platform/eclipse.platform.ui/pull/1035

Yes, but the unit test that are changed with this commit do highlight that it's still possible to get the ExpandableNode returned by JFace's `getSelection()`. I see my proposal here is a continuation of #1035

> Here no one client is broken without explicitly accepting that new invariant and consequences.
> ...
> But see above, I'm pretty confident that stability shouldn't be affected


That's unfortunately not entirely true, and the various "fixes" like the ones that were implemented in JDT's and other viewers are demonstrating the risk associated with that change for established viewers.
I've explained it in my previous message, but let me repeat it: some APIs (for example CommonNavigator) do expose a viewer instance to clients, including 3rd-party clients. If this client moves to "incremental display" because it's best for it, they would hope and even expect that clients are not broken by usage of this new flag, and may just enable incremental displayt because it's nice and easy to do so; and miss to think about the issue that consumers need to consider the ExpandableNode in the selection, and actually make those 3rd party consumers prone to ClassCastExceptions and so on.
So direct clients of the new API do it on their own risk, but when they start exposing the viewer, it's the clients of this clients that can crash the viewer because of the change in the invariant.

> And if someone enables it and finds new issue, we would be happy to fix it (as with any other code change).

This change does intend to fix the issue before someone finds it in a release ;)

> all new API is marked as experimental anyway. 

Right, but we should still try to make it as robust and stable as possible before releasing it.

> So if there will be a major refactoring later we should be fine changing this experimental API. Just consider this as a Java "preview" feature that might change. Everyone is warned - if you turn it on, you might need to adopt your code in the future. 

I've seen "experimental" APIs that got adopted too quickly by too many people so they got practically set stone after 1st release, and providers were unable to do the fix they had planned for the future. For Java they have a strict policy, enabling preview features is a burden that makes it very clear it can change. In Eclipse Platform, people may bind to experimental APIs without even noticing it, it's too easy to over-use an API, the "preview" approach of Platform is to weak to be reliable.
Hence why I want to push for highest quality ASAP, before the release.

I will investigate the "Link with editor" thing today.

mickaelistria:
So it seems like "Link with editor" doesn't work anyway as per https://github.com/eclipse-platform/eclipse.platform.ui/issues/1014 ?

iloveeclipse:
> So it seems like "Link with editor" doesn't work anyway as per #1014 ?

It works partly - it selects the node that "hides" expected element. So user sees what need to be expanded.

mickaelistria:
The reason why tweaks are needed in JDT seems to be another isuse with implementation in JFace: `getRawChildren()` for the ExpandableNode should *not* return children as the invariant in JFace is that the raw children are actual children, that are to be displayed as children.
Here, the API has been bent to serve a different purpose and it leads it to returning incorrect values that require adaptation in downstream consumers.

mickaelistria:
Although I would like to spend more time on this topic, I have some other priorities and am pessimistic about the possibility of improving this code substantially before the release. I'm giving up.

raghucssit:
@mickaelistria I will checkout this and fix what's remaining.

akurtakov:
Is this one still needed?

iloveeclipse:
> Is this one still needed?

It is obsoleted meanwhile. All points are fixed by @raghucssit in other PR's .

PR #1065, opened 2023-08-24
Opening a directory shows existing projects in Project Explorer

yuzawa-san:
The existing behavior is to always run the import wizard, but this is not necessary if the directory being opened is the root of an existing project. In that case, the project will be shown in the Project Explorer. If the project is open, it will be refreshed. If the project is closed, it will be opened. Additionally, I tried to make the code flow a little easier to follow and with less 'arrow'. There are decent amount of whitespace changes. If there is an easier way to do this please advise, but I tried to lean on the existing view/selection methods and commands.

Rationale:

- I was looking at https://github.com/desktop/desktop/issues/16991 and per [their requirements](https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md) the opening of directories should be properly supported (i.e. you open a github repo and can choose "open repo in eclipse"). However, the current behavior is not useful for already imported projects since the wizard says it's already imported. This change will make the open directory call more useful.

Screen recording:

https://github.com/eclipse-platform/eclipse.platform.ui/assets/1082334/c8db553a-cf03-49a8-a8d0-905eb14c6283

Scenarios in video (the mac `open` command line tool is used):

- open a file in a project
- open an existing project which is open is refreshed
- open an existing project which is closed is opened
- open another directory (opens import wizard)

### DIFF
--- a/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
+++ b/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2021 IBM Corporation and others.
+ * Copyright (c) 2010, 2021, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -14,6 +14,8 @@
  *                       at a specific line/col
  *     Nitin Dahyabhai - Bug 567708 - Support for MultiPageEditorParts
  *                       containing a TextEditor page
+ *     James Yuzawa    - 'Open Directory' action shows existing projects in
+ *                       Project Explorer
  ******************************************************************************/
 
 package org.eclipse.ui.internal.ide.application;
@@ -25,11 +27,18 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 
+import org.eclipse.core.commands.Command;
+import org.eclipse.core.commands.ExecutionEvent;
+import org.eclipse.core.commands.common.CommandException;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.jface.dialogs.MessageDialog;
+import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.jface.wizard.WizardDialog;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
@@ -38,14 +47,20 @@ import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.IViewPart;
+import org.eclipse.ui.IWorkbenchCommandConstants;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.WorkbenchException;
+import org.eclipse.ui.commands.ICommandService;
 import org.eclipse.ui.ide.IDE;
 import org.eclipse.ui.internal.ide.IDEWorkbenchMessages;
 import org.eclipse.ui.internal.ide.IDEWorkbenchPlugin;
 import org.eclipse.ui.internal.wizards.datatransfer.SmartImportWizard;
+import org.eclipse.ui.navigator.resources.ProjectExplorer;
+import org.eclipse.ui.part.ISetSelectionTarget;
 import org.eclipse.urischeme.IUriSchemeProcessor;
 import org.osgi.framework.Bundle;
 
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
```diff
@@ -183,90 +198,119 @@ public void run() {
...
+				if (shell != null) {
+					if (shell.getMinimized())
+						shell.setMinimized(false);
+					shell.forceActive();
+				}
+				try {
+					if (details.fileInfo.isDirectory()) {
+						// Open a directory
+						File directory = new File(details.fileStore.toURI());
+						for (IProject project : ResourcesPlugin.getWorkspace().getRoot().getProjects()) {
```
BeckerWdf, line 226: May it be that some of the methods in `ResourcesPlugin.getWorkspace().getRoot().getProjects()` return null?

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
```diff
@@ -183,90 +198,119 @@ public void run() {
...
-								Bundle textEditorBundle = Platform.getBundle(TEXTEDITOR_BUNDLE_NAME);
-								if (textEditorBundle != null) {
-									Class<?> textEditorClass = textEditorBundle.loadClass(TEXTEDITOR_CLASS_NAME);
-									if (textEditorClass != null) {
-										Object textEditor = invoke(openEditor, "getAdapter", //$NON-NLS-1$
-												new Class[] { Class.class }, new Object[] { textEditorClass });
-										if (textEditor != null) {
-											openEditor = (IEditorPart) textEditor;
-										}
+								IViewPart view = page.showView(ProjectExplorer.VIEW_ID);
```
BeckerWdf, line 234: So this is specific to the project explorer view. But what about other views like "Package Explorer" or other "Navigator-Views"?

This will open the project explorer view if it is not yet open, correct? Is this really what a user wants?

mickaelistria (approved):
Thank you. This patch is interesting and implementation looks good.
Eclipse Platform is currently in stabilization/RC phase for one more week and we typically avoid feature addition in that phase. We will try to merge it soon after we can start feature development for next/4.30 release.

BeckerWdf:
It would be nice to have some kind of automated tests for this as well.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
iloveeclipse: no

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
iloveeclipse: I think this is OK, considering that "old" Navigator is not there, Package Explorer may not be installed (CDT etc), Git repo Explorer is Git specific.

BeckerWdf: But do we really need this. What would happen if we don't show and refresh that view. Wouldn't the editor do that refresh implicitly when the file is shown in the editor?

iloveeclipse: I would expect it would be surprising if someone wants to see a directory (from command line) but IDE would "silently" do that in background and no visual feedback is given. But may be I misunderstood the request here. If I understand it right, one passes directory path to IDE and IDE is supposed to show appropriate resource in a view. For that I would consider showing the (potentially hidden / closed) view as "must have" part of the enhancement.

BeckerWdf: > I would expect it would be surprising if someone wants to see a directory (from command line) but IDE would "silently" do that in background and no visual feedback is given. But may be I misunderstood the request here. If I understand it right, one passes directory path to IDE and IDE is supposed to show appropriate resource in a view. For that I would consider showing the (potentially hidden / closed) view as "must have" part of the enhancement.

You are right. For a directory this makes sense - as no editor opens. But what if we open a file that is inside an project in the workspace? I think in that case we should not "touch" the view but leave it to the editor to "refresh" the view.

iloveeclipse: OK, that's correct, view should only be opened for directories.

iloveeclipse: But we are inside `if (details.fileInfo.isDirectory()` so that is already given

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
```diff
@@ -183,90 +198,119 @@ public void run() {
...
+										IWorkbenchCommandConstants.PROJECT_OPEN_PROJECT);
+									if (command != null && command.isHandled() && command.isEnabled()) {
+										command.executeWithChecks(new ExecutionEvent());
 									}
 								}
+								return;
+							}
+						}
+						// Fall back to import
+						SmartImportWizard wizard = new SmartImportWizard();
```
BeckerWdf, line 249: Starting the smart import wizard may be refactored out into a private method to avoid code duplication

BeckerWdf (dismissed):

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/DelayedEventsProcessor.java
yuzawa-san: Can you please elaborate on the deduplication? I'm not seeing the SmartImportWizard used anywhere else in this file. It seems to be used in OpenFolderCommand and ConfigureProjectHandler but slightly differently.

BeckerWdf: > Can you please elaborate on the deduplication? I'm not seeing the SmartImportWizard used anywhere else in this file. It seems to be used in OpenFolderCommand and ConfigureProjectHandler but slightly differently.

Sorry. I now looked again. There's no code duplication. I mis-read this.

mickaelistria:
Build log shows that a version bump is necessary since we just release 4.29
```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.2:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.ui.ide.application: Only qualifier changed for (org.eclipse.ui.ide.application/1.5.100.v20230911-0711). Expected to have bigger x.y.z than what is available in baseline (1.5.100.v20230821-1342) -> [Help 1]
```
Can you please bump bundle version by +0.0.100 ?

yuzawa-san:
@mickaelistria I think I bumped it correctly and it seems to work locally.

@BeckerWdf Do you have any advice on how to test this? That entire class seems quite hard to test. Would a test of the innermost logic suffice? I could break the crux of my change into an easier to test static method.

mickaelistria:
It has been measured that adding tests after completion of the feature is usually not profitable (iirc it is discussed in *The Economics of Software Quality*), so while addition of tests is always welcome as a good thing, I wouldn't make it a requirement for a merge here.

Issue #735, opened 2023-04-27
New setting: show icons in editor area


albertmink:
During development the number of opened views might increase significantly, see the views at the bottom. Here the titles cannot be shown properly and in can get worse easily.
<img width="1344" alt="Screenshot 2023-04-27 at 11 58 53" src="https://user-images.githubusercontent.com/38354196/234829046-5c2f7368-f028-4f9e-aef9-df370c511da9.png">

This issue propose to add a functionality that allows to hide icons. This provides more space for the titles see mock-up
<img width="1474" alt="lots_of_views_no_icon" src="https://user-images.githubusercontent.com/38354196/234828797-892feb49-3b83-4209-9aa1-43992625588e.png">

UX perspective:
Without icons (in the bottom section) the editor gains focus, which could be favourable over the additional icon (that help of simple identify views).

jukzi:
Feel free to provide a PR that optionally removes icons. But by default i want to see icons and the typical way to prevent that many views is to switch between perspectives.

albertmink:
We propose to add two (one shown above already) options to allow further visual customising in the page 'Preferences' -> 'Appearance'.
To start with, adding a new section with the checkboxes (default: not checked/ticked ➡️  won't change the recent behaviour)

Tab icons and texts:
- [ ] Show icons only in editor area
- [ ] Show full text for all tabs except in editor area

~~Screenshot for checked boxes will be added below~~ Get the preference page up and running as the first step.

jukzi:
Sounds reasonable. Do you really have problems with many tabs in the java-IDE or rather problems on a product based on it?

albertmink:
From my daily work 
<img width="1159" alt="Screenshot 2023-06-15 at 12 40 04" src="https://github.com/eclipse-platform/eclipse.platform.ui/assets/38354196/ebf3d567-5075-4435-a50f-7ce0bb549559">

jukzi:
"ABAP" -> @sratz can you look at it, please?

sratz:
> "ABAP" -> @sratz can you look at it, please?

Our team is currently working on this and we will provide a pull-request soon.

vogella:
To many options should be avoided, why the distinction of icons in the editor area? Is this really necessary?

Phillipus:
Removing the icon won't make much difference if the view title is quite long.

> Too many options should be avoided,

Agreed

albertmink:
[...]

PR #1071, opened 2023-08-29
Add Options to hide tab icons and and show full text in view area

praveen-skp:
Issue: #735 

Currently, the UI/UX for Eclipse views does not scale properly when a perspective contains a lot of views. The current solution for scaling views in Eclipse is to provide lots of different perspectives and use them to group the views. If only a limited number of views are visible it means that the whole title of each view is visible and the usability is good but if the number views increase significantly then the following issues occur:

1) The tab titles get so small that only the view icons are visible. This means users have to learn all icons which is not ideal.
2) The overall look&feel of the IDE feels cluttered.
3) Lots of view icons draw attention away from the editor area. However, when playing the role as a classic IDE for a programming language the editor area should be most prominent.

Changes:
- As all three points are based on personal preferences, new eclipse preferences has been added for handling Tab icons and titles in View areas by providing two options 
a. Always show full titles 
b. Hide icons
Tab Icons and text preferences

- Corresponding change to CTabRendering to handle these preferences and render the image and text for tabs in view areas accordingly
    - As a pre-requisite, added property "showSelectedImage" to CTabFolder : [Pull request](https://github.com/eclipse-platform/eclipse.platform.swt/pull/785)  

- Added two CSS properties handler
  - CSSPropertyTabTextMinimumCharactersSWTHandler: An swt property to set the minimum characters for CTabFolder
  - CSSPropertySelectedImageVisibleSWTHandler: An swt property to set if the selected tab's image should be visible or not

This a prototype and would like to get feedback


albertmink:
@praveen-skp can you provide a screenshot from the whole IDE, not only a view area (with hidden icons)? Such an example helps to explain the addressed change in _overall look&feel_

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/custom/CSSPropertySelectedImageVisibleSWTHandler.java
```diff
@@ -0,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2016 IBM Corporation and others.
```
BeckerWdf: Adapt copyright header to 2023 and "SAP SE..."

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/custom/CSSPropertyTabTextMinimumCharactersSWTHandler.java
```diff
@@ -0,0 +1,45 @@
+package org.eclipse.e4.ui.css.swt.properties.custom;
```
BeckerWdf, line 13: add copyright header

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -54,7 +59,36 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
...
 	 */
 	public static final boolean USE_ROUND_TABS_DEFAULT = false;
+	/**
+	 * A named preference for setting CTabFolder's to be rendered without icons
+	 * <p>
+	 * The default value for this preference is: <code>false</code> (render
+	 * CTabFolder's with icons)
+	 * </p>
+	 */
+	public static final String HIDE_ICONS = "HIDE_ICONS"; //$NON-NLS-1$
```
BeckerWdf: I would say "HIDE_ICONS_FOR_VIEW_TABS" better describes this.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -54,7 +59,36 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
...
+	 */
+	public static final boolean HIDE_ICONS_DEFAULT = false;
+	/**
+	 * A named preference for setting CTabFolder's to show full text
+	 * <p>
+	 * The default value for this preference is: <code>false</code> (render
+	 * CTabFolder's without full text)
+	 * </p>
+	 */
+	public static final String SHOW_FULL_TEXT = "SHOW_FULL_TEXT"; //$NON-NLS-1$
```
BeckerWdf: "SHOW_FULL_TEXT_FOR_VIEW_TABS"

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -139,6 +173,8 @@ public CTabRendering(CTabFolder parent) {
 		parent.addDisposeListener(e -> preferences.removePreferenceChangeListener(this));
 
 		cornerRadiusPreferenceChanged();
+		showFullText();
```
BeckerWdf: should be "showFullTextPreferenceChanged"

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -139,6 +173,8 @@ public CTabRendering(CTabFolder parent) {
...
+		hideIcons();
```
BeckerWdf: should be "hideIconsPreferenceChanged"

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -1264,13 +1300,61 @@ private void cornerRadiusPreferenceChanged() {
...
 	@Override
 	public void preferenceChange(PreferenceChangeEvent event) {
-		if (!USE_ROUND_TABS.equals(event.getKey())) {
+		if (event.getKey().equals(USE_ROUND_TABS)) {
+			cornerRadiusPreferenceChanged();
+		} else if (event.getKey().equals(HIDE_ICONS)) {
+			hideIcons();
+		} else if (event.getKey().equals(SHOW_FULL_TEXT)) {
+			showFullText();
+		} else {
```
BeckerWdf: else return can be omitted

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -180,6 +190,50 @@ private void createThemeIndependentComposits(Composite comp) {
...
+		boolean parentState = parent.getSelection();
+		dependent.setEnabled(parentState);
+
+		SelectionListener listener = new SelectionListener() {
+			@Override
+			public void widgetSelected(SelectionEvent e) {
+				boolean state = parent.getSelection();
+				dependent.setEnabled(state);
+
+				if (state == false)
```
BeckerWdf: Formatting seems not to be according to the standards. Didn't you format it? Shouldn't it be 
if (!state) {
...

BeckerWdf (changes requested):

praveen-skp:
Screenshots for the whole IDE:
Default
Show full title
Show full title and hide icons

iloveeclipse:
Few things:
- if no icons shown for views, it would be nice to show **active** view with an icon
- Instead of hiding (small) icons and showing only (long) view titles, why not add an option to show no view titles. In this case, views will be shown by icons and only active one will have icon and title. **That** will save place and clearly set focus on what is active in each view stack.

shubhamWaghmare-sap:
@iloveeclipse 
> Few things:
> 
> * if no icons shown for views, it would be nice to show **active** view with an icon

This is already technically possible using the SWT API. We can consider incorporating this as a future update if required.

On a side note, we also observed that showing the icon (only) for **active** view could potentially increase screen flickering, as the text for the tabs would be moved each time a tab is switched. Thus the impact on user experience should be evaluated before proceeding.

> * Instead of hiding (small) icons and showing only (long) view titles, why not add an option to show no view titles. In this case, views will be shown by icons and only active one will have icon and title. **That** will save place and clearly set focus on what is active in each view stack.

Showing only icons is also technically possible and here is what it would look like:

image

We had contemplated this option and consulted with UX experts who indicated the following potential issues related to accessibility:

- It would be difficult for users to identify a tab just by the icon.
- Users would need to learn/remember the icons for different views, which is not ideal.
- Figuring out if a view is already open or not would not be a quick task.
- Precise mouse movements would be needed to hit the required tab.

Moreover, the intention behind the change was not on saving space but to keep the focus only on editor area which is the _main actor_ in IDE. Thus the icons for all tabs from the view stack (_supporting actors_) are to be hidden, which might draw attention away from the editor area otherwise. 
Also the idea is to align the look & feel with other IDEs out there like VS code, where tabs in view stack always show full texts but no icons, which gives a cleaner look.

BeckerWdf:
Regarding "only show icons": It looks much more cluttered compared to the current state (Icon and Text). I have the feeling this is a move into the wrong direction.

humphreygao:
I think it is a good idea, you can provide this 2 options(show full titles and hide icons) and let users to decide. And it would be good to always show editor tabs' full name especially when you open  many editors.

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/custom/CSSPropertySelectedImageVisibleSWTHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import org.w3c.dom.css.CSSValue;
+
+public class CSSPropertySelectedImageVisibleSWTHandler extends AbstractCSSPropertySWTHandler {
+
+	@Override
+	public void applyCSSProperty(Control control, String property, CSSValue value, String pseudo, CSSEngine engine)
+			throws Exception {
+		boolean isSelectedImage = (Boolean) engine.convert(value, Boolean.class, null);
+		if (control instanceof CTabFolder) {
+			CTabFolder folder = (CTabFolder) control;
```
HeikoKlare:
```suggestion
		if (control instanceof CTabFolder folder) {
```

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/custom/CSSPropertySelectedImageVisibleSWTHandler.java
```diff
@@ -0,0 +1,42 @@
...
+			folder.setSelectedImageVisible(isSelectedImage);
+		}
+	}
+
+	@Override
+	public String retrieveCSSProperty(Control control, String property, String pseudo, CSSEngine engine)
+			throws Exception {
+		if (control instanceof CTabFolder) {
+			CTabFolder folder = (CTabFolder) control;
```
HeikoKlare:
```suggestion
		if (control instanceof CTabFolder folder) {
```

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/custom/CSSPropertySelectedImageVisibleSWTHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import org.eclipse.swt.custom.CTabFolder;
+import org.eclipse.swt.widgets.Control;
+import org.w3c.dom.css.CSSValue;
+
+public class CSSPropertySelectedImageVisibleSWTHandler extends AbstractCSSPropertySWTHandler {
+
+	@Override
+	public void applyCSSProperty(Control control, String property, CSSValue value, String pseudo, CSSEngine engine)
+			throws Exception {
+		boolean isSelectedImage = (Boolean) engine.convert(value, Boolean.class, null);
```
HeikoKlare: I do not understand the name of the variable. Shouldn't it be something like `isImageVisibleForSelectedTab` or the like?

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/properties/custom/CSSPropertyTabTextMinimumCharactersSWTHandler.java
```diff
@@ -0,0 +1,57 @@
...
+			CTabFolder folder = (CTabFolder) control;
+			folder.setMinimumCharacters(minimumCharacters);
+		}
+	}
+
+	@Override
+	protected String retrieveCSSProperty(Control control, String property, String pseudo, CSSEngine engine)
+			throws Exception {
+		if (control instanceof CTabFolder) {
+			CTabFolder folder = (CTabFolder) control;
```
HeikoKlare:
```suggestion
		if (control instanceof CTabFolder folder) {
```

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -1264,13 +1301,57 @@ private void cornerRadiusPreferenceChanged() {
...
 	private IEclipsePreferences getSwtRendererPreferences() {
 		return InstanceScope.INSTANCE.getNode("org.eclipse.e4.ui.workbench.renderers.swt"); //$NON-NLS-1$
 	}
+
+	private boolean isPartOfEditorStack() {
+		MUIElement element = (MUIElement) parent.getData(AbstractPartRenderer.OWNING_ME);
+		if (element != null && element.getTags().contains(EditorTag)) {
+			return true;
+		}
+		return false;
```
HeikoKlare:
```suggestion
		return element != null && element.getTags().contains(EditorTag);
```

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -1264,13 +1301,57 @@ private void cornerRadiusPreferenceChanged() {
...
+		} else if (event.getKey().equals(SHOW_FULL_TEXT_FOR_VIEW_TABS)) {
+			showFullTextForViewTabsPreferenceChanged();
+		}
+	}
+
+	private void showFullTextForViewTabsPreferenceChanged() {
+		boolean showFullText = getShowFullTextForViewTabsPreference();
+		if (!isPartOfEditorStack()) {
+			if (showFullText) {
+				Optional<Integer> max = Arrays.stream(parent.getItems()).map(CTabItem::getText).map(String::length)
```
HeikoKlare: Please rename the variable to something self-explaining like `lengthOfLongestItemText`

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -1264,13 +1301,57 @@ private void cornerRadiusPreferenceChanged() {
...
+						.max(Integer::compare);
+				parent.setMinimumCharacters(max.orElseGet(() -> 9999));
```
HeikoKlare: Please assign this magic value to an explaining variable.

HeikoKlare (commented):
Thanks for the proposal. I really like it.

I have some minor code style comments and one general remark:
I am wondering whether it may be useful to allow the configuration for each part stack (not necessarily as part of this PR but as a future extension). You could then define the visualization that fits most for one specific stack of views. E.g., you may have one stack of views that you usually distinguish by their (full) names, so that you want to hide the icons and always show the full texts, and you may have a (horizontally) small stack of views of which you know well their icons, so that only showing the icons may be perfect for that stack (this would require to have the option of only showing icons as proposed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1071#issuecomment-1717766464). This would relieve you from deciding about the visualization of tab items for the whole workbench. As a concrete example, in a "usual" perspective setup such as the Java perspective, displaying the full names of the tabs may be useful for the views in the stack at the bottom of the workbench window, while it may not be desired for the stack at the right side of the workbench window, as the names of one or two views can easily require all the available space for the tabs.
As an RCP application provider, you could even provide perspectives that are predefined accordingly. The preferences could then be used to define the initial behavior of newly created stacks.

BeckerWdf:
> I think it is a good idea, you can provide this 2 options(show full titles and hide icons) and let users to decide. And it would be good to always show editor tabs' full name especially when you open many editors.

I would try to keep the number of configurations options as small as possible. Let's start with the current implementation of this PR. Let's see that what kind of feedback we get. We can add more configuration options later on. But pls. let's start simple.

HeikoKlare:
One further thing I forgot to mention: when removing the icons, it becomes really hard to see where one tab item ends and the next begins, as there are no borders between them. With the icons enabled, you directly see where the items starts. So maybe there should be some kind of small separator to see where a new tab item starts?

BeckerWdf:
> One further thing I forgot to mention: when removing the icons, it becomes really hard to see where one tab item ends and the next begins, as there are no borders between them. With the icons enabled, you directly see where the items starts. So maybe there should be some kind of small separator to see where a new tab item starts?

You are right. But is it necessary to see where "inactive" view tabs start and end exactly. If you want to bring it to the top you simple click on the text. Pls. look how VsCode does it - I also don't see separators there...

HeikoKlare:
> You are right. But is it necessary to see where "inactive" view tabs start and end exactly. If you want to bring it to the top you simple click on the text. Pls. look how VsCode does it - I also don't see separators there...

That's true. My concerns are not with respect to where the tab starts and ends exactly but with respect to identifying the text that belongs together, i.e., to a single tab, at all. The difference I see in comparison to VS Code is the size of the space between the texts. Maybe we could increase the padding to have a bigger difference between the size of a whitespace and the size of the space between tabs, so to make it easier to find the individual tabs?

For example, in the following tab folder, I find it quite hard to see where tabs start and end. It takes some time (or knowledge) to identify whether there are tabs "CSS" and "Scratch Pad" or whether it is a single "CSS Scratch Pad". Either I know that, I have to take a deep look for quite some time to distinguish the space between tabs and the whitespace between words, or I have to hover with the mouse to see the highlighting.

image

In VS Code, I find it fairly easy to identify the individual tabs.
image

BeckerWdf:
> Maybe we could increase the padding to have a bigger difference between the size of a whitespace and the size of the space between tabs, so to make it easier to find the individual tabs?

Good idea.

shubhamWaghmare-sap:
> Maybe we could increase the padding to have a bigger difference between the size of a whitespace and the size of the space between tabs, so to make it easier to find the individual tabs?

Earlier the padding was only added to the selected tab. 
Now the padding is uniformly added across all tabs (selected and unselected) where tab icon would not be visible.
This is how it looks now and the change is done in https://github.com/eclipse-platform/eclipse.platform.swt/pull/785
image

For example, notice the significant space between _Type Hierarchy_ tab and _CSS Scratch Pad_ tab.

HeikoKlare (commented):
With some further adaptations of the required SWT change https://github.com/eclipse-platform/eclipse.platform.swt/pull/785, this looks good to me.

I would be in favor of even larger paddings when no icons are shown, as also mentioned in the SWT PR (https://github.com/eclipse-platform/eclipse.platform.swt/pull/785#pullrequestreview-1753347506), which could look like this:
image
But that's a matter of personal preference.

BeckerWdf (approved):

HeikoKlare:
@shubhamWaghmare-sap Was it intended that the minimum number of characters shown for tabs now is 1 in case "always show full titles" is not enabled? That way it seems to become impossible to navigate between source code files 😇 
image

thomasritter:
Just tested the iBuild. In general it looks much cleaner. We should think about making the option "Always show full titles" the default, in the future. When the icons are visible everything looks great. Once I switch off the icons the tab title spacing is not ideal when the tab is in focus, in my opinion. When I look at the titles it feels like they are not bound to the beginning of the tab. They are leaning to the right. The reason for this is that the titles are not aligned in the center. I would recommend we align them. The left red box should be flush with the beginning of the tab. See screenshot:

alignment_tab_titles_no_icons

shubhamWaghmare-sap:
> @shubhamWaghmare-sap Was it intended that the minimum number of characters shown for tabs now is 1 in case "always show full titles" is not enabled? That way it seems to become impossible to navigate between source code files 😇 image

@HeikoKlare It seems that before this change, the minimum number of characters for tabs _"in view areas"_ was always 1. Also the newly added preferences do not affect the _"editor area"_.
The below screenshot corresponds to a commit before this change, where, we have a chevron in _"editor area"_ and only 1 character visible in _"view area"_. (same is the case when "always show full titles" is not enabled )

image

So,
1. Do you observe this issue in the _Editor area_ or in _View area_?
2. If you observe it in _View area_, is it expected to have a higher number for default minimum characters and have chevron instead ?
3. Is there a reason to suspect it to be a regression?

HeikoKlare:
Maybe I erroneously related the changed behavior to this PR. For me, the behavior changed in the _editor area_. This is how it looks like:
image

Behavior for the view area is fine for me, as it has been like this before, and you can usually also distinguish views by icons (in comparison to files in the editor area, where the single first letter does not help).

The changed behavior for the editor area (assuming that it is related to this PR) actually is a regression for me, because for me it is now ususable without "always show full titles" activated.

> Once I switch off the icons the tab title spacing is not ideal when the tab is in focus, in my opinion. When I look at the titles it feels like they are not bound to the beginning of the tab. They are leaning to the right. The reason for this is that the titles are not aligned in the center.

The SWT implementation for tab folders with disabled icons centers the text (like, e.g., VS code does). @thomasritter What do you mean by "leaning to the right"? The text actually should be centered. And making it left aligned will look strange, as then you have a huge space to the right as a separator to the next tab.

thomasritter:
@HeikoKlare If the tab is selected the area in which the text should be aligned runs from the left edge to the right X. See this screenshot. I added the black box. You can see that the text is not centered. It is too much to the right.

alignment_tab_titles_no_icons

HeikoKlare:
Thanks for clarification! I got it. The problem is that centering is performed w.r.t. the "text area", but the tab also has a padding on its own. We have to fix that in SWT.

thomasritter:
Ok, here a fake screenshot in which I fixed it manually and centered the text. This is how I would expect it to look like.

alignment_tab_titles_centered

HeikoKlare:
Seeing the fake screnshot: do you really think that it looks better that way? It seems like my eyes expect the text to be centered "within the padding" of the tab, like in the current behavior. So the current behavior actually looks more "natural" to me than the proposed one.

thomasritter:
@HeikoKlare In order to get away from "feeling" and more towards empirical data I looked at the most widely used tab design and measured how they position the content (text + icon). Have a look at this diagram:

image

As you can see, almost all tab titles are leaning slightly to the left side of the tab. Only Intellij uses a centered design. As most humans read from left to right this makes a lot of sense to me. The content needs to feel anchored to the left side. This is not the case with the current design. My proposal would be to follow the padding of the Eclipse tab with icon or at least do not have more padding in the front than after the (x) button.

By the way thanks for sending me down this rabbit hole :) Was interesting to realize that the browser vendors are using very similar spacing for their tabs.

HeikoKlare:
Thanks for collecting that data, @thomasritter!

You compare the iconless tab design with other applications using icons in tabs. The reason why left alignment and small paddings work when using icons is that the icons work as a separator between individual tabs, which allows you to easily find the start/end of a tab.

When using an iconless design, you need to apply larger paddings to be able to distinguish tab start/end from whitespaces between words in a single tab (see my [previous comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1071#issuecomment-1735717715)). From my point of view, using paddings as separators only works with centered text. Left-aligning the tab text looks really weird:
image

To support this with some empirical data: VS code does it the same way (in the screenshot, the "Terminal" tab is moved between the "Output" and "Debug Console" tabs to show where VS code puts the border):
image

Issue #1070, opened 2023-08-28
Code Mining: Redraw of end-of-line annotation is skipped


mx990:
When redrawing an end-of-line annotation in `AbstractInlinedAnnotation.redraw()`, the argument for `GC gc` is always `null`:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/81e7679279080525779750c0415907d7326015b6/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/AbstractInlinedAnnotation.java#L136

This ultimately causes `InlinedAnnotationDrawingStrategy.drawAfterLine(...)` to return without doing anything:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/81e7679279080525779750c0415907d7326015b6/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationDrawingStrategy.java#L213-L216

The relevant call stack is: 
```
InlinedAnnotationDrawingStrategy.drawAfterLine(LineContentAnnotation, GC, StyledText, int, int, Color) line: 215
InlinedAnnotationDrawingStrategy.draw(LineContentAnnotation, GC, StyledText, int, int, Color) line: 205
InlinedAnnotationDrawingStrategy.draw(AbstractInlinedAnnotation, GC, StyledText, int, int, Color) line: 136
CodeMiningLineContentAnnotation(AbstractInlinedAnnotation).lambda$0(StyledText) line: 136
[...]
```

mickaelistria:
Would you be interested in submitting a PR to fix it?

mx990:
The issue initially occurred when using debug code minings in https://eclipse.dev/4diac/. When modifying a variable though the _Variables_ view, the values shown in the debug code minings were not updated. It was necessary to click into the editor (or otherwise cause a redraw) to see the current values. I subsequently narrowed it down to the problem described above.

I have added a minimal (non-)working example based on the existing `org.eclipse.jface.text.examples.codemining.CodeMiningDemo` in a separate PR https://github.com/eclipse-platform/eclipse.platform.ui/pull/1076.

Steps to reproduce (using https://github.com/eclipse-platform/eclipse.platform.ui/pull/1076):
1. Run the modified `org.eclipse.jface.text.examples.codemining.CodeMiningDemo` from `org.eclipse.jface.text.examples`
2. Modify the label for the end-of-line code minings in the text field at the top

Current behavior (master):
[...]

PR #1076, opened 2023-09-01
Code Mining: Example for redraw of end-of-line annotation #1070

mx990:
This change modifies the code mining example to show the redraw of end-of-line annotations. A text field is added at the top to change the label of end-of-line code minings through user input.


mickaelistria (approved):
Will merge it when 4.29 release is complete and 4.30 is open for business.

PR #1077, opened 2023-09-01
IStatusLineManager.getProgressMonitor(): fix multiple invocations

jukzi:
ignore call to done() if beginTask() was not called!

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/61

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLineManager.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLineManager.java
@@ -149,15 +149,29 @@ public class StatusLineManager extends ContributionManager implements
 				: new NullProgressMonitor();
 
 		return new IProgressMonitor() {
+			private volatile boolean taskStarted;
 
 			@Override
 			public void beginTask(String name, int totalWork) {
+				if (taskStarted) {
+					throw new IllegalStateException("beginTask must only be called once per instance"); //$NON-NLS-1$
+				}
+				taskStarted = true;
+				// According to the IProgressMonitor javadoc beginTask() must only be called
+				// once on a given progress monitor instance.
+				// However it works in this case multiple times if done() was called in between.
 				progressDelegate.beginTask(name, totalWork);
-
 			}
 
 			@Override
 			public void done() {
+				if (!taskStarted) {
+					// ignore call to done() if beginTask() was not called!
+					// Otherwise an otherwise already started delegate would be finished
+					// see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/61
+					return;
+				}
+				taskStarted = false;
 				progressDelegate.done();
 			}
 
@@ -182,6 +196,9 @@ public class StatusLineManager extends ContributionManager implements
 
 			@Override
 			public void setTaskName(String name) {
+				if (!taskStarted) {
+					throw new IllegalStateException("call to beginTask() missing"); //$NON-NLS-1$
+				}
 				progressDelegate.setTaskName(name);
 
 			}
@@ -194,6 +211,9 @@ public class StatusLineManager extends ContributionManager implements
 
 			@Override
 			public void worked(int work) {
+				if (!taskStarted) {
+					throw new IllegalStateException("call to beginTask() missing"); //$NON-NLS-1$
+				}
 				progressDelegate.worked(work);
 			}
 
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/action/IStatusLineManager.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/action/IStatusLineManager.java
@@ -31,7 +31,7 @@ import org.eclipse.swt.graphics.Image;
  */
 public interface IStatusLineManager extends IContributionManager {
 	/**
-	 * Returns a progress monitor which reports progress in the status line.
+	 * Creates a new progress monitor which reports progress in the status line.
 	 * Note that the returned progress monitor may only be accessed from the UI
 	 * thread.
 	 *
@@ -39,6 +39,7 @@ public interface IStatusLineManager extends IContributionManager {
 	 *
 	 *         Note: There is a delay after a beginTask message before the
 	 *         monitor is shown. This may not be appropriate for all apps.
+	 * @see IProgressMonitor
 	 */
 	public IProgressMonitor getProgressMonitor();
 


iloveeclipse:
This caused lot of test fails, because Java editor can't be created anymore.

https://download.eclipse.org/eclipse/downloads/drops4/I20230911-1800/testResults.php

iloveeclipse:
```
Caused by: java.lang.IllegalStateException: beginTask must only be called once per instance
	at org.eclipse.jface.action.StatusLineManager$1.beginTask(StatusLineManager.java:157)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:483)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:428)
	at org.eclipse.jdt.internal.core.JavaModelOperation.run(JavaModelOperation.java:731)
	at org.eclipse.jdt.internal.core.JavaModelOperation.runOperation(JavaModelOperation.java:806)
	at org.eclipse.jdt.internal.core.CompilationUnit.becomeWorkingCopy(CompilationUnit.java:98)
	at org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.createFileInfo(CompilationUnitDocumentProvider.java:1015)
	at org.eclipse.ui.editors.text.TextFileDocumentProvider.connect(TextFileDocumentProvider.java:481)
	at org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.connect(CompilationUnitDocumentProvider.java:1260)
	at org.eclipse.ui.texteditor.AbstractTextEditor.doSetInput(AbstractTextEditor.java:4177)
	at org.eclipse.ui.texteditor.StatusTextEditor.doSetInput(StatusTextEditor.java:262)
	at org.eclipse.ui.texteditor.AbstractDecoratedTextEditor.doSetInput(AbstractDecoratedTextEditor.java:1474)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaEditor.internalDoSetInput(JavaEditor.java:2556)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaEditor.doSetInput(JavaEditor.java:2529)
	at org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitEditor.doSetInput(CompilationUnitEditor.java:1415)
	at org.eclipse.ui.texteditor.AbstractTextEditor.lambda$1(AbstractTextEditor.java:3167)
	at org.eclipse.jface.operation.ModalContext.runInCurrentThread(ModalContext.java:434)
	at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:352)
	at org.eclipse.ui.internal.WorkbenchWindow.lambda$7(WorkbenchWindow.java:2346)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:67)
	at org.eclipse.ui.internal.WorkbenchWindow.run(WorkbenchWindow.java:2344)
	at org.eclipse.ui.texteditor.AbstractTextEditor.internalInit(AbstractTextEditor.java:3184)
```

bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLineManager.java
```diff
@@ -194,6 +211,9 @@ public void subTask(String name) {
 
 			@Override
 			public void worked(int work) {
+				if (!taskStarted) {
```
laeubi, line 214: I'm not sure if this is covered by the API that one *must* have started the monitor before reporting work (even though it seems useful), maybe just ignore that call then? or simply store the work amount and apply it once the monitor is started...

bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLineManager.java
```diff
@@ -149,15 +149,29 @@ public IProgressMonitor getProgressMonitor() {
 				: new NullProgressMonitor();
 
 		return new IProgressMonitor() {
+			private volatile boolean taskStarted;
 
 			@Override
 			public void beginTask(String name, int totalWork) {
+				if (taskStarted) {
+					throw new IllegalStateException("beginTask must only be called once per instance"); //$NON-NLS-1$
```
laeubi, line 157: I think for this to become useful one should record the callstack of the first invocation to give a hint where this has originated.

The easiest way would be to construct a new Runtime exception e.g. with message "first invocation here" and use that as the cause of the exception

bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLineManager.java
```diff
@@ -149,15 +149,29 @@ public IProgressMonitor getProgressMonitor() {
 				: new NullProgressMonitor();
 
 		return new IProgressMonitor() {
+			private volatile boolean taskStarted;
```
laeubi, line 152: This is not sufficient for thread safety, so either it should be an atomicboolean using the compare and set operation below or a simple boolean, marking a value as volatile "just in case" does not help much and gives a false impression.

jukzi:
> This caused lot of test fails, because Java editor can't be created anymore.

At the first glance i think this is an error in the call site to reuse org.eclipse.ui.editors.text.TextFileDocumentProvider.getProgressMonitor().
I will look into it.

bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLineManager.java
jukzi: IProgressMonitor:25
```
When a task begins, a <code>beginTask(String, int)
 * </code> notification is reported, followed by any number and mixture of 
 * progress reports (<code>worked()</code>)
```

bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLineManager.java
jukzi: IProgressMonitor itself is not designed to be threadsafe. look for example at the SubMonitor implementation which is inherently unsafe

iloveeclipse:
> At the first glance i think this is an error in the call site to reuse

I would propose to revert all places where `throw new IllegalStateException` was added to the monitor implementation.
While clients may make mistakes operating the monitor, the API contract of IProgressMonitor doesn't say anything about throwing unchecked exceptions if some operations aren't executed in expected order. 

Changing very old and common API behavior like in this PR I would consider as a breaking change.

The most problematic part of it is that the breaking change can't be found by compilation, static code analysis and even tests - because depending on the timing the problematic monitor code might be not used at all in tests, and will explode in the production.

jukzi:
> I would propose to revert all places where `throw new IllegalStateException` was added to the monitor implementation. While clients may make mistakes operating the monitor, the API contract of IProgressMonitor doesn't say anything about throwing unchecked exceptions if some operations aren't executed in expected order.
> 
> Changing very old and common API behavior like in this PR I would consider as a breaking change.
> 
> The most problematic part of it is that the breaking change can't be found by compilation, static code analysis and even tests - because depending on the timing the problematic monitor code might be not used at all in tests, and will explode in the production.

You can change it to log the error for example in M3, but first we should fix the indicated errors. - which seems to be pretty simple  - see commits.

iloveeclipse:
> You can change it to log the error for example in M3

This is too late. We use nightly I builds in validation tests and can't afford wait for M3.

Adding code that potentially breaks application just to see where the behavior is broken is surely possible on an internal project, but not on the SDK level.

jukzi:
> This is too late. We use nightly I builds in validation tests and can't afford wait for M3.

you could take the chance to fix your repos too if needed.

iloveeclipse:
> you could take the chance to fix your repos too if needed. 

Jörg, please see 

> The most problematic part of it is that the breaking change can't be found by compilation, static code analysis and even tests - because depending on the timing the problematic monitor code might be not used at all in tests, and will explode in the production.

jukzi:
> will explode in the production.

i doubt you use I-Build in production.

iloveeclipse:
Jörg, could you please simply remove Exceptions? You can't assume anything about how SDK is used by consumers, except that if it can fail it will fail.

jukzi:
i can assume that you don't need I Build in production.
BTW it's very easy to find if you have any code calling org.eclipse.ui.editors.text.TextFileDocumentProvider.getProgressMonitor(). -and make sure it calls done as it should.
I have no concerns to remove exceptions as soon as we have verified the Eclipse IDE does not trigger it.

iloveeclipse:
> BTW it's very easy to find if you have any code calling org.eclipse.ui.editors.text.TextFileDocumentProvider.getProgressMonitor(). -and make sure it calls done as it should

Don't get me wrong, but monitor can be passed around and from here on it isn't easy anymore. Also it is not related to TextFileDocumentProvider only. 

> I have no concerns to remove exceptions as soon as we have verified the Eclipse IDE does not trigger it

Not sure how you plan to verify it, but please provide a PR.

jukzi:
> Not sure how you plan to verify it, but please provide a PR.

?? review the suggested changes, please, submit and then we hopefully see tomorrow ibuild passes.

iloveeclipse:
> ?? review the suggested changes, please, submit and then we hopefully see tomorrow ibuild passes.

The build passes (but as explained above, it cannot validate error free state after this patch, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/1107).

Please provide a PR to convert new exceptions to the log entries.

PR #1078, opened 2023-09-01
ViewerCell: eager NPE, final members #177

jukzi:
Do not reuse ViewerCell instance - that only caused trouble and was clearly forbidden by
org.eclipse.jface.viewers.ViewerColumn.refresh(ViewerCell). Instead let the GC do his job.

Declaring the constants final formally requires API change. But it was never intended that the constants could be changed.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/177


HeikoKlare (commented):
Changes look good to me. It makes sense to set the `ViewerCell`'s column immutable and create new cell instances instead of reusing and changing the same one (both from a design perspective as well as considering the number bugs related to the existing functionality).

I guess that the original functionality was implemented for performance reasons? But I do not see why the proposed change (with additional object instantiation and cleanup) should have a noticable impact on performance.

@jukzi There is one test failing because of the proposed change `Bug203657TableViewerTest.testBug203657`. I would propose to remove the test (class) as the change made it obsolete (and the test was quite bad anyway 🙂).

jukzi:
> @jukzi There is one test failing because of the proposed change `Bug203657TableViewerTest.testBug203657`. I would propose to remove the test (class) as the change made it obsolete (and the test was quite bad anyway 🙂).

good idea! thanks

PR #1080, opened 2023-09-04
bug 569286: remove msg to post stacktrace

jukzi:
we already got traces, but nobody cares


HeikoKlare (commented):
Should we represent the referenced bug as a GitHub issue if it is (still) relevant? I think it will get more attention here than in the "old" bug tracker, in particular since there is some current discussion on the topic. Then the exception message could also be changed to reference the GitHub issue.

jukzi:
> Should we represent the referenced bug as a GitHub issue if it is (still) relevant? I

Don' add reference to github unless somebody promisses to dig deep into the error. Otherwise it's like dev/null

HeikoKlare:
True. I just wanted to make the point that it's less likely that someone will dig into the issue if they do not find it (due to only having an eye on GitHub issues and not on the old bug tracker).

HeikoKlare (approved):

Issue #1084, opened 2023-09-06
Startup/shutdown performance: unnecessary processing of activity events


HeikoKlare:
## Issue
The activity management in `MutableActivityManager` and `ActivityPersistenceHelper` is inefficient when dealing with higher amounts of activities. Activity enablements are unnecessarily processed and persisted upon different events:
1. All activities enablements are persisted upon every activity enablement change in the preference store, even if only the enablement of an expression-controlled activity changes (which is not part of the persisted states):
https://github.com/eclipse-platform/eclipse.platform.ui/blob/f4989c93f6e599d977cabf7b039836a771a3b035/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/ActivityPersistanceHelper.java#L60-L62
2. Enablement changes of expression-controlled activities are processed (and persisted as stated in 1.) even if the workbench is already closing. This has no effect as the activity support is disposed anyway during workbench close. Since the activity management is disposed _after_ the service locator, which produces further events due to the evaluation service being removed, it receives and processes these additional events unnecessarily.

## Reproduction

This unnecessary processing probably has no noticable effect in many products, as the types and amounts of activities are not high enough to produce a perceivable performance impact. In larger products this can, however, have severe impacts. In our product the effect is an overhead during startup and shutdown of the workbench of multiple seconds.

I have created a _very_ artificial demo project with a large amount of defined activities and a large amount of enablement changes during startup/shutdown to make the impact obvious:

[ManyActivitiesPerformanceDemo.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/12539744/ManyActivitiesPerformanceDemo.zip)

## Evaluation

I have executed that demo with the current state of implementation and compared it to improved implementations that address the mentioned issue (can be seen in PR #1085 and #1086).

On my machine, the results with the demo project across several runs were as follows:

Startup time:
Without unnecessary persistence: 350-500 ms
With unnecessary persistence: 1500-1700 ms

Shutdown time:
With early disposal: 350-450 ms
Without proper disposal: 1300-1450 ms

## How to proceed

I have started two PRs that address the mentioned issues (#1085, #1086). The changes of those PRs are the basis for the evaluation results mentioned above. In case you already see any side effects of addressing those issue that I am currently not aware of or, in general, reasons why the current behavior is required, I would be glad if you can let me know.

vogella:
[...]

PR #1085, opened 2023-09-06
Avoid unnecessary persistence of activity enablements #1084

HeikoKlare:
The `ActivityPersistenceHelper` persists all activity enablements upon every notification about a changed activity enablement it receives. Persistence is, however, only done and thus necessary for activities that are not expression-controlled. So the current implementation performs unnecessary persistence operations.

With this change, the ActivityPersistenceHelper only persists the enablements when the enablement of a non-expression-controlled activity has changed. To this end, the received ActivityManagerEvent is extended by the information about whether such an activity was affected. This information is either directly passed to the event construction in case it has already been calculated in the calling context or is calculated lazily when accessing the value.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1084.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/activities/ActivityManagerEvent.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/activities/ActivityManagerEvent.java
@@ -14,7 +14,10 @@
 
 package org.eclipse.ui.activities;
 
+import static java.util.Collections.emptySet;
+
 import java.util.Set;
+import java.util.stream.Stream;
 import org.eclipse.ui.internal.util.Util;
 
 /**
@@ -40,22 +43,30 @@ public final class ActivityManagerEvent {
 	private boolean enabledActivityIdsChanged;
 
 	/**
+	 * Indicates whether enabled IDs of non-expression-controlled activities have
+	 * changed. The value is calculated from changed enabled activity IDs given to
+	 * the constructor if known by the caller or is set to <code>null</null> and
+	 * will be calculated on demand upon first access to the value.
+	 */
+	private Boolean enabledNonExpressionControlledActivityIdsChanged = null;
+
+	/**
 	 * The set of activity identifiers (strings) that were defined before the change
-	 * occurred. If the defined activities did not changed, then this value is
+	 * occurred. If the defined activities did not change, then this value is
 	 * <code>null</code>.
 	 */
 	private final Set<String> previouslyDefinedActivityIds;
 
 	/**
 	 * The set of category identifiers (strings) that were defined before the change
-	 * occurred. If the defined category did not changed, then this value is
+	 * occurred. If the defined category did not change, then this value is
 	 * <code>null</code>.
 	 */
 	private final Set<String> previouslyDefinedCategoryIds;
 
 	/**
 	 * The set of activity identifiers (strings) that were enabled before the change
-	 * occurred. If the enabled activities did not changed, then this value is
+	 * occurred. If the enabled activities did not change, then this value is
 	 * <code>null</code>.
 	 */
 	private final Set<String> previouslyEnabledActivityIds;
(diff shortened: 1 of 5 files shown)


HeikoKlare:
Failing test is documented: #926

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/activities/ActivityManagerEvent.java
```diff
@@ -214,4 +302,34 @@ public boolean haveDefinedCategoryIdsChanged() {
...
+
+	private static boolean haveEnabledNonExpressionControlledActivityIdsChanged(IActivityManager activityManager,
+			Set<String> previousIds) {
+		Set<String> currentIds = activityManager.getEnabledActivityIds();
+		Set<String> unchangedIds = new HashSet<>(currentIds);
+		unchangedIds.retainAll(previousIds);
+		Set<String> changedIds = new HashSet<>(previousIds);
+		changedIds.addAll(currentIds);
+		changedIds.removeAll(unchangedIds);
+		return changedIds.stream().anyMatch(id -> activityManager.getActivity(id).getExpression() == null);
```
fedejeanne: You could avoid creating new `Set`s if you do it like this:

```java
Set<String> currentIds = activityManager.getEnabledActivityIds();

Stream.concat(currentIds.stream(), previousIds.stream()) // the sum...
		.filter(id -> !currentIds.contains(id) || !previousIds.contains(id))// ...minus the intersection
		.anyMatch(id -> activityManager.getActivity(id).getExpression() == null);
```

You could even experiment concatenating 2 `.parallelStream()`s and see if it improves/degrades performance.

HeikoKlare: Thanks! Good point. It should be even more efficient to treat the streams individually, as then only one `contains` checks for every element is necessary:
```java
	Set<String> currentIds = activityManager.getEnabledActivityIds();
	Stream<String> addedIds = currentIds.stream().filter(id -> !previousIds.contains(id));
	if (addedIds.anyMatch(this::isIdOfNonExpressionControlledActivity)) {
		return true;
	}
	Stream<String> removedIds = previousIds.stream().filter(id -> !currentIds.contains(id));
	return removedIds.anyMatch(this::isIdOfNonExpressionControlledActivity);

...

private boolean isIdOfNonExpressionControlledActivity(String id) {
	return activityManager.getActivity(id).getExpression() == null;
}
```
Actually, this method only exists for compatibility reasons. The actual goal is that the field `enabledNonExpressionControlledActivityIdsChanged` is set via the constructor call as the calculation has implicitly been done before. Thus, in the demo scenario changes to this method have no performance impact. Even if changing the code to explicitly performing the calculation everytime the value is accessed has no measurable performance impact, as the performance issue addressed by this PR does not come from this calculation but from the code that is executed when not performing this check at all.

I have updated the functionality in b8c6a1e accordingly.

akurtakov:
Looks like this one broke javadoc generation https://download.eclipse.org/eclipse/downloads/drops4/I20230919-1800/compilelogs/platform.doc.isv.javadoc.txt

laeubi:
Also reported here (search for `javadoc:jar`):

https://github.com/eclipse-platform/eclipse.platform.ui/runs/16931217956

HeikoKlare:
Sorry for that. I'll provide a fix in a moment.

Could we have those checks integrated into mandatory GH Actions checks to have them represented as actual errors of the PR? Or is there a (non-technical) reason for not dong so?

HeikoKlare:
> Also reported here (search for `javadoc:jar`):
> 
> https://github.com/eclipse-platform/eclipse.platform.ui/runs/16931217956

I think the ones reported there are not related to this PR but pre-existing issues in Javadoc generation, as they appear in every Jenkins log. In particular, the warnings also occur in the build for the PR with the fixes for the broken Javadoc in this PR: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1142/checks?check_run_id=16954953767

akurtakov:
The I-build javadoc generation cares about public API only while the GH action looks at all the sources.  We never managed to get that to a clean state to even consider making it fail the build. It's an extra issue (not investigated yet AFAICT) that some javadoc errors are not catched in the Javadoc report but are only in the Maven one which is full with other issues and thus even less people notice these. 
Long story short - it's a a web of issues so I kindly ask everyone to just try to not let warnings stay around when noticed, if enough people do that at some point we should be able to raise certain things to errors and make them more obvious. Help on the releng side to get better configs, improving tools and etc. is also highly appreciated.

HeikoKlare:
Thanks for clarification, @akurtakov! Good to know about that situation and have it in mind when looking at PR check results and logs.

Issue #1005, opened 2023-08-11
Random failing VirtualTableViewerTest.testRenameWithSorter and VirtualTableViewerTest.testContains 


HeikoKlare:
The test `org.eclipse.jface.tests.viewers.VirtualTableViewerTest.testRenameWithSorter` randomly fails, e.g. https://download.eclipse.org/eclipse/downloads/drops4/I20230810-1800/testresults/html/org.eclipse.jface.tests_ep429I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

```
sorted first expected:<0-[0 name-999]9> but was:<0-[9 name-]9>

junit.framework.ComparisonFailure: sorted first expected:<0-[0 name-999]9> but was:<0-[9 name-]9>
at junit.framework.Assert.assertEquals(Assert.java:100)
at junit.framework.TestCase.assertEquals(TestCase.java:253)
at org.eclipse.jface.tests.viewers.VirtualTableViewerTest.testRenameWithSorter(VirtualTableViewerTest.java:244)
```

HeikoKlare:
The test `org.eclipse.jface.tests.viewers.VirtualTableViewerTest.testContains` also randomly fails, e.g., 
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1086/checks?check_run_id=16671995346 
```
element must be available on the viewer
junit.framework.AssertionFailedError: element must be available on the viewer
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.jface.tests.viewers.TableViewerTest.testContains(TableViewerTest.java:200)
```

vogella:
Seems to fails all the time on Mac. @sratz could you have a look if this can be fixed or if it should be disabled on Mac?

BeckerWdf:
also reported in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1283

PR #1086, opened 2023-09-06
Dispose activity management during shutdown completely and early #1084

### LINKED ISSUE #14: Disable Event Processing During Workbench Shutdown
### Current Behavior

The `enabledWhenListener` of the `MutableActivityManager` processes events, also during workbench shutdown. Depending on the number of events, this can lead to high delays during shutdown (in our products up to several seconds).

https://github.com/eclipse-platform/eclipse.platform.ui/blob/1bd79e4cc55ce3f287dcc2151ad96e19e07a19a0/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/activities/MutableActivityManager.java#L597-L613

The highest time-consuming operation is the update call to the `ActivityPersistenceHelper`, which persists the state of all activity enablements upon changing a single one. In the shutdown scenario where many activities are disabled, for each disablement the whole number of activities is processed and their enablements are stored in the preference store. This is unnecessary, as the shutdown performs a save operation afterwards anyway.

### Expected Behavior

This event processing during shutdown should either be
* deactivated during shutdown if it is not necessary,
* deactivatable through some hook to speedup shutdown (works for us), or
* improved in its performance

### Proposed Solution

Disable `ActivityPersistenceHelper::saveEnabledStates()` during workbench shutdown, as it will be performed when shutting down the persistence helper anyway.

### Additional Information

Problem analysis has shown that not only shutdown is affected by unnecessary event processing but also startup. The issue has been documented together with a reproduction example in eclipse-platform/eclipse.platform.ui#1084.

The proposed solution is split into one affecting startup and one affecting shutdown:
eclipse-platform/eclipse.platform.ui#1085
eclipse-platform/eclipse.platform.ui#1086

HeikoKlare:
The `WorkbenchActivitySupport` is currently disabled quite late during workbench shutdown. In particular, it is performed after the service locator disposal, which produces several events for the disablement of expression-controlled activities that are processed by the activity management. Since the activity management is shut down afterwards anyway, the processing of these events is unnecessary.

This change ensures that activities enablement changes are not unnecessarily processed during workbench shutdown. It consists of two parts:
1. It performs the disposal of activity management and its persistence handler early in the workbench shutdown process (in particular before the service locator disposal).
2. It adds proper dispose functionality for the IActivityManager, such that disposing the workbench activity support disposes the activity manager, which then removes all its listeners to not react to further events.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1084.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -3019,17 +3019,26 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 		workbenchListeners.clear();
 
 		cancelEarlyStartup();
-		if (workbenchService != null)
+		if (workbenchService != null) {
 			workbenchService.unregister();
+		}
 		workbenchService = null;
 
-		if (e4WorkbenchService != null)
+		if (e4WorkbenchService != null) {
 			e4WorkbenchService.unregister();
+		}
 		e4WorkbenchService = null;
 
 		// for dynamic UI
 		registry.removeRegistryChangeListener(extensionEventHandler);
 		registry.removeRegistryChangeListener(startupRegistryListener);
+
+		// shut down activity helper before disposing workbench activity support;
+		// dispose activity support before disposing service locator to avoid
+		// unnecessary activity disablement processing
+		activityHelper.shutdown();
+		workbenchActivitySupport.dispose();
+		WorkbenchHelpSystem.disposeIfNecessary();
 
 		// Bring down all of the services.
 		serviceLocator.dispose();
@@ -3038,11 +3047,7 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 		getDisplay().removeFilter(SWT.MouseDown, backForwardListener);
 		backForwardListener = null;
 
-		workbenchActivitySupport.dispose();
-		WorkbenchHelpSystem.disposeIfNecessary();
-
 		// shutdown the rest of the workbench
-		activityHelper.shutdown();
 		uninitializeImages();
 		if (WorkbenchPlugin.getDefault() != null) {
 			WorkbenchPlugin.getDefault().reset();
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/AbstractActivityManager.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/AbstractActivityManager.java
@@ -22,6 +22,8 @@ import org.eclipse.ui.activities.IActivityManagerListener;
 public abstract class AbstractActivityManager implements IActivityManager {
 	private ListenerList<IActivityManagerListener> activityManagerListeners;
 
+	private boolean disposed;
+
 	protected AbstractActivityManager() {
 	}
 
@@ -60,4 +62,15 @@ public abstract class AbstractActivityManager implements IActivityManager {
 			activityManagerListeners.remove(activityManagerListener);
 		}
 	}
+
+	@Override
+	public void dispose() {
+		activityManagerListeners.clear();
+		disposed = true;
+	}
+
+	protected boolean isDisposed() {
+		return disposed;
+	}
+
 }
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/MutableActivityManager.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/MutableActivityManager.java
@@ -134,6 +134,23 @@ public final class MutableActivityManager extends AbstractActivityManager
 	}
 
 	@Override
+	public void dispose() {
+		super.dispose();
+		clearExpressions();
+		activityRegistry.removeActivityRegistryListener(activityRegistryListener);
+		activitiesById.clear();
+		activityRequirementBindingsByActivityId.clear();
+		activityPatternBindingsByActivityId.clear();
+		categoriesById.clear();
+		categoryActivityBindingsByCategoryId.clear();
+		categoryDefinitionsById.clear();
+		definedActivityIds.clear();
+		definedCategoryIds.clear();
+		enabledActivityIds.clear();
+		identifiersById.clear();
+	}
+
+	@Override
 	synchronized public IActivity getActivity(String activityId) {
 		if (activityId == null) {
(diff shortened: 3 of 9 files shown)


HeikoKlare:
Failing tests are documented: #926, #1005

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/AbstractActivityManager.java
```diff
@@ -60,4 +62,16 @@ public void removeActivityManagerListener(IActivityManagerListener activityManag
 			activityManagerListeners.remove(activityManagerListener);
 		}
 	}
+
+	@Override
+	public void dispose() {
+		this.activityManagerListeners.clear();
+		this.activityManagerListeners = null;
+		this.isDisposed = true;
```
fedejeanne: NIT: no need to use `this`

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/AbstractActivityManager.java
```diff
@@ -22,6 +22,8 @@
 public abstract class AbstractActivityManager implements IActivityManager {
 	private ListenerList<IActivityManagerListener> activityManagerListeners;
 
+	private boolean isDisposed = false;
```
fedejeanne: IMO `disposed` would be better, otherwise it looks like a method.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -3031,18 +3031,21 @@ private void shutdown() {
 		registry.removeRegistryChangeListener(extensionEventHandler);
 		registry.removeRegistryChangeListener(startupRegistryListener);
 
+		// shutdown activity helper before disposing workbench activity support;
+		// dispose activity support before disposing service locator to avoid
+		// unnecessary activity disablement processing
```
fedejeanne: Would it be possible to test this? If I'm reading it correctly, this is the cornerstone of the PR, right? I mean the order is relevant.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/MutableActivityManager.java
```diff
@@ -133,6 +133,23 @@ public MutableActivityManager(ITriggerPointAdvisor triggerPointAdvisor, IActivit
 		readRegistry(true);
 	}
 
+	@Override
+	public void dispose() {
+		super.dispose();
+		clearExpressions();
+		this.activityRegistry.removeActivityRegistryListener(activityRegistryListener);
+		this.activitiesById = null;
```
sratz: The API doc of `dispose` says `The behavior of all other methods after disposing is undefined.`, however, in most cases this will now lead to a NPE.

How likely are such scenario?

And, if not entirely unlikely, should we instead be more lenient here by keeping the fields and just clearing them?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/AbstractActivityManager.java
iloveeclipse: Seeing this code, can we please don't init it to default value?
`private boolean disposed;` should be enough

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/AbstractActivityManager.java
HeikoKlare: all unnecessary usages of `this` fixed with  ada4186

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/AbstractActivityManager.java
HeikoKlare: agree with both comments, fixed with ada4186

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/activities/MutableActivityManager.java
HeikoKlare: Thanks for pointing that out. From my point of view, scenarios where an activity manager is (unintendedly) used after disposing it are unlikely. The API is internal and we know about its current usages, which is basically only during shutdown at which no further access to it is performed.

I see two lines of argumentation: 
1. You could argue that throwing NPEs in case of inappropriate API usage is good because you get fast feedback that you are doing unintended things. 
2. You could argue that the risk of NPEs is unnecessary as there is no risk of memory leaks and no performance benfit when nulling the values instead of clearing them. And the occurence of NPEs on inappropriate API usage may not occur during development but may be restricted to specific usage scenarios because of a race condition.

I think that both argumentations are valid but I tend to follow your proposal of being lenient (so going with argument 2). I adapted the PR with ada4186.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
HeikoKlare: Thanks @fedejeanne, that's a good idea. I would actually not like to test the execution order, as it should be an implementation detail of the method. Instead, I would propose to test the effect that I want to achieve or avoid, respectively. In this case, This PR aims at avoiding unnecessary activity manager events being fired during workbench shutdown, So I've added a test that verifies this behavior in 26964bb. The test fails when executed on the code without the changes in this PR, so it's actually a regression test for the according changes.

fedejeanne: Good idea! And you're right: the less one tests the implementation details, the more stable the tests are.

Thanks for adding the regression test.

vogella:
@HeikoKlare can this be merged?

HeikoKlare:
Yes, it's ready to merge. I've only been waiting for the end of the M1 freeze period.

vogella:
> Yes, it's ready to merge. I've only been waiting for the end of the M1 freeze period.

Master is open now, so please merge after the verification build.

HeikoKlare:
Failing tests are unrelated and documented in #294 and #1183.

PR #1088, opened 2023-09-08
Only persist then view part name when a custom name has been set

### LINKED ISSUE #763: Problems view is not translated when changing IDE language
When changing the Eclipse to e.g. German, the title of the Problems view is not correctly translated. Weirdly enough, the correct name is shown in the drop-down menu and other views such as the Progress are correctly translated.
 
image

ptziegler:
The user can set a custom name when opening a new view. This value has
to be persisted when closing the Eclipse, otherwise the view would fall
back to its default name. We must only persist the part name when a
custom name has been set, in order to prevent saving a translated name.

Resolves #763

### DIFF
--- a/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.ide; singleton:=true
-Bundle-Version: 3.21.100.qualifier
+Bundle-Version: 3.21.200.qualifier
 Bundle-Activator: org.eclipse.ui.internal.ide.IDEWorkbenchPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %Plugin.providerName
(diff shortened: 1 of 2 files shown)


mickaelistria:
This seems like very custom code for a very common case... How do other views handle it? Why is the view name persisted in the memento for that view?

ptziegler:
>  How do other views handle it?

The e.g. Progress view implements the ViewPart directly, so they never run into this problem.

> Why is the view name persisted in the memento for that view?

After a little bit of digging, I was able to trace this behavior back to [Bug 203767](https://bugs.eclipse.org/bugs/show_bug.cgi?id=203767) - [Markers] multiple instances of the new view a bit confusing to work with.

And I think my change would break this behavior... I sounds like you're somehow able to specify a custom name when opening a new view and this is what's then written to the memento. Except that the name is _always_ written to the memento, which sounds like the real problem. The proper solution would then be to _only_ persist the part name, when a custom name has been set. Would you agree?

ptziegler:
This is the handler responsible for the custom name:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/f4989c93f6e599d977cabf7b039836a771a3b035/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/OpenMarkersViewHandler.java#L60-L63

ptziegler:
As a remark: The name of the view is only updated when it is initialized. When the Eclipse is restarted and the view is not focused, it will keep its old name, because then the EPart label is used. But that's a completely different beast...

In any case, the custom part name is stored in a local variable and used instead of the default name, if initialized.

An additional remark:
This change won't "fix" older views. But for such a minor issue, I think that's acceptable. The user simply has to close and re-open the view and then the problem should sort itself out. Reason being that we can't retroactively determine, whether the label is a custom label or the default label, just with a different locale.

mickaelistria:
Thanks!

PR #1089, opened 2023-09-11
Add method for optionally returning the first element of a given type

laeubi:
IStructuredSelection#getFirstElement is widely used but because selections are not generic returns a plain Object resulting in different boilerplate patterns across the code including optimistic casting, instance of and similar. In the case for an empty selection null is returned adding some risk for NPE if not used appropriate and even though there is an isEmpty() method Optionals usually offer a much more convenient way to handle the absence of a value.

This adds a new method to access the first element with an Optional describing the result and specify a desired type of element fixing both the risk of null-pointer-access, remove the boilerplate for additional cast/instanceof checks as well as allowing easy to understand chaining and default value handling.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
@@ -15,6 +15,10 @@ package org.eclipse.jface.viewers;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * A selection containing elements.
@@ -22,12 +26,71 @@ import java.util.List;
 @SuppressWarnings("rawtypes")
 public interface IStructuredSelection extends ISelection, Iterable {
 	/**
-	 * Returns the first element in this selection, or <code>null</code>
-	 * if the selection is empty.
+	 * Returns the first element in this selection, or <code>null</code> if the
+	 * selection is empty.
 	 *
 	 * @return an element, or <code>null</code> if none
 	 */
-	public Object getFirstElement();
+	default Object getFirstElement() {
+		return stream().findFirst().orElse(null);
+	}
+
+	/**
+	 * Returns an optional describing the first element of the requested type or an
+	 * empty optional if the selection is empty or does not contain any compatible
+	 * element in regard to the specified type
+	 *
+	 * @param type the expected type
+	 * @param <T>  the desired type
+	 * @return an optional describing the result
+	 * @since 3.32
+	 */
+	default <T> Optional<T> getFirstElementOf(Class<T> type) {
+		return streamOf(type).findFirst();
+	}
+
+	/**
+	 * Returns an optional describing the single selected element of the requested
+	 * type or an empty optional if the selection is empty or does not contain
+	 * exactly one compatible element in regard to the specified type
+	 *
+	 * @param type the expected type
+	 * @param <T>  the desired type
+	 * @return an optional describing the result
+	 * @since 3.32
+	 */
+	default <T> Optional<T> getSingleElementOf(Class<T> type) {
+		if (size() == 1) {
+			return Optional.ofNullable(getFirstElement()).filter(type::isInstance).map(type::cast);
+		}
+		return Optional.empty();
+	}
+
+	/**
+	 * Returns the elements in this selection as a stream.
+	 *
+	 * @return a (possibly empty) stream of the elements contained in this selection
+	 * @since 3.32
+	 */
+	@SuppressWarnings("unchecked")
+	default Stream<Object> stream() {
+		return StreamSupport.stream(spliterator(), false);
+	}
+
+	/**
+	 * Return all the elements in this selection as a stream that are of the given
+	 * type.
+	 *
+	 * @param type the desired type of elements
+	 * @param <T>  the element type
+	 * @return a stream of compatible elements
+	 * @since 3.32
+	 */
+	@SuppressWarnings("unchecked")
+	default <T> Stream<T> streamOf(Class<T> type) {
+		Objects.requireNonNull(type);
+		return StreamSupport.stream(spliterator(), false).filter(type::isInstance).map(type::cast);
+	}
 
 	/**
 	 * Returns an iterator over the elements of this selection.
@@ -35,7 +98,9 @@ public interface IStructuredSelection extends ISelection, Iterable {
 	 * @return an iterator over the selected elements
 	 */
 	@Override
-	public Iterator iterator();
+	default Iterator iterator() {
+		return stream().iterator();
+	}
 
 	/**
 	 * Returns the number of elements selected in this selection.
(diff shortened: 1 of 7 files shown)


mickaelistria:
Can you please include as part of your PR some consumer in Platform that would take advantage of this new APIs?

laeubi:
> Can you please include as part of your PR some consumer in Platform that would take advantage of this new APIs?

Done, I used one in JFace bundle directly here.

mickaelistria:
I fully support this addition, but we first need API tools to cope with it.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
```diff
@@ -15,20 +15,40 @@
...
+	 * Returns an optional describing the first element of the requested type or an
+	 * empty optional if the selection is empty or does not contain any compatible
+	 * element in regard to the specified type
+	 *
+	 * @param type the expected type
+	 * @param <T>  the desired type
+	 * @return an optional describing the result
+	 * @since 3.32
+	 */
+	default <T> Optional<T> getFirstElement(Class<T> type) {
```
sratz: `first` has different semantics in the new method than in `Object getFirstElement()`

What about

`default Stream<T> getElements(Class<T> type)`

instead, returning a stream of *all* elements in the selection of given type.

It's easy to extend this with `findFirst().orElse(null)` in consumer code, but also would allow use-cases where more than one element is wanted.

laeubi: `Stream` is a separate thing for me and last time I tried this a year ago or so it resulted in endless discussions so I'd better like to start with simpler things now.

Also it *might* be more performant for some implementations to only fetch a few / items than prepare a full stream.

Yes the method has a slightly more relaxed semantics here but I assume this is never an issue anyways.

Of course if there later is a stream() method the default impl could be changed accordingly :-)

mickaelistria: I read `getFirstElement` as the implementation tells, otherwise, I would expect the method to be named `getFirstElementAs(Class<T>)`. Indeed most of the current code can be substituted by some `getFirstElementAs(Class<T>)` without changing the behavior, it would be safer to keep the behavior for consumers.
@laeubi Are you aware of many consumers getSelection actually iterate over the whole selection filtering for a particular type?

About going to streams, we can also do something more precise or powerful technically, but here the proposal is meant to replace the very common pattern we see everywhere. If the new API is noticeably more complex than the legacy pattern, then the API proposal is less beneficial as it will be less used...

laeubi:
@sratz @mickaelistria if there is support from you I can also adjust this to include Streams, I just wanted to avoid that this getting (gain) abandon because of to many competing things...

what I wanted to have for strutucred selection as well is 

- stream()
- streamOf()
- singleElement()

as these are the most annoying things currently when using it where I write the same code over and over again...

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
laeubi: @sratz I renamed the method now to `getFirstElementOf(...)` to make the difference more clear.

mickaelistria:
To be honest, I don't remember the reason what we don't already have `IStructuredSelection.stream()` already. I remember it was proposed and I fully support the idea in principle.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
laeubi: Most consumer simply want *an element* in contrast to really *the first element*, also most viewers only contain the same type of elements (even though elements might have different subclasses), I just wanted to avoid making it more complex than it actually is in practice ... I hope the rename makes it now more clear and easy to understand.

mickaelistria: >  I hope the rename makes it now more clear and easy to understand.

The issue with this method as opposed to `getFirstElementAs` is that consumers cannot just replace safely: if they're in a strange state where they care about the 1st element only (which is the one implemented in the vast majority of consumers), and the 1st is not of the right type but the 2nd is, then adopting this method will cause a change in the behavior.
So if this is not an equivalent replacement, it's much harder to adopt it safely.

laeubi: I think it is safe in almost all cases, because if you look at where it is used it either is checked that there is only one element or the caller is just to lazy to check more elements I can't even think about a case where I want the user to only select a specific first element, I just want the user to select *at least one* that I can process....

mickaelistria: > I think it is safe in almost all cases,

While I agree, there is a major different between "we think it is safe in almost all cases" and "we're certain it's safe in all cases" ;)

laeubi:
I added the stream methods as well...

laeubi:
I now decided to better start with the smallest possible change of adding a stream() method here:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1099

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
```diff
@@ -15,27 +15,92 @@
...
+	 * Returns an optional describing the single selected element of the requested
+	 * type or an empty optional if the selection is empty or does not contain
+	 * exactly one compatible element in regard to the specified type
+	 *
+	 * @param type the expected type
+	 * @param <T>  the desired type
+	 * @return an optional describing the result
+	 * @since 3.32
+	 */
+	default <T> Optional<T> getSingleElementOf(Class<T> type) {
```
HannesWell, line 62: What about `getOnlyElementAs(Class)`?

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
```diff
@@ -15,27 +15,92 @@
...
+	 * Returns an optional describing the first element of the requested type or an
+	 * empty optional if the selection is empty or does not contain any compatible
+	 * element in regard to the specified type
+	 *
+	 * @param type the expected type
+	 * @param <T>  the desired type
+	 * @return an optional describing the result
+	 * @since 3.32
+	 */
+	default <T> Optional<T> getFirstElementOf(Class<T> type) {
```
HannesWell, line 48: Just `getFirst(Class)` would probably make it more natural to read:
`selection.getFirst(String.class)`

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
```diff
@@ -15,27 +15,92 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * A selection containing elements.
  */
 @SuppressWarnings("rawtypes")
 public interface IStructuredSelection extends ISelection, Iterable {
```
HannesWell, line 27: I have the impression that with all the default implementations we now have an endless loop (`iterator()` uses `stream()` and `stream()` the spliterator(), which uses `iterator()`). I think it would be good to enforce at least one implementation.
`iterator()` is probably always possible in a simple way.

laeubi: stream should be the method to actually implement, everything else could be derived from it... but I would leave out that for now anyways and split this into smaller chunks.

PR #1054, opened 2023-08-21
Split org.eclipse.search into core and ui bundles

robstryker:
This PR aims to split org.eclipse.search into core and UI bundles. 
Package names are preserved in all cases (though I appreciate a review to verify). 
The existing bundle maintains its name. A new bundle is added with .core. 
The new core bundle is re-exported to consumers so the result should be seamless to them (I believe, but I appreciate smarter people verifying). 

The main sticking point in this refactor was the TextSearchVisitor requiring access to the UI to discover dirty editors. Separating out the UI logic and finding a way to make it available to core was the bulk of the work. 

While some API-level solutions could in theory work (ie, when org.eclipse.search starts, set a dirtyEditorParticipant on the core plugin directly), the solution I've chosen was an internal extension point.  If this is not acceptable, we could always switch back to an API-only solution without the need of an extension point. I'm willing to do whatever the community thinks is best. 

The build passes locally and a smoke-test on my side has verified that search functions as expected both for files that match on-disk contents, as well as dirty editors.

PR #1098, opened 2023-09-11
DO NOT PUSH, TEST

robstryker:
This PR is to see whether a clean build with minimal changes expresses the same api analysis errors visible in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1054


mickaelistria:
Looks like things are all right now and we don't need that test any more.

PR #1099, opened 2023-09-12
Add stream support to IStructuredSelection

laeubi:
Streams offer much more flexibility as they can be filtered, mapped and finally collected to any useful representation of the data.

This adds a new `stream() `method to `IStructuredSelection` so consumers of selections can benefit from this without any additional casting or boilerplate code (example of code simplification included in this PR), implement it as a default method for backward compatibility of implementors and adjust the implementation to return a slightly more efficient variant.

This is a minimal change extracted from
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1089

and hopefully less controversial to go in soon.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,11 +10,14 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - add support for stream() method
  *******************************************************************************/
 package org.eclipse.jface.viewers;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * A selection containing elements.
@@ -57,4 +60,15 @@ public interface IStructuredSelection extends ISelection, Iterable {
 	 * @return the selected elements as a list
 	 */
 	public List toList();
+
+	/**
+	 * Returns the elements in this selection as a <code>Stream</code>.
+	 *
+	 * @return the selected elements as a stream
+	 * @since 3.32
+	 */
+	@SuppressWarnings("unchecked")
+	default Stream<Object> stream() {
+		return StreamSupport.stream(spliterator(), false);
+	}
 }
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StructuredSelection.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StructuredSelection.java
@@ -18,6 +18,7 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Stream;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.jface.resource.JFaceResources;
@@ -207,6 +208,14 @@ public class StructuredSelection implements IStructuredSelection {
 		return Arrays.asList(elements == null ? new Object[0] : elements);
 	}
 
+	@Override
+	public Stream<Object> stream() {
+		if (isEmpty()) {
+			return Stream.empty();
+		}
+		return Arrays.stream(elements);
+	}
+
 	/**
 	 * Internal method which returns a string representation of this
 	 * selection suitable for debug purposes only.
@@ -215,6 +224,7 @@ public class StructuredSelection implements IStructuredSelection {
 	 */
 	@Override
 	public String toString() {
-		return isEmpty() ? JFaceResources.getString("<empty_selection>") : toList().toString(); //$NON-NLS-1$
+		return isEmpty() ? JFaceResources.getString("<empty_selection>") //$NON-NLS-1$
+				: Arrays.toString(elements);
 	}
 }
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
```diff
@@ -57,4 +60,15 @@ public interface IStructuredSelection extends ISelection, Iterable {
 	 * @return the selected elements as a list
 	 */
 	public List toList();
+
+	/**
+	 * Returns the elements in this selection as a <code>Stream</code>.
+	 *
+	 * @return the selected elements as a stream
+	 * @since 3.31
```
HannesWell: Since this is an API addition, don't we need a minor version bump here?

HannesWell (approved):
I like it.

I assume TreeSelection does not need an override.

In general, wouldn't it make sense to make it a generic-type `IStructuredSelection<T>` and return T as element type in `getFirstElement()`, `iterator()`, `toList()` and `stream()`?

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
laeubi: > Since this is an API addition, don't we need a minor version bump here?

API tools says no and complains otherwise, so who m I to fight the API tools? :-D

laeubi:
> In general, wouldn't it make sense to make it a generic-type IStructuredSelection<T> and return T as element type in getFirstElement(), iterator(), toList() and stream()?

Theoretically yes, in practice this breaks compatibility and would require a "JFace 2.0" that likley never happens, so all one can do is lower the pain with adding new features and leave the rest alone.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
mickaelistria: A minor bump is indeed necessary.
There is a new .api_filter added for this method. If you remove the API filter, don't APT Tools warn you about it?

laeubi: The API filter is because otherwise it wants a **major** (4.0.0) version bump. If I increment the version anyways it complains that the bump is unnecessary and i need another filter for that because it complains that version already was bumped before....

merks: Make IStructuredSelection generic?  Noooo!  😱   Mostly one gets a structured selection from/in places where there is no known T...

laeubi: And how does it matter if you don't know `T`  in contrast of having no `T` possibly defined? ;-)

Beside that this might be true for some platform usages, but the usual usage is you create a tree and add a selection listener to it, so if the Tree would be parameterized with `<T>` and my Selection Listener would as well be parametrized with `<T>` my selection can easily be parametrized to the known type `T` ... nerveless this is absolut **NOT IN SCOPE** of this change...!

mickaelistria:
> The API filter is because otherwise it wants a major (4.0.0) version bump. If I increment the version anyways it complains that the bump is unnecessary and i need another filter for that because it complains that version already was bumped before....

The minor version bump for API addition is a strong requirement that we can't escape. Can this work using those 2 filters?

laeubi:
I now added the second filter and incremented the version... API tools is a mystery for me but maybe @vik-chand can tell if it is the right approach.

iloveeclipse:
> I now added the second filter and incremented the version... API tools is a mystery for me but maybe @vik-chand can tell if it is the right approach.

There is no mystery. Your API baseline should be not 4.28 but 4.29 RC2a. With the right baseline you don't need a second filter.

merks:
> > I now added the second filter and incremented the version... API tools is a mystery for me but maybe @vik-chand can tell if it is the right approach.
> 
> There is no mystery. Your API baseline should be not 4.28 but 4.29 RC2a. With the right baseline you don't need a second filter.

And for that you need to do what I mentioned here:

https://github.com/eclipse-platform/.github/issues/150

iloveeclipse:
> Tere is no mystery. Your API baseline should be not 4.28 but 4.29 RC2a. With the right baseline you don't need a second filter

OK, I was faster as API tooling. Now build finished and reports an error. So the problem is a combination of suppression of an API error adding the new method on an interface and version bump needed for exact that reason.

laeubi:
> There is no mystery. Your API baseline should be not 4.28 but 4.29 RC2a. With the right baseline you don't need a second filter.

The build also complains:

> build / Verify Linux: bundles/org.eclipse.jface/META-INF/MANIFEST.MF#L5
> The minor version should be the same for version 3.32.0, since no new APIs have been added since version 3.31.0

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StructuredSelection.java
```diff
@@ -215,6 +225,7 @@ public List toList() {
 	 */
 	@Override
 	public String toString() {
-		return isEmpty() ? JFaceResources.getString("<empty_selection>") : toList().toString(); //$NON-NLS-1$
+		return isEmpty() ? JFaceResources.getString("<empty_selection>") //$NON-NLS-1$
+				: stream().map(String::valueOf).collect(Collectors.joining(", ", "[", "]")); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
```
szarnekow: I'd do `Arrays.toString(elements)` instead to simplify the code.

mickaelistria: @laeubi can you please address this suggestion?

laeubi:
I hopefully have ow addressed the API tools error

mickaelistria (approved):
That's all good to me, and I would happily use this API in the future!
Let's try to merge soon (before end of week) if no-one has objection.

laeubi:
Instable Test documented here:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/1139

PR #1100, opened 2023-09-12
fix StatusLine progress

jukzi:
Reusing a IProgressManager requires to finally call
progressMonitor.done() before using it again.

see
https://bugs.eclipse.org/bugs/show_bug.cgi?id=485748
and
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/773

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/editors/text/TextFileDocumentProvider.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/editors/text/TextFileDocumentProvider.java
@@ -557,29 +557,36 @@ public class TextFileDocumentProvider implements IDocumentProvider, IDocumentPro
 		LocationKind locationKind= null;
 
 		file= adaptable.getAdapter(IFile.class);
-		if (file != null) {
-			IPath location= file.getFullPath();
-			locationKind= LocationKind.IFILE;
-			manager.connect(location, locationKind,getProgressMonitor());
-			fileBuffer= manager.getTextFileBuffer(location, locationKind);
-		} else {
-			ILocationProvider provider= adaptable.getAdapter(ILocationProvider.class);
-			if (provider instanceof ILocationProviderExtension) {
-				URI uri= ((ILocationProviderExtension)provider).getURI(element);
-				if (ResourcesPlugin.getWorkspace().getRoot().findFilesForLocationURI(uri).length == 0) {
-					IFileStore fileStore= EFS.getStore(uri);
-					manager.connectFileStore(fileStore, getProgressMonitor());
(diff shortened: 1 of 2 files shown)


jukzi:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1077#issuecomment-1714948144

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/editors/text/TextFileDocumentProvider.java
```diff
@@ -557,30 +557,35 @@ protected FileInfo createFileInfo(Object element) throws CoreException {
...
-					fileBuffer= manager.getFileStoreTextFileBuffer(fileStore);
-				}
-			}
-			if (fileBuffer == null && provider != null) {
-				IPath location= provider.getPath(element);
-				if (location == null)
-					return null;
-				locationKind= LocationKind.NORMALIZE;
-				manager.connect(location, locationKind, getProgressMonitor());
+		IProgressMonitor progressMonitor= getProgressMonitor();
```
iloveeclipse, line 560: Can this change on TextFileDocumentProvider be in a dedicated commit with proper commit message explaining the reason?

iloveeclipse (changes requested):

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/editors/text/TextFileDocumentProvider.java
jukzi: good enough?

iloveeclipse: > good enough?

Unfortunately github doesn't allow to comment on commit messages.

The current message says: 

> Reusing a IProgressManager requires to finally call progressMonitor.done() before using it again.

This is incorrect. Correct would be:

`Reusing a monitor received from IStatusLineManager requires to finally call progressMonitor.done() before using it again.`

Note: this info should be added to the `org.eclipse.jface.action.IStatusLineManager.getProgressMonitor()`  API.

In general this change in the IStatusLineManager is counter-intuitive, because the "modern" API like `SubMonitor` explicitly says there shouldn't be need to call `done()` at all.

I don't want to start discussion about all this monitor API mess, because it is something you haven't touched but please update commit message and IStatusLineManager javadoc.

jukzi: done

iloveeclipse (approved):

PR #1104, opened 2023-09-12
[PoC] Fix #1102 - IContentAssistProcessor.computeCompletionProposals dont't know if it's part of autoassist or on-demand assist

### LINKED ISSUE #1102: IContentAssistProcessor.computeCompletionProposals dont't know if it's part of autoassist or on-demand assist
During work on one of WWD bug (eclipse-wildwebdeveloper/wildwebdeveloper#1316) I realized that due Eclipse limitation is not possible to fix LSPE in clean way (eclipse/lsp4e#799)

For now IContentAssistProcessor.computeCompletionProposals is completely blind, and don't know if it was invoked by AutoAssistListener (triggered by char and some delay) or on demand (for example via ctrl/cmd+space)

Would by nice to have third parameter in #computeCompletionProposals, int/enum mode, or separate optional method computeAutoCompletionProposals (with default to computeCompletionProposals) that will allow introduce some differences depending to mode

To implement this, the eldest of the PDT :-) introduces own, extremely ugly, PHPContentAssistant

If you agree I could take care of this, but tell me which way you prefer ;-)

zulus:
This is [PoC] please comment

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistRequest.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistRequest.java
@@ -1,0 +1,55 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Dawid Pakuła and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Dawid Pakuła - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jface.text.contentassist;
+
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.ITextViewer;
+
+/**
+ * Describes the state that the content assistant is in when completing proposals.
+ * <p>
+ * Clients may use this class.
+ * </p>
+ *
+ * @since 3.24
+ * @noimplements
+ */
+public interface IContentAssistRequest {
+
+	/**
+	 * @return the viewer whose document is used to compute the proposals
+	 */
+	public ITextViewer getViewer();
+
+	/**
+	 * @return the document connect to viewer
+	 */
+	public IDocument getDocument();
+
+	/**
+	 * @return document offset
+	 */
+	public int getOffset();
+
+	/**
+	 * @return flag for auto-assist
+	 */
+	public boolean isAutoActivated();
+
+	/**
+	 * @return flag for incremental calls {@link IContentAssistantExtension2}
+	 */
+	public boolean isIncremental();
+
+}
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/ContentAssistRequest.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/ContentAssistRequest.java
@@ -1,0 +1,89 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Dawid Pakuła and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Dawid Pakuła - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jface.internal.text;
+
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.ITextViewer;
+import org.eclipse.jface.text.contentassist.IContentAssistRequest;
+import org.eclipse.jface.text.contentassist.IContentAssistantExtension2;
+
+/**
+ * Describes the state that the content assistant is in when completing proposals.
+ *
+ * @since 3.24
+ */
+public final class ContentAssistRequest implements IContentAssistRequest {
+	/**
+	 * Creates a new event.
+	 *
+	 * @param viewer text viewer
+	 * @param offset assist offset
+	 * @param autoActivated whether content assist was triggered by auto activation
+	 * @param incremental incremental flag {@link IContentAssistantExtension2}
+	 */
+	public ContentAssistRequest(ITextViewer viewer, int offset, boolean autoActivated, boolean incremental) {
+		this.fViewer= viewer;
+		this.fOffset= offset;
+		this.fAutoActivated= autoActivated;
+		this.fIncremental= incremental;
+	}
+
+	/**
+	 * the viewer whose document is used to compute the proposals
+	 */
+	private ITextViewer fViewer;
+
+	/**
+	 * an offset within the document for which completions should be computed
+	 */
+	private int fOffset;
+
+	/**
+	 * Tells, whether content assist was triggered by auto activation.
+	 */
+	private boolean fAutoActivated;
+
+	/**
+	 * Tells, whether content assist was triggered incrementally
+	 */
+	private boolean fIncremental;
+
+
+	@Override
+	public ITextViewer getViewer() {
+		return fViewer;
+	}
+
+	@Override
+	public IDocument getDocument() {
+		return fViewer.getDocument();
+	}
+
+	@Override
+	public int getOffset() {
+		return fOffset;
+	}
+
+	@Override
+	public boolean isAutoActivated() {
+		return fAutoActivated;
+	}
+
+	@Override
+	public boolean isIncremental() {
+		return fIncremental;
+	}
+
+
+}
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     Mickael Istria (Red Hat Inc.) - [251156] async content assist
+ *     Dawid Pakuła - [#1102] isAutoActivated flag
  *******************************************************************************/
 package org.eclipse.jface.text.contentassist;
 
@@ -149,8 +150,8 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 	}
 
 	/**
-	 * This methods differs from its super as it will show the list of proposals that
-	 * gets augmented as the {@link IContentAssistProcessor#computeCompletionProposals(ITextViewer, int)}
+	 * This methods differs from its super as it will show the list of proposals that gets augmented
+	 * as the {@link IContentAssistProcessor#computeCompletionProposals(IContentAssistRequest)}
 	 * complete. All computations operation happen in a non-UI Thread so they're not blocking UI.
 	 */
 	@Override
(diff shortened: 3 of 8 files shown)


bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistRequest.java
```diff
@@ -0,0 +1,95 @@
...
+ * Describes the state that the content assistant is in when completing proposals.
+ * <p>
+ * Clients may use this class.
+ * </p>
+ *
+ * @since 3.24
+ * @see IContentAssistProcessorExtension2
+ * @noinstantiate This class is not intended to be instantiated by clients
+ */
+public final class ContentAssistRequest {
```
laeubi: An interface would be more suitable (and flexible) here I think

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
```diff
@@ -0,0 +1,113 @@
...
+	 * @param viewer the viewer whose document is used to compute the proposals
+	 * @param offset an offset within the document for which completions should be computed
+	 * @return an array of completion proposals or <code>null</code> if no proposals are possible
+	 */
+	@Deprecated
+	default IContextInformation[] computeContextInformation(ITextViewer viewer, int offset) {
+		return computeContextInformation(new ContentAssistRequest(viewer, offset, false, false));
+	}
+
+	static IContentAssistProcessorExtension2 adapt(IContentAssistProcessor processor) {
```
laeubi: You should better use an `AdapterFactory` for this purpose

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
```diff
@@ -0,0 +1,113 @@
...
+
+import org.eclipse.jface.text.ITextViewer;
+
+/**
+ * Extension interface of {@link IContentAssistProcessor} that forward information about
+ * autoActivation
+ *
+ * @since 3.24
+ */
+public interface IContentAssistProcessorExtension2 extends IContentAssistProcessor {
```
laeubi:
```suggestion
public interface IContentAssistProcessorExtension2 extends IContentAssistProcessor2 {
```

it then should also replace/deprecate the methods there that currently take a viewer+offset.

mickaelistria: Why not default methods?

mickaelistria: I don't think we should make the interface extend IContentAssistProcessor by default as it doesn't bring any value. We should document that IContentAssistProcess can be decorated with IContentAssistProcessor2 and annotation that IContentAssistProcessor2 can be implementing IContentAssistProcessor; but I see no immediate value in mandating the hierarchy between both.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistRequest.java
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Creates a new event.
+	 *
+	 * @param viewer text viewer
+	 * @param offset assist offset
+	 * @param autoActivated whether content assist was triggered by auto activation
+	 * @param incremental incremental flag {@link IContentAssistantExtension2}
+	 * @since 3.20
+	 */
+	public ContentAssistRequest(ITextViewer viewer, int offset, boolean autoActivated, boolean incremental) {
```
mickaelistria: If it's `@noinstantiate`, then can't we make the constructed package protected?
But thinking about it, some consumers may like to instantiate it for testing or to reuse the capability of contentAssistant in different context (eg to compute some quick-fixes). So the `@noinstantiate` seems a bit constraining and would maybe better removed.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistRequest.java
zulus: Sure

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
zulus: I just used same approach as IContentAssitProcessorExtension :) but I'll do as you suggest

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistRequest.java
zulus: I'll transform to interface + private and/or inner class

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
zulus: This was my original idea but in this case IDE will not suggest you to implement them immediately

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
zulus: For now I didn't introduced extra IAdapterFactory, it feels to me like to much overhead

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistRequest.java
zulus: I dropped @noinstantiate and prepared internal class

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
zulus: Parent is necessary if we want keep default methods. Without this Implementing both IContentAssistProcessor and IContentAssistProcessorExtensions2 will generate conflict

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
zulus: I made it a little bit different and I hope more flexible:

1. Two inner methods inside ContentAssistant that just verify instance of (no extra classes)
2. Extra IContentAssistantExtension5 with "buildAssistantRequest"

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistRequest.java
zulus: Done

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistRequest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.jface.text.ITextViewer;
+
+/**
+ * Describes the state that the content assistant is in when completing proposals.
+ * <p>
+ * Clients may use this class.
+ * </p>
+ *
+ * @since 3.24
+ * @see IContentAssistProcessorExtension2
```
mickaelistria: Should have `@noimplements`

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
mickaelistria: I don't get it: why would we want to keep default methods here?
Using an IContentAssistProcessor instead of an IContentAssistProcessor2 would be relatively simple to do internally without exposing default methods, wouldn't it?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistantExtension5.java
```diff
@@ -0,0 +1,29 @@
...
+ * function:
+ * <ul>
+ * <li>allows to build IContentAssitRequest</li>
+ * </ul>
+ *
+ * @since 3.24
+ */
+public interface IContentAssistantExtension5 {
+
+	IContentAssistRequest buildAssistantRequest(int offset, boolean isAutoAssit, boolean isIncremental);
```
mickaelistria: Can't this be internal to ContentAssistant?

zulus: According to defaults, without them if want switch to IContentAssistRequest methods only, you still have to build IContentAssistProcessor with old one

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistRequest.java
zulus: This together with IContentAssistantExtension5 introduce more flexibility. For example JDT, PDT, DLTK, CDT build own ContentAssistInvocationContext to keep similar information (their processors have to be programmatically reqistered to assistant) . In this case will be possible to stop using extra ICompletionListener (see ContentAssistEvent).

mickaelistria: How an extensible interface would change the fact that computer that require custom context would need to plug it in the content assistant? And isn't the goal of this work exactly that the IContentAssistRequest to exactly cover all the informations that would allow a CompletionProcessor to pick the right choice, without need for extra context to be injected?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
mickaelistria: OK, I see that we have the `IContentAssistant.getContentAssistProcessor(String contentType)` that we want to keep working with new implementations as well...

mickaelistria: I don't think this argument is very strong: if one is to implement a ContentAssistProcessor, the IDE wouldn't suggest them to implement IContentAssistProcessExtension2 and its method either. So to me, the effort to discover and adopt those new methods is very comparable, and even smaller IMO with default method if they're invoked by the CotnentAssistant (to transitively delegate to older variants) and part of the call stack that we see when debugging. We could even consider deprecating the older methods with a note that this code isn't invoked directly any more byt the ContentAssistant and to recommend implementing the newer variants, and then add some default empty implementations.

zulus: I prepared defaults and dropped IContentAssistProcessorExtension

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistRequest.java
zulus: I'm just looking for opportunity to via one patch,  make possible to always use built-in content assistant processor, especially with GenericTextEditor. In current state If you want introduce categorization (repeated invoke in popup) and/or add some extra information/filters in completion processor, you have to introduce ICompletionListener programmatically connected to concrete ContentAssistant. But for now I'll drop IContentAssistantExtension5 and start working on this, maybe for GenericEditorContentAssistant only

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistProcessorExtension2.java
laeubi: It is better to separate the concerns instead of cluttering the API it also makes it much easier if later adjustments are needed because the interfaces/classes do not really change, so even if there is a small overhead it is worth the effort.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/IContentAssistRequest.java
zulus: Ok what now?

zulus:
@mickaelistria @laeubi any comments? Or should I prepare api_platform exceptions and merge as-is

PR #1105, opened 2023-09-13
Split org.eclipse.search into core and ui bundles

mickaelistria:
Move parts that don't require UI into a `search.core` bundle. Legacy `.search` bundle now contains the bits that require UI.
The `.search.core` plugin is now reusable in deployment which do not include UI bundles.

Existing extension-points are preserved, without changing id/namespace.

Introduce and Use an OSGi (declarative) service to pass the strategy to handle editor-specific dirty document from UI (.search) to Core (.search.core)

This PR is a reoganized variant of https://github.com/eclipse-platform/eclipse.platform.ui/pull/1054


mickaelistria:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1105 adds the copyright header and reorganizes the commits here. I'll merge that other one when it passes and then will close this one.

PR #1108, opened 2023-09-13
Eliminate includes of bundles transitively required by batik

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1349

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -109,35 +109,7 @@
          unpack="false"/>
 
    <plugin
-         id="org.w3c.css.sac"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
          id="org.apache.batik.util"
-         download-size="0"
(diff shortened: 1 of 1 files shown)


laeubi:
I think you can even remove the batik ones all together here...

merks:
> I think you can even remove the batik ones all together here...

Yes, who know what other transitive dependencies aren't really needed.  But I'd like to keep the focus here very small to solve the problem at hand...

merks:
The Code Analysis errors are reported here:

https://github.com/eclipse-platform/eclipse.platform.ui/issues/1109

The continuous-integration/jenkins/pr-head fails because of :

> 11:26:53.491 [ERROR] Cannot resolve project dependencies:
> 11:26:53.491 [ERROR]   You requested to install 'org.eclipse.equinox.p2.iu; org.eclipse.search.core 0.0.0' but it could not be found

Lots of javadoc errors too, but those don't seem to fail the build...

Some tests failed on Windows.  It doesn't look related.

@akurtakov 

All failures look completely unrelated to my changes.  I assume it's okay to merge these changes?

PR #1115, opened 2023-09-14
AccumulatingProgressMonitor: log warning like StatusLineManager

jukzi:
to get meaningfull stacktrace - as AccumulatingProgressMonitor uses asyncExec which does not reveal the caller of the method.


iloveeclipse:
@jukzi : could you please create ticket & target it to M3 to disable logging here and in the manager?

jukzi:
By now i don't know any case where it hits. If it does not hit i see no reason to disable it.

iloveeclipse:
> By now i don't know any case where it hits.

First one: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1125

iloveeclipse:
Yet another one: https://bugs.eclipse.org/bugs/show_bug.cgi?id=582445

PR #1119, opened 2023-09-14
o.e.search regression: extension point has changed name. Fixes #1118

### LINKED ISSUE #1118: [search] Regression caused by https://github.com/eclipse-platform/eclipse.platform.ui/pull/1054
Moving the search engine extension point into a new search.core plugin has caused the extension point's name to change, which is an API change and unacceptable. 

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1054 for breaking change

The extension point definition should be moved back into o.e.search.  The o.e.search.core extension can still load the extensions, but the point must be defined in the previous location.

robstryker:


mickaelistria:
I'm sincerlu not sure whether the bug is in your change or in PDE not honoring namespaces well as they are defined in the schema; but in any case, I think it's fine to merge this change to make everything happy, and maybe consider re-evaluating the extension point definition move some time later.

robstryker:
Let me re-verify this on Monday before we merge. I might have misunderstood the error messages I was seeing as the extension point was missing, when it was just the javadoc stuff. 

I'll leave this open but dive deeper come Monday.

mickaelistria:
If it's only about documentation, then I think you can just tweak the doc files, applying the same rule as if new extension point was created: https://wiki.eclipse.org/How_to_add_things_to_the_Eclipse_doc

robstryker:
Nah I just dug deeper. The bug is real. This PR should fix it. 

From what I can tell, the extension loader does not read the actual schema file at all. It only reads the definition line in the plugin.xml and so (as far as I can tell) completely ignores the `targetNamespace` flag in the schema file. 

And if you think about it, I don't think the runtime cares to read or verify the schema file at all, not even later. It's not doing the work of verifying all required attributes are present. Pretty sure the consumer of the extension point does that... at least I always have in the past. I would always load the configuration elements and verify all required attributes were present. I never trusted the platform to do that for me. 

I wouldn't put any faith in the target-namespace flag of the exsd files in the future. I just don't think the platform actually reads them? Or at least I've seen no evidence thus far it reads them. I wouldn't be surprised if the schema files were mostly for pde / dev time and nothing more.

mickaelistria:
> And if you think about it, I don't think the runtime cares to read or verify the schema file at all

all that makes sense; but then as long as the extension point id doesn't change externally and extensions still use the target extension-point id (ie it's still `org.eclipse.search.textSearchEngine`), then the consuming code just requests extensions to `org.eclipse.search.textSearchEngine` and those extensions do not depend at all on where is the extension-point definition...

merks:
Isn't it the case that the definition can use a fully qualified name such that the symbolic name of the containing bundle doesn't matter?  E.g., 

image

mickaelistria:
See build error

> [ERROR] Failed to execute goal org.eclipse.tycho:tycho-source-plugin:4.0.2:plugin-source (plugin-source) on project org.eclipse.search.core: /home/jenkins/agent/workspace/eclipse.platform.ui_PR-1119/bundles/org.eclipse.search.core/build.properties: src.includes value(s) [schema/] do not match any file

The build.properties file needs to forget about hte schema/ folder.

robstryker:
> Isn't it the case that the definition can use a fully qualified name such that the symbolic name of the containing bundle doesn't matter? E.g.,
> 
> image

So I just tested this. When I go to the plugin.xml editor in org.eclipse.search.core and attempt to mark the extension point id as "org.eclipse.search.textSearchEngine", the editor presents a warning to me. 

`Illegal value 'org.eclipse.search.textSearchEngine' for attribute "id".  Legal characters are "a-z", "A-Z", "0-9", and "_"`

So the plugin.xml editor refuses to accept it, but, perhaps that's just a warning and not representative of the underlying behavior?  So I kept it and ran it and saw what id the extension was added to the model with:  `org.eclipse.search.core.org.eclipse.search.textSearchEngine`

So no... I don't know what equinox has done to accomplish that, but it doesn't seem to be available for common extensions.

mickaelistria:
I guess the trick is the one we see in the adapters.exsd file:
```xml
<schema targetNamespace="org.eclipse.core.runtime">
<annotation>
      <appInfo>
         <meta.schema plugin="org.eclipse.core.runtime" id="adapters" name="Adapters"/>
      </appInfo>
...
```
which seems to force things to PDE.

Anyway, it's still unclear to me what is the issue that makes extension point isn't working any more for consumer side at runtime?

If it's only documentation (eg https://github.com/eclipse-platform/eclipse.platform.ui/issues/1120 ), then manipulating the index to refer to the proper location of extension-point definition can be done as described in https://wiki.eclipse.org/How_to_add_things_to_the_Eclipse_doc .

merks:
There really are an **awful lot** of occurrences of that:

image

robstryker:
I have been unable to get this to work without moving the extension back. I would suggest / request that this patch be accepted.

iloveeclipse:
Since it seem to be unclear what exectly is broken now, can we have a unit test that tries to use this extension point and would fail without this patch?

mickaelistria:
Please see #1122 (which works for me)

laeubi:
Just in case someone is interested in the details, the Extension registry is managed by **Contributors** in the case of OSGi, the **Bundle** is _contribution_ an _extension_, if you lookup an extension it is done by the full qualified name that is `<contributor name>.<extension name>` where the `<contributor name>` is the bundle symbolic name in case of the OSGi Factory used here:

https://github.com/eclipse-equinox/equinox/blob/9ef5cd8ec7d8c4a4db71424d27d13496dd1ffba7/bundles/org.eclipse.equinox.registry/src/org/eclipse/core/runtime/ContributorFactoryOSGi.java#L39-L55

If you look at the code there is one "trick" for migration, you can move the extension point to a **fragment** with two hosts (the "old" bundle and the "new" bundle), then you can use the extension point with the two bundles (but different full qualified names!). One might enhance the concept here so that even two extensions are published in the case of multiple hosts...

The `targetNamespace` of an XML document [is completely independent of that](https://www.oracle.com/technical-resources/articles/srivastava-namespaces.html)!

mickaelistria:
@robstryker Which patch do you think is best to try merging? This one or #1122 ?

robstryker:
Closing in deference to the seeming superior https://github.com/eclipse-platform/eclipse.platform.ui/pull/1122

PR #1124, opened 2023-09-17
Limit the visibility of "Find References" to the active Generic Editor

oleosterhagen:
The visibility of this menu item must be restricted because other editors like the Xtext Editor come with their own "Find References" action.

Fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=564610
Also referenced in https://gitlab.eclipse.org/eclipse-wg/ide-wg/ide-wg-dev-funded-efforts/ide-wg-dev-funded-program-planning-council-top-issues/-/issues/30

With this change the _Xtext Editor_ only shows its own _Find References_ action. I have also tested (with Typescript files) that this menu item is still visible and functional in the _Generic Editor_.


mickaelistria:
Thank you!

PR #1133, opened 2023-09-18
File Search should offer to search in all opened editors

### LINKED ISSUE #1116: File Search should offer to search in all opened editors
In the dialog below I miss the possibility to search & replace in **all** opened editors, not only in the currently active one.

image

I do not see any good reason why we should allow to search / replace in only the currently opened editor, so I would propose to change the dialog above and 
- either change the option "Resource in active editor" to "Resources in opened editors"
- or just add the option "Resources in opened editors"

iloveeclipse:
Adds "Opened editors" scope to the "Text" search.

image

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1116

### DIFF
--- a/bundles/org.eclipse.search/new search/org/eclipse/search/ui/ISearchPageContainer.java
+++ b/bundles/org.eclipse.search/new search/org/eclipse/search/ui/ISearchPageContainer.java
@@ -62,6 +62,13 @@ public interface ISearchPageContainer {
 	public static final int SELECTED_PROJECTS_SCOPE= 3;
 
 	/**
+	 * Current Project scope (value <code>4</code>).
+	 *
+	 * @since 3.16
+	 */
+	public static final int OPENED_EDITORS_SCOPE = 4;
+
+	/**
 	 * Returns the selection with which this container was opened.
 	 *
 	 * @return the selection passed to this container when it was opened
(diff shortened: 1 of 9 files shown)


iloveeclipse:
@ Anyone in platform UI: any comments on this feature?

laeubi:
You can trigger everyone with @eclipse-platform/eclipse-platform-committers

Beside that I think its fine but I'm a bit unsure about the wording, because "Opened editors" seems strange (how do I search in "Closed Editors", are not all editors open anyways?

What about "Files opened in Editors" (as it is a **File**-Search Dialog) .. another alternative would be to have another Tab named "Editors" as many options here might be irrelevant in an Editor search, e.g. will Editors that are not file based ignored? Why should I then not be able to search in not file based editors (I assume any editor that provides an `IDocument` should be searchable?

mickaelistria:
I support the feature overall.
I think "Files opened in editors" is indeed a good suggestion.
Another tab is IMO to be avoided as it makes UI more and more complex and harder and harder to discover. And also, semantically, we're still doing a File Search, and just changing the scope of the search, so the proposal fits well.

iloveeclipse:
OK, going to change the wording to:
image

jukzi:
in Notepad++ it's called "in All Opened Documents", which i frequently use
image
therefore i suggest to word it
`Scope: All Opened Documents`

laeubi:
> therefore i suggest to word it
> Scope: All Opened Documents

As mentioned above this does **NOT** search in _documents_ only in files, and editor *might* be file based or not but both have a document.

iloveeclipse:
If there are no further objections / change requests, I would prefer to merge this PR now.

iloveeclipse:
Test failures in github actions known already:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/751
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

PR #1138, opened 2023-09-19
tests: avoid Zip Slip

jukzi:


tests/org.eclipse.core.filebuffers.tests/src/org/eclipse/core/filebuffers/tests/FileTool.java
```diff
@@ -63,6 +63,9 @@ public static void unzip(ZipFile zipFile, File dstDir) throws IOException {
 					continue;
 				}
 				String entryName = entry.getName();
+				if (!new File(dstDir, entryName).toPath().normalize().startsWith(dstDir.toPath().normalize())) {
```
laeubi, line 66: why not use Path API in the first place here and below?

jukzi: copy & paste from advisory

Issue #921, opened 2023-07-07
org.eclipse.ui.tests do not work in parallel to other tests


laeubi:
As discovered here https://github.com/eclipse-platform/eclipse.platform.ui/pull/917 when running `org.eclipse.ui.tests` in parallel with other tests.

iloveeclipse:
@laeubi : what is your expectation from **UI** tests sharing same host, how they should test in parallel? 

With permanent focus stealing etc the only way to run them in parallel is to open a dedicated vnc session and run every UI test inside such vnc session.

If maven/tycho doesn't offer such functionality, one don't need to create multiple bugs here, it will never work reliably.

laeubi:
> what is your expectation from UI tests sharing same host, how they should test in parallel?

The same I can use different Eclipse instances in parallel even with other applications on the same host maschine.

> With permanent focus stealing etc the only way to run them in parallel is to open a dedicated vnc session and run every UI test inside such vnc session.

If the test is not about focus itself one usually don't need to rely on focus, e.g. SWTBot offers methods that take a control to interact with the UI in a focus independent way, also one might use some kind of mutext-lock if exclusive screen access is required.

If you look here:
https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-917/1/

only a few tests actually fail so there are others that can work perfectly fine so it is not completely hopeless.

iloveeclipse:
> only a few tests actually fail

... in this concrete execution, since it is running in parallel, it will be permanent "russian roulette" which one will fail.

HeikoKlare:
What about disabling parallel execution only for "affected" bundles, i.e., those may having race conditions when executed in parallel, for now? I see that these bundles require most of the overall execution time of a build, but starting with whatever can be parallelized now makes the CI even a bit faster.

Finding all the test cases with potential race conditions during parallel execution may be some effort, but maybe it's worth taking a look at the tests / perform some full parallel builds and have a tracking issue for all test cases that probably need to be adapted for parallel execution. Then it's easier to decide whether it's feasible and worth the effort.

[...]

Issue #920, opened 2023-07-07
org.eclipse.jface.tests.databinding do not work in parallel to other tests


laeubi:
As discovered here https://github.com/eclipse-platform/eclipse.platform.ui/pull/917 when running `org.eclipse.jface.tests.databinding` in parallel with other tests.

Issue #919, opened 2023-07-07
org.eclipse.jface.text.tests do not work in parallel to other tests


laeubi:
As discovered here https://github.com/eclipse-platform/eclipse.platform.ui/pull/917 when running `org.eclipse.jface.text.tests` in parallel with other tests.

Issue #918, opened 2023-07-07
org.eclipse.ui.workbench.texteditor.tests do not work in parallel to other tests


laeubi:
As discovered here https://github.com/eclipse-platform/eclipse.platform.ui/pull/917 when running `org.eclipse.ui.workbench.texteditor.tests` in parallel with other tests.

PR #1148, opened 2023-09-21
Enable org.eclipse.ui.tests.rcp in Tycho build

HeikoKlare:
See https://github.com/eclipse-platform/eclipse.platform.ui/discussions/1143

### DIFF
--- a/tests/org.eclipse.ui.tests.rcp/build.properties
+++ b/tests/org.eclipse.ui.tests.rcp/build.properties
@@ -26,5 +26,7 @@ output.. = bin/
 # Maven properties, see https://github.com/eclipse/tycho/wiki/Tycho-Pomless
 # This plug-in's name does not comply with Tycho's naming convention for test-plugins -> packaging type has to be specified explicitly
 pom.model.packaging = eclipse-test-plugin
-# Test not yet ready for Surefire
-pom.model.property.skipTests = true
+pom.model.property.testClass = org.eclipse.ui.tests.rcp.RcpTestSuite
+# Tests have to run headless, with an application
+pom.model.property.tycho.surefire.useUIHarness = false
+pom.model.property.tycho.surefire.useUIThread = true


HeikoKlare:
Failing test is documented in #370.

Any objections on merging this test (re-)activation?

iloveeclipse:
> Any objections on merging this test (re-)activation?

The only one objection I see is that after merging so many repos into platform.ui the test times increased a lot, jenkins needs now nearly a hour to complete. Would be nice if someone could split test execution on jenkins based on changed code. 

I believe that was proposed on one of the "merge git repo" actions.

HeikoKlare:
I see you point. Splitting up test execution absolutely makes sense. Even for the GH actions builds, it could make sense to have different jobs for subsets of the test projects to reduce the time for getting feedback when submitting a PR. Seems to be a important topic to me, but should be considered independent from this PR.

Concerning the tests enabled by this PR: they add some few seconds (less than 10) to test execution time. At the same time, they seem to be the only tests that actually test some startup/shutdown behavior of the workbench (correct me if I'm wrong), so it's well invested time (from my point of view). Also note that this PR only enables the tests in the Tycho build executed by GH actions. I did not find information in the Jenkins logs that the test bundle is considered in those builds at all, so I think this PR will not affect the Jenkins builds.

akurtakov:
FWIW, only 1 repo (platform.text) was merged into platform.ui the rest were merged into platform one.

HeikoKlare:
> FWIW, only 1 repo (platform.text) was merged into platform.ui the rest were merged into platform one.

True. If I'm not mistaken, that merge by itself doubled the build time (from ~30 minutes to ~1 hour).

I hope you also see the general build time topic as something to be addressed independent from this PR and that it should not prevent this test reactivation to be performed.

akurtakov:
For the sake of having correct data:
* 4.28 (aka prior merge) - 36 mins (https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/R4_28_maintenance/)
* 4.29 (aka after the merge) - 45 mins (https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/R4_29_maintenance/)

So slow down is 9 mins there must be other things slowing down the build. 
Most likely build time depends on load of the CI servers - as per https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/ build time ranges from 41 to 56 mins.

akurtakov:
I've rebased the patch on master so there is another build to compare with the first one (56 mins). There will most likely be multi minute difference just proving your point that speeding up the build is not an issue here.

HeikoKlare:
Yes, build times vary a lot. I just considered the times the whole GitHub actions CI check took [right before the merge](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/5370682845) and [right after the merge](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/5399689866). But actually it does not really matter what the reason for the build times is and what effect the merge had, as no one wanted to challenge the decision for performing the merge :-)

Build times added through test activation in this PR can also be seen in the logs. E.g., the Tycho build in the GH actions run on Windows took around 12 seconds in its last execution, see https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/6259047161/job/16994277236#step:7:20179.

akurtakov:
Next build took 49 mins on Jenkins (https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1148/2/).

akurtakov:
Merging.

HeikoKlare:
Thanks @akurtakov!

laeubi:
Just to clarify here about the varying build times, actually it is not only the test-execution/compilation itself, it is also API analysis that takes a considerable amount of time, some bundles require **minutes** to complete because they have large dependency chains and currently API tools recompile them to perform the analysis and in general is quite slow.

There is already a switch to skip API analysis if the bundle itself was replaced by the baseline, so if there is no change in the bundles this is skipped and save us some time, but I don't think this will work smoothly for test as it will require a deep analysis of the dependencies and likely will always return true as we have only some very big test bundles instead of small module local ones, given the tight interaction of platform code, literally **any** change can break anything.

Next there are several issues I created that currently seem to hinder us using parallel execution (and therefore maybe use more CPUs) but the conclusion was more "it's impossible" or "to risky":

- https://github.com/eclipse-platform/eclipse.platform.ui/issues/921
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/920
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/919
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/918

so at the moment I do not see much room for improvement without major effort what might outweigh some additional minutes to wait for the verification build.

laeubi:
I have now created a profiler PR here:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1153

Issue #339, opened 2022-09-16
NavigationHistoryEntry logic can lead to entrys being merged together when they shouldn't be


Greglar4:
In the mergeInto method of NavigationHistoryEntry, it checks if the EditorInput equals the EditorInput of the currentEntry parameter, and merges the NavigationHistoryEntries if so.

However, in  our use case, we have two different editors whose EditorInputs are FileEditorInputs, and they can be editors for the same file. The equals method in FileEditorInputs returns true if the two FileEditorInputs are associated with the same underlying file.

Therefore, the NavigationHistoryEntries are merged together if the user navigates between two editors which edit the same underlying file, even though these two editors are very different editors, and should be treated as different editors for navigation purposes.

merks:
Maybe org.eclipse.ui.internal.NavigationHistoryEntry.mergeInto(NavigationHistoryEntry should be implemented like this to also take the editor ID into account:
```
	/**
	 * Merges this entry into the current entry. Returns true if the merge was
	 * possible otherwise returns false.
	 */
	boolean mergeInto(NavigationHistoryEntry currentEntry) {
		if (Objects.equals(editorInfo.editorInput, currentEntry.editorInfo.editorInput)
				&& Objects.equals(editorInfo.editorID, currentEntry.editorInfo.editorID)) {
			if (location != null) {
				if (currentEntry.location == null) {
					currentEntry.location = location;
					location = null;
					return true;
				}
				return location.mergeInto(currentEntry.location);
			} else if (currentEntry.location == null) {
				return true;
			}
		}
		return false;
	}
```

It's really easy to set up a development environment to contribute a pull request:

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#create-an-eclipse-development-environment

Greglar4:
I could contribute a pull request with this change. Would it be at all possible for this to be downbranched into Eclipse 2020-12? If not, can you think of a way we could implement this change ourselves? I was looking at it and it seemed that to edit the NavigationHistoryEntry we would have to edit NavigationHistory, which then seemed to require that we go all the way up and implement our own Workbench.

mickaelistria:
@Greglar4 Platform project requires that incoming changes are based on top of master first. Once approved for master, we can on request backport the change to older branches; however Eclipse project wouldn't produce a new build of 2020-12/4.18 so you'd need to build it on your end anyway.

mickaelistria:
[...]

PR #1150, opened 2023-09-21
Navigation History Entry - Issue #339

Greglar4:
Changes the implementation of the mergeInto method of the class NavigationHistoryEntry. This is to fix the reported github issue #339, where a user has multiple different editors whose EditorInputs are both FileEditorInputs with the same underlying file. Thus the editorInputs of the multiple editors were considered the same, and when navigating between the editors, the navigation history entries were being merged, incorrectly.

The supplied change checks both if the editor inputs and editorIDs of two navigation history entries are the same before merging the two navigation history entries.


Greglar4:
I failed the ECA agreement check, but have now agreed to the ECA agreement in my Eclipse account.

mickaelistria (dismissed):
2 editors using the same FileEditorInput usuually use the same underlying IDocument and are necessarily synchronized. The history is not really per editor but per document; so I don't think this patch is really valid.

merks:
I've run into this problem too with clients.  In my case they were not using IDocuments and I worked around it with this approach using an editor ID.  Even if documents are shared and synchronized, the different editors can be in different locations in the document in the different editors.  So to me this looks like a good change that can only improve the user's experience. 

@mickaelistria I get the impression that your comment is somehow based on the concept editing history where synchronized shared document are relevant, but this change is about navigation history, i.e., what the left and right arrow in the toolbar do.  Is there perhaps a misunderstanding here?  This navigation history is an extensible framework and non-textual editor can plug into that...

mickaelistria (commented):
@merks You're right, sorry, I misunderstood. Then it all makes more sense to me, I'm removing my negative review as it's not relevant here and leave it up to you and @Greglar4 to complete it as I've just demonstrated I'm not really reliable on that piece of code ;)

merks:
@Greglar4 

It looks like the builds themselves are broken right now, unrelated to you changes, so I'll need to kick them again later...

The ECA is failing again, but that was passing before I think.  Are you have to access this page:

https://api.eclipse.org/git/eca/status/gh/eclipse-platform/eclipse.platform.ui/1150

For me it shows only a partial email address, but there is a field where you can enter a full email address; I just don't know if those things work for you....h

Greglar4:
Checking my email address at the Git ECA Validation tool has it pass, but I noticed my github profile did not have an email address selected. I updated it to the email address that passes the validation tool, is this sufficient?

Greglar4:
I have also edited my author and commiter email on the actual commit, if that was necessary.

merks:
Can you amend the commit, ensure that the author email of the commit is the one associated with your eclipse account and force push that?

merks:
The ECA passed so it looks promising.

merks:
FYI, I'm just testing if merge in master fixes the build problems...

PR #1152, opened 2023-09-21
Remove "classic" Search API

akurtakov:
It was not used for ages and has been deprecated since 2016. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=487303 for actual deprecation


vogella:
Nice one, thanks. I remember we tried to remove it earlier but failed due to the "old style of handling things".

PR #1153, opened 2023-09-22
Enable maven profiler to see what adds to the build times

laeubi:
This add the maven profiler and enables to output the results to the console.

This is not really meant to be merged, just to see what currently takes the most time when building a PR.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -5,4 +5,9 @@
 		<artifactId>tycho-build</artifactId>
 		<version>4.0.4</version>
 	</extension>
+	<extension>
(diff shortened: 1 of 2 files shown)


laeubi:
@akurtakov looking at the profiling results, the javadoc generation takes considerable time. I can try to skip this for plugins that have not changed but of course then we only get new issues reported (not existing ones).

akurtakov:
Please no. There are enough times javadoc bites us to want it to be in better shape.

akurtakov:
I think this draft can be closed now, no?

laeubi:
It depends it can still be used to see what is taking most of the time, I'll rebase to get a new measure now.

Issue #1158, opened 2023-09-22
Mute monitor logs from AccumulatingProgressMonitor.beginTask


iloveeclipse:
Follow up on https://github.com/eclipse-platform/eclipse.platform.ui/pull/1115.

While running latest I Build with our application tests I observe a lot of cases where clients can't do anything meaningful to prevent error log polluted with "beginTask should only be called once per instance" warnings from AccumulatingProgressMonitor.

In the stack below framework started beginTask in the progress dialog while waiting on a job and in the job client code simply used given job monitor to call SubMonitor.convert(monitor). This is fully valid use case, but there is no way to avoid the warning logged.

```
!ENTRY org.eclipse.jface 2 0 2023-09-20 20:32:21.203
!MESSAGE beginTask should only be called once per instance. At least call done() before further invocations
!STACK 0
java.lang.IllegalStateException: beginTask should only be called once per instance. At least call done() before further invocations
	at org.eclipse.jface.operation.AccumulatingProgressMonitor.beginTask(AccumulatingProgressMonitor.java:148)
	at org.eclipse.ui.internal.progress.ProgressManager$JobMonitor.lambda$0(ProgressManager.java:265)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1092)
	at org.eclipse.ui.internal.progress.ProgressManager$JobMonitor.beginTask(ProgressManager.java:265)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:483)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:449)
	at some.client.method()
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:43)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
Caused by: java.lang.IllegalStateException: beginTask(, 1000) was called here previously
	at org.eclipse.jface.operation.AccumulatingProgressMonitor.beginTask(AccumulatingProgressMonitor.java:155)
	at org.eclipse.ui.internal.progress.ProgressManager$JobMonitor.addProgressListener(ProgressManager.java:244)
	at org.eclipse.ui.internal.progress.ProgressMonitorFocusJobDialog$3.runInUIThread(ProgressMonitorFocusJobDialog.java:213)
[...]

PR #1161, opened 2023-09-22
Examples of valid use of monitors that log warnings now

iloveeclipse:
Import org.eclipse.ui.examples.job project, start Eclipse from debugger, open "Job Factory" view, toggle "user job" and click on "Create jobs" button. It will flood the error log with warnings.

image

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1158.


jukzi:
thanks, i will look into it.

PR #1174, opened 2023-09-26
Fix code mining sometimes not rendering

### LINKED ISSUE #1170: [Code Mining] rendering issues with folding
Folded ranges seem to cause code-mining to not render in some cases.
Please see attached screencast
[codemining.webm](https://github.com/eclipse-platform/eclipse.platform.ui/assets/469058/67cc6b19-98b9-4b0d-9eba-4c499929f0c3)

It looks like the condition evaluating whether to render some code mining or not fails at taking folding/projection into account.

mickaelistria:
Because of confusion between document and widget offset in some comparison.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1170


mickaelistria:
Failure seems unrelated and not reproducible on all environments. Everything else seems fine according to Jenkins build. Let's merge.

PR #1175, opened 2023-09-28
Align Snippet 030 with 043 Snippet

vogella:
Snippet 043 uses var for the DatabindingContext so do the same for 030, this is minimal change.

Note: This is just an example for an client, so feel free to disregard this change.


laeubi:
Why not "align" it the other way round? Using `var` here seem not bring any benfit and making it harder to read...

vogella:
Closing as this is just a small example and @laeubi disagrees with this change.

PR #1178, opened 2023-09-29
Clean Up: Add final modifier to private fields

jukzi:


HeikoKlare (commented):
Makes sense to have these variables declared as `final`. If I see correctly, only some versions need to be bumped and then this PR can be merged?

jukzi:
ignoring random failing HoverTest, UIEditWorkingSetWizardAuto

https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

PR #1132, opened 2023-09-18
[Modern Find/Replace] Refactor FindReplaceDialog

Wittmaxi:
The FindReplaceDialog's business-logic is now handled in the FindReplacer.
The FindReplacer communicates using a FindReplaceStatus object which is polled by FindReplaceDialog as needed and allows for displaying correct messages in the Interface.

This Refactoring is required so that the business-logic can be reused in the new FindReplace overlay, as proposed in issue #1090

PR #1181, opened 2023-09-29
Increase Version Number to accomodate #1132

Wittmaxi:
As discussed in #1132


Wittmaxi:
@vogella

vogella:
I think you plan to keep the API stable? So only increase the minor segment by 100

vogella:
3.17.200 -> 3.17.300

Wittmaxi:
Yes, you are absolutely right, the API should stay stable. I have updated my PR :) @vogella

iloveeclipse (changes requested):
This PR is not needed, version was already bumped for 4.30 via 324229d513c8735524422904c659d8b1404183bb

vogella:
@Wittmaxi so the build failure of https://github.com/eclipse-platform/eclipse.platform.ui/pull/1132 is not related to missing version increase?

HeikoKlare:
I think the build there is failing, because the `FindAndReplace` class is currently marked `public`, which would be a public API change. Reducing its visibility will resolve the version problem.

PR #1186, opened 2023-10-03
Hinder multiple completion proposal popups (fix #1185)

### LINKED ISSUE #1185: Possible problem in file org/eclipse/jface/text/contentassist/CompletionProposalPopup.java#L1248
Hi.

According to the discussion, I transformed it into an issue. Attached you can find: i) Main.java a very brief example to reproduce the problem; simply start the application while in the editor, press ctrl+space multiple times (many times, to be able to see that there are actually many overlapped content assistants open - see figure1).

The problem looks to be at https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java#L1248

Rationale
- condition 1241 is true only when the listener is armed and it is armed if and only if the widgettoken is acquired by the ContentAssistant (see code of addContentAssistListener)
- if the listener is installed, the ContentAssistant is displayed
- however, a second ContentAssistant cannot acquire the widgettoken and the execution gets to line 1248 and thus, the  (nth) ContentAssistant is still displayed (?!?!). Instead of going to 1248, I would expect the execution to get to 1252 and hide the additional ContentAssist (actually, this looks to be the case earlier - see this blame view where hide() is executed when the listener/widgettoken cannot be acquired https://github.com/eclipse-platform/eclipse.platform.ui/blame/a1ac8bebc4317a8647163943422a9ed27907aedc/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java#L1225)

### Discussed in https://github.com/eclipse-platform/eclipse.platform.ui/discussions/1184

<div type='discussions-op-text'>

<sup>Originally posted by **petrufm** October  2, 2023</sup>
Hi,

In file https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java#L1248 I observed it would be possible to make visible a content assist without arming its CompletionProposalPopup (this) listener. Is this intended?  Among others, multiple content assists may be made visible simultaneously.

Thanks.</div>

[Main.java.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/12789284/Main.java.zip)
figure1

petrufm:
Fixes #1185

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java
@@ -1245,9 +1245,6 @@ class CompletionProposalPopup implements IContentAssistListener {
 			if (showPopup) {
 				configureAndMakeVisible();
 			}
-		} else if (!fProposalShell.isVisible() && showPopup) {
-			// if a previous call did initialize but did make the popup visible, this will make sure the popup will be visible.
-			configureAndMakeVisible();
 		} else
 			hide();
 	}


mickaelistria:
Please use a more descriptive commit message; the commit messages matter much more on the long run that the issue in the tracker, so they're expected to be helpful to other developers more directly and not just referencing an issue.

Something like
```
Fix possible multiple completion proposal popup

Prevents showing a popup in case ....
```

I hope to give a try to this patch in the Generic Editor with asynchronous completion tomorrow.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java
```diff
@@ -1245,9 +1245,6 @@ void displayProposals(boolean showPopup) {
 			if (showPopup) {
 				configureAndMakeVisible();
 			}
-		} else if (!fProposalShell.isVisible() && showPopup) {
```
iloveeclipse, line 1248: This was added via 78fc9f6f6566323c691f1fee1eb14de2f27e7d34.

@gayanper : could you please explain why this block was needed and how removing it would affect your original use case? Unfortunately there were no ticket reported...

petrufm:
Done.

Regarding the async completion. Yeah, it would be good to try the patch. To help here a little: please note that the addContentAssistListener uses acquireWidgetToken which at its turn uses requestWidgetToken. In the documentation of the latter it is said "Requests the widget token from this token owner. Returns true if the token has been acquired or **is already owned by the requester**". Thus, I would expect that an async popup that was already initialised but not displayed, will be displayed in line 1246 (since the content asist already has the token and it will get it again during listener (re)install). I hope this will help a little.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java
gayanper: I'm not sure if I remember correctly, but I do see that the comment is not correctly phrased. But I guess there was a scenario where the documentListinser was already installed but the shell was not visible. Couldn't remember if it was due to the completion popup was canceled due to more keypress been done.

mickaelistria:
I'm tempted to merge it in the current form so we will test it more extensively in practice before next milestone and have proper time and enough usage feedback to decide whether we need to fix or revert. @gayanper any objection here?

PR #1072, opened 2023-08-29
TreeViewer: auto expansion for elements with single children

Wittmaxi:
As implementation of #1063, This PR addresses the need to automatically expand paths which consist of chained singular nodes, eg:

```
com.wittmaxi.plugin
 └─src
    └─org.foo.com
       └─Bar.java
```
It introduces two new API-Methods into TreeViewer:

-setAutoExpandTrivialPathsLevel
-getAutoExpandTrivialPathsLevel

The PR also uses the new features in the Project Explorer, thus resolving #1063

PR #1191, opened 2023-10-05
Refactored AbstractTreeViewer: extracted setRedraw

Wittmaxi:
Extracted from  #1072

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -112,6 +112,13 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	private boolean isTreePathContentProvider = false;
 
 	/**
+	 * Keeps track of how many methods have called
+	 * {@link #executeWithRedrawDisabled } before it could reset the state of the
+	 * tree.
+	 */
+	private int redrawCount = 0;
+
+	/**
 	 * Safe runnable used to update an item.
 	 */
 	class UpdateItemSafeRunnable extends SafeRunnable {
@@ -774,16 +781,12 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	 */
 	public void collapseToLevel(Object elementOrTreePath, int level) {
 		Assert.isNotNull(elementOrTreePath);
-		Control control = getControl();
-		try {
-			control.setRedraw(false);
+		executeWithRedrawDisabled(() -> {
 			Widget w = internalGetWidgetToSelect(elementOrTreePath);
 			if (w != null) {
 				internalCollapseToLevel(w, level);
 			}
-		} finally {
-			control.setRedraw(true);
-		}
+		});
 	}
 
 	/**
@@ -1143,19 +1146,18 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	public void expandToLevel(Object elementOrTreePath, int level, boolean disableRedraw) {
 		if (checkBusy())
 			return;
-		Control control = getControl();
-		try {
-			if (disableRedraw) {
-				control.setRedraw(false);
-			}
+
+		Runnable expandToLevelOperation = () -> {
 			Widget w = internalExpand(elementOrTreePath, true);
 			if (w != null) {
 				internalExpandToLevel(w, level);
 			}
-		} finally {
-			if (disableRedraw) {
-				control.setRedraw(true);
-			}
+		};
+
+		if (disableRedraw) {
+			executeWithRedrawDisabled(expandToLevelOperation);
+		} else {
+			expandToLevelOperation.run();
 		}
 	}
 
@@ -1606,15 +1608,12 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	@Override
 	protected void inputChanged(Object input, Object oldInput) {
 		preservingSelection(() -> {
-			Control tree = getControl();
-			tree.setRedraw(false);
-			try {
+			executeWithRedrawDisabled(() -> {
+				Control tree = getControl();
 				removeAll(tree);
 				tree.setData(getRoot());
 				internalInitializeTree(tree);
-			} finally {
-				tree.setRedraw(true);
-			}
+			});
 		});
 	}
 
@@ -2272,12 +2271,10 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 
 	@Override
 	protected void labelProviderChanged() {
-		// we have to walk the (visible) tree and update every item
-		Control tree = getControl();
-		tree.setRedraw(false);
-		// don't pick up structure changes, but do force label updates
-		internalRefresh(tree, getRoot(), false, true);
-		tree.setRedraw(true);
+		executeWithRedrawDisabled(() -> {
+			Control tree = getControl();
+			internalRefresh(tree, getRoot(), false, true);
+		});
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -774,16 +774,15 @@ public void collapseAll() {
...
-			Widget w = internalGetWidgetToSelect(elementOrTreePath);
-			if (w != null) {
-				internalCollapseToLevel(w, level);
+		executeWithRedrawDisabled(new Runnable() {
+			@Override
+			public void run() {
+				Widget w = internalGetWidgetToSelect(elementOrTreePath);
+				if (w != null) {
+					internalCollapseToLevel(w, level);
+				}
```
HeikoKlare: You can simplify this by putting into a lambda expression:
```suggestion
		executeWithRedrawDisabled(() -> {
			Widget w = internalGetWidgetToSelect(elementOrTreePath);
			if (w != null) {
				internalCollapseToLevel(w, level);
			}
```
This also applies to the other instantiations of `Runnable`.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -3401,4 +3396,38 @@ public boolean contains(Object parent, Object element) {
 		return false;
 	}
 
+	/**
+	 * Transactionally perform an action on the TreeViewer without redrawing
+	 * regularly. Disables redraw, executes toExecute and enables redraw again.
```
HeikoKlare: nit: put `toExecute` in `@code{toExecute}` to make it easier to see that it's a code reference

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -3401,4 +3396,38 @@ public boolean contains(Object parent, Object element) {
...
+			if (shouldDisableRedraw) {
+				control.setRedraw(false);
+			}
+			toExecute.run();
+		} finally {
+			if (shouldDisableRedraw) {
+				control.setRedraw(true);
+			}
+		}
+	}
```
HeikoKlare: This method is currently misleading, as its name states that it disables redraw while it accepts an argument that defines whether it actually does so.
I propose either of the folowing:
- Inline this code as it is only used once.
- Rename the method to something like `execute(Runnable toExecute, boolean withRedrawDisabled)`. In that case, the Javadoc has to be corrected as well (the second parameter is not documented).

laeubi:
I think the wording "transactionally" is completely confusing here, a [transaction in computer sience](https://en.wikipedia.org/wiki/Transaction_processing)  something completely different than what happening here.

also `execute` is way to generic here, i would suggest using `runWithRedraw` or similar. Also note that this is not really safe to be called in a recursive manner as it does not restore the previous state, it simply assumes a state and in some cases reset it to this assumption. simple example I have previously already called `treeviewer.getControl().setRedraw(false)` this state will not be restored when I `labelProviderChanged` is called for whatever reason in between.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -1143,20 +1139,12 @@ public void expandToLevel(Object elementOrTreePath, int level) {
 	public void expandToLevel(Object elementOrTreePath, int level, boolean disableRedraw) {
 		if (checkBusy())
 			return;
-		Control control = getControl();
-		try {
-			if (disableRedraw) {
-				control.setRedraw(false);
-			}
+		execute(() -> {
```
laeubi: You should extract the runnable here and call run here directly if `disableRedraw = false` then you don't need the paramter what makes the whole thing way mor confusing then required.

HeikoKlare:
> I think the wording "transactionally" is completely confusing here, a [transaction in computer sience](https://en.wikipedia.org/wiki/Transaction_processing) something completely different than what happening here.

👍 

> Also note that this is not really safe to be called in a recursive manner as it does not restore the previous state, it simply assumes a state and in some cases reset it to this assumption.

True. Even though this is existing behavior, it seems to be reasonable to store the initial redrawing state and reset this state after execution. The method should then document that in case the executed operation modifies the redraw state, this will be overwritten afterwards.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
HeikoKlare: I agree. That conforms to what I meant with the first option in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1191#discussion_r1347120412 and I think it is actually the better of the two options.

Wittmaxi:
@HeikoKlare there is currently no (... I couldn't find any ...) function that returns a redraw state in the public API of `Control`. The windows-Implementation has a method called `getDrawing` which isn't public API. The GTK-Implementation does not have that method. Hence, I see myself forced to invalidate the previous draw-state of the tree, unless you have a better idea?

At the very least, I can mark this with a comment in the JavaDoc, warning of invalidated states.
Since the tree is expected to "redraw by default" anyway, I don't think that it's too big of a stretch to assume the tree is always redrawing, except for operations that happen inside of `executeWithredrawDisabled` - maybe we can find a good solution for recursive functions that call `executeWithRedrawDisabled` themselves?

HeikoKlare:
I see. When I take a look at all the callers of `setRedraw`, then most (all?) of them use is to temporarily disable redraw. I.e., the all follow the pattern of disabling redraw, doing something, and reenabling it afterwards. Many of them even don't re-enable redraw in a finally block to leave the system in a proper state in case something fails.

So I think this `executeWithoutRedraw` method should actually rather be placed inside `Control` and used by all these callers. To ensure that the method behaves as expected, it could ...
1. be  `synchronized` to avoid concurrent calls, and
2. maintain a number/stack of calls to ensure that redraw is only re-enabled when returning from the last call

Then we would have:
1. A reusable logic to execute something on a control while redrawing is disabled.
2. A fail-safe implementation of that logic, which always restores the old state, even if the code to be executed fails.
3. A thread-safe and recursion-aware implementation of this logic.

iloveeclipse:
> 1. be  `synchronized` to avoid concurrent calls

SWT is single threaded, whoever calls redraw **must** be in UI thread. Please don't add synchronized in UI / SWT layer.

HeikoKlare:
> > 1. be  `synchronized` to avoid concurrent calls
> 
> SWT is single threaded, whoever calls redraw **must** be in UI thread. Please don't add synchronized in UI / SWT layer.

Would probably be unnecessary anyway if the callers are tracked as proposed in 2. (which reduces to tracking the number of recursions in case of single-threaded access).

vogella:
What is the status here?

HeikoKlare:
This is to be supseded by a more general PR performing the refactoring at the level of `Controls` rather than on the `AbstractTreeViewer` only.

vogella:
> supseded

So we can close this one? Do you have a link to the new PR?

HeikoKlare:
From my point of view, we can close this one. A new PR does not exist yet. It will require some effort to adapt all places where redrawing is temporarily disabled, but I would prefer that over a local solution that only improves the behavior for one single kind of control. Maybe we can use the quiet period until 4.31 development opens to prepare an according PR.

Wittmaxi:
Closed with regard to Heiko's comment, created #1285

PR #1196, opened 2023-10-06
Use pomless build for org.eclipse.e4.tools.test

vogella:


vogella:
Not sure if I should add skipAPIAnalysis to build.properties, as test typically do not have API, I think we should be fine here.

Issue #1199, opened 2023-10-08
Can/should we remove batik includes from


merks:
The most recent update to batik required a "qualifier touch":

https://github.com/eclipse-platform/eclipse.platform.ui/commit/dde3b606e801b648b466f5ed9fe52e4915f95298

If we look at the batik dependencies, , ignoring feature dependencies, we see that only the org.eclipse.e4.ui.css.core bundle depends on a batik package:

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/208716/920b0a62-7123-418d-a4ce-aeb59c24e95e)

The rest of the batik dependencies are transitive dependencies of the directly required org.apache.batik.css.

So my sense is that we really don't need to specify any dependencies on batik in the feature.  By removing those includes, we make the Platform more flexible to accommodate newer versions of batik and of course we make it easier to update the target platform in the future without "qualifier touching".

@akurtakov  @iloveeclipse 

Do you have any concerns?

merks:
I assume this is fine if the builds are. I hope everyone is fine with that.

PR #1193, opened 2023-10-06
Increase build duration timeout

mickaelistria:
As more and more tests are running as part of this build, we need to increase the timeout.

PR #1200, opened 2023-10-08
Remove batik includes from the org.eclipse.e4.rcp feature

merks:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1199

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -102,20 +102,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.apache.batik.css"
-         download-size="0"
-         install-size="0"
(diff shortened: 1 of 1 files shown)


merks:
@iloveeclipse @akurtakov 

It looks like builds are taking so long (> 60 minutes) such that they abort.  😱

mickaelistria:
I've tried #1193 about build duration, but it appears there could be a real issue to fix somewhere instead of just increasing timeout.

laeubi:
You can always rebase this PR and look at the output to find out what takes the time:

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1153

merks:
I kicked another build manually and that completed and passed in 51 minutes.  That’s close to timeout.

PR #1213, opened 2023-10-11
Offer simplified version to create LocalResourceManager

vogella:
Frequently LocalResourceManager(JFaceResources.getResources(), parent) is used in client code. This add a helper method on JFaceResource to simplify such call.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/JFaceResources.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/JFaceResources.java
@@ -32,6 +32,7 @@ import org.eclipse.jface.preference.PreferenceDialog;
 import org.eclipse.jface.wizard.Wizard;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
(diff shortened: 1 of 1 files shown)


vogella:
Random failing test already known, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

Planning to merge soon cc @laeubi @HannesWell

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/JFaceResources.java
```diff
@@ -583,6 +584,23 @@ public static void setFontRegistry(FontRegistry registry) {
...
+	 * control.
+	 *
+	 * shortcut for
+	 * <code>LocalResourceManager(JFaceResources.getResources(), owner)</code>
+	 *
+	 * @param owner control whose disposal will trigger cleanup of everything in the
+	 *              registry.
+	 */
+	static LocalResourceManager managerFor(Control owner) {
+		return new LocalResourceManager(getResources(), owner);
```
laeubi, line 602: I would use control.getData/setData to cache the instance created so multiple calls could be performed without creating multiple managers.

HannesWell: Can you estimate how often that will hit the cache?
Since the local-manager is a child from the global one caching would not save SWT-resources but mainly only another map.

vogella:
I think caching can be applied in a different commit if desired. This one had the intention to provide a simpler way to create a new LocalResourceManager which I think it does.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/JFaceResources.java
```diff
@@ -583,6 +584,23 @@ public static void setFontRegistry(FontRegistry registry) {
...
+	 * should not be used to allocate any resource that must outlive the given
+	 * control.
+	 *
+	 * shortcut for
+	 * <code>LocalResourceManager(JFaceResources.getResources(), owner)</code>
+	 *
+	 * @param owner control whose disposal will trigger cleanup of everything in the
+	 *              registry.
+	 */
+	static LocalResourceManager managerFor(Control owner) {
```
szarnekow, line 601: Why is this package private?

vogella: > Why is this package private?

Thanks for the catch. Will be fixed with 
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1218

PR #1216, opened 2023-10-12
Add empty tree functionality to CheckedTreeSelectionDialog

### LINKED ISSUE #1215: Allow a dialog that uses CheckedTreeSelectionDialog to have empty tree
A number of JDT dialogs use SourceActionDialog which sub-classes CheckedTreeSelectionDialog.  In some cases, such as Generate Constructor from Fields... it is possible to deselect all fields and the operation succeeds, however, if a user has a class that has no fields, the dialog is not opened and fails with an error dialog.  Opening the dialog with an empty tree does not work because the CheckedTreeSelectionDialog will change the status to error.

This issue proposes allowing a sub-class of CheckedTreeSelectionDialog to specify if an empty tree is allowed with the default being no.

This is to fix: https://bugs.eclipse.org/bugs/show_bug.cgi?id=552556

jjohnstn:
- add new isEmptyTreeAllowed() protected method to CheckedTreeSelectionDialog to allow sub-classes to permit empty trees
- fixes: #1215

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -229,6 +229,16 @@ public class CheckedTreeSelectionDialog extends SelectionStatusDialog {
 	}
 
 	/**
+	 * Returns whether an empty tree is allowed
+	 *
+	 * @return true if empty tree is ok, false otherwise
+	 * @since 3.131
+	 */
+	protected boolean isEmptyTreeAllowed() {
+		return false;
+	}
+
+	/**
 	 * Validate the receiver and update the status with the result.
 	 *
 	 */
@@ -238,6 +248,11 @@ public class CheckedTreeSelectionDialog extends SelectionStatusDialog {
 				fCurrStatus = fValidator.validate(fViewer.getCheckedElements());
 				updateStatus(fCurrStatus);
 			} else if (!fCurrStatus.isOK()) {
+				fCurrStatus = new Status(IStatus.OK, PlatformUI.PLUGIN_ID, IStatus.OK, "", //$NON-NLS-1$
+						null);
+			}
+		} else if (isEmptyTreeAllowed()) {
+			if (!fCurrStatus.isOK()) {
 				fCurrStatus = new Status(IStatus.OK, PlatformUI.PLUGIN_ID, IStatus.OK, "", //$NON-NLS-1$
 						null);
 			}
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
@@ -102,7 +102,7 @@ public class CheckedTreeSelectionDialog extends SelectionStatusDialog {
 	 * @param parent             The shell to parent from.
 	 * @param labelProvider      the label provider to render the entries
 	 * @param contentProvider    the content provider to evaluate the tree structure
-	 * @param isEmptyTreeAllowed true if an empty tree can be input, false otherwise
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
```diff
@@ -228,6 +228,16 @@ public void setSize(int width, int height) {
 		fHeight = height;
 	}
 
+	/**
+	 * Returns whether an empty tree is allowed
+	 *
+	 * @return true if empty tree is ok, false otherwise
+	 * @since 3.131
+	 */
+	protected boolean isEmptyTreeAllowed() {
```
mickaelistria: Do you want this flag to change for a given instance? If not, what about making it a constructor argument instead?

jjohnstn: I did not want the flag to change.  I'll add a new constructor as suggested.

mickaelistria:
In which circumstances would one want to not allow empty trees? Do we really have to make it a new flag? Can't we just allow empty trees as default and sole behavior?

jjohnstn:
The dialog is meant to show a set of input and allow the user to check/uncheck various elements.  The current long-time default is to flag an empty tree as an error and to post a message that is specified by the sub-class - the logic being the dialog isn't needed if there isn't anything for the end-user to choose and no action is usually possible.  It is sub-classed by various UI actions such as Generate toString, Create unimplemented constructors.  In almost all cases, you don't want to bring up the dialog if there isn't anything to do and you often have to police if all items are deselected as an action isn't possible.  A user opened a valid issue that even though the action to Create a constructor using fields had no fields to use, it should create the empty constructor.  There may be further actions in the future that may allow an empty tree but it is understandable that the default behavior is to enforce input is available to display.

jjohnstn:
@mickaelistria Can you please review?  This is needed prior to fixing the JDT side.  The windows failure does not appear to have anything to do with my change.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
```diff
@@ -82,6 +82,8 @@ public class CheckedTreeSelectionDialog extends SelectionStatusDialog {
 
 	private int fStyle = SWT.BORDER;
 
+	private boolean fIsEmptyTreeAllowed;
```
mickaelistria:
```suggestion
	private final boolean fAllowEmptyTree;
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
```diff
@@ -94,6 +96,21 @@ public CheckedTreeSelectionDialog(Shell parent, ILabelProvider labelProvider,
 		this(parent, labelProvider, contentProvider, SWT.BORDER);
 	}
 
+	/**
+	 * Constructs an instance of <code>ElementTreeSelectionDialog</code>.
+	 *
+	 * @param parent             The shell to parent from.
+	 * @param labelProvider      the label provider to render the entries
+	 * @param contentProvider    the content provider to evaluate the tree structure
+	 * @param isEmptyTreeAllowed true if an empty tree can be input, false otherwise
```
mickaelistria:
```suggestion
	 * @param allowEmptyTree <code>true</code> if an empty tree can be input, <code>false</code> if an empty tree must be treated as an error
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/CheckedTreeSelectionDialog.java
```diff
@@ -228,6 +245,16 @@ public void setSize(int width, int height) {
 		fHeight = height;
 	}
 
+	/**
+	 * Returns whether an empty tree is allowed
+	 *
+	 * @return true if empty tree is ok, false otherwise
+	 * @since 3.131
+	 */
+	protected boolean isEmptyTreeAllowed() {
```
mickaelistria: Is this method actually consumed in any current client? ie, could it be removed?

mickaelistria (commented):
I added a few comments, nitpicking mostly for the internal part (eg docs, names...); but the question as to whether we still need an API method is worth being discussed before merge.

jjohnstn:
@mickaelistria Please review for M3.  I have used the new final field as suggested and removed the protected method as asked.

mickaelistria (approved):

PR #1218, opened 2023-10-13
Make JFaceResources#managerFor public

vogella:


iloveeclipse:
Ideally every new API should have a test added.

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/resources/JFaceResourcesTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.Test;
+
+public class JFaceResourcesTest {
+
+	@Test
+	public void ensureLocalResourceManagerCanBeCreated() {
+		Display d = Display.getDefault();
+		Shell shell = new Shell(d);
+		Composite composite = new Composite(shell, SWT.NONE);
+		LocalResourceManager localResourceManager = JFaceResources.managerFor(composite);
```
HannesWell, line 33: Could we check here if `localResourceManager.getDevice()` returns the display `d`?
Just to have any kind of validation.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/JFaceResources.java
```diff
@@ -597,8 +597,9 @@ public static void setFontRegistry(FontRegistry registry) {
 	 *
 	 * @param owner control whose disposal will trigger cleanup of everything in the
 	 *              registry.
+	 * @return LocalResourceManager
 	 */
-	static LocalResourceManager managerFor(Control owner) {
+	static public LocalResourceManager managerFor(Control owner) {
```
HannesWell, line 603: The usual convention is to mention the visibility first.
```suggestion
	public static LocalResourceManager managerFor(Control owner) {
```

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/resources/JFaceResourcesTest.java
szarnekow: We should probably dispose the shell, too. And check that the resource manager is disposed afterwards?

szarnekow:
Since the original ticket was motivated by repeated code, it would be great if the new API was used throughout the codebase (local to this repository).

vogella:
> Since the original ticket was motivated by repeated code, it would be great if the new API was used throughout the codebase (local to this repository).

I agree, could be a good follow-up PR.

vogella:
> Ideally every new API should have a test added.

Done

vogella:
Virtual test failure fails randomly, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005

PR #1219, opened 2023-10-13
Mark deprecated Dialog constants from 2005 for removal

vogella:
Marked as deprecated since 2005, should be marked for removal for the future.


vogella:
[VirtualTable ](https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005) unrelated.

PR #1220, opened 2023-10-13
Update the JFace.action tests to Junit4

vogella:
Instead of inheriting TestCase these tests use now a JUnit rule for setup and teardown.


HeikoKlare (approved):
Changes look good. Thanks for migrating to JUnit 4!
Checks only seem to be broken because of API tooling issue: https://github.com/eclipse-pde/eclipse.pde/issues/782.

vogella:
Thanks @HeikoKlare

Issue #1226, opened 2023-10-17
NotificationBuilder API does not allow to position the notification on the bottom of the Shell


marcushoepfner:
Refer to issue https://github.com/eclipse-platform/eclipse.platform/issues/740

The shell cannot be given to the builder in NotificationPopup

marcushoepfner:
Snippets should be adapted as well:

> Note: It looks like the new builder snippets are also not added to the JFace Snippet page, https://wiki.eclipse.org/JFaceSnippets. We should do this and potentially migrate this page to source control as the wiki will disagree soon.

marcushoepfner:
@vogella I have changed the Snippet081AbstractNotificationPopup you have mentioned.
Look how short this has become with all the new APIs. But is not about AbstractNotificationPopup anymore, so I'm not sure whether this one should be changed. WDYT? just rename the class? 
I don't see a reason to propose older APIs when we have nice new ones.
```java
package org.eclipse.jface.notifications;

import org.eclipse.jface.widgets.WidgetFactory;
import org.eclipse.swt.SWT;
import org.eclipse.swt.events.SelectionEvent;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Button;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class Snippet081AbstractNotificationPopup {

	private Display display = new Display();
	private final Shell shell = WidgetFactory.shell(SWT.NONE).layout(new FillLayout(SWT.VERTICAL)).create(display);

	public static void main(String[] args) {
		new Snippet081AbstractNotificationPopup().start();
	}

	private void start() {
		WidgetFactory.button(SWT.PUSH).text("Hello World").onSelect(this::select).create(shell);
		WidgetFactory.button(SWT.PUSH).text("I am a notification popup").onSelect(this::select).create(shell);
		WidgetFactory.button(SWT.PUSH).text("Press me!").onSelect(this::select).create(shell);

		shell.setSize(500, 500);
		shell.open();

		while (!shell.isDisposed()) {
			if (!display.readAndDispatch())
				display.sleep();
		}
		display.dispose();
	}

	private void select(SelectionEvent event) {
		String text = ((Button) event.widget).getText();
		NotificationPopup.forShell(shell).delay(10000).text(text).build().open();
	}
}
```

vogella:
[...]

PR #1227, opened 2023-10-17
Notification Popup Builder API for shell

marcushoepfner:
Allow setting the shell in the builder API for Notification Popups.

Issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1226


vogella:
LGTM, unrelated test failure is documented here https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

Issue #1177, opened 2023-09-28
Add constructor to ImageDescriptor for simple creation based on the bundle


vogella:
To construct an ImageDescriptor from the same plug-in I have to write a lot of code. Example;:

```

ResourceManager resourceManager = new LocalResourceManager(JFaceResources.getResources(),
					goToLocationButton);
			Bundle bundle = FrameworkUtil.getBundle(getClass());
			URL goToFolderUrl = FileLocator.find(bundle, IPath.fromOSString("icons/full/obj16/goto_input.png"), //$NON-NLS-1$
					null);
			goToLocationButton.setImage(resourceManager.create(ImageDescriptor.createFromURL(goToFolderUrl)));

```
Could we add a constructor for this (relatively) frequent use case? Or do we already have easier API and I just didn't notice?

vogella:
Adding @HannesWell as he has driven lots of API simplifications in the past.

laeubi:
@vogella why using the bundle/FileLocator here?

You can simply use `getClass().getResource("/icons/full/obj16/goto_input.png")`

laeubi:
@vogella there is even a convenience method already:

`org.eclipse.jface.resource.ImageDescriptor.createFromFile(Class<?>, String)`

vogella:
@laeubi I use Github search and used the first fitting entry https://github.com/search?q=repo%3Aeclipse-platform%2Feclipse.platform.ui%20resourceManager.create&type=code I will try to simplify ResourceInfoPage.java so that next time, I can copy and paste this code. :-)

laeubi:
> I use Github search and used the first fitting entry 

Sadly Platform itsel do not always serve as the best (modern) example so its never bad to look at the class used itself :-)

vogella:
> > I use Github search and used the first fitting entry
> 
> Sadly Platform itsel do not always serve as the best (modern) example so its never bad to look at the class used itself :-)

:-) Or once we note, we just update the platform code.... I will try to do this for this use case.

vogella:
Thanks @laeubi for the reminder

HannesWell:
> @vogella there is even a convenience method already:
> 
> `org.eclipse.jface.resource.ImageDescriptor.createFromFile(Class<?>, String)`

We could consider to add another overload that detemines the caller class using StackWalker?


> Sadly Platform itsel do not always serve as the best (modern) example so its never bad to look at the class used itself :-)

[...]

PR #1231, opened 2023-10-18
ImageDescriptor API for creating based on the bundle

vogella:
Simplifies the access of an image in a bundle.

Test needs to be added, see FileImageDescriptorTest

For #1177

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
@@ -18,6 +18,8 @@ import java.net.URI;
 import java.net.URL;
 import java.util.function.Supplier;
 
+import org.eclipse.core.runtime.FileLocator;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.swt.SWTException;
 import org.eclipse.swt.graphics.Device;
 import org.eclipse.swt.graphics.Image;
@@ -26,6 +28,8 @@ import org.eclipse.swt.graphics.ImageDataProvider;
 import org.eclipse.swt.graphics.PaletteData;
 import org.eclipse.swt.graphics.RGB;
 import org.eclipse.swt.widgets.Display;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
 
 /**
  * An image descriptor is an object that knows how to create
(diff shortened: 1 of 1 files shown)


vogella:
@HeikoKlare are you already updating  FileImageDescriptorTest somewhere to Junit4? I did not yet add tests to this API, as I did not want to interfere with your JFace Junit 4 work.

vogella:
@ptziegler please review

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -87,6 +91,21 @@ public static ImageDescriptor createFromFile(Class<?> location, String filename)
...
+	 * Creates and returns a new image descriptor from a plug-in.
+	 *
+	 * @param location to determine the bundle which contains the file
+	 * @param filePath path to the file relative in the bundle
+	 * @return a new image descriptor
+	 * @since 3.32
+	 */
+	public static ImageDescriptor createFromBundle(Class<?> location, String filePath) {
+		Bundle bundle = FrameworkUtil.getBundle(location);
+		URL url = FileLocator.find(bundle, IPath.fromOSString(filePath),
```
ptziegler, line 104: This will most likely throw a NullPointerException in case `FrameworkUtil.getBundle()` returns null. I think an explicit null check, which then simply returns `getMissingImageDescriptor()`, would make sense.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -87,6 +91,21 @@ public static ImageDescriptor createFromFile(Class<?> location, String filename)
...
 
+	/**
+	 * Creates and returns a new image descriptor from a plug-in.
+	 *
+	 * @param location to determine the bundle which contains the file
+	 * @param filePath path to the file relative in the bundle
+	 * @return a new image descriptor
+	 * @since 3.32
+	 */
+	public static ImageDescriptor createFromBundle(Class<?> location, String filePath) {
```
ptziegler, line 102: What happens if you want to load an image from a different bundle where you don't have a class?
In WindowBuilder, we have something like this, for example:

> ResourceManager.getPluginImage("org.eclipse.jdt.ui", "/icons/full/elcl16/ch_cancel.png");

Perhaps it makes sense to add a second (or even third) method:

```
ImageDescriptor createFromBundle(Class<?> location, String filePath) {
  return createFromBundle(FrameworkUtil.getBundle(location), String filePath);
}

ImageDescriptor createFromBundle(String symbolicName, String filePath) {
  return createFromBundle(Platform.getBundle(symbolicName), String filePath);
}

ImageDescriptor createFromBundle(Bundle bundle, String filePath) {
  ...
}
```

HeikoKlare:
> @HeikoKlare are you already updating FileImageDescriptorTest somewhere to Junit4? I did not yet add tests to this API, as I did not want to interfere with your JFace Junit 4 work.

No, I have not started working on that yet (and will probably not be able to do so for at least the next week), so feel free to add tests there. Thanks for asking!

ptziegler:
Is there anything else I can help out with here? Because I'd be really awesome if this change can make it into the 2023-12.

PR #1234, opened 2023-10-20
Remove TestRunLogUtil

### LINKED ISSUE #1229: Eclipse UI Tests log seem to include useless setup / teardown output
Looking at the Eclipse UI Tests build logs I see lots of:

```
----- testGetDescriptor
testGetDescriptor: setUp...
testGetDescriptor: tearDown...
```
Is this useful? I typically not interested in lot of output in case tests are running fine, but more interested in output if something goes wrong. This makes it harder in my opinion to find the relevant output.

WDYT @fedejeanne and @HeikoKlare? Is this useful output for you (as you are currently working on the tests).

fedejeanne:
The class is only used to add a TestWatcher that could write to the standard output when starting/finishing a test, but this information is unnecessary since the name of the test currently running is already logged to the console in the CI build.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1229


HeikoKlare:
If I see correclty, `TestRunLogUtil` is part of the public API, so you cannot remove it without a major version increment. In case this PR shall be merged, I'd propose to keep the class and mark it as deprecated (for removal), so in case anyone outside uses it, they will be informed.

fedejeanne:
I'm retracting this PR since it doesn't really solve all of the problems discussed in #1229 . Without a better alternative for logging (other than using `System.out.println()`), there is no easy way to see which test is hanging in real time

PR #1239, opened 2023-10-23
TarInputStream: Prevent Implicit narrowing conversion

BeckerWdf:
https://github.com/eclipse-platform/eclipse.platform.ui/security/code-scanning/4

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarInputStream.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarInputStream.java
@@ -27,10 +27,10 @@ import java.nio.charset.StandardCharsets;
  */
 public class TarInputStream extends FilterInputStream
 {
-	private int nextEntry = 0;
-	private int nextEOF = 0;
-	private int filepos = 0;
-	private int bytesread = 0;
+	private long nextEntry = 0;
+	private long nextEOF = 0;
+	private long filepos = 0;
+	private long bytesread = 0;
 	private TarEntry firstEntry = null;
 	private String longLinkName = null;
 
@@ -79,7 +79,7 @@ public class TarInputStream extends FilterInputStream
 	 * @return false if the entry has already been passed
 	 */
 	boolean skipToEntry(TarEntry entry) throws TarException, IOException {
-		int bytestoskip = entry.filepos - bytesread;
+		long bytestoskip = entry.filepos - bytesread;
 		if(bytestoskip < 0) {
 			return false;
 		}
@@ -310,7 +310,7 @@ public class TarInputStream extends FilterInputStream
 			return -1;
 		}
 		if(len > nextEOF) {
-			len = nextEOF;
+			len = (int) nextEOF;
 		}
 		int size = super.read(b, off, len);
 		nextEntry -= size;
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarEntry.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarEntry.java
@@ -23,7 +23,7 @@ public class TarEntry implements Cloneable
 	private String name;
 	private long mode, time, size;
 	private int type;
-	int filepos;
+	long filepos;
 
 	/**
 	 * Entry type for normal files.
@@ -42,7 +42,7 @@ public class TarEntry implements Cloneable
 	 * @param name filename
 	 * @param pos position in the file in bytes
 	 */
-	TarEntry(String name, int pos) {
+	TarEntry(String name, long pos) {
 		this.name = name;
 		mode = 0644;
 		type = FILE;


jukzi:
I think it will resolve the warning, but probably not allow to decode TARs with huge files.
See also https://github.com/eclipse-pde/eclipse.pde/pull/750/files for basically the same problem. It would be cool if you could invest the time for an actual test

BeckerWdf:
> It would be cool if you could invest the time for an actual test

You mean a test, if we could decode large tar files?

jukzi:
> You mean a test, if we could decode large tar files?

yes, files that are larger then 4G - to let size not fit in int :-)

BeckerWdf:
> > You mean a test, if we could decode large tar files?
> 
> yes, files that are larger then 4G - to let size not fit in int :-)

But fixing this code warning is something different then supporting 4G sized tars, isn't it?

jukzi:
well, the warning showed that the code will probably fail on large files. So instead of only getting rid of the warning, we might solve the problem the root cause.

merks:
Perhaps we have bigger fish to fry than writing a test case to test that we can untar a 4G tar file?

laeubi:
> Perhaps we have bigger fish to fry than writing a test case to test that we can untar a 4G tar file?

Please se the previous discussion. So if we don't *care* about the problem, *suppressing* the warning is defiantly a no-go as this will then never ever be fixed/noticed again and should just live with the warning.

In all other cases, simply migrating *everything to long* will fix the problem and is much less work than all the discussions around ;-)

merks:
@laeubi 

Sorry you are definitely correct that one should fix a problem, not just suppress a warning about a problem.  I was just hoping no one was going to create a 4G test case!  😱

laeubi:
If one absolutely like of course one *might* do a local test with such a large tar, what is not really that isoteric if you look at [what tar was intended for](https://en.wikipedia.org/wiki/Tar_(computing)) that is creating file-system dumps on tape disks, so with todays terabytes of harddisk size it can easily exceed 4 GB without a problem.

Nevertheless for my concerns it would be enough to just change all members to long especially the `filepos` what sadly is not detected at all as a problem by the automatic scanner and why I always suggest a manual review of such automatic suggestions instead of blindly apply such "fix"...

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarInputStream.java
```diff
@@ -27,10 +27,10 @@
  */
 public class TarInputStream extends FilterInputStream
 {
-	private int nextEntry = 0;
+	private long nextEntry = 0;
 	private int nextEOF = 0;
```
laeubi:
```suggestion
	private long nextEOF = 0;
```

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarInputStream.java
```diff
@@ -27,10 +27,10 @@
...
 	private int filepos = 0;
```
laeubi:
```suggestion
	private long filepos = 0;
```

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarInputStream.java
BeckerWdf: Done

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarInputStream.java
BeckerWdf: Done

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/TarInputStream.java
```diff
@@ -323,7 +323,7 @@ public int read(byte[] b, int off, int len) throws IOException {
 			return -1;
 		}
 		if(len > nextEOF) {
-			len = nextEOF;
+			len = (int) nextEOF;
```
laeubi, line 313: Why cast to int here? I think we need to check here first if it can be cast to an int without exceeding MAX_INT and if throw an IOException.

laeubi: Thinking a bit more this can not happen ... so ignore this.

laeubi (approved):
Looks sane now.

laeubi:
To make things worse we also have `org.eclipse.equinox.internal.p2.core.helpers.TarEntry` and `org.eclipse.pde.api.tools.internal.util.TarEntry` as a copy of this class... and cosequently `org.eclipse.pde.api.tools.internal.util.TarFile` and `org.eclipse.equinox.internal.p2.core.helpers.TarInputStream` that need the same fix...

BeckerWdf:
> To make things worse we also have `org.eclipse.equinox.internal.p2.core.helpers.TarEntry` and `org.eclipse.pde.api.tools.internal.util.TarEntry` as a copy of this class... and cosequently `org.eclipse.pde.api.tools.internal.util.TarFile` and `org.eclipse.equinox.internal.p2.core.helpers.TarInputStream` that need the same fix...

I can provide fixes for them in a separate PR.

BeckerWdf:
failed test are unrelated to this change

BeckerWdf:
@jukzi : Is this change ok for you?

merks:
Warning, this is a hijack.  It seems one can't have too many TarEntry classes:

image

That's what happens when no single one of them wants to be API.   

API does exist though:

image

jukzi:
> @jukzi : Is this change ok for you?

As long as nobody tests we don't know if it makes anything better, but it doesn't look worse. I abstain from vote.

PR #1252, opened 2023-10-27
javadoc: fix some @see

jukzi:


bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/ExpandableComposite.java
```diff
@@ -887,10 +887,10 @@ public int getExpansionStyle() {
 	/**
 	 * Changes expanded state.
 	 *
-	 * Equivalent to <code>setExpanded(expanded, false)</code>. @see {@link #setExpanded(boolean, boolean)}.
+	 * Equivalent to <code>setExpanded(expanded, false)</code>.
 	 *
-	 * @param expanded
-	 *            the new expanded state
+	 * @param expanded the new expanded state
```
marcushoepfner, line 892: typically the left hand part is in. a new line for parameters isn't it.
Don't want to be picky, but this was something which again would need to be fixed in the future. maybe you wanna fix it now.

jukzi: However it's not an error/warning to have a new line. It could be solved with reformating the file(s). But i am scared to do it because it creates a giant touch to the blame.

PR #1253, opened 2023-10-28
Restore command state support #925

basilevs:
Restore state support for command handlers as documented in [IObjectWithState](https://github.com/eclipse-platform/eclipse.platform.ui/blob/50553ea8426afdc390e14676abf8906c44d184a4/bundles/org.eclipse.core.commands/src/org/eclipse/core/commands/IObjectWithState.java#L29) Javadoc.
See previous work in [Bug 549802](https://bugs.eclipse.org/bugs/show_bug.cgi?id=549802)
Fixes #925

This PR assumes that `IObjectWithState` protocol is not deprecated. If it is feasible to deprecate it, do not merge this PR.

Command state support had been lost during migration to E4 as proxies failed to properly support `IObjectWithState` protocol or ignored it altogether.

- [HandlerProxy](https://github.com/eclipse-platform/eclipse.platform.ui/blob/50553ea8426afdc390e14676abf8906c44d184a4/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/handlers/HandlerProxy.java#L63)
- [HandlerServiceHandler](https://github.com/eclipse-platform/eclipse.platform.ui/blob/50553ea8426afdc390e14676abf8906c44d184a4/bundles/org.eclipse.e4.core.commands/src/org/eclipse/e4/core/commands/internal/HandlerServiceHandler.java#L36)
- [E4HandlerProxy](https://github.com/eclipse-platform/eclipse.platform.ui/blob/50553ea8426afdc390e14676abf8906c44d184a4/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/handlers/E4HandlerProxy.java#L53)

do now support the protocol. Following tests are modified to rely on command state polling and reenabled:

StateTest
ToggleStateTest
RadioStateTest

A new test WorkbenchStateTest demonstrates how state propagation works in workbench environment without explicit polling.

Main changes:
- internal APIs are extended to implement IObjectWIthState
- HandlerServiceHandler is now stateful

PR #1254, opened 2023-10-28
Test PR builds API verification #1253

basilevs:
Builds are failing even if API is unchanged.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench; singleton:=true
-Bundle-Version: 3.131.0.qualifier
+Bundle-Version: 3.131.100.qualifier
 Bundle-Activator: org.eclipse.ui.internal.WorkbenchPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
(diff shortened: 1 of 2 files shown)


basilevs:
@merks @akurtakov this PR contains no changes, but the API verification fails due to removal of MultiPageEditor.

merks:
I thought Alex deleted that a while back and suppressed the error…

basilevs:
> I thought Alex deleted that a while back and suppressed the error…

The diagnostics only kicks in, when version of the plugin is modified, which makes testing complicated.  I guess enabling diagnostics for all plugins on every build slows down builds too much?

basilevs:
I will prepare a proper MAJOR version bump in  a dedicated MR.

merks:
We’ll have to wait for Alex. We’ll need to suppress the error. But usually one sees the error in the ide and use quick fix for that.

basilevs:
> We’ll have to wait for Alex. We’ll need to suppress the error. But usually one sees the error in the ide and use quick fix for that.
Story goes:

- I did not bump any versions and IDE was silent.
- Then Jenkins complained about changes qualifier, but not patch version.
- I've modified the patch version.
- The IDE did show the API verification error and suggested to bump the major version.
- So I did.

So to bump version properly at least one Jenkins build is required. Probably due to the nature of problem filtering.

UPDATE: the error appears on every minor and patch bump, as API problem filters match version exactly and are invalidated on every version change.

merks:
> I will prepare a proper MAJOR version bump in a dedicated MR.

No major version bump!  Ever!!

akurtakov:
> > I will prepare a proper MAJOR version bump in a dedicated MR.
> 
> No major version bump! Ever!!

Let's not say "ever". Who knows what comes tomorrow ?

mickaelistria:
Never say never ;)

PR #1255, opened 2023-10-28
Migrate from javax.inject/annotation to jakarta.inject/annotation

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1056


mickaelistria:
If not already done (the change is too big to easily ensure it),can we keep 1 small test with javax.inject to ensure backward-compatibility?

HannesWell:
> If not already done (the change is too big to easily ensure it),can we keep 1 small test with javax.inject to ensure backward-compatibility?

The `org.eclipse.e4.core.tests`, which seem to test most of the injector and associated classes are currently still using the `javax` annotations. My suggestion is to just duplicate all the tests into a second test-plugin org.eclipse.e4.core.javax.tests that duplicates the tests as they are (using javax) and then the 'original' tests are migrated to the jakarta annotations as well.
When removing javax support the project `org.eclipse.e4.core.javax.tests` can then be deleted too.
But that is the in eclipse.platform and we could completely eclipse.platform.ui.

laeubi:
> My suggestion is to just duplicate all the tests into a second test-plugin 

Do we really need a seperate plugin or is a dedicated package sufficient?

HannesWell:
> > My suggestion is to just duplicate all the tests into a second test-plugin
> 
> Do we really need a seperate plugin or is a dedicated package sufficient?

I'm currently working on making the javax annotations optional and therefore would like to have for the jakarta-annotations tests a test-runtime (at least with Tycho) that does not contain the javax-annotations to verify if that really works.

HannesWell:
> > If not already done (the change is too big to easily ensure it),can we keep 1 small test with javax.inject to ensure backward-compatibility?
> 
> The `org.eclipse.e4.core.tests`, which seem to test most of the injector and associated classes are currently still using the `javax` annotations. My suggestion is to just duplicate all the tests into a second test-plugin org.eclipse.e4.core.javax.tests that duplicates the tests as they are (using javax) and then the 'original' tests are migrated to the jakarta annotations as well. When removing javax support the project `org.eclipse.e4.core.javax.tests` can then be deleted too. But that is the in eclipse.platform and we could completely eclipse.platform.ui.

I have now created https://github.com/eclipse-platform/eclipse.platform/pull/791 for that purpose.

mickaelistria:
> I have now created https://github.com/eclipse-platform/eclipse.platform/pull/791 for that purpose.

Thanks!

PR #1266, opened 2023-11-02
Remove some deprecated SubProgressMonitor

jukzi:
new SubProgressMonitor -> SubMonitor.convert


bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/CreateChangeOperation.java
```diff
@@ -119,20 +119,20 @@ public void run(IProgressMonitor pm) throws CoreException {
 				int allTicks= conditionTicks + rtp.getCreateChangeTicks() + rtp.getInitializeChangeTicks();
 				pm.beginTask("", allTicks); //$NON-NLS-1$
 				pm.subTask(""); //$NON-NLS-1$
-				fCheckConditionOperation.run(new SubProgressMonitor(pm, conditionTicks));
+				fCheckConditionOperation.run(SubMonitor.convert(pm, conditionTicks));
 				RefactoringStatus status= fCheckConditionOperation.getStatus();
 				if (status != null && status.getSeverity() < fConditionCheckingFailedSeverity) {
-					fChange= fRefactoring.createChange(new SubProgressMonitor(pm, rtp.getCreateChangeTicks()));
+					fChange= fRefactoring.createChange(SubMonitor.convert(pm, rtp.getCreateChangeTicks()));
```
kthoms, line 125: Is this really safe? The Javadoc of convert() says

> Since it is illegal to call beginTask on the same IProgressMonitor more than once, the same instance of IProgressMonitor must not be passed to convert more than once.

Do you know if this is still valid, and this here is an example of the unrecommended usage?

jukzi: It works in this case, because convert() has a special case for monitor "instanceof SubMonitor" not calling beginTask again.
A Junit that calls this operation is for example org.eclipse.jdt.ui.tests.refactoring.RenamePackageTests.testHierarchicalJUnit() which also still works if i replace NullProgress with org.eclipse.core.tests.harness.FussyProgressMonitor.FussyProgressMonitor() and assertUsedUp().

I agree that this code does not look clean but it was not before and i have no intention to manually clean-up i.e. rework all that stuff.

It was not possible to get rid of all SubProgressMonitor with this search&replace (see https://github.com/eclipse-equinox/equinox/discussions/392 ) so we may also just leave the code as-is and the PRs.

jukzi: RenamePackageTests.test2() can show 
that ProcessorBasedRefactoring.checkFinalConditions(IProgressMonitor)
illegally calls beginTask twice, both with and without this PR.

```
java.lang.AssertionError: beginTask may only be called once (old name=Checking preconditions...)
	at org.junit.Assert.assertFalse(Assert.java:65)
	at org.eclipse.core.tests.harness.FussyProgressMonitor.beginTask(FussyProgressMonitor.java:96)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:557)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:512)
	at org.eclipse.ltk.core.refactoring.participants.ProcessorBasedRefactoring.checkFinalConditions(ProcessorBasedRefactoring.java:226)
	at org.eclipse.jdt.ui.tests.refactoring.RenamePackageTests.test2(RenamePackageTests.java:1141)


```

kthoms: Thanks for the clarification. Then it is fine for me.

akurtakov:
@jukzi if you're confident about this one please land it in tomorrow at latest so it doesn't have to wait for March release.

jukzi:
can wait for m1. nobody really needs it

jukzi:
I'll submit this as i also fixed the wrong progress using SubMonitor now

jukzi:
Oh no, i reviewed some cases that don't have test cases and almost everywhere it should be `subMonitor.split()` instead of `SubMonitor.convert()`. That's too much work for manual conversion.

jukzi:
fix only the errors: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1277

PR #1270, opened 2023-11-03
Condense test name output

vogella:
To shorten the build log and make is easier to find the right elements only the executed method names are written to the log.

We now see:

----- testGetSharedImages
----- testGetWorkingSetManager
----- testGetPrefereneManager
----- testGetEditorRegistry
----- testGetWorkbenchWindows
----- testGetPerspectiveRegistry

Before that change we saw:

----- testGetSharedImages
testGetSharedImages: setUp...
testGetSharedImages: tearDown...

----- testGetWorkingSetManager
testGetWorkingSetManager: setUp...
testGetWorkingSetManager: tearDown...

----- testGetPrefereneManager
testGetPrefereneManager: setUp...
testGetPrefereneManager: tearDown...

----- testGetEditorRegistry
testGetEditorRegistry: setUp...
testGetEditorRegistry: tearDown...

----- testGetWorkbenchWindows
testGetWorkbenchWindows: setUp...
testGetWorkbenchWindows: tearDown...

----- testGetPerspectiveRegistry
testGetPerspectiveRegistry: setUp...
testGetPerspectiveRegistry: tearDown...

For #1229

### DIFF
--- a/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/TestRunLogUtil.java
+++ b/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/TestRunLogUtil.java
@@ -24,8 +24,8 @@ public final class TestRunLogUtil {
 	private static final String LINE_BREAK = System.lineSeparator();
 
 	/**
-	 * A {@link TestWatcher} to log test start and end. Should be added as @Rule in
-	 * every JUnit4 test class (which not already extends {@link UITestCase}) as
+	 * A {@link TestWatcher} to log current test method name. Can be added as @Rule
+	 * in every JUnit4 test class (which not already extends {@link UITestCase}) as
 	 *
 	 * <pre>
 	 * &#64;Rule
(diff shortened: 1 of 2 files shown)


iloveeclipse (changes requested):
The point of original code is to ducument which test started and finished properly. This can't be achieved with a single line. I honestly don't understand what are you trying to achieve here. Reducing output will not help anyone, in contrast, it will make understanding of test failures harder.

If you aren't interested in test log output, simply **don't read it**, but do not remove it!

jukzi (changes requested):
For me it was helpful to see both start and stop of test in the logs, when analyzing build errors. Because otherwise it was sometimes not clear at which test the build crashed. The text  "setUp"/"tearDown..." was however never intuitive. could also be something like start/stop, but i need both events!

vogella:
> If you aren't interested in test log output, simply don't read it, but do not remove it!

The reason why I wanted to have it shorter is that scrolling in the Github output to find the right output is just slow with a lot (and in my opinion) unnecessary output. But as you guys are using it, lets keep it.

PR #1281, opened 2023-11-06
fix NON-NLS warnings

jukzi:


jukzi:
ignoring random fails

Issue #949, opened 2023-07-20
Save the bounds (size, location) of dialogs


luapfr:
As it has the possibility of resizing the screens, it would be very good to save the size of each of the screens, including the settings Form

BeckerWdf:
Can you please elaborate a bit more in details that the issue is?

luapfr:
for you to understand better see this post

https://github.com/dbeaver/dbeaver/issues/20498

BeckerWdf:
So it's basically this issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=563362 ?

luapfr:
that same

What do you say about this problem any solution ?

marcushoepfner:
Just to understand (and in order to avoid that other have to follow several links), this is about Preference Dialog correct?
You request that is shall have
- the same size
- and same position (?)
as last time you have closed it.

Is my understanding correct?

luapfr:
>the same size
>and same position (?)
>as last time you have closed it.


Yes












Just to understand (and in order to avoid that other have to follow several links), this is about Preference Dialog correct?
You request that is shall have

the same size
and same position (?)
as last time you have closed it.

Is my understanding correct?

—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you authored the thread.


-- 
Enviado do Vivaldi Mail. Baixe o Vivaldi de graça em vivaldi.com

luapfr:
Not only preferences all screens should be saved the size and position 


Thankful.




Em Terça-feira, 25 de julho de 2023 05:04:19 (-03:00), marcushoepfner escreveu:



Just to understand (and in order to avoid that other have to follow several links), this is about Preference Dialog correct?
You request that is shall have

the same size
and same position (?)
as last time you have closed it.

Is my understanding correct?

—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you authored the thread.


-- 
Enviado do Vivaldi Mail. Baixe o Vivaldi de graça em vivaldi.com

marcushoepfner:
What do you mean by all screens?
Sorry, but if you do not provide concrete examples and always only throw a few words in, we will not really making progress here.

I agree on the preference dialog, but what else is missing?
Please provide screen shots and explanation what is missing and what you expect.

This is an open source project. Some maintainers are even maintaining this in their free time and I find it not fair, to just throw a few word. Please give detailed descriptions.

[...]

PR #1283, opened 2023-11-07
Store wizard dialog bounds.

marcushoepfner:
WizardDialog asks the actual wizard for dialog bounds.

Issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/949

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
@@ -32,7 +32,9 @@ import org.eclipse.core.runtime.ListenerList;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.jface.dialogs.ControlEnableState;
+import org.eclipse.jface.dialogs.DialogSettings;
 import org.eclipse.jface.dialogs.IDialogConstants;
+import org.eclipse.jface.dialogs.IDialogSettings;
 import org.eclipse.jface.dialogs.IMessageProvider;
 import org.eclipse.jface.dialogs.IPageChangeProvider;
 import org.eclipse.jface.dialogs.IPageChangedListener;
@@ -1565,5 +1567,15 @@ public class WizardDialog extends TitleAreaDialog implements IWizardContainer2, 
 				}
 			});
 		}
+	}
+
+	@Override
+	protected IDialogSettings getDialogBoundsSettings() {
+		String name = getClass().getSimpleName() + ".dialogBounds"; //$NON-NLS-1$
+		IDialogSettings dialogSettings = getWizard().getDialogSettings();
+		if (dialogSettings == null) {
+			return null;
+		}
+		return DialogSettings.getOrCreateSection(dialogSettings, name);
 	}
 }


BeckerWdf (approved):

merks:
Given IWizard.getDialogSettings is documented like this: 
```
	/**
	 * Returns the dialog settings for this wizard.
	 * <p>
	 * The dialog store is used to record state between
	 * wizard invocations (for example, radio button selections,
	 * last directory, etc.).
	 * </p>
	 *
	 * @return the dialog settings, or <code>null</code> if none
	 */
	IDialogSettings getDialogSettings();
```
I think this code can NPE.  I'm also not sure one should decide on behalf of all subclasses of this framework class they they should necessarily store their bounds whenever possible...

BeckerWdf:
> I'm also not sure one should decide on behalf of all subclasses of this framework class they they should necessarily store their bounds whenever possible...

Why do you think that. Isn't storing this always a good thing? Could you think of a use-case where a wizard dialog should not remember it's size and position?

bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
```diff
@@ -1572,4 +1574,10 @@ public void run() {
 			});
 		}
 	}
+
+	@Override
+	protected IDialogSettings getDialogBoundsSettings() {
+		IDialogSettings dialogSettings = getWizard().getDialogSettings();
+		return DialogSettings.getOrCreateSection(dialogSettings, getClass().getSimpleName());
```
laeubi: You need to check for null here, also choosing a section based on the classname might clash and one better choose something that appends a key (e.g. 
```suggestion
		return DialogSettings.getOrCreateSection(dialogSettings, getClass().getSimpleName()+".dialogBounds");
```

merks:
I thought maybe things might open on a different monitor rather than on the same monitor as where the window is currently, but testing that, it seems not to be the case.  So yes, it's hard to imagine not wanting the information to be remembered.

laeubi:
What eclipse definitely lacks is a way to manage/reset such stored items, or  haven't found out yet how to do so...

bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
marcushoepfner: done

bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
```diff
@@ -1566,4 +1568,11 @@ public void run() {
 			});
 		}
 	}
+
+	@Override
+	protected IDialogSettings getDialogBoundsSettings() {
+		String name = getClass().getSimpleName() + ".dialogBounds"; //$NON-NLS-1$
+		IDialogSettings dialogSettings = getWizard().getDialogSettings();
```
BeckerWdf, line 1575: dialogSettings might be null

marcushoepfner: Oh damn yes. Saw all the failing tests now. Fixed with next push

BeckerWdf (approved):

BeckerWdf:
test errors are runrelated. See: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005

iloveeclipse:
@marcushoepfner : this change seem to introduce a regression in JDT, see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1082.

Can you please investigate?

PR #1288, opened 2023-11-07
Fix random failing PositionTrackerTest #882

### LINKED ISSUE #882: org.eclipse.search.tests.filesearch.PositionTrackerTest fails randomly
In the Build [I20220706-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220706-1800/testresults/html/org.eclipse.search.tests_ep425I-unit-macM1-java17_macosx.cocoa.aarch64_17.html), the below tests in org.eclipse.search.tests.filesearch.PositionTrackerTest failed on macM1

- testInsertAt0
```
java.lang.AssertionError
at org.junit.Assert.assertNotNull(Assert.java:713)
at org.junit.Assert.assertNotNull(Assert.java:723)
at org.eclipse.search.tests.filesearch.PositionTrackerTest.checkInsertAtZero(PositionTrackerTest.java:130)
at org.eclipse.search.tests.filesearch.PositionTrackerTest.testInsertAt0(PositionTrackerTest.java:68) 
```
- testInsertInsideMatch
```
java.lang.AssertionError
at org.junit.Assert.assertNotNull(Assert.java:713)
at org.junit.Assert.assertNotNull(Assert.java:723)
at org.eclipse.search.tests.filesearch.PositionTrackerTest.checkInsertInsideMatch(PositionTrackerTest.java:99)
at org.eclipse.search.tests.filesearch.PositionTrackerTest.testInsertInsideMatch(PositionTrackerTest.java:82) 
```

HeikoKlare:
The test cases in `PositionTrackerTest` are randomly failing because the text editor part initializes a `DocumentLineDiffer` that concurrently modifies the document, thus affecting the matches that are validated via ssertions in these tests.
    
This change disables some properties that result in the `DocumentLineDiffer` not being instantiated during the test execution. It also reenables one of the test cases for MacOS, which should be fixed by this change.
In addition, the change improves the failing assertion, which is a null check that currently gives no context information. This change improves the error message in case the assertion fails.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/882

### DIFF
--- a/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/PositionTrackerTest.java
+++ b/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/PositionTrackerTest.java
@@ -19,11 +19,13 @@ import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assume.assumeFalse;
 
 import org.junit.Before;
 import org.junit.ClassRule;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
 
 import org.eclipse.core.runtime.jobs.Job;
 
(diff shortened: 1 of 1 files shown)


HeikoKlare:
@iloveeclipse Since you have already analyzed this test failure and found the reasons (see [here](https://github.com/eclipse-platform/eclipse.platform.ui/issues/882#issuecomment-1651233956)): 

> Hmm, it is actually worse. The two are working on different document instances that are created on same file. Even worse, one is org.eclipse.jface.text.Document (not synchronized) and another one is org.eclipse.core.internal.filebuffers.SynchronizableDocument.

Do you think the approach in this PR is reasonable? It deactives two properties, such that a text editor does not open a second document instance anymore. In consequence, this avoids a race condition between the two documents applying changes and the test should hopefully run deterministically.
I know this solution is far from perfect, both in terms of creating such a fitting environment for the test as well as in terms of relying on modifications of specific properties to create such an environment. But I see this as the best option to achieve reliable and stable tests that validate the behavior of `PositionTracker` in isolation.

HeikoKlare:
Closing as this should be obsolete due to #1428

PR #1290, opened 2023-11-07
Remove cancel dialog during Eclipse shutdown

### LINKED ISSUE #1269: Closing Eclipse IDE shows flashing popup if closing is really fast
These days Eclipse closes really fast for me, I assume due to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1084 and other performance enhancements.

It looks like we still show the "Shutting down" dialog, at least a dialog flashes up, but it vanishes so fast again, that I cannot read it. Do avoid such flashing dialogs it would be nicer, if we only show the dialog if shutting down takes longer than for example 1000 ms.

@HeikoKlare WDYT?

vogella:
Removes the dialog during Eclipse shutdown which only flashes up in my Eclipse SDK build without any option to read it or interact with it.

The currently dialog behavior is also highly inconsistent or non-functional.

The cancel dialog during Eclipse shutdown, does not allow to cancel the shutdown, it allow affects the history pruning in the hard-code assumption that pruning happens in  (Math.abs(total - 4.0)).

Fixes #1269

### DIFF
--- a/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
+++ b/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
@@ -17,7 +17,6 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.ide.application;
 
-import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.text.Collator;
 import java.util.ArrayList;
@@ -41,8 +40,8 @@ import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Platform;
-import org.eclipse.core.runtime.ProgressMonitorWrapper;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.ISchedulingRule;
 import org.eclipse.core.runtime.jobs.Job;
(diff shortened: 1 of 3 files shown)


HeikoKlare:
I will have a look once we have agreed on the solution concept in #1269 to not have separate discussions until we have agreed on an approach.

basilevs:
Is not this the only way to resolve a live lock between hanged jobs? I remember this dialog saving me a couple of times, because each job has its own cancel button (red square) and plugins often fight each other for workspace (and other) resources. I could stop less important jobs and allow more important jobs (like saving resources) to complete without waiting for their opponents.

HeikoKlare:
> Is not this the only way to resolve a live lock between hanged jobs? I remember this dialog saving me a couple of times, because each job has its own cancel button (red square) and plugins often fight each other for workspace (and other) resources. I could stop less important jobs and allow more important jobs (like saving resources) to complete without waiting for their opponents.

This is only about a progress dialog for the save operation performed during workbench shutdown, so it will not affect any other job progress dialog shown. @basilebs You probably refer to progress dialogs popping up while using the IDE and not during shutdown, right?

basilevs:
> You probably refer to progress dialogs popping up while using the IDE and not during shutdown, right?

Nope, a dialog with a list of jobs pops up on shutdown, if jobs are long running.

HeikoKlare:
I see, but then you probably refer to the "ordinary" progress dialog that shows all the jobs that are still running during shutdown? The one in this PR is really only a single progress bar for the save operation in which you can currently cancel one subtask of the save operation. It does not allow to cancel any other jobs.

basilevs:
> The one in this PR is really only a single progress bar for the save operation in which you can currently cancel one subtask of the save operation

As long as there is only a single progress bar, it is indeed different from the one I've thought about. Thanks and sorry for the confusion.

HeikoKlare:
Thanks for the feedback, then the usecase you mention should not be affected. And no problem, thanks for asking! Could have also been that we missed some use case of the removed functionality.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
```diff
@@ -539,31 +468,9 @@ protected void disconnectFromWorkspace() {
...
-			status
-					.merge(new Status(IStatus.ERROR,
-							IDEWorkbenchPlugin.IDE_WORKBENCH, 1,
-							IDEWorkbenchMessages.InternalError, e
-									.getTargetException()));
-		} catch (InterruptedException e) {
+		} catch (CoreException e) {
 			status.merge(new Status(IStatus.ERROR,
 					IDEWorkbenchPlugin.IDE_WORKBENCH, 1,
 					IDEWorkbenchMessages.InternalError, e));
```
HeikoKlare, line 476: As mentioned [here](https://github.com/eclipse-platform/eclipse.platform.ui/issues/1269#issuecomment-1801276486), I would propose to execute the save operation in a `busyCursorWhile` of the `ProgressManager` , so that the UI does not freeze in case saving takes long. Instead a busy cursor will be shown first and when it takes a "long time" (currently more than 800ms), it will open a progress dialog.

```java
ProgressManager.getInstance().busyCursorWhile(monitor -> {
	try {
		status.merge(((Workspace) ResourcesPlugin.getWorkspace()).save(true, true, monitor));
	} catch (CoreException e) {
		status.merge(new Status(IStatus.ERROR,
					IDEWorkbenchPlugin.IDE_WORKBENCH, 1,
					IDEWorkbenchMessages.InternalError, e));
	}
});
```

HeikoKlare (commented):
I like this simplification and the removal of the broken opportunity to cancel the history cleaning step. I would only propose a change to avoid a freezing UI in case the save operation takes rather long by using a busy cursor, which automatically opens a progress dialog in case the operation takes too long.

HeikoKlare:
@vogella just out of curiousity: why did you close your PR? Wouldn't reducing complexity by removing some unnecessary code have been good?

vogella:
> @vogella just out of curiousity: why did you close your PR? Wouldn't reducing complexity by removing some unnecessary code have been good?

I agree that reducing complexity would been good, but currently I do not work on Eclipse code and I assume nobody else would continue this work. But if you are interested in reviewing this effort that would be great. Either restore this PR and update it, or simply create a new one.

Issue #1261, opened 2023-10-30
Second IElementUpdater is not called


basilevs:
While investigating failure of  `org.eclipse.ui.tests.commands.CommandEnablementTest.testRestoreContributedUI()` I've found that `org.eclipse.ui.internal.WorkbenchHandlerServiceHandler.updateElement(UIElement, Map)` queries local context, but does not track it.
When handlers are changed due to a change of activation conditions, neither `ICommandService` nor `WorkbenchHandlerServiceHandler` are notified about the fact. This results in loss of menu/toolbar updates from subsequent handlers implementing `IElementUpdater`.

Here is a part of the test demonstrating the problem (after fixing irrelevant issues):
```
	@Test
	public void testRestoreContributedUI() throws Exception {

		Field iconField = CommandContributionItem.class.getDeclaredField("icon");
		iconField.setAccessible(true);

		Field labelField = CommandContributionItem.class.getDeclaredField("label");
		labelField.setAccessible(true);

		String menuId = "org.eclipse.ui.tests.Bug275126";
		IWorkbenchWindow window = UITestCase.openTestWindow();
		MenuManager manager = new MenuManager("testRestoreContributedUI", menuId);
		manager.createContextMenu(window.getShell());
		IMenuService menuService = fWorkbench.getService(IMenuService.class);
		menuService.populateContributionManager(manager, MenuUtil.menuUri(menuId));

		String text1 = "text1";
		String text2 = "text2";

		// contributed from plugin.xml
		String contributedLabel = "Contributed Label";

		MenuItem menuItem = getTheOnlyItem(manager);

		// default handler
		assertEquals(contributedLabel, menuItem.getText());
		assertNotNull(menuItem.getImage());

		UpdatingHandler handler1 = new UpdatingHandler(text1);
		activation1 = handlerService.activateHandler(CMD3_ID, handler1, new ActiveContextExpression(CONTEXT_TEST1,
				new String[] { ISources.ACTIVE_CONTEXT_NAME }));
		UpdatingHandler handler2 = new UpdatingHandler(text2);
		activation2 = handlerService.activateHandler(CMD3_ID, handler2, new ActiveContextExpression(CONTEXT_TEST2,
				new String[] { ISources.ACTIVE_CONTEXT_NAME }));

		contextActivation1 = contextService.activateContext(CONTEXT_TEST1);
		UITestCase.processEvents();
		menuItem = getTheOnlyItem(manager);
[...]

PR #1291, opened 2023-11-08
Fix handler leak #925

### LINKED ISSUE #925: Command handler state is not observed
This is discussed in [Bug 549802](https://bugs.eclipse.org/bugs/show_bug.cgi?id=549802)

Corresponding tests are disabled:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/50553ea8426afdc390e14676abf8906c44d184a4/tests/org.eclipse.ui.tests/Eclipse%20UI%20Tests/org/eclipse/ui/tests/commands/StateTest.java#L43
https://github.com/eclipse-platform/eclipse.platform.ui/blob/50553ea8426afdc390e14676abf8906c44d184a4/tests/org.eclipse.ui.tests/Eclipse%20UI%20Tests/org/eclipse/ui/tests/commands/ToggleStateTest.java#L45

Commands try to subscribe to handler state in `org.eclipse.core.commands.Command.addState(String, State)` and `org.eclipse.core.commands.Command.setHandler(IHandler)`, but handler proxies fail to propagate `org.eclipse.core.commands.IObjectWithState` protocol.

Offending proxies:

- org.eclipse.ui.internal.WorkbenchHandlerServiceHandler
- org.eclipse.ui.internal.handlers.E4HandlerProxy

While `org.eclipse.ui.internal.handlers.HandlerProxy.handleStateChange(State, Object)` does propagate the states to the actual handler, it is never called as upstream proxies do not invoke `HandlerProxy.addState()`

[Attached is a simple plugin with a stateful command handler](https://github.com/eclipse-platform/eclipse.platform.ui/files/11995015/handler_state_fail.zip), events from which fail to propagate to command framework.
The menu item is "File/I should blink". It does not change state even if menu is reopened.

### Workaround
if command id is known, update command state through service locator:
```
	@Override
	public void setEnabled(Object evaluationContext) {
		super.setEnabled(evaluationContext);
		boolean isRegisteredIdle;
		try {
			Stream<JavaThread> threads = getThreads(evaluationContext);
			isRegisteredIdle = threads.parallel()
					.anyMatch(Activator.getDefault()::isIdle);
		} catch (ExecutionException e) {
			LOG.error("Failed to process selected threads", e);
			isRegisteredIdle = false;
		}
		
		IWorkbenchSite site;
		try {
			site = HandlerUtil.getActiveSiteChecked(adapt(evaluationContext));
		} catch (ExecutionException e) {
			throw new IllegalStateException(e);
		}
		State state = site.getService(ICommandService.class).getCommand("org.basilevs.jstackfilter.eclipse.ui.markidle").getState(RegistryToggleState.STATE_ID);
		state.setValue(isRegisteredIdle);
	}
```

basilevs:
Fixes #925
Fixes a regression introduced in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1253

Handlers can store references to external objects so cleaning them up when another handler is activated is not quick enough. WeakReference is enough to detect handler switch, so we use that.

The main problem with this fix is the loss of side-effects on cleanup of old instances of `IObjectWithState` as `removeState()` is never called for them.

### DIFF
--- a/bundles/org.eclipse.core.commands/src/org/eclipse/core/commands/IObjectWithState.java
+++ b/bundles/org.eclipse.core.commands/src/org/eclipse/core/commands/IObjectWithState.java
@@ -31,12 +31,18 @@ public interface IObjectWithState {
 	/**
 	 * Adds state to this object.
 	 *
-	 * @param id
-	 *            The identifier indicating the type of state being added; must
-	 *            not be <code>null</code>.
-	 * @param state
-	 *            The new state to add to this object; must not be
-	 *            <code>null</code>.
+	 * Usually, consumers will call <code>removeState()</code> with a matching id in
+	 * the same lifecycle of IObjectWithState. However, this behavior is not
+	 * guaranteed. Implementors should not rely on <code>removeState()</code> for
+	 * resource disposal. The recommended practice is to free resources associated
+	 * with non-removed states in some kind of dispose() method.
+	 *
+	 * @see AbstractHandlerWithState
+	 *
+	 * @param id    The identifier indicating the type of state being added; must
+	 *              not be <code>null</code>.
+	 * @param state The new state to add to this object; must not be
+	 *              <code>null</code>.
 	 */
 	public void addState(String id, State state);
 
--- a/bundles/org.eclipse.e4.core.commands/src/org/eclipse/e4/core/commands/internal/HandlerServiceHandler.java
+++ b/bundles/org.eclipse.e4.core.commands/src/org/eclipse/e4/core/commands/internal/HandlerServiceHandler.java
@@ -14,6 +14,7 @@
 
 package org.eclipse.e4.core.commands.internal;
 
+import java.lang.ref.WeakReference;
 import org.eclipse.core.commands.AbstractHandlerWithState;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
@@ -43,7 +44,7 @@ public class HandlerServiceHandler extends AbstractHandlerWithState {
 
 	protected final String commandId;
 	// Remove state from currentStateHandler when it goes out of scope
-	private IObjectWithState currentStateHandler;
+	private WeakReference<IObjectWithState> currentStateHandler = new WeakReference<IObjectWithState>(null);
 
 	public HandlerServiceHandler(String commandId) {
 		this.commandId = commandId;
@@ -250,11 +251,12 @@ public class HandlerServiceHandler extends AbstractHandlerWithState {
 		} else {
 			typed = null;
 		}
-		IObjectWithState oldHandler = currentStateHandler;
+
+		IObjectWithState oldHandler = currentStateHandler.get();
 		if (oldHandler == typed) {
 			return;
 		}
-		currentStateHandler = typed;
+		currentStateHandler = new WeakReference<>(typed);
 		for (String id : getStateIds()) {
 			if (oldHandler != null) {
 				oldHandler.removeState(id);


laeubi:
> The main problem with this fix is the loss of side-effects on cleanup of old instances of IObjectWithState as removeState() is never called for them.

Is this relevant if the (referenced) object is garbage collected?

basilevs:
> > The main problem with this fix is the loss of side-effects on cleanup of old instances of IObjectWithState as removeState() is never called for them.

> Is this relevant if the (referenced) object is garbage collected?

Yes, as side effects may be important.

laeubi:
> Yes, as side effects may be important.

Can you explain that? Maybe `IObjectWithState` should have some kind of way to register a `DisposeListener` then instead?

basilevs:
> > Yes, as side effects may be important.
> Can you explain that? Maybe `IObjectWithState` should have some kind of way to register a `DisposeListener` then instead?

When State is added, implementor of IObjectWithState may opt to subscribe to another event source, or otherwise allocate resources. The contract of IObjectWithState is that every `addState()` is paired with `removeState()` and listeners and resource allocation have a chance to be reverted in `removeState()`. I've fixed `AbstractHandlerWithState` to handle disposal gracefully (calling `removeState()` if Command Framework have not already done so, but in general such care is not implied by `IObjectWithState` contract (as can be seen in original implementation of `AbstractHandlerWithState` which did not care).

iloveeclipse:
I haven't followed what about original issue is, but using weak reference for something that is very important for that feature would mean, after a "gc" this feature may not work as expected ?

basilevs:
> I haven't followed what about original issue is, but using weak reference for something that is very important for that feature would mean, after a "gc" this feature may not work as expected ?

Yes, there is a corner case, when `IObjectWithState` may allocate resources in `addState()` and fail to dispose them as `removeState()` is never called. Usually the disposal should be done in `dispose()` so it should not happen often.

The original idea behind the feature will work as intended, as handlers that are collected are usually no longer interested in command state.

laeubi:
> Yes, there is a corner case, when IObjectWithState may allocate resources and fail to dispose them in dispose() method, as removeState() is never called.

I think one can assume that dispose has to remove every remaining resources. The paring is only required if the regitrar object itself is removed and one don't want to leak there.

@basilevs please (as a follow up) enhance the javadoc of `IObjectWithState` that disposing has to cleanup any remaining `addState(...)` if you think there is any doubt and at best check all implementation of `IObjectWithState` in the platform.

iloveeclipse:
Have you tried your fix with jdt leak tests - are they "green" now?

basilevs:
> Have you tried your fix with jdt leak tests - are they "green" now?

Yes, the were red before the fix and became green after. They were hard to launch on Java 17 as many `--add-opens` were missing from pom file (not sure how that works in CI), but I managed.

Missing VM arguments were:

```
--add-opens jdk.management.jfr/jdk.management.jfr=ALL-UNNAMED
--add-opens jdk.jfr/jdk.jfr.internal.management=ALL-UNNAMED
--add-opens jdk.crypto.ec/sun.security.ec=ALL-UNNAMED
```

basilevs:
@laeubi 
> I think one can assume that dispose has to remove every remaining resources. The paring is only required if the regitrar object itself is removed and one don't want to leak there.

Unfortunately, there is no `dispose()` in `IObjectWithState`. In particular `org.eclipse.core.commands.Command` also implements `IObjectWithState` but should not have `dispose()` method or its equivalent.

> @basilevs please (as a follow up) enhance the javadoc of `IObjectWithState` that disposing has to cleanup any remaining `addState(...)` if you think there is any doubt and at best check all implementation of `IObjectWithState` in the platform.

I will add some vague wording.  There are no "working" implementations of IObjectWithState besides the ones I touched.

```java
        /**
	 * Adds state to this object.
	 *
	 * Usually, consumers will call <code>removeState()</code> with a matching id in
	 * the same lifecycle of IObjectWithState. However, this behavior is not
	 * guaranteed. Implementors should not rely on <code>removeState()</code> for
	 * resource disposal. The recommended practice is to free resources associated
	 * with non-removed states in some kind of dispose() method.
	 *
	 * @see AbstractHandlerWithState
	 *
	 * @param id    The identifier indicating the type of state being added; must
	 *              not be <code>null</code>.
	 * @param state The new state to add to this object; must not be
	 *              <code>null</code>.
	 */
	public void addState(String id, State state);
```

basilevs:
I have considered to track handler removals with `IEclipseContext.runAndTrack()` like suggested in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1261 but as there are multiple contexts involved, I'm afraid this would cause listener leak.

iloveeclipse:
Let's merge that for the next build. Any objections?

laeubi:
> Let's merge that for the next build. Any objections?

go for it...

iloveeclipse:
@basilevs : thanks for a fast workaround.

PR #1287, opened 2023-11-07
Remove blank Javadoc

jukzi:
This commit cleans up Javadoc that does not add information. It resolves ecj warnings:
 `Javadoc: Description expected after ...`
It helps to prevent future empty javadoc by disabling missingJavaDoc warnings. This resolves
 `Javadoc: Missing ...`

The modification is a result of regular expression search&replace:

in files `*.java`
 `^[\s]*\*[\s]*(@return|@param[\s]*[^\s]+|@throws[\s]*[^\s]+)\R([\s]*\*[\s]*@|[\s]*\*/\R)`
 ->`$2`
 `^([\s]*\*[\s]*\R)([\s]*\*/\R)`
 ->`$2`
 `^[\S\t ]*/\*\*\R[\s]*\*/\R`
 ->``

in files `org.eclipse.jdt.core.prefs`
 `org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc(Comments|Tags)\=[^\s]*`
 ->`org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc$1\=ignore`

PR #1295, opened 2023-11-09
various javadoc fixes

jukzi:
reduce ecj warnings in workspace


jukzi:
has to wait for https://github.com/eclipse-platform/eclipse.platform.ui/pull/1287 due to conflicts

jukzi:
@akurtakov  @iloveeclipse ci failed with `Java heap space` on this small change

iloveeclipse:
> @akurtakov @iloveeclipse ci failed with `Java heap space` on this small change

I would propose to create a ticket and discuss there possible solutions.

jukzi:
Did it fail because the today changes in sum are as worse a a single big change and we will have to wait for another i-build so that this repo's build will work again?

akurtakov:
> Did it fail because the today changes in sum are as worse a a single big change and we will have to wait for another i-build so that this repo's build will work again?

It is actually quite likely as all these bundle changes will be compared as long as they are not in the I-builds repo.

jukzi:
unrelated test errors on linux

PR #1299, opened 2023-11-10
Disable some org.eclipse.jdt.core.compiler.problem

jukzi:
To reduce problem markers in workspace.

autoboxing=ignore
missingDefaultCase=ignore
rawTypeReference=ignore
undocumentedEmptyBlock=ignore
uncheckedTypeOperation=ignore

Those are no real problems and ignored by developers for years.

Only for projects that actually have shown many of those markers. For projects with few markers "@SuppressWarnings" was added instead.


jukzi:
ignoring unrelated failing test

PR #1301, opened 2023-11-10
Initialize cause of some InterruptedExceptions

fedejeanne:
Pack an `OperationCanceledException` inside the `InterruptedException` when this is thrown due to the user canceling an operation. This makes it easier to determine the cause of the `InterruptedException` from outside the methods (i.e. in the catch-clauses). Adapt the method `ModalContext::checkCanceled` accordingly. Let `WorkspaceModifyOperation::run` use the original `OperationCanceledException` if provided and only create a new one as a fallback.

Same idea as https://github.com/eclipse-platform/eclipse.platform.ui/pull/1259

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
@@ -14,6 +14,7 @@
 package org.eclipse.ui.actions;
 
 import java.lang.reflect.InvocationTargetException;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IWorkspaceRunnable;
@@ -103,20 +104,19 @@ public abstract class WorkspaceModifyOperation implements IRunnableWithProgress,
 	@Override
 	public synchronized final void run(IProgressMonitor monitor)
 			throws InvocationTargetException, InterruptedException {
-		final InvocationTargetException[] iteHolder = new InvocationTargetException[1];
+		AtomicReference<InvocationTargetException> rethrownInvocationTargetException = new AtomicReference<>();
+		AtomicReference<InterruptedException> rethrownInterruptedException = new AtomicReference<>();
 		try {
 			IWorkspaceRunnable workspaceRunnable = pm -> {
 				try {
 					execute(pm);
 				} catch (InvocationTargetException e1) {
-					// Pass it outside the workspace runnable
-					iteHolder[0] = e1;
+					rethrownInvocationTargetException.set(e1);
 				} catch (InterruptedException e2) {
-					// Re-throw as OperationCanceledException, which will be
-					// caught and re-thrown as InterruptedException below.
-					throw new OperationCanceledException(e2.getMessage());
+					rethrownInterruptedException.set(e2);
 				}
-				// CoreException and OperationCanceledException are propagated
+				// CoreException and unchecked exceptions (e.g. OperationCanceledException) are
+				// propagated to the outer catch
 			};
 			// if we are in the UI thread, make sure we use progress monitor
 			// that spins event loop to allow processing of pending asyncExecs
@@ -137,9 +137,13 @@ public abstract class WorkspaceModifyOperation implements IRunnableWithProgress,
 			interruptedException.initCause(e);
 			throw interruptedException;
 		}
-		// Re-throw the InvocationTargetException, if any occurred
-		if (iteHolder[0] != null) {
-			throw iteHolder[0];
+
+		// Re-throw any exceptions caught while running the IWorkspaceRunnable
+		if (rethrownInvocationTargetException.get() != null) {
+			throw rethrownInvocationTargetException.get();
+		}
+		if (rethrownInterruptedException.get() != null) {
+			throw rethrownInterruptedException.get();
 		}
 	}
 	@Override


fedejeanne:
The check failures don't seem to be related to my changes since I haven't changed any API.

iloveeclipse:
it's tycho problem, see eclipse-tycho/tycho#3019

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
```diff
@@ -114,6 +114,13 @@ public synchronized final void run(IProgressMonitor monitor)
 				} catch (InterruptedException e2) {
 					// Re-throw as OperationCanceledException, which will be
 					// caught and re-thrown as InterruptedException below.
+
+					// This is the best case scenario because the original exception has the exact
+					// line where the cancelation happened (better for debugging)
+					if (e2.getCause() instanceof OperationCanceledException oce)
+						throw oce;
+
+					// fallback: create a new exception
```
HeikoKlare: Why has this exception to be repackaged outside the workspace runnable at all? The current handling is broken, as the catch block for the execution of the workspace runnable adds _every_ `OperationCanceledException` as cause to an `InterruptedException` although the `OperationCanceledException` is only used as temporary non-checked exception wrapped (see comment in lines 115--116), so this is not suppoed to be passed to the method consumer.

From my understand, what this code is simply trying to do is to pass the `InterruptedException` outside of the runnable and throw it there. So the handling could be simply done like for the invocation target exception in line 113.

HeikoKlare (commented):
I agree that the existing exception handling in `WorkspaceModifyOperation` is not correct, but the fix does seem to solve the problem completey (see my detailed comment).

I do not understand the reason for all the other changes, like I have argued for the according changes to the platform repository [here](https://github.com/eclipse-platform/eclipse.platform/pull/853#pullrequestreview-1726718911).

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
fedejeanne: > From my understand, what this code is simply trying to do is to pass the InterruptedException outside of the runnable and throw it there.

That was my understanding too, the thing is that simply storing the exception in `iteHolder` like it's done in line 113 would not skip the execution of the rest of the code outside the inner `try-catch`-block while throwing an `OperationCanceledException` would.
The objective of this PR is simply adding (in this particular case: **improving**) details to the `InterruptedException`, not to change the logic.

Another option to preserve the logic would be to check if `iteHoldler` holds an `OperationCanceledException` and, if it does, skip processing all pending _asyncExecs_, but that code wouldn't look much better.

HeikoKlare: > That was my understanding too, the thing is that simply storing the exception in `iteHolder` like it's done in line 113 would not skip the execution of the rest of the code outside the inner `try-catch`-block while throwing an `OperationCanceledException` would.

Can you please detail that? I do not see any further code that would be executed then.

fedejeanne: If line 113 is reached then this piece of code will still be invoked:
```java
// if we are in the UI thread, make sure we use progress monitor
// that spins event loop to allow processing of pending asyncExecs
if (monitor != null && PlatformUI.isWorkbenchRunning()
		&& !PlatformUI.getWorkbench().isStarting()) {
	Display display = PlatformUI.getWorkbench().getDisplay();
	if (!display.isDisposed()
			&& display.getThread() == Thread.currentThread()) {
		monitor = new EventLoopProgressMonitor(monitor);
	}
}
IDEWorkbenchPlugin.getPluginWorkspace().run(workspaceRunnable,
		rule, IResource.NONE, monitor);
```

But if an exception is thrown then that piece of code will not run.

HeikoKlare: That code is always invoked, as line 113 is within a lambda that is triggered in line 138.

fedejeanne: You're right, thanks. That means I can safely use the same mechanism in the whole method. 

Done in b6c33cdd30a6d9eb63686e48173e7322cb17f6cf

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
```diff
@@ -103,20 +103,17 @@ protected abstract void execute(IProgressMonitor monitor)
 	@Override
 	public synchronized final void run(IProgressMonitor monitor)
 			throws InvocationTargetException, InterruptedException {
-		final InvocationTargetException[] iteHolder = new InvocationTargetException[1];
+		final Exception[] exceptionHolder = new Exception[1];
```
HeikoKlare: Can you please change this into `AtomicReference` or similiar? Misusing arrays for passing references is a bad pattern, both in terms of comprehensibility as well because it's not thread-safe (no problem here, but bad example for other places).

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
```diff
@@ -103,20 +103,17 @@ protected abstract void execute(IProgressMonitor monitor)
...
 		try {
 			IWorkspaceRunnable workspaceRunnable = pm -> {
 				try {
 					execute(pm);
-				} catch (InvocationTargetException e1) {
+				} catch (InvocationTargetException | InterruptedException | CoreException e1) {
```
HeikoKlare: Why is `CoreException` caught now? In fact, the changed code completely swallows a `CoreException`. An  IWorkspaceRunnable` is allowed to throw a `CoreException` and the containing code already properly handles that.

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
```diff
@@ -103,20 +103,17 @@ protected abstract void execute(IProgressMonitor monitor)
...
-					iteHolder[0] = e1;
-				} catch (InterruptedException e2) {
-					// Re-throw as OperationCanceledException, which will be
-					// caught and re-thrown as InterruptedException below.
-					throw new OperationCanceledException(e2.getMessage());
+					exceptionHolder[0] = (e1 instanceof CoreException ce) //
+							? new InvocationTargetException(ce) //
+							: e1;
 				}
-				// CoreException and OperationCanceledException are propagated
```
HeikoKlare, line 119: I think this comment is rather useful to understand the outside catch, so it should stay.

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
```diff
@@ -137,9 +134,12 @@ public synchronized final void run(IProgressMonitor monitor)
...
-		// Re-throw the InvocationTargetException, if any occurred
-		if (iteHolder[0] != null) {
-			throw iteHolder[0];
+
+		// Re-throw the InvocationTargetException, if any occurred inside the IWorkspaceRunnable
+		if (exceptionHolder[0] instanceof InvocationTargetException ite) {
+			throw ite;
+		} else if (exceptionHolder[0] instanceof InterruptedException ie) {
+			throw ie;
 		}
```
HeikoKlare: No need to distinguish the kind of exception, as long as it's not null.

fedejeanne: How would you write it?

HeikoKlare: Now I see, the problem is that the `exceptionHolder` was generalized to `Exception` and the method declares specific types of exception. But then I would propose to store the typed exceptions separately, i.e., in different variables, rather than re-extracting them using `instanceof`.

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
fedejeanne: Done in 12615e9efe647076b42e23ad1130bd965296e783

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
fedejeanne: Good point, this is unnecessary. 

Done in 12615e9efe647076b42e23ad1130bd965296e783

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
fedejeanne: Done in 12615e9efe647076b42e23ad1130bd965296e783

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/WorkspaceModifyOperation.java
fedejeanne: Thank you for clarifying it but I'd rather keep it as it is now: adding more variables only adds complexity to the method and it doesn't really improve readability.

HeikoKlare: Your choice to select some other solution, but the current proposal contains an incomplete instanceof switch, which will eventually break.

fedejeanne: Then I must have misunderstood your proposal with the 2 variables because the way I imagined it would have also ended up being an incomplete switch in disguise. 

Could you please provide a snippet?

HeikoKlare: Something like:
```java
		AtomicReference<InvocationTargetException> rethrownInvocationTargetException = new AtomicReference<>();
		AtomicReference<InterruptedException> rethrownInterruptedException = new AtomicReference<>();
		try {
			IWorkspaceRunnable workspaceRunnable = pm -> {
				try {
					execute(pm);
				} catch (InvocationTargetException e1) {
					rethrownInvocationTargetException.set(e1);
				} catch (InterruptedException e2) {
					rethrownInterruptedException.set(e2);
				}
				// CoreException and OperationCanceledException are propagated
			};
			...
		} catch (CoreException e) {
			throw new InvocationTargetException(e);
		} catch (OperationCanceledException e) {
			InterruptedException interruptedException = new InterruptedException(e.getMessage());
			interruptedException.initCause(e);
			throw interruptedException;
		}
		if (rethrownInvocationTargetException.get() != null) {
			throw rethrownInvocationTargetException.get();
		}
		if (rethrownInterruptedException.get() != null) {
			throw rethrownInterruptedException.get();
		}
```

fedejeanne: That's actually the way I imagined it: the 2 `if`-statements at the end are also "a `switch` in disguise".
Or am I missing something?

HeikoKlare: I think I don't get your point. Where do you see a switch? The method throws exactly those kinds of exceptions that you can assign within the runnable. There is no chance the runnable assigns a type of exception that is not rethrown in the method (unlike when having a generic `Exception` variable from which only specific types are extracted).

fedejeanne: What I mean with "a `switch` in disguise" is that when using several variables, you still need to exhaust all possibilities and query all variables for `null`.

If `SomeNewCheckedException` comes along then you still need to add this at the end:

```java
if (someNewCheckedException.get() != null) {
	throw someNewCheckedException.get();
}
```

This is the typical place were one does a "copy-paste mistake" that may end up like...

```java
if (someNewCheckedException.get() != null) {
	throw rethrownInterruptedException.get();
}
```
One ends up having a new variable and a new exclusive `catch`-clause (just more code to maintain) without really solving the problem of having to exhaust every possibility.

HeikoKlare: > If `SomeNewCheckedException` comes along then you still need to add this at the end:

That's exactly the point. You _must_ do this. When just having an Exception-typed variable, you are not forced to add code for another exception, which results in that exception being swallowed. This is a risk and will eventually lead to a bug, like in almost every case where `instanceof` is applied without providing handling for every type.

As mentioned before: You do not necessarily need to use my proposal, just choose any option that ensures that a complete switch is performed on the possible types.

fedejeanne: > When just having an Exception-typed variable, you are not forced to add code for another exception...

Now I get what you mean: you mean that using `<Exception>` as the type is like saying _"This methods already handles every kind of `Exception`"_ (which is not true because only those types that are covered in the `if`-statements are actually handled) while using specific subclasses of `Exception` says _"This method only handles these specific kind of exceptions"_ (which basically says the developer: _"you want it, you add it"_)

Is a bit more code but if it helps to avoid any misinterpretation of the current implementation then it's worth it. 

I'm sold! 💰  --> 5b150d44e71720bbf3a917bdf6bb92cc1912555d

Thank you for clarifying it :)

bundles/org.eclipse.jface/src/org/eclipse/jface/operation/ModalContext.java
```diff
@@ -255,7 +258,9 @@ public static boolean canProgressMonitorBeUsed(IProgressMonitor monitor1, IProgr
 	 */
 	public static void checkCanceled(IProgressMonitor monitor) throws InterruptedException {
 		if (monitor.isCanceled()) {
-			throw new InterruptedException();
+			InterruptedException interruptedException = new InterruptedException();
+			interruptedException.initCause(new OperationCanceledException());
+			throw interruptedException;
```
HeikoKlare: This is semantically incorrect. A monitor may be cancelled for other reasons than the one stated in an `OperationCanceledException`, which is tied to the user canceling an operation. In addition, there is no information gain in adding another exception as a cause to the `InterruptedException`, as no further data is added to the exception and the stack trace will be the same. So it only add overhead due to exception generation.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/deferred/LazySortedCollection.java
```diff
@@ -386,7 +387,9 @@ private final int partition(int subTree, FastProgressReporter mon) throws Interr
 			}
 
 			if (mon.isCanceled()) {
-				throw new InterruptedException();
+				InterruptedException interruptedException = new InterruptedException();
+				interruptedException.initCause(new OperationCanceledException());
+				throw interruptedException;
```
HeikoKlare: Like above: semantically incorrect and unnecessary. The exception is only used for internal control flows of the class and not passed outside through public API, so no need to introduce this overhead.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/FileSystemExportOperation.java
```diff
@@ -227,7 +228,9 @@ protected void exportFile(IFile file, IPath location)
 						.queryOverwrite(properPathString);
 
 				if (overwriteAnswer.equals(IOverwriteQuery.CANCEL)) {
-					throw new InterruptedException();
+					InterruptedException interruptedException = new InterruptedException();
+					interruptedException.initCause(new OperationCanceledException());
+					throw interruptedException;
```
HeikoKlare: Same as above: Cancel may not only occur upon user cancellation. All consumers of the exception (`WizardFileSystemResourceExportPage1`) do not use the exception contents.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/WizardFileSystemResourceImportPage1.java
```diff
@@ -1122,9 +1123,7 @@ public void filterElements(Collection files,
 				}
 				Iterator filesList = files.iterator();
 				while (filesList.hasNext()) {
-					if (monitor.isCanceled()) {
-						throw new InterruptedException();
-					}
+					ModalContext.checkCanceled(monitor);
```
HeikoKlare: This is only okay when not changing the `ModalContext.checkCanceled` implementation, as the exception is only used with this method (line 1170), so no need to add overhead.

fedejeanne:
I reverted some changes, I hope it's looking good now @HeikoKlare ?

HeikoKlare (approved):
Change looks good to me now.

fedejeanne:
Test failures are unrelated: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005, #1337 , #1338

PR #1303, opened 2023-11-13
NLS missing messages in Generic Editor messages

vogella:
GotoMatchingBracket_error_noMatchingBracket in:
org.eclipse.ui.internal.genericeditor.messages
and GotoMatchingBracket_error_bracketOutsideSelectedElement are missing.

First one can be seen if you open the source tab of the target editor.


vogella:
@merks you were faster... :-)

merks:
🚀

PR #1305, opened 2023-11-14
Fix display used for ImageRegistry creation

### LINKED ISSUE #15: Prefer Display From Workbench When Initializing ImageRegistry

HeikoKlare:
The `AbstractUIPlugin`, which is implemented by several UI plugins, creates and provides an  ImageRegistry`. This registry is initialized with the display of the current thread, if the current thread has a display, and with the display of the workbench otherwise. In consequence, if the first access to the ImageRegistry happens from a thread with a (temporary) display that is disposed at some point in time, all images will (unexpectedly) be disposed as well.  This can currently only happen
on Windows, as Linux and MacOS do not allow multiple display.

This change adapts the `ImageRegistry` creation in the `AbstractUIPlugin` so that the display of the workbench is used, if available. The current thread's display is only used if no workbench is running. This results in the `ImageRegistry` being disposed with the workbench instead of any temporary display. On systems that have no support for multiple displays, the change will simply have no effect. It also adds a regression test for the unexpected behavior.

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/WorkbenchPluginTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/WorkbenchPluginTest.java
@@ -1,0 +1,94 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
+package org.eclipse.ui.tests.api;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assume.assumeFalse;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.eclipse.core.runtime.Platform.OS;
+import org.eclipse.core.tests.harness.TestBarrier2;
+import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.jface.resource.ImageRegistry;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.plugin.AbstractUIPlugin;
+import org.junit.Test;
+
+public class WorkbenchPluginTest {
+	/**
+	 * The test is supposed to initialize the image registry of an AbstractUIPlugin
+	 * from another thread. To this end, we use a new plug-in instance to simulate
+	 * the behavior of any specialization of an AbstractUIPlugin (like
+	 * WorkbenchPluginTest) being first accessed from another thread.
+	 */
+	private final AbstractUIPlugin testPlugin = new AbstractUIPlugin() {
+	};
+
+	@Test
+	public void testGetImageRegistryFromAdditionalDisplay() {
+		assumeFalse("multiple displays are not allowed on Linux", OS.isLinux());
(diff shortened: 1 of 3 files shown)


tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/WorkbenchPluginTest.java
```diff
@@ -0,0 +1,85 @@
...
+		Image fileIcon = registry.get(filename);
+		if (fileIcon != null) {
+			return fileIcon;
+		}
+		ImageDescriptor imageDescriptor = PlatformUI.getWorkbench().getEditorRegistry().getImageDescriptor(filename);
+		if (imageDescriptor != null) {
+			registry.put(filename, imageDescriptor);
+		}
+		return registry.get(filename);
+	}
```
fedejeanne: This one is pretty similar to `org.eclipse.ui.ISharedImages.getImage(String)` and it currently only works because the `EditorRegistry` ends up providing a default image (see `org.eclipse.ui.internal.registry.EditorRegistry.getDefaultImage()`). You could directly call `org.eclipse.ui.ISharedImages.getImage(String)` and use the descriptor of the default image:

```suggestion
	private Image getFileIcon() {
		return PlatformUI.getWorkbench().getSharedImages().getImage(ISharedImages.IMG_OBJ_FILE);
	}
```

fedejeanne (changes requested):
Looks good, I only have a minor suggestion to reuse an existing method and search for an existing default image (which ends up happening anyway with the current code).

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/WorkbenchPluginTest.java
HeikoKlare: I think I do not really understand the suggestion. The provided test is a regression test for the issue fixed with the PR. When I change the test code as proposed, the test will succeed, but it will not be a regression test for the issue anymore (it then already succceeds with the existing code).

fedejeanne: You're right, you're testing `AbstractUIPlugin` through `TestPlugin`. In that case I suggest you make the test independent from `TestPlugin` in order to clarify this intent, you could use: `ImageRegistry registry = new AbstractUIPlugin() {}.getImageRegistry();`

fedejeanne (approved):

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/WorkbenchPluginTest.java
HeikoKlare: Yes, that's a good idea. In particular because it is important that the tested plug-in's image registry is initialized from another thread, which may not be guaranteed for the `TestPlugin` in case some other test already accessed it. I have adapter the test in e3af6f0 with one difference to your proposal: the plugin is kept as an instance variable as otherwise every call to an image access will be done on a new plug-in instance (which would then again lead to the test not being a regression test for the fixed issue).

PR #1307, opened 2023-11-15
Restarting workspace should take into account changes to eclipse.ini

### LINKED ISSUE #166: Restarting workspace after feature installation does not take into account changes to eclipse.ini
Take a feature installation that has a p2.inf file as follows:

```
instructions.configure=\
org.eclipse.equinox.p2.touchpoint.eclipse.addJvmArg(jvmArg:--add-opens=java.base/java.security=ALL-UNNAMED);

instructions.unconfigure=\
org.eclipse.equinox.p2.touchpoint.eclipse.removeJvmArg(jvmArg:--add-opens=java.base/java.security=ALL-UNNAMED);
```

When this feature is installed via the UI, eclipse will come up with a dialog asking you to restart the IDE to apply the software update. When you select "Restart Now", the workspace will restart. The expected changes listed above will be physically added to the installation's eclipse.ini file. But when the workspace starts, these specific add-opens will not be active. 

If you then go restart the workbench one more time, the changes are applied to the running VM as expected. 

This would indicate a bug in the p2's software update - restart action where it fails to properly recognize or apply changes to eclipse.ini.

umairsair:
Background:
===========
There are 3 return codes supported by Eclipse
- 0  - Normal exit
- 23 - Restart the eclipse with previous launcher arguments
- 24 - Restart the eclipse with the arguments provided by Eclipse application itself

The above mentioned problem is not with the exit code 23. The reason is that the eclipse launcher restarts itself in this case, hence eclipse.ini is also reloaded. However eclipse launcher does not restart itself in case of code 24 and parses the arguments provided by Eclipse application in shared memory and simply relaunches java, which means that eclipse.ini will not be reloaded.

For restarts, eclipse should be using code 23 but we had to switch to code 24 to fix following issue.
[[WorkbenchLauncher] Restart asks for the workspace to run](https://bugs.eclipse.org/bugs/show_bug.cgi?id=264072)
And then the fix of following built up on it.
[[Workspace chooser] "Older Workspace Version" dialog: "Cancel" button shouldn't close Eclipse but go back to workspace chooser](https://bugs.eclipse.org/bugs/show_bug.cgi?id=538830)

Fix:
====
For code 24, the arguments in shared memory are for launcher, so instead of launcher just parsing the args and relaunching java, it should restart the launcher with appropriate arguments. With relaunching the launcher, eclipse.ini will be reloaded like its done for code 23.

Eclipse launcher is updated to relaunch for code 24 as well. Moreover, there are 3 new launcher arguments introduced, 2 are internal to launcher and one is external to launcher and can be set by eclipse application as a relaunch argument.

- `--launcher.oldUserArgsStart` and `--launcher.oldUserArgsEnd`: The user arguments passed to eclipse launcher for first start are tracked between these arguments during relaunches. E.g., if eclipse was started as `eclipse A B C` where A B C are some arguments to launcher and for relaunch, Eclipse application mentioned `X Y Z` arguments then relaunch command will be `eclipse --launcher.oldUserArgsStart A B C --launcher.oldUserArgsEnd A B C X Y Z`. Moreover, launcher relaunch will ignore all arguments between and including `--launcher.oldUserArgsStart` and `--launcher.oldUserArgsEnd`.
- `--launcher.skipOldUserArgs`: If eclipse application wants to relaunch with provided arguments only and ignore the older user args then it can mention this argument. So for the case where eclipse was started as `eclipse A B C` and Eclipse application mentioned `X Y Z` as relaunch arguments along with `--launcher.skipOldUserArgs` then relaunch command will be
`eclipse --launcher.oldUserArgsStart A B C --launcher.oldUserArgsEnd X Y Z` where launcher relaunch will ignore all arguments between and including `--launcher.oldUserArgsStart` and `--launcher.oldUserArgsEnd`.

For restarts, Eclipse application just mentions the `-data` argument now as relaunch arguments instead of mentioning complete java args. However, user can still mention more arguments for relaunch and Eclipse restart will respect them as well and append the `-data` argument at the end of user arguments.

See also https://github.com/eclipse-equinox/equinox/pull/401
Fixes https://github.com/eclipse-equinox/p2/issues/166

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -139,6 +139,7 @@ import org.eclipse.jface.util.Util;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.window.IShellProvider;
 import org.eclipse.jface.window.Window;
+import org.eclipse.osgi.internal.location.LocationHelper;
 import org.eclipse.osgi.service.runnable.StartupMonitor;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
@@ -290,12 +291,8 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 	public static final String EDITOR_TAG = "Editor"; //$NON-NLS-1$
 
-	private static final String PROP_VM = "eclipse.vm"; //$NON-NLS-1$
-	private static final String PROP_VMARGS = "eclipse.vmargs"; //$NON-NLS-1$
-	private static final String PROP_COMMANDS = "eclipse.commands"; //$NON-NLS-1$
 	public static final String PROP_EXIT_CODE = "eclipse.exitcode"; //$NON-NLS-1$
 	private static final String CMD_DATA = "-data"; //$NON-NLS-1$
-	private static final String CMD_VMARGS = "-vmargs"; //$NON-NLS-1$
 
 	private static final class StartupProgressBundleListener implements SynchronousBundleListener {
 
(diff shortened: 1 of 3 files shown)


umairsair:
@sravanlakkimsetti any comments here or are we good?

sravanlakkimsetti (approved):

PR #1309, opened 2023-11-16
Adding Command Icon To Application Model Editor Tree #1308

### LINKED ISSUE #1308: Command Icon are not aligned to other elements in the Application Model Editor Tree
Currently, Command Icons are not visible in the Tree of the Application Model Editor.
Icons of other Elements, like Parts, are visible in the Tree. 
grafik

It would be nicer if this would look like this:
grafik

N1k145:
Currently, Command Icons are not visible in the Tree of the Application Model Editor.
Icons of other Elements, like Parts, are visible in the Tree. This aligns this behavior.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1308

### DIFF
--- a/tools/bundles/org.eclipse.e4.tools.emf.ui/src/org/eclipse/e4/tools/emf/ui/common/component/AbstractComponentEditor.java
+++ b/tools/bundles/org.eclipse.e4.tools.emf.ui/src/org/eclipse/e4/tools/emf/ui/common/component/AbstractComponentEditor.java
@@ -128,8 +128,7 @@ public abstract class AbstractComponentEditor<M> {
 	}
 
 	protected void setElementId(Object element) {
-		if (getEditor().isAutoCreateElementId() && element instanceof MApplicationElement) {
-			final MApplicationElement el = (MApplicationElement) element;
+		if (getEditor().isAutoCreateElementId() && element instanceof MApplicationElement el) {
 			if (el.getElementId() == null || el.getElementId().trim().length() == 0) {
 				el.setElementId(Util.getDefaultElementId(((EObject) getMaster().getValue()).eResource(), el,
 						getEditor().getProject()));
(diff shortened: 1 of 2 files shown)


BeckerWdf (approved):

vogella:
@N1k145 loosely related question, if I assign an icon to a command and use this in menu / toolbar contribution, should I see the icon in the app? Last time I tried, this did not work, but this is a few weeks ago

N1k145:
@vogella I don't think this is supported, Also not with the E3 Command Icons. For the Menu and Toolbar Contributions, the icons are usually set directly.
But I think that would be a nice addition

vogella:
> I don't think this is supported,

Why would I set the icon for the e4 commands in this case? Just to "document" which icon should be used?

N1k145:
@vogella For the QuickAccess (Ctrl + 3)
That is the only place where the icons are used, that I know of. 
But they behave the same as the E3 Command Icons so when there is another user they appear also there

vogella:
> @vogella For the QuickAccess (Ctrl + 3) That is the only place where the icons are used, that I know of. But they behave the same as the E3 Command Icons so when there is another user they appear also there

Ah, thanks for the info. I was trying in a pure e4 RCP application (in which Quick Access cannot be used) and failed to see a difference. Now I understand the intention of this development. Thanks, very useful for the e4 contributions to teh IDE now that I know about it. :-)

N1k145:
@vogella 
Thanks, but I like your Idea to use the Icon as a fallback when no other icon is defined in a Button or Menu Item, maybe I have a look sometime in the future.

Feel free to merge this PR as I don't have write access

fedejeanne:
It's looking good...

image

The test failure is unrelated, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/882#issuecomment-1857649717.

Should I rebase and merge, @vogella ?

fedejeanne (approved):

vogella:
@fedejeanne sounds good

PR #1312, opened 2023-11-17
Fixes #1311 - move some classes from o.e.jface.text to org.eclipse.text

### LINKED ISSUE #1311: Majority of org.eclipse.jface.text.rules package could live in org.eclipse.text for non-UI consumption
jdt-ls makes use of clones of a handful of classes in the org.eclipse.jface.text.rules package.  If these classes were moved down into org.eclipse.text, they could be more easily consumed by non-UI bundles. 

Since org.eclipse.text already includes many org.eclipse.jface.text.* packages in it, this would not be a large or unexpected departure from current practice.

robstryker:


HannesWell:
Instead of adding more split-packages (which is a major problem if one wants to use Import-Package), what about introducing another bundle for the non UI text parts?

merks:
I have the same concern. The next time the signing certificate updates, we will need to be careful that both bundles hosting the split package are touched if either is touched.  That’s a recurring problem.

laeubi:
I don't think an additional bundle is possible as you need to change the package name, so maybe check if one can simply move *all* classes of package `org.eclipse.jface.text.rules`?

robstryker:
Unfortunately DefaultDamagerRepairer and RuleBasedDamagerRepairer (deprecated) are in the package and have UI elements.

robstryker:
So... I understand the complaint about split packages. I really do. 

However, it appears to me (and I could be wrong here) that the sole purpose of org.eclipse.text bundle was to take the core parts of org.eclipse.jface.text and separate them out. This bundle pair already has 4 packages split between each other, and adding a fifth can't possibly be that detrimental? 

I understand the desire to simply move things to a new bundle, or to a new package name, but the package is public API and so cannot be changed in this way. 

I understand the desire to move all classes to org.eclipse.text instead, but as mentioned above, two of those classes have UI APIs in them. So that's a non-starter as well. 

As far as I can tell, there's one other possibility:  essentially duplicate the classes into org.eclipse.text, and then have the versions in org.eclipse.jface.text simply extend the versions in org.eclipse.text and delete the body of these classes except for necessary constructors etc. This is a viable solution, but it's messy and leads to duplicate class names, which can be annoying. And we still end up with too many class files. It also gets messy at ensuring the pre-existing interfaces maintain the right parameters (ie, the org.eclipse.jface.text versions of the interfaces would need the org.eclipse.jface.text versions of the parameters). 

I honestly would submit that while split packages are generally not to be preferred, in this specific case, where a plugin pair already share 4 split packages, where the two bundles already seem tied to each other, where the alternatives (like moving all classes in the package or duplicating the classes and extending them) are more work and more confusion than is desirable, I would suggest that the split package here is not as bad as it may appear at first glance.

laeubi:
If there are already split packages I agree that it does not make things worse.

HannesWell:
> If there are already split packages I agree that it does not make things worse.

Yes, that's right.
It was more a general remark that it would be good to think about how to resolve it on the long run, especially if we want to use more Import-Package instead of require bundle. Before this PR I wasn't aware that a split package also exists with jface.text/eclipse.text. I only knew about o.e.core.runtime and o.e.equinox.common.

robstryker:
There are alternatives to simply splitting a package from one bundle to another when the classes are all API. You all have listed some.  

1) Try to move the whole package. This is ideal if possible. 

2) For relatively flat class hierarchies and when option 1 doesn't apply, you can create new classes in a new package in the destination bundle, and have the classes in the source bundle simply extend them. This becomes harder, though, when you have fairly robust class hierarchies since java does not support multiple inheritance. It gets very messy after the second level of inheritance and you cannot reduce the set of superclasses so you get boxed in to a specific solution which doesn't really work. 

3) The old way:  make new, and deprecate old. This one is to some extent the worst option because consumers (who likely aren't updating their plugins the way they used to anymore) would need to migrate or re-release eventually. I've been trying to avoid this if at all possible.

mickaelistria:
I see several projects to `Import-Package: org.eclipse.jface.text.rules` ( https://github.com/search?q=org.eclipse.jface.text.rules++path%3A*%2FMANIFEST.MF&type=code ); would those remain working and capable of referencing moved classes (eg `IToken`) with this refactoring (thanks to the `org.eclipse.jface.text` re-exporting `org.eclipse.text) ?

I tried such a project
MANIFEST.MF
```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-SymbolicName: sandboxPlugin
Bundle-Version: 1.0.0.qualifier
Import-Package: org.eclipse.jface.text.rules
Require-Bundle: org.junit,
 junit-jupiter-api
Bundle-RequiredExecutionEnvironment: JavaSE-17
Automatic-Module-Name: sandboxPlugin
```
with simple test class checking I can reference the types
```java
package sandboxplugin;

import static org.junit.jupiter.api.Assertions.*;

import org.eclipse.jface.text.rules.IToken;
import org.eclipse.jface.text.rules.Token;
import org.junit.jupiter.api.Test;

class SomeTest {

	@Test
	void test() {
		IToken token = new Token(0);
		assertEquals(0, token.getData());
	}

}
```
PDE and JUnit Plugin test were happy, so it looks like the Import-Package still resolves to those classes.
I then tried to modify the test to reference the other/untouched half of the split package only
```java
package sandboxplugin;

import static org.junit.jupiter.api.Assertions.assertNotNull;

import org.eclipse.jface.text.rules.DefaultDamagerRepairer;
import org.eclipse.jface.text.rules.RuleBasedPartitionScanner;
import org.junit.jupiter.api.Test;

class SomeTest {

	@Test
	void test() {
		DefaultDamagerRepairer repairer = new DefaultDamagerRepairer(new RuleBasedPartitionScanner());
		assertNotNull(repairer);
	}

}
```
and PDE still likes it, Junit Test plugin launch as well.

So it seems to me that this change has in practice no impact at all on the consumers, as both Require-Bundle and Import-Package leads to complete resolution of the package just like if it were not split.
This change is good to go in, unless someone proves it's open to regressions before we merge.

@robstryker can you please rebase?

mickaelistria (approved):

robstryker:
@mickaelistria  we have done similar pushdowns in the past. It seems that if the package is split, and one plugin requires the other be present, then there is absolutely no harm to pushing the classes down. 

At one point I do recall, for public API, that there was some javadoc issue? Perhaps @akurtakov  might remember where that occurred? 

I've rebased.

robstryker:
Recent build failures seem unrelated. 

```
Results:

Failures: 
  UIComparePreferencesAuto.testCompareViewersPref:57 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 190
  DeprecatedUIPreferencesAuto.testAppearancePref:61 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  DeprecatedUIPreferencesAuto.testDefaultTextEditorPref:67 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  DeprecatedUIPreferencesAuto.testFileEditorsPref:73 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  DeprecatedUIPreferencesAuto.testLocalHistoryPref:79 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  DeprecatedUIPreferencesAuto.testPerspectivesPref:85 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  DeprecatedUIPreferencesAuto.testWorkbenchPref:55 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  UIDialogsAuto.testCopyMoveProject:102 Warning: Button {Use &default location}
	Actual Width -> 147
	Recommended Width -> 149
  UIDialogsAuto.testEditorSelection:132 Warning: Button {&Internal editors}
	Actual Width -> 115
	Recommended Width -> 117
  UIPreferencesAuto.testAppearancePref:63 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  UIPreferencesAuto.testDefaultTextEditorPref:69 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  UIPreferencesAuto.testFileEditorsPref:75 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  UIPreferencesAuto.testLocalHistoryPref:81 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  UIPreferencesAuto.testPerspectivesPref:87 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
  UIPreferencesAuto.testWorkbenchPref:57 Warning: Button {Always r&un in background}
	Actual Width -> 184
	Recommended Width -> 186
Errors: 
  DeprecatedUIWizardsAuto.testNewProjectPage1:260 » SWT Widget is disposed
  DeprecatedUIWizardsAuto.testNewProjectPage2:285 » SWT Widget is disposed
  UIWizardsAuto.testNewProjectPage1:261 » SWT Widget is disposed
  UIWizardsAuto.testNewProjectPage2:285 » SWT Widget is disposed

Tests run: 1576, Failures: 15, Errors: 4, Skipped: 119
```

laeubi:
> as both Require-Bundle and Import-Package leads to complete resolution of the package just like if it were not split.

I just wanted to mention that one needs to take care on Import-Package because i**f you use classes from both splits** in the same class you must use Require Bundle! And if you don't you actually need to qualify the import package to tell what bundle you want it to be choosen, so we should simply proceed in diverging things as much as possible and users of the o.e.jface.text need to use require bundle sadly in this case.

robstryker:
> I just wanted to mention that one needs to take care on Import-Package because i**f you use classes from both splits** in the same class you must use Require Bundle! And if you don't you actually need to qualify the import package to tell what bundle you want it to be choosen, so we should simply proceed in diverging things as much as possible and users of the o.e.jface.text need to use require bundle sadly in this case.

I will keep this in mind in the future. I try to migrate code when possible, either to new packages, or migrate the entire package together. But in cases like this, where there's already a tight-coupling between the two, and an old and very public API that really should not be changed, the options really are limited. Basically, these two bundles were already so tightly related with so many split  packages that it's almost a lost cause to try to improve the situation. 

The real issue here is that such a large amount of the various eclipse projects depend on public API utility classes that live in UI bundles, and they use these classes in yet more APIs, and thus the dependency on UI bundles spreads, making it difficult to provide a similar level of functionality in a headless environment (like jdt.ls). 

I will continue trying to separate out logic where I can, with the goal of using new packages, or migrating whole packages together. But in situations where a very tight coupling with lots of split packages already exists, I believe I will continue along my current path as it can't make the situation much worse, to be honest.

laeubi:
> I will continue trying to separate out logic where I can, with the goal of using new packages, or migrating whole packages together

:+1: please do so! as you mentioned there is no easy way out and everything is better than being stuck forever!

PR #1313, opened 2023-11-18
try eclipse tool for finding compiler warnings

Bananeweizen:
See eclipse-platform/.github#144. Need to check the Eclipse CI build afterwards.


Bananeweizen:
did not work

PR #1318, opened 2023-11-20
enable explicitlyClosedAutoCloseable=warning

jukzi:


vogella:
I would set it to error, given the number of warnings, new warnings will be ignored.

But if you prefer warning, fine for me, please merge in this case.

jukzi:
> given the number of warnings

i work on that

Issue #1074, opened 2023-08-31
Adding a child to a ParStack that is not currently visible creates a new one in the current perspective


laeubi:
I want to add a `Part` to a `PartStack` that is currently not visible in a pure E4 application perspective.

This currently leads to a new part stack created in the _current_ perspective and the part not added to the _invisible_ one so if one later switch to that perspective the part is missing.

To reproduce one needs two perspectives (A+B) and a partstack X in B:

1. Make A the active perspective.
2. Have some code (e.g. a button) that creates a new `MPart` and search for X, e.g. `modelService.find("X", application);`
3. The add the new part to that stack `stack.getChildren().add(newPart);`
4. Now one can see a new Partstack created in `A` switching to `B` shows that `X` does **not** contain any new parts.

laeubi:
@mickaelistria does this really fix the described issue here? The PR talks about restarts while this bug talks about invisible perspective stacks.

mickaelistria:
I guess it's a type in the commit message of https://github.com/eclipse-platform/eclipse.platform.ui/commit/0bf3af9269e0b3816e2beff16cb8b76cb76d7f56 which has caused this ticket to be closed.
But indeed, it'snot related, so let's reopen (and don't hesitate to do it without waiting for my feedback in the future if you're confident enough about it)

Issue #1047, opened 2023-08-18
Multiselection: lost after restart


Wittmaxi:
# How to reproduce
1) Press Ctrl+F
2) Enter a search term for a word that is contained multiple times in the file
3) Press "Select All"
4) Immediately close eclipse
5) Reopen eclipse
 
I am not sure if there are other way to select multiple words at once in eclipse. If that mechanism exists, it would be interesting to see if the same behaviour still applies.

# Expected behaviour
After restarting eclipse, I expect the same - multiple - areas to still be part of my selection, instead of having one selection that spans from the first to the last selection.

# Screenshots
After Step 3)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/2249e12b-5454-4141-8fc2-fe081cc51d0b)
After step 5)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/68a84df0-863c-4b0d-b1f1-4ddd34d8f22a)

# My setup
Eclipse IDE for RCP and RAP Developers (includes Incubating components)

Version: 2023-06 (4.28.0)
Build id: 20230608-1333

jukzi:
@mickaelistria  caused by implementing multiselection: https://github.com/eclipse-platform/eclipse.platform.ui/commit/13e3e7a5866fb13f479253382030907238a54ee2

@Wittmaxi If you need that feature please contribute a PR for saving multiselection

mickaelistria:
It indeed looks like the editor only persists simple selections properly: block selection is discarded and multi-selection is joined (as described in this issue).
We would need a way to improve how the workspace persists the selections to better cover the various type. This would happen in a combination of changes in `AbstractTextEditor#saveState` and `AbstractTextEditor#doRestoreState` to support multiple selections. A PR would be welcome (note that backward compatibiity must be guaranteed, so `doRestoreState` would need to still be able to part legacy memento.

PR #1319, opened 2023-11-21
Fix Bug #1074

Wittmaxi:
Fixes Bug #1047: `Multiselection: lost after restart`
When saving it's state, an AbstractEditor will now consider MultiSelection.

This PR is a proposal for improvement and requires a few more iterations before release.
In particular, I would like to check:
- Do I break backwards-compatibility with prior Mementos?
- Are there some special AbstractTextEditor's which break because of my change?
- Is the format I chose for saving the different regions of a MultiSelection sane?

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.ResourceBundle;
@@ -154,6 +155,7 @@ import org.eclipse.jface.text.ITextViewerExtension8;
 import org.eclipse.jface.text.ITextViewerExtension8.EnrichMode;
 import org.eclipse.jface.text.IUndoManager;
 import org.eclipse.jface.text.IUndoManagerExtension;
+import org.eclipse.jface.text.MultiTextSelection;
 import org.eclipse.jface.text.Position;
 import org.eclipse.jface.text.Region;
 import org.eclipse.jface.text.TabsToSpacesConverter;
@@ -291,7 +293,8 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 	protected static final String TAG_SELECTION_OFFSET= "selectionOffset"; //$NON-NLS-1$
 
 	/**
-	 * Tag used in the {@link IMemento} when saving and restoring the editor's selection length.
+	 * Tag used in the {@link IMemento} when saving and restoring the editor's
+	 * selection length.
 	 *
 	 * @see #saveState(IMemento)
 	 * @see #restoreState(IMemento)
@@ -7049,9 +7052,15 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 	@Override
 	public void saveState(IMemento memento) {
 		ISelection selection= doGetSelection();
-		if (selection instanceof ITextSelection) {
-			memento.putInteger(TAG_SELECTION_OFFSET, ((ITextSelection)selection).getOffset());
-			memento.putInteger(TAG_SELECTION_LENGTH, ((ITextSelection)selection).getLength());
+		if (selection instanceof IMultiTextSelection multiSelect && multiSelect.getRegions().length > 1) {
+			for (int i = 0; i < multiSelect.getRegions().length; i++) {
+				memento.putInteger(TAG_SELECTION_OFFSET + i, multiSelect.getRegions()[i].getOffset());
+				memento.putInteger(TAG_SELECTION_LENGTH + i, multiSelect.getRegions()[i].getLength());
+			}
+
+		} else if (selection instanceof ITextSelection singleSelect) {
+			memento.putInteger(TAG_SELECTION_OFFSET, singleSelect.getOffset());
+			memento.putInteger(TAG_SELECTION_LENGTH, singleSelect.getLength());
 		}
 		final StyledText textWidget= fSourceViewer.getTextWidget();
 		memento.putInteger(TAG_SELECTION_TOP_PIXEL, textWidget.getTopPixel());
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -7074,15 +7097,11 @@ protected boolean containsSavedState(IMemento memento) {
 	 * @since 3.3
 	 */
 	protected void doRestoreState(IMemento memento) {
-		Integer offset= memento.getInteger(TAG_SELECTION_OFFSET);
-		if (offset == null)
+		ISelection savedSelection = readSelectionFromMemento(memento);
```
Wittmaxi, line 7095: A common pattern across other selection-using methods of AbstractTextEditor is checking with `getSelectionProvider().isValidSelection(selection)` - it might be useful to add that check inside of this code?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -7095,6 +7114,38 @@ protected void doRestoreState(IMemento memento) {
...
+					continue;
+				}
+
+				Integer length = memento.getInteger(TAG_SELECTION_LENGTH + i);
+				if (length == null) {
+					continue;
+				}
+				regions.add(new Region(offset, length));
+			}
+			return new MultiTextSelection(new Document(), regions.toArray(new IRegion[0]));
```
Wittmaxi: Is using an empty document for the `MultiTextSelection` permissible in this case? If not, which document should I pass to it?

mickaelistria: No, it should be something like `getDocumentProvider().getDocument(getEditorInput())`

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -7043,9 +7054,18 @@ public void restoreState(IMemento memento) {
 	@Override
 	public void saveState(IMemento memento) {
 		ISelection selection= doGetSelection();
-		if (selection instanceof ITextSelection) {
-			memento.putInteger(TAG_SELECTION_OFFSET, ((ITextSelection)selection).getOffset());
-			memento.putInteger(TAG_SELECTION_LENGTH, ((ITextSelection)selection).getLength());
+		if (selection instanceof IMultiTextSelection multiSelect && multiSelect.getRegions().length > 1) {
+			memento.putBoolean(TAG_IS_MULTISELECTION, true);
+			memento.putInteger(TAG_MULTISELECTION_COUNT, multiSelect.getRegions().length);
```
mickaelistria: I don't think those 2 flags are necessary, they're redundant. You can just check the memento to deduce those data.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -7062,6 +7082,9 @@ public void saveState(IMemento memento) {
 	 * @since 3.3
 	 */
 	protected boolean containsSavedState(IMemento memento) {
+		if (memento.getBoolean(TAG_IS_MULTISELECTION) == Boolean.TRUE) {
```
mickaelistria:
```suggestion
		if (memento.getInteger(TAG_SELECTION_OFFSET + "0") != null) {
```

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
mickaelistria: If it wasn't done earlier, don't bother adding it now. Maybe in a future PR.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -7095,6 +7114,38 @@ protected void doRestoreState(IMemento memento) {
 			textWidget.setHorizontalPixel(horizontalPixel.intValue());
 	}
 
+	@SuppressWarnings("boxing")
+	private ISelection readSelectionFromMemento(IMemento memento) {
+		if (memento.getBoolean(TAG_IS_MULTISELECTION)) {
```
mickaelistria:
```suggestion
		if (memento.getInteger(TAG_SELECTION_OFFSET + "0") != null) {
```
and them something like
```java
int i = 0;
Integer offset = null;
while ((offset = memento.getInteger(TAG_SELECTION_OFFSET + i)) != null) {
	Integer length = memento.getInteger(TAG_SELECTION_LENGTH + i);
	if (length == null) {
		continue;
	}
	regions.add(new Region(offset, length));
}
```

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -7095,6 +7114,38 @@ protected void doRestoreState(IMemento memento) {
...
+			int selectionRegions = memento.getInteger(TAG_MULTISELECTION_COUNT);
+			Vector<IRegion> regions = new Vector<>();
```
mickaelistria: Since the only operation here is appending content, using `LinkedList` might be more suitable.

mickaelistria (commented):
This is definitely going in the right direction. I've submitted some proposals to minimize the amount of extra fields in the memento.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
Wittmaxi: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1320
Documented here. Feel free to close if you feel like that Issue is too nitpicky

Wittmaxi:
@mickaelistria I have addressed all of your concerns. After a bit of manual testing, I am more confident in this PR.

mickaelistria (changes requested):
Code is good to be merged, thanks.
However, can ayou please reformat the commit message. The usual best practice is to mention what's fixed in the commit header, then an optional details block and then the reference to the fixes issue.

For example here it could be
```
Restore editor MultiSelection on restart

Stores the multi-selection for the text editor in the memento; and restores it on restart.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1074
```

Wittmaxi:
@mickaelistria I changed the message of the commit!

mickaelistria (approved):
Thank you. We'll merge it as soon as the 4.31 stream is open for development (in a couple of weeks IIRC)

PR #1321, opened 2023-11-22
Correct output and source specifications in build.properties

HeikoKlare:
* Add missing output entries
* Remove source entries in org.eclipse.ui.themes, which has no source folders
* Fix classpath with unseparated bin folders and adapt .gitignore in org.eclipse.ui.examples.javaeditor
* Add missing inclusion of generated classes in binary output for org.eclipse.e4.ui.examples.job


jukzi:
that was quick, do you have a script for that?

HeikoKlare:
No, I do that by hand, but it's a pretty fast task. Since there are some special cases like duplicate entries, outdated output definitions etc., I don't expect defining a script to pay off.

But note this is still a draft, so I will probably have to make some fixes before everything is fine.

HeikoKlare:
The changes to `build.properties` seem to require micro version bumps, as the API tools otherwise show a validation error because of changes made to the bundle. I already added the bumps for those bundles that have not been touched since the last release. I guess we will have to add version bump for all other affected bundles once 4.31 stream opens.

PR #1324, opened 2023-11-22
Ensure that the search Result project is open

### LINKED ISSUE #1322: Eclipse text search mostNestedProject should not include closed projects when detecting which files to open
Summary

When having nested projects some are open and some are closed when searching for textual content:
- the textual search shows the results of the open projects only
- but when you double click on the search results it will send you to the mostNestedProject which may be closed
- Resulting in opening an empty editor

Steps to reproduce

- Import a project from existing workspace (A)
- Create another project (B)
- Drag folder from A project into B
- Select Link to files and folders
- Search for a keyword in that folder then click to open it

What is the current bug behavior?
The result is resource "A/file.txt" doesn't exist

What is the expected correct behavior?
It should open the file in the open projects only.

Relevant logs and/or screenshots
image

org.eclipse.search.internal.ui.text.FileSearchPage.mostNestedEquivalent(IFile)
image

This should check the file exists in an open projects only

MKamelEJ:
Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1322

### DIFF
--- a/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchPage.java
+++ b/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchPage.java
@@ -276,6 +276,7 @@ public class FileSearchPage extends AbstractTextSearchViewPage implements IAdapt
 		ITextFileBuffer textFileBuffer = textFileBufferManager.getTextFileBuffer(resource.getFullPath(),
 				LocationKind.IFILE);
 		return Arrays.stream(resource.getWorkspace().getRoot().findFilesForLocationURI(resource.getLocationURI())) //
+				.filter(aFile -> aFile.getProject().isAccessible()) // Check the project is Open
 				.min(Comparator.comparingInt(aFile -> aFile.getFullPath().segments().length)) //
 				.filter(aFile -> {
 					ITextFileBuffer buffer = textFileBufferManager.getTextFileBuffer(aFile.getFullPath(),


bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchPage.java
```diff
@@ -276,6 +276,7 @@ private IFile mostNestedEquivalent(IFile resource) {
 		ITextFileBuffer textFileBuffer = textFileBufferManager.getTextFileBuffer(resource.getFullPath(),
 				LocationKind.IFILE);
 		return Arrays.stream(resource.getWorkspace().getRoot().findFilesForLocationURI(resource.getLocationURI())) //
+		        .filter(aFile -> aFile.getProject().isOpen()) // Check File is in an Open project
```
mickaelistria: What about
```suggestion
				.filter(IResource::isAccessible)
```
?

MKamelEJ: Although not sure if it will fix it this way
isAccessible == exists, not sure if in a closed project this would always be false.
`Returns whether this resource is accessible. For files and folders, this is equivalent to existing; for projects, this is equivalent to existing and being open. The workspace root is always accessible.`

mickaelistria: I think you're right, so let's stick to your proposal. Thanks for checking anyway!

PR #1325, opened 2023-11-22
Remove Platform Text Oomph setup

HeikoKlare:
Since the platform.text repository has been merged into the platform.ui one, the Oomph setup for Platform Text became obsolete, as everything is now part of the Platform UI setup.

This change removes the unnecessary `org.eclipse.text.releng` project containing the Oomph setup file.


akurtakov:
Makes perfect sense.

PR #1329, opened 2023-11-26
Support multiple IPresentationReconciler in generic editor

### LINKED ISSUE #1328: [generic editor] Support  multiple IPresentationReconcilers
Currently `org.eclipse.ui.internal.genericeditor.ExtensionBasedTextViewerConfiguration.getPresentationReconciler(ISourceViewer)` only use the first `IPresentationReconciler` it finds in the registry. Not that this is random, it currently hinders extensions to provide additional presentation for different content types.

Instead it should only use the first if there is only one and otherwise install a wrapper that delegates to the individual items.

@mickaelistria WDYT?

laeubi:
Currently only one `IPresentationReconciler` is supported and if there are multiple ones the first is chosen.

This adds support for multiple `IPresentationReconciler` being registered, if there is only one the instance is returned, otherwise a wrapper is created that delegates to the individual ones.

Fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/1328

### DIFF
--- a/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/CompositePresentationReconciler.java
+++ b/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/CompositePresentationReconciler.java
@@ -1,0 +1,60 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


mickaelistria (approved):
That would be good to merge it during 4.31 M1.
Note that there will be cases where multiple reconcilers may provide different conflicting files. But we can cover it later.

laeubi:
@mickaelistria can you probably help with `org.eclipse.ui.genericeditor.tests.StylingTest.testStyle`? I assume now multiple providers are supported somehow in the test another than the expected one is fetched...

Issue #649, opened 2023-03-20
Lazy TreeViewer reveal is quadratic  on Linux


basilevs:
Moved from [Bug 516281](https://bugs.eclipse.org/bugs/show_bug.cgi?id=516281)
On large lazy tree models, `TreeViewer.setSelection()` and `TreeViewer.reveal()` are very slow.

# Test steps

- Import [attached Eclipse project](https://github.com/eclipse-platform/eclipse.platform.ui/files/11019153/bug516281.zip)   in a PDE

- Add native SWT fragment to included launch configuration TreeViewerPerformanceTest.
- Alternatively create your own launch configuration to start TreeViewerPerformanceTest Junit plug-in test.
- Run the configuration/test

## Expected
For virtual trees, the time to render should not depend on the model size.

## Actual
On MacOS and Windows the time taken is at most linear, on Linux, reveal is very slow (quadratic).

### MacOS
52 ms to process 10000 items
458 ms to process 100000 items
The time to select scales as O(n^0.94) as the size of the model grows from 10000 to 100000
36 ms to process 10000 items
384 ms to process 100000 items
The time to reveal scales as O(n^1.02) as the size of the model grows from 10000 to 100000
### Windows
9 ms to process 10000 items
12 ms to process 100000 items
The time to select scales as O(n^0.16) as the size of the model grows from 10000 to 100000
6145 ms to process 10000 items
59166 ms to process 100000 items
The time to reveal scales as O(n^0.98) as the size of the model grows from 10000 to 100000
### Linux
5 ms to process 10000 items
34 ms to process 100000 items
The time to select scales as O(n^0.83) as the size of the model grows from 10000 to 100000
3888 ms to process 10000 items
396560 ms to process 100000 items
The time to reveal scales as O(n^2.01) as the size of the model grows from 10000 to 100000

Example of a hot stacktrace:
```
"main" #1 prio=5 os_prio=0 cpu=73458.71ms elapsed=85.50s tid=0x00007fa6c4026000 nid=0x1b6ab3 runnable  [0x00007fa6c9e34000]
   java.lang.Thread.State: RUNNABLE
	at org.eclipse.swt.internal.gtk.GTK.gtk_tree_store_set(Native Method)
	at org.eclipse.swt.widgets.Tree.getId(Tree.java:254)
	at org.eclipse.swt.widgets.Tree._getItem(Tree.java:203)
[...]

PR #1331, opened 2023-11-27
[performance] faster Tree expand - eclipse.platform.swt#901

jukzi:
- insert every item at index 0 (in reverse order)

https://github.com/eclipse-platform/eclipse.platform.swt/issues/901

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TreeViewer.java
@@ -1150,8 +1150,9 @@ public class TreeViewer extends AbstractTreeViewer {
 			item.dispose();
 
 			// create children on parent
-			for (Object element : children) {
-				createTreeItem(parent, element, -1);
+			// For performance insert every item at index 0 (in reverse order):
+			for (int i = children.length - 1; i >= 0; i--) {
+				createTreeItem(parent, children[i], 0);
 			}
 
 			// reset the selection. client's selection listener should not be triggered.
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -46,6 +46,7 @@ import org.eclipse.swt.events.TreeListener;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Item;
+import org.eclipse.swt.widgets.Tree;
 import org.eclipse.swt.widgets.TreeItem;
 import org.eclipse.swt.widgets.Widget;
 
@@ -362,8 +363,9 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 
 		// Optimize for the empty case
 		if (items.length == 0) {
-			for (Object element : elements) {
-				createTreeItem(widget, element, -1);
+			// For performance insert every item at index 0 (in reverse order):
+			for (int i = elements.length - 1; i >= 0; i--) {
+				createTreeItem(widget, elements[i], 0);
 			}
 			return;
 		}
@@ -846,8 +848,9 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 				} else {
 					children = getSortedChildren(parentElement);
 				}
-				for (Object element : children) {
-					createTreeItem(widget, element, -1);
+				// For performance insert every item at index 0 (in reverse order):
+				for (int i = children.length - 1; i >= 0; i--) {
+					createTreeItem(widget, children[i], 0);
 				}
 			}
 		} finally {
@@ -856,16 +859,16 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	}
 
 	/**
-	 * Creates a single item for the given parent and synchronizes it with the
-	 * given element.
+	 * Creates a single item for the given parent and synchronizes it with the given
+	 * element. The fastest way to insert many items is documented in
+	 * {@link TreeItem#TreeItem(Tree,int,int)}
 	 *
-	 * @param parent
-	 *            the parent widget
-	 * @param element
-	 *            the element
-	 * @param index
-	 *            if non-negative, indicates the position to insert the item
-	 *            into its parent
+	 * @param parent  the parent widget
+	 * @param element the element
+	 * @param index   if non-negative, indicates the position to insert the item
+	 *                into its parent
+	 * @see org.eclipse.swt.widgets.TreeItem#TreeItem(org.eclipse.swt.widgets.Tree,
+	 *      int, int)
 	 */
 	protected void createTreeItem(Widget parent, Object element, int index) {
 		Item item = newItem(parent, SWT.NULL, index);
@@ -1831,6 +1834,7 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 				return;
 			}
 			createChildren(widget, false);
+			// XXX for performance widget should be expanded after expanding children:
 			if (widget instanceof Item) {
 				setExpanded((Item) widget, true);
 			}
@@ -1844,6 +1848,7 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 					}
 				}
 			}
+			// XXX expanding here fails on linux
 		}
 	}
 


laeubi:
@SyntevoAlex should this happen for all platforms or only for Windows?

akurtakov:
Jface should be ws agnostic.

laeubi:
> Jface should be ws agnostic.

That's true but still we can have platform switches in it if required.

SyntevoAlex:
I suggest adding a code comment with explanation.

> should this happen for all platforms or only for Windows?

Doing it on all platforms won't hurt, I think. But I didn't test it myself. You could play with my `Bug575787_Tree_SetItemCount_Perf` snippet (already in SWT)

jukzi:
> I suggest adding a code comment with explanation.

i added comments and reverse order

SyntevoAlex:
Good. Note that I didn't test anything.

SyntevoAlex (commented):
Looks reasonable to me, but I didn't test it.

jukzi:
I also added the setRedraw(false) trick, which cuts downs setImage() and setText() time when i sort the Problems view with many elements. for example ~4000 elements org.eclipse.swt.widgets.Display.readAndDispatch () reduces from 1300ms to 600ms

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TreeViewer.java
```diff
@@ -363,7 +363,9 @@ void preservingSelection(Runnable updateCode, boolean reveal) {
 		// updateCode.run():
 		insidePreservingSelection = true;
 		try {
+			tree.setRedraw(false);
 			super.preservingSelection(updateCode, reveal);
+			tree.setRedraw(true);
```
basilevs: Should `setRedraw(true)` be in a finally block?

basilevs:
I've tried the test reproducing the problem I'm working on from https://github.com/eclipse-platform/eclipse.platform.ui/issues/649
## GTK
No regressions
### Before change
﻿210 ms to process 10000 items
202 ms to process 100000 items
The time to select scales as O(n^-0.02) as the size of the model grows from 10000 to 100000
1160 ms to process 10000 items
83267 ms to process 100000 items
The time to reveal scales as O(n^1.86) as the size of the model grows from 10000 to 100000
### After change
196 ms to process 10000 items
399 ms to process 100000 items
The time to select scales as O(n^0.31) as the size of the model grows from 10000 to 100000
1260 ms to process 10000 items
86350 ms to process 100000 items
The time to reveal scales as O(n^1.84) as the size of the model grows from 10000 to 100000
## MacOS
Slowdown?
### Before the change
53 ms to process 10000 items
430 ms to process 100000 items
The time to select scales as O(n^0.90) as the size of the model grows from 10000 to 100000
38 ms to process 10000 items
381 ms to process 100000 items
The time to reveal scales as O(n^1.00) as the size of the model grows from 10000 to 100000
### After the change
110 ms to process 10000 items
1005 ms to process 100000 items
The time to select scales as O(n^0.96) as the size of the model grows from 10000 to 100000
91 ms to process 10000 items
944 ms to process 100000 items
The time to reveal scales as O(n^1.01) as the size of the model grows from 10000 to 100000

basilevs:
### MacOS increased tree size
The change definitely increases a time to reveal an item in a large tree
### Before
413 ms to process 100000 items
4265 ms to process 1000000 items
The time to select scales as O(n^1.01) as the size of the model grows from 100000 to 1000000
389 ms to process 100000 items
4133 ms to process 1000000 items
The time to reveal scales as O(n^1.03) as the size of the model grows from 100000 to 1000000
### After
969 ms to process 100000 items
9748 ms to process 1000000 items
The time to select scales as O(n^1.00) as the size of the model grows from 100000 to 1000000
963 ms to process 100000 items
9959 ms to process 1000000 items
The time to reveal scales as O(n^1.01) as the size of the model grows from 100000 to 1000000

jukzi:
> The change definitely increases a time to reveal an item in a large tree

@SyntevoAlex any idea?
@basilevs can you please share the testprogram? i want to try it on windows

basilevs:
@juksi https://github.com/eclipse-platform/eclipse.platform.ui/issues/649

https://github.com/eclipse-platform/eclipse.platform.ui/files/11019153/bug516281.zip

jukzi:
> @Juksi #649

jukZi :-)
thanks,. can you post measurements if you combine this PR with 649? all platforms, before+after 
This PR contains 3 different optimizations: 

- move setExpanded() after expanding the children
- insert every item at index 0 (in reverse order)
- setRedraw

can you please figure out which of them does cause regression on Mac?

basilevs:
> can you post measurements if you combine this PR with 649?

#649 is WIP, there is nothing to combine, I've tested on master and on your branch
> all platforms, before+after

I don't have Windows env and have already covered Linux and MacOS

> This PR contains 3 different optimizations: 
>     * move setExpanded() after expanding the children
>     * insert every item at index 0 (in reverse order)
>     * setRedraw
> can you please figure out which of them does cause regression on Mac?

I've removed new `setRedraw()` lines from `TreeViewer` and got old performance:

```
395 ms to process 100000 items
4172 ms to process 1000000 items
The time to select scales as O(n^1.02) as the size of the model grows from 100000 to 1000000
366 ms to process 100000 items
4070 ms to process 1000000 items
The time to reveal scales as O(n^1.05) as the size of the model grows from 100000 to 1000000
```

This is a major surprise - I assumed `setRedraw()` was a natural wrapper for any batch operations.

basilevs:
Performance tests do not work, but are necessary. Life is pain.

jukzi:
on windows:
```
before PR:
3 ms to process 10000 items
6 ms to process 100000 items
The time to select scales as O(n^0,24) as the size of the model grows from 10000 to 100000
1094 ms to process 10000 items
10413 ms to process 100000 items
The time to reveal scales as O(n^0,98) as the size of the model grows from 10000 to 100000

without redraw(false)
3 ms to process 10000 items
6 ms to process 100000 items
The time to select scales as O(n^0,27) as the size of the model grows from 10000 to 100000
995 ms to process 10000 items
9993 ms to process 100000 items
The time to reveal scales as O(n^1,00) as the size of the model grows from 10000 to 100000

with redraw(false)
5 ms to process 10000 items
13 ms to process 100000 items
The time to select scales as O(n^0,38) as the size of the model grows from 10000 to 100000
3105 ms to process 10000 items
30691 ms to process 100000 items
The time to reveal scales as O(n^0,99) as the size of the model grows from 10000 to 100000
```

=> that test can not show that this PR  can make some things faster, but also no regression
=> with this test redraw(false) is not good. probably because the tested view is much less complicated then the Problem View i measured with. So probably it's better to use redraw(false) not in general but may be in Problem View

basilevs:
@jukzi  `reveal()` operation is not used in Problem View. Project Explorer or Outline views are a better match for my test.

jukzi:
let forget about redraw(false), it's already set in org.eclipse.ui.internal.views.markers.UIUpdateJob.runInUIThread(IProgressMonitor)

basilevs (approved):

basilevs:
@jukzi have you tried using` TreeItem.setItemCount()` instead of creating items one-by-one? IMO, it should dramatically increase performance.

jukzi:
> @jukzi have you tried using` TreeItem.setItemCount()` instead of creating items one-by-one? IMO, it should dramatically increase performance.

I have not tried that. i don't understand how it should be used. Is that replacement for multiple `new TreeItem(parent, -1)` ?

basilevs:
> Is that replacement for multiple `new TreeItem(parent, -1)` ?

Yes. The change would be to replace multiple new `TreeItem()` with `TreeItem.setItemCount()`
than traverse those with `getItems()`, `updateItem()`and `updatePlus()`. Performance gain is not guaranteed and would require testing (as `getItems()` may be slow enough to negate any gains).

jukzi:
We got a Problem: this change keeps failing on linux - and only linux  - 

[org.eclipse.jface.tests.viewers.VirtualLazyTreeViewerTest.testExpandToLevel](https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1331/8/testReport/junit/org.eclipse.jface.tests.viewers/VirtualLazyTreeViewerTest/testExpandToLevel/)
[org.eclipse.jface.tests.viewers.TreeViewerWithLimitTest.testRemoveItemsAtParent](https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1331/8/testReport/junit/org.eclipse.jface.tests.viewers/TreeViewerWithLimitTest/testRemoveItemsAtParent/)

The code should be theoretically OS independent. i am on windows. @basilevs can you look why it fails on linux?

laeubi:
@jukzi first basic step would be to enhance the test with *what* assertions failed and probably adding a message that helps in understand the test failure.

basilevs:
The test fails to expand second level of the tree:
Screenshot 2023-12-01 at 13 55 19

jukzi:
> Yes. The change would be to replace multiple new `TreeItem()` with `TreeItem.setItemCount()`
> than traverse those with `getItems()`, `updateItem()`and `updatePlus()`. Performance gain is not guaranteed and would require testing (as `getItems()` may be slow enough to negate any gains).

that seems to be not usable in TreeViewer  because
org.eclipse.jface.viewers.TreeViewer.newItem(Widget, int, int) is protected - i.e. in principle it could have been overloaded by clients.
Even though `@noextend This class is not intended to be subclassed by clients.` the reality looks different:
image

jukzi:
linux VirtualLazyTreeViewerTest.testExpandToLevel
> The test fails to expand second level of the tree:

but why - @SyntevoAlex ?

basilevs:
`newItem()` is already not used in `org.eclipse.jface.viewers.TreeViewer.setChildCount(Object, int)` used by lazy TreeViewer and likely in other places.

basilevs:
None of new `createTreeItem()` lines are called in the test. So problem is probably with move of `setExpanded()`

UPDATE: confirmed, moving it to the old spot fixes one test (I only check one test)

jukzi:
> UPDATE: confirmed, moving it to the old spot fixes one test (I only check one test)

i will split that commit

basilevs:
There is a hint in Tree:
https://github.com/basilevs/eclipse.platform.swt/blob/4f69de6866723e1e139a5762d993c5dabf2c4799/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Tree.java#L2545-L2566
```java
long gtk_row_has_child_toggled (long model, long path, long iter) {
	/*
	* Feature in GTK. The expanded state of a row that lost
	* its children is not persisted by GTK. So, the row
	* doesn't exhibit the expanded state after obtaining the
	* children. The fix is to preserve the expanded state
	* and use this callback, as it is invoked when a row has
	* gotten the first child row or lost its last child row.
	*/
	int [] index = new int [1];
	GTK.gtk_tree_model_get (modelHandle, iter, ID_COLUMN, index, -1);
	if (index [0] >= items.length) return 0;
	TreeItem item = items [index [0]];
	if (item == null) return 0;
	int childCount = GTK.gtk_tree_model_iter_n_children (modelHandle, item.handle);
	if (childCount != 0 && item.isExpanded) {
		OS.g_signal_handlers_block_matched (handle, OS.G_SIGNAL_MATCH_DATA, 0, 0, 0, 0, TEST_EXPAND_ROW);
		GTK.gtk_tree_view_expand_row (handle, path, false);
		OS.g_signal_handlers_unblock_matched (handle, OS.G_SIGNAL_MATCH_DATA, 0, 0, 0, 0, TEST_EXPAND_ROW);
	}
	return 0;
}
```
But the condition was not used during my debugging.

basilevs:
Idea - Tree does not recheck states of children when item is expanded. So if state did not stick to GTK model the first time, when its parent is expanded the expansion state is not reapplied.

Effectively the tree only works as expected when it is manipulated in a way end user would - top to bottom.
This seems to be a bug in GTK Tree implementation, as other platforms persist expanded state of invisible item.

jukzi:
> None of new `createTreeItem()` lines are called in the test. So problem is probably with move of `setExpanded()`

i created an experiment: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1357/files
it seems to be SLOWER on windows to use setItemCount

jukzi:
> This seems to be a bug in GTK Tree implementation, as other platforms persist expanded state of invisible item.

sounds reasonable

jukzi:
only unrelated test errors: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1351

laeubi:
> > This seems to be a bug in GTK Tree implementation, as other platforms persist expanded state of invisible item.
> 
> sounds reasonable

Why is it not a bug in **windows** that expanding invisible childs is faster than expand visible ones? Just a thought ;-)

jukzi:
> Why is it not a bug in **windows** that expanding invisible childs is faster than expand visible ones? Just a thought ;-)

Invisible things don't need to be drawn.

SyntevoAlex:
> linux VirtualLazyTreeViewerTest.testExpandToLevel
> 
> > The test fails to expand second level of the tree:
> 
> but why - @SyntevoAlex ?

I glanced at `gtk_tree_view_real_expand_row()` in GTK sources and I had the impression that "item is expanded" and "internal data for child items exist" are the same thing. In other words, child items are only created when parent is expanded. This might explain why it's not possible to modify item that is not expanded yet.

If you care to make a pure SWT snippet, I could spend a bit of time to check it with native debugger.

jukzi:
> If you care to make a pure SWT snippet, I could spend a bit of time to check it with native debugger.

@basilevs can you create a pure SWT reproducer for GTK?

basilevs:
@jukzi no, but I assume https://github.com/basilevs/eclipse.platform.swt/blob/158e4c6d7e70a9c305ea3e788928b7ff6419edb6/tests/org.eclipse.swt.tests/JUnit%20Tests/org/eclipse/swt/tests/junit/performance/Test_org_eclipse_swt_widgets_Tree.java#L262 would work fine. Try it to see if results match your report.

basilevs:
Following test fails on last line on GTK. If user manually "helps" to expand "item2" node, the test instantly passes.
It does not fail if order of expansions is reversed. 
```diff
diff --git a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Tree.java b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Tree.java
index 0ad17d5..86ada48 100644
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Tree.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Tree.java
@@ -1188,6 +1188,35 @@
 		new TreeItem(item_0, 0, 0);
 		assertEquals(10, item_0.getItemCount());
 	});
 }
 
+@Test
+public void test_persistExpandStatusForInvisibleItems() throws InterruptedException {
+	tree.dispose();
+	tree = new Tree(shell, SWT.VIRTUAL);
+	setWidget(tree);
+	shell.setLayout(new FillLayout());
+	SwtTestUtil.openShell(shell);
+	int itemCount[] = new int[] {1};
+	List<TreeItem> dataRequests = new ArrayList<>();
+	tree.addListener(SWT.SetData, event -> {
+		TreeItem item = (TreeItem) event.item;
+		dataRequests.add(item);
+		item.setText("item"+itemCount[0]++);
+		item.setItemCount(1);
+	});
+	TreeItem item1 = new TreeItem(tree, SWT.NONE);
+	TreeItem item2 = item1.getItem(0);
+	TreeItem item3 = item2.getItem(0);
+	assertTrue(dataRequests.remove(item1));
+	assertTrue(dataRequests.remove(item2));
+	assertTrue(dataRequests.isEmpty());
+	item2.setExpanded(true);
+	assertTrue(dataRequests.isEmpty());
+	item1.setExpanded(true);
+	// Item 3 is now visible and should request data
+	SwtTestUtil.processEvents(10000, () -> !dataRequests.isEmpty());
+	assertFalse(dataRequests.isEmpty());
+}
+
 }
```

jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform.ui/issues/275

PR #1334, opened 2023-11-27
Push down HTML Printer and various associated classes from o.e.jface …

robstryker:
…to org.eclipse.text

Signed-off-by: Rob Stryker <stryker@redhat.com>

Part 2

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/SingleCharReader.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/SingleCharReader.java
@@ -13,59 +13,15 @@
  *******************************************************************************/
 package org.eclipse.jface.internal.text.html;
 
-import java.io.IOException;
-import java.io.Reader;
+import org.eclipse.text.readers.SingleCharacterReader;
 
 
 /**
  * <p>
  * Moved into this package from <code>org.eclipse.jface.internal.text.revisions</code>.</p>
+ *
+ * @deprecated Please use SingleCharacterReader
  */
-public abstract class SingleCharReader extends Reader {
+public abstract class SingleCharReader extends SingleCharacterReader {
 
-	/**
-	 * @see Reader#read()
-	 */
-	@Override
-	public abstract int read() throws IOException;
-
-	/**
-	 * @see Reader#read(char[],int,int)
-	 */
-	@Override
-	public int read(char cbuf[], int off, int len) throws IOException {
-		int end= off + len;
-		for (int i= off; i < end; i++) {
-			int ch= read();
-			if (ch == -1) {
-				if (i == off)
-					return -1;
-				return i - off;
-			}
-			cbuf[i]= (char)ch;
-		}
-		return len;
-	}
-
-	/**
-	 * @see Reader#ready()
-	 */
-	@Override
-	public boolean ready() throws IOException {
-		return true;
-	}
-
-	/**
-	 * Returns the readable content as string.
-	 * @return the readable content as string
-	 * @exception IOException in case reading fails
-	 */
-	public String getString() throws IOException {
-		StringBuilder buf= new StringBuilder();
-		int ch;
-		while ((ch= read()) != -1) {
-			buf.append((char)ch);
-		}
-		return buf.toString();
-	}
 }
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/SingleCharReader.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/SingleCharReader.java
@@ -22,6 +22,7 @@ import org.eclipse.text.readers.SingleCharacterReader;
  *
  * @deprecated Please use SingleCharacterReader
  */
+@Deprecated
 public abstract class SingleCharReader extends SingleCharacterReader {
 
 }
--- a/bundles/org.eclipse.text/src/org/eclipse/text/html/CoreHTMLPrinter.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/text/html/CoreHTMLPrinter.java
@@ -1,0 +1,465 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2018 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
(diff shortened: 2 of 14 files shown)


robstryker:
This PR looks ready for review? I can only see 1 test fail on mac and it seems unrelated to my code?

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreRGB.java
```diff
@@ -0,0 +1,231 @@
...
+ * object has been inserted.
+ * </p>
+ * <p>
+ * This class is a clone of org.eclipse.swt.RGB without any UI dependencies
+ * </p>
+ *
+ * @since 3.14
+ */
+
+public final class CoreRGB implements Serializable {
```
mickaelistria: As it's a new API, let's try to make it as clean as possible from the beginning:
* What about making it a record?
* What about calling it just RGB anyway (and people who need to disambiguize SWT vs text would just use package name)?

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreRGB.java
```diff
@@ -0,0 +1,231 @@
...
+* @param hue the hue value for the HSB color (from 0 to 360)
+* @param saturation the saturation value for the HSB color (from 0 to 1)
+* @param brightness the brightness value for the HSB color (from 0 to 1)
+*
+* @exception IllegalArgumentException <ul>
+*    <li>ERROR_INVALID_ARGUMENT - if the hue is not between 0 and 360 or
+*    the saturation or brightness is not between 0 and 1</li>
+* </ul>
+*
+* @since 3.2
```
mickaelistria: All those seem incorrect as the class is new.

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreHTMLPrinter.java
```diff
@@ -0,0 +1,465 @@
...
+import org.eclipse.core.runtime.Platform;
+
+/**
+ * Provides a set of convenience methods for creating HTML pages.
+ * <p>
+ * Moved into this package from <code>org.eclipse.jface.internal.text.revisions</code>.</p>
+ *
+ * @since 3.14
+ */
+public class CoreHTMLPrinter {
```
mickaelistria: Can't we find a nicer name for this new API?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/SingleCharReader.java
```diff
@@ -13,59 +13,15 @@
...
-	}
-
-	/**
-	 * @see Reader#ready()
-	 */
-	@Override
-	public boolean ready() throws IOException {
-		return true;
-	}
+public abstract class SingleCharReader extends SingleCharacterReader {
```
mickaelistria, line 26: Let's add the official `@Deprecated` java annotation as well.
Should we plan and announce removal?

mickaelistria (approved):
The overall idea is more than sane.
However, as we're basically creating some new API, it seems best to use it as an opportunity to refresh the code rather than duplicating old stuff.

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreRGB.java
robstryker: I have no problem naming it RGB and using package to disambiguate. I think making it a record has a few drawbacks.  First, there's absolutely no reason to mix things up if we don't need to. I prefer this class match the org.eclipse.swt version as closely as possible, with the same exceptions being thrown, the same logic, and essentially the same interface.

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreRGB.java
robstryker: Fixing it

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreHTMLPrinter.java
robstryker: Switched to BaseHTMLPrinter.  

Other options such as MickaelIsAwesomeHTMLPrinter and RobsLinguisticallyDeficientHTMLPrinter were rejected as too verbose.

robstryker:
The test failures appear to be in the `Eclipse UI Tests ` bundle and do not appear related to these changes at all. But I welcome a second look.

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreHTMLPrinter.java
mickaelistria: Who rejected those proposals? We should get the PMC to vote!
I was more wondering whether this class is actually a printer or something else. As a non-native english speaker, I'm naturally linguistically deficient in English, and to me, a printer is mostly 🖨️ and this class doesn't seem really related to 🖨️. I would go for a more typical Java name such as `HTMLBuilder` or as it's mostly stateless, something like `HTMLToolbox` or something like that. Do 👷 or 🧰 sound worse than 🖨️ for what we're trying to do here?

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreRGB.java
mickaelistria: OK, I see that record cannot hide the canonical constructor (so we could still have checks with the record); which is a pity.

bundles/org.eclipse.text/src/org/eclipse/text/html/RGB.java
```diff
@@ -0,0 +1,231 @@
...
+
+	/**
+	 * the green component of the RGB
+	 */
+	public int green;
+
+	/**
+	 * the blue component of the RGB
+	 */
+	public int blue;
```
mickaelistria: Can those 3 become final?

bundles/org.eclipse.text/src/org/eclipse/text/html/CoreHTMLPrinter.java
laeubi: > I'm naturally linguistically deficient in English, and to me, a printer is mostly 🖨️ and this class doesn't seem really related to 🖨️

I think this is related to printing on the screen or similar like in "System.out.**print**ln" also java has **Print**Writer class...

mickaelistria: Yes, I get it. Yet I'm not sure this class is really about printing on the screen, it's more about building HTML strings as far as I could see.

robstryker:
Changed to HTMLBuilder, made RGB fields final. 

Let's get this pushed plz ;)

PR #1336, opened 2023-11-28
Make TableColumnViewerLabelProvider class public

### LINKED ISSUE #1335: TableColumnViewerLabelProvider should be public

## Suggestion

I'm currently in the process of cleaning up a dialog which internally uses a TableViewer. This viewer is using a custom LabelProvider. Because this provider doesn't support tooltips, I want to switch to the ColumnLabelProvider and then use `ColumnViewerToolTipSupport.enableFor(...)`

Because we have a fairly complex label provider, I don't really want to implement everything from scratch again. So my initial idea was to use the [TableColumnViewerLabelProvider](https://github.com/eclipse-platform/eclipse.platform.ui/blob/d1a48f627950632d08827ac883e14d1e07759f22/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TableColumnViewerLabelProvider.java) (a wrapper), to reuse the already existing code. Except, this class is package-private and therefore not accessible.

Weirdly enough, this class is extended by the [TreeColumnViewerLabelProvider](https://github.com/eclipse-platform/eclipse.platform.ui/blob/d1a48f627950632d08827ac883e14d1e07759f22/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TreeColumnViewerLabelProvider.java), which _is_ public. Meaning right now I have a code snippet of the form:

```
TableViewer tableViewer = new TableViewer(...)
TableViewerColumn tableViewerColumn = new TableViewerColumn(tableViewer, ...)
tableViewerColumn.setLabelProvider(new TreeColumnViewerLabelProvider(...));
```

Which works perfectly fine, but looks really weird...

Looking at the history, the class was marked as experimental, when first created. This marker was removed with 8fb6198fecaf93552ac7007e9104360adfdca565, but the class was never made public. I.e. I think this is merely an oversight.

**My suggestion:** Make this class public so that it is accessible by other projects.

## Community

ptziegler:
The class remained package-private, even after the experimental marker was removed with 8fb6198fecaf93552ac7007e9104360adfdca565, seemingly due to an oversight.

This is reinforced by the fact that the TreeColumnViewerLabelProvider, which directly extends the TableColumnViewerLabelProvider is public.

This leads to weird situations where one has to use the TreeColumnViewerLabelProvider for both TreeViewers and TableViewers, even though its name would only suggest one is compatible...

Resolves #1335

### DIFF
--- a/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface;singleton:=true
-Bundle-Version: 3.32.0.qualifier
+Bundle-Version: 3.33.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.jface,
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TableColumnViewerLabelProvider.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TableColumnViewerLabelProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2015 IBM Corporation and others.
+ * Copyright (c) 2006, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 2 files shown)


vogella:
Can you add a version bump for y+1 z=0 in the x.y.z.qualifier scheme.

ptziegler:
> Can you add a version bump for y+1 z=0 in the x.y.z.qualifier scheme.

Done... I assume the remaining errors are because the baseline is still 2023-09, while the manifest has already been updated for the 2023-12.

image

merks:
I’ll fix the setup to update the baseline tomorrow.  It’s impossible to keep up with the 18 hours of activity every day.  Open master at 18:00 and the floodgates open.

vogella:
Thanks @ptziegler

PR #1341, opened 2023-11-29
Set RAT expression to null when deactivating context

lredor:
If UpdateExpression updating is set to false, the expression will be no longer evaluated. So it can be set to null.
This avoids potential memory leaks as for example on Sirius [1] discussed here [2].

[1] https://github.com/eclipse-sirius/sirius-desktop/issues/178
[2] https://github.com/eclipse-platform/eclipse.platform/discussions/889

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/contexts/ContextService.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/contexts/ContextService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2015 IBM Corporation and others.
+ * Copyright (c) 2005, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -180,6 +180,9 @@ public final class ContextService implements IContextService {
 			final UpdateExpression rat = activationToRat.remove(activation);
 			if (rat != null) {
 				rat.updating = false;
+				// The updating status has been set to false, so the expression is no longer
+				// useful. Set it to null to avoid potential memory leak.
+				rat.expression = null;
 				if (rat.cached != null && rat.cached != EvaluationResult.FALSE) {
 					contextService.deactivateContext(activation.getContextId());
 				}


mickaelistria:
Do you have any idea why references to UpdateExpression are kept? Is the issue really the `.expression` field, or the fact that UpdateExpression should be discarded sooner?

> If UpdateExpression updating is set to false, the expression will be no longer evaluated.

Are we certain of that assumption. Ie can one set to `false` to temporary disable it, and re-enable it later?

lredor:
> Do you have any idea why references to UpdateExpression are kept? Is the issue really the .expression field, or the fact that UpdateExpression should be discarded sooner?

Unfortunately, I don't know. I discovered this part of code during my analysis. I think that the UpdateExpression should be discarded sooner, but I don't know how and if it is really the case. This is the goal of the opened discussion on "Eclipse Platform GitHub":https://github.com/eclipse-platform/eclipse.platform/discussions/889 (to have expert feedback). AFAIK, this PR fixes my memory leak, but maybe with a hack. There is potentially always a leak, because the UpdateExpression is always here, but smaller because the `org.eclipse.ui.internal.expressions.ActivePartExpression` and its associated `org.eclipse.ui.IWorkbenchPart` are no longer kept.

> Are we certain of that assumption. Ie can one set to false to temporary disable it, and re-enable it later?

Currently, the value of field `updating` is changed only in `deactivateContext()` method. But it's not impossible that in the future the logic will be changed. I can complete the PR with a test `expression != null` in method `org.eclipse.ui.internal.contexts.ContextService.UpdateExpression.changed(IEclipseContext)` to avoid a NPE in a case of a future  logic change.

laeubi:
I verified that the object is never resued again after it was deactivated, so the change seems save here.

But this leads to another question, as the class is releasing all its reference to the `UpdateExpression` there must still be someone holding a reference to the `RunAndTrack` itself. I can't see how a RAT can be canceled beside it is being called and return `false` so it might be worth to investigate.

The javadoc claims that "The runnable does not need to be explicitly unregistered from this context" but I can't see how it can become removed if no longer needed, form the implementation the RAT is only removed if the context is disposed otherwise, so we still leaking the instance at least.

lredor:
Thank you for the merge. I've tested on [Sirius side](https://github.com/eclipse-sirius/sirius-desktop/issues/178#issuecomment-1859918496) and it's OK.

PR #1342, opened 2023-11-29
Move to Jenkins pod that can use 8 GB memory instead of 4

iloveeclipse:
See related
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3896
- https://github.com/eclipse-cbi/cbi/wiki#resource-pack
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1287

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,7 @@ pipeline {
 		disableConcurrentBuilds(abortPrevious: true)
 	}
 	agent {
-		label "centos-latest"
+		label "centos-8-8gb"
 	}
 	tools {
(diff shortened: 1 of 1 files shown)


akurtakov:
Note that this will reduce parallel jobs by half as 8gb is counted as two jobs.

iloveeclipse:
> Note that this will reduce parallel jobs by half as 8gb is counted as two jobs.

OK. So we keep with 4GB?

akurtakov:
I believe it's better to keepn4gb  if it's a single pr that has that need

PR #1349, opened 2023-11-30
Disable CSS styling of text editors

vogella:
Migrated the Gerrit from Bug 559321, done by Thomas Wolf.

AbstractTextEditor and subclasses manage their appearance via the editor preferences. If CSS styling is enabled, CSS will override the preferences. So disable CSS styling, and use only the preferences. The CSS engine does set these preferences on theme switches, so the appearance does follow the theme, but now also obeys any preference changes the users makes.

(Note that preference changes get lost on theme switches. But that is unrelated; it appears to be bug 544714.)


sratz:
This is a duplicate of #1036

PR #1352, opened 2023-11-30
Using 4.0.4 for tycho-build extensions

vogella:


akurtakov:
Unrelated test failures and it reduces the mismatch in tycho versions thus pushing.

PR #1357, opened 2023-12-01
WIP SHOWCASE [performance] Tree batch expand

jukzi:
[win]:
OLD implementation:
3 ms to process 10000 items
5 ms to process 100000 items
The time to select scales as O(n^0,19) as the size of the model grows from 10000 to 100000
1068 ms to process 10000 items
10672 ms to process 100000 items
The time to reveal scales as O(n^1,00) as the size of the model grows from 10000 to 100000

NEW batch implementation:
3 ms to process 10000 items
20 ms to process 100000 items  <----- WORSE
The time to select scales as O(n^0,75) as the size of the model grows from 10000 to 100000
1053 ms to process 10000 items
10487 ms to process 100000 items
The time to reveal scales as O(n^1,00) as the size of the model grows from 10000 to 100000

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -46,6 +46,7 @@ import org.eclipse.swt.events.TreeListener;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Item;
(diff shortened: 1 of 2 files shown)


jukzi:
@basilevs

basilevs:
Well, if the experimant is a failure, lets reject it.

jukzi:
> Well, if the experimant is a failure, lets reject it.

i wonder if you want to test it on MAC/linux

basilevs:
No, if this is a regression on Windows, the only way this would be useful, is to add platform checks in TreeViewer, which would be unacceptable.

PR #1362, opened 2023-12-03
Recommend OSGi EventHandler for APP_STARTUP_COMPLETE over IStartup

HannesWell:
And migrate `MonitoringStartup` accordingly.

This also uses the `org.osgi.service.event.propertytypes.EventTopics` `ComponentPropertyType` for which PDE has very recently got support for via https://github.com/eclipse-pde/eclipse.pde/pull/923.

Instead of just recommending the OSGi EventHandler over IStartup, we could also deprecate `IStartup` and the associcate extension-point in favor of this approach? I can then migrate the only remaining usage in the SDK in P2.

### DIFF
--- a/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringStartup.java
+++ b/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringStartup.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (C) 2014, 2019 Google Inc and others.
+ * Copyright (C) 2014, 2023 Google Inc and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -17,21 +17,27 @@
 package org.eclipse.ui.internal.monitoring;
 
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.e4.ui.workbench.UIEvents;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.swt.widgets.Display;
-import org.eclipse.ui.IStartup;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.internal.monitoring.preferences.MonitoringPreferenceListener;
 import org.eclipse.ui.monitoring.PreferenceConstants;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.event.Event;
+import org.osgi.service.event.EventHandler;
+import org.osgi.service.event.propertytypes.EventTopics;
 
 /**
  * Starts the event loop monitoring thread. Initializes preferences from {@link IPreferenceStore}.
  */
-public class MonitoringStartup implements IStartup {
+@Component(service = EventHandler.class)
+@EventTopics(UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
+public class MonitoringStartup implements EventHandler {
 	private EventLoopMonitorThread monitoringThread;
 
 	@Override
-	public void earlyStartup() {
+	public void handleEvent(Event event) {
 		if (monitoringThread != null) {
 			return;
 		}
--- a/bundles/org.eclipse.ui.monitoring/.project
+++ b/bundles/org.eclipse.ui.monitoring/.project
@@ -30,6 +30,11 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.ds.core.builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
--- a/bundles/org.eclipse.ui.monitoring/.settings/org.eclipse.pde.ds.annotations.prefs
+++ b/bundles/org.eclipse.ui.monitoring/.settings/org.eclipse.pde.ds.annotations.prefs
@@ -1,0 +1,7 @@
+dsVersion=V1_4
+eclipse.preferences.version=1
+enabled=true
+generateBundleActivationPolicyLazy=true
+path=OSGI-INF
+validationErrorLevel=error
+validationErrorLevel.missingImplicitUnbindMethod=error
--- a/bundles/org.eclipse.ui.monitoring/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.monitoring/META-INF/MANIFEST.MF
@@ -12,5 +12,9 @@ Export-Package: org.eclipse.ui.internal.monitoring;x-internal:=true,
  org.eclipse.ui.monitoring;x-internal:=true
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.jface;bundle-version="[3.10.0,4.0.0)",
- org.eclipse.ui;bundle-version="[3.106.0,4.0.0)"
+ org.eclipse.ui;bundle-version="[3.106.0,4.0.0)",
+ org.eclipse.e4.ui.workbench;bundle-version="[1.15.300,2.0.0)"
+Service-Component: OSGI-INF/org.eclipse.ui.internal.monitoring.MonitoringStartup.xml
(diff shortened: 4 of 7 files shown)


bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringStartup.java
```diff
@@ -17,21 +17,27 @@
...
+import org.osgi.service.event.EventHandler;
+import org.osgi.service.event.propertytypes.EventTopics;
 
 /**
  * Starts the event loop monitoring thread. Initializes preferences from {@link IPreferenceStore}.
  */
-public class MonitoringStartup implements IStartup {
+@Component(service = EventHandler.class)
+@EventTopics(value = UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
+public class MonitoringStartup implements EventHandler {
```
laeubi, line 36: You can even define a new composite annotation like this

```
@Component(service = EventHandler.class)
@EventTopics(value = UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
public @interface StartupComponent {

}
```

what the can be used like this 
```suggestion
@StartupComponent
public class MonitoringStartup implements EventHandler {
```

bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringStartup.java
```diff
@@ -17,21 +17,27 @@
...
+import org.osgi.service.event.Event;
+import org.osgi.service.event.EventHandler;
+import org.osgi.service.event.propertytypes.EventTopics;
 
 /**
  * Starts the event loop monitoring thread. Initializes preferences from {@link IPreferenceStore}.
  */
-public class MonitoringStartup implements IStartup {
+@Component(service = EventHandler.class)
+@EventTopics(value = UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
```
laeubi:
```suggestion
@EventTopics(UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
```

HannesWell: Done.
This was a left-over while investigating https://github.com/eclipse-pde/eclipse.pde/issues/973 a bit.

bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringStartup.java
HannesWell: If I try to add for example to the `UILifeCycle` class:
```
@Component(service = EventHandler.class)
@EventTopics(value = APP_STARTUP_COMPLETE)
public static @interface StartupCompleted {
}
```
I get the error `Invalid component implementation class 'StartupCompleted': annotation type.`

laeubi: Maybe `@Component` can't be placed on meta annotation, maybe its just a bug in PDE-DS ...

sratz:
An `IStartup`s execution can be prevented via _Window -> Preferences -> General -> Startup and Shutdown -> Plug-ins activated on startup_ ('activated' is a bit misleading here, but the list contains all plug-ins contributing an `IStartup` extension and controls whether those are actually started or not).

This gives users a choice to enable or disable certain startups.

I suppose the OSGI mechanism does not respect the preference page?

In that case I don't think `EventHandler` can be considered a drop-in replacement over `IStartup` and `IStartup` does have its right to exist. The Javadoc should probably be updated to make this distinction of whether an end-user may disable it or not.

vogella:
I'm in favor of deprecating IStartup and marking it for deletion. The user option to disable IStartup was always very dangerous in my opinion as this would allow the user to disable potential critical infrastructure.

sratz:
> I'm in favor of deprecating IStartup and marking it for deletion. The user option to disable IStartup was always very dangerous in my opinion as this would allow the user to disable potential critical infrastructure.

I'm not sure, I consider the preference a feature, e.g. we have a 'Release notes' feature for our product that is triggered via an `IStartup`, which users can disable there if they don't like it.

With the two alternatives we now have to chance to separate mandatory/critical startups from optional ones and convert those to OSGI event handlers.

My vote would be to state the fact in the Javadoc of `IStartup` and give providers the choice.

laeubi:
> I'm not sure, I consider the preference a feature, e.g. we have a 'Release notes' feature for our product that is triggered via an IStartup, which users can disable there if they don't like it.

I think that would be **much** more suitable to have a custom preference (page) for that that having to go into some generic dialogs.

vogella:
> > I'm not sure, I consider the preference a feature, e.g. we have a 'Release notes' feature for our product that is triggered via an IStartup, which users can disable there if they don't like it.
> 
> I think that would be **much** more suitable to have a custom preference (page) for that that having to go into some generic dialogs.

+1

BeckerWdf:
> > I'm not sure, I consider the preference a feature, e.g. we have a 'Release notes' feature for our product that is triggered via an IStartup, which users can disable there if they don't like it.
> 
> I think that would be **much** more suitable to have a custom preference (page) for that that having to go into some generic dialogs.

But that's we is out there today. I assume other RCP products may also use this feature. If platform now changes how this behaves: this is incompatible and will force RCP implementers to adapt their code. And what's the value of it?

Besides that: I as a user find it quite nice to have one central place where I can turn off / on such start ups.

vogella:
FWIW - I have never seen (nor recommended) using IStartup in RCP applications. The ability to disable them, make it impossible to rely on the functionality for the regular use case that "something should start early".

vogella:
@HannesWell can we apply this? The recommendation is fine for now IMHO. Once we have removed the usage of IStartup from p2 we can discuss to deprecate it, even though maybe not mark it for deletion yet due to the concerns from @BeckerWdf and @sratz. But it might be good to guide new users to use the new approach).

Please also add to N&N so that people can find it.

HannesWell:
> An `IStartup`s execution can be prevented via _Window -> Preferences -> General -> Startup and Shutdown -> Plug-ins activated on startup_ ('activated' is a bit misleading here, but the list contains all plug-ins contributing an `IStartup` extension and controls whether those are actually started or not).
> 
> This gives users a choice to enable or disable certain startups.

Thank you @sratz for mentioning this, I didn't knew it. And yes, at least to my knowledge, one cannot disable the execution of OSGi event handlers.

And like Lars I would consider this, at least in general, a dangerous feature if it is used for critical tasks. And since it was not mention in the doc, I'm not sure if everyone that uses it is aware of it (I wasn't).
Like Christoph suggested I think it would be more suitable to have dedicated and explicit preferences for tasks that are optional and where the user should be able to disable it.

> Besides that: I as a user find it quite nice to have one central place where I can turn off / on such start ups.

To some extend I understand that demand. But not everything that runs at start-up currently implements IStartup. Many things happens in Activators or code triggered by activators. So you will only see a sub-set of everything any ways.

> @HannesWell can we apply this? The recommendation is fine for now IMHO. Once we have removed the usage of IStartup from p2 we can discuss to deprecate it, even though maybe not mark it for deletion yet due to the concerns from @BeckerWdf and @sratz. But it might be good to guide new users to use the new approach).
> 

Applying this to the Freeze-Monitor currently requires the latest Features in the PDE DS support which was just implemented by @laeubi in https://github.com/eclipse-pde/eclipse.pde/pull/923. It is currently only available only available in I-builds.
I don't know if it is an acceptable requirement for other platform contributors?
Furthermore with the objections raised we have to think again if the Freeze-Monitor should possible be disabled by the user?

HannesWell:
I also changed the language in the doc to be less strict.

> Furthermore with the objections raised we have to think again if the Freeze-Monitor should possible be disabled by the user?

In fact there is already a dedicated preference:
grafik

So it would just be the question if it is OK to require the latest PDE DS version for developing the platfrom, but if one only uses the latest milestone in most cases it is probably also ok if the freeze-monitoring does not work temporarily.

vogella:
Lets reduce our usage of IStartup, the deprecation may come later, we may want to "only" deprecate  for now and not mark for deletion so that new usages of this "feature" are avoided.

PR #1364, opened 2023-12-04
Wrapped the OperationCanceledException in InterruptedException

amartya4256:
https://github.com/eclipse-platform/eclipse.platform/pull/853

Wrapped the OperationCanceledException in InterruptedException

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/operation/ModalContext.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/operation/ModalContext.java
@@ -434,7 +434,9 @@ public class ModalContext {
 				runnable.run(progressMonitor);
 			}
 		} catch (OperationCanceledException e) {
-			throw new InterruptedException();
+			InterruptedException interruptedException = new InterruptedException(e.getLocalizedMessage());
+			interruptedException.initCause(e);
+			throw interruptedException;
 		} catch (InvocationTargetException | InterruptedException | ThreadDeath e) {
 			// Make sure to propagate ThreadDeath, or threads will never fully
 			// terminate.


merks:
I'm not sure this is a generally and universally necessary change.

In any case, please don't merge in commits to your PR.  Keep the PR with a single commit and rebase on master when necessary.  The PR in this state current cannot be properly reviewed or processed...

merks:
@amartya4256 

That's better.  😄 


@HeikoKlare 

Are such changes generally wanted?

HeikoKlare:
I think this kind of change makes sense, as there are several points in the code at which `InterruptedExceptions` are thrown in order to make a checked exception out of an `OperationCanceledException` without containing information about the original exception. This makes it impossible to eventually find out where the exception was originally thrown and what its cause was (an actual operation cancellation or, for example, an interrupted thread).

As far as I know, the first kind of change improving this exception handling was https://github.com/eclipse-platform/eclipse.platform.ui/commit/d8c2f8631708c3420e9d5b2581d2dbcd66153dbf
And another recent change that was driven by need to finding out the reasons for a thrown exception was https://github.com/eclipse-platform/eclipse.platform.ui/pull/1259

Do you see any negative impact in these changes, @merks?

merks:
I don’t see anything harmful. It’s just annoying (very verbose) that there is no convenient constructor. Which begs the question why there is no such constructor?  Is it an unusual pattern to throw this type of exception as being caused by something else?  Might we encapsulate the conversation somewhere?  E.g., a toInterruptedException method on OperationCanceledException?

HeikoKlare:
First: You are right, the repetition is verbose and error prone and we should choose some better solution.

This kind of exception wrapping seems to exist for a long time and I do not completely understand why. The obvious thing is that `OperationCanceledExceptions` are unchecked while `InterruptedExceptions` are checked and maybe `InterruptedExceptions` were chosen as a wrapper to make the exception checked as they were the semantically "best" fitting one (but that's only my guess). That's why I think this pattern is rather unusual and can only be found when wrapping `OperationCanceledExceptions` into checked ones. And since the checked exceptions are part of the API, we can also not exchange them easily.

To avoid the repetition, I would propose the following: add some wrapping method to the `OperationCanceledException`, such as `InterruptedException wrapIntoInterruptedException()` or `void rethrowAsInterruptedException() throws InterruptedException`. We cannot provide a convience constructor for the `InterruptedException`, as it's part of the JDK, so we can only provide the functionality at the `OperationCanceledException` or at some independent place such as a factory class.

merks:
Yes I would be much happier to see a concise, single-line replacement.  The rethrow method is interesting but the compiler won’t know that flow will not continue as it knows when there is an explicit throw statement

HannesWell:
> Yes I would be much happier to see a concise, single-line replacement.  The rethrow method is interesting but the compiler won’t know that flow will not continue as it knows when there is an explicit throw statement 

That's right. But if the wrap method returns an exception one can just write`throw` before the method call.

merks:
Exactly. That is why I think a method that returns the wrapping exception would be ideal.

laeubi:
> Which begs the question why there is no such constructor?

The answer is simple, because this exception is actually not to be created by user code except you have cleared the interrupted **flag(!)** so it can never be really "caused" by something see:
https://docs.oracle.com/javase/8/docs/api/java/lang/InterruptedException.html

>> Thrown when a **thread** is waiting, sleeping, or otherwise occupied, **and the thread is interrupted**, either before or during the activity. 

Nothing of those is happening here! Sadly InterruptedException is misuses at several places in eclipse while the proper one to use would be [CancellationException](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/CancellationException.html
) but also in this case there is no causing exception because this can't be caused by another exception, in such case one has to use [ExecutionException](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ExecutionException.html) a good example is the method [get(long timeout, TimeUnit) unit)](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/FutureTask.html#get(long,%20java.util.concurrent.TimeUnit)) of future task 

> Is it an unusual pattern to throw this type of exception as being caused by something else? Might we encapsulate the conversation somewhere? E.g., a toInterruptedException method on OperationCanceledException?

I would avoid / replace this entirely instead of making it more convenient to follow this wrong pattern.

HeikoKlare:
> I would avoid / replace this entirely instead of making it more convenient to follow this wrong pattern.

I agree that no new code should follow this pattern, but we cannot easily change existing code either, as in some cases it affects public API (the `InterruptedException` is part of the method signature), also see https://github.com/eclipse-platform/eclipse.platform/pull/853#issuecomment-1808244183

Also note that this misused exception is only passed outside internal code rather rarely. If I remember correctly we now have around 5 places touched in Platform and Platform UI code. 

> Nothing of those is happening here! Sadly InterruptedException is misuses at several places in eclipse while the proper one to use would be [CancellationException](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/CancellationException.html) but also in this case there is no causing exception because this can't be caused by another exception, in such case one has to use [ExecutionException](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ExecutionException.html) a good example is the method [get(long timeout, TimeUnit) unit)](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/FutureTask.html#get(long,%20java.util.concurrent.TimeUnit)) of future task

It's correct that these exception would fit better, but please also note that they have been introduced into Java after at least some of this Eclipse code has been written initially.

bundles/org.eclipse.search/new search/org/eclipse/search2/internal/ui/InternalSearchUI.java
```diff
@@ -228,7 +228,9 @@ private IStatus doRunSearchInForeground(final SearchJobRecord rec, IRunnableCont
 						throw new InvocationTargetException(new CoreException(status));
 					}
 				} catch (OperationCanceledException e) {
-					throw new InterruptedException();
+					InterruptedException interruptedException = new InterruptedException(e.getLocalizedMessage());
+					interruptedException.initCause(e);
+					throw interruptedException;
```
HeikoKlare: In this case, the exception is only misused to transfer an event. This code is inside a try-catch that transforms the exception into a status value, so there is no benefit in adding further information to it:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/6e1920564ae8a68e344f22c3d5c37a2243887e13/bundles/org.eclipse.search/new%20search/org/eclipse/search2/internal/ui/InternalSearchUI.java#L220-L244

HeikoKlare (changes requested):

laeubi:
> It's correct that these exception would fit better, but please also note that they have been introduced into Java after at least some of this Eclipse code has been written initially.

But even before it was wrong to use that and eclipse has `OperationCanceledException` for the good old days... also just because its in the API it is not appropriate to use it for different purpose ;-)

HeikoKlare:
> But even before it was wrong to use that and eclipse has `OperationCanceledException` for the good old days... also just because its in the API it is not appropriate to use it for different purpose ;-)

I totally agree with you :-) I just tried to understand (and partly justify) the decision made long years ago. But, anyway, we cannot change that anymore, so my primary goals would be to (1) avoid that this kind of misuse is repeated in new code and (2) that we make the best of the existing code in terms of providing information about the wrapped exception in case it is provided to non-internal code.

HeikoKlare (approved):

HeikoKlare:
@amartya4256 can you please rebase on master instead of merging master into this PR?

bundles/org.eclipse.jface/src/org/eclipse/jface/operation/ModalContext.java
```diff
@@ -435,7 +435,9 @@ private static void runInCurrentThread(IRunnableWithProgress runnable, IProgress
 				runnable.run(progressMonitor);
 			}
 		} catch (OperationCanceledException e) {
-			throw new InterruptedException();
+			InterruptedException interruptedException = new InterruptedException(e.getLocalizedMessage());
```
szarnekow, line 437: I miss a systematic approach here. In the same file `org.eclipse.jface.operation.ModalContext.run(IRunnableWithProgress, boolean, IProgressMonitor, Display)` also created an InterruptedException. Why isn't that code path updated?

HeikoKlare: That code path was already updated some weeks ago to add the causing exception (the `OperationCanceledException`) to the `InterruptedException`. The systematics in this PR is that we went through all the Platform UI code to see where this pattern of misuse of `InterruptedException` happened as well (and was not "improved" yet). The place touched here is the only remaining one we found.
We already discussed systematic ways of wrapping the `OperationCanceledException` in this PR, but since we agree that this pattern is just legacy and should not be implemented anywhere else, I agree with @laeubi (https://github.com/eclipse-platform/eclipse.platform.ui/pull/1364#issuecomment-1839991041) that we should not make this pattern more convenient to implement.

szarnekow: Oh, my bad. I was looking at a different local version of that file.

PR #1372, opened 2023-12-05
Format edited lines in org.eclipse.ltk.core.refactoring

vogella:
Common setting on most platform plug-ins should also applied to migrated plug-ins from JDT


vogella:
Know test errors on Linux, planning to merge once good old slow Mac is done.

PR #1375, opened 2023-12-05
Fix Javadoc view not starting

akurtakov:
Null values passed are preserved and sent to HTMLBuilder which sets them to default colors if needed.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/HTMLPrinter.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/HTMLPrinter.java
@@ -63,6 +63,11 @@ public class HTMLPrinter {
 	}
 
 	private static org.eclipse.text.html.RGB fromRGB(RGB val) {
+		// Preserve	null RGB as HTMLBuilder contains the default colors and sets them accordingly
+		// in case of null parameter passed
+		if (val == null) {
+			return null;
+		}
 		return new org.eclipse.text.html.RGB(val.red, val.green, val.blue);
 	}
 	private static void cacheColors(Display display) {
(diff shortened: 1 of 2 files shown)


akurtakov:
The actual code fix  for https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/973

akurtakov:
Not the brightest fix but better than copy this code everywhere fromRGB is used.

akurtakov:
@iloveeclipse What's your opinion here?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/HTMLPrinter.java
```diff
@@ -63,6 +63,11 @@ private static void installColorUpdater(final Display display) {
 	}
 
 	private static org.eclipse.text.html.RGB fromRGB(RGB val) {
+		// Preserve	null RGB as HTMLBuilder contains the default colors and sets them accordingly
```
iloveeclipse, line 66: The problem here is the second public constructor that initializes everything **without** null check. Means, this would not work for client code that creates `HTMLBuilder(RGB bg, RGB fg, RGB link, RGB alink)` with some "null" color.
I haven't reviewed original PR that introduced this constructor, and it should not be used yet widely. So I think this PR should add some assert in that constuctor.

akurtakov: The NPE seen happens before even getting into HTMLBuilder but I'll add guards for that too.

akurtakov: No null colors now with latest revision.

akurtakov:
Once verifications are green I plan to merge this one unless there are objections as this is will fix the immediate issue. Further fixes/improvements can be done after that in other PRs.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/HTMLPrinter.java
iloveeclipse: Thanks

iloveeclipse (approved):

akurtakov:
Unrelated test failure on MacOS only.

PR #1379, opened 2023-12-05
[memory] optimize e4 ClassUtils.getPackageName()

jukzi:
to return always the same String Instance
Saves all duplicate Strings in
org.eclipse.e4.ui.css.swt.dom.WidgetElement.namespaceURI

it's only 1 MB.. but why not...
image


vogella:
Cool, LGTM. thanks.

PR #1381, opened 2023-12-06
Add "Advanced Text Editor" strategy to the list of possible strategies

### LINKED ISSUE #1380: Add "Generic Text Editor" strategy to the list of possible strategies
I'm trying to use tm4e as a lightweight universal syntax highlighting addon for generic editor for all possible file types one can see flying around, but I'm hindered by the fact that I have to associate each and every "unassociated" file type with generic editor manually.

Platform provides a way to do so, but currently it only provides a choice between "Text Editor", "System Text Editor", "System or Text ..." and "Ask...".

So the easiest way to get "generic" syntax support for most of the used file types is to extend the options provided in this selection box to "Generic Text Editor", which offers more functionality as "Text".

I think this shouldn't harm anyone to have more available choices:

image

I will push a PR with proposal. With that one, opening python files with nothing else installed in the SDK shows immediately syntax (and of course for many other supported files).

iloveeclipse:
Extend the options provided in "Open unassociated files with:" selection box to "Advanced Text Editor" (and "System Editor, if none: Advanced Text Editor"), which offer more functionality as "Text" by default.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1380

### DIFF
--- a/bundles/org.eclipse.ui.genericeditor/plugin.xml
+++ b/bundles/org.eclipse.ui.genericeditor/plugin.xml
@@ -260,4 +260,19 @@
             contentType="org.eclipse.core.runtime.text">
       </reconciler>
    </extension>
+   <extension
+         point="org.eclipse.ui.ide.unassociatedEditorStrategy">
+      <strategy
+            class="org.eclipse.ui.internal.genericeditor.SystemEditorOrGenericTextEditorStrategy"
+            id="org.eclipse.ui.internal.genericeditor.systemEditorOrGenericEditor"
+            interactive="false"
+            label="%systemEditorOrGenericEditorStrategy">
+      </strategy>
+      <strategy
+            class="org.eclipse.ui.internal.genericeditor.GenericTextEditorStrategy"
+            id="org.eclipse.ui.internal.genericeditor.genericTextEditor"
+            interactive="false"
+            label="%genericEditorStrategy">
+      </strategy>
+   </extension>
 </plugin>
--- a/bundles/org.eclipse.ui.genericeditor/plugin.properties
+++ b/bundles/org.eclipse.ui.genericeditor/plugin.properties
@@ -37,3 +37,5 @@ command.toggle.highlight.name = Toggle Highlight
 menu.source.label = Source
 gotoMatchingBracketCommand_name =  Go to Matching Bracket
 gotoMatchingBracketCommand_description = Moves the cursor to the matching bracket
+systemEditorOrGenericEditorStrategy=System Editor; if none: Advanced Text Editor
+genericEditorStrategy=Advanced Text Editor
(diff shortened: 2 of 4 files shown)


laeubi (approved):
This is a good improvement!

bundles/org.eclipse.ui.genericeditor/plugin.xml
```diff
@@ -260,4 +260,19 @@
             contentType="org.eclipse.core.runtime.text">
       </reconciler>
    </extension>
+   <extension
```
szarnekow, line 263: Is there a way to hide this UI contribution for certain file types in a product?

szarnekow: Use case:
Show `Advanced Text Editor` for XML files in the workspace, but never for a file extension `xyz` which is specifically targeted by IDE support.

szarnekow: Oh, I misread. I was under the assumption that this is also shown in the `Open With` but that's not the case, is it?

iloveeclipse: > Is there a way to hide this UI contribution for certain file types in a product?

What do you mean by "hide"? Hide from the drop down list? Not that I know.

> Show Advanced Text Editor for XML files in the workspace, but never for a file extension xyz which is specifically targeted by IDE support.

This extension is only for **not accociated** file types. So it will be used only if there is no **other** editor known for given file.

iloveeclipse: **And** it is **not default** for now.

iloveeclipse: So if you get it, the only change will be new option in the drop down, but **no behavior** change otherwise anywhere - until user / product select this strategy to be default one.

iloveeclipse: > Oh, I misread. I was under the assumption that this is also shown in the `Open With` but that's not the case, is it?

Nope.

mickaelistria:
While I welcome this change, I wonder whether we shouldn't try to find a better user-facing name for the generic editor. What about calling it "Generic Source Editor" or "Generic Code Editor" ?

iloveeclipse:
> While I welcome this change, I wonder whether we shouldn't try to find a better user-facing name for the generic editor. What about calling it "Generic Source Editor" or "Generic Code Editor" ?

Maybe you should start a dedicated discussion thread for that, to avoid hijacking this PR for a **longer** discussion :-)

mickaelistria:
Ok, I've open https://github.com/eclipse-platform/eclipse.platform.ui/issues/1385

Issue #811, opened 2023-06-09
Random failing ImportExistingProjectsWizardTest.test21FullBuildAfterImportedProjects


jukzi:
```
Expected no auto-build triggers expected:<0> but was:<1>
java.lang.AssertionError: Expected no auto-build triggers expected:<0> but was:<1>
	at org.eclipse.ui.tests.datatransfer.ImportTestUtils$TestBuilder.assertFullBuildWasDone(ImportTestUtils.java:97)
	at org.eclipse.ui.tests.datatransfer.ImportExistingProjectsWizardTest.test21FullBuildAfterImportedProjects(ImportExistingProjectsWizardTest.java:1059)
```

HeikoKlare:
Occurred again in recent I-Build: https://download.eclipse.org/eclipse/downloads/drops4/I20230723-1800/testresults/html/org.eclipse.ui.tests_ep429I-unit-cen64-gtk3-java20_linux.gtk.x86_64_20.html

PR #1383, opened 2023-12-06
Solve some "Discouraged access" by using API

jukzi:


jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform.ui/issues/811

PR #1384, opened 2023-12-06
some @SuppressWarnings("restriction") on PDE/JDT internals

jukzi:
To get rid of all "Discouraged access" warnings


jukzi:
ignoring unrelated https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #1389, opened 2023-12-07
Fix and enable MediaRulesTest:testMediaRule

### LINKED ISSUE #533: Enable CSS tests
Tracking ticket for:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1389

The PR has been created two years ago. Please check if it can/should be finalized or otherwise just close it.

amartya4256:
Enabled some disabled tests which aren't in broken state anymore and fixed other disabled tests.
https://github.com/eclipse-platform/eclipse.platform/issues/525

In the MediaRulesTest class, we test for if a CSS rule with @media tag is considered on parsing. Ideally we want the rule to be ignored.
Suppose we have a CSS string like:
@media screen, print {
					BODY { background-color: red }
					}
 In the previous version of the test (which was disabled), it was expected that the rule "BODY { background-color: red }" under @media tag would be completely ignored. But now the behaviour has changed. It seems like the rules under @media are loaded indeed on parsing but the attributes are ignored, i.e. it looks like "BODY {  }".

However, this behaviour doesn't have any side effect since a rule with no attributes neither change anything nor resets anything.

### DIFF
--- a/tests/org.eclipse.e4.ui.tests.css.core/src/org/eclipse/e4/ui/tests/css/core/parser/MediaRulesTest.java
+++ b/tests/org.eclipse.e4.ui.tests.css.core/src/org/eclipse/e4/ui/tests/css/core/parser/MediaRulesTest.java
@@ -15,12 +15,12 @@
 package org.eclipse.e4.ui.tests.css.core.parser;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import org.eclipse.e4.ui.tests.css.core.util.ParserTestUtil;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
-import org.w3c.dom.css.CSSRuleList;
 import org.w3c.dom.css.CSSStyleSheet;
 
 /**
@@ -29,18 +29,16 @@ import org.w3c.dom.css.CSSStyleSheet;
 public class MediaRulesTest {
 
 	@Test
-	@Disabled("//THIS TEST KNOWN TO FAIL Dec 16/08")
 	void testMediaRule() throws Exception {
 		String css = """
-			@media screen, print {
-			BODY { line-height: 1.2 }
-			}
-			Label { background-color: #FF0000 }""";
+				@media screen, print {
+				BODY { line-height: 1.2 }
+				}
+				Label { background-color: #FF0000 }""";
 		CSSStyleSheet styleSheet = ParserTestUtil.parseCss(css);
 		assertNotNull(styleSheet);
-		CSSRuleList rules = styleSheet.getCssRules();
-
-
-		assertEquals(1, rules.getLength());
+		assertEquals(2, styleSheet.getCssRules().getLength());
+		assertFalse(styleSheet.getCssRules().item(0).getCssText().contains("line-height"));
+		assertTrue(styleSheet.getCssRules().item(1).getCssText().contains("background-color"));
 	}
 }
--- a/tests/org.eclipse.e4.ui.tests.css.core/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.e4.ui.tests.css.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.e4.ui.tests.css.core
-Bundle-Version: 1.302.700.qualifier
+Bundle-Version: 1.302.800.qualifier
 Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.swt,
  org.eclipse.e4.ui.css.core,


tests/org.eclipse.e4.ui.tests.css.core/src/org/eclipse/e4/ui/tests/css/core/parser/MediaRulesTest.java
```diff
@@ -29,18 +30,23 @@
...
-
+		List<String> cssRules = IntStream.range(0, styleSheet.getCssRules().getLength())
+				.mapToObj(i -> styleSheet.getCssRules().item(i).getCssText()).toList();
 
-		assertEquals(1, rules.getLength());
+		assertEquals(1, cssRules.stream().filter(it -> it.contains("line-height")).toList().size());
+		assertEquals(1, cssRules.stream().filter(it -> it.contains("background-color"))
+				.toList().size());
+		assert (cssRules.contains("background-color: #FF0000"));
+		assert (cssRules.contains("line-height: 1.3"));
```
fedejeanne: This one is the wrong `assert`, you probably meant `assertTrue`. In any case, please use HarmCrest / AssertJ assertions, they read better.
```java
import static org.assertj.core.api.Assertions.assertThat;
...
      assertThat(cssRules).contains(...); // you could also use "containsOnly" or some other one
```

fedejeanne (changes requested):

fedejeanne:
The failure of `org.eclipse.ui.tests.progress.ProgressViewTests.testItemOrder` might be unrelated to your changes (and it probably is). Please look for any existing issue on that (and, if there is none, please create one) and mention it here.

tests/org.eclipse.e4.ui.tests.css.core/src/org/eclipse/e4/ui/tests/css/core/parser/MediaRulesTest.java
```diff
@@ -29,18 +30,22 @@
...
-			@media screen, print {
-			BODY { line-height: 1.2 }
-			}
-			Label { background-color: #FF0000 }""";
+				@media screen, print {
+					BODY { line-height: 1.2 }
+					Label { background-color: #FFFFFF }
+				}
+				BODY { line-height: 1.3 }
+				Label { background-color: #FF0000 }""";
```
fedejeanne, line 37: You tested the same thing twice here: that an attribute outside the `@media` is not overwritten by those inside the `@media`.
You could test this only once and also test that the order in which the `@media` statements appear doesn't play any role _i.e._ use this...

```suggestion
@media screen, print {
	BODY { line-height: 1.2 }
	Label { background-color: #FFFFFF }
}
BODY { line-height: 1.3 }

@media screen, print {
	BODY { line-height: 1.4 }
}
""";
```

... and adapt the assertions (see below)

tests/org.eclipse.e4.ui.tests.css.core/src/org/eclipse/e4/ui/tests/css/core/parser/MediaRulesTest.java
```diff
@@ -29,18 +30,22 @@
...
-
+		List<String> cssRules = IntStream.range(0, styleSheet.getCssRules().getLength())
+				.mapToObj(i -> styleSheet.getCssRules().item(i).getCssText()).toList();
 
-		assertEquals(1, rules.getLength());
+		assertThat(cssRules.stream().filter(it -> it.contains("background-color")).toList())
+		.containsOnly("Label { background-color: rgb(255, 0, 0); }");
+		assertThat(cssRules.stream().filter(it -> it.contains("line-height")).toList())
+				.containsOnly("BODY { line-height: 1.3; }");
 	}
```
fedejeanne:
```suggestion
	CSSStyleSheet styleSheet = ParserTestUtil.parseCss(css);
	assertNotNull(styleSheet);

	// This one is provided only inside the @media so it shouldn't be there
	assertThat(findCssRuleThatContains(styleSheet, "background-color")).isEmpty();

	// This one is provided outside the @media and it shouldn't be overwritten by
	// any of them
	assertThat(findCssRuleThatContains(styleSheet, "line-height")).containsOnly("BODY { line-height: 1.3; }");
}

private Stream<String> findCssRuleThatContains(CSSStyleSheet styleSheet, String text) {
	Stream<String> cssRulesText = IntStream.range(0, styleSheet.getCssRules().getLength())
			.mapToObj(i -> styleSheet.getCssRules().item(i).getCssText());
	return cssRulesText.filter(r -> r.contains(text));
}
```

I extracted a private method to improve readability.

fedejeanne (changes requested):
Better. You could add a use case that checks that the order doesn't play a role and extract a private method to improve readability

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ButtonElement.java
```diff
@@ -78,6 +78,9 @@ public boolean isPseudoInstanceOf(String s) {
 		if ("checked".equalsIgnoreCase(s)) {
 			return this.isSelected;
 		}
+		if ("selected".equalsIgnoreCase(s)) {
+			return this.isSelected;
+		}
```
fedejeanne: Merge this two

tests/org.eclipse.e4.ui.tests.css.core/src/org/eclipse/e4/ui/tests/css/core/parser/MediaRulesTest.java
fedejeanne: You still need to change the content of `String css`

tests/org.eclipse.e4.ui.tests.css.swt/src/org/eclipse/e4/ui/tests/css/swt/CTabItemTest.java
```diff
@@ -356,7 +356,6 @@ void testSelectedShowClose2() {
 		}
 	}
 
-	@Disabled("test was commented before bug 443094")
```
fedejeanne: Does this test suddenly work again? Are you sure it doesn't fail randomly?

fedejeanne (changes requested):

bundles/org.eclipse.e4.ui.css.swt/src/org/eclipse/e4/ui/css/swt/dom/ButtonElement.java
fedejeanne: Actually: is `checked` even valid? I couldn't find any usage in any CSS file

vogella: I think it is valid, IIRC

fedejeanne: You're right, it's a valid selector: https://www.w3schools.com/cssref/sel_checked.php

tests/org.eclipse.e4.ui.tests.css.swt/src/org/eclipse/e4/ui/tests/css/swt/CTabItemTest.java
```diff
@@ -356,7 +356,6 @@ void testSelectedShowClose2() {
...
 	@Test
```
fedejeanne: Let's assume this one is still not fixed, ok? If you want to activate it, you can do it on a separate PR and monitor the tests the following day. 
```suggestion
	@Disabled("test was commented before bug 443094")
	@Test
```

fedejeanne (approved):

fedejeanne (dismissed):
`testSelectPseudo` is failing. Please investigate and fix or undo changes in it

HeikoKlare (approved):
The reenabled test is properly and successfully executed in the CI runs.

@fedejeanne any concerns remaining on your side?

fedejeanne:
> @fedejeanne any concerns remaining on your side?

No concerns

PR #1392, opened 2023-12-09
Replace deprecated PluginVersionIdentifier

HannesWell:


tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/keys/Bug36420Test.java
```diff
@@ -78,9 +76,7 @@ public void testImportKeyPreferences() throws CoreException,
 		// of bundles
 		String[] pluginIds = Platform.getExtensionRegistry().getNamespaces();
 		for (String pluginId : pluginIds) {
-			preferences.put(pluginId, new PluginVersionIdentifier(
```
HannesWell, line 81: I wonder if this ever worked before because the Properties class assumes that all keys and values are actually Strings when storing the properties?

laeubi: `Properties` extends `Map<Object, Object>` so as long as you do not call `getProperty("...")` and use `map.get(...)` instead it would work.

HannesWell: But Properties.store0() (which is called by `store()`) casts each entry to a `String`.

Any way, this seems to work as well. :)

HannesWell:
Test failures are unrelated.

PR #1396, opened 2023-12-11
Replace apitools javadoc tags with annotations

akurtakov:
Identified in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1655 that so far annotations were missing in generated .api_description files.
Local build shown that everything is still picked up with annotations only, even the one that wasn't before (LazyResourceManager) and caused this issue.


akurtakov:
Unrelated failure on MacOS only.

Issue #1180, opened 2023-09-29
Onboarding widgets cover Editor-Content after additional WorbenchWindow-Close


Rainer7000:
This Problem is a bit tricky to reproduce, because it involves a few preconditions.
The Problem was introduced with Eclipse 2023-09, alle previous versions workt fine.
I used the old EMail-Example-Application in Eclipse3 compatibility as a starting-point to reproduce it.
I have a Toolbar-Action that opens verry simple EditorParts (just a left and a right label) which works fine.
A second Toolbar-Action opens an additional Toolbar-Window with a different Perspective.
If the additional Toolbar-Window is closed while the editorArea of the primary WorkbenchWindow ist empty,
the content of the editor is never rendered (it's not a sizing/layouting problem, i checked that with a ControlListener).
The Problem does not occur with ViewParts instead of EditorParts and it also does not occur when one EditorPart 
is still visible when the additional window ist closed.
![rcp_editors](https://github.com/eclipse-platform/eclipse.platform.ui/assets/57557879/efbaa701-5cdb-4c92-bd73-0756d6e12746)
![rcp_window_open](https://github.com/eclipse-platform/eclipse.platform.ui/assets/57557879/b3623b2d-5a77-41cd-92dd-72da22e5b6a2)
![rcp_editors_empty](https://github.com/eclipse-platform/eclipse.platform.ui/assets/57557879/3bc5b85b-6fcf-4167-9a3f-80f957b39416)

tivervac:
@Rainer7000 We might be seeing similar issues. What does `not rendering` mean? In our case, events are still being processed (moving windows, copying and pasting...) but nothing is drawn. Making the window smaller than larger again just causes the resized parts to go full black

Rainer7000:
@tivervac Yes, I think this is the same behaviour as in our case. The content of the editor tab (Composites, Labels etc.) gets created but nothing is drawn. The tab in the third image in my original post stays gray but should look like the tab in the first image with a left and right label on white background.

Rainer7000:
After coloring my test-composites (left=red, right=green) I found out that after closing the second window there is a area with the color of the window background drawn in front of everything. You can see this if you enlarge the first window after the effect occours. It seems to have the size of the editor-content when the second window is closed.
I hope this helps to find the bug.
[...]

PR #1400, opened 2023-12-12
Fix onboarding - don't keep onboarding related controls as state

marcushoepfner:
Issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1180

The StackRenderer may be used several times e.g. when a second Eclipse window is opened via Menu -> Window -> new Window. When this second window is closed, the onboarding controls are disposed. When these controls have been kept as state (fields), the logic does not work anymore any strange ui behaviour can be observed. E.g. Onboarding controls are visible even when an editor is open.

Solution: remove the state and get the controls from the model element. In order to find them easier the controls store an ID in their data attribute.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.e4.core.contexts.IEclipseContext;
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -249,13 +258,32 @@ void subscribePerspectiveSwitched(
...
-				.filter(c -> c != onboardingText).forEach(Control::dispose);
+		Label onboardingImage = getChild(onboardingComposite, ONBOARDING_IMAGE, Label.class);
+		Label onboardingText = getChild(onboardingComposite, ONBOARDING_TEXT, Label.class);
+
+		if (onboardingImage == null || onboardingImage.isDisposed() || onboardingText == null
+				|| onboardingText.isDisposed()) {
+			return;
+		}
+
+		Stream<Control> commandsStream = Arrays.stream(onboardingComposite.getChildren())
```
BeckerWdf: you could inline "commandsStream" in the same way as in the previous version

BeckerWdf (approved):

BeckerWdf (approved):

BeckerWdf (approved):

PR #1402, opened 2023-12-12
Workaround for changed IStatusLineManager.getProgressMonitor() API

### LINKED ISSUE #1398: IllegalStateException reported in the log on opening files or navigating in browser
- Start Eclipse 4.30
- Open any svg file 
- Warning below reported into the log
```
java.lang.IllegalStateException: call to beginTask() missing
	at org.eclipse.jface.action.StatusLineManager$1.worked(StatusLineManager.java:221)
	at org.eclipse.ui.internal.browser.BrowserViewer$3.changed(BrowserViewer.java:421)
	at org.eclipse.swt.browser.WebKit.lambda$15(WebKit.java:2492)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5039)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4519)
```

Same in green:
- Start Eclipse 4.30 on Linux
- Open "Internal browser"
- Set any url and go there
- **Multiple** warnings like above reported in the log

This is a regression/side effect of https://github.com/eclipse-platform/eclipse.platform.ui/pull/1077.

iloveeclipse:
The change in 7fe83e4b8b7d62b0b8b31971389599afaf0a27ca broke method contract from `IStatusLineManager.getProgressMonitor()`. It was originally stating "returns monitor", not it says "creates new monitor".

So the browser listener is not able now to use
`IStatusLineManager.getProgressMonitor()` directly and has to keep a local monitor instance.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1398

### DIFF
--- a/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
+++ b/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
@@ -399,6 +399,8 @@ public class BrowserViewer extends Composite {
 		});
 
 		browser.addProgressListener(new ProgressListener() {
+			private IProgressMonitor monitor;
+
 			@Override
 			public void changed(ProgressEvent event) {
 					//System.out.println("progress: " + event.current + ", " + event.total); //$NON-NLS-1$ //$NON-NLS-2$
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Tested on  RHEL 7.9 Linux (Webkit) / Windows 11 (IE), works.
Note: the original code works on Windows (IE) because there are only calls to `beginTask()` and `done()` on the monitor, never calls to `worked()`.

jukzi:
> It was originally stating "returns monitor", not it says "creates new monitor".

The implementation returned a new instance before that. There is so many code in the eclipse codebase where "get" creates something.. . The only change was the check and a proper documentation.

The sourcecode of org.eclipse.help.ui.internal.views.BrowserPart seems to have the same error.

iloveeclipse:
> The sourcecode of org.eclipse.help.ui.internal.views.BrowserPart seems to have the same error.

Thanks, I will provide a patch. To reproduce: open Help view and play around with navigation... Same misery...

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform/pull/1002

PR #1412, opened 2023-12-14
Remove dead call in ShellActivationListener

vogella:
No need to call an empty method, move the documentation comment to the previous method call


vogella:
Know unrelated test error https://github.com/eclipse-platform/eclipse.platform.ui/issues/275

PR #1413, opened 2023-12-14
Apply Make inner classes static where possible clean-up

vogella:
Using "Make inner classes static where possible" JDT UI clean-up on platform UI


akurtakov:
Mac and Windows tests are failing. Do anyone care about these OSes enough to look into fixing tests on them?

PR #186, opened 2022-07-05
Mark ShowViewHandler(boolean makeFast) for deletion

vogella:
Fast views are not supported since multiple years in the Eclipse IDE,
hence lets mark the constructur for this for removal.

PR #1423, opened 2023-12-15
Showviewhandler deprecate useless constructor

vogella:


vogella:
Replacement for https://github.com/eclipse-platform/eclipse.platform.ui/pull/186

PR #1426, opened 2023-12-15
Enable viewer limits by default

### LINKED ISSUE #1425: Enable viewer limits by default
As discussed in #818, we had few open issues and/or concerns regarding "viewer limit" functionality merged in 4.30. 
It was decided to postpone enablement to the next release M1.

Here we are, I will push a PR that sets the default limit to 1000, which is what we believe the most practical value on most platforms to avoid UI freezes in large viewers.

iloveeclipse:
Set default workbench preference for the view limit to 1000, which is what we believe the most practical value on most platforms to avoid UI freezes in large viewers.

Note: this limit can also be disabled in products by specifying `org.eclipse.ui.workbench/largeViewLimit=0` in the plugin_customization.ini file.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1425


vogella:
> Note: this limit can also be disabled in products by specifying org.eclipse.ui.workbench/largeViewLimit=0 in the plugin_customization.ini file.

This MUST be added to the N&N so that people can find this.

PR #1428, opened 2023-12-18
Fix LineAnnotationManagerTest + random failing PositionTrackerTest #882

### LINKED ISSUE #882: org.eclipse.search.tests.filesearch.PositionTrackerTest fails randomly

HeikoKlare:
The LineAnnotationManagerTest is broken. Its LineBaseFileSearch throws exceptions leading to no search results being provided to the test to validate, which in turn results in a succeeding test. This also leads to subsequently executed search (or tests relying on them) to fail. Because of the current configuration of the AllFileSearchTests test suite, the PositionTrackerTest as the subsequent one randomly fails.

This change brings the LineAnnotationManagerTest into a state that (1) does not produce (silent) exceptions, (2) actually operates on search results and (3) successfully validates their representation in the annotation model.
To this end, it does the following:
- Make the LineBaseFileSearch thread safe, as it is used concurrently
- Provide a dummy LineElement for the search results, as the currently passed null element leads to an exception
- Make the LineAnnotationManagerTest validate on line positions rather than text positions, since that is what the annotation model actually contains
- Improve assertion messages in order to ease analyzing eventual further / upcoming test failures

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/882

### DIFF
--- a/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/LineAnnotationManagerTest.java
+++ b/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/LineAnnotationManagerTest.java
@@ -13,8 +13,9 @@
  *******************************************************************************/
 package org.eclipse.search.tests.filesearch;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.hasItem;
 
 import java.util.ArrayList;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


HeikoKlare:
Supercedes #1288

BeckerWdf (approved):

PR #1431, opened 2023-12-19
Properly assign values to fields in ResourceInfoPage

HeikoKlare:
The creation logic of the ResourceInfoPage creates two text input fields that are not assigned to the according fields of the class but stored as equally named local variables. This produces warnings and is obviously unintended as no value is assigned to the fields. So currently all logic defined for the case that the fields are non-null is never executed.


mickaelistria:
Thanks

PR #1432, opened 2023-12-19
Suppress warnings for fields hiding other fields

HeikoKlare:
There are several fields that hide fields of superclasses because of having the same name and produce according warnings. In some cases this is intended or necessary, so the warnings should be suppressed. These cases comprise:
* Class-specific IDs and default objects (like EMPTY)
* Fields hiding superclass fields that should not be accessed

This change adds the according suppressions.


jukzi (approved):
image

PR #1433, opened 2023-12-19
Fix several warnings due to name clashes

HeikoKlare:
Resolves warnings of kind "hiding field" by either
* removing the variable if it is unnecessary or a duplicate, or
* renaming the hiding or the hidden variable, or
* making the method static in case a parameter is hiding a field.

Together with #1431, #1432, this should resolve all warnings of type "hiding field" in this repository. To avoid faulty name replacements, all name changes have been performed via warnings-free refactorings,


HeikoKlare:
This change only applies refactorings to fix a number of warnings:
image

Since recent changes became conflicting to this PR, I would like to merge it soon to avoid further necessity of readaptation. In case there are any objections on merging, please let me know.

PR #1439, opened 2023-12-20
Enable eager resolve due to heavy interaction between platform+ui

laeubi:


.mvn/maven.config
```diff
@@ -1 +1,2 @@
 -Pbuild-individual-bundles
+-Dtycho.target.eager=true
```
iloveeclipse, line 2: Sorry if stupid question , but isn't that the opposite of PR message?

laeubi: Commit message was wrong, fixed that.

PR #1448, opened 2023-12-21
Api removal document

vogella:


vogella:
Fits better into eclipse-platform https://github.com/eclipse-platform/eclipse.platform/pull/1064

PR #1455, opened 2024-01-02
adopt standard platform check utility

elsazac:
1.Make use of standard utility function for platform check wherever applicable.
2.standardize field naming conventions

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
@@ -46,14 +46,14 @@ import org.eclipse.swt.widgets.Widget;
  * @since 3.4
  */
 public abstract class AbstractColumnLayout extends Layout {
-	private static int columnTrim;
+	private static int COLUMN_TRIM;
 	static {
 		if (Util.isWindows()) {
-			columnTrim = 4;
+			COLUMN_TRIM = 4;
 		} else if (Util.isMac()) {
-			columnTrim = 24;
+			COLUMN_TRIM = 24;
 		} else {
-			columnTrim = 3;
+			COLUMN_TRIM = 3;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
```diff
@@ -46,14 +46,14 @@
  * @since 3.4
  */
 public abstract class AbstractColumnLayout extends Layout {
-	private static int COLUMN_TRIM;
+	private static int columnTrim;
```
mickaelistria: Here, the goal with upper case is to emphasize that the value is in practice final and mustn't nor shouldn't be changed upon access. Please keep this convention.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TableLayout.java
```diff
@@ -39,21 +39,21 @@
...
-	 * To ensure there are N pixels available for the content of the column,
-	 * assign N+COLUMN_TRIM for the column width.
+	 * The number of extra pixels taken as horizontal trim by the table column. To
+	 * ensure there are N pixels available for the content of the column, assign
+	 * N+columnTrim for the column width.
 	 *
 	 * @since 3.1
 	 */
-	private static int COLUMN_TRIM;
+	private static int columnTrim;
```
mickaelistria: Same.

mickaelistria (changes requested):

bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
elsazac: I have reverted the case change, however this scenario looks strange to me - The field cannot be declared ~~static~~ final because it value neither known at compile time nor at construction time, and yet it holds a final value :)

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TableLayout.java
elsazac: ditto

bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
mickaelistria: The value is known at class loading time, the `static { ... }` block after it will be executed as soon as class is loaded, and will set the value dynamically. It's not actually, but the name convention is a reminder that this static field must be consumed as if it were final.

mickaelistria (approved):

PR #1459, opened 2024-01-03
replace anonymous runnable class with lambda

elsazac:


bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/link/contentassist/CompletionProposalPopup2.java
```diff
@@ -872,18 +872,14 @@ public void processEvent(VerifyEvent e) {
 	private void filterProposals() {
 		++ fInvocationCounter;
 		Control control= fViewer.getTextWidget();
-		control.getDisplay().asyncExec(new Runnable() {
+		control.getDisplay().asyncExec(() -> {
```
mickaelistria, line 876: This is not equivalent and will most likely cause bugs:

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/link/contentassist/CompletionProposalPopup2.java
```diff
@@ -872,18 +872,14 @@ public void processEvent(VerifyEvent e) {
...
 			long fCounter= fInvocationCounter;
```
mickaelistria: Before, fCounter is initialized just once, when `filterProposals` is invoked and the `Runnable` is created, ie before actually running the Runnable. While with the change, the value is set at execution of the Runnable, leading to...

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/link/contentassist/CompletionProposalPopup2.java
```diff
@@ -872,18 +872,14 @@ public void processEvent(VerifyEvent e) {
...
-			@Override
-			public void run() {
-
-				if (fCounter != fInvocationCounter)
```
mickaelistria, line 880: This condition not being always true (before) but being always true (after)

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContextInformationPopup.java
```diff
@@ -876,30 +876,24 @@ private void validateContextInformation() {
 		 */
 		if (!isValid(fContextInfoPopup))
 			return;
-
-		fContextInfoPopup.getDisplay().asyncExec(new Runnable() {
-
-			private ContextFrame fFrame= fContextFrameStack.peek();
```
mickaelistria: A similar pattern here: the frame is set once at instantiation for some reason, not on the execution. This looks like an actual intent to capture the "initial value before execution".

mickaelistria (changes requested):
I don't think this change is safe. Usually, a class that defines a field shouldn't be turned into a lambda as the field lifecycle (which value they get and where) matters.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/link/contentassist/CompletionProposalPopup2.java
elsazac: will moving `fCounter`  assignment outside of the lambda solve the problem?

mickaelistria: Yes, I think it would.

elsazac: @mickaelistria I have made that change accordingly and removed the problematic lambda. please have a look.

mickaelistria (approved):

mickaelistria:
Thank you!

PR #1462, opened 2024-01-04
simplify collection emptiness check

elsazac:
Use isEmpty() API of collections to simplify the emptiness check. reduces one extra comparison.


akurtakov (approved):
Thanks!

PR #1463, opened 2024-01-04
Leverage type inference for generics

elsazac:
Use <> for type inference wherever applicable

### DIFF
--- a/bundles/org.eclipse.core.databinding.observable/src/org/eclipse/core/databinding/observable/value/DateAndTimeObservableValue.java
+++ b/bundles/org.eclipse.core.databinding.observable/src/org/eclipse/core/databinding/observable/value/DateAndTimeObservableValue.java
@@ -116,7 +116,7 @@ public class DateAndTimeObservableValue extends AbstractObservableValue<Date> {
 	}
 
 	// One calendar per thread to preserve thread-safety
-	private static final ThreadLocal<Calendar> calendar = new ThreadLocal<Calendar>() {
+	private static final ThreadLocal<Calendar> calendar = new ThreadLocal<>() {
 		@Override
 		protected Calendar initialValue() {
 			return Calendar.getInstance();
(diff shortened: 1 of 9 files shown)


bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/application/ui/menu/util/MenuAdapterFactory.java
```diff
@@ -104,7 +104,7 @@ public boolean isFactoryForType(Object object) {
 	 * <!-- end-user-doc -->
 	 * @generated
 	 */
-	protected MenuSwitch<Adapter> modelSwitch = new MenuSwitch<Adapter>() {
+	protected MenuSwitch<Adapter> modelSwitch = new MenuSwitch<>() {
```
akurtakov: This is generated code so shouldn't be manually edited. @merks should there be some update in EMF itself or regeneration should help?

merks: The generator uses the configured cleanup actions which would make this change, but I it doesn't apply such an action if the generated file is unchanged by generation.  So better would be to do a manual cleanup of the entire project and then regenerate the model(s).

merks:
Please don't change generated files, i.e., files with ` * @generated` in the Javadoc.

I will apply the cleanup to projects with models in them and regenerate them to ensure they remain consistent.

merks:
FYI, I've done a general cleanup and regeneration via this PR:

https://github.com/eclipse-platform/eclipse.platform.ui/pull/1465

merks (changes requested):
Please remove the changes to generated files.  Those are handled by this PR:

https://github.com/eclipse-platform/eclipse.platform.ui/pull/1465

elsazac:
Thanks for the review @akurtakov @merks .
I reverted all the changes in `org.eclipse.e4.ui.model.workbench` bundle, assuming that's where all the generated code go ( I never knew they were generated code). Let me know if this is fine.

merks (approved):
It looks fine now.  Thanks.  😄

PR #1464, opened 2024-01-04
safe guard against NPE

elsazac:


mickaelistria:
Is a null check the best approach? Shouldn't the case of `tbsize == null` or `msize == null` throw an exception anyway? Or is the `null` case a possibly normal execution.

elsazac:
> Is a null check the best approach? Shouldn't the case of `tbsize == null` or `msize == null` throw an exception anyway? Or is the `null` case a possibly normal execution.

My idea was to make sure null cases do not break the code, as opposed to throwing exceptions from such a low level function that may not be handled downstream meaningfully. Do you think explicitly throwing exception here is a good idea?

laeubi:
At least `tbsize` is already checked for `null` before so it seems a valid case...

mickaelistria:
> Do you think explicitly throwing exception here is a good idea?

Note sure at all; that's why I'm asking your opinion here ;)

> At least tbsize is already checked for null before so it seems a valid case...

That's a good hint, thanks!

PR #1467, opened 2024-01-05
relax throws clause in overridden methods that do not throw

elsazac:
Refs: https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8.3

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/browser/DefaultWorkbenchBrowserSupport.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/browser/DefaultWorkbenchBrowserSupport.java
@@ -53,7 +53,7 @@ public class DefaultWorkbenchBrowserSupport extends AbstractWorkbenchBrowserSupp
 	}
 
 	protected IWebBrowser doCreateBrowser(int style, String browserId, String name, String tooltip)
-			throws PartInitException {
+	{
 		return new DefaultWebBrowser(this, browserId);
 	}
 
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/browser/DefaultWorkbenchBrowserSupport.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/browser/DefaultWorkbenchBrowserSupport.java
@@ -15,7 +15,6 @@ package org.eclipse.ui.internal.browser;
 
 import java.util.HashMap;
 import java.util.Map;
-import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.browser.AbstractWorkbenchBrowserSupport;
 import org.eclipse.ui.browser.IWebBrowser;
 
@@ -59,7 +58,7 @@ public class DefaultWorkbenchBrowserSupport extends AbstractWorkbenchBrowserSupp
 
 	@Override
 	public IWebBrowser createBrowser(int style, String browserId, String name, String tooltip)
-			throws PartInitException {
+	{
 		IWebBrowser browser = findBrowser(browserId == null ? getDefaultId() : browserId);
 		if (browser != null) {
 			return browser;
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/browser/DefaultWorkbenchBrowserSupport.java
```diff
@@ -53,7 +53,7 @@ IWebBrowser findBrowser(String id) {
 	}
 
 	protected IWebBrowser doCreateBrowser(int style, String browserId, String name, String tooltip)
```
laeubi, line 54: Please check the java warning

laeubi (dismissed):
It seems you need to adjust the javadoc as well then

elsazac:
> It seems you need to adjust the javadoc as well then

@laeubi looks like the javadoc of the overridden methods are dynamically adjusted as per the declaration.

before:
image (1)

after:
image (2)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/browser/DefaultWorkbenchBrowserSupport.java
elsazac: The java warnings are preexisting.

before:
image (3)

image (5)


after:
image (6)

laeubi:
Please check the build that shows new warnings, this maybe because you removed the exception from one place then another one becomes no longer throwing the exception as for example here:

https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1467/1/java/new/source.ddf8b0c4-8449-4f37-a47f-0accdbbf3117/#62

elsazac:
> Please check the build that shows new warnings, this maybe because you removed the exception from one place then another one becomes no longer throwing the exception as for example here:
> 
> https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1467/1/java/new/source.ddf8b0c4-8449-4f37-a47f-0accdbbf3117/#62

@laeubi Thanks for the clarification. Now I see where it is coming from. please have a look.

laeubi:
Build still shows one new compiler warning even though Jenkins seem to have problems parsing them:

https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1467/3/java/new/type.1251191584/

elsazac:
> Please check the build that shows new warnings, this maybe because you removed the exception from one place then another one becomes no longer throwing the exception as for example here:
> 
> https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1467/1/java/new/source.ddf8b0c4-8449-4f37-a47f-0accdbbf3117/#62

@laeubi we did a detailed investigation, and could not find any new warnings.
 - no of preexisting warnings : 2374
 - no of throws clause removed : 3
 - no of additional warnings introduced : 2
 - and hence no of additional throws clause removed : 2
 - no of current (with this commit) warnings : 2371
 
 I undone everything, and introduce the changes one by one and confirm that there is no additional warnings introduced.The yellow strip that Jenkins shows should contain three preexisting warnings (on unused parameters), probably it was hidden behind the throws clause warning and got highlighted after this commit - that's my theory.

laeubi:
> The yellow strip that Jenkins shows should contain three preexisting warnings (on unused parameters), probably it was hidden behind the throws clause warning and got highlighted after this commit

Its possible that this is now counted as new, can you just fix the warnings there?

elsazac:
> > The yellow strip that Jenkins shows should contain three preexisting warnings (on unused parameters), probably it was hidden behind the throws clause warning and got highlighted after this commit
> 
> Its possible that this is now counted as new, can you just fix the warnings there?

``` java
protected IWebBrowser doCreateBrowser(int style, String browserId, String name, String tooltip)
	{
		return new DefaultWebBrowser(this, browserId);
	}
```
this is the code. The warning complaints about unused params. I refrained from removing the params because it's a protected method (overridden methods may choose to use these params). what do you suggest?

merks:
Annotations can be used:

image

mickaelistria (approved):
Code is better, no new warning introduced, let's merge.
Other issues can be considered seperately (more easier tracking and better productivity)

laeubi:
`DefaultWorkbenchBrowserSupport` is an internal class so we can change it anytime and anyway...

PR #1469, opened 2024-01-05
convert legacy iterator to forEach lambda

elsazac:


bundles/org.eclipse.core.databinding.property/src/org/eclipse/core/internal/databinding/property/value/MapDelegatingValueObservableMap.java
```diff
@@ -162,36 +162,34 @@ private MapDiff<K, V> convertDiff(MapDiff<? extends K, ? extends I> diff) {
 			Map<K, V> newValues = new HashMap<>();
 
 			Set<? extends K> addedKeys = diff.getAddedKeys();
-			for (K key : addedKeys) {
+			addedKeys.forEach(key -> {
```
laeubi, line 166: Can you explain why this style is better the a classic foreach?

Also the commit message do not reflect this well as this is not an iterator (and I can't see iterators are legacy).

bundles/org.eclipse.core.databinding.property/src/org/eclipse/core/internal/databinding/property/value/MapDelegatingValueObservableMap.java
```diff
@@ -162,36 +162,34 @@ private MapDiff<K, V> convertDiff(MapDiff<? extends K, ? extends I> diff) {
...
+			changedKeys.forEach(key -> {
 				I oldMasterValue = diff.getOldValue(key);
 				I newMasterValue = diff.getNewValue(key);
 
 				V oldValue = cache.get(oldMasterValue);
 				V newValue = cache.get(newMasterValue);
 
 				if (Objects.equals(oldValue, newValue)) {
-					it.remove();
+					changedKeys.remove(key);
```
laeubi: I think iterator is required here because otherwise removing items from set might throw concurrent modification exception or is it guaranteed?

If one wants to "modernize" things, `removeIf(...)` might be the construct to choose here

bundles/org.eclipse.core.databinding.property/src/org/eclipse/core/internal/databinding/property/value/MapDelegatingValueObservableMap.java
```diff
@@ -270,12 +268,11 @@ public Object getValueType() {
 
 	private Set<K> keysFor(I masterValue) {
 		Set<K> keys = new HashSet<>();
-
-		for (Map.Entry<K, I> entry : masterMap.entrySet()) {
+		masterMap.entrySet().forEach(entry -> {
```
laeubi: same here what is the benefit of this construct? looking at the code I would say a more modern approach would be  using stream with filter and collect them into a set.

laeubi (changes requested):
Please try to explain why you think this improves anything also I'm not sure it will work to remove

bundles/org.eclipse.core.databinding.property/src/org/eclipse/core/internal/databinding/property/value/MapDelegatingValueObservableMap.java
elsazac: Thanks for the suggestion, made the changes by making use of `removeIf` API.

bundles/org.eclipse.core.databinding.property/src/org/eclipse/core/internal/databinding/property/value/MapDelegatingValueObservableMap.java
elsazac: made those changes as per your suggestion. please have a look.

bundles/org.eclipse.core.databinding.property/src/org/eclipse/core/internal/databinding/property/value/MapDelegatingValueObservableMap.java
elsazac: my understanding is `lambda` is preferred over traditional iterations(with trivial `for` loops for explicit `collection` iterators). Let me know if that's not the case., and I can remove this change.

iloveeclipse (changes requested):
Given that the change is not actually an improvement, I would prefer not to merge this PR.
There is no reason (neither style nor performance nor anyone) why it should be better as original code.

merks:
I'm certainly hoping that we don't see all for loops turned into lambdas just because it's possible and it's modern.  (At best it performs no worse than before, though, granted performance won't matter is most cases.)

elsazac:
I did some online search, and did not find any conclusive evidence that suggest lambda is an improvement over iterators.
So I acknowledge and respect your views.Thanks.

mickaelistria:
@elsazac Thanks for the contribution anyway. 1 aspect I would like to add "against" using lambdas for those simpler cases is that lambdas tend to make debugging slightly harder. So while they're usually great at making code more expressive and more focused with less boilerplate, in such cases where they're not an obvious win on performance, maintenance or readability, I think they shouldn't be forced in.

PR #1470, opened 2024-01-05
remove stale bundle dependencies

elsazac:
identified by `organise manifest` tool from PDE.

### DIFF
--- a/bundles/org.eclipse.ui/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui/META-INF/MANIFEST.MF
@@ -9,6 +9,7 @@ Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.ui.internal;x-internal:=true
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
+ org.eclipse.swt;bundle-version="[3.124.0,4.0.0)";visibility:=reexport,
  org.eclipse.jface;bundle-version="[3.31.0,4.0.0)";visibility:=reexport,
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui/META-INF/MANIFEST.MF
```diff
@@ -9,9 +9,7 @@ Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.ui.internal;x-internal:=true
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
- org.eclipse.swt;bundle-version="[3.124.0,4.0.0)";visibility:=reexport,
```
laeubi, line 12: Reexports can not be removed even if they are unused

laeubi (changes requested):

bundles/org.eclipse.ui/META-INF/MANIFEST.MF
elsazac: Thanks for the review. reverted the `SWT` entry that has the reexport visibility. please have a look.
btw I was using the `PDE` tool to clean up the dependencies. Do you think the tool needs to care about the visibility attribute?

iloveeclipse (changes requested):
`org.eclipse.core.expressions` extension point is used in this bundle, so it can't be removed from dependencies

iloveeclipse:
See my comment above, closing

bundles/org.eclipse.ui/META-INF/MANIFEST.MF
laeubi: It should have an option to keep such entries, in general one *can* remove, but for eclipse platform it would be an API break, so if you can enhance/improve it it would be great.

PR #1472, opened 2024-01-08
Add filter text to the Feature about dialog

elsazac:
Looks like the filter feature was omitted for the feature list because its a relatively smaller set . However we are seeing users with multipage feature list, and it makes sense to add a filter feature.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
@@ -49,6 +49,7 @@ import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Table;
 import org.eclipse.swt.widgets.TableColumn;
 import org.eclipse.swt.widgets.TableItem;
 import org.eclipse.swt.widgets.Text;
@@ -70,25 +71,21 @@ public class AboutFeaturesPage extends ProductInfoPage {
 
 		@Override
 		public String getColumnText(Object element, int columnIndex) {
-			if (element instanceof AboutBundleGroupData) {
-				AboutBundleGroupData data = (AboutBundleGroupData) element;
-				switch (columnIndex) {
-				case 0:
-					return data.getProviderName();
-				case 1:
-					return data.getName();
-				case 2:
-					return data.getVersion();
-				case 3:
-					return data.getId();
-				}
+			if (element instanceof AboutBundleGroupData data) {
+
+				return switch (columnIndex) {
+				case 0 -> data.getProviderName();
+				case 1 -> data.getName();
+				case 2 -> data.getVersion();
+				case 3 -> data.getId();
+				default -> ""; //$NON-NLS-1$
+				};
 			}
 			return ""; //$NON-NLS-1$
 		}
 
 		@Override
 		public Image getColumnImage(Object element, int columnIndex) {
-			// TODO Auto-generated method stub
 			return null;
 		}
 	}
@@ -108,7 +105,7 @@ public class AboutFeaturesPage extends ProductInfoPage {
 
 	private static final int COLUMNS_ID = IDialogConstants.CLIENT_ID + 3;
 
-	private TableViewer table;
+	private Table table;
 
 	private Label imageLabel;
 
@@ -175,7 +172,7 @@ public class AboutFeaturesPage extends ProductInfoPage {
 	 * selected feature.
 	 */
 	private void handlePluginInfoPressed() {
-		TableItem[] items = table.getTable().getSelection();
+		TableItem[] items = table.getSelection();
 		if (items.length <= 0) {
 			return;
 		}
@@ -196,7 +193,7 @@ public class AboutFeaturesPage extends ProductInfoPage {
 		moreButton = createButton(parent, MORE_ID, WorkbenchMessages.AboutFeaturesDialog_moreInfo);
 		pluginsButton = createButton(parent, PLUGINS_ID, WorkbenchMessages.AboutFeaturesDialog_pluginsInfo);
 		createButton(parent, COLUMNS_ID, WorkbenchMessages.AboutFeaturesDialog_columns);
-		TableItem[] items = table.getTable().getSelection();
+		TableItem[] items = table.getSelection();
 		AboutBundleGroupData info = null;
 		if (items.length > 0) {
 			info = (AboutBundleGroupData) items[0].getData();
@@ -258,7 +255,7 @@ public class AboutFeaturesPage extends ProductInfoPage {
 
 		textManager = new AboutTextManager(text);
 
-		TableItem[] items = table.getTable().getSelection();
+		TableItem[] items = table.getSelection();
 		if (items.length > 0) {
 			updateInfoArea((AboutBundleGroupData) items[0].getData());
 		}
@@ -278,16 +275,18 @@ public class AboutFeaturesPage extends ProductInfoPage {
 
 		initializeBundleGroupInfos();
 
-		table = new TableViewer(parent, SWT.H_SCROLL | SWT.V_SCROLL | SWT.SINGLE | SWT.FULL_SELECTION | SWT.BORDER);
+		TableViewer tableViewer = new TableViewer(parent,
+				SWT.H_SCROLL | SWT.V_SCROLL | SWT.SINGLE | SWT.FULL_SELECTION | SWT.BORDER);
 
+		table = tableViewer.getTable();
 		GridData gridData = new GridData(GridData.FILL, GridData.FILL, true, true);
 		gridData.heightHint = convertVerticalDLUsToPixels(TABLE_HEIGHT);
-		table.getTable().setLayoutData(gridData);
-		table.getTable().setHeaderVisible(true);
+		table.setLayoutData(gridData);
+		table.setHeaderVisible(true);
 
-		table.getTable().setLinesVisible(true);
-		table.getTable().setFont(parent.getFont());
-		table.getTable().addSelectionListener(widgetSelectedAdapter(e -> {
+		table.setLinesVisible(true);
+		table.setFont(parent.getFont());
+		table.addSelectionListener(widgetSelectedAdapter(e -> {
 			// If there is no item, nothing we can do.
 			// See https://bugs.eclipse.org/bugs/show_bug.cgi?id=266177
 			if (e.item == null)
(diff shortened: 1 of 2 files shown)


laeubi:
@elsazac can you share a screenshot before -> after ?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
```diff
@@ -239,18 +271,23 @@ protected void createInfoArea(Composite parent) {
...
 
+		final Text filterText = new Text(parent, SWT.BORDER | SWT.SEARCH | SWT.ICON_CANCEL | SWT.ICON_SEARCH);
+		filterText.setLayoutData(GridDataFactory.fillDefaults().create());
+		filterText.setMessage(WorkbenchMessages.AboutPluginsDialog_filterTextMessage);
+		filterText.setFocus();
+
 		initializeBundleGroupInfos();
 
-		table = new Table(parent, SWT.H_SCROLL | SWT.V_SCROLL | SWT.SINGLE | SWT.FULL_SELECTION | SWT.BORDER);
+		table = new TableViewer(parent, SWT.H_SCROLL | SWT.V_SCROLL | SWT.SINGLE | SWT.FULL_SELECTION | SWT.BORDER);
```
laeubi: you should rename this to 
```suggestion
		tableViewer = new TableViewer(parent, SWT.H_SCROLL | SWT.V_SCROLL | SWT.SINGLE | SWT.FULL_SELECTION | SWT.BORDER);
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
```diff
@@ -239,18 +271,23 @@ protected void createInfoArea(Composite parent) {
...
 
 		GridData gridData = new GridData(GridData.FILL, GridData.FILL, true, true);
 		gridData.heightHint = convertVerticalDLUsToPixels(TABLE_HEIGHT);
-		table.setLayoutData(gridData);
-		table.setHeaderVisible(true);
+		table.getTable().setLayoutData(gridData);
```
laeubi: This could then be extracted into a local variable "table"

elsazac:
> @elsazac can you share a screenshot before -> after ?

before:

image

after:

image

after filter is applied:

image

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
elsazac: @laeubi Thanks for the quick review, noted the suggestions.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
```diff
@@ -61,6 +66,33 @@
...
+					return data.getProviderName();
+				case 1:
+					return data.getName();
+				case 2:
+					return data.getVersion();
+				case 3:
+					return data.getId();
+				}
+			}
+			return ""; //$NON-NLS-1$
```
mickaelistria: Just to promote newer Java
```suggestion
			return element instanceof AboutBundleGroupData data ?
				switch (columnIndex) {
					case 0 -> data.getProviderName();
					case 1 -> data.getName();
					case 2 -> data.getVersion();
					case 3 -> data.getId();
					default -> "";  //$NON-NLS-1$
				} : "";  //$NON-NLS-1$
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
```diff
@@ -61,6 +66,33 @@
...
+		}
+
+		@Override
+		public Image getColumnImage(Object element, int columnIndex) {
+			// TODO Auto-generated method stub
```
mickaelistria: Please remove comment if there is no suitable image to show.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
```diff
@@ -76,7 +108,7 @@ public class AboutFeaturesPage extends ProductInfoPage {
 
 	private static final int COLUMNS_ID = IDialogConstants.CLIENT_ID + 3;
 
-	private Table table;
+	private TableViewer table;
```
mickaelistria: This could still be the `Table` as it's the element that's used in all methods.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
mickaelistria: And the TableViewer could most likely be a local variable (and Table remain the field)

mickaelistria (changes requested):
Some minor code commends inline.
On a more general note, wouldn't it be simpler to use a FilteredTree instead? (it's s sincere question, not a request)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
elsazac: addressed the changes.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
elsazac: ditto

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutPluginsPage.java
```diff
@@ -693,6 +693,11 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
 			return matcher.match(data.getName()) || matcher.match(data.getProviderName())
 					|| matcher.match(data.getId());
 		}
+		else if (element instanceof AboutBundleGroupData) {
```
mickaelistria, line 696:
```suggestion
		else if (element instanceof AboutBundleGroupData data) {
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutPluginsPage.java
```diff
@@ -693,6 +693,11 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
...
+			AboutBundleGroupData data = (AboutBundleGroupData) element;
```
mickaelistria:
```suggestion
		else if (element instanceof AboutBundleGroupData data) {
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
elsazac: This syntax is giving me error "data cannot be resolved to a type". so I used the long notation, still making use of the modern switch case semantics. let me know if this is fine with you.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
elsazac: done.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
elsazac: done.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
mickaelistria: OK thanks.
I'm wondering whether the error is normal. Ie does `javac` manage to process it?

elsazac:
> Some minor code commends inline. On a more general note, wouldn't it be simpler to use a FilteredTree instead? (it's s sincere question, not a request)

I haven't studied about FilteredTree enough to make this change now. as the code is working as expected , I would like to leave the design as is for the time being and look for improvements in the future. Hope it's fine with you?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutPluginsPage.java
elsazac: done.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/about/AboutFeaturesPage.java
elsazac: This is what I get, looks like Javac is unable to recognise it.
image

iloveeclipse: Which Eclipse version do you have? Please try to use latest SDK build, there were few fixes in switch area.

elsazac: I am using this version of Eclipse SDK
```
Version: 2024-03 (4.31)
Build id: I20240101-1800
```

did the fix come in the last few days?

mickaelistria: I'm seeing the same error with Version: 2024-03 (4.31) / Build id: I20240105-1800 , updated about 2 hours ago.
```java
class A {
	String method(Object element, int columnIndex) {
		return element instanceof String data ?
			switch (columnIndex) {
				case 0 -> data;
				case 1 -> data.toUpperCase();
				default -> "";
			} : "";
	}
}
```
makes JDT complain with

> Description	Resource	Path	Location	Type
data cannot be resolved	A.java	/sandboxJava/src/sandboxJava	line 7	Java Problem
data cannot be resolved to a variable	A.java	/sandboxJava/src/sandboxJava	line 6	Java Problem

while `javac A.java` generates a fine `A.class` file.

@elsazac as you were the lucky one spotting the issue, would you mind reporting it to JDT Core?

mickaelistria (approved):

Issue #370, opened 2022-09-30
[Tests] Random failing ProgressContantsTest.testKeepOneProperty


ktatavarthi:
org.eclipse.ui.tests.progress.ProgressContantsTest.testKeepOneProperty fails in the I-build [I20220928-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220928-1800) on [Windows ](https://download.eclipse.org/eclipse/downloads/drops4/I20220928-1800/testresults/html/org.eclipse.ui.tests_ep426I-unit-win32-java11_win32.win32.x86_64_11.html) and [MacM1](https://download.eclipse.org/eclipse/downloads/drops4/I20220928-1800/testresults/html/org.eclipse.ui.tests_ep426I-unit-macM1-java17_macosx.cocoa.aarch64_17.html) with the below error

```
Only one finished job should be kept in view expected:<1> but was:<20>

junit.framework.AssertionFailedError: Only one finished job should be kept in view expected:<1> but was:<20>
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.Assert.failNotEquals(Assert.java:329)
at junit.framework.Assert.assertEquals(Assert.java:78)
at junit.framework.Assert.assertEquals(Assert.java:234)
at junit.framework.TestCase.assertEquals(TestCase.java:377)
at org.eclipse.ui.tests.progress.ProgressContantsTest.testKeepOneProperty(ProgressContantsTest.java:259) 
```

jukzi:
i can reproduce this error locally

jukzi:
i can still reproduce this error when reverting eclipse.platform to 149a08f433580abfecff3a9cbf73250638133aa8 (2022-09-05) - before the JobManager.join changes!

jukzi:
i can still reproduce this error when reverting eclipse.platform.ui to 180c5b6b70f00a875dfb2da6271e4bf7948d42ea (2022-09-21) - before my ui changes.

iloveeclipse:
Unfortunately I can't reproduce on Linux, may be that is Windows specific.

> i can still reproduce this error when reverting eclipse.platform.ui to https://github.com/eclipse-platform/eclipse.platform.ui/commit/180c5b6b70f00a875dfb2da6271e4bf7948d42ea (2022-09-21) - before my ui changes.

OK, so that could be unrelated. The problem is that there is no one looking on Windows test fails, because there are too many of them, so it could be even already in 4.25.

[...]

PR #1474, opened 2024-01-08
ltk: use SubMonitor

jukzi:
to get rid of deprecated SubProgressMonitor


jukzi:
randoms fails on MAC:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/811  
https://github.com/eclipse-platform/eclipse.platform.ui/issues/370

PR #1477, opened 2024-01-09
show fully resolved ws path in launcher dialog

### LINKED ISSUE #279: Workspace selection dialog allows relative paths and that leads to unexpected workspace path
At least on macOS and Eclipse 2022-06, the dialog where the path to the Eclipse workspace is specified does not resolve the tilde character (`~`), referring to the user's home directory (It does work when you click on browse and then start typing, since this is probably a system UI)

Instead of resolving `~` to the user's home directory, the path is treated as being relative to the Eclipse application folder, in my case `/Applications/Eclipse.app/Contents/MacOS/~/test`.

You can imagine what happened next: After I found the actual path using the "browse" button (because the drop down menus also only show `~/test`), I used Terminal to cd into /Applications/Eclipse.app/Contents/MacOS and tried to `rm -rf ~`...

I guess I should have finished my morning coffee first. Thankfully, I have recent backups of my actual home directory.

tilde-of-death

elsazac:
Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/279

with normal relative path referencing (good case):
image

with `~` in the path referencing (bad case) :
image

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
@@ -49,6 +49,7 @@ import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.layout.BorderData;
 import org.eclipse.swt.layout.BorderLayout;
+import org.eclipse.swt.layout.FillLayout;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.layout.RowData;
@@ -58,6 +59,7 @@ import org.eclipse.swt.widgets.Combo;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.DirectoryDialog;
+import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.Link;
 import org.eclipse.swt.widgets.Menu;
 import org.eclipse.swt.widgets.MenuItem;
@@ -94,6 +96,7 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 	private Form recentWorkspacesForm;
 
 	private Button defaultButton;
+
 
 	/**
 	 * Create a modal dialog on the argument shell, using and updating the
@@ -447,11 +450,14 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 	}
 
 	protected Combo createPathCombo(Composite panel) {
-		text = new Combo(panel, SWT.BORDER | SWT.LEAD | SWT.DROP_DOWN);
+		Composite c = new Composite(panel, SWT.NONE);
+		FillLayout layout = new FillLayout(SWT.VERTICAL);
+		c.setLayout(layout);
+		text = new Combo(c, SWT.BORDER | SWT.LEAD | SWT.DROP_DOWN);
 		new DirectoryProposalContentAssist().apply(text);
 		text.setTextDirection(SWT.AUTO_TEXT_DIRECTION);
 		text.setFocus();
-		text.setLayoutData(new GridData(400, SWT.DEFAULT));
+		Label label = new Label(c, SWT.BORDER | SWT.LEAD);
 		text.addModifyListener(e -> {
 			Button okButton = getButton(Window.OK);
 			if(okButton != null && !okButton.isDisposed()) {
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -464,6 +471,14 @@ protected Combo createPathCombo(Composite panel) {
 					}
 				}
 				okButton.setEnabled(nonWhitespaceFound);
+				try {
+				File location = new File(characters);
+				label.setText(
+						IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + location.getCanonicalPath());
```
mickaelistria: We shouldn't show the canonical path here as we most likely don't want to resolve symlinks. What you want is more likely a normalized path. Something like `location.getAbsoluteFile().toPath().normalize()`

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -464,6 +471,14 @@ protected Combo createPathCombo(Composite panel) {
 					}
 				}
 				okButton.setEnabled(nonWhitespaceFound);
+				try {
+				File location = new File(characters);
+				label.setText(
```
mickaelistria, line 479: The label should only be visible (ie not empty) if the input is different from the resolved normalized location. When the path is absolute and doesn't contain `.` or `..` or whatever (basically when the resolved location is equal to the input) it would be better to not display it/make the label empty.

mickaelistria (changes requested):

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
elsazac: Thanks for the suggestion. I tried this and it works.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
elsazac: given the main use case is to show the user  what the path would resolve to , how this will work for `~` character?
for example, if the path contains `~`, then the absolute normalised path will make no change and hence both will be the same. In this scenario the label should show up to warn the user, while it will not show up according to your suggestion?

mickaelistria: OK, I understand the purpose.
If `~` is an exceptional case, we can maybe just show a warning, setting the label to something like `⚠️ '~' is not expanded, resolved path is ...`?

elsazac: @mickaelistria I followed your suggestion, however I have difficulty in rendering the image. with the pushed changes, I am getting NPE. I know the reason as the variable is uninitialised but the same logic is working elsewhere in the project. Can you advise?

`java.lang.NullPointerException: Cannot invoke "org.eclipse.e4.ui.workbench.IResourceUtilities.imageDescriptorFromURI(org.eclipse.emf.common.util.URI)" because "this.resUtils" is null
`

mickaelistria: Don't bother with the image, just use the warning emoji in messages.properties/ChooseWorkspaceDialog_TildeNonExpandedWarning0 (IIRC you should use the escaped encoding of the emoji in properties file, eg `\u26A0\uFE0F`

elsazac: @mickaelistria I have addressed the requested changes. Please have a look.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -464,6 +470,30 @@ protected Combo createPathCombo(Composite panel) {
...
+				if (!characters.equalsIgnoreCase(normalisedPath)
+						&& !characters.equalsIgnoreCase(normalisedPathWithSeperator)) {
+				label.setText(
+						IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0
+								+ location.getAbsoluteFile().toPath().normalize());
+				}
+				else {
+					label.setText(""); //$NON-NLS-1$
+				}
+				if (normalisedPath.indexOf("~") != -1) //$NON-NLS-1$
```
iloveeclipse: Is this really `indexOf("~")` and not `startsWith("~")` ?
Matching `../../~/../` makes no sense for me

elsazac: @iloveeclipse I agree with your assertion and I have made the changes.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -464,6 +470,30 @@ protected Combo createPathCombo(Composite panel) {
...
+				}
+				if (normalisedPath.startsWith("~")) //$NON-NLS-1$
+				{
+					label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_TildeNonExpandedWarning0
+							+ IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0
+							+ location.getAbsoluteFile().toPath().normalize());
+
+				}
+
+
```
mickaelistria: Can we get rid of some whitespaces here?

elsazac: Yes. I have removed the whitespaces.

mickaelistria (approved):

mickaelistria:
Thanks!

PR #1482, opened 2024-01-09
Fix faulty <li> tags in javadoc

HeikoKlare:
Fixes: `tag not allowed here: <li>`


HeikoKlare:
Search in log before (e.g. recent master build https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/199):
image

Search in log after:
image
Match being only the logged commit message.

PR #1484, opened 2024-01-09
Fix faulty braces around @return

HeikoKlare:
Braces around @return are only valid if no description is provided, as the return description is then used for the general description.

This adresses the two javadoc warnings reported in Jenkins:
image


HeikoKlare:
image

PR #1485, opened 2024-01-09
Fix faulty javadoc references

HeikoKlare:
Fixes all logged javadoc errors of type "reference not found":
* Remove faulty or outdated javadoc containing faulty reference
* Remove obsolete `@see SUPERCLASS_DOCUMENTATION` javadoc for overwritten methods
* Correct faulty references to types and types if they are accessible
* Replace type links with code tags in case a references type is accessible (i.e., not on the bundle's classpath)


HeikoKlare:
Search in log before (e.g. recent master build https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/207):
image

Search in log after:
image

PR #1486, opened 2024-01-09
Disable local parent resolution as it is not longer part of the same

kwin:
repository


laeubi (changes requested):
This is used in the [aggregator builds](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/) in local builds this is resolved from the snapshots repository.

jukzi:
Closing because of to many faults: negative review, conflict, no progress

PR #1488, opened 2024-01-10
Include validation result into diagnostic message of fragment merge

krutelp:
Include validation result into diagnostic message of fragment merge. Without validation result the error is not visible to the user and requires debugging.


vogella:
LGTM

PR #1491, opened 2024-01-11
remove label border and fix button layout in the launch dialog

### LINKED ISSUE #1492: SWT layout in Choose Workspace Dialog broken in I20240109-1800

## Steps to reproduce

Download and launch the latest SDK nightly build.

Before (4.30):

image

After (4.31I):

image

I believe this is a result of https://github.com/eclipse-platform/eclipse.platform.ui/pull/1477, which shows the absolute path if only a relative path is given. The problem arises when the path is already fully qualified, which makes the empty field look out-of-place.

* Eclipse SDK
* Version: 2024-03 (4.31)
* Build id: I20240109-1800
* OS: Windows 10, v.10.0, x86_64 / win32
* Java vendor: Eclipse Adoptium
* Java runtime version: 21.0.1+12-LTS
* Java version: 21.0.1

## Community

elsazac:
fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1492
Refs: #1477

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
@@ -453,6 +453,32 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 		return panel;
 	}
 
+	private void populateResolvedPathLabel(Label label, String characters) {
+		try {
+			File location = new File(characters);
+			String normalisedPath = ""; //$NON-NLS-1$
+			if (location.exists() && location.isDirectory()) {
+				normalisedPath = location.getAbsoluteFile().toPath().toRealPath(LinkOption.NOFOLLOW_LINKS).normalize()
+						.toString();
+			} else {
+				normalisedPath = location.getAbsoluteFile().toPath().normalize().toString();
+			}
+			String normalisedPathWithSeperator = normalisedPath + File.separator;
+			if (characters.startsWith("~")) //$NON-NLS-1$
+			{
+				label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_TildeNonExpandedWarning0
+						+ IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + System.lineSeparator()
+						+ location.getAbsoluteFile().toPath().normalize());
+			} else if (!characters.equalsIgnoreCase(normalisedPath)
+					&& !characters.equalsIgnoreCase(normalisedPathWithSeperator)) {
+				label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + System.lineSeparator()
+						+ location.getAbsoluteFile().toPath().normalize());
+			} else {
+				label.setText(""); //$NON-NLS-1$
+			}
+		} catch (Exception ex) {
+		}
+	}
 	protected Combo createPathCombo(Composite panel) {
 
 		Composite c = createBrowseComposite(panel.getParent());
@@ -474,35 +500,11 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 					}
 				}
 				okButton.setEnabled(nonWhitespaceFound);
-				try {
-					File location = new File(characters);
-					String normalisedPath = ""; //$NON-NLS-1$
-					if (location.exists() && location.isDirectory()) {
-						normalisedPath = location.getAbsoluteFile().toPath().toRealPath(LinkOption.NOFOLLOW_LINKS)
-								.normalize().toString();
-					} else {
-						normalisedPath = location.getAbsoluteFile().toPath().normalize().toString();
-					}
-					String normalisedPathWithSeperator = normalisedPath + File.separator;
-					if (characters.startsWith("~")) //$NON-NLS-1$
-					{
-						label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_TildeNonExpandedWarning0
-								+ IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0
-								+ System.lineSeparator() + location.getAbsoluteFile().toPath().normalize());
-					} else if (!characters.equalsIgnoreCase(normalisedPath)
-							&& !characters.equalsIgnoreCase(normalisedPathWithSeperator)) {
-						label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0
-								+ System.lineSeparator() + location.getAbsoluteFile().toPath().normalize());
-					} else {
-						label.setText(""); //$NON-NLS-1$
-					}
-
-				}
-				catch (Exception ex) {
-				}
+				populateResolvedPathLabel(label, characters);
 			}
 		});
 		setInitialTextValues(text);
+		populateResolvedPathLabel(label, text.getText());
 		return text;
 	}
 
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
@@ -17,6 +17,7 @@ package org.eclipse.ui.internal.ide;
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.LinkOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -427,7 +428,10 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 	 * information and a drop-down of the most recently used workspaces.
 	 */
 	protected Control createWorkspaceBrowseRow(Composite parent) {
-		Composite panel = createBrowseComposite(parent);
+		Composite parentpanel = createBrowseComposite(parent);
+		FillLayout layout = new FillLayout(SWT.VERTICAL);
+		parentpanel.setLayout(layout);
+		Composite panel = createBrowseComposite(parentpanel);
 
 		createPathCombo(panel);
 
@@ -440,9 +444,9 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 		Composite panel = new Composite(parent, SWT.NONE);
 
 		BorderLayout layout = new BorderLayout();
-		layout.marginHeight = convertVerticalDLUsToPixels(IDialogConstants.VERTICAL_MARGIN);
-		layout.marginWidth = convertHorizontalDLUsToPixels(IDialogConstants.HORIZONTAL_MARGIN);
-		layout.spacing = convertVerticalDLUsToPixels(IDialogConstants.VERTICAL_SPACING);
+		layout.marginHeight = convertVerticalDLUsToPixels(IDialogConstants.VERTICAL_MARGIN) / 2;
+		layout.marginWidth = convertHorizontalDLUsToPixels(IDialogConstants.HORIZONTAL_MARGIN) ;
+		layout.spacing = convertVerticalDLUsToPixels(IDialogConstants.VERTICAL_SPACING) / 2;
 		panel.setLayout(layout);
 		panel.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
 		panel.setFont(parent.getFont());
@@ -450,14 +454,14 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 	}
 
 	protected Combo createPathCombo(Composite panel) {
-		Composite c = new Composite(panel, SWT.NONE);
-		FillLayout layout = new FillLayout(SWT.VERTICAL);
-		c.setLayout(layout);
-		text = new Combo(c, SWT.BORDER | SWT.LEAD | SWT.DROP_DOWN);
+
+		Composite c = createBrowseComposite(panel.getParent());
+
+		text = new Combo(panel, SWT.BORDER | SWT.LEAD | SWT.DROP_DOWN);
 		new DirectoryProposalContentAssist().apply(text);
 		text.setTextDirection(SWT.AUTO_TEXT_DIRECTION);
 		text.setFocus();
-		Label label = new Label(c, SWT.BORDER | SWT.LEAD);
+		Label label = new Label(c, SWT.NONE);
 		text.addModifyListener(e -> {
 			Button okButton = getButton(Window.OK);
 			if(okButton != null && !okButton.isDisposed()) {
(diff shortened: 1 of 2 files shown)


elsazac:
This is how the launch dialog looks now. The warning message is fully visible now and the browse button is back to the normal alignment. But as a whole the label is little bit down from the previous look, I think. I tried moving it up but could not and I have run out of ideas. If anyone has a suggestion on how to fix this, please let me know.

image

elsazac:
@jukzi 
Can you test this in Windows and see if https://github.com/eclipse-platform/eclipse.platform.ui/issues/279#issuecomment-1884464012 is resolved?

elsazac:
@jukzi 
Regarding this https://github.com/eclipse-platform/eclipse.platform.ui/issues/279#issuecomment-1884475265, I am unable to reproduce this in Mac and I don't have Windows. Is it fair to suspect the continuous key press call back  https://github.com/eclipse-platform/eclipse.platform.ui/blob/307a46097bbccb0af721f9c28e17f7c48923ed5c/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java#L461-L493 and subsequent path computation logic as consuming CPU ? I have made the conditions in the call back more strict which reduces the path computation logic. Let me know if the crash is still happening.

elsazac:
@jukzi 
Regarding https://github.com/eclipse-platform/eclipse.platform.ui/issues/279#issuecomment-1884500242, as others already mentioned, I am not devising this path instead just displaying what is being computed inside. It is not in the scope of this PR to modify/optimise that.

jukzi:
the border seems to be gone. but text is not aligned and the window seems to be inappropriatly high compared to old version:
image
also you should catch such errors:
```
java.nio.file.InvalidPathException: UNC path is missing sharename: \\a
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:118)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:232)
	at java.base/java.io.File.toPath(File.java:2387)
	at org.eclipse.ui.internal.ide.ChooseWorkspaceDialog.lambda$10(ChooseWorkspaceDialog.java:474)
```

would also look better if the path's are aligned such that the text "Resolved absolute path:" is not in front but below

jukzi:
> I am not devising this path instead just displaying what is being computed inside. It is not in the scope of this PR to modify/optimise that.

that's ok - just be sure it is really the path that will later be used an not computed any other way.

elsazac:
@jukzi 

Regarding https://github.com/eclipse-platform/eclipse.platform.ui/issues/279#issuecomment-1884506054, Not sure I followed the suggestion fully, there was an advice to switch from canonicalised path to normalised path .Without knowing clear benefit of one over the other in this context I am hesitant to make this change.

jukzi:
> Without knowing clear benefit of one over the other in this context I am hesitant to make this change.

1. to be consistent with at least ChooseWorkspaceWithSettingsDialog.toggleDecoForSettingsImportButtons():231 you need to use NOFOLLOW_LINKS, i.e. Path.toRealPath(NOFOLLOW_LINKS)
2. also note that ChooseWorkspaceWithSettingsDialog.getInitialBrowsePath:547 looks like relative paths are intended to be resolved against user.dir
3. also ChooseWorkspaceWithSettingsDialog.workspaceSelected:260 will later remove directional marker characters

i think the most consistent way would be if all those spread logic would be gathered in ChooseWorkspaceDialog.getWorkspaceLocation()

elsazac:
>also you should catch such errors:

done. enclosed the whole file access code in try catch block.

>would also look better if the path's are aligned such that the text "Resolved absolute path:" is not in front but below

done.
image

elsazac:
> 1. to be consistent with at least ChooseWorkspaceWithSettingsDialog.toggleDecoForSettingsImportButtons():231 you need to use NOFOLLOW_LINKS, i.e. Path.toRealPath(NOFOLLOW_LINKS)

This throws ` java.nio.file.NoSuchFileException: /Applications/Eclipse 4.app/Contents/MacOS/~` for non existent paths.
> 2. also note that ChooseWorkspaceWithSettingsDialog.getInitialBrowsePath:547 looks like relative paths are intended to be resolved against user.dir

Sorry as I said earlier, I am not planning to cover it here.

laeubi:
@ptziegler can you help with testing this?

laeubi:
> This throws ` java.nio.file.NoSuchFileException: /Applications/Eclipse 4.app/Contents/MacOS/~` for non existent paths.

I think one should check if path is a directory if yes, then use `toRealPath` otherwise use `normalize` (that works on non existing path) maybe.

iloveeclipse:
Last state still not nice, see picture
image

merks:
Yes, there remains significant room for improvement.  Even the text `expanded, Resolved` with a capitalized word after the comma looks poor.

ptziegler:
@laeubi Things are a little busy right now, but I think I can spare some time for tests.

ptziegler:
That said, the weird button layout is back, once I expand the "Copy Settings" tree
(Note: Done via the "Switch Workspace" button)

image

iloveeclipse:
What about following proposal: I will push a full revert of original commit, and one can continue here with improving things as long as we need to get something acceptable?

I honestly can't see the new dialog appearance anymore.

elsazac:
> What about following proposal: I will push a full revert of original commit, and one can continue here with improving things as long as we need to get something acceptable?
> 
> I honestly can't see the new dialog appearance anymore.

I am working on it. There are multiple problems to fix and it is taking time but I am making progress.

elsazac:
outstanding issues:
 - unaligned text and label: looks like the image rendition causes the unalignment; still debugging
 - unwanted space in the top margin: identified the issue as a side effect of doubling of vertical margin and spacing in a border layout. I am able to fix this by cutting the margins space to half of the previous value.
 - expansion of browse button when expandable composite is touched; investigating
 
I can assert that these are doable in reasonable amount of time and hence request not to revert the original PR.

elsazac:
Looks like I have fixed all the issues found in the previous PR. @iloveeclipse @jukzi @ptziegler @merks Please have a look and let me know if you are happy with the changes. below pasting screenshots for your convenience.

Launch window:
Screenshot 2024-01-12 at 12 28 39 PM

with ~ character: 
Screenshot 2024-01-12 at 12 28 54 PM

with relative path:
image

with Expandable composite expanded: 
image

jukzi:
before i test myself, please remove the extra space below the combo in your screenshot:
image

ptziegler:
That's looking much smoother now.
But as a minor concern: With the default size, the label for the resolved path is cut off, so I assume one has to also request a re-layout.

The gap between the label and the combo box also looks a little weird. How difficult would it be to get rid of it?

image

mickaelistria:
I see in the screenshot that the "Browse" button doesn't seem fully aligned with the text field (maybe a couple of pixels more at the bottom). Can this also be considered as part of this change? I imagine it's something like the gridLayout margins being different.

merks:
> I see in the screenshot that the "Browse" button doesn't seem fully aligned with the text field (maybe a couple of pixels more at the bottom). Can this also be considered as part of this change? I imagine it's something like the gridLayout margins being different.

I think you'll find such pixel polishing would be more widely needed, e.g, 

image

image

I think a button has a certain height and a text field has a certain height and they would need to have the same height.  Let's not go further off the deep end on this change set.

BeckerWdf:
> What about following proposal: I will push a full revert of original commit, and one can continue here with improving things as long as we need to get something acceptable?

Sounds like a good approach.

elsazac:
>before i test myself, please remove the extra space below the combo in your screenshot:

image

@jukzi 
I have reduced the spacing and vertical margin associated with the combo and the label by half. Are you happy with this? Its possible to reduce it further to any arbitrary value but reducing to half looks reasonable to me because: 
 - it is aligned with the existing logic of deriving space and margin from the dialog's default 
 - reducing the spacing further will also reduce the spacing between the top combo and its top boundary.
 - the label looks decently positioned between the combo and the expandable composite.

elsazac:
> That's looking much smoother now. But as a minor concern: With the default size, the label for the resolved path is cut off, so I assume one has to also request a re-layout.

@ptziegler Could you test now?

jukzi:
still broken on win:
image

elsazac:
> I see in the screenshot that the "Browse" button doesn't seem fully aligned with the text field (maybe a couple of pixels more at the bottom). Can this also be considered as part of this change? I imagine it's something like the gridLayout margins being different.

@mickaelistria Thanks for the suggestion and even for providing the hint for the fix. However, given that this misalignment is a preexisting condition I would like to limit the changes of this PR to the side effects caused by #1477 only. Please let me know if you are fine with that approach?

elsazac:
> still broken on win: 

Could you please elaborate? Is it that the label is partly hidden or some other problem. Its not clear to me with the yellow highlighter.

jukzi:
Also: when i start the dialog from File/switch workspace, then the hint is initially empty:
image
and as soon as the "copy settings" is expanded the "Browse" button is again to high:
image
i think theese error have been reported now multiple times. please fix them before asking for review again

jukzi:
> Could you please elaborate? Is it that the label is partly hidden or some other problem.

the text is not fully visible.

jukzi:
When the filename is illegal the Hint is not updated anymore but stops at the last legal name
image
it should instead say the path is illegal

elsazac:
> it should instead say the path is illegal

I don't think so. the expectation is to show what path it will resolve to, not to validate the path itself

jukzi:
> I don't think so. the expectation is to show what path it will resolve to

and if does not resolve? Exception -> show its illegal! Currently it shows a path that has nothing todo with the current input :-(

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -440,24 +444,50 @@ protected Composite createBrowseComposite(Composite parent) {
...
+						+ IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + System.lineSeparator()
+						+ location.getAbsoluteFile().toPath().normalize());
+			} else if (!characters.equalsIgnoreCase(normalisedPath)
+					&& !characters.equalsIgnoreCase(normalisedPathWithSeperator)) {
+				label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + System.lineSeparator()
+						+ location.getAbsoluteFile().toPath().normalize());
+			} else {
+				label.setText(""); //$NON-NLS-1$
+			}
+		} catch (Exception ex) {
```
jukzi, line 479: ignoring exception forgets to update label

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -440,24 +444,50 @@ protected Composite createBrowseComposite(Composite parent) {
...
+			String normalisedPathWithSeperator = normalisedPath + File.separator;
+			if (characters.startsWith("~")) //$NON-NLS-1$
+			{
+				label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_TildeNonExpandedWarning0
+						+ IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + System.lineSeparator()
+						+ location.getAbsoluteFile().toPath().normalize());
+			} else if (!characters.equalsIgnoreCase(normalisedPath)
+					&& !characters.equalsIgnoreCase(normalisedPathWithSeperator)) {
+				label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + System.lineSeparator()
+						+ location.getAbsoluteFile().toPath().normalize());
```
jukzi, line 475: should show normalisedPath instead

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -440,24 +444,50 @@ protected Composite createBrowseComposite(Composite parent) {
...
+				normalisedPath = location.getAbsoluteFile().toPath().toRealPath(LinkOption.NOFOLLOW_LINKS).normalize()
+						.toString();
+			} else {
+				normalisedPath = location.getAbsoluteFile().toPath().normalize().toString();
+			}
+			String normalisedPathWithSeperator = normalisedPath + File.separator;
+			if (characters.startsWith("~")) //$NON-NLS-1$
+			{
+				label.setText(IDEWorkbenchMessages.ChooseWorkspaceDialog_TildeNonExpandedWarning0
+						+ IDEWorkbenchMessages.ChooseWorkspaceDialog_ResolvedAbsolutePath0 + System.lineSeparator()
```
jukzi, line 470: if "~" is a legal path in linux, it should still show the (legal) normalisedPath

merks:
This reminds of all the work that was done in installer for the infinite number of ways that people can enter bogus information for the Install Folder location:

image

image

image

elsazac:
> and if does not resolve? Exception -> show its illegal! Currently it shows a path that has nothing todo with the current input :-(

I don't get any exceptions in Mac, Looks like this is again windows specific.

image

jukzi:
on mac an linux you could probably find other illegal characters like 0 byte

elsazac:
> Also: when i start the dialog from File/switch workspace, then the hint is initially empty: 

This is fixed now, initialise the label along with the combo.

>and as soon as the "copy settings" is expanded the "Browse" button is again to high:  

This again looks like a windows specific behaviour(I don't see this happening in Mac). I can relate to this because the button's height hint is left open in the dialog so its possible that the button gets resized as per the layout logic:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/2f1e3be312d3bebb8c544257751b3d075f32e615/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/Dialog.java#L956-L962

>i think theese error have been reported now multiple times. please fix them before asking for review again

It's possible to explicitly apply the height hint for the button for this dialog but I cannot test it as I don't have Windows.Is there anyone else willing to test if I push a change?

laeubi:
I have now looked at the implementation an I think this is simply because @elsazac tried to merge the layout of the combo with a label that does not really will work well here.  Actually `FillLayout` is not suitable here and the reason it might look odd.

I'll try to propose a (small) fix here that then can be used to further improve this in general the Layout of the launcher dialog is a bit odd anyways last time I tried to improve it, the main problem is that SWT lacks good layout managers, I recently added BorderLayout that has some better behavior and proposed already to have even more (modern) Layouts to SWT that might be required to fix it all together but thats not finished yet.

laeubi:
Please see
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1502

this of course can be further enhanced but should fix the layout issue already.

elsazac:
Closing this, as https://github.com/eclipse-platform/eclipse.platform.ui/pull/1502 is a more appropriate PR that effectively resolves the layout issue and has already been merged.

Issue #1183, opened 2023-09-29
Random failing VirtualLazyTreeViewerTest.testContains


HeikoKlare:
The test case `org.eclipse.jface.tests.viewers.VirtualLazyTreeViewerTest.testContains` is randomly failing, e.g., here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1086/checks?check_run_id=17254557444

```
element must be available on the viewer
junit.framework.AssertionFailedError: element must be available on the viewer
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.jface.tests.viewers.AbstractTreeViewerTest.testContains(AbstractTreeViewerTest.java:318)
```

May be related to #1005.

fedejeanne:
Failed again in [this build](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/6772043819/job/18403747805?pr=1280) of #1280

jukzi:
seems to fail on mac reproducible
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1537/checks?check_run_id=20724274542

```
element must be available on the viewer
java.lang.AssertionError: element must be available on the viewer
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.jface.tests.viewers.AbstractTreeViewerTest.testContains(AbstractTreeViewerTest.java:342)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

HeikoKlare:
I did some debugging of the issue on Mac but since the issue occurs randomly, it is hard to reproduce (in particular in debug mode).

Several other test cases have been disabled for trees with lazy content providers on Mac (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=347491). When comparing the involved functionality of those tests with the one in the this random failing test, the failures seem to have the same root cause. Thus, I propose to disable this test case on Mac just like the others. The complete functionality for lazy content providers seems to somehow be broken on Mac anyway (as can be seen by all the other test cases being disabled).

HeikoKlare:
Since the error message is not unique and line numbers change, just for the record: The assertions that fails it the first within the test case that expects the viewer to contains something, i.e.:
[...]

PR #1500, opened 2024-01-15
Apply theme color to status message in FindReplaceDialog

HeikoKlare:
The status shown in the `FindReplaceDialog` is currently written in red (on input error) or in black. When setting the black font color, it does not consider the theme style applied to the dialog.

With this change, the status message color is properly set to the themed one used by the dialog. When using light theme, nothing changes.

Before:
image

After:
image

Light theme:
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -1304,8 +1304,9 @@ class FindReplaceDialog extends Dialog {
 
 		String dialogMessage = status.accept(new FindReplaceLogicMessageGenerator());
 		fStatusLabel.setText(dialogMessage);
-		fStatusLabel.setForeground(null);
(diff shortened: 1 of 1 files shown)


BeckerWdf (approved):

vogella:
LGTM, thanks @HeikoKlare 

Heiko, you are our test master, so I assume you know how to handle the failing / removed tests.

HeikoKlare:
Yes, failing tests seem to be random failures as a result of regression tests added with #1132. I'm currently investigating the cause.

AObuchow:
@HeikoKlare just wanted to thank you for working on resolving this, as this bug has been present for [many years](https://github.com/AObuchow/Eclipse-Spectrum-Theme/issues/136) :)

AObuchow (approved):

HeikoKlare:
Thanks for the pointer, @AObuchow! Then this should also solve the downstream theme issue :slightly_smiling_face: 

I have also verified that nothing breaks on Linux (on which theming the message has properly worked before, as also noted in AObuchow/Eclipse-Spectrum-Theme#36).

image

Test failures have been resolved and the remaining failure is an independent, random fail documented in #1183

PR #1501, opened 2024-01-15
Move identification of find/replace operation success to status object

HeikoKlare:
The recently introduced `IStatus` objects used by the `FindReplaceLogic` to notify consumers (like the `FindReplaceDialog`) about the result of a find/replace operation only provide the information about the success of an operation implicitly. This is why the FindReplaceDialog currently needs to consider the type of an `IStatus` to extract this information.

With this change, the `IStatus` objects provide success information via a `wasSuccessful()` method, which relieves the `FindReplaceDialog` from manually distinguishing and processing different types of status objects.

This is a deferred design improvement (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/1132#pullrequestreview-1820230660) as a planned follow-up to #1132,

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/FindAllStatus.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/FindAllStatus.java
@@ -34,4 +34,9 @@ public class FindAllStatus implements IFindReplaceStatus {
 		return true;
 	}
 
+	@Override
+	public boolean wasSuccessful() {
+		return selectCount > 0;
+	}
+
 }
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/FindStatus.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/FindStatus.java
@@ -40,4 +40,11 @@ public class FindStatus implements IFindReplaceStatus {
 		return messageCode != StatusCode.READONLY;
 	}
 
+	@Override
+	public boolean wasSuccessful() {
+		// Also report StatusCode.WRAPPED as unsuccessful, because it implicitly
+		// includes NO_MATCH, as before wrapping no further match was found
+		return false;
+	}
+
 }
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/IFindReplaceStatus.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/IFindReplaceStatus.java
@@ -14,8 +14,8 @@
 package org.eclipse.ui.internal.findandreplace.status;
 
 /**
- * Interface for statuses that can occur while performing
- * Find/Replace-operations.
+ * Interface for statuses that can occur while performing find/replace
+ * operations.
  */
 public interface IFindReplaceStatus {
 	public <T> T accept(IFindReplaceStatusVisitor<T> visitor);
@@ -25,4 +25,10 @@ public interface IFindReplaceStatus {
 	 * that the target is writable on replace operations}
 	 */
 	public boolean isInputValid();
+
+	/**
+	 * {@return whether the find/replace operation was successful, i.e., whether
+	 * some matching string was found or replaced}
+	 */
+	public boolean wasSuccessful();
 }
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/InvalidRegExStatus.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/InvalidRegExStatus.java
@@ -41,4 +41,9 @@ public class InvalidRegExStatus implements IFindReplaceStatus {
 		return false;
 	}
 
+	@Override
+	public boolean wasSuccessful() {
+		return false;
+	}
+
 }
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/NoStatus.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/NoStatus.java
@@ -32,4 +32,9 @@ public class NoStatus implements IFindReplaceStatus {
 		return true;
 	}
 
+	@Override
+	public boolean wasSuccessful() {
+		return true;
+	}
+
 }
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/ReplaceAllStatus.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/ReplaceAllStatus.java
@@ -34,4 +34,9 @@ public class ReplaceAllStatus implements IFindReplaceStatus {
 		return true;
 	}
 
+	@Override
+	public boolean wasSuccessful() {
+		return replaceCount > 0;
+	}
+
 }
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -68,11 +68,7 @@ import org.eclipse.ui.internal.findandreplace.FindReplaceMessages;
 import org.eclipse.ui.internal.findandreplace.HistoryStore;
 import org.eclipse.ui.internal.findandreplace.IFindReplaceLogic;
 import org.eclipse.ui.internal.findandreplace.SearchOptions;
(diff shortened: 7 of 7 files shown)


laeubi:
@HeikoKlare An Status can also carry a `CODE` (integer) what is usually used to have additional information.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/IStatus.java
```diff
@@ -14,8 +14,8 @@
 package org.eclipse.ui.internal.findandreplace.status;
 
 /**
- * Interface for statuses that can occur while performing
- * Find/Replace-operations.
+ * Interface for statuses that can occur while performing find/replace
+ * operations.
  */
 public interface IStatus {
```
laeubi: Why one needs a new `IStatus` and can't use the Eclipse `IStatus` interface here?

merks: I think it's **really unfortunate** to reuse the name IStatus because that can only lead to unfortunate confusion and we already see that here.  When I saw the comment from the PR in my inbox, I thought, "Are you kidding my, they are doing to change IStatus?!"...

HeikoKlare: This is only an internal status object for exchange of very specific information between find/replace business logic and UI, for which Eclipse's `IStatus` is neither required nor really appropriate. From a pure technical perspective, the name of this interface is `[...].findandreplace.IStatus`, as, of course, the package hierarchy also belongs to the name (c.f. types like `Event`).
But I totally see your point that the name clash is bad for something so essential as `IStatus` and that it makes sense to give it a more specific name. I will extend the PR to also adapt the name of that interface.

merks: @HeikoKlare 

Yes, renaming would be great because I bet @iloveeclipse thumbs up on @laeubi's comment would not happen (and the comment itself) if there wasn't a name collision.

image

I.e., there is no relation between these two and so one would never compare them and suggest to use the other if this one had a more specific name the reflected it's specific purpose within the find/replace infrastructure.

laeubi: @HeikoKlare why is `IStatus` (eclipse) not appropriate here? I must confess I have not reviewed the original in detail so its maybe a bit late now but I don't see this status being much different (except that you can visit childs, what could be either a static helper method or be added to `IStatus` (eclipse) as well...

HeikoKlare: Eclipse's `IStatus` is a general concept to provide all kinds of status information about the outcome of an operation to be processed uniformly at different places. Here, we just need a very restricted set of specific status information that is only passed between the `FindReplaceLogic` and its consumers (i.e. find/replace UIs).
Using Eclipse's `IStatus` would result in worse design, as we have to encode information into "magic" integer values rather than providing a suitable API and as we would have to make the status objects provide messages, whereas we introduced these status objects to move the responsibility for generating a UI representation (such as a message) away from the business logic into the UI. At the same time, at least from my point of view there is no real benefit in using Eclipse's `IStatus` here. If you see a benefit in using Eclipse's `IStatus` that justifies the loss in specificity/design, please let me know and we can of course adapt that (everything here is internal anyway).

merks: I will assert again that we would not be having this discussion at all if it had a different name to begin with.  :scream:  I don't think we need to make this foot fit into a the IStatus shoe...

laeubi: My only concern here is that you have an interface and then check for specific implementations... that actually shows that the "clean api design" is actually not an api (maybe a facade) so probably you more want a search context (!) instead that allows to pass information you need around and then maybe has a Type enum...

HeikoKlare: > My only concern here is that you have an interface and then check for specific implementations... that actually shows that the "clean api design" is actually not an api (maybe a facade) so probably you more want a search context (!) instead that allows to pass information you need around and then maybe has a Type enum...

That's exactly what this PR is supposed to fix 🙂

laeubi: If you look at the public API of the `IStatus` classes:

| **IStatus (search)** | **IStatus (eclipse)**           |
|----------------------|---------------------------------|
| wasSuccesfull        | isOk                            |
| isInputValid         | getSeverity() == IStatus.CANCEL |
| accept               | getChildren                     |

for me it looks like as the public api interface does matches 1:1 but IStatus(Eclipse) is a bit more options so for me it seems good to reuse an existing api/interface than inventing a new one.

HeikoKlare: Please don't get me wrong: I am totally in favor of reusing implementations whenever possible, as it has a clear benefit (changes to that code, be they corrective, adaptive or perfective, apply to all usages). But this is not about code duplication but about an interface, which is designed to fit for a specific purpose. And, unfortunately, I do not understand the general idea of reusing an existing (rather generic) interface just because you can implement it in some way that makes it (syntactically) applicable in some scenario, because I do not see any benefit in that. Maybe you can clarify what you would expect from that reuse.
On the contrary, I see significant drawbacks in using a general-purpose rather than a tailored interface: the interface that we use here is _small_. It provides two methods that you could mimic with Eclipse's `IStatus` and provides one additional method (`accept()`, which is _not_ equal to `getChildren()`) that is not provided by Eclipse's `IStatus` at all. Eclipse's `IStatus` has 9 methods of which 7 would _not_ be used, so some default implementations for these methods without any meaning would be required. For the remaining two methods, you would need to document how the general interface is mapped to the specifics needed in this scenario, which in particular concerns the implicit meaning of `severity = IStatus.CANCEL`. You would have to place the documentation of current `isInputValid()` at some less reasonable point than at an according method that is called to retrieve that state.

And as I mentioned, Eclipse's `IStatus` assumes that a message is provided. The status objects for find/replace results are supposed to explicitly _not_ provide a message, as the status is only supposed to carry an abstract description of the status while the decision how to report it should be up to the UI. That is something that is infeasible for Eclipse's `IStatus` due to its open usage scope allowing all kinds of unknown subclasses, but it is reasonable in case all subclasses are known (like in this case).

HeikoKlare: I've opened #1503 for the rename to keep concerns separate.

laeubi: As said I'm not in the details and finally its up to you to decide, so I just can share some observations here, e.g. if you look at

`InvalidRegExStatus` it is used when an exception is caught. wrapping an exception, that's exactly what `IStatus` already offers. Then it has a getMessage() method what actually will be similar to what `IStatus` is doing in the first place with a plain `Status.error(....)`.

Then there is `NoStatus` what does effectively is what we Have with `Status.OK_STATUS`, then there is `ReplaceAllStatus` what carries an int that is later used to construct a message with it, so could be Status.ok("the message") ... FindStatus could be an OK message with a CODE and so on. 

So it *looks* like everything could be mapped already without any new interface or any new implementation at all ...

laeubi:
Looking a bit closer `IStatusVisitor` seems to be more a replacement for a `instanceof` check and the only implementation is `FindReplaceLogicMessageGenerator` so for me seems overhead here and could be removed at all. All implementation of the interface always ever only "visit" the instance iteself `visit(IStatus status)` is even never called at all.

HeikoKlare:
> Looking a bit closer `IStatusVisitor` seems to be more a replacement for a `instanceof` check

It is not a replacement, it is a proper way of dealing with the tyranny of the dominant decomposition. With Java 21, an alternative (that would also fit here) will be record patterns with sealed classes.

> All implementation of the interface always ever only "visit" the instance iteself `visit(IStatus status)` is even never called at all.

Yes, that's how a visitor is implemented 🙂

laeubi:
It don't feel to bring any benefit here I must confess, but as said actually its you choice it seems a lot of additional implementation to save a single instanceof check...

vogella (approved):

HeikoKlare:
I really like such discussions about different perspectives on design. Though I have to admit that I do not really know how to proceed with the discussion if only selective parts of my responses are considered. I tried to explain in quite some detail the rationale behind the current design from a software engineering perspective, but if it is only about feeling and how to save code
> it seems a lot of additional implementation to save a single instanceof check

and not about design, I think the essential problem is that we simply follow different goals.

If there is interest in continuing with the discussion, let's open an issue for that. The current design decision has been made with #1132 and only became visible with this PR, which just makes an improvement to the existing design.

The failing test is independent and already documented: #1005.

PR #1502, opened 2024-01-15
Fix the layout of new workspace label for unexpected path

laeubi:
Currently the label is combined with the combo what gives strange results.

This moves the label out of the this into an own row and also hide it if there is no information to show.


elsazac:
Thanks @laeubi for the PR. The layout looks good. I have tested this only on Mac and one thing that I noticed is, the warning symbol isn't shown.

image

laeubi:
I don't know where the warning symbol is supposed to come from but at least under SWT I haven't seen it before. If we really want a warning icon ist maybe better to use a decorator on the combo or similar, for now I would just go without it.

laeubi:
@elsazac the warning is now back!

This is how it looks on linux
grafik

elsazac:
> @elsazac the warning is now back!

Yes. I can see that now. This looks good to me. 
Thanks @laeubi

merks:
It (Switch workspace) looks like this on Windows:

image

iloveeclipse:
Let's merge that, it is by any means better as current state.

PR #1503, opened 2024-01-15
Rename status for find/replace from IStatus to IFindReplaceStatus

HeikoKlare:
To avoid confusion with the existing `IStatus` interface for representing Eclipse operation results, this change renames the specific status interface for find/replace action results from `IStatus` to `IFindReplaceStatus`. It also adapts the `IStatusVisitor` to `IFindReplaceStatusVisitor` accordingly.

See the discussion in #1501.


vogella:
LGTM and to avoid any bad naming in tomorrows I-build, I suggest to merge this. If I do not see negative comments soon, I merge this one (or @HeikoKlare can merge it now, if he thinks that is the right thing to do).

PR #1506, opened 2024-01-15
Make test cases in FindReplaceDialogTest independent from each other

HeikoKlare:
The test cases in `FindReplaceDialogTest` currently depend on each other. The class defines the tests to be executed in lexicographic order, as test cases may not reset settings in the `FindReplaceDialog` they have changed during execution. In addition, the tests do not only use reflection to access the SWT controls but also to call internal logic as selection events are not simulated properly.

In order to ensure proper test isolation and make tests run (and potentially fail) in a reproducible and runtime-equivalent setup, this change makes the `FindReplaceDialog` settings being reset on each execution and adds assertions for a proper setup in each test case. It also ensures that selection events are executed properly without calling internal methods of the `FindReplaceDialog`. Finally, the change also adds a missing default setup for the search scope in the FindReplaceDialog that was revealed by the improved test setup.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -209,6 +209,7 @@ class FindReplaceDialog extends Dialog {
 	public FindReplaceDialog(Shell parentShell) {
 		super(parentShell);
 		findReplaceLogic = new FindReplaceLogic();
+		findReplaceLogic.activate(SearchOptions.GLOBAL);
 		fParentShell = null;
 		fDialogPositionInit = null;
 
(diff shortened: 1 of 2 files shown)


HeikoKlare:
This should resolve the random test failures that occurred in [a recent CI run for #1500](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1500/checks?check_run_id=20487004545) or at least give better insights concerning the reasons for such failures due to proper test isolation.

testDisableWholeWordIfNotWord (org.eclipse.ui.workbench.texteditor.tests.FindReplaceDialogTest)
```
java.lang.AssertionError
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.eclipse.ui.workbench.texteditor.tests.FindReplaceDialogTest.testDisableWholeWordIfNotWord(FindReplaceDialogTest.java:178)
```

testDisableWholeWordIfRegEx (org.eclipse.ui.workbench.texteditor.tests.FindReplaceDialogTest)
```
java.lang.AssertionError
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.eclipse.ui.workbench.texteditor.tests.FindReplaceDialogTest.testDisableWholeWordIfRegEx(FindReplaceDialogTest.java:152)
```

Issue #1517, opened 2024-01-18
Random failing PlatformUITest#testCreateAndRunWorkbench()


HeikoKlare:
The test case `org.eclipse.ui.tests.rcp.PlatformUITest#testCreateAndRunWorkbench()` has been randomly failing in recent builds. One recent failure happened in this build: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1514/checks?check_run_id=20614845943

```
Async from qualified thread did not run during startup.  See RCPTestWorkbenchAdvisor.preStartup() expected:<true> but was:<false>
java.lang.AssertionError: Async from qualified thread did not run during startup.  See RCPTestWorkbenchAdvisor.preStartup() expected:<true> but was:<false>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.eclipse.ui.tests.rcp.PlatformUITest.testCreateAndRunWorkbench(PlatformUITest.java:72)
```

vogella:
## Why the Phaser approach was failing

The initial fix attempt used `java.util.concurrent.Phaser` to synchronize the async/sync operations, but this approach had fundamental issues:

### Problem 1: Timing vs. Execution
The Phaser was counting arrivals when operations were **scheduled**, not when they **executed**:
```java
display.asyncExec(() -> {
    asyncWithDisplayAccess = !isSTARTED();
});
// Phaser counted down here - but the runnable hasn't executed yet!
asyncPhaser.arriveAndDeregister();
```

This meant `postStartup()` could mark `started=true` before the runnables actually ran on the UI thread, defeating the entire purpose of the synchronization.

### Problem 2: Event Processing Issues
The inline event processing using `Display.getCurrent()` and `readAndDispatch()` wasn't reliable during the startup phase, leading to timeouts.

### Solution: CountDownLatch
The fix uses `CountDownLatch` that counts down **inside** the runnables after they execute:
```java
display.asyncExec(() -> {
    asyncWithDisplayAccess = !isSTARTED();
    displayAccessLatch.countDown();  // Count down AFTER execution
});
```

This ensures `postStartup()` waits for actual execution, not just scheduling. The approach is:
1. Initialize latch with count=2 (for async and sync with DisplayAccess)
2. Each runnable counts down after setting its variable
3. `postStartup()` waits for latch to reach zero
[...]

PR #1518, opened 2024-01-18
Migrate RCP tests to JUnit 5 and improve startup assertions #1517

HeikoKlare:
This change migrates the (non-performance) tests in `org.eclipse.ui.tests.rcp` to JUnit 5. Assertions are also replaced with JUnit 5 assertions. Since most `assertEquals()` calls contained actual and expected in wrong order, they do not have to be changed as usual when migrating from JUnit 4 to 5 assertions.

In addition, the assertions in `PlatformUITest#testCreateAndRunWorkbench()` are improved so that all of them are evaluated even if multiple of them fail. This serves as a means to ease identifying the reasons for this test case to randomly fail.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1517

### DIFF
--- a/tests/org.eclipse.ui.tests.rcp/RCP Test Suite.launch
+++ b/tests/org.eclipse.ui.tests.rcp/RCP Test Suite.launch
@@ -1,47 +1,50 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <launchConfiguration type="org.eclipse.pde.ui.JunitLaunchConfig">
-<stringAttribute key="application" value="org.eclipse.pde.junit.runtime.coretestapplication"/>
-<booleanAttribute key="askclear" value="false"/>
-<booleanAttribute key="automaticAdd" value="true"/>
-<booleanAttribute key="automaticValidate" value="false"/>
-<stringAttribute key="bootstrap" value=""/>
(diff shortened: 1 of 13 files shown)


tests/org.eclipse.ui.tests.rcp/RCP Test Suite.launch
```diff
@@ -1,47 +1,50 @@
...
-<stringAttribute key="product" value="org.eclipse.sdk.ide"/>
-<booleanAttribute key="run_in_ui_thread" value="true"/>
-<booleanAttribute key="show_selected_only" value="false"/>
-<stringAttribute key="templateConfig" value=""/>
-<booleanAttribute key="tracing" value="false"/>
-<booleanAttribute key="useCustomFeatures" value="false"/>
-<booleanAttribute key="useDefaultConfig" value="true"/>
-<booleanAttribute key="useDefaultConfigArea" value="false"/>
-<booleanAttribute key="useProduct" value="false"/>
+    <stringAttribute key="application" value="org.eclipse.pde.junit.runtime.coretestapplication"/>
```
HeikoKlare, line 3: In my opinion, it makes sense to keep this launch configuration since these tests have to be executed in headless mode, which otherwise has to be configured manually (and this necessity must also be known).

PR #1519, opened 2024-01-19
avoid overwrite popup when possible

### LINKED ISSUE #617: Improvement for the WizardProjectsImportPage
Hello,

The org.eclipse.core.internal.resources.Project class has recently been changed. The '.settings' folder is automatically created containing a preference file with the encoding in it.

When i use the 'import existing projects within workspace' wizard the following happens:
- i select multiple projects (with already an '.settings' folder) 
- Select the option 'copy projects into workspace'
- For every project an popup is shown that there is already existing '.settings' folder.

From a customer perspective this popup is strange because this is triggered because of the working within the WizardProjectsImportPage. First a project is created and opened within the workspace and then the resources are copied.

It would be nice if this popup would not be shown.

Kind regards,

Richard

elsazac:
When files are copied as part of project import , the overwrite flag is set for certain files but was never in effect- the prompt logic was hiding the overwrite check. To fix flip the order

Fixes: #617


jukzi:
@richardgroote does this commit meet your requirements? can you test please?

elsazac:
Let me know if anything pending on this PR ?

PR #1522, opened 2024-01-19
Add a warning when the path itself is invalid

### LINKED ISSUE #1513: ChooseWorkspaceDialog resolved path issues
no reason to Log an exception when user enters illegal path:
```
java.nio.file.InvalidPathException: Illegal char <*> at index 39: C:\Users\jkubitz\platforM-2024-01-10\ws*
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:204)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:175)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:231)
	at java.base/java.io.File.toPath(File.java:2401)
	at org.eclipse.ui.internal.ide.ChooseWorkspaceDialog.getUnexpectedPathHint(ChooseWorkspaceDialog.java:466)
```

and the resolved path is still wrong:
image

and still does not show the right capitalization:
image

laeubi:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1513

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
@@ -17,6 +17,8 @@ package org.eclipse.ui.internal.ide;
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


laeubi:
@jukzi @HeikoKlare please verify, I'm on Linux and it seems almost all characters are allowed

jukzi:
null (ASCII 0) is not allowed on linux

laeubi:
> null (ASCII 0) is not allowed on linux

Yes and it is super easy to type null (ASCII 0)  on a standard keyboard 🙄

> I'm on Linux and it seems **almost** all characters are allowed

jukzi:
`To achieve the same in Ubuntu, you must type Ctrl + Shift + U, and then type Unicode codepoint.`
https://askubuntu.com/questions/88347/how-can-i-type-ascii-characters-like-alt-numpad-in-windows

laeubi:
> `To achieve the same in Ubuntu, you must type Ctrl + Shift + U, and then type Unicode codepoint.` https://askubuntu.com/questions/88347/how-can-i-type-ascii-characters-like-alt-numpad-in-windows

Or as windows users that complained that they can easily get their Eclipse broken to at least verify it works for their cases if they don't mind to just fix it ...

jukzi (changes requested):
tested: catches the most annoying exception, but still path not capitalized and launch button still enabled with wrong path

mickaelistria (changes requested):
We can always discuss hypothetical cases where a bit of code is not sufficient; but in the meantime, this change is clearly an improvement over current master, fixes a known issue and won't introduce any regression; so let's merge it as soon as PR build validates.

mickaelistria (approved):

laeubi:
> but still path not capitalized

I don't understand what this should mean, a **capitalized** path is not mandatory on windows and I don't see how it would improve anything here.

jukzi:
The problem is that it does not fully fix 1522 and the contributors keep fixing only part of problems, that way the IDE becomes more and more corrupt.

laeubi:
> that way the IDE becomes more and more corrupt.

If your IDE is corrupt you probably want to contact your local anti-corruption office 🙈

jukzi:
> I don't understand what this should mean, a **capitalized** path is not mandatory on windows and I don't see how it would improve anything here.

The path shown in the hint does not shown the path that is actually later used and shown later

mickaelistria:
> the contributors keep fixing only part of problems

Yes, and that's good; we should welcome that as it is. Asking for more and adding pressure when the contributor see no clear value gain is reducing the time-to-market of incoming contributions and reduces agility of the project and its ability to welcome innovation. Doing things incrementally is usually more productive, particularly as we don't have control on developer resources to always ask for more.
The criterion for a merge is "Does it make something work better? Does it cause some new issue?", as long as answers are respectfully yes and no, we must merge the contribution and deliver the value, and thank the contributors.

laeubi:
> The path shown in the hint does not shown the path that is actually later used

The dialog can only give a hint, the raw string is only passed further down to other parts in the system that process the string in different ways, so without major refactoring you can't archive what you describe here and it has never worked like this before.

So from my point of view we can try to catch some common problems that *likely* will cause issues/confusion, if you want to go further feel free to propose a PR for that, at best all this would be working on (java!) Path API from top to bottom of course.

jukzi:
The problem is not an "incremental" work but stopping at a point that is worse then before. If someone wants a new feature it should be complete. Better revert the whole thing if the contributors don't plan to work on a consistent solution.

jukzi (approved):
that  looks nice 
image

laeubi:
The main problem is that more effort is put into complaining and arguing than actually fixing a trivial bug.

Also such concerns should be raised when reviewing features and not afterwards, also please note that **I don't have created, reviewed or asked for that feature** I only helped with a layout issue (whether or not it is new does not matter in my opinion)!

Nevertheless I now tested this on Linux and it works for all obvious cases as expected:

- [x] Empty path -> no label, button disabled
- [x]  `~` in path -> shows warning
- [x] Non writable path -> shows warning, button disabled

HeikoKlare (approved):
Thanks for this improvement, @laeubi! The code looks good and works as expected. It is clearly an improvement over the current state: the input does not freeze anymore but shows a hint. So I agree with the argument of @mickaelistria that this should be merged as soon as possible.

I have not yet verifed whether it also solves the problem of Eclipse crashing once an invalid path has been entered and written tot the preferences, but since it is related to the same place in the code, I will expect that to be fixed as well.

jukzi:
> I have not yet verifed whether it also solves the problem of Eclipse crashing once an invalid path has been entered and written tot the preferences, but since it is related to the same place in the code, I will expect that to be fixed as well.

based on the stacktrace in the log, it would not happen anymore, and with button disabled its impossible to get into that state anyways :-)

HeikoKlare:
> based on the stacktrace in the log, it would not happen anymore, and with button disabled its impossible to get into that state anyways :-)

That's partly true, as button is _not_ disabled in that case:
image

mickaelistria:
Build passed and 3 positive review, let's merge.
Thansk to everyone here.

PR #1524, opened 2024-01-19
fix missing hashCode

jukzi:
"The type % should also implement hashCode() since it overrides Object.equals()"


jukzi:
ignoring random fail: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1525

PR #1536, opened 2024-01-22
make path names UNIX-friendly - search bundle

elsazac:
remove white spaces from path names


merks:
I don't think you've fixed the .classpath file:

image

I generally prefer to show all .* files.

image

image

Not to digress on the actual issue being addressed here, I wonder if the ant scripts (such as the one modified by this PR) are just noise from ancient build days:

image

elsazac:
Thanks @merks. Fixed. please have a look.

merks:
I'm going out for a while.  I'll check the build results when I'm back...

PR #1537, opened 2024-01-22
use try-with-resource where hinted

jukzi:


bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/part/ResourceTransfer.java
```diff
@@ -129,25 +129,21 @@ protected void javaToNative(Object data, TransferData transferData) {
 
 		int resourceCount = resources.length;
 
-		try {
-			ByteArrayOutputStream out = new ByteArrayOutputStream();
-			DataOutputStream dataOut = new DataOutputStream(out);
+		ByteArrayOutputStream out = new ByteArrayOutputStream();
```
akurtakov, line 132: Shouldn't out variable be handled by try-with-resources too? With that change it's no longer closed but it used to be in the old code.

HeikoKlare: I guess this is because `ByteArrayOutputStream` does not need to be closed (`close()` actually is a no-op there).

jukzi:
ignoring https://github.com/eclipse-platform/eclipse.platform.ui/issues/1183

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/part/ResourceTransfer.java
jukzi: see ByteArrayOutputStream#close - its only memory not an actual resource

PR #1541, opened 2024-01-22
Fix replace behavior of FindReplaceLogic/Dialog for RegEx search #1540

### LINKED ISSUE #1540: Replace/Find button skips every second occurrence

## Steps to reproduce

- Create an "Untitled Text Document"
- Enter following content:
```
﻿﻿testa!
testb!
testc!
```
- Set the cursor to the start of the document
- Hit Command+F
- Fill "Find" field with `(\w)!`
- Fill "Replace with" field with `\1 !`
- Set checkbox "Regular expressions"
- Hit Find button
- Hit Replace/Find button

Screenshot 2024-01-22 at 21 34 03

I expected:
```
﻿﻿testa !
testb!
testc!
```

But got:
```
﻿﻿testa!
testb !
testc!
```
Screenshot 2024-01-22 at 21 34 28

* Ubuntu 22.04.3 LTS
* Eclipse SDK: Version: 2024-03 (4.31) Build id: I20240121-1800

## Community

HeikoKlare:
The replace functionality of the `FindAndReplaceLogic` checks whether the current selection fits to the current search string and otherwise performs a find operation first. To this end, it compares the text selection with the search string, even if regex search is used. In regex search mode, this check will usually fail and a find operation is performed even though a correct string is already selected. This results in the replacement of the next instead of the current match.

With this change, the check for whether the currently selected string fits to the search string properly considers regex search mode. In consequence, all replace operations consider the current selection for replacement if fitting. It also adds according regression tests.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1540

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
 import org.eclipse.swt.custom.BusyIndicator;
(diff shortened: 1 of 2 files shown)


HeikoKlare:
Failing test is documented as random failure: #751

PR #1543, opened 2024-01-23
show search pattern as the quick search dialog title

### LINKED ISSUE #478: Quick Search show search term in title bar to support multiple search windows
Please display the search term in the title bar of the search window.
Request:  when multiple Quick Search windows are open, currently they are all titled "Quick Search".  When minimized to the title bar, they are anonymous, making it difficult to find the one desired and to avoid duplicate windows (when duplicates are not desired; sometimes I want to keep old search results AND have new search results).  
Please display the search term in the title bar.
Currently using:
Eclipse IDE for Embedded C/C++ Developers (includes Incubating components)

Version: 2022-06 (4.24.0)
Build id: 20220609-1112
With QuickSearch plugin:
	Eclipse.org	Quick Search	1.1.300.v20211203-1705	org.eclipse.text.quicksearch

elsazac:
Fixes: #478

### DIFF
--- a/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
+++ b/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
@@ -1268,6 +1268,7 @@ public class QuickSearchDialog extends SelectionStatusDialog {
 	 */
 	protected void applyFilter(boolean force) {
 		QuickTextQuery newFilter = createFilter();
+		getShell().setText(Messages.QuickSearchDialog_title + " - " + pattern.getText()); //$NON-NLS-1$
 		if (this.searcher==null) {
 			if (!newFilter.isTrivial()) {
 				//Create the QuickTextSearcher with the inital query.


bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
```diff
@@ -1268,6 +1268,9 @@ protected QuickTextQuery createFilter() {
 	 */
 	protected void applyFilter(boolean force) {
 		QuickTextQuery newFilter = createFilter();
+		if(newFilter != null) {
+			getShell().setText(newFilter.getPatternString());
```
merks: I don't think the pattern string should replace the title.   The new filter can't be null and the pattern string of that filter is the same as pattern.getText().

elsazac: image

@merks I removed the null check and used `pattern.getText` instead of `getPatternString`. However I don't know what is the benefit of one over the other. I am able to get the title changed in both the ways.

merks:
The search string should not replace the descriptive title. It should be added to the title.

elsazac:
> The search string should not replace the descriptive title. It should be added to the title.

done. Thanks

merks:
I’m on a phone away from a computer.   Maybe making it look more like the workbench title shows the selection would be better.

elsazac:
> I’m on a phone away from a computer. Maybe making it look more like the workbench title shows the selection would be better.

@merks Quick search dialog is a modeless one, so it allows multiple search children. So it makes sense for individual child dialogs to show pattern texts instead of the workbench, or am I misreading your suggestion.?

merks:
I think the workbench separates the title from the selection text with a hyphen not a colon. I’m just suggesting to use a similar style.

elsazac:
> I think the workbench separates the title from the selection text with a hyphen not a colon. I’m just suggesting to use a similar style.

ok, understood the suggestion. I have made the changes.

PR #1544, opened 2024-01-23
fix "Potential null pointer access" warnings

jukzi:


jukzi:
image

BeckerWdf:
> image

Where is that from?

jukzi:
ignoring random fail on mac: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005

jukzi:
> Where is that from?

click here: 
image

PR #1551, opened 2024-01-24
[test] fix "The switch statement should have a default case"

jukzi:


jukzi:
unrelated fail on mac:

```
Two files should be selected by default
junit.framework.AssertionFailedError: Two files should be selected by default
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.testMultiSelectionAndSomeInitialNonExistingSelectionWithInitialPattern(ResourceInitialSelectionTest.java:283)
```

PR #1554, opened 2024-01-24
fix "static" problem markers

jukzi:


jukzi:
somehow jenkins reports more issues fixed then this commit is big, looks like an error :-(
image

HeikoKlare:
> somehow jenkins reports more issues fixed then this commit is big, looks like an error :-(

Probably because the reference build (249) is not the current master build (252). There have been more recent master builds in which several issues have been fixed, e.g., https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/252/

jukzi:
> Probably because the reference build (249) is not the current master build (252).

Then it's wrong that either
a) it compares to an outdated build or
b) the code is build as if rebased instead in the state as commited - @iloveeclipse there was setting for that behavior somewhere

HeikoKlare:
I agree that this does not make sense. A potential reason is that build 252 also had one _new_ issue and was thus not marked as the new reference build. Maybe that is the configuration that should be fixed: last successful master build must always be marked as reference build.

jukzi:
> last successful master build must always be marked as reference build.

do you know how to change that?

jukzi:
After rebase the numbers are correctly shown  - fixed 2:
image
image

HeikoKlare:
> do you know how to change that?

Unfortunately not. It was only a guess based on the results of those builds. If I understand the [documentation ](https://www.jenkins.io/doc/pipeline/steps/warnings-ng/)correctly, as long as quality gates passes, the reference build should be updated. That would mean there has to be a different reason, as the quality gate passed even though a new issue was introduced.

Issue #891, opened 2022-05-26
Press Save or Enter Empty multiple times to appear in a code prompt box


baily-s:
Pressing the Save shortcut multiple times or entering other invalid characters will pop up a code prompt box.

mickaelistria:
Can you please provide detailed steps to reproduce from a fresh and recent installation of Eclipse IDE?

baily-s:
1. Mac M1 chip version
2. Eclipse 2022-3 version
3. Set auto activation triggers for java:.abcdefghijklmnopqrstuvwxyz
4. At this time, press the shortcut key (ctrl+s) multiple times and a prompt box will pop up
5. I guess auto activation triggers should be triggered, but my intention is to save
thakns

mickaelistria:
I cannot reproduce it on Fedora 36.

elsazac:
The linked PR has been merged, so this can be closed.

PR #1556, opened 2024-01-25
spare control characters from content assist trigger

elsazac:
Fixes :#891

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
@@ -340,7 +340,7 @@ public class ContentAssistant implements IContentAssistant, IContentAssistantExt
 			if (e.character == 0 && (e.keyCode & SWT.KEYCODE_BIT) == 0)
 				return;
 
-			if (e.character != 0 && (e.stateMask == SWT.ALT))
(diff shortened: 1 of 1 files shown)


mickaelistria:
Is Ctrl+Space trigger still working with this change?

elsazac:
> Is Ctrl+Space trigger still working with this change?

Yes it works.

elsazac:
I think `CTRL+SPACE` is handled in ProposalSelectionListener and does not intervene here.

merks:
I added some print statements to see when the method is called.  It's kind of confusing that sometimes it's called but often it's not.  I.e., one would expect it not to be called by Ctrl-S because that triggers a save action.  That's sort of true, but if the editor isn't dirty the action isn't triggered and the event does come through.  That would be the reason why pressing Ctrl-S a couple of times causes the issue.

Then I wondered about the origin  "ALT" and saw it come from this:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=176698

So I suppose this change is safe, though it still feels slightly uncomfortable.  

@akurtakov 

I'm not sure who else would be "qualified" to review this change?

@szarnekow

Often Sebastian has a keen eye.

mickaelistria (approved):
I also spent some time looking at this in more details earlier today, and this seems like a good fix. I'm not aware of anyone who would miss a content-assist popup when hitting a key with Ctrl or Cmd keys on; it seems totally safe to assume someone hitting one of those keys is *not* willing to enter text directly in the editor but want to operate a more complex operations, potentially outside of the editor, and as a consequence, that they're not in a case where automatic trigger of content-assist make sense.

merks:
@mickaelistria 

I appreciate the confirmation!

PR #1559, opened 2024-01-25
remove deprecated CSS element

### LINKED ISSUE #159: Remove deprecated show-close CSS support
Since 2012 we are issuing a warning if the deprecated swt-close CSS property is used. We should remove this property.

elsazac:
Fixes: #159


vogella:
Thanks @elsazac

PR #1592, opened 2024-01-30
increase transparency of showKeyPopUp

### LINKED ISSUE #48: The key binding shown hides cursor
There are those settings in Preferences -> keys to show the key binding when command is invoked, by mouse and/or keyboard.
I find this very helpful when programming with shared screen.

However, when the cursor is at the end of the file, each keystroke (e.g. Ctrl Arrow left/right) brings up the window and it hides the cursor position. That is very annoying.
At would be helpful if that window is direct to another location when the focus is in text editor and the cursor at the bottom.

elsazac:
Fixes: #48 

image

The text behind is partly visible now though at the expanse of transparency to the key shortcut text

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/keys/show/ShowKeysUI.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/keys/show/ShowKeysUI.java
@@ -223,7 +223,7 @@ public class ShowKeysUI implements IDisposable {
 
 			Color color = JFaceResources.getColorRegistry().get(POPUP_COLOR_BG);
 			newShell.setBackground(color);
-			newShell.setAlpha(170);
+			newShell.setAlpha(130);
 		}
 
 		@Override


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/keys/show/ShowKeysUI.java
```diff
@@ -223,7 +223,7 @@ protected void configureShell(Shell newShell) {
 
 			Color color = JFaceResources.getColorRegistry().get(POPUP_COLOR_BG);
 			newShell.setBackground(color);
-			newShell.setAlpha(170);
+			newShell.setAlpha(100);
```
BeckerWdf: Maybe now it's a little bit too transparent. I will test it locally later and check if I can come up with a better value.

BeckerWdf: How about 130?
```suggestion
			newShell.setAlpha(130);
```

BeckerWdf (changes requested):

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/keys/show/ShowKeysUI.java
elsazac: Thanks @BeckerWdf. I have made the changes. Please have a look

laeubi:
Please also test with a light theme / background ...

BeckerWdf:
> Please also test with a light theme / background ...

I did. In the light theme it also works fine - maybe even better...

elsazac:
> Please also test with a light theme / background ...

image

This is the appearance in light theme.Does this look good?

BeckerWdf:
image

BeckerWdf (approved):

laeubi:
It is a bit hard to tell, but can we make the **background** only transparent but the text **opaque**.... the screen looks like underlying text is shining though the text what makes it very hard to read.

laeubi:
> image

I find the text underneath (Moves the ...) hardly readable now if I wouldn't know what is supposed to be there...

elsazac:
> I find the text underneath (Moves the ...) hardly readable now if I wouldn't know what is supposed to be there...

@laeubi For some reason I can't see the picture you posted.

laeubi:
It is from this comment https://github.com/eclipse-platform/eclipse.platform.ui/pull/1592#issuecomment-1916751813

PR #1355, opened 2023-12-01
Splash Screen getting Flipped in MacOs Sonoma.

elsazac:
Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/751

Due to a bug in MacOS Sonoma ,Splash Screen gets flipped.As a workaround the image is flipped and returned, so that the splash screen is displayed correctly.

PR #1596, opened 2024-01-30
Future-proof workaround for flipped splash screen in macOS Sonoma

sratz:
The workaround should keep working even when macOS major version moves beyond 14.

Follow-up on #1355.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -851,7 +851,7 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 		if (splashLoc != null) {
 			try (InputStream input = new BufferedInputStream(new FileInputStream(splashLoc))) {
 				background = getImage(display, input);
-			} catch (SWTException | IOException e) {
+			} catch (SWTException | IOException | NumberFormatException e) {
 				StatusManager.getManager().handle(StatusUtil.newStatus(WorkbenchPlugin.PI_WORKBENCH, e));
 			}
 		}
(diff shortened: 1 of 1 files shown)


Phillipus:
Isn't there a better way to get the OS version from `OS` class? I'll check...

sratz:
> Isn't there a better way to get the OS version from `OS` class? I'll check...

There is, but `OS` isn't platform-independent so we can't use that from jface/workbench.

Phillipus:
> > Isn't there a better way to get the OS version from `OS` class? I'll check...
> 
> There is, but `OS` isn't platform-independent so we can't use that from jface/workbench.

Right, I should have checked before mentioning it.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -867,7 +867,7 @@ private static Image getImage(Display display, InputStream input) {
 			 * (https://github.com/eclipse-platform/eclipse.platform.swt/issues/772) ,Splash
 			 * Screen gets flipped.As a workaround the image is flipped and returned.
 			 */
-			if (System.getProperty("os.version").startsWith("14")) { //$NON-NLS-1$ //$NON-NLS-2$
+			if (Integer.parseInt(System.getProperty("os.version").split(".")[0]) >= 14) { //$NON-NLS-1$ //$NON-NLS-2$
```
elsazac, line 870: @sratz looks like the assumption here is, its not going to be fixed in v15 or beyond. Is that the case, do you have any tech note or article from Apple in this regard ? Thanks.

sratz: I am not aware of any communication from Apple regarding this.

So it's unclear whether this is a problem on cocoa side or on SWT side.

Hopefully we can find a proper fix for this soon (at which point this workaround would be removed completely anyways).

Until we do, I think it's best to assume that it's a problem in SWT and if we do nothing the issue will also show on macOS >= 15.

PR #1597, opened 2024-01-31
display install root in the plugins view

### LINKED ISSUE #1363: In the installation details add an option to the Plug-ins view to show there installation root(s)
Assume the following situation, I go the the about eclipse and then show the installation details like here

grafik

now I see I have two times Xbase Runtime library and wondering why I have two of them.

What would be very useful is that I can right-click on it and show me the installation root(s) of the bundle, that is what has requested it to be there so it can't be purged from the profile.

Another thing would be to have a button to ask to uninstall that bundle and then Eclipse tells me what else needs to be uninstalled as a consequence of that.

elsazac:
Fixes: #1363 

image

@laeubi Please let me know if this is what you need? also please note that I have addressed only the first part of your RFE in this PR.


laeubi:
@elsazac thanks for working on this but I think there is a missunderstanding, the goal is to display what of the root IUs of the profile (e.b. QueryUtil.uservisible) is responsible for a given unit to be installed.

Issue #1314, opened 2023-11-19
Action labels missing from Undo/Redo menu items since 2023-09 (Eclipse 4.29.0)


pcdavid:
**Steps to reproduce**:

1. Open a plain text file.
2. Type some text.
3. Open the "Edit" menu.

**Expected**: the "Undo" entry is labeled "Undo Typing".

**Actual result**: the menu entry is simply labeled "Undo", with no reference to the action that will be undone (or redone, the same applies to the "Redo" menu entry).

Below: first screenshot on 2023-06, second on 2023-09.

![2023-06](https://github.com/eclipse-platform/eclipse.platform.ui/assets/10608/af306195-7142-4f6f-9002-541f7c385815)
![2023-09](https://github.com/eclipse-platform/eclipse.platform.ui/assets/10608/d815409b-0ef9-4ca4-afb3-a579f06021a0)

I first saw this when a lot of our tests (in Sirius) failed on 2023-12, but it is not specific to Sirius or even anything related to modeling. As indicated above, I reproduce it with a plain text file.

On Eclipse 2023-06, this works as expected. The regression appears in 2023-09 (and is present in the upcoming 2023-12).

pcdavid:
It seems to have been caused by https://github.com/eclipse-platform/eclipse.platform.ui/pull/978.
Reverting `WorkbenchActionBuilder` to the version before that change fixes the issue for me (the action labels are back, e.g. "Undo Typing" and "Redo Typing").

BeckerWdf:
@N1k145: Can you pls. have a look at it and provide a fix for that regression?

N1k145:
@BeckerWdf @pcdavid 
I had a quick look and I'm able to reproduce this. 
I have the feeling that this is somewhat unsupported by the other implementation, but I will have a look.
The change I made was basically switching the menu entry to a mechanism supporting E4 so that E4 contributions can provide their own handlers.

merks:
This issue was raised in the EMF forum too:

https://www.eclipse.org/forums/index.php/t/1113985/

N1k145:
@pcdavid @BeckerWdf @merks 
I analyzed the issue:
The old implementation (pure E3) did take the name of the Action and had a listener when the Action changes that refreshed the UI controls.

For this, there is currently no support in E4, which is used by the new implementation. A Handler does not provide a name, only a command.
[...]

PR #1602, opened 2024-01-31
IHandler.getHandlerLabel should be @since 3.12

merks:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1314


N1k145 (approved):
@merks thank you, you were quicker than the Github Notification on the other PR
It seems that the version of the plugin was increased just in the commit before mine was merged and the not being up-to-date to master was the issue

PR #1619, opened 2024-02-01
Dialog Images: suppress removal warning

jukzi:
The field Dialog.DLG_IMG_ERROR has been deprecated since version 2023-12 and marked for removal


jukzi:
fixes 9
image

PR #1622, opened 2024-02-04
PDE: Restarting development eclipse does not relaunch

umairsair:
- In development mode, changing restart parameters is not allowed, hence exit code 24 is not supported. Previously Workbench.buildCommandLine(..) used to return null for development mode, hence exit code was not set by Workbench.restart(..). And IDEApplication.start(..) used to take care of restarting development mode eclipse with exit code 23.
- Now in [PR#1307](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1307), Workbench.buildCommandLine(..) does not check for development mode, hence Workbench.restart(..) is setting return code 24.

Fix:
====
Similar to Workbench.setRestartArguments(..), Workbench.restart(..) also checks for development mode and explicitly sets the exit code 23

Related to https://github.com/eclipse-equinox/p2/issues/166


merks:
Yes, that looks correct and then the 24 doesn't not surface as a return value.

PR #1625, opened 2024-02-05
Link documentation for the "Appearance" preference page

BeckerWdf:
There is documentation available for the "Appearance" preference page but it's help context was not added to this preference page. So if a user pressed the "?" icon on the preference dialog this specific documentation was not show.


sratz (approved):

mickaelistria (approved):
Good finding!

BeckerWdf:
When doing this I also found out that not all the preference pages use the same style for their titles. Will be fixed with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1771

Issue #1628, opened 2024-02-05
Jenkins ECJ plugin detects way too less warnings 


jukzi:
same as for jdt.ui  (https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1169):
Jenkins show only 423 issues (https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/314/eclipse/), while workspace shows many more:

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/3c75459a-0dfe-4237-8881-9c8a3c4f64f1)

akurtakov:
The parent pom still overrides the warnings https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/15b58fce6c38cd475a541e2cf41b5a274194cfd5/eclipse-platform-parent/pom.xml#L96 . Please try removing this line and see the results.

akurtakov:
Even https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/15b58fce6c38cd475a541e2cf41b5a274194cfd5/eclipse-platform-parent/pom.xml#L105 has to be removed so test bundles warnings are the same too.

PR #1631, opened 2024-02-05
set deprecation=error / analyze #1628

jukzi:
AbstractKeyFormatter has deprecation warnings in workspace but none shown in Jenkins

https://github.com/eclipse-platform/eclipse.platform.ui/issues/1628


jukzi:
@laeubi do you mean something like this?

akurtakov:
In order to achieve workspace and cli builds warning parity there should be no overrides as explained in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1628#issuecomment-1926807026 . Tweaking per bundle warnings should affect both after that.

laeubi:
> @laeubi do you mean something like this?

yes

jukzi:
such test seems not to help as error fails fast

laeubi:
> such test seems not to help as error fails fast

It does help as it shows you the issue is actually detected and your setting is taken into account.

PR #1633, opened 2024-02-05
Replace sleak tool with

vogella:
org.eclipse.swt.graphics.Resource.reportNonDisposed

Sleak was never re-written to be used in RCP 4 applications and org.eclipse.swt.graphics.Resource.reportNonDisposed is anyhow 10x better to find such leaks.


akurtakov:
Shouldn't we remove Sleak in this case?

iloveeclipse:
> Shouldn't we remove Sleak in this case?

It works (and was fixed to work fine) in SDK, also it is use model differs to SWT resource leaks as it does not depend on GC.

Issue #1427, opened 2023-12-18
Random failure of SmartImportTests.testConfigurationIgnoreNestedProjects


fedejeanne:
Failed on [this build](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/7223222480/job/19681830746) of #1404 
```java
org.eclipse.ui.tests.datatransfer.SmartImportTests.testConfigurationIgnoreNestedProjects -- Time elapsed: 0.689 s <<< FAILURE!
junit.framework.AssertionFailedError: Root project shouldn't have been configured
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.Assert.assertFalse(Assert.java:39)
	at junit.framework.TestCase.assertFalse(TestCase.java:208)
	at org.eclipse.ui.tests.datatransfer.SmartImportTests.testConfigurationIgnoreNestedProjects(SmartImportTests.java:273)
...
```

PR #1645, opened 2024-02-06
Unify Hover-Behavior Between Light and Dark Theme

BeckerWdf:
In the "Dark" theme the background color of unselected tabs in CTabFolders changes color when user hovers over them with the mouse cursor. This helps a lot to see where an inactive tab starts and ends.

This change adds the same behavior also for the light theme.

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/light/e4-light_tabstyle.css
+++ b/bundles/org.eclipse.ui.themes/css/light/e4-light_tabstyle.css
@@ -30,6 +30,7 @@
 	padding: 0px;
 	color: '#org-eclipse-ui-workbench-INACTIVE_TAB_TEXT_COLOR';
 	swt-draw-custom-tab-content-background: true;
+	swt-unselected-hot-tab-color-background:'#org-eclipse-ui-workbench-ACTIVE_TAB_BG_START';
 }
 
 .MPartStack.active {


BeckerWdf:
This is how the dark theme already looks like:
Screen Recording dark

This how this looks like on macOS before this change
Screen Recording before

This how this looks like on macOS after this change

Screen Recording after

BeckerWdf:
I think this also helps a lot for user that have the "Hide icon" setting on the "Appearance" preference page (#1071) turned on.

BeckerWdf:
@HeikoKlare and @thomasritter: What do you think?

HeikoKlare:
I agree that the proposed behavior is desirable. On Windows, the behavior is already as expected without this PR: hovering over unselected tab items (no matter if the tab folder has focus or not) highlights them. In the following screenshot, I hover over the "Declaration" tab. Note that the color is also lighter than the one selected in this PR:
image

Unfortunately, with this change, it does now work anymore on Windows. Behavior is then as follows: hovering over a tab item only highlights it if the tab folder does _not_ have focus. And in that case, the highlight color is notably darker than in the current implementation. Again, I hover over the "Declaration" tab:
image

So there seem to be different theme colors and/or different CSS interpretations at least in the Mac and Windows implementations.

BeckerWdf:
> I agree that the proposed behavior is desirable. On Windows, the behavior is already as expected without this PR: hovering over unselected tab items (no matter if the tab folder has focus or not) highlights them. In the following screenshot, I hover over the "Declaration" tab. Note that the color is also lighter than the one selected in this PR: image

Can you tell me what color definition is used here?

> Unfortunately, with this change, it does now work anymore on Windows. Behavior is then as follows: hovering over a tab item only highlights it if the tab folder does _not_ have focus. And in that case, the highlight color is notably darker than in the current implementation. Again, I hover over the "Declaration" tab: image
> 
> So there seem to be different theme colors and/or different CSS interpretations at least in the Mac and Windows implementations.

That's strange.
For windows light the following CSS file is used: e4_default_win.css. This includes /css/e4_basestyle.css, which in turn includes /css/light/e4-light_tabstyle.css.
For mac light the following CSS file is used: e4_default_mac.css. This includes /css/e4_basestyle.css, which is the same as for windows.

So the difference has to be in the CSS definitons inside e4_default_win.css. And in this file we have (besides some ColorDefinitions) only this:

```
.MPartStack {
	swt-simple: true;
}

CTabFolder.MArea {
	background-color: #F0F0F0;
	swt-selected-tab-fill: #F0F0F0;
	swt-unselected-tabs-color: #F0F0F0;
	swt-outer-keyline-color: #F0F0F0;
	swt-inner-keyline-color: #F0F0F0;
	swt-tab-outline: #F0F0F0;
}

CTabFolder Canvas {
	background-color: #F8F8F8;
}
```

Which of these properties causes the hover-behaviour on windows?

HeikoKlare:
I changed all of them, but none had any effect on the hover behavior. Probaby the hover behavior is simply the default behavior of the OS widget.

But I now found why your proposal "breaks" behavior on Windows: the highlight color you chose is exactly the same as the background color of a tab folder that has focus. You can see that in the following screenshot: in both cases, I hover over the "Declaration" tab. The tab folder at the top has focus, while the one at the bottom has not.

image

Using the background color for active rather than inactive tabs would fix that. That is actually the color that Windows already uses by default, so it would change nothing on Windows compared to existing behavior.

bundles/org.eclipse.ui.themes/css/light/e4-light_tabstyle.css
```diff
@@ -30,6 +30,7 @@
 	padding: 0px;
 	color: '#org-eclipse-ui-workbench-INACTIVE_TAB_TEXT_COLOR';
 	swt-draw-custom-tab-content-background: true;
+	swt-unselected-hot-tab-color-background:'#org-eclipse-ui-workbench-INACTIVE_TAB_BG_START';
```
HeikoKlare:
```suggestion
	swt-unselected-hot-tab-color-background:'#org-eclipse-ui-workbench-ACTIVE_TAB_BG_START';
```

BeckerWdf: Thanks for finding out. I took over your proposal. Can you pls. test again?

HeikoKlare:
I tested on Windows and Linux and both look good now 👍 That means: the patch does not change any behavior for these systems, as both Win32 and GTK seem to have the hover highlighting behavior as default.

Just one thing I found when taking another look at the CSS configurations: usually there are color keys "start" and "end", probably in case the theme defines some gradient? E.g. 
```
'#org-eclipse-ui-workbench-ACTIVE_TAB_BG_START' '#org-eclipse-ui-workbench-ACTIVE_TAB_BG_END' 100% 100%;
```
Maybe the added configuration should also consider that, even though gradients are rather uncommon now?

BeckerWdf:
> I tested on Windows and Linux and both look good now 👍 That means: the patch does not change any behavior for these systems, as both Win32 and GTK seem to have the hover highlighting behavior as default.
> 
> Just one thing I found when taking another look at the CSS configurations: usually there are color keys "start" and "end", probably in case the theme defines some gradient? E.g.
> 
> ```
> '#org-eclipse-ui-workbench-ACTIVE_TAB_BG_START' '#org-eclipse-ui-workbench-ACTIVE_TAB_BG_END' 100% 100%;
> ```
> 
> Maybe the added configuration should also consider that, even though gradients are rather uncommon now?

I also thought about the gradient and explicitly opted against it. It's uncommon in todays UIs. Should we merge this as is?

vogella:
Comment to the same start and end gradient colour: IIRC the CSS engine only supported gradients hence to remove it I had to set both to the same value. It would better if the css API supports a single colour as we could simplify our CSS files but I never found the time to investigate/ implement this.

HeikoKlare (approved):
From my side, this is ready to merge.

BeckerWdf:
Why is the CI build failing now?
`Remote call to JNLP4-connect connection from 10.40.66.231/10.40.66.231:56210`

HeikoKlare:
Build is failing because of 
`META-INF/MANIFEST.MF:0 The re-exported type org.eclipse.swt.ole.win32.OLE has been removed from org.eclipse.jface_3.32.0` etc.
These errors also occur in master builds: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/335/console

This is an SWT issue (I will inspect if that's covered by one of the issues processed today).

BeckerWdf:
> Build is failing because of `META-INF/MANIFEST.MF:0 The re-exported type org.eclipse.swt.ole.win32.OLE has been removed from org.eclipse.jface_3.32.0` etc. These errors also occur in master builds: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/335/console
> 
> This is an SWT issue (I will inspect if that's covered by one of the issues processed today).

Thanks for looking into it.

BeckerWdf:
> Comment to the same start and end gradient colour: IIRC the CSS engine only supported gradients hence to remove it I had to set both to the same value. It would better if the css API supports a single colour as we could simplify our CSS files but I never found the time to investigate/ implement this.

the property 'swt-unselected-hot-tab-color-background' doesn't support gradients but only a single color 
see:

```
	@Override
	protected void applyCSSProperty(Control control, String property,
			CSSValue value, String pseudo, CSSEngine engine) throws Exception {
		if (!(control instanceof CTabFolder) || value.getCssValueType() != CSSValue.CSS_PRIMITIVE_VALUE) {
			return;
		}

		if (UNSELECTED_HOT_TAB_COLOR_BACKGROUND.equals(property)) {
			Color newColor = (Color) engine.convert(value, Color.class, control.getDisplay());
			CTabFolderRenderer renderer = ((CTabFolder) control).getRenderer();
			if (renderer instanceof ICTabRendering) {
				((ICTabRendering) renderer).setUnselectedHotTabsColorBackground(newColor);
			}
		}
	}
```

iloveeclipse:
> Build is failing because of `META-INF/MANIFEST.MF:0 The re-exported type org.eclipse.swt.ole.win32.OLE has been removed from org.eclipse.jface_3.32.0` etc. These errors also occur in master builds: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/335/console
> 
> This is an SWT issue (I will inspect if that's covered by one of the issues processed today).

I've reported https://github.com/eclipse-platform/eclipse.platform.ui/issues/1654

BeckerWdf:
> > Build is failing because of `META-INF/MANIFEST.MF:0 The re-exported type org.eclipse.swt.ole.win32.OLE has been removed from org.eclipse.jface_3.32.0` etc. These errors also occur in master builds: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/335/console
> > This is an SWT issue (I will inspect if that's covered by one of the issues processed today).
> 
> I've reported #1654

Thanks.

BeckerWdf:
The two failing tests are already reported in:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1427

PR #1665, opened 2024-02-10
Allow user select default editor per content type

### LINKED ISSUE #1659: Extend editor specification with editor precedence based on content type bindings
## Issue

It seems, the Eclipse platform does not provide any way of specifying when to prefer which editor in case we have multiple editors bound to the same content type. For example, having two C++ editors registered for the C++ content type (e.g. content type ID org.eclipse.cdt.core.cxxSource) and trying to find the preferred (or "default") editor for a cpp file with `IDE.getEditorDescriptor(path, true, false)`, the Eclipse platform will just [select the first compatible editor that if finds](https://github.com/eclipse-platform/eclipse.platform.ui/blob/d7676d31c86c5ff586a9fa091b2cb81981ffd754/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/registry/EditorRegistry.java#L1341).

When specifying an editor in an editor extension, we can set the default attribute to true, but that value seems not to be considered when selecting editors by content type bindings.

image

This issue might be related to #1532.

## Suggestion

It would be nice if there were an extension point where we can specify an editor precedence based on the content type binding, e.g. prefer LSP-based C++ editor over the classical C++ editor from CDT or prefer the generic text editor over the default text editor. The extension point should provide default settings that can be changed via preferences based on the user needs.

Having this, the Eclipse Platform's editor choice based on the content type should consider the specified / configured editor precedence. I.e. in case of multiple editors registered for the same content type `IDE.getEditorDescriptor(path, true, false)` would not just return an arbitrary editor that matches the content type, but return the one preferred by the user (or plugin specifications).

iloveeclipse:
Extended "Content Types" preference page with possibility to set one of the editors configured for a content type to "default" editor.

The preference will be saved into `org.eclipse.ui.workbench` preference store.

The key is a combination of a prefix with content type id, the value is the editor id, for example to associate base XML content type with generic editor, following is saved:

`defaultEditorForContentType_org.eclipse.core.runtime.xml=org.eclipse.ui.genericeditor.GenericEditor`

By default nothing is changed for IDE, but products can set defaults using the product customization, e.g. for the example above it would be:

`org.eclipse.ui.workbench/defaultEditorForContentType_org.eclipse.core.runtime.xml=org.eclipse.ui.genericeditor.GenericEditor`

image

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1659

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
@@ -99,6 +99,9 @@ public interface IPreferenceConstants {
 	// Preference key for default editors
 	String DEFAULT_EDITORS_CACHE = "defaultEditorsCache"; //$NON-NLS-1$
 
+	// Preference key prefix for default editor associated with a content type
+	String DEFAULT_EDITOR_FOR_CONTENT_TYPE = "defaultEditorForContentType_"; //$NON-NLS-1$
+
 	// Tab width = tab height * scalar value
 	String EDITOR_TAB_WIDTH = "EDITOR_TAB_WIDTH"; //$NON-NLS-1$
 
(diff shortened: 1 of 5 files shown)


iloveeclipse:
@mickaelistria , @travkin79: if there are no objections, I would like to merge that now.

travkin79:
No objections from my side. Thank you very much, @iloveeclipse!

mickaelistria:
I have not looked at the implementation detail, but I think the feature and the implementation make a lot of sense to be added. So no objection from my side, all the contrary!

BeckerWdf:
@iloveeclipse: Can you pls. also extend the documentation for that preference page?

PR #1666, opened 2024-02-11
Fix ignored external folding reconcile strategies

ghentschke:
Since the change which comes with commit 14f3660 'add always default folding reconciler' is faulty, it should be removed. Reason:
hasExternalFoldingAnnotations is always false because containsExternalFoldingAnnotations() can  never return true, because FoldingAnnotation is a non public class which should not be used by external folding reconciler strategies, like the one provided by LSP4E.

Therefore the external folding strategy provided via extension point is not considered.


ghentschke:
@mickaelistria can you please take look.

mickaelistria:
Do you think it would be affordable to create a test case for this story?

ghentschke:
What you think should be tested (better)? There are already unit tests for Folding [here](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/FoldingTest.java) including tests for the default and custom folding reconciler.

mickaelistria:
@ghentschke OK, let's merge.

PR #1669, opened 2024-02-12
Improve Double Separator Handling

### LINKED ISSUE #1668: Double Separators can be created with empty DynamicMenuContributions

## Steps to reproduce

AFAIK, JFace has a logic to remove double Separators from Context Menus. But this only works if they are next to each other after the contributions are collected. When some of these contributions do not contribute anything, the double Separators are not removed.

I tried:
grafik
The Second Contribution in this screenshot creates a dynamic menu contribution and a separator before that. 
The dynamic contribution does not add any elements (This could be the case that in the current selection, none of the possible results are valid, depending on implementation)

I expected: That the double and end separators are removed

But got: 
There is a Separator at the end of the Context Menu that is not removed.
grafik

* OS & version: Windows 11 and Linux
* Eclipse IDE/Platform version (as shown in *Help > About*): latest 2024-03

## Community

N1k145:
Fixes #1668

The Double Separator Handling is improved to support cases where empty contributions where located between separators

Steps to reproduce:
Create a menu looking like this:
* Action
* Separator
* Dynamic Contribution (not contributing anything)
* Separator
* Action

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/action/MenuManager.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/action/MenuManager.java
@@ -740,28 +740,17 @@ public class MenuManager extends ContributionManager implements IMenuManager {
 	protected void update(boolean force, boolean recursive) {
 		if (isDirty() || force) {
 			if (menuExist()) {
-				// clean contains all active items without double separators
 				IContributionItem[] items = getItems();
+
+				// clean contains all active items
 				List<IContributionItem> clean = new ArrayList<>(items.length);
-				IContributionItem separator = null;
+
 				for (IContributionItem item : items) {
 					IContributionItem ci = item;
 					if (!isChildVisible(ci)) {
 						continue;
 					}
-					if (ci.isSeparator()) {
-						// delay creation until necessary
-						// (handles both adjacent separators, and separator at end)
-						separator = ci;
-					} else {
-						if (separator != null) {
-							if (clean.size() > 0) {
-								clean.add(separator);
-							}
-							separator = null;
-						}
-						clean.add(ci);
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/action/MenuManager.java
```diff
@@ -740,28 +740,17 @@ protected void doItemFill(IContributionItem ci, int index) {
 	protected void update(boolean force, boolean recursive) {
 		if (isDirty() || force) {
 			if (menuExist()) {
-				// clean contains all active items without double separators
 				IContributionItem[] items = getItems();
+
+				// clean contains all active items
 				List<IContributionItem> clean = new ArrayList<>(items.length);
```
BeckerWdf, line 746: as "cleaning up" the double separators is not done later: is "clean" still a good name?

N1k145: To be honest, I think "clean" was never a good name.
But after the change, the not visible items are still "cleaned" from the list

BeckerWdf:
failing test is already reported in https://github.com/eclipse-platform/eclipse.platform.ui/issues/751

BeckerWdf:
If nobody objects I plan to merge this tomorrow.

BeckerWdf:
@N1k145 for your PR

iloveeclipse:
Could you please add the test to the test suite? Otherwise it will not run.

N1k145:
> Could you please add the test to the test suite? Otherwise it will not run.

Thank you for noticing, I'm so used to JUnit where there are no test suites anymore.

I added it in the new PR.

PR #1676, opened 2024-02-13
Fix wrong String split regex when checking OS version

Phillipus:
The "." character should be escaped as `"\\."` otherwise we get an empty String array


sratz (approved):

Phillipus:
@sratz Thanks for review. Shall I commit or would you prefer to?

sratz:
Thanks!

Test failure on macOS is unrelated.

PR #1678, opened 2024-02-13
PartRenderingEngineTests: fix ThreadDeath

jukzi:
"The type ThreadDeath has been deprecated since version 20 and marked for removal"

### DIFF
--- a/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
+++ b/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
@@ -153,8 +153,11 @@ public class PartRenderingEngineTests {
 		try {
 			advisor.eventLoopException(ex);
 		} catch (Throwable t) {
-			if (t instanceof ThreadDeath td) {
-				throw td;
+			// The type ThreadDeath has been deprecated since version 20
+			// and marked for removal
+			if ("ThreadDeath".equals(t.getClass().getSimpleName())) {
+				// don't catch ThreadDeath by accident
+				throw t;
 			}
 			// couldn't handle the exception, print to console
 			t.printStackTrace();


tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
```diff
@@ -153,8 +153,11 @@ private void handle(Throwable ex, IEventLoopAdvisor advisor) {
 		try {
 			advisor.eventLoopException(ex);
 		} catch (Throwable t) {
-			if (t instanceof ThreadDeath td) {
-				throw td;
+			// The type ThreadDeath has been deprecated since version 20
+			// and marked for removal
+			if ("ThreadDeath".equals(t.getClass().getSimpleName())) {
```
HannesWell, line 158: Is there a reason, why you didn't use the fully qualified name `java.lang.ThreadDeath`?
Using that would be saver.

haubi: @jukzi Two further thoughts here - especially for when someone discovers this when looking for some template on what to do with that deprecation warning:
- This would break when the throwable actually _extends_ ThreadDeath.
- This would hide a compilation error when ThreadDeath really is being removed.

Instead of _silently_ hiding the warning, I would prefer a general discussion of what to use _instead_ of ThreadDeath:
Are you aware of such a discussion somewhere already, or maybe even it's outcome meanwhile, and can share the link, even here?

laeubi: > Instead of silently hiding the warning, I would prefer a general discussion of what to use instead of ThreadDeath

As this is a testclass and hopefully no one ever calls Thread.stop() somewhere (**deprecated since Java 1.2!**) I would simply remove that thing altogether!

haubi:
ACK - moved this question to https://github.com/eclipse-rap/org.eclipse.rap/issues/267.

tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
haubi: ACK - moved this question to https://github.com/eclipse-rap/org.eclipse.rap/issues/267.

jukzi: As long as tests can use Thread.stop() i would keep the code. I wonder how org.eclipse.jdt.debug.core.IJavaThread.stop(IJavaObject) works.

laeubi: I can't find any test really calling `Thread.stop()` and the javadoc clearly state that this is dangerous and can corrupt memory. So I would say even if tests do this we should not support such behavior.

`IJavaThread.stop(IJavaObject)` is documented to call `Thread.stop()` but I cant really find any usage of this as well, so probably we should prepare to remove that API as well.

PR #1680, opened 2024-02-14
Suppress warnings about deprecated Preferences in legacy Code

jukzi:


akurtakov:
Please don't submit for the reason described at https://github.com/eclipse-platform/eclipse.platform/pull/1204#issuecomment-1943498574

PR #1681, opened 2024-02-14
Suppress warnings about deprecated org.eclipse.ui.keys in legacy Code

jukzi:


akurtakov:
Please don't push for the reason described at https://github.com/eclipse-platform/eclipse.platform/pull/1204#issuecomment-1943498574

vogella:
Seems like a again and again upcoming topic if we should hide warnings or not. @jukzi you can send a request for decisioin to the PMC mailing list or start a committer vote on this one. 

Currently it is at least one committer against hidding more warnings, so this is a no-go until you got a different PMC decsion or committer vote.

jukzi:
at least for this commit this is hillarious: it is only the package org.eclipse.ui.internal.keys which imports deprecated org.eclipse.ui.keys and nothing else beside java.util

laeubi:
If code relies on deprecated functionality it should be itself be deprecated (what will make the deprecation warning go away as far as I know).

akurtakov:
Even if it's only java.util that way a new deprecation similar to https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Date.html#%3Cinit%3E(int,int,int) (which is most likely not used in the package at all but I haven't checked) will go unnoticed with this approach.

PR #1684, opened 2024-02-14
Fix NPE in Windows Defender autofix Powershell script on first exclusion

HannesWell:
If a user has no process-exclusion defined in its Windows Defender configuration yet, the expression
`(Get-MpPreference).ExclusionProcess` returns null in the Powershell script to perform the exclusion.
If a null reference is read, the variable is now initialized with an empty set.


HannesWell:
Verified that it is now working as expected.
Everyone that used yesterdays I-build and had run the exclusion check should update its installation and rerun the exclusion check via the preferences `General -> Startup and shutdown`.

PR #1687, opened 2024-02-15
deprecate org.eclipse.ui.keys for removal

jukzi:
(except IBindingService)

Those classes have been deprecated with Bug 82854 (memory issues), replacing Objects with primitives.They are only used internal in org.eclipse.ui.internal.keys and tests

https://bugs.eclipse.org/bugs/show_bug.cgi?id=82854

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/keys/CharacterKey.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/keys/CharacterKey.java
@@ -16,7 +16,6 @@ package org.eclipse.ui.keys;
 
 import java.util.SortedMap;
 import java.util.TreeMap;
-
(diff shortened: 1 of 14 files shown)


akurtakov:
Looks good for me. Please look for usages of these classes marked for removal in the whole aggregator and if there are any we should deprecate such API for removal too and if internal remove directly.

jukzi:
> whole aggregator

is there anything outside the platform workspace?

akurtakov:
> > whole aggregator
> 
> is there anything outside the platform workspace?

Most likely not. This is in order to be on the safe side and prevent some I-build breakage.

jukzi:
It does not cause any issue in the workspace. What exactly do you want me to check beside that?

akurtakov:
Potential usages in eclipse.platform, pde, jdt.ui, etc. repos. If you're confident there are not such please proceed and merge.

PR #1689, opened 2024-02-16
Keep theme coloring of cells in FocusCellOwnerDrawHighlighter

Christopher-Hermann:
In dark theme, the cell back ground color are colored in light colors when using a FocusCellOwnerDrawHighlighter. By setting the background color alpha to 0, the color is not even draw and the initial color is kept.

**Coloring on macOS before the fix:**
Bildschirmfoto 2024-02-16 um 10 20 47

**Coloring on macOS after the fix:**
Bildschirmfoto 2024-02-16 um 10 21 38

**Reproduce/Test**
To test the issue on your platform, import the project [focusCellDemo.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/14308999/focusCellDemo.zip) and execute command "Open focus cell demo" to open a dialog with a demo viewer.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java
@@ -17,6 +17,7 @@
 package org.eclipse.jface.viewers;
 
 import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.GC;
@@ -106,9 +107,13 @@ public class FocusCellOwnerDrawHighlighter extends FocusCellHighlighter {
 
 	private void removeSelectionInformation(Event event, ViewerCell cell) {
 		GC gc = event.gc;
-		gc.setAlpha(0);// Don't draw background to keep cell coloring of theme
+		if (Platform.WS_GTK.equals(Platform.getWS())) {
+			// On GTK, the line is highlighted even though the SELECTED flag is removed. To
+			// fix this issue, the background must be overwridden
+			gc.setBackground(cell.getViewerRow().getBackground(cell.getColumnIndex()));
+			gc.fillRectangle(cell.getBounds());
+		}
 		gc.setForeground(cell.getViewerRow().getForeground(cell.getColumnIndex()));
-		gc.fillRectangle(cell.getBounds());
 		event.detail &= ~SWT.SELECTED;
 	}
 
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java
@@ -17,7 +17,6 @@
 package org.eclipse.jface.viewers;
 
 import org.eclipse.core.runtime.Assert;
-import org.eclipse.core.runtime.Platform;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.GC;
@@ -107,7 +106,7 @@ public class FocusCellOwnerDrawHighlighter extends FocusCellHighlighter {
 
 	private void removeSelectionInformation(Event event, ViewerCell cell) {
 		GC gc = event.gc;
-		if (Platform.WS_GTK.equals(Platform.getWS())) {
+		if ("gtk".equals(SWT.getPlatform())) { //$NON-NLS-1$
 			// On GTK, the line is highlighted even though the SELECTED flag is removed. To
 			// fix this issue, the background must be overwridden
 			gc.setBackground(cell.getViewerRow().getBackground(cell.getColumnIndex()));
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java
@@ -106,7 +106,7 @@ public class FocusCellOwnerDrawHighlighter extends FocusCellHighlighter {
 
 	private void removeSelectionInformation(Event event, ViewerCell cell) {
 		GC gc = event.gc;
-		gc.setBackground(cell.getViewerRow().getBackground(cell.getColumnIndex()));
+		gc.setAlpha(0);// Don't draw background to keep cell coloring of theme
 		gc.setForeground(cell.getViewerRow().getForeground(cell.getColumnIndex()));
 		gc.fillRectangle(cell.getBounds());
 		event.detail &= ~SWT.SELECTED;
--- a/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
@@ -37,8 +37,7 @@ Export-Package: org.eclipse.jface,
 Require-Bundle: org.eclipse.swt;bundle-version="[3.111.0,4.0.0)";visibility:=reexport,
  org.eclipse.core.commands;bundle-version="[3.4.0,4.0.0)";visibility:=reexport,
  org.eclipse.equinox.common;bundle-version="[3.18.0,4.0.0)",
- org.eclipse.equinox.bidi;bundle-version="[0.10.0,2.0.0)";resolution:=optional,
- org.eclipse.core.runtime;bundle-version="[3.31.0,4.0.0)"
+ org.eclipse.equinox.bidi;bundle-version="[0.10.0,2.0.0)";resolution:=optional
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Import-Package: javax.xml.parsers,
  org.osgi.framework;version="[1.8.0,2.0.0)",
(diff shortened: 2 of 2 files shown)


BeckerWdf:
The JavaDoc of the class says:

> A concrete implementation of {@link FocusCellHighlighter} using by setting the control into owner draw mode and *highlighting* the currently selected cell.

Does your change now no longer highlight the current selected cell?
Where does the `getBackground(cell.getColumnIndex()) `call get it's color from? Maybe we *only* need to override a default color in the dark theme.

Christopher-Hermann:
> Does your change now no longer highlight the current selected cell? 

The cell is still highlighted, see the screenshot. The problem was that the rest of the row should be colored as if it wasn't selected. This was done by setting the background to `cell.getViewerRow().getBackground(cell.getColumnIndex())`, but somehow this returns the color of the light theme. I did not find out why this is the case. By setting the background alpha to 0, nothing is draw and the color is kept as it was never selected.

> Where does the getBackground(cell.getColumnIndex()) call get it's color from? Maybe we only need to override a default color in the dark theme.

Yes, good point. The problem is that the table rows are displayed in black and a very dark grey in alternating order. And as said, `cell.getViewerRow().getBackground(cell.getColumnIndex())` does somehow returns the wrong color. I did not find where the actuall cell color is set for dark mode. I will check again if I can find some more information.

Christopher-Hermann:
The problem is that the color for table row is never set, neither in the light mode nor in the dark mode. So the default color of the viewer is used, which is just a pure white:
```
Color defaultBackground () {
	return display.getWidgetColor (SWT.COLOR_LIST_BACKGROUND);
}
```

So if possible the best solution would be to set the color into the viewer row, but as said, the alternating color must be followed and I cannot find any coding where this is done. So if someone has an idea or a hint where I can find coding related to this it would be nice.

BeckerWdf:
> The problem is that the color for table row is never set, neither in the light mode nor in the dark mode. So the default color of the viewer is used, which is just a pure white:
> 
> ```
> Color defaultBackground () {
> 	return display.getWidgetColor (SWT.COLOR_LIST_BACKGROUND);
> }
> ```
> 
> So if possible the best solution would be to set the color into the viewer row, but as said, the alternating color must be followed and I cannot find any coding where this is done. So if someone has an idea or a hint where I can find coding related to this it would be nice.

Where is this alternating of the color done?

Christopher-Hermann:
> > The problem is that the color for table row is never set, neither in the light mode nor in the dark mode. So the default color of the viewer is used, which is just a pure white:
> > ```
> > Color defaultBackground () {
> > 	return display.getWidgetColor (SWT.COLOR_LIST_BACKGROUND);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So if possible the best solution would be to set the color into the viewer row, but as said, the alternating color must be followed and I cannot find any coding where this is done. So if someone has an idea or a hint where I can find coding related to this it would be nice.
> 
> Where is this alternating of the color done?

I'm not sure, I cannot find the coding for this one. I think this is a Mac only feature, at least on windows the alternating of the color is not done.

BeckerWdf:
> > > The problem is that the color for table row is never set, neither in the light mode nor in the dark mode. So the default color of the viewer is used, which is just a pure white:
> > > ```
> > > Color defaultBackground () {
> > > 	return display.getWidgetColor (SWT.COLOR_LIST_BACKGROUND);
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So if possible the best solution would be to set the color into the viewer row, but as said, the alternating color must be followed and I cannot find any coding where this is done. So if someone has an idea or a hint where I can find coding related to this it would be nice.
> > 
> > 
> > Where is this alternating of the color done?
> 
> I'm not sure, I cannot find the coding for this one. I think this is a Mac only feature, at least on windows the alternating of the color is not done.

Ah ok. So this may be in the native part of SWT.

BeckerWdf (approved):

BeckerWdf:
This issue also exists in the light theme on macOS but it's not so apparent there: For lines with light background it's correct and for grey lines the difference between grey and white can easily been overlooked.
I find the fix of making the cell transparent (so that the correct background shines through) instead of setting the background very elegant.
I propose to merge this once master is open for the next release.
@HeikoKlare: Can you test this on windows and linux? Just to be sure we don't break anything there?

HeikoKlare:
I have tested the change using the provided demo shell on Windows (Windows 11 Enterprise, Version 21H2, Build 22000.2777) without noticing any difference in behavior in highlighting behavior. Below, you can find two example screenshots with and without the patch.

Without the patch:
image

With the patch:
image

I'll also test on Linux and post the results.

HeikoKlare:
Here are my test results with the provided demo shell on Linux (Ubuntu 22.04.4 LTS, Gnome 42.9, X11). There is a noticable difference on Linux, and I think the existing implementation is actually motivated by the behavior on GTK: It seems like on GTK the whole row is highlighted and setting the individual background color for the cells ensures that for non-selected cells the highlighting is overwritten. You can see in the screenshots below, that this is a kind of hacky solution, as you can still see the row highlighted behind the cell, but at least the user can see which cell is selected. After applying this patch, it will become impossible to know which cell is selected.

Without the patch:
image

With the patch:
image

BeckerWdf:
> Here are my test results with the provided demo shell on Linux (Ubuntu 22.04.4 LTS, Gnome 42.9, X11). There is a noticable difference on Linux, and I think the existing implementation is actually motivated by the behavior on GTK: It seems like on GTK the whole row is highlighted and setting the individual background color for the cells ensures that for non-selected cells the highlighting is overwritten. You can see in the screenshots below, that this is a kind of hacky solution, as you can still see the row highlighted behind the cell, but at least the user can see which cell is selected. After applying this patch, it will become impossible to know which cell is selected.
> 
> Without the patch: image
> 
> With the patch: image

So should we limit @Christopher-Hermann patch to only set the transparency on macOS as on windows the problem doesn't exist and on GTK it's even harmfull?

HeikoKlare:
> So should we limit @Christopher-Hermann patch to only set the transparency on macOS as on windows the problem doesn't exist and on GTK it's even harmfull?

If there is no alternative, I think that would be the way to go. However, I would try to avoid OS-specific behavior whenever possible. Do we know why the original line is working on every OS except Mac? More precisely, do we know why the background color returned by the viewer row is "wrong" on Mac? Since it works on the other OS', the returned color does not seem to be non-themed in general.
https://github.com/eclipse-platform/eclipse.platform.ui/blob/ea878173798e7ed395a425db3e3ec8bfc93be46c/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java#L109

Christopher-Hermann:
`cell.getViewerRow().getBackground(cell.getColumnIndex())` is returning the background color of the viewer (on every OS). The background color is explicit set to unset on MacOS to keep the alternating coloring that is provided in MacOS (can also be seen in the files explorer on Mac): 
https://github.com/eclipse-platform/eclipse.platform.ui/blob/cb121fe0f1b0bb270d8cfbc266092df0fd1de743/bundles/org.eclipse.ui.themes/css/e4-dark_mac.css#L43-L49

By unsetting the color for Trees/Tables on MacOs, the default color is returned which is just white.

HeikoKlare:
Thanks for explanation, @Christopher-Hermann!

I currently don't see any generic, i.e., OS-independent, solution. My understanding (from some experiments I made) is as follows: On Windows and Mac, it is sufficient to remove the `SWT.SELECTED` flag from the event details (and to restore the text color, i.e., the foreground color). On GTK, however, the line is highlighted even though the `SWT.SELECTED` flag is removed, which is why the background has to be overwritten explicitly. This also results in weird appearance, as only the cell backgrounds are corrected but not the borders cell and row borders.

So in conclusion: the "special case" for me is GTK and not Mac, which is why I would not propose to apply the alpha-value solution for Mac. Instead I would restrict the application of the background color to GTK (or even better to ensure that the row is not highlighted on GTK at all, like it is done on the other OS').

BeckerWdf:
> Instead I would restrict the application of the background color to GTK (or even better to ensure that the row is not highlighted on GTK at all, like it is done on the other OS').

At @Christopher-Hermann is this something you think can do?

Christopher-Hermann:
Yes, done. Thanks for testing and the hint @HeikoKlare. The fix works fine on MacOS

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/FocusCellOwnerDrawHighlighter.java
```diff
@@ -106,9 +107,13 @@ private void markFocusedCell(Event event, ViewerCell cell) {
 
 	private void removeSelectionInformation(Event event, ViewerCell cell) {
 		GC gc = event.gc;
-		gc.setBackground(cell.getViewerRow().getBackground(cell.getColumnIndex()));
+		if (Platform.WS_GTK.equals(Platform.getWS())) {
```
HeikoKlare:
```suggestion
		if (SWT.getPlatform().equals("gtk")) {
```

HeikoKlare (commented):
Thanks for updating the PR! I've tested the current changes on all three platforms and they work as expected.

Just one minor proposal for improvement: instead of adding further platform dependencies, I suggest to use the abilities for GTK identification provided by SWT, as the dependency to SWT exists anyway.

HeikoKlare (approved):
Still works as expected. Thank you!
Ready to be merged from my side.

BeckerWdf:
> Still works as expected. Thank you! Ready to be merged from my side.

Same for me. I will merge once the validation-build looks good.

BeckerWdf:
Test failures are pre-existing and already reported in 
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1427
https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

Merging this.
Thanks @Christopher-Hermann for providing the fix.

Issue #1688, opened 2024-02-15
Dark theme: code completion selection does not have enough contrast -> align with other popular dark themes


thomasritter:
Yesterday, I had an interesting call with a user of our source based editor. The user uses Eclipse on a large monitor with the Eclipse dark theme. He showed me that the contrast of the code completion is not high enough for him. Therefore, he always:

a) Opens the code completion
b) Uses tab to highlight the selection. With the highlight the contrast is much better

Here the comparison between:

Selected via tab:
<img width="427" alt="image" src="https://github.com/eclipse-platform/eclipse.platform.ui/assets/41768/e2b08145-cc17-4193-a9e6-62452a0d5545">

Without tab selection:
<img width="381" alt="image" src="https://github.com/eclipse-platform/eclipse.platform.ui/assets/41768/e9acc405-06c3-450b-82fb-78186078d5b5">

Now, imagine the first selection on a large monitor with a high resolution. The contrast is simply not enough.

Here some examples how other IDEs render the selection without the need to press tab.

Intellij:
<img width="336" alt="Screenshot 2024-02-15 at 09 02 53" src="https://github.com/eclipse-platform/eclipse.platform.ui/assets/41768/bcdb509d-dcfb-4dee-b976-64da088ca439">

VS Code:
<img width="684" alt="Screenshot 2024-02-15 at 09 02 41" src="https://github.com/eclipse-platform/eclipse.platform.ui/assets/41768/cd297995-ec76-4326-9f0c-a948bb54c9ab">

Both use a light blue which provides more contrast.

I would propose that the dark theme in Eclipse gets adjusted, too. Having to use tab feels like a hack/workaround. What do you think? @BeckerWdf 

- [X ] I verified [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/) doesn't provide the feature.

HeikoKlare:
I agree that having better highlighting would be great. Note that on Windows the issue is even worse, as highlighting in dark theme is _always_ hardly visible: https://github.com/eclipse-platform/eclipse.platform.swt/issues/811

Also note that the GTK implementation already does what you expect. Highlighting of the selected element is always done in a well visible way (no matter whether using tab to have focus or not):
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/755472/582395ef-ffb7-4042-804a-258a2625a905)

[...]

PR #1690, opened 2024-02-16
Highlighting problem when using the dark theme on Windows eclipse-platform/eclipse.platform.swt#811

### LINKED ISSUE #811: Highlighting problem when using the dark theme on Windows.

Christopher-Hermann:
Highlighting problem when using the dark theme on Windows https://github.com/eclipse-platform/eclipse.platform.swt/issues/811

JFace viewer are using OS selection color to highlight the selected item. On some OS this is not accessible. In particular when the eclipse is used with dark theme and the OS is used with light theme, this can cause really bad UI/UX experience. With this change, the selection color can be changed via color preference in the settings of eclipse.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/811

**Example**
Selection color on Windows Server in dark theme looks really bad:
272533915-9b63234a-6644-4174-9f2a-42b610eb5d66

**Testing**
On start up of eclipse, the selection colors in column viewers (table and tree viewer) should look like before.
On Windows, the dark theme should predefine the selection colors to fix the bad coloring on some Windows platforms.

In the preferences (Preferences > General > Appearance > Colors and Fonts), there are 4 new properties for the coloring of the viewer selection:

- Basic > Selection foreground color for cell viewer
- Basic > Selection background color for cell viewer
- Basic > Selection foreground color for cell viewer without focus
- Basic > Selection background color for cell viewer without focus
Changing these colors should affect the coloring of the selection in viewers. The coloring should immediately be applied as soon as the apply button is confirmed. The reset button should reset the coloring to the OS specific highlight coloring, expect for Windows dark theme, where the predefined colors should be restored.

To test the in different kind of viewers, import project [viewerSelectionDemo.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/14309383/viewerSelectionDemo.zip) and execute the command "Open Viewer Selection Demo". There a bunch of different viewers with different selection colors are rendered to test the coloring for focused/not focused viewers.

**Open Questions/Issues**
This will have impact on coloring for all eclipse installations on all OSs and themes. The default coloring of the OS is overwritten in any case, even if it is not needed. Maybe we should only overwrite the selection color if this is enabled/requested by the user via a preference setting?

### DIFF
--- a/tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.jface.text.tests
-Bundle-Version: 3.13.500.qualifier
+Bundle-Version: 3.13.600.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: 
--- a/tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
@@ -22,7 +22,8 @@ Require-Bundle:
  org.eclipse.text.tests;bundle-version="[3.5.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.ui.workbench.texteditor,
- org.eclipse.test.performance;bundle-version="3.13.0"
+ org.eclipse.test.performance;bundle-version="3.13.0",
+ org.mockito.mockito-core;bundle-version="5.12.0"
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Eclipse-BundleShape: dir
 Automatic-Module-Name: org.eclipse.jface.text.tests
--- a/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/TableOwnerDrawSupportTest.java
+++ b/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/TableOwnerDrawSupportTest.java
@@ -1,0 +1,93 @@
+package org.eclipse.jface.internal.text;
+
+import static org.mockito.Mockito.doReturn;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.RGB;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Table;
+import org.eclipse.swt.widgets.TableItem;
+
+import org.eclipse.jface.resource.ColorRegistry;
+import org.eclipse.jface.resource.JFaceResources;
+
+public class TableOwnerDrawSupportTest {
+
+	private static final int EVENT_Y= 2;
+
+	private static final int EVENT_HEIGHT= 3;
+
+	private Table table;
+
+	private Event event;
+
+	private RGB selectedRowBackgroundColor= new RGB(0, 0, 0);
+
+	private RGB selectedRowForegroundColor= new RGB(1, 1, 1);
+
+	private RGB selectedRowBackgroundColorNoFocus= new RGB(2, 2, 2);
+
+	private RGB selectedRowForegroundColorNoFocus= new RGB(3, 3, 3);
+
+	private GC gc;
+
+	@Before
+	public void setup() {
+		Shell shell= new Shell(Display.getDefault());
+		Table originalTable= new Table(shell, SWT.SINGLE);
+		//Spy in order to overwrite isFocusControl()
+		table= Mockito.spy(originalTable);
+
+		TableItem tableItem= Mockito.mock(TableItem.class);
+		Mockito.when(tableItem.getParent()).thenReturn(table);
+
+		gc= Mockito.mock(GC.class);
+		event= new Event();
+		event.type= SWT.PaintItem;
+		event.gc= gc;
+		event.detail= SWT.SELECTED;
+		event.item= tableItem;
+		event.y= EVENT_Y;
+		event.height= EVENT_HEIGHT;
+
+		Display.getDefault();
+		ColorRegistry colorRegistry= JFaceResources.getColorRegistry();
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_BACKGROUND", new RGB(0, 0, 0));
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_FOREGROUND", new RGB(1, 1, 1));
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_BACKGROUND_NO_FOCUS", new RGB(2, 2, 2));
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_FOREGROUND_NO_FOCUS", new RGB(3, 3, 3));
+
+		TableOwnerDrawSupport.install(table);
+	}
+
+	@Test
+	public void testPaintSelectionFocus() {
+		doReturn(true).when(table).isFocusControl();
+
+		table.notifyListeners(SWT.PaintItem, event);
+
+		Mockito.verify(gc).setBackground(new Color(selectedRowBackgroundColor));
+		Mockito.verify(gc).setForeground(new Color(selectedRowForegroundColor));
+		Mockito.verify(gc).fillRectangle(0, EVENT_Y, table.getBounds().width, EVENT_HEIGHT);
+	}
+
+	@Test
+	public void testPaintSelectionNoFocus() {
+		doReturn(false).when(table).isFocusControl();
+
+		table.notifyListeners(SWT.PaintItem, event);
+
+		Mockito.verify(gc).setBackground(new Color(selectedRowBackgroundColorNoFocus));
+		Mockito.verify(gc).setForeground(new Color(selectedRowForegroundColorNoFocus));
+		Mockito.verify(gc).fillRectangle(0, EVENT_Y, table.getBounds().width, EVENT_HEIGHT);
+	}
+
+}
--- a/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/contentassist/CompletionTableDrawSupportTest.java
+++ b/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/contentassist/CompletionTableDrawSupportTest.java
@@ -1,0 +1,92 @@
+package org.eclipse.jface.internal.text.contentassist;
+
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.collection.IsArrayWithSize.arrayWithSize;
+import static org.hamcrest.collection.IsArrayWithSize.emptyArray;
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.StyleRange;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.RGB;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Table;
+import org.eclipse.swt.widgets.TableItem;
+
+import org.eclipse.jface.internal.text.TableOwnerDrawSupport;
+import org.eclipse.jface.resource.ColorRegistry;
+import org.eclipse.jface.resource.JFaceResources;
+
+public class CompletionTableDrawSupportTest {
+
+	@Test
+	public void testInstall() {
+		Shell shell= new Shell();
+		Table table= new Table(shell, SWT.NONE);
+
+		assertThat(table.getListeners(SWT.MeasureItem), emptyArray());
+		assertThat(table.getListeners(SWT.EraseItem), emptyArray());
+		assertThat(table.getListeners(SWT.PaintItem), emptyArray());
+
+		CompletionTableDrawSupport.install(table);
+
+		assertThat(table.getListeners(SWT.MeasureItem), arrayWithSize(1));
+		assertThat(table.getListeners(SWT.MeasureItem)[0], instanceOf(TableOwnerDrawSupport.class));
+		assertThat(table.getListeners(SWT.EraseItem), arrayWithSize(1));
+		assertThat(table.getListeners(SWT.EraseItem)[0], instanceOf(TableOwnerDrawSupport.class));
+		assertThat(table.getListeners(SWT.PaintItem), arrayWithSize(1));
+		assertThat(table.getListeners(SWT.PaintItem)[0], instanceOf(TableOwnerDrawSupport.class));
+	}
+
+	@Test
+	public void testStoreStyleRanges() {
+		Shell shell= new Shell();
+		Table table= new Table(shell, SWT.NONE);
+		TableItem tableItem= new TableItem(table, SWT.NONE);
+		StyleRange[] ranges= new StyleRange[] {};
+
+		CompletionTableDrawSupport.storeStyleRanges(tableItem, 2, ranges);
+
+		assertEquals(ranges, tableItem.getData("styled_ranges2"));
+	}
+
+	@Test
+	public void testPaintNonFocusSelectionInFocusColors() {
+		int EVENT_Y= 2;
+		int EVENT_HEIGHT= 3;
+
+		Shell shell= new Shell();
+		Table table= new Table(shell, SWT.NONE);
+		TableItem tableItem= new TableItem(table, SWT.NONE);
+
+		ColorRegistry colorRegistry= JFaceResources.getColorRegistry();
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_BACKGROUND", new RGB(0, 0, 0));
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_FOREGROUND", new RGB(1, 1, 1));
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_BACKGROUND_NO_FOCUS", new RGB(2, 2, 2));
+		colorRegistry.put("org.eclipse.ui.workbench.SELECTED_CELL_FOREGROUND_NO_FOCUS", new RGB(3, 3, 3));
+
+		GC gc= Mockito.mock(GC.class);
+		Event event= new Event();
+		event.type= SWT.PaintItem;
+		event.gc= gc;
+		event.detail= SWT.SELECTED;
+		event.item= tableItem;
+		event.y= EVENT_Y;
+		event.height= EVENT_HEIGHT;
+
+		CompletionTableDrawSupport.install(table);
+
+		table.notifyListeners(SWT.PaintItem, event);
+
+		Mockito.verify(gc).setBackground(new Color(new RGB(0, 0, 0)));
+		Mockito.verify(gc).setForeground(new Color(new RGB(1, 1, 1)));
+		Mockito.verify(gc).fillRectangle(0, EVENT_Y, table.getBounds().width, EVENT_HEIGHT);
+	}
+
+}
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.text
-Bundle-Version: 3.25.100.qualifier
+Bundle-Version: 3.25.200.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -10,6 +10,7 @@ Export-Package:
  org.eclipse.jface.contentassist.images,
  org.eclipse.jface.internal.text;x-internal:=true,
  org.eclipse.jface.internal.text.codemining;x-internal:=true,
+ org.eclipse.jface.internal.text.contentassist;x-internal:=true,
  org.eclipse.jface.internal.text.html;x-friends:="org.eclipse.ant.ui, org.eclipse.jdt.ui, org.eclipse.ltk.ui.refactoring, org.eclipse.pde.ui, org.eclipse.ui.editors, org.eclipse.xtext.ui",
  org.eclipse.jface.internal.text.link.contentassist;x-internal:=true,
  org.eclipse.jface.internal.text.revisions;x-internal:=true,
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/TableOwnerDrawSupport.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/TableOwnerDrawSupport.java
@@ -25,6 +25,9 @@ import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Table;
 import org.eclipse.swt.widgets.TableItem;
 
+import org.eclipse.jface.resource.ColorRegistry;
+import org.eclipse.jface.resource.JFaceResources;
+
 
 /**
  * Adds owner draw support for tables.
@@ -42,6 +45,10 @@ public class TableOwnerDrawSupport implements Listener {
 
 	public static void install(Table table) {
 		TableOwnerDrawSupport listener= new TableOwnerDrawSupport(table);
+		installListener(table, listener);
+	}
+
+	protected static void installListener(Table table, Listener listener) {
 		table.addListener(SWT.Dispose, listener);
 		table.addListener(SWT.MeasureItem, listener);
 		table.addListener(SWT.EraseItem, listener);
@@ -70,7 +77,7 @@ public class TableOwnerDrawSupport implements Listener {
 		return (StyleRange[])item.getData(STYLED_RANGES_KEY + column);
 	}
 
-	private TableOwnerDrawSupport(Table table) {
+	protected TableOwnerDrawSupport(Table table) {
 		int orientation= table.getStyle() & (SWT.LEFT_TO_RIGHT | SWT.RIGHT_TO_LEFT);
 		fSharedLayout= new TextLayout(table.getDisplay());
 		fSharedLayout.setOrientation(orientation);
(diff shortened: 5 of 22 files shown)


fedejeanne:
@Christopher-Hermann thank you for looking into this! I personally use the dark theme and am sometimes annoyed that it doesn't highlight stuff properly (at least on Windows).

I tested the viewer (on Windows 10) and I see this:

highlighting_issue_dark_theme_TestViewer

I like it!

Would it be possible to somehow also fix the auto-completion dialog though?

highlighting_issue_dark_theme_autocompletion

Or maybe I am just testing this the wrong way?

Here are the GIFs in raw format: [gifs.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/15075456/gifs.zip)

Christopher-Hermann:
> Would it be possible to somehow also fix the auto-completion dialog though?

I just looked into this exact issue today, see #1688. I hope I can finish the work on this the next weeks.

Christopher-Hermann:
Not sure about the design of TableOwnerDrawSupport and the sub class CompletionTableDrawSupport. Maybe someone has an better idea how to overwrite the selection color in CompletionTableDrawSupport.

Christopher-Hermann:
I would appreciate some assistance on deciding how to proceed with this PR.

**Summary**
Four new color options are now available in the preferences to style table and tree selections. Users can even customize these from the preferences dialog for a more personalized user experience:
image

This offers a significant UI/UX enhancement, especially on Windows, where the default OS coloring resulted in a poor user experience:
image

With these new color preferences, table and tree viewer selection colors remain consistent across all platforms, and can be tailored to individual user preferences. However, changing the selection color for every Eclipse installation might result into problems on certain platforms.

For instance, screenshots taken from Lars Vogels' tutorials on Linux Ubuntu show the default selection color to be orange:
image

Moreover, MacOS users can specify custom selection colors in the OS (Akzentfarbe), which is currently recognized by Eclipse:
image

**Open Questions**
- Can we universally alter the default selection color across all operating systems?
- Would it be better to make the selection coloring optional, perhaps by including an additional setting within the preferences?
- Is it possible to consider OS selection coloring when initially launching Eclipse, while still providing users with the option to modify the coloring in the preferences?

fedejeanne:
> **Open Questions**
> 
>     * Can we universally alter the default selection color across all operating systems?
> 
>     * Would it be better to make the selection coloring optional, perhaps by including an additional setting within the preferences?
> 
>     * Is it possible to consider OS selection coloring when initially launching Eclipse, while still providing users with the option to modify the coloring in the preferences?

I don't know how the following could be achieved but my personal favorite would be:
* Provide a new button to let the 4 new configurations use the "system color" (similar to _Use System Font_)
* Use the system color by default: I would expect that Mac and Linux still look good when using the dark theme (also, the green "Akzentfarbe" should be used in Mac)
* (Optionally) Override the system color for Windows when using the dark theme (if the decision on this is blocking the PR, just postpone it to another PR)

IMO letting the users fix the coloring problem on their own by actively changing it would be good enough for this PR. 

My two cents on the issue :-)

HeikoKlare:
First of all, I really like and appreciate this contribution, as this dark theme highlighting problem on Windows is quite annoying. So thank you for working on that, @Christopher-Hermann!

I have to admit that I had no time yet to look into the proposed solution, so I can only give input in terms of "solution-independent expectations". Maybe they can help you in some way; otherwise feel free to ignore them:
* From a pure user experience point of view, I would expect that on Linux and Mac everything stays as it is and on Windows the dark theme by default uses the new colors (as there are no native colors to be reasonably used).
* From a more technical point of view, I would expect that by default the native colors are used and they can optionall be overwriting via preferences. On Windows, the theme could then overwrite these values by default. I am not sure whether that's possible with the preferences mechanism.

Christopher-Hermann:
> * From a more technical point of view, I would expect that by default the native colors are used and they can optionall be overwriting via preferences. On Windows, the theme could then overwrite these values by default. I am not sure whether that's possible with the preferences mechanism.

Yes, I also prefer this approach. I will have a look if something like this can be done within the preferences. I can imagine that taking the OS colors on the initial eclipse startup can also cause troubles when the OS color settings are changed in a later point in time. But I will have a look.

> * (Optionally) Override the system color for Windows when using the dark theme (if the decision on this is blocking the PR, just postpone it to another PR)

If the preference thing is not possible (or at least not with manageable effort) I will go for this approach to provide a quick fix at least for windows users. I think that's where the problem is most annoying. An then have a look on a separate PR.

Christopher-Hermann:
I achieved the desired behavior, at least for MacOS it's working fine:

- System highlight color is set as viewer selection color on the startup of eclipse.
- The color can be overwritten by the user via the color preferences.
- For dark theme on Windows, the system colors are overwritten by the theme settings.

It would be nice if someone can test the latest changes on other platforms. As soon as testing looks good I will have a look if unit tests can be added.

fedejeanne:
@Christopher-Hermann could you please describe how you tested and what is the expected behavior? 

Please explain which menus/preferences should be changed in the "Testing" section of this PR for future reference. I can test in Linux if it's ready in the next couple of hours. Thank you.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionTableDrawSupport.java
```diff
@@ -0,0 +1,40 @@
+package org.eclipse.jface.text.contentassist;
```
BeckerWdf: Copyright Header is missing

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,65 @@
+package org.eclipse.jface.viewers;
```
BeckerWdf: Copyright Header is missing

Christopher-Hermann:
> Please explain which menus/preferences should be changed in the "Testing" section of this PR for future reference. I can test in Linux if it's ready in the next couple of hours. Thank you.

You can use the ZIP file provided under Testing here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1690#issue-2138265188
There are a bunch of Viewers rendered where the selection coloring with and without the fix can be compared.

Furthermore, the new selection color should be applied in all standard eclipse views using column viewer, for example project explorer, outline, etc.

On Linux, the behavior should be:
- Every thing is colored as it was without the change (in light and dark theme)
- Going to Preferences > General > Appearance > Colors and Fonts and changing the color settings for the selection (Baisc > Selection background color for cell viewer, etc.)
- After changing the color settings, the viewers from the test ZIP and all Eclipse views should adapt the selection color accordingly

fedejeanne:
@Christopher-Hermann thank you for the details.

Changes look good in Linux and the colors can be changed. I also  briefly tested in Windows and it looks OK.

Please remember to:
* Edit the _Testing_ section of this PR
* Mention the 4 added configurations: where are they and what are their names?
* Mention the expected state of the buttons according to the platform (Windows overrides the defaults so the state of the "Restore" button is different than on the other 2 platforms)
* Mention a "quickly noticeable" change once you hit the "Apply" button (you should even see it in the preferences dialog).

The reason I insist on this is because it should be easy to see how to test and what to expect in which specific platform for newcomers. It would also give me and other reviewers an idea of how you tested and we could expand on that idea and test for stuff that you might have overlooked.

### My findings
I also found that, in Linux:
* The system values for "Selection background color for cell viewer without focus" (white) and "Selection foreground color for cell viewer without focus" (gray) don't work well together (look at the 2nd and 4th columns in the screenshot below)
* Some borders are still painted on the right columns when working with cell selection (look at the 4th and 5th column of the screenshot below)
 
Screenshot from 2024-05-07 11-05-22

Christopher-Hermann:
> * The system values for "Selection background color for cell viewer without focus" (white) and "Selection foreground color for cell viewer without focus" (gray) don't work well together (look at the 2nd and 4th columns in the screenshot below)

I noticed the same (not as bad as on Linux, but it could still be better) for MacOS. Maybe we need to define a color for the non focus selection. This shouldn't be affected by the OS highlight color anyway.

Christopher-Hermann:
> * Some borders are still painted on the right columns when working with cell selection (look at the 4th and 5th column of the screenshot below)

Not related to this change, I opened another issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1878. You can assign the issue to me, I will look into it later.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
BeckerWdf: Mh. I still don't see the copyright header.

Christopher-Hermann: Done

BeckerWdf (approved):

BeckerWdf:
Would be good to have some unit tests for the changes.
Also it would be nice if somebody on windows could test this PR.

tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
```diff
@@ -22,7 +23,10 @@ Require-Bundle:
  org.eclipse.text.tests;bundle-version="[3.5.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.ui.workbench.texteditor,
- org.eclipse.test.performance;bundle-version="3.13.0"
+ org.eclipse.test.performance;bundle-version="3.13.0",
+ org.mockito.mockito-core;bundle-version="5.11.0",
+ net.bytebuddy.byte-buddy;bundle-version="1.14.13",
+ net.bytebuddy.byte-buddy-agent;bundle-version="1.14.13"
```
fedejeanne: Do you really need these?

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/contentassist/CompletionTableDrawSupportTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Table;
+import org.eclipse.swt.widgets.TableItem;
+
+import org.eclipse.jface.internal.text.TableOwnerDrawSupport;
+import org.eclipse.jface.resource.ColorRegistry;
+import org.eclipse.jface.resource.JFaceResources;
+
+public class CompletionTableDrawSupportTest {
```
fedejeanne, line 26: Could you please rename the tests in this class and start them with `test...` ? I know it's not a rule but it makes it way easier for me to easily spot test methods in search results, call hierarchies, etc.

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/TableOwnerDrawSupportTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Table;
+import org.eclipse.swt.widgets.TableItem;
+
+import org.eclipse.jface.resource.ColorRegistry;
+import org.eclipse.jface.resource.JFaceResources;
+
+public class TableOwnerDrawSupportTest {
```
fedejeanne, line 22: Could you please rename the tests in this class and start them with `test...` ? I know it's not a rule but it makes it way easier for me to easily spot test methods in search results, call hierarchies, etc.

fedejeanne (changes requested):
I took a peek at the latest changes in b9ac80115739502ddb19e8e5cddeb98f089328a9 and I have some minor comments.

I didn't test the changes yet, I will try to come back to this later this week.

tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
Christopher-Hermann: Yes. byte-buddy is needed for Mockito, Mockito is needed to mock the static class GC.

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/TableOwnerDrawSupportTest.java
Christopher-Hermann: Done

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/contentassist/CompletionTableDrawSupportTest.java
Christopher-Hermann: Done

tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
fedejeanne: Hm, I thought it should be automatically added since I didn't see any other references to byte-buddy in the WS (but several references to Mockito).
The tests are failing (invalid access). Can you please look into that?

Christopher-Hermann:
For the failing unit test I will need some help. On my local machine I fixed the test by adding the packages as export package:
`org.eclipse.jface.internal.text.contentassist;x-internal:=true,` in bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
`org.eclipse.jface.internal.text.contentassist,` as export package in tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF

Not sure if something else needs to be done to fix this in the GitHub run

alexanderKononov-sap:
Thank you Christopher for taking on this task!
I have just tested it under Windows 11, and it works as expected.
- New preferences take effect immediately after pressing "Apply"
- Reset to default works
- In all views that I tested, the selection had new colors that I set
- Also in the code completion pop-over

The only thing I have noticed, is that in your test view, the 4th and 5th trees look the same. I assume that this is a copy-paste error in the test view, but if not, please check it out.
image

fedejeanne:
> For the failing unit test I will need some help. On my local machine I fixed the test by adding the packages as export package: `org.eclipse.jface.internal.text.contentassist;x-internal:=true,` in bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF `org.eclipse.jface.internal.text.contentassist,` as export package in tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
> 
> Not sure if something else needs to be done to fix this in the GitHub run

Maybe something similar to the problem we faced before, @HeikoKlare ? I mean the one fixed by https://github.com/eclipse-platform/eclipse.platform.swt/commit/43cf2953c7fe32ed4ef24c4c14c4cdee9c2fd1cd.

HeikoKlare:
The test is failing because the it tries to access a protected method (the constructor `TableOwnerDrawSupport(Table)`) without having access to it. The reason is that the test is placed in a test plug-in, which (in an OSGi environment) uses a different class loader than the to-be-tested plug-in, so that even though test class and tested class are placed in the same package they cannot access protected or package-visible methods of each other (even though one might assume that when considering ordinary Java visibility behavior).

If the to-be-tested functionality is not part of plug-in's public API but should still be tested in isolation (in terms of a proper unit test), you  need to use one of the options discussed in https://github.com/eclipse-platform/.github/discussions/203, namely either plug-in-internal tests (as done in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1203 for SWT) or test fragments.

Maybe in this case it could be easier to instantiate the `TableOwnerDrawSupport` via public API (i.e., via `TableOwnerDrawSupport::install(Table)`) instead of providing a more complicated test setup?

Christopher-Hermann:
> Maybe in this case it could be easier to instantiate the `TableOwnerDrawSupport` via public API (i.e., via `TableOwnerDrawSupport::install(Table)`) instead of providing a more complicated test setup?

Done, thanks for your help. 

There are still some build failure and a test failure. It seems to be related to:
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.ui.themes: Only qualifier changed for (org.eclipse.ui.themes/1.2.2400.v20240522-0941). Expected to have bigger x.y.z than what is available in baseline (1.2.2400.v20240213-1133)
I will check on that

BeckerWdf:
Let's wait with version number increases once the freeze period is over and development for the next releases starts.

HeikoKlare:
In addition to the errors due to required version increments and API tools failures, there also seem to be some compilation problems:
```
Error:  Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.8-SNAPSHOT:compile (default-compile) on project org.eclipse.jface.text.tests: Compilation failure: Compilation failure: 
Error:  /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/internal/text/contentassist/CompletionTableDrawSupportTest.java:[37] 
Error:  	CompletionTableDrawSupport.install(table);
Error:  	^^^^^^^^^^^^^^^^^^^^^^^^^^
Error:  Access restriction: The type 'CompletionTableDrawSupport' is not API (restriction on classpath entry '/home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.jface.text/target/org.eclipse.jface.text-3.25.100-SNAPSHOT.jar')
[...]
```

Maybe those are worth to fix before the next development cycle starts.

merks:
There are 16 commits in this PR.  That seems like 15 too many…

Christopher-Hermann:
> There are 16 commits in this PR. That seems like 15 too many…

It's called iterative development, should be no problem with the squash and merge function.
Since I'm not that experienced with the Eclipse build pipelines I need to trail and error...

merks:
Personally I prefer to see one commit that’s amended 

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-a-pull-request

At some point there will need to be one commit with a good commit description.

Christopher-Hermann:
I always try to avoid using amend and force push, at least when collaborating with others. But for this PR your are right, I think there are no others working on this PR

Christopher-Hermann:
But back to work, do you have any suggestion how to increase the bundle version?
org.eclipse.ui.themes 1.2.2400 -> 1.2.2500?
org.eclipse.jface 3.34.0 -> 3.35.0?
Not sure if there are any other versions to increase.

HeikoKlare:
> But back to work, do you have any suggestion how to increase the bundle version?
> org.eclipse.ui.themes 1.2.2400 -> 1.2.2500?
> org.eclipse.jface 3.34.0 -> 3.35.0?
> Not sure if there are any other versions to increase.

Taking a quick look at the changes, I don't see any additions to public API, so I would only expect micro version bumps for all affected bundles to be required:
* `org.eclipse.jface.text`
* `org.eclipse.ui.themes`
* `org.eclipse.ui.workbench`
* `org.eclipse.ui`
* `org.eclipse.jface`: this one already has a micro version bump for the new development cycle. But you already applied a minor version bump here, so I guess that was somehow necessary?

Some of the bumps might become obsolete rather soon once other PRs affecting those bundles are merged in the new development cycle.

BeckerWdf:
This branch has conflicts that must be resolved

iloveeclipse:
@Christopher-Hermann : **please** squash all commits to one and provide appropriate commit message.

BeckerWdf:
> @Christopher-Hermann : **please** squash all commits to one and provide appropriate commit message.

Wouldn't it make sense to keep the version bumps as separate commits?

iloveeclipse:
> Wouldn't it make sense to keep the version bumps as separate commits?

Yes, in case the change need to be reverted, it is very handy.

Christopher-Hermann:
MacOS build failures seems to be unrelated

BeckerWdf (approved):

BeckerWdf:
@fedejeanne are your comments addressed in the meantime?

Christopher-Hermann:
> @fedejeanne are your comments addressed in the meantime?

Only open point is the byte-buddy dependency. At least for executing the unit tests on the local machine, the dependency is needed.
I could give it a try to remove it and check if the build can execute the unit tests without the dependency, but it feels wrong to have tests that only work on the build machine.
I think the dependency is needed since the static GC class is mocked. Most probably no other test is mocking static classes, but I did not find a better way to unit test this.

fedejeanne (changes requested):
The build is failing, I see that the tests in Mac are not running "due to previous errors" (says the log).

Maybe this one:

```
Error:  [API ERROR] File MANIFEST.MF at line 7: The minor version should be incremented in version 1.5.400, since execution environments have been changed since version 1.5.400 (location: /Users/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.e4.ui.di/META-INF/MANIFEST.MF)
```

merks:
> The build is failing, I see that the tests in Mac are not running "due to previous errors" (says the log).
> 
> Maybe this one:
> 
> ```
> Error:  [API ERROR] File MANIFEST.MF at line 7: The minor version should be incremented in version 1.5.400, since execution environments have been changed since version 1.5.400 (location: /Users/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.e4.ui.di/META-INF/MANIFEST.MF)
> ```

This looks just like this error:

https://github.com/eclipse-pde/eclipse.pde/issues/1283

Hopefully not a regression, but I don't see this error in the workspace this morning. 

In the history we can see it was changed long ago:

image

So the error seems bogus.

BeckerWdf:
> Hopefully not a regression, but I don't see this error in the workspace this morning.

So should we simply rebase this PR and see if a new build is fine again?

tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
fedejeanne: This is odd: no other bundle explicitly imports the bytebuddy bundle (only the package) and yet they all work. Also, by looking at the dependencies of the bundle `org.mockito.mockito-core` I see:

For version `5.11.0`:
image

For version `5.12.0`:
image

Maybe the tests run locally with `5.11.0` and that extra dependency to `5.12.0` breaks stuff? 

Maybe you could investigate that further and try what happens if:
1. You explicitly set the version to `5.12.0`
2. You don't set a version (which one is used then?)
3. You set the minimum version to `2.13.0`, like other test bundles do (which one is used then?)

fedejeanne:
> > Hopefully not a regression, but I don't see this error in the workspace this morning.
> 
> So should we simply rebase this PR and see if a new build is fine again?

@BeckerWdf I just rebased, hopefully the error will be gone and the tests (in Mac) will run too

merks:
Yes, rebasing is a good idea.

tests/org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
Christopher-Hermann: I changed to 5.12 and removed the dependency to byte-buddy. On my local machine the unit tests doesn't work, but maybe it's caused by a odd local setup. Let's see if the build server can run the tests...

Christopher-Hermann: Seems to work, at least the linux build does work. Windows and Mac should be unrelated. Thanks for your help 👍

Christopher-Hermann:
Rebase is not working. But the error is not reliable reproducible, sometime the Windows, sometime the Mac and sometime the Linux build is failing.

BeckerWdf:
failing test is unrelated and documented in https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

BeckerWdf:
@fedejeanne: Are you ok with merging this PR?

fedejeanne (approved):
Thank you @Christopher-Hermann for this, it looks really good!
I'm sorry for the delay, I'm swamped. 

I took a final look into it and noticed that you exported some packages in `org.eclipse.jface.text.tests`, which I assume was necessary because of Mockito? In any case, I removed the exports and the tests ran just fine on my machine so I tooked the liberty of pushing my changes into your PR so you don't need to wait for me again.

All checks pass so this one is ready to go ✅ 

Thank you again and thank you @BeckerWdf for your input too!

jukzi:
please have a look at test failures: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1955

jukzi:
And may this relate too? https://github.com/eclipse-platform/eclipse.platform.swt/issues/1275

BeckerWdf:
> please have a look at test failures: #1955

Are these tests not executed on the PR's verification build?

jukzi:
> Are these tests not executed on the PR's verification build?

I don't know. Sometimes the tests behave just different in Integration Builds or on OSs/java versions not tested on Ci.

BeckerWdf:
> > Are these tests not executed on the PR's verification build?
> 
> I don't know. Sometimes the tests behave just different in Integration Builds or on OSs/java versions not tested on Ci.

So if that's the case that's really sad. Once should have confidence that everything is ok if the verification builds are green.

merks:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1955#issuecomment-2162243203 for why this is broken.

I.e., code should not assume any specific color key is registered if that plugin itself did not register that color key.  I'm not sure if JFace is supposed to work standalone too, but in that case, you can't ever assume some color is registered if it's done via a plugin.xml.

iloveeclipse:
This caused https://github.com/eclipse-platform/eclipse.platform.ui/issues/1956 regression too.

Phillipus:
> I'm not sure if JFace is supposed to work standalone too,

Yes, it is. See https://wiki.eclipse.org/JFace#Using_JFace_outside_the_Eclipse_platform

Acru:
Is this still being worked on?  Sad to see it was reverted due to side effects, rather than being improved.

I've been looking for a fix for this, where the default highlighting in my setup is such that it is nearly impossible to tell when something is highlighted, as it is nearly the same color as the background.

Edit: I'm also having the same problem with enabled toolbar buttons actually, eg. with the perspective toolbar I can't tell which perspective is selected, though this pull doesn't help with that...

Christopher-Hermann:
It's still on my TODOs list but unfortunately I don't really have time at the moment for this topic.
The regressions/side effects were only related to Linus OS and it's really complex for me to setup, test and fix this on Linux.

Acru:
That is too bad.

In the mean time I have been trying to figure out where the toolbar button toggle highlighting is actually being done, though it seems that the highlighting is somewhere separate from the toolitems in org.eclipse.swt.widgets?

Acru:
Unless I am missing something, it looks like toolbar item highlighting is also entirely provided by the OS's implementation and theme coloring, and I did see there is an outstanding issue for windows 11 as well, in that regard; https://github.com/eclipse-platform/eclipse.platform.swt/issues/217#issue-1278004939

edcache:
When will this option be available?

PR #1693, opened 2024-02-17
Windows Defender autofix preference adjustments

HannesWell:
This PR enhances the Windows Defneder autofix implemented in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1453 in two parts
- Make Windows Defender Autofix preference names more expressive
- Remove win.defender.startup.check.skip preference if disabled for all
Removing the preference node from the user-scope instead of setting the value to false (if it was true before) allows the preference service to consider the value of higher-level scopes like the default scope defined in a product.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -240,7 +240,11 @@ public class WindowsDefenderConfigurator implements EventHandler {
 
 	public static void savePreference(IScopeContext scope, String key, String value) throws CoreException {
 		IEclipsePreferences preferences = getPreference(scope);
-		preferences.put(key, value);
+		if (value != null) {
+			preferences.put(key, value);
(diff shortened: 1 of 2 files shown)


HannesWell:
@iloveeclipse, @akurtakov at this change a change requires PL approval, right?
I think it would be good to submit this now so that the enhanced preference constant can be used permanently by providers of custom products. What do you think?

HannesWell:
I want to add a bit more context, why I think this is important:
Before this change if a user once activated the setting `Skip exclusion check on startup for all new Eclipse-based installations` and then disables it again the value `false` is persisted in the user-scope. Because the `user-sope` has priority over the `default-scope` and the search stops at the first encounter of a key, even if one has defined `skip=true` in a product's preference customization the startup check and its pop-up show up for new or newly updated installations, even if it was disabled in the products preference installation.
In my opinion the startup check should only run if neither the user has disabled it nor it was disabled for the product.

HannesWell:
If @akurtakov or @iloveeclipse are not available, @merks can you have a look at this as PMC member?
If submitted, I think this should be submitted rather sooner than later.

merks (approved):
We should merge this sooner rather than later.

PR #1685, opened 2024-02-14
Suppress Windows Defender Autofix startup check in test-runtimes

HannesWell:
Only run the startup check if the running application is `org.eclipse.ui.ide.workbench`, which runs in a usual Eclipse IDE. Requiring this specific app-id prevents the start-up check and consequently a pop-up in test applications launched by PDE or Tycho for OSGi Junit tests or general console applications (that still have the `o.e.ui.workbench` plugin installed) on Windows.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1683

Issue #1683, opened 2024-02-14
[Win] Most tests in I20240213-1800 DNF waiting on "WindowsDefenderConfigurator"


iloveeclipse:
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1453.

Most of Windows test hang in https://download.eclipse.org/eclipse/downloads/drops4/I20240213-1800/testResults.php
See
- https://download.eclipse.org/eclipse/downloads/drops4/I20240213-1800/logs.php#console
- https://download.eclipse.org/eclipse/downloads/drops4/I20240213-1800/testresults/ep431I-unit-win32-java17_win32.win32.x86_64_17/org.eclipse.ant.tests.ui.testplugin.AntUITests.txt

**That must be fixed for M3.**

Thread dump:
```
DumpStackTracesTimer almost reached timeout '7200000'.
totalMemory:            268435456
freeMemory (before GC): 212480400
freeMemory (after GC):  219865896
Thread dump 0 at 2024-02-14 02:17:45 +0000:
java.lang.Exception: ThreadDump for thread "Bundle File Closer"
	at java.base@17.0.5/java.lang.Object.wait(Native Method)
	at java.base@17.0.5/java.lang.Object.wait(Object.java:338)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.getNextEvent(EventManager.java:396)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:337)
java.lang.Exception: ThreadDump for thread "Signal Dispatcher"
java.lang.Exception: ThreadDump for thread "Worker-7: Windows Defender Exclusion Check"
	at java.base@17.0.5/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base@17.0.5/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:252)
	at java.base@17.0.5/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:717)
	at java.base@17.0.5/java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1074)
	at java.base@17.0.5/java.util.concurrent.Semaphore.tryAcquire(Semaphore.java:415)
	at org.eclipse.ui.internal.PendingSyncExec.acquire(PendingSyncExec.java:39)
[...]

PR #1694, opened 2024-02-17
Support running Windows Defender Autofix startup check in custom apps

HannesWell:
Currently the startup check for the Windows Defender Autofix only runs for the `org.eclipse.ui.ide.workbench`.
This adds the possibility to permit the startup check for other applications by setting the preference
`windows.defender.startup.check.app` to the permitted appId.

This is a follow-up for https://github.com/eclipse-platform/eclipse.platform.ui/pull/1685 to further enhance the possible configuration of the Windows Defender autofix implemented in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1453.

@merks I guess Oomph uses a custom application to run the Installer, doesn't it?
With this you could add `org.eclipse.ui/windows.defender.startup.check.app=<Oomph-Installer-app-id>` to the Installer's `preferenceCustomization` properties file, if you want to enable the startup check for the installer itself.
In order to make it then run this would require that the `UIEvents.UILifeCycle.APP_STARTUP_COMPLETE` event is fired in the installer, but I don't know if that's the case or if the installer uses the E4 UI workbench as well or a completely custom UI.
In case of the latter you could simply instantiate the WindowsDefenderConfigurator class directly.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -76,21 +76,18 @@ import org.osgi.service.prefs.BackingStoreException;
 @EventTopics(UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
 public class WindowsDefenderConfigurator implements EventHandler {
 	private static final String PREFERENCE_EXCLUDED_INSTALLATION_PATH = "windows.defender.excluded.path"; //$NON-NLS-1$
+	private static final String PREFERENCE_STARTUP_CHECK_APP = "windows.defender.startup.check.app"; //$NON-NLS-1$
 	public static final String PREFERENCE_STARTUP_CHECK_SKIP = "windows.defender.startup.check.skip"; //$NON-NLS-1$
 	public static final boolean PREFERENCE_STARTUP_CHECK_SKIP_DEFAULT = false;
 
 	@Reference
-	private IPreferencesService preferences;
+	protected IPreferencesService preferences;
 
 	@Override
 	public void handleEvent(Event event) {
 		if (runStartupCheck()) {
 			Job job = Job.create(WorkbenchMessages.WindowsDefenderConfigurator_statusCheck, m -> {
 				SubMonitor monitor = SubMonitor.convert(m, 10);
-				if (preferences.getBoolean(PI_WORKBENCH, PREFERENCE_STARTUP_CHECK_SKIP,
-						PREFERENCE_STARTUP_CHECK_SKIP_DEFAULT, null)) {
-					return;
-				}
 				Optional<Path> installLocation = getInstallationLocation();
 				if (installLocation.isPresent()) {
 					String checkedPath = getPreference(ConfigurationScope.INSTANCE)
@@ -107,9 +104,15 @@ public class WindowsDefenderConfigurator implements EventHandler {
 		}
 	}
 
-	private static boolean runStartupCheck() {
-		if (Platform.isRunning() && Platform.OS.isWindows() && !Platform.inDevelopmentMode()) {
-			return "org.eclipse.ui.ide.workbench".equals(getRunningApplicationId()); //$NON-NLS-1$
+	protected boolean runStartupCheck() {
+		if (Platform.OS.isWindows() && !Platform.inDevelopmentMode()) {
+			if (preferences.getBoolean(PI_WORKBENCH, PREFERENCE_STARTUP_CHECK_SKIP,
+					PREFERENCE_STARTUP_CHECK_SKIP_DEFAULT, null)) {
+				return false;
+			}
+			String permittedApp = preferences.getString(PI_WORKBENCH, PREFERENCE_STARTUP_CHECK_APP,
+					"org.eclipse.ui.ide.workbench", null); //$NON-NLS-1$
+			return permittedApp.equals(getRunningApplicationId());
 		}
 		return false;
 	}
@@ -150,8 +153,8 @@ public class WindowsDefenderConfigurator implements EventHandler {
 		if (!isWindowsDefenderActive(monitor.split(1))) {
 			return Boolean.FALSE;
 		}
-
-		HandlingOption decision = askForDefenderHandlingDecision();
+		Display display = Display.getDefault();
+		HandlingOption decision = askForDefenderHandlingDecision(display);
 		if (decision != null) {
 			switch (decision) {
 			case EXECUTE_EXCLUSION -> {
@@ -160,9 +163,8 @@ public class WindowsDefenderConfigurator implements EventHandler {
 					savePreference(ConfigurationScope.INSTANCE, PREFERENCE_EXCLUDED_INSTALLATION_PATH,
 							installLocation.map(Path::toString).orElse("")); //$NON-NLS-1$
 				} catch (IOException e) {
-					PlatformUI.getWorkbench().getDisplay()
-							.syncExec(() -> MessageDialog.openError(null, "Exclusion failed", //$NON-NLS-1$
-									bindProductName(WorkbenchMessages.WindowsDefenderConfigurator_exclusionFailed)));
+					display.syncExec(() -> MessageDialog.openError(null, "Exclusion failed", //$NON-NLS-1$
+							bindProductName(WorkbenchMessages.WindowsDefenderConfigurator_exclusionFailed)));
 				}
 			}
 			case IGNORE_THIS_INSTALLATION -> savePreference(ConfigurationScope.INSTANCE, PREFERENCE_STARTUP_CHECK_SKIP,
@@ -172,12 +174,12 @@ public class WindowsDefenderConfigurator implements EventHandler {
 		return decision == HandlingOption.EXECUTE_EXCLUSION ? Boolean.TRUE : null;
 	}
 
-	private static HandlingOption askForDefenderHandlingDecision() {
+	private static HandlingOption askForDefenderHandlingDecision(Display display) {
 		String message = bindProductName(WorkbenchMessages.WindowsDefenderConfigurator_exclusionCheckMessage);
 
-		return PlatformUI.getWorkbench().getDisplay().syncCall(() -> {
+		return display.syncCall(() -> {
 			HandlingOption[] choice = new HandlingOption[] { null };
-			MessageDialog dialog = new MessageDialog(Display.getCurrent().getActiveShell(),
+			MessageDialog dialog = new MessageDialog(display.getActiveShell(),
 					WorkbenchMessages.WindowsDefenderConfigurator_statusCheck, null, message, MessageDialog.INFORMATION,
 					0, IDialogConstants.PROCEED_LABEL, IDialogConstants.CANCEL_LABEL) {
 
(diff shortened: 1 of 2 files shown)


merks:
Unfortunately code like this is already a show stopper because there is no workbench:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/7d3801bfb9caba51a3d751c771c39b5dfbd2a104/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java#L176

With all the private statics, there's really not much that can be specialized or reused here.

merks (approved):
Looks good.

HannesWell:
> Unfortunately code like this is already a show stopper because there is no workbench:

Thanks Ed for this hint, this is indeed a problem.
I have now added an distinction between the case if the workbench is running or not. I just used PlatformUI.getWorkbench().getDisplay() since that method says
```
Code should always ask the workbench for the display rather than rely on Display.getDefault().
```

But actually it should be save to always use `Display.getDefault()` since by default the start-up check only runs after app-startup has completed and then the UI is already up and running.
And when searching through the platform code I see a lot of hits for `Display.getDefault()`. So I will change that too.

I'm a bit reculant to make more code non static because then I would have to create a dummy instance in `runCheckEnforced()` just to call `runExclusionCheck()`. But if there are no other solutions this can of course be done.

I also now skipped the link if the PlatformUI workbench is not running, but I think this should be configured further.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -359,4 +364,8 @@ private static List<String> runProcess(List<String> command, IProgressMonitor mo
 		}
 	}
 
+	private static Display getDisplay() {
+		return PlatformUI.isWorkbenchRunning() ? PlatformUI.getWorkbench().getDisplay() : Display.getDefault();
```
laeubi: Why don't you let inject a `UISyncronze` if you are already using E4 here?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -173,7 +175,7 @@ private static Boolean runExclusionCheck(IProgressMonitor m, Optional<Path> inst
 	private static HandlingOption askForDefenderHandlingDecision() {
 		String message = bindProductName(WorkbenchMessages.WindowsDefenderConfigurator_exclusionCheckMessage);
 
-		return PlatformUI.getWorkbench().getDisplay().syncCall(() -> {
+		return getDisplay().syncCall(() -> {
```
laeubi: here you fetch the display, why not use it below for the dialog as well and instead fetch it again.
Beside that if you want the Shell, your probably should inject that as well...

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -76,11 +76,12 @@
 @EventTopics(UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
 public class WindowsDefenderConfigurator implements EventHandler {
```
laeubi, line 77: I would use E4 plugin here instead of OSGi as you can then inject everything you need and still get notified by events

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -202,6 +204,9 @@ protected Control createCustomArea(Composite parent) {
 							}).create(parent);
 					GridDataFactory.swtDefaults().indent(0, 5).applyTo(keepScanning);
 
+					if (!PlatformUI.isWorkbenchRunning()) {
```
laeubi, line 209: A custom app can still use the e3 compatibility later and then there will be a workbench running... so the comment is quite misleading.

laeubi:
@HannesWell have you considered making this just an own *bundle* instead of making this part of the workbench.ui?
Then it could use even a platform filter and you save you from checking that it is on windows at all, user can actively decide to include it in their products as well without having some preferences.

Also as you are using an UI event here, it seems more suitable to make this a plugin with `fragment.e4xmi` targeting any application so it could be used in pure E4 RCP apps as well.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
HannesWell: Right, corrected that comment.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
HannesWell: Good point about the double fetching, reused that.

HannesWell:
> @HannesWell have you considered making this just an own _bundle_ instead of making this part of the workbench.ui? Then it could use even a platform filter and you save you from checking that it is on windows at all, user can actively decide to include it in their products as well without having some preferences.
> 
> Also as you are using an UI event here, it seems more suitable to make this a plugin with `fragment.e4xmi` targeting any application so it could be used in pure E4 RCP apps as well.

Yes I have considered making it an own bundle, but as explained in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1683#issuecomment-1945630875 I think this would be too much overhead for a single class and functionality and the savings would not justify the extra work.

And regarding using more E4 functionality: In the `WindowsDefenderConfigurator` only the default start-up check reacts to the E4 `APP_STARTUP_COMPLETE` event since that's the one fired, at least for an Eclipse IDE application. Otherwise it is not specifically tied to E4. Actually I tried to use as generic API as possible but for example because `o.e.ui.dialogs.PreferencesUtil` is needed this can't be move further up, without splitting this class. And at this time I don't want to change this that fundamentally anymore.
If there is a real need further larger enhancements need to be done in the next development cycle.

laeubi:
> And regarding using more E4 functionality: In the WindowsDefenderConfigurator only the default start-up check reacts to the E4 APP_STARTUP_COMPLETE event since that's the one fired, at least for an Eclipse IDE application. Otherwise it is not specifically tied to E4

Well E4 is actually the way to go, your are using a lot of legacy (E3) API here performing additional check that you don't need if you just go the E4 way of DI, and E4 works also for E3 compatibility apps but the other way not...

>  think this would be too much overhead for a single class and functionality and the savings would not justify the extra work

It doesn't matter how many classes are there, Bundles should be made so they represent different functionality, and this one seems very special and a lot of effort is taken to detect hings, disable things and so on so it just looks as one example where separation is actually useful instead of place it into some common bundle...

HannesWell:
> > And regarding using more E4 functionality: In the WindowsDefenderConfigurator only the default start-up check reacts to the E4 APP_STARTUP_COMPLETE event since that's the one fired, at least for an Eclipse IDE application. Otherwise it is not specifically tied to E4
> 
> Well E4 is actually the way to go, your are using a lot of legacy (E3) API here performing additional check that you don't need if you just go the E4 way of DI, and E4 works also for E3 compatibility apps but the other way not...

Can you please explain to me where exactly I use a lot of legacy API?
I'm currently only aware of `PlatformUI` and that's only used to prevent the creation of the link to the preferences, based on the assumption that without a workbench the preference UI is not available.
And since I use `PreferencesUtil`, which resides in `org.eclipse.ui.workbench` just like the `WindowsDefenderConfigurator` I don't see much potential to further reduce the dependencies, without splinting the configurator in an abstract and Eclipse workbench specific part.
But I would be curios to learn a more general way to open the preference page, if a preferences UI available.


> 
> > think this would be too much overhead for a single class and functionality and the savings would not justify the extra work
> 
> It doesn't matter how many classes are there, Bundles should be made so they represent different functionality, and this one seems very special and a lot of effort is taken to detect hings, disable things and so on so it just looks as one example where separation is actually useful instead of place it into some common bundle...

There is one medium sized method with checks, I would not consider this a lot of effort (even if it took two rounds to adjust that).
Anyways, as usual there are many different ways to solve a problem. Each has its pros and cons. But so late in the development cycle I don't want to change the path taken before. If you think it is important we can discuss that again for the next release cycle. But to be frank, I would have appreciated such suggestions during the initial implementation phase. PR https://github.com/eclipse-platform/eclipse.platform.ui/pull/1453 was open for over one and a half month and there was not a single comment on this topic.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -150,8 +153,8 @@ private static Boolean runExclusionCheck(IProgressMonitor m, Optional<Path> inst
 		if (!isWindowsDefenderActive(monitor.split(1))) {
 			return Boolean.FALSE;
 		}
-
-		HandlingOption decision = askForDefenderHandlingDecision();
+		Display display = Display.getDefault();
```
laeubi, line 156: Should this not use `getDisplay()`?

HannesWell: You mean the one from before?
I have have replaced that now with just this single call to `Display.getDefault()`. The UI is already up an running when this code is called so it seems save to just call `Display.getDefault()` and not depend on PlatformUI.

laeubi: The `Platform` is only the OSGi framework (Equinox), but you most probably want that the Workbench is running (`PlatformUI`), as mentioned before you are using an E4 event here so that should be safe, but why then check for a running platform at all?

HannesWell: You mean in `runStartupCheck()` the line ?
```
Platform.isRunning() && Platform.OS.isWindows() && !Platform.inDevelopmentMode()
```
I just added it as a extra safety net since it looks like that `Platform.OS.isWindows()` and `Platform.inDevelopmentMode()` will throw an exception if the platform is not running. Although I agree with you that it should always be running if that code is executed.

laeubi: Yes because if platform is not running there cant be no workbench, so no event.

On the other hand you assume that the Workbench is running and don't want to be safe there by not using `PlatformUI` ;-)
 if you want to be safe, you still need to use `UISynchronize` instead of Display ...

HannesWell: Removed that now along with another minor adjustment.
If someone calls the (internal) `WindowsDefenderConfigurator.handleEvent()` directly and the platform is not running (for whatever reason), one can just override the entire `runStartupCheck()` method.

HannesWell: > On the other hand you assume that the Workbench is running and don't want to be safe there by not using PlatformUI ;-)

Since I replaced the `getDisplay()` method by `Display.getDefault()` the only reference to `PlatformUI` is the call to `PlatformUI.isWorkbenchRunning()`, which is used to guard the creation of the link-text that can open the preference dialog. That should always be save.

PR #1697, opened 2024-02-19
fix ErrorDialog AUTOMATED_MODE logging

jukzi:
tests should not log IllegalStateException when opening ErrorDialog is expected.

https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/395


iloveeclipse (changes requested):
Exception was added on purpose, so one can see which tests are trying to open dialogs - because if they do so, it is in most cases an error and not intended.

So I don't think this PR should be merged.

PR #1701, opened 2024-02-19
Ensure 'WinDefend' service is running before querying MpComputerStatus

### LINKED ISSUE #1696: IOException in WindowsDefenderConfigurator on startup

I'm in the unfortunate situation where we use an anti-virus program with even worse performance problems than the Windows Defender. The WindowsDefenderAutofix which was merged recently doesn't seem to handle this tool very well.

Unless explicitly disabled in the preferences, starting the Eclipse now always also shows the following popup:
image

## Steps to reproduce

- Install Cortex XDR and use it instead of the Windows Defender
- Restart the Eclipse IDE

I tried:
Not much, given that I'm not able to configure Cortex in any capacity.

I expected:
The startup task to be skipped, as Windows Defender is inactive.

But got: 
The error dialog shown above.

## Here is some relevant log output

```
java.io.IOException: Process terminated with unexpected result:

	at org.eclipse.ui.internal.WindowsDefenderConfigurator.isWindowsDefenderActive(WindowsDefenderConfigurator.java:289)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.runExclusionCheck(WindowsDefenderConfigurator.java:153)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.lambda$0(WindowsDefenderConfigurator.java:100)
```

The problem seems to be the value of the `AMRunningMode` property. Given the configuration described above, I would expect it to be "Passive Mode", but instead it is an empty string. I assume that this is because the following criteria is not satisfied: 
[Passive mode is only available for devices that are onboarded to Microsoft Defender for Endpoint and that meet certain requirements.](https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-windows?view=o365-worldwide#use-powershell-to-check-the-status-of-microsoft-defender-antivirus)

Any thoughts on this?

* OS & version:
  - Windows 10 PRO
  - Version10.0.19045 Build 19045
* Eclipse IDE/Platform version : 
  -   eclipse.buildId=4.31.0.I20240218-1800
  -   java.version=21.0.1
  -   java.vendor=Eclipse Adoptium

## Community

HannesWell:
Follow-up on https://github.com/eclipse-platform/eclipse.platform.ui/pull/1453.
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1696

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -30,6 +30,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IProduct;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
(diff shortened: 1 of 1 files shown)


HannesWell:
@iloveeclipse or @ptziegler do you want to review this or try it out?
If I get no feedback, I'll submit this before midnight to have this available in tonight's I-build.

ptziegler:
I don't see anything odd, though I won't be able to actually test it until I'm back at work tomorrow morning.

HannesWell:
> I don't see anything odd, though I won't be able to actually test it until I'm back at work tomorrow morning.

No problem and thanks for having a look.
Then I will just continue as said above and submit it, once the build is green.

PR #1702, opened 2024-02-20
Throttler: fix initial timestamp

jukzi:
Unlucky value range of System.nanoTime() could have prevented throttled Runnable to run.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
@@ -14,6 +14,7 @@
  *******************************************************************************/
 package org.eclipse.jface.util;
 import java.time.Duration;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.swt.SWTException;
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
```diff
@@ -54,34 +61,40 @@ public Throttler(Display display, Duration minWaitTime, Runnable runnable) {
...
 		int minWaitBetweenRunMillis = (int) minWaitTime.toMillis();
-		Runnable runner = () -> {
+		Runnable runner = () -> { // Always runs in Display Thread
 			scheduled.set(false);
 			runnable.run();
-			lastRunNanos = System.nanoTime();
+			long nanoTime = System.nanoTime();
+			if (nanoTime == 0) {
+				// prevent further 0 values.
+				nanoTime = -1;
```
mickaelistria, line 70: When can this happen? In such case, shouldn't we fall back to using 1000 * System,currentTimeMillis()?

jukzi: It may happen with chance of ~ 1/Long.MAXVALUE - which is most unlikely. When called in that unfortunate instant using the previous nanoTime  (-1) may execute the next event just 1 nanosecond too early - which feels OK.

mickaelistria: ok

Issue #1695, opened 2024-02-19
remove 500ms PostSelectionDelay


jukzi:
currently org.eclipse.jface.util.OpenStrategy.TIME is set to 500ms which has the following effect:
After selecting something (mouse or keeyboard) in the editor there will be a 500ms delay before the UI is update.. for example 
-  the link in the Outline view, 
-  Status Bar entry for warnings,
-  highlighting the selected word, 
- ...? 
example:
![500msdelay](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/0875bd97-6b54-406b-b8d4-bda2c0811dd2)

without delay all those features are almost instant:
![no_delay](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/e922f9f8-abf1-487e-a4c3-71c15a7710ce)

I don't know of any benefit of the delay. Anybody does? I thought it might help to avoid too much events when let the cursor be pressed for scrolling. But those events are ignored anyway until key is released. Only if i use mouse to multiselect multiple lines there are many events, but i don't experience any lag due to it. It looks much more responsive when the outline view is updated instantaniously.
![mouseMultiselect](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/ac5fae4f-e735-49c2-84a4-db4d2e71e5b3)

how i stumpled across that delay (i always thought the UI is just so slow that it need so much time to update): There currently is an error when you select a type and press ctrl+t within <500ms. The type hierarchy will use the statusline for progressreport but after 500ms the text is overridden by Java Editors Problem hint. (until now i never noticed the problem hint in the status bar, probably because it has that 500ms lag)

mickaelistria:
> I don't know of any benefit of the delay. Anybody does? I thought it might help to avoid too much events when let the cursor be pressed for scrolling.

I also suspect this delay is meant to not trigger action while user is still actively interacting. Users repeatedly pressing a key is a possible and frequent case I believe.
Does the "blame" tell more about when/why this delay was introduce?

> But those events are ignored anyway until key is released

Is this true on all OSs?

jukzi:
> Does the "blame" tell more about when/why this delay was introduce?

no. its from initial.

> Is this true on all OSs?

i tested on win only.

vogella:
CPUs were slower in the old days. IIRC I added a similar delay in the p2 installation dialog as it was constantly freezing if you typed. 

[...]

PR #1703, opened 2024-02-20
TextViewer: use Throttler for firePostSelectionChanged #1695

jukzi:
Avoids a 500ms delay after changing the selection, like higlighting the
selected word.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/1695

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/multipageeditor/MultiVariablePageTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/multipageeditor/MultiVariablePageTest.java
@@ -150,7 +150,7 @@ public class MultiVariablePageTest extends UITestCase {
 		IPostSelectionProvider postProvider = (IPostSelectionProvider) sp;
 
 		fPostCalled = 0;
-		ISelectionChangedListener listener = event -> ++fPostCalled;
+		ISelectionChangedListener listener = event -> fPostCalled += sp != event.getSelectionProvider() ? 1 : 0;
 
 		try {
 			postProvider.addPostSelectionChangedListener(listener);
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.text
-Bundle-Version: 3.25.0.qualifier
+Bundle-Version: 3.25.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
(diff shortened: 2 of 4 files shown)


merks (commented):
I'm curious in general here if there needs to be a check that the widget is not disposed?  Maybe the `getDisplay() != null)` check was implicitly doing that in the old logic?

vogella:
Thanks @jukzi. Please extend the commit message, "immediately calls firePostSelectionChanged by chance." is not very telling. The project guidelines are that the commit message should describe the change and not the ticket.

jukzi:
> I'm curious in general here if there needs to be a check that the widget is not disposed? Maybe the `getDisplay() != null)` check was implicitly doing that in the old logic?

done - good catch.

jukzi:
> Please extend the commit message

done. ok? Otherwise please make a suggestion.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/multipageeditor/MultiVariablePageTest.java
```diff
@@ -155,9 +155,9 @@ public void testPostSelection() throws Throwable {
 		try {
 			postProvider.addPostSelectionChangedListener(listener);
 			editor.setPage(1);
-			assertEquals(1, fPostCalled);
-			editor.setPage(0);
```
mickaelistria: Is removing that necessary for the test to pass? If a test is broken, then isn't it a regression?

jukzi: i am not removing it i updated the expectation form 1 to 2, since we now get an immediate event.

mickaelistria: I think what's in the test is what is expected (1 event for each `setPage`), what's is the new test seems bogus: 1 `setPage` => 2 events ?

jukzi: there are two totally different senders for the both events. 1 for a page change and 1 for the included selection change which  used to come later

mickaelistria: ah ok, the test doesn't make it clear. Then can you please add a note about it in new comments inside the tests. Someone seeing this new test would easily be confused without such explanation.

jukzi: i now filter for the  SelectionProvider that is tested

mickaelistria: Perfect, thanks!

mickaelistria (approved):
Let's merge it as soon as CI pass so we get it in early in 4.32 to adapt if necessary.

(related) We should also consider, in a separate PR, changing the value of OpenStrategy.getPostSelectionDelay() to some duration that better matches "a bit more than the typical duration between 2 arrow keystorkes".

PR #1706, opened 2024-02-20
fix lost StatusLine message

jukzi:
StatusLine message is used both as ProgressMonitor and to show problems.

When a problem message is reset while ProgressMonitor is running the progress message was lost.
When a ProgressMonitor finished while a problem was shown the problem message was lost;

Happened for example while opening Quick-Type Hierarchy within <500ms after selecting a line that had a Problem marker.


jukzi:
Example with fix: (warning is preserved):
StatusLine text

PR #1707, opened 2024-02-21
LogView: Option to activate on Error only

jukzi:
For the case that someone is not interested in A LogView popping up for info items.


jukzi:
before:
image
after:
image

bundles/org.eclipse.ui.views.log/src/org/eclipse/ui/internal/views/log/messages.properties
```diff
@@ -47,7 +47,9 @@ LogView_confirmOverwrite_message = File " {0}" exists. Would you like to overwri
 LogView_confirmDelete_deleteButton = &Delete All Events
 LogView_operation_importing = Importing log...
 LogView_operation_reloading = Reloading...
-LogView_activate = &Activate on new events
+LogView_activate = &Activate on any new event
+LogView_activateWarn = Activate on new warnings
+LogView_activateError = Activate on new error only
```
HeikoKlare: I like the proposal.
Instead of making the user assume that probably "activate on new warnings" also includes activation on error, I would make that explicit.
Nitpicky: it should at least be "warning" instead of "warnings" to be consistent with the the other entries.
```suggestion
LogView_activateWarn = Activate on new warning or error
LogView_activateError = Activate on new error
```

jukzi: ok

merks:
I too like the idea!

bundles/org.eclipse.ui.views.log/src/org/eclipse/ui/internal/views/log/messages.properties
BeckerWdf: > I like the proposal. Instead of making the user assume that probably "activate on new warnings" also includes activation on error, I would make that explicit. Nitpicky: it should at least be "warning" instead of "warnings" to be consistent with the the other entries.

This also fits more to the filter options of the problems view:
image
where it's also made explicit.

BeckerWdf (approved):

PR #1708, opened 2024-02-21
[memory] optimize e4 ClassUtils.getSimpleName()

jukzi:
to return always the same string instance
Saves all duplicate Strings in
org.eclipse.e4.ui.css.swt.dom.WidgetElement.localName

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/utils/ClassUtils.java
+++ b/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/utils/ClassUtils.java
@@ -13,22 +13,31 @@
  *******************************************************************************/
 package org.eclipse.e4.ui.css.core.utils;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
 /**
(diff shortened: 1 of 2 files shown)


jukzi:
Duplicates Strings for example while showing quick type hierarchy:
image

bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/utils/ClassUtils.java
```diff
@@ -22,13 +22,7 @@ public class ClassUtils {
...
 	 */
 	public static String getSimpleName(Class<?> c) {
-		String name = c.getName();
-		int index = name.lastIndexOf('.');
-		if (index > 0) {
-			name = name.substring(index + 1, name.length());
-		}
-
-		return name.replace('$', '-');
+		return c.getSimpleName();
```
laeubi: will this really replace `$` with `-` as before?

jukzi: good point i will fix that

PR #1710, opened 2024-02-21
Handle enabled tamper-protection in Win-Defender Autofix startup check

### LINKED ISSUE #1709: Windows Defender exclusion cannot be added if Tamper Protection is active on the machine (e.g. InTune managed coroporate device)
The Windows Defender auto fix implemented with https://github.com/eclipse-platform/eclipse.platform.ui/pull/1453 does not work on devices where exclusion rules are tamper-protected.

The PowerShell script to add exclusions will run successfully, but the entry will not actually be added, causing the popup to re-appear next time Eclipse is started.

See https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/manage-tamper-protection-intune?view=o365-worldwide#tamper-protection-for-antivirus-exclusions

The Windows Defender auto fix implementation should check whether exclusions are tamper-protected before showing the popup.

Microsoft describes a method to check this:

https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/manage-tamper-protection-intune?view=o365-worldwide#how-to-determine-whether-antivirus-exclusions-are-tamper-protected-on-a-windows-device

I.e., the auto fix should not be proposed when `TPExclusions` in registry key
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Defender\Features` has value `1`.

HannesWell:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1709

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -150,7 +150,7 @@ public class WindowsDefenderConfigurator implements EventHandler {
 	 *         Windows Defender is inactive and null if the process was aborted.
 	 */
 	private static Boolean runExclusionCheck(IProgressMonitor m, Optional<Path> installLocation) throws CoreException {
-		SubMonitor monitor = SubMonitor.convert(m, 4);
+		SubMonitor monitor = SubMonitor.convert(m, 5);
 		if (!isWindowsDefenderServiceRunning(monitor.split(1)) || !isWindowsDefenderActive(monitor.split(1))) {
 			return Boolean.FALSE;
 		}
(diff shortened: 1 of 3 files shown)


BeckerWdf (approved):

HannesWell:
If one selects to exclude the current installation and tamper protection is enabled the following dialog pops up after one has hit `Proceed` on the 'main' dialog:

grafik

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -159,6 +159,12 @@ private static Boolean runExclusionCheck(IProgressMonitor m, Optional<Path> inst
 		if (decision != null) {
 			switch (decision) {
 			case EXECUTE_EXCLUSION -> {
+				if (isExclusionTamperProtectionEnabled(monitor.split(1))) {
+					display.syncExec(() -> MessageDialog.openError(null, "Exclusion failed", //$NON-NLS-1$
+							bindProductName(WorkbenchMessages.WindowsDefenderConfigurator_exclusionFailed_Protected)));
+					savePreference(ConfigurationScope.INSTANCE, PREFERENCE_STARTUP_CHECK_SKIP, "true"); //$NON-NLS-1$
+					return null; // Consider selection as 'aborted' and don't show the dialog again on startup
```
HannesWell, line 166: At this place I think there is not a real difference if it is persisted that this installation is excluded or should be ignored. In both cases the startup-check does show show up again and I think that's what we want here.
The same applies for the return type. The only caller handling the return type is only interested about the case when false is returned,which means Win-Defender is inactive to show a corresponding result dialog. But in this case we already have the mentioned extra handling so nothing else should be shown.

merks: Please try to get this merged today.  +1 PMC approval.

Issue #1570, opened 2024-01-26
Too many open Editors - can't read name


jukzi:
open all files in org.eclipse.jdt.internal.compiler.ast
this is how it used to look:
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/0f1dc09e-2c6c-401a-9556-12f6f5af43fb)

this is how it looks now:
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/f73c5b86-1dd4-490b-96fd-4b3210c9f799)

jukzi:
eclipse.buildId=4.31.0.I20240124-1800
java.version=21.0.2

HeikoKlare:
fyi: maybe #1071 is the reason, as also reported in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1071#issuecomment-1887786168

jukzi:
i also get a lot of exceptions in the log while doing it:
don't know if its related
```
org.eclipse.jface.text.BadLocationException
	at org.eclipse.jface.text.AbstractDocument.addPosition(AbstractDocument.java:341)
	at org.eclipse.core.internal.filebuffers.SynchronizableDocument.addPosition(SynchronizableDocument.java:212)
	at org.eclipse.jdt.internal.ui.javaeditor.SemanticHighlightingPresenter.updatePresentation(SemanticHighlightingPresenter.java:214)
	at org.eclipse.jdt.internal.ui.javaeditor.SemanticHighlightingPresenter.lambda$0(SemanticHighlightingPresenter.java:149)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4046)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3662)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:823)
	at org.eclipse.jface.window.Window.open(Window.java:799)
	at org.eclipse.ui.internal.views.log.EventDetailsDialog.open(EventDetailsDialog.java:198)
	at org.eclipse.ui.internal.views.log.EventDetailsDialogAction.run(EventDetailsDialogAction.java:102)
	at org.eclipse.ui.internal.views.log.LogView.lambda$2(LogView.java:599)
[...]

PR #1712, opened 2024-02-22
Improve 'EditorStack' detection in CTabRendering

praveen-skp:
Issue:  #1570.

The current CTabRendering check if a tab is part of editor area, by checking if the tab has the 'EditorStack' tag. If the tag is missing, then the tab is considered as part of view area.
This fix enhances the tab identification by including a check based upon the location, even when the 'EditorStack' tag is absent.

This is a proposal, since this implementation is not working because the injection does not work and so, the getElementLocation throws null pointer exception.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -27,7 +27,11 @@ import org.eclipse.core.runtime.preferences.IEclipsePreferences.PreferenceChange
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.e4.ui.internal.css.swt.ICTabRendering;
 import org.eclipse.e4.ui.internal.workbench.swt.AbstractPartRenderer;
+import org.eclipse.e4.ui.model.application.ui.MContext;
 import org.eclipse.e4.ui.model.application.ui.MUIElement;
+import org.eclipse.e4.ui.workbench.modeling.EModelService;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.CTabFolder;
 import org.eclipse.swt.custom.CTabFolderRenderer;
@@ -88,8 +92,6 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 
 	private static int MIN_VIEW_CHARS = 1;
 	private static int MAX_VIEW_CHARS = 9999;
-
-	private static final String EditorTag = "EditorStack"; //$NON-NLS-1$
 
 	// Constants for circle drawing
 	static enum CirclePart {
@@ -1330,7 +1332,15 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 
 	private boolean isPartOfEditorStack() {
 		MUIElement element = (MUIElement) parent.getData(AbstractPartRenderer.OWNING_ME);
-		return element != null && element.getTags().contains(EditorTag);
+		EObject root = EcoreUtil.getRootContainer((EObject) element, true);
+		if (root instanceof MContext context) {
+			EModelService eModelService = context.getContext().get(EModelService.class);
+			if (eModelService != null) {
+				int location = eModelService.getElementLocation(element);
+				return (location & EModelService.IN_SHARED_AREA) != 0;
+			}
+		}
+		return false;
 	}
 
 	private boolean getHideIconsForViewTabsPreference() {


praveen-skp:
Another approach is by adding a getModelService method in the StackRenderer and use the following code to get the locaiton.
int location = ((StackRenderer) element.getRenderer()).getModelService().getElementLocation(element);

iloveeclipse:
> This is a proposal, since this implementation is not working 

I've converted this PR to draft.

laeubi:
@praveen-skp model elements should implement `MContext#getContext()` where you can get an `IEclipseContext` you can use to lookup the required service.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -1330,7 +1335,9 @@ private IEclipsePreferences getSwtRendererPreferences() {
 
 	private boolean isPartOfEditorStack() {
 		MUIElement element = (MUIElement) parent.getData(AbstractPartRenderer.OWNING_ME);
-		return element != null && element.getTags().contains(EditorTag);
+		int location = eModelService.getElementLocation(element);
+		return (element != null && element.getTags().contains(EditorTag))
+				|| ((location & EModelService.IN_SHARED_AREA) != 0);
```
laeubi: I don't think this is correct, there can be many `Area`(s) in a RCP application and kust because they are shared don't make them the editor area.

merks:
You are trying to use the injected field during construction but it will be injected only after construction:

image

laeubi:
> You are trying to use the injected field during construction but it will be injected only after construction:

E4 also supports constructor injection (just in case), but I think the fix is invalid anyways.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
BeckerWdf: But have a look into StackRenderer#createWidget. There we have the same check:
```
		// Adjust the minimum chars based on the location
		int location = modelService.getElementLocation(element);
		if ((location & EModelService.IN_SHARED_AREA) != 0) {
			tabFolder.setMinimumCharacters(MIN_EDITOR_CHARS);
			tabFolder.setUnselectedCloseVisible(true);
		} else {
			tabFolder.setMinimumCharacters(MIN_VIEW_CHARS);
			tabFolder.setUnselectedCloseVisible(false);
		}
```

laeubi: It is obviously not the same check ;-)

But if we are sure we want to use the same the EditorTag check can completely be removed as it would be redundant then. Please note that usually there is only one editor `Area` but different part stacks.

So if this one is MISSING the Editor is just added to a ramdom Partstack, its good that this is happily the "right" area here... but this does not fix the missing EditorTag.

BeckerWdf: > It is obviously not the same check ;-)

Really?

`(location & EModelService.IN_SHARED_AREA) != 0)` and
`(location & EModelService.IN_SHARED_AREA) != 0)` look quite identical for me. Am I overlooking something?

merks:
I'm not sure the following observations are best made in the issue or here in this PR.  

When trying to reproduce the cause of the problem I already notice this problem, i.e., if you drag an editor to create a split editor area, the second area's tabs are squished:

image

If you now close all the editors and open new editors, those tabs are squished as well:

image

So it's highly likely that anyone who ever create such a split editor area will always see this problem and will them permanently see this problem...

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
laeubi: `(element != null && element.getTags().contains(EditorTag))`
is not the same as 
`(location & EModelService.IN_SHARED_AREA) != 0)`

so who ever did the tag check instead of the location check has did this probably for a purpose, now the EditorTag check even seem to become obsolete as the EditorArea is almost always shared (I can't think about a reason why one don't want this), also as I mentioned Eclipse uses the ID to determine it is "the editor stack" so why not consider this as well and so on.

merks: The following works for me:
```
	private boolean isPartOfEditorStack() {
		MUIElement element = (MUIElement) parent.getData(AbstractPartRenderer.OWNING_ME);
		EObject root = EcoreUtil.getRootContainer((EObject) element, true);
		if (root instanceof MContext context) {
			EModelService eModelService = context.getContext().get(EModelService.class);
			if (eModelService != null) {
				int location = eModelService.getElementLocation(element);
				return (location & EModelService.IN_SHARED_AREA) != 0;
			}
		}
		return false;
	}
```
EcoreUtil.getRootContainer is null safe.

It looks like this where otherwise all the tabs in both areas would be squished:

image

BeckerWdf: > The following works for me

So does this mean we have consensus that the PR with the changes proposed by @merks are good and would fix the issue (even when it does not fix the root cause of the disappearing tag)?

merks: Given that this can be demonstrated to fix the current problem, I _think_ the worst thing that could happen is that in some places the tabs might not be squished where you guys want them squished.  But my opinion is just one opinion...

HeikoKlare: I can confirm that the fix proposed by Ed solves the problem for me. And I agree that avoiding squished editor tabs outweighs the risk of some view tabs not being squished when they should be.

BeckerWdf:
I just updated the PR with @merks proposal.
Feel free to approve and merge once CI is green.

HeikoKlare (approved):
I must not merge during freeze period, but I guess @merks will do?

merks:
@iloveeclipse 

I just wanted to double check that you too are fine with merging this.  It has my approval on behalf of the PMC...

iloveeclipse:
> 
> I just wanted to double check that you too are fine with merging this. It has my approval on behalf of the PMC...

I'm not familiar with this code. Any chance to have a test case added?

merks:
If someone else wants to a test case, which I think is relatively challenging, they may do so, but I personally don't have the time for that.  I believe right now such time would likely be better invested in figuring out why the editor stack area tag is getting lost and while doing that, create a test case for that more general problem which has gone undetected until now...

In any case, I don't want that to hold up getting this into an I-Build.

BeckerWdf:
> I believe right now such time would likely be better invested in figuring out why the editor stack area tag is getting lost and while doing that, create a test case for that more general problem which has gone undetected until now...

I second that. Once we have fixed the root cause we can then also remove this workaround here again.
Thanks to @merks for approving and merging.

merks:
I still wonder about the case where the second editor area's tab are squished.  If makes me wonder if any such new editor area should have a tag.  But indeed, that is a separate issue that is significantly harder to address, though I see @HeikoKlare ❤️ has made some interesting progress as a starting point!

PR #1714, opened 2024-02-24
Activate and Fix FindReplaceLogicTest

Wittmaxi:
The FindReplaceLogic was not included in the TestSuite and thus not executed by CI. Additionally, I fixed a bug where the tests would not run correctly on linux and mac because windows-style line-breaks were assumed in the Strings.


Wittmaxi:
This PR does not touch any production code, it only enables and fixes a Unit test and thus increases our confidence in the production code.
The tests are Green - in particular the FindReplaceLogic tests:
grafik

Unless you have further concerns, @HeikoKlare, I am confident we are ready to merge this PR.

HeikoKlare (approved):
Thanks for activating and fixing the tests! Since only tests are affected, I think we can merge this even though we are in RC2 period.

vogella:
+1 for merge, tests are not affected by the release freeze

PR #1716, opened 2024-02-25
Fix calculation of minimum characters for view tab titles when 'Always show full titles' preference is enabled

### LINKED ISSUE #1715: Preference to 'always show full titles' for tabs in view areas is not correctly applied to newly added tab in existing view area
Bug for #1071 

#1071 added preferences to _always show full titles_ for tabs in view areas. This preference gets correctly applied to existing tabs in the view area and for newly added tabs having title length less than or equal to the maximum title length amongst existing tabs.

But if the newly added tab has title length greater than the maximum title length amongst existing tabs, the title/text of this new tab 
is truncated. Thus, the full title is not visible for this new tab.

## Steps to reproduce

1. Add enough views/tabs in a view area such that titles of tabs are not fully visible.
2. From appearance preferences, enable _Always show full titles_ preference.
       image
     -  Full titles for tabs will be shown and additional tabs will be put into a chevron.
           image
3. Add a new view, with title length, greater than the maximum title length amongst existing tabs
    - For example, added _Abstract Syntax Tree_ view.
         image

Expected: *Full title of the _Abstract Syntax Tree_ view should be visible*

But got: *Title of  _Abstract Syntax Tree_ view tab is truncated*

* OS & version: Windows 11
* Eclipse IDE : Version 2024-03 (4.31), Build I20240223-1800

shubhamWaghmare-sap:
Fixes #1715 

- Sets the minimum characters to be Max( Longest tab text length, MAX_VIEW_CHARS), where MAX_VIEW_CHARS = 9999.
- In most cases, the minimum characters would be set to MAX_VIEW_CHARS
- In rare cases, there could be a view tab with text/title length greater than MAX_VIEW_CHARS. 
   - The minimum characters would be set to the longest tab text length.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -20,7 +20,6 @@ package org.eclipse.e4.ui.workbench.renderers.swt;
 import java.lang.reflect.Field;
 import java.util.Arrays;
 import java.util.Objects;
-import java.util.Optional;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.PreferenceChangeEvent;
@@ -1306,10 +1305,9 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 		boolean showFullText = getShowFullTextForViewTabsPreference();
 		if (!isPartOfEditorStack()) {
 			if (showFullText) {
-				Optional<Integer> lengthOfLongestItemText = Arrays.stream(parent.getItems()).map(CTabItem::getText)
-						.map(String::length)
-						.max(Integer::compare);
-				parent.setMinimumCharacters(lengthOfLongestItemText.orElseGet(() -> MAX_VIEW_CHARS));
+				int lengthOfLongestItemText = Arrays.stream(parent.getItems()).map(CTabItem::getText)
+						.map(String::length).max(Integer::compare).orElse(0);
+				parent.setMinimumCharacters(Math.max(MAX_VIEW_CHARS, lengthOfLongestItemText));
 			} else {
 				parent.setMinimumCharacters(MIN_VIEW_CHARS);
 			}
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -88,7 +88,7 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 	public static final boolean SHOW_FULL_TEXT_FOR_VIEW_TABS_DEFAULT = false;
 
 	private static int MIN_VIEW_CHARS = 1;
-	private static int MAX_VIEW_CHARS = 999999;
+	private static int MAX_VIEW_CHARS = Integer.MAX_VALUE;
 
 	// Constants for circle drawing
 	static enum CirclePart {
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -18,7 +18,6 @@
 package org.eclipse.e4.ui.workbench.renderers.swt;
 
 import java.lang.reflect.Field;
-import java.util.Arrays;
 import java.util.Objects;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
@@ -34,7 +33,6 @@ import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.CTabFolder;
 import org.eclipse.swt.custom.CTabFolderRenderer;
-import org.eclipse.swt.custom.CTabItem;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
@@ -90,7 +88,7 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 	public static final boolean SHOW_FULL_TEXT_FOR_VIEW_TABS_DEFAULT = false;
 
 	private static int MIN_VIEW_CHARS = 1;
-	private static int MAX_VIEW_CHARS = 9999;
+	private static int MAX_VIEW_CHARS = 999999;
 
 	// Constants for circle drawing
 	static enum CirclePart {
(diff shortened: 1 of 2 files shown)


merks:
This looks generally mostly bogus to me.    

If I add some print statements;

image

This is what I see printed when starting an IDE with the preference already set:
```
lengthOfLongestItemText=0
minimumCharacters=9999
lengthOfLongestItemText=0
minimumCharacters=9999
lengthOfLongestItemText=0
minimumCharacters=9999
```
So generally this logic is an expensive way to set the min to 9999, except (in principle) if some view has a title longer than that, which seems unlikely at best, and would not fit on a normal screen anyway.  But note that this appears to be computed primarily when there are no tabs yet, i.e., `parent.getItems()` returns an empty array on startup, hence the 0 in the print statement.

Why do we need to compute a minimum rather than using a reasonably large hard coded minimum.

shubhamWaghmare-sap:
> So generally this logic is an expensive way to set the min to 9999, except (in principle) if some view has a title longer than that, which seems unlikely at best, and would not fit on a normal screen anyway. But note that this appears to be computed primarily when there are no tabs yet, i.e., `parent.getItems()` returns an empty array on startup, hence the 0 in the print statement.
> 
> Why do we need to compute a minimum rather than using a reasonably large hard coded minimum.

@merks  I agree. Technically, just setting the minimum characters to a really large value would suffice.

**Possible change:**
image

The computation only makes sense (and has any effect) in the extreme case when a tab has a title longer than MAX_VIEW_CHARS. 
Since this is highly unlikely, I'll make the change and directly set the minimum characters to MAX_VIEW_CHARS, when full titles are to be shown and wait for further review.

MAX_VIEW_CHARS changed to 999999.

merks:
I think even in the extreme case that a view title is longer MAX_VIEW_CHARS, this computation is generally useless because it's happening before the tabs are even populated.  So setting a hard coded max is fine, but I really question such a huge "magic" number.  I.e., 999999 characters will not fit on any monitor, so to my mind, it's either there is no limit, e.g., MAX_INT, or there is a reasonable limit, e.g., something that fits on a monitor.

Can you please test what actually happens when a view title is stupidly big, e.g., 1000 characters?

merks:
@BeckerWdf 

I'm not sure that a fix is really badly needed for the release.  Generally what I see is that the min is being set to 9999 on startup because there are no tabs.  So there only seems to be a corner case of changing the preference while tabs are already showing and then opening a tab with a longer title.

Of course setting a proper min also seems to be risk free, given the code is guarded by a preference that is not enabled by default.  But huge, arbitrary magic numbers make me ask yet more questions.

BeckerWdf:
> I'm not sure that a fix is really badly needed for the release

I don't think that is's badly needed. It's a bug, but it's not a regression compared to the previous release and the feature is off by default.

merks:
This is what happens when I make a view actually have a title with around 1000 characters.  The drop down expands onto another monitor:

image

So there is a pre-existing bug that the drop down is not bound by the current monitor...

shubhamWaghmare-sap:
> So setting a hard coded max is fine, but I really question such a huge "magic" number. I.e., 999999 characters will not fit on any monitor, so to my mind, it's either there is no limit, e.g., MAX_INT, or there is a reasonable limit, e.g., something that fits on a monitor.

I believe we would not know how many characters a monitor would be able to fit (today and in future). Thus the limit could be MAX_INT. It's better than setting arbitrary number.

merks:
FYI, here is how it looks when I set the min to 100 but the title is more than a 1000 characters:

image

shubhamWaghmare-sap:
> Of course setting a proper min also seems to be risk free, given the code is guarded by a preference that is not enabled by default. 

Even if the preference is enabled by default (in future), I don't see a risk as such. 🤔

BeckerWdf:
> FYI, here is how it looks when I set the min to 100 but the title is more than a 1000 characters:
> 
> image

Having view with such long names anyway is not a thing which has a good UX. So I think we don't need to optimize for such cases. Setting a large "magic" number. If its 99999 or 999999 or... does not make much difference for me.

@merks: Is it still possible and meaningful for 2024-03? If yes it would be nice to include it as it fixes a big and the fix has a low risk. If not I also don't have an issue with that either. In that case we merge it once the next dev-cycle starts. It's totally up to you to decide.

merks:
It's really a unlikely corner case that I think it not worth fixing (for this release).

BeckerWdf:
> It's really a unlikely corner case that I think it not worth fixing (for this release).

That's ok.

jukzi:
Is this PR ready to merge?

BeckerWdf:
> Is this PR ready to merge?

I would say: Yes. Thanks for asking.

jukzi (approved):

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -91,7 +88,7 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 	public static final boolean SHOW_FULL_TEXT_FOR_VIEW_TABS_DEFAULT = false;
 
 	private static int MIN_VIEW_CHARS = 1;
-	private static int MAX_VIEW_CHARS = 9999;
+	private static int MAX_VIEW_CHARS = 999999;
```
merks: Please either stop using a magic number or explain and document in the code why  exactly the huge value 999999, rather than 999998, 1000000, or Integer.MAX_VALUE.  Especially because MAX_VALUE makes it clear that there is effectively no minimum.

shubhamWaghmare-sap: I agree. I have set _Integer.MAX_VALUE_ as MAX_VIEW_CHARS instead of any other large number. This constant is itself self explanatory and safe guards in case a tab with really large title text is possible (Mega screen).

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -88,7 +88,7 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 	public static final boolean SHOW_FULL_TEXT_FOR_VIEW_TABS_DEFAULT = false;
 
 	private static int MIN_VIEW_CHARS = 1;
-	private static int MAX_VIEW_CHARS = 999999;
+	private static int MAX_VIEW_CHARS = Integer.MAX_VALUE;
```
merks, line 91: Very good.

BeckerWdf:
CI says:

Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.7-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.ui.workbench.renderers.swt: Only qualifier changed for (org.eclipse.e4.ui.workbench.renderers.swt/0.16.300.v20240311-1145). Expected to have bigger x.y.z than what is available in baseline (0.16.300.v20240223-1002) -> [Help 1]

@shubhamWaghmare-sap: You have to update the version of org.eclipse.e4.ui.workbench.renderers.swt to 0.16.400 as the new dev cycle has started.

BeckerWdf:
So the build looks good to me.
"pr-head" says:

```
17:33:20.290 [INFO] ------------------------------------------------------------------------
17:33:20.290 [INFO] BUILD SUCCESS
17:33:20.290 [INFO] ------------------------------------------------------------------------
17:33:20.290 [INFO] Total time:  41:24 min
17:33:20.291 [INFO] Finished at: 2024-03-11T17:33:20Z
17:33:20.291 [INFO] ------------------------------------------------------------------------
Exception in thread "Thread-3" java.lang.NoClassDefFoundError: org/eclipse/jgit/internal/JGitText
	at org.eclipse.jgit.internal.util.ShutdownHook.cleanup(ShutdownHook.java:85)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.ClassNotFoundException: org.eclipse.jgit.internal.JGitText
	... 2 more
```

What's the issue here?

Failing tests are pre-existing.

Any objections if I merge this PR this evening?

jukzi:
> What's the issue here?

the build is marked unstable because of
https://github.com/eclipse-pde/eclipse.pde/issues/1193

3 API markers considered "new" despite they are unrelated and old:
image

The jgit issue is totally unrelated (https://github.com/eclipse-jgit/jgit/issues/17) and does not contribute to build mark broken.

BeckerWdf:
@jukzi: So you don't have objections to merge this?

jukzi:
I have no clue what this PR is about, but technically you can consider it green - it is totally OK to merge it now.

merks:
FYI,

This is just an annoying, but harmless/irrelevant thing when using Tycho 4.0.6:
```
Exception in thread "Thread-3" java.lang.NoClassDefFoundError: org/eclipse/jgit/internal/JGitText
	at org.eclipse.jgit.internal.util.ShutdownHook.cleanup(ShutdownHook.java:85)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.ClassNotFoundException: org.eclipse.jgit.internal.JGitText
	... 2 more
```

BeckerWdf:
@merks Thanks for merging.

Issue #865, opened 2023-05-26
Typing open bracket should surround the selected text element with brackets


vogella:
If I select an text element and type an opening bracket, e.g. "(" VScode the selected work is surrounded by brackets, if I do the same in Eclipse the work is deleted.

I think the VScode behavior is much nicer and noticed this difference while testing @mickaelistria LS-JDT integration for Eclipse.

Would be also adjust the behavior in Eclipse? If yes, which editor would be good to implement this? Text editor or generic editor?

BeckerWdf:
@vogella: Sorry I didn't get it. What is a "text element" and was is the "selected work"?

vogella:
Type WORD in an editor
Double-click on WORD to select it
Now type "(" -> WORD gets deleted in Eclipse

Do the same in code, the result is: -> (WORD)

mickaelistria:
Note that since we already have in most editors some strategy to auto-insert a closing brackets, this request is most likely just an extension of the existing strategies to tweak the handling of insertion with a non-empty selection. For the generic editor (and TM4E that already provide all the matching bracket smartness for a given language), it's mostly a matter of updating the `org.eclipse.tm4e.languageconfiguration.internal.LanguageConfigurationAutoEditStrategy`, for JDT it looks like it's a matter of changing CompilationUnitEditor.verifyKey() to add the overridden content between `				buffer.append(character)` and `buffer.append(closingCharacter)`.

mickaelistria:
FWIW, this just got implemented in TM4E (so it would serve many editors already, but not default JDT one): https://github.com/eclipse/tm4e/commit/501ede88852ec05ac9f0414c75783effa33cb1e9

vogella:
Awesome, thanks @lathapatil 

Also thanks @mickaelistria. I hope at some point the base JDT editor is changed to the generic code editor to benefit from such developments.

sratz:
The feature is now always active, unless overridden in code.

Shouldn't this feature be made optional using a preference page and/or coupled to the 'Smart Insert Mode'?

vogella:
Optional features are hard be discover for users. Why do you think it should not be active by default?

merks:
The way I read it, this is suggesting for a way to control the behavior with no comment about what the default should be...

sratz:
Yes, enabled by default is is fine, but I think there should be a way to disable this functionality, similar to Java-specific preferences:

[...]

PR #1718, opened 2024-02-26
New Bracket support 

Dinesh0723:
Introduced a new Bracket support.

Fixes:https://github.com/eclipse-platform/eclipse.platform.ui/issues/865

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultDocumentAdapter.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultDocumentAdapter.java
@@ -16,6 +16,7 @@ package org.eclipse.jface.text;
 
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
@@ -220,7 +221,26 @@ class DefaultDocumentAdapter implements IDocumentAdapter, IDocumentListener, IDo
 	@Override
 	public void replaceTextRange(int pos, int length, String text) {
 		try {
-			fDocument.replace(pos, length, text);
+			String[][] bracketTypes= {
+					{ SpecialCharacterConstants.openParentheses, SpecialCharacterConstants.openSquareBracket, SpecialCharacterConstants.openCurlyBracket, SpecialCharacterConstants.openAngleBracket,
+							SpecialCharacterConstants.singleQuote, SpecialCharacterConstants.doubleQuotes },
+					{ SpecialCharacterConstants.closeParentheses, SpecialCharacterConstants.closeSquareBracket, SpecialCharacterConstants.closeCurlyBracket,
+							SpecialCharacterConstants.closeAngleBracket, SpecialCharacterConstants.singleQuote, SpecialCharacterConstants.doubleQuotes }
+			};
+			boolean containsSpecialBrackets= Arrays.asList(bracketTypes[0]).contains(text);
+			TextSelection textSelection= new TextSelection(fDocument, pos, length);
+			//Checks if the text contains special type of brackets, the length of the selected text is greater than 0 & the selected text is not empty.
+			if (containsSpecialBrackets && textSelection.getLength() > 0 && !textSelection.getText().trim().isEmpty()) {
+				int index= Arrays.asList(bracketTypes[0]).indexOf(text);
+				if (index >= 0) {
+					String currentOpeningBracket= bracketTypes[0][index];
+					String currentClosingBracket= bracketTypes[1][index];
+					text= currentOpeningBracket + textSelection.getText() + currentClosingBracket;
+					fDocument.replace(pos, length, text);
+				}
+			} else {
+				fDocument.replace(pos, length, text);
+			}
 		} catch (BadLocationException x) {
 			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 		}
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SpecialCharacterConstants.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SpecialCharacterConstants.java
@@ -1,0 +1,43 @@
+/*******************************************************************************
+ * Copyright (c) 2023 ETAS GmbH and others, all rights reserved.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SpecialCharacterConstants.java
```diff
@@ -0,0 +1,43 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     ETAS GmbH - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.jface.text;
+
+/**
+ * @since 3.25
```
jukzi, line 18: @Dinesh0723 in principle the code and feature looks good. But i would not add new API for these otherwise unused constants. Please inline them

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultDocumentAdapter.java
```diff
@@ -220,7 +221,26 @@ public String getTextRange(int offset, int length) {
...
 	public void replaceTextRange(int pos, int length, String text) {
 		try {
-			fDocument.replace(pos, length, text);
+			String[][] bracketTypes= {
+					{ SpecialCharacterConstants.openParentheses, SpecialCharacterConstants.openSquareBracket, SpecialCharacterConstants.openCurlyBracket, SpecialCharacterConstants.openAngleBracket,
+							SpecialCharacterConstants.singleQuote, SpecialCharacterConstants.doubleQuotes },
+					{ SpecialCharacterConstants.closeParentheses, SpecialCharacterConstants.closeSquareBracket, SpecialCharacterConstants.closeCurlyBracket,
+							SpecialCharacterConstants.closeAngleBracket, SpecialCharacterConstants.singleQuote, SpecialCharacterConstants.doubleQuotes }
+			};
+			boolean containsSpecialBrackets= Arrays.asList(bracketTypes[0]).contains(text);
```
jukzi, line 230: You can store this List as constant "private final static", please. Hint: better use` List.of(...) `

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SpecialCharacterConstants.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class SpecialCharacterConstants {
```
laeubi, line 20:
```suggestion
class SpecialCharacterConstants {
```

should work as well....

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SpecialCharacterConstants.java
```diff
@@ -0,0 +1,43 @@
...
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     ETAS GmbH - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.jface.text;
```
mickaelistria, line 15: Please move it to org.eclipse.jface.text.source package, which already hosts some related feature such as ICharacterPairMatcher.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SpecialCharacterConstants.java
mickaelistria, line 20: Yes, please keep non-API, and better call it "DefaultCharacterPairs" here and host also the pairs in that class.

mickaelistria (changes requested):
We don't want to hardcode such feature into JFace, we want JFace to make such feature possible but not to be the host of the edition smartness. And this is usually done by the IAutoEditStrategies registered by the editor on the viewer.
Editors such as the Java one or the Generic Editor already host some smartness regarding finding or inserting matching brackets. We should stay one the same layer instead of hardcoding it into JFace.
Concretely, this feature could be implemented in some IAutoEditStrategy that would be added to the viewer via the SourceViewerConfiguration. JDT already has some `CompilationUnitEditor.getPeerCharacter()` to reuse; and the other case would be mostly the generic editor, where the "autoEdits" are usually provided by TM4E. So TM4E could be improved to turn "replace `foo` by `(`" into `replace `foo` by `(foo)`" according to the defined character matching rules.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SpecialCharacterConstants.java
lathapatil: @mickaelistria  As per my understanding, a non-API does not include public classes. If we move the class to the org.eclipse.jface.text.source package, it will not be possible to access the constants from DefaultDocumentAdapter.java, which is located in the org.eclipse.jface.text package.

Therefore, Is it better to inline them in DefaultDocumentAdapter.java as 
String[][] bracketTypes = {
                { "(", "[", "{", "<", "'", "\"" },
                { ")", "]", "}", ">", "'", "\"" }
        };
        
  Please correct me if I have misunderstood.

lathapatil:
> Editors such as the Java one or the Generic Editor already host some smartness regarding finding or inserting matching brackets. We should stay one the same layer instead of hardcoding it into JFace.

@mickaelistria  I will explore on this further if the approach provided in the PR is not acceptable

mickaelistria:
> @mickaelistria I will explore on this further if the approach provided in the PR is not acceptable

Yes please, I think it would be better.
Also, I would recommend that you start by implementing an automated test that describes the scenario you wish to support.

lathapatil:
@mickaelistria 
I have written a strategy for this feature and registered it in the `SourceViewer` class. It is currently functioning across all types of editors (tested with Text, Generic, Java, Ant, etc.).

Could you please confirm if registering the strategy in the [`SourceViewer`](https://github.com/eclipse-platform/eclipse.platform.ui/blob/fc44255456951ee0727f4fc02c9f61b7c97e9b2d/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java#L546) class (in the `Configure` method) is the correct approach?

mickaelistria:
> Could you please confirm if ...

I'm not sure. Feel free to push the code to this PR and I'll have a look.

PR #1722, opened 2024-02-27
Add Windows Defender preference keywords

Bananeweizen:
When running into the first problem with the new Windows Defender tooling, it was unclear where to search in preferences, since no matching keyword is defined.


BeckerWdf (approved):

BeckerWdf:
I think this a low-risk change. Should we still merge in the RC phase? What do you think @merks ?

merks:
This looks safe and good. +1 from PMC.  Merge when ready.

BeckerWdf:
@Bananeweizen : Thanks for providing the PR

Issue #195, opened 2022-07-07
[Tests] ProgressViewTests.testItemOrder is flaky 


ktatavarthi:
The testcase org.eclipse.ui.tests.progress.ProgressViewTests.testItemOrder keeps failing randomly in the I-build

This issue is created to track the [bug 553430](https://bugs.eclipse.org/bugs/show_bug.cgi?id=553430)

ktatavarthi:
This test failed in the I-Build [I20220706-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220706-1800/testresults/html/org.eclipse.ui.tests_ep425I-unit-macM1-java17_macosx.cocoa.aarch64_17.html) on macM1 with the below error

```
Wrong job order: arrays first differed at element [0]; expected:<1. User Job(1193)> but was:<4. Usual job 2(1196)>

Wrong job order: arrays first differed at element [0]; expected:<1. User Job(1193)> but was:<4. Usual job 2(1196)>
at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:78)
at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:28)
at org.junit.Assert.internalArrayEquals(Assert.java:534)
at org.junit.Assert.assertArrayEquals(Assert.java:285)
at org.eclipse.ui.tests.progress.ProgressViewTests.testItemOrder(ProgressViewTests.java:193) 
```

ktatavarthi:
This test failed in the I-Build [I20220928-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220928-1800/testresults/html/org.eclipse.ui.tests_ep426I-unit-macM1-java17_macosx.cocoa.aarch64_17.html) on macM1 with the below error

```
Wrong job order: arrays first differed at element [0]; expected:<1. User Job(1193)> but was:<4. Usual job 2(1196)>

Wrong job order: arrays first differed at element [0]; expected:<1. User Job(1193)> but was:<4. Usual job 2(1196)>
at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:78)
at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:28)
at org.junit.Assert.internalArrayEquals(Assert.java:534)
at org.junit.Assert.assertArrayEquals(Assert.java:285)
[...]

PR #1726, opened 2024-02-29
Additional Completion Information: remove delay & timeout #1695

jukzi:
Removes 500ms lag for javadoc information in the completion dialog. Sets related delays to 0 and fix code to handle such short delays.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/1695

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AdditionalInfoController.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AdditionalInfoController.java
@@ -52,7 +52,9 @@ class AdditionalInfoController extends AbstractInformationControlManager {
 	 * @since 3.2
 	 */
 	private static abstract class Timer {
-		private static final int DELAY_UNTIL_JOB_IS_SCHEDULED= 50;
+		/** delay until additional information is calculated **/
+		private static final int DELAY_UNTIL_JOB_IS_SCHEDULED= 0;
+ 
 
 		/**
 		 * A <code>Task</code> is {@link Task#run() run} when {@link #delay()} milliseconds have
@@ -324,11 +326,12 @@ class AdditionalInfoController extends AbstractInformationControlManager {
 
 		private void schedule(Task task, long current) {
 			fTask= task;
-			long nextWakeup= current + task.delay();
-			if (nextWakeup <= current)
+			long delay= task.delay();
+			if (delay == Long.MAX_VALUE)
+				// sleep until notify
 				fNextWakeup= Long.MAX_VALUE;
 			else
-				fNextWakeup= nextWakeup;
+				fNextWakeup= current + delay;
 		}
 
 		private boolean isExt5(ICompletionProposal p) {
@@ -376,7 +379,11 @@ class AdditionalInfoController extends AbstractInformationControlManager {
 
 		void allowShowing() {
 			fAllowShowing= true;
-			triggerShowing();
+			// triggerShowing() with fCurrentInfo==null would implicitly disallow any later showing
+			if (fCurrentInfo != null) {
+				// only if information is already available:
+				triggerShowing();
+			}
 		}
 	}
 	/**
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
@@ -75,7 +75,6 @@ import org.eclipse.jface.contentassist.ISubjectControlContentAssistProcessor;
 import org.eclipse.jface.dialogs.IDialogSettings;
 import org.eclipse.jface.preference.JFacePreferences;
 import org.eclipse.jface.util.Geometry;
-import org.eclipse.jface.util.OpenStrategy;
 import org.eclipse.jface.viewers.ISelectionChangedListener;
 
 import org.eclipse.jface.text.BadLocationException;
@@ -1580,8 +1579,11 @@ public class ContentAssistant implements IContentAssistant, IContentAssistantExt
 		fInternalListener= new InternalListener();
 
 		AdditionalInfoController controller= null;
-		if (fInformationControlCreator != null)
-			controller= new AdditionalInfoController(fInformationControlCreator, OpenStrategy.getPostSelectionDelay());
+		if (fInformationControlCreator != null) {
+			/** delay until each additional information is shown **/
+			int delay= 0;		
+			controller= new AdditionalInfoController(fInformationControlCreator, delay);
+		}
 
 		fContextInfoPopup= fContentAssistSubjectControlAdapter.createContextInfoPopup(this);
 		fProposalPopup= fContentAssistSubjectControlAdapter.createCompletionProposalPopup(this, controller, fAsynchronous);
--- a/tests/org.eclipse.ui.genericeditor.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.ui.genericeditor.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.genericeditor.tests;singleton:=true
-Bundle-Version: 1.3.400.qualifier
+Bundle-Version: 1.3.500.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.ui.genericeditor.tests,
--- a/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/CompletionTest.java
+++ b/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/CompletionTest.java
@@ -27,6 +27,7 @@ import static org.junit.Assume.assumeFalse;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Hashtable;
 import java.util.LinkedList;
@@ -212,16 +213,19 @@ public class CompletionTest extends AbstratGenericEditorTest {
 	}
 
 	public static Shell findNewShell(Set<Shell> beforeShells, Display display, boolean expectShell) {
-		Shell[] afterShells = Arrays.stream(display.getShells())
-				.filter(Shell::isVisible)
-				.filter(shell -> !beforeShells.contains(shell))
-				.toArray(Shell[]::new);
+		Shell[] afterShells = findNewShells(beforeShells, display).toArray(Shell[]::new);
 		if (expectShell) {
 			assertEquals("No new shell found", 1, afterShells.length);
 		}
 		return afterShells.length > 0 ? afterShells[0] : null;
 	}
-
+	public static Collection<Shell> findNewShells(Set<Shell> beforeShells, Display display) {
+		List<Shell> result= Arrays.stream(display.getShells())
+				.filter(Shell::isVisible)
+				.filter(shell -> !beforeShells.contains(shell))
+				.toList();
+		return result;
+	}
 	@Test
 	public void testCompletionFreeze_bug521484() throws Exception {
 		assumeFalse("test fails on Mac, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/906", Util.isMac());
@@ -248,7 +252,7 @@ public class CompletionTest extends AbstratGenericEditorTest {
 		emulatePressLeftArrowKey();
 		final Set<Shell> beforeShells = Arrays.stream(editor.getSite().getShell().getDisplay().getShells()).filter(Shell::isVisible).collect(Collectors.toSet());
 		DisplayHelper.sleep(editor.getSite().getShell().getDisplay(), 200);
-		this.completionShell= findNewShell(beforeShells, editor.getSite().getShell().getDisplay(), true);
+		this.completionShell= findNewShells(beforeShells, editor.getSite().getShell().getDisplay()).stream().filter(s->Arrays.stream(s.getChildren()).allMatch(w->w instanceof Table) ).findFirst().get();
 		final Table completionProposalList = findCompletionSelectionControl(this.completionShell);
 		checkCompletionContent(completionProposalList);
 	}


BeckerWdf (approved):
I manually tested this change with JDT's element info.
Also with the element info in our ABAP Development Tools.
This works as expected. Element Info in code completion is not much more snappy.

What about the Implementation in AdditionalInfoController2? Will you adapt it as well?

BeckerWdf:
Let's merge this early once dev for 4.32 opens.

BeckerWdf:
Pls. also have a look at the failing tests.

jukzi:
```
java.lang.AssertionError: No new shell found expected:<1> but was:<2>
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.eclipse.ui.genericeditor.tests.CompletionTest.findNewShell(CompletionTest.java:220)
	at org.eclipse.ui.genericeditor.tests.CompletionTest.testMoveCaretBackUsesAllProcessors_bug522255(CompletionTest.java:251)
```

```
Wrong job order: arrays first differed at element [0]; expected:<1. User Job(3523)> but was:<4. Usual job 2(3526)>
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:78)
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:28)
	at org.junit.Assert.internalArrayEquals(Assert.java:534)
	at org.junit.Assert.assertArrayEquals(Assert.java:285)
	at org.eclipse.ui.tests.progress.ProgressViewTests.testItemOrder(ProgressViewTests.java:192)
```

mickaelistria (changes requested):
In most cases, computing extra information for a completion item is expensive (and it's actually expensive  JDT and Javadoc rendering, and expensive for LSP-based editors), and it often creates no value as someone who would hit repeatedly the down key in the completion list would not need the intermediary extra documentations and only wants the one for the target. Extra information is useless in many cases, so I don't think constantly computing extra useless information is a feature here, I see it more as a way to overload CPU with uninteresting operations.
IMO, a delay is necessary for best usability and performance; removing it would be a regression. A better value for the delay could maybe be found though.

BeckerWdf:
> In most cases, computing extra information for a completion item is expensive (and it's actually expensive JDT and Javadoc rendering, and expensive for LSP-based editors), and it often creates no value as someone who would hit repeatedly the down key in the completion list would not need the intermediary extra documentations and only wants the one for the target. Extra information is useless in many cases, so I don't think constantly computing extra useless information is a feature here, I see it more as a way to overload CPU with uninteresting operations. IMO, a delay is necessary for best usability and performance; removing it would be a regression. A better value for the delay could maybe be found though.

There are two delays 50 ms befor starting the computation. and 400 ms *after* the computation before it's displayed.
At least the 400ms don't help with the issues you talk about (as the calculation is done anyways).

jukzi:
> In most cases, computing extra information for a completion item is expensive

Please see the analysis in linked issue: the "timeout" did not prevent computing that information anyway. Please also note that this implementation was tested with both javadoc and even ABAPs network stuff and did not show a problem. If you find any performance issues please report them with a way how to reproduce.

mickaelistria:
Both computating the data (first delay) and rendering it (second delay) can be expensive. But I don't think it makes a lot of sense to separate those. A single delay to rule them all would be enough, and there is indeed no value in computing the data if it's not to render it ASAP.

mickaelistria:
> If you find any performance issues please report them with a way how to reproduce.

Can you please check the overall CPU load of browsing a list of proposals? ie if I'm hitting the down arrow key 5 times to get the 6th completion item, I believe this change will cause 5 times more computations than before, for no obvious value.
It may not be perceivable by users, but more CPU for not clear additional value is a performance regresion.

jukzi:
> ie if I'm hitting the down arrow key 5 times to get the 6th completion item

sorry but i am unable to press down key 5 times within 50ms

laeubi:
> > ie if I'm hitting the down arrow key 5 times to get the 6th completion item
> 
> sorry but i am unable to press down key 5 times within 50ms

This will probably help you:

https://www.youtube.com/watch?v=GvxjnW7cx20

Anyways the second delay is as mentioned before to prevent flickering (e.g. when scrolling through items by *holding the key down(!)*), or by moving the mouse around and to adhere to the principle that only if I'm "long enough" over an item I get additional information, e.g. for Tooltips see this one:

https://uxpickle.com/delay-before-the-tooltip-shows-up/

jukzi:
> ProgressViewTests

random failing see https://github.com/eclipse-platform/eclipse.platform.ui/issues/195

> CompletionTest

fixed the test

jukzi:
> Anyways the second delay is as mentioned before to prevent flickering

There is no flickering, the additional shell stays open anyway. please stick to real issues that are reproducible.

laeubi:
> > Anyways the second delay is as mentioned before to prevent flickering
> 
> There is no flickering, the additional shell stays open anyway. please stick to real issues that are reproducible.

But only if the "real" content is always the same for all elements what would make such information quite useless.

Just see this example where the size and contents "jumps around" as one moves the mouse (still with delay so I need to hold over this):

https://github.com/eclipse-platform/eclipse.platform.ui/assets/1331477/b9034ee3-d46d-40c7-9b69-362a94720cd4

now assume that **on every move over the document** the popup is show immediately that will flicker around all the way an will at least drive me crazy, and that's the purpose of these delays, so even if the information is there it is only show when the user actually stops for a while.

jukzi:
> now assume that **on every move over the document** the popup

That popup is not a completion proposal. It is totally unrelated to the completion popup target by this PR.

laeubi:
> > now assume that **on every move over the document** the popup
> 
> That popup is not a completion proposal. It is totally unrelated to the completion popup target by this PR.

It shows the problem and the background for delays, just open a completion proposal with many proposals, do completion, wait for the popup to show up and then hold down the down key.

**With delay**: It shows the last item unless you release the key
**Without delay**: the text changes (depending on how fast it could be computed) causing "flickering" as it is constantly redrawn.

jukzi:
> **Without delay**: the text changes (depending on how fast it could be computed) causing "flickering" as it is constantly redrawn.

Mind to share a video of such flickering and how to reproduce? To me it looks just smooth.

laeubi:
> > **Without delay**: the text changes (depending on how fast it could be computed) causing "flickering" as it is constantly redrawn.
> 
> Mind to share a video of such flickering and how to reproduce? To me it looks just smooth.

If you like watching text-movies it looks smooth, but any animation any change or alike will catch up the eye so one should just avoid that, it is already moving the highlight of the selection, any other thing is just distracting: https://uxdesign.cc/the-ultimate-guide-to-proper-use-of-animation-in-ux-10bd98614fa9

BeckerWdf:
> Fix CompletionTest to find the completionShell

Jörgs Change does not affect this. It's about the element information in the completion popup. Pls. really do test the correct stuff.

BeckerWdf:
> If you like watching text-movies it looks smooth, but any animation any change or alike will catch up the eye so one should just avoid that, it is already moving the highlight of the selection, any other thing is just distracting: https://uxdesign.cc/the-ultimate-guide-to-proper-use-of-animation-in-ux-10bd98614fa9

This article speaks about animations. I don't see how this is related to this PR.

BeckerWdf:
> and even ABAPs network stuff and did not show a problem.

to be honest there is not problem with out "network stuff" because we anyway have a throttling in pace that send's requests to the server ever 400 ms at max. If the last request was triggered not longer then 400 ms in the past we don't trigger a new one. We have this in place since ages. Sending a request to the server on every selection change immediately wouldn't be a good idea.

laeubi:
Yes for sure everything is always unrelated so why asking ;-)

It was asked why there are two delays, I explained why in UX design one has such delays --> unrelated

It was asked why there is flickering if it is "smooth" (aka animation) , I gave a reference how animations are correctly used --> unrelated 

Jörg itself fixed a test about completion that fails now he change things --> unrelated

BeckerWdf:
> and rendering it (second delay) can be expensive

Pls. provide proof for this. I did test this. Showing the anyway computed information immediately did not cause any issue on my computer. I did not see that the selection did move slower and less smooth when simply holing the down key down.

Any pls. guys: Holding down the key for seconds is not a real-word use case - it's synthetic test case. In real word users might press down the cursor keys maybe 5-6 times repeatedly fast after each other. If the needed proposal is not found they  might simply continue typing in the editor to narrow down the list of completions.

Maybe we can agree on keeping the 50ms that throttles the computation - but I don't see any value (in todays reality with fast computers) in waiting half a second before rendering the information.

BeckerWdf:
> Yes for sure everything is always unrelated so why asking ;-)
> 
> It was asked why there are two delays, I explained why in UX design one has such delays --> unrelated
> 
> It was asked why there is flickering if it is "smooth" (aka animation) , I gave a reference how animations are correctly used --> unrelated
> 
> Jörg itself fixed a test about completion that fails now he change things --> unrelated

@laeubi: 
Your communication behavior is not constructive and goal-oriented. It does not comply with many points of the Code of Conduct of the Eclipse Foundation (https://www.eclipse.org/org/documents/Community_Code_of_Conduct.php). For example :
- Using welcoming and inclusive language
- Actively encouraging all voices
- Helping others bring their perspectives and listening actively. 
- Being respectful of differing viewpoints and experiences
- Gracefully accepting constructive criticism

Please let's try to come back to a constructive discussion.

BeckerWdf:
> and fix code to handle such short delays.

This really concerns me a bit. Might there be code outside of platform / JDT / ... that might break due to this change? @jukzi what's your feeling on this?

laeubi:
> Please let's try to come back to constructive discusion.

You mean starting each sentence where you want to actually *force* someone by "please"?
Or to simply qualify every unwanted thing as "totally unrelated" or even "not a real-word" ... but yes for sure its all me being not so friendly and welcoming any change.

It was _repeatably_ discussed and accepted that a lower delay is suitable, but every time is is insisted that it **must be zero** (because it is meant to actually fix a different problem) and every one that don't agree is just an non real world idiot. So these things tend to always running in circles.

Also _please_ note that I didn't _blocked_ this PR in any way I just gave some _hints_ why "delay = zero" is _probably_ a bad choice from UX POV. 

So maybe it would help if one not only demands things from other but actively following the mentions Code of Conduct and to be a role model in this regards.

jukzi:
> This really concerns me a bit. Might there be code outside of platform / JDT / ... that might break due to this change? @jukzi what's your feeling on this?

The comment was only because the Additional Completion internally had a problem with substracting those values, making wrong assumptions. Also all Completiontests which assert a delay may of cause break, but just of a needless assumption. 
If any problem remains its likely to be fixable. Fixing the code could even stay when reverting the values.

BeckerWdf:
> > This really concerns me a bit. Might there be code outside of platform / JDT / ... that might break due to this change? @jukzi what's your feeling on this?
> 
> The comment was only because the Additional Completion internally had a problem with substracting those values, making wrong assumptions. Also all Completiontests which assert a delay may of cause break, but just of a needless assumption. If any problem remains its likely to be fixable. Fixing the code could even stay when reverting the values.

But could you also imagine implementations out in the wild that might break because platform now is faster?

BeckerWdf:
> but every time is is insisted that it **must be zero** and every one that don't agree is just an non real world idiot

This is not true. *I* did not insist that every delay *must* be zero. See:

> Maybe we can agree on keeping the 50ms that throttles the computation - but I don't see any value (in todays reality with fast computers) in waiting half a second before rendering the information.

and I also didn't us the word *idiot*.

jukzi:
> because it is meant to actually fix a different problem

The Statusline problem was only the initial trigger to stumble across this delay. This change is totally unrelated to that - which will be fixed by https://github.com/eclipse-platform/eclipse.platform.ui/pull/1706. 

> is just an non real world idiot.
> ...
> following the mentions Code of Conduct

That's a strange combination of sentences.

jukzi:
> But could you also imagine implementations out in the wild that might break because platform now is faster?

As any change can break adopters (https://www.hyrumslaw.com/) -of cause. We could also provide system properties for the actual timings to allow a revert without touching the binary.

mickaelistria:
> In real word users might press down the cursor keys maybe 5-6 times repeatedly fast after each other.

Yes, this is indeed the case to focus on here. And I don't think this proposal here which is about computing and rendering the information for the 5 hits is a better solution for than reducing the delay to a typical duraction of repeated hits (which I imagine is around ~300ms but is probably documented in some UX recommendation somewhere), which is probably what the `OpenStrategy.getPostSelectionDelay()` is already trying to capture.
Note that I think removing the hardcoded 50ms delay here makes sense, but I think the `OpenStrategy.getPostSelectionDelay()` should still be obeyed (even if we change it to some other value later as part of #1695 ).

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AdditionalInfoController.java
```diff
@@ -52,7 +52,9 @@ class AdditionalInfoController extends AbstractInformationControlManager {
 	 * @since 3.2
 	 */
 	private static abstract class Timer {
-		private static final int DELAY_UNTIL_JOB_IS_SCHEDULED= 50;
+		/** delay until additional information is calculated **/
+		private static final int DELAY_UNTIL_JOB_IS_SCHEDULED= 0;
```
mickaelistria, line 56: As mentioned on more general comments, I think this value would better be removed all together and only OpenStrategy.getPostSelectionDelay() be used.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
```diff
@@ -1580,8 +1579,11 @@ protected void install() {
 		fInternalListener= new InternalListener();
 
 		AdditionalInfoController controller= null;
-		if (fInformationControlCreator != null)
-			controller= new AdditionalInfoController(fInformationControlCreator, OpenStrategy.getPostSelectionDelay());
+		if (fInformationControlCreator != null) {
+			/** delay until each additional information is shown **/
```
mickaelistria, line 1583: I think it's better to keep te OpenStrategy.getPostSelection() delay here (and probably improve its value in another PR to match the case of repeated hits on arrow keys)

mickaelistria (changes requested):

jukzi:
> What about the Implementation in AdditionalInfoController2? Will you adapt it as well?

I don't know when/how it is used and don't want to touch things blindly. Can you hint to a usecase?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AdditionalInfoController.java
jukzi: As there seems to be no mutual consent if this delay makes sense at all or if the removal could break anything i prefer to stick with 0 for easier tuning.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
jukzi: As @laeubi mentioned: in other usecases (javadoc hover) a delay in other usecases may make sense.  Therefore i prefer to touch only usecase by usecase. In this usecase the delay seem to not show any positive value.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AdditionalInfoController.java
mickaelistria: OK.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
mickaelistria: > Therefore i prefer to touch only usecase by usecase.

The thing here is that you're not touching usecase by usecase, but you're touching all use-cases at once. Also, if a delay make sense for some use-cases, it's likely to make sense for some other.

mickaelistria (commented):
I'm still not convinced that this approach is viable. If the issue is mostly to improve reactivity, I think the right place to achieve that widely is to set a better value to OpenStrategy.getPostSelectionDelay().

The ContentAssistant could be widely improved (eg to use a Throttler, to remove the "data" delay vs the popup delay ...) but it seems to be a wider scope than what's intended in this patch.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
jukzi: enlighten me; what is any other usecase here then showing the additional completion window?  This value was only used to show the wrong (outdated) information for a fixed 0.5 sec and if the computation took too long to never show it. My guess is you still have not tried out what this value does. That was more a bug then a feature.

mickaelistria: > what is any other usecase here then showing the additional completion window?

Unless we're talking about different thing, any IContentAssistProcessor can return an IContentAssistProposal that returns a `getAdditionalProposalInfo` (which may be long to compute) and that will show an additional completion window (which may be long to render) in order to show completion details. CDT, LSP4E-based projects such as Wild Web Developer or ShellWax or m2e... and many others are example.

>  This value was only used to show the wrong (outdated) information for a fixed 0.5 sec 

I suspect the value here is used to prevent from extra computation and flickering with new information if users performs other actions in a short time. Retaining an outdated value/popup for some ms is more comfortable than "blinking" popup.
Again, the root cause of the whole discussion is that 0.5s is too much and doesn't fit the reality of keyboard navigation. Just moving this value to ~0.3s (in `OpenStrategy`) would make a big difference without challenging all the existing performance or UX optimizations.

jukzi: Just that there is no flickering nor computational saving nor any blinking popup related to this value. The popup open if and only if you press the accelerator and closes when you escape or commit.
That we additional reduce OpenStrategy for other usecase is still an option. Even if rendering takes long it does so after the delay. If you  press keys faster then the rendering (which i still doubt or would qualify you for the tetris world record) the rendering is skipped anyway by the existing double check that the proposal is still the current. This state machine already has a build in throttler. just hard to see because it is a bit overcomplicated

akurtakov:
Nothing happened here for more than a year and ECA of submitter is no longer active thus closing this one.

PR #1730, opened 2024-03-05
AboutPluginsPage: show signed Date in column (lazy, sortable)

jukzi:
Also adapt column width to show most values without ellipsis


jukzi:
before:
image

after:
image

merks:
That's a lot of detail.  Do you think it's ever important to know the time of day, down to the exact second?

jukzi:
> That's a lot of detail. Do you think it's ever important to know the time of day, down to the exact second?

For me its most important see the day. I just took the format that was already used in that dialog.

jukzi:
i'll change to short:
image

PR #1738, opened 2024-03-07
fix TextViewer postSelectionChanged for jdt ParameterNamesCodeMiningTest

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1236

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
@@ -97,13 +97,27 @@ public class Throttler {
 	 * returning.
 	 */
 	public void throttledExec() {
+		throttledExec(true);
+	}
+
+	/**
+	 * Like {@link #throttledExec()} but if called from Display Thread guaranteed to
+	 * return before Runnable is run
+	 *
+	 * @since 3.34
+	 */
+	public void throttledAsyncExec() {
+		throttledExec(false);
+	}
+
+	private void throttledExec(boolean allowSyncExec) {
 		if (display.isDisposed()) {
 			return;
 		}
 		if (scheduled.compareAndSet(false, true)) {
 			boolean exception = true;
 			try {
-				if (Thread.currentThread() == display.getThread()) {
+				if (allowSyncExec && Thread.currentThread() == display.getThread()) {
 					timerExec.run();
 				} else {
 					display.asyncExec(timerExec);
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewer.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewer.java
@@ -1541,11 +1541,6 @@ public class TextViewer extends Viewer implements
 	 */
 	private List<ISelectionChangedListener> fPostSelectionChangedListeners;
 	/**
-	 * Queued post selection changed events count.
-	 * @since 3.0
-	 */
-	private final int[] fNumberOfPostSelectionChangedEvents= new int[1];
-	/**
 	 * Last selection range sent to post selection change listeners.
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
```diff
@@ -97,13 +97,27 @@ public Throttler(Display display, Duration minWaitTime, Runnable runnable) {
...
+
+	private void throttledExec(boolean allowSyncExec) {
 		if (display.isDisposed()) {
 			return;
 		}
 		if (scheduled.compareAndSet(false, true)) {
 			boolean exception = true;
 			try {
-				if (Thread.currentThread() == display.getThread()) {
+				if (allowSyncExec && Thread.currentThread() == display.getThread()) {
```
iloveeclipse, line 120: Somehow the condition and the code do not match. `if(allowSyncExec) timerExec else asyncExec` makes no sense. Should it be something like `if forceTimerExec timerExec else asyncExec`? 

TimerExec is also asynchronous, argument name is misleading.

jukzi: "allow"+"sync" vs "async"

bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
```diff
@@ -97,13 +97,27 @@ public Throttler(Display display, Duration minWaitTime, Runnable runnable) {
 	 * returning.
 	 */
 	public void throttledExec() {
+		throttledExec(true);
+	}
+
+	/**
+	 * Like {@link #throttledExec()} but guaranteed to not run the Runnable before
+	 * this method returns.
```
iloveeclipse: `guaranteed to not run the Runnable before this method returns.`
The code doesn't match the contract.
One can put breakpoint at line 124 and the code on display will run before the return, assuming current call is not in the UI thread.

jukzi: valid point. i fixed the javadoc

bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
iloveeclipse: `allowSyncExec` -> `allowTimerExec` ? 
Just reading the code for the first time one could assume here is something that is supposed to run synchronous, but it is never the case.

jukzi: I would not mind to rename anything, but actually timerExec is a Runnable that is just run() synchronous when allowSyncExec and runs runner.run() synchronous. Maybe the name "timerExec" does confuse you more

jukzi:
The failing MultiVariablePageTest.testContextActivation is unrelated regression from 0532901782056835d47d4d13741a1492eb06815c
@iloveeclipse good to merge?

bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java
iloveeclipse: `timerExec is a Runnable that is just run() synchronous`

Are we talking about `Display.timerExec()`? It is asynchronous.

jukzi: No, about Throttler.timerExec which is run() in 121 and runs Runner in 79 without Display.timerExec() as in 85

iloveeclipse: OK, I see.

iloveeclipse:
I'm fine with that.

PR #1739, opened 2024-03-07
Jenkinsfile: Not fail build on failed test

jukzi:
Not fail build on failed test, to still show the ecj warnings if tests fail. Copied from jdt


jukzi:
Currently all builds fail due to unrelated swt change, which has the effect that no warnings are shown.
With this change the build will only show `unstable ` result but also related warnings.

jukzi:
```
[Pipeline] archiveArtifacts
Archiving artifacts
‘*.log’ doesn’t match anything, but ‘tests/org.eclipse.ui.tests/target/work/data/.metadata/*.log’ does. Perhaps that’s what you mean?
No artifacts found that match the file pattern "*.log,*/target/work/data/.metadata/*.log,*/tests/target/work/data/.metadata/*.log,apiAnalyzer-workspace/.metadata/*.log". Configuration error?
```

huh?

mickaelistria:
The message is relatively clear, it tells that the regexp doesn't match any file. The regexp could add `tests/‘tests/*/target/work/data/.metadata/*.log`

jukzi:
The build now shows the warnings, is marked as unstable but still "fails" on github - as supposed 
image
image
image
the *.log error is unrelated

Issue #1740, opened 2024-03-08
[Find/Replace] Search-scopes don't update


Wittmaxi:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

1) Open a file
2) select any area of text
3) Open the Find/Replace-Dialog
4) select another area of text
5) press "selected lines" in the Find/Replace-Dialog
6) see that the old area is used as search-scope, instead of using the new scope

I expected the search-scope to update with the new selection

But got the old search-scope that was first deactivated and re-activated again.

## Tested under this environment:
Eclipse SDK

Version: 2024-03 (4.31)
Build id: I20240229-0520

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

jukzi:
i wonder why i can not reproduce that bug

Wittmaxi:
@jukzi This is the behavior on my end:
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/b4764fe5-b396-4da3-8286-c55b90910bc2)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/2cae8bcc-edcd-41c4-8833-540f1816783e)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/e69f0801-91b1-458b-af9b-93f64f7dc8b6)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/57e97108-da77-43a1-8961-5a6fce1c6bb0)

I have implemented a Fix+Regression-Test [here](https://github.com/Wittmaxi/eclipse.platform.ui/commit/6ebfd8a03f1b7e9f97786cc1d86bb1653d410fb7)

jukzi:
tried to reproduce again: this time i faced this:
1. activate "Selected lines"
2. select other lines (without closing dialog)
=> "Selected lines" is deselect and Scope=All is set. WTF

HeikoKlare:
I can reproduce the bug:
[...]

PR #1741, opened 2024-03-08
Fix: "Search-scopes don't update #1740"

Wittmaxi:
Updates the search-scopes when a new selection is created. Also implements a regression-test to the FindReplaceDialogTest-class.
This test will be re-implemented to test the behaviour in both the FindReplaceOverlay and the FindReplaceDialog.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -50,7 +50,6 @@ import org.eclipse.ui.texteditor.IFindReplaceTargetExtension2;
 public class FindReplaceLogic implements IFindReplaceLogic {
 	private IFindReplaceStatus status;
 	private IFindReplaceTarget target;
-	private IRegion oldScope;
 	private Point incrementalBaseLocation;
 
 	private boolean isTargetSupportingRegEx;
@@ -65,7 +64,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 
 		switch (searchOption) {
 		case GLOBAL:
-			useSelectedLines(false);
+			unsetSearchScope();
 			break;
 		case FORWARD:
 		case INCREMENTAL:
@@ -73,7 +72,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 				initIncrementalBaseLocation();
 			}
 			break;
-			// $CASES-OMITTED$
+		// $CASES-OMITTED$
 		default:
 			break;
 		}
@@ -86,7 +85,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		}
 
 		if (searchOption == SearchOptions.GLOBAL) {
-			useSelectedLines(true);
+			initializeSearchScope();
 		}
 
 		if (searchOption == SearchOptions.FORWARD && shouldInitIncrementalBaseLocation()) {
@@ -152,10 +151,8 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	/**
 	 * Tells the dialog to perform searches only in the scope given by the actually
 	 * selected lines.
-	 *
-	 * @param selectedLines <code>true</code> if selected lines should be used
 	 */
-	private void useSelectedLines(boolean selectedLines) {
+	private void initializeSearchScope() {
 		if (shouldInitIncrementalBaseLocation()) {
 			initIncrementalBaseLocation();
 		}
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -169,7 +169,7 @@ private void useSelectedLines(boolean selectedLines) {
 		if (selectedLines) {
 
 			IRegion scope;
-			if (oldScope == null) {
+			if (target.getSelection().y > 1 || oldScope == null) {
 				Point lineSelection = extensionTarget.getLineSelection();
```
HeikoKlare: What is the reason for not using the currently selected line if there is no selection but just the cursor being in some line? I would expect that the line in which the cursor is currently placed is taken as selected line when activating the option.

As a more general question: why do we need to store the `oldScope` at all? I would expect scope to be set once the option is activated and whenever it is reset to then use the selection at that point in time (rather than an old one).

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -366,6 +378,37 @@ public void testFindWithWholeWordEnabledWithMultipleWords() {
 		assertEquals(dialog.findCombo.getText().length(), (target.getSelection()).y);
 	}
 
+	@Test
+	public void testGlobalSearchChangeScope() {
+		openTextViewerAndFindReplaceDialog("line1\nline2\nline3");
+
+		fTextViewer.setSelection(new TextSelection(6, 11));
+		dialog.globalRadioButton.setSelection(false);
+		dialog.selectedRangeRadioButton.setSelection(true);
```
HeikoKlare: Isn't it also possible to set the already provided `globalRadioButton` to `false`?

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -366,6 +378,37 @@ public void testFindWithWholeWordEnabledWithMultipleWords() {
...
+		dialog.selectedRangeRadioButton.notifyListeners(SWT.Selection, null);
+		runEventQueue();
```
HeikoKlare: Why is it necessary to run the event queue here?

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -366,6 +378,37 @@ public void testFindWithWholeWordEnabledWithMultipleWords() {
...
 	}
 
+	@Test
+	public void testGlobalSearchChangeScope() {
+		openTextViewerAndFindReplaceDialog("line1\nline2\nline3");
+
+		fTextViewer.setSelection(new TextSelection(6, 11));
+		dialog.globalRadioButton.setSelection(false);
+		dialog.selectedRangeRadioButton.setSelection(true);
+		dialog.selectedRangeRadioButton.notifyListeners(SWT.Selection, null);
```
HeikoKlare: Please use the existing `select()` and `deselect()` operation instead of reimplementing the same logic.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -90,22 +93,31 @@ private class DialogAccess {
...
+
 		Button regExCheckBox;
 
 		private Supplier<Shell> shellRetriever;
 
 		private Runnable closeOperation;
 
+		public Combo replaceCombo;
+
+		public Button selectedRangeRadioButton;
```
HeikoKlare: I would propose to place the added fields at more reasonable positions, i.e., `replaceCombo` after `findCombo` without `public` modifier, `selectedRangeRadioButton` after `globalRadioButton` without public modifier, and `replaceAllButton` and `replaceAllButton` not in between the state-changing buttons.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -366,6 +378,37 @@ public void testFindWithWholeWordEnabledWithMultipleWords() {
 		assertEquals(dialog.findCombo.getText().length(), (target.getSelection()).y);
 	}
 
+	@Test
+	public void testGlobalSearchChangeScope() {
```
HeikoKlare: Shouldn't this be a test case of the `FindReplaceLogic` rather than (or in addition to) the `FindReplaceDialog`? It already affects the business logic, no matter which UI is used to execute it.

Wittmaxi: I have removed the test and added a similar test in FindReplaceLogicTest with this commit: https://github.com/eclipse-platform/eclipse.platform.ui/commit/fa4cfb1121b8ac0f2fe8f1bbb0ae46cc656dc22e

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: I have removed the test and added a similar test in FindReplaceLogicTest with this commit: https://github.com/eclipse-platform/eclipse.platform.ui/commit/fa4cfb1121b8ac0f2fe8f1bbb0ae46cc656dc22e

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: I have removed the test and added a similar test in FindReplaceLogicTest with this commit: https://github.com/eclipse-platform/eclipse.platform.ui/commit/fa4cfb1121b8ac0f2fe8f1bbb0ae46cc656dc22e

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: I have removed the test and added a similar test in FindReplaceLogicTest with this commit: https://github.com/eclipse-platform/eclipse.platform.ui/commit/fa4cfb1121b8ac0f2fe8f1bbb0ae46cc656dc22e

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: I have removed the test and added a similar test in FindReplaceLogicTest with this commit: https://github.com/eclipse-platform/eclipse.platform.ui/commit/fa4cfb1121b8ac0f2fe8f1bbb0ae46cc656dc22e

HeikoKlare (approved):
Changes are sound and work as expected. Regression test is appropriate for the fixed issue.

Thanks for the fix!

Issue #1063, opened 2023-08-24
Project Explorer: automatically expand nested folders


Wittmaxi:
In the project explorer and similar treeviews, when expanding a folder that contains a singular folder, I expect that folder to recursively expand as well.

### Examples
In these cases, I expect all folders to open recursively
```
com.wittmaxi.plugin
 └─src
    └─org.foo.com
       └─Bar.java
```
```
com.wittmaxi.plugin
 └─src
    ├─org.foo.com
    │  └─Bar.java
    └─org.baz.com (empty)
```
#### In Eclipse
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/0de3a584-b46a-4d4d-be70-1b61902c4136)
(click on  `src`)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/c93279a7-7ffe-42ef-b551-3f263c95748c)
(click on `com.vogella.contribute.parts`)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/eb2cb49c-a4d6-485a-99b2-4eade953e6c2)

#### In IntelliJ
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/9b2f1672-4e58-429d-a1ba-3075ad30ba2f)
(click on `src`)
![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/16443184/2c05bcf7-9765-479e-870a-3adead4ead50)

Note: eclipse shows a lot of these nested folders in form of "packages"

### Implementation
It makes sense to implement this feature as an option in the SWT-Treeview-Widget, which would allow even more views to benefit from this behaviour.

I am ready to create a PR that implements this feature and currently looking for feedback on my proposal.

HeikoKlare:
I would appreciate providing the proposed behavior. It could make sense to (optionally) provide this for all kinds of (SWT) trees but it is of course particularly useful for views visualizing folder structures, such as in the project or package explorer.

First, some investigation where the functionality needs to be placed should be conducted. At least on Windows, `TreeItem.setExpanded()` does only seem to be called from specific operations, such as double-clicking a node or using "expand all". Only clicking the "expand marker" next to a node does not invoke that method. I did not find any other method in `TreeItem` handling expand events, so maybe the "simple expand functionality" is provided by the operation system's control. In that case, OS events conducting the expand operation would have to be perceived and processed accordingly.
[...]

PR #1742, opened 2024-03-08
Automatically Expand Paths of single children in the Project-Explorer #1063

Wittmaxi:
Allows a user to automatically fully open nested paths that consist of only one expandable folder at each level.

grafik
(click on src automatically opens all levels shown below src)

I have set the auto-expansion to be at most 10 nodes deep:
I don't want infinite expansion to avoid loading a lot of lazy-loaded nodes. I think 10 nodes is enough to have a realistic benefit while not becoming too much.

For reference, this is how other tools handle this case:
- IntelliJ will auto-expand to infinite depth (tested with 106 nested folders)
- VScode will auto-expand the entire depth but will not actually display the folders and rather show the entire path in the folder-node: like so: "\a\...\a\a\a\a"

A few things are yet to be discussed (help wanted!):

- Should this be always enabled by default or should there be a preference to toggle this behavior?
- What should be the default "auto-expansion-depth"? 10 Nodes? Or infinitely many? Should there be a preference for those values?

### DIFF
--- a/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/navigator/resources/ProjectExplorer.java
+++ b/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/navigator/resources/ProjectExplorer.java
@@ -28,6 +28,7 @@ import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.SafeRunner;
+import org.eclipse.jface.viewers.AbstractTreeViewer;
 import org.eclipse.jface.viewers.DoubleClickEvent;
 import org.eclipse.jface.viewers.ILabelProvider;
 import org.eclipse.jface.viewers.IStructuredSelection;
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Be aware about https://github.com/eclipse-platform/eclipse.platform.swt/issues/1120 on Mac, not that this would cause troubles there.

HeikoKlare:
> Be aware about [eclipse-platform/eclipse.platform.swt#1120](https://github.com/eclipse-platform/eclipse.platform.swt/issues/1120) on Mac, not that this would cause troubles there.

Thanks for the pointer, @iloveeclipse! To my understanding, the problem in https://github.com/eclipse-platform/eclipse.platform.swt/issues/1120 is that there is a large number of elements to be expanded, as with the Option+Click every element in the whole subtree is expanded. So how long that operation takes depends on the size of the subtree.
The automatic expansion implemented by this PR does only apply if there is a single child element and is limited by a depth of 10. It is equivalent to manually performing 10 expansions on tree items and is even limited by adding at most 10 shown elements to the tree in addition to the (potentially much higher) number of child elements of the last one to be automatically extended. In particular, it will never apply to the dependencies entries (JRE and Plug-in dependencies) of a project, which have huge subtrees.
So I do not see any risk for a performance problem related to the issue in https://github.com/eclipse-platform/eclipse.platform.swt/issues/1120.

To be sure, I have also validated the PR on a Mac and found that it behaves as explained above. I did not find any performance issues yet. I was only able to produce the problems of https://github.com/eclipse-platform/eclipse.platform.swt/issues/1120 by applying the complete subtree expansion via Option+Click.

HeikoKlare (approved):
The change looks good to me, works as expected and makes sense to me. It reduces the effort to expand "trivial paths" in the project explorer, like done in IntelliJ or VScode.

Failing test is already documented as randomly failing: #294.

PR #1752, opened 2024-03-13
Extend BadLocationException by illegal argument information

jukzi:
As logged for example during
ConsoleDocumentAdapterTests.testInvalidInvocations()

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/ListLineTracker.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/ListLineTracker.java
@@ -174,8 +174,10 @@ abstract class ListLineTracker implements ILineTracker {
 	public final IRegion getLineInformation(int line) throws BadLocationException {
 		int lines= fLines.size();
 
-		if (line < 0 || line > lines)
-			throw new BadLocationException();
+		if (line > lines)
+			throw new BadLocationException("line > lines: " + line + " > " + lines);  //$NON-NLS-1$//$NON-NLS-2$
+		if (line < 0)
+			throw new BadLocationException("line < 0: " +line);  //$NON-NLS-1$
 
 		if (lines == 0)
 			return new Line(0, 0);
--- a/bundles/org.eclipse.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.text
-Bundle-Version: 3.14.0.qualifier
+Bundle-Version: 3.14.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
@@ -1075,8 +1075,12 @@ public abstract class AbstractDocument implements IDocument, IDocumentExtension,
 
 	@Override
 	public void replace(int pos, int length, String text, long modificationStamp) throws BadLocationException {
-		if ((0 > pos) || (0 > length) || (pos + length > getLength()))
-			throw new BadLocationException();
+		if (pos + length > getLength())
+			throw new BadLocationException("end > length: " + (pos + length) + " > " + getLength());  //$NON-NLS-1$//$NON-NLS-2$
+		if (length < 0)
+			throw new BadLocationException("length < 0: " +pos);  //$NON-NLS-1$
+		if (pos < 0)
+			throw new BadLocationException("pos < 0: " +pos);  //$NON-NLS-1$
 
 		DocumentEvent e= new DocumentEvent(this, pos, length, text);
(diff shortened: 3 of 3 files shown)


bundles/org.eclipse.text/src/org/eclipse/jface/text/ListLineTracker.java
```diff
@@ -174,8 +174,10 @@ public final IRegion getLineInformationOfOffset(int position) throws BadLocation
 	public final IRegion getLineInformation(int line) throws BadLocationException {
 		int lines= fLines.size();
 
-		if (line < 0 || line > lines)
-			throw new BadLocationException();
+		if (line > lines)
+			throw new BadLocationException("line > lines: " + line + " > " + lines);  //$NON-NLS-1$//$NON-NLS-2$
```
laeubi, line 178: In general I like this as good errors are essential, but the wording is highly confusing here and need one to look at the code what lines / line means.

So I would suggest something like this (also on theother places)
```suggestion
			throw new BadLocationException("The given line "+ line + "  is larger as the number of lines currently tracked (" + lines+")");  //$NON-NLS-1$//$NON-NLS-2$
```

Also using String.format will save you multiple NON_NLS and make it easier to find out how the message will look like.

bundles/org.eclipse.text/src/org/eclipse/jface/text/ListLineTracker.java
```diff
@@ -174,8 +174,10 @@ public final IRegion getLineInformationOfOffset(int position) throws BadLocation
...
+		if (line < 0)
+			throw new BadLocationException("line < 0: " +line);  //$NON-NLS-1$
```
laeubi, line 180: In such case I think something like
```suggestion
			throw new BadLocationException("Negative values are not allowed for parameter line..");  //$NON-NLS-1$
```

jukzi:
The idea & style was copy pasted from org.eclipse.jface.text.ListLineTracker.getLineNumberOfOffset(int). Its good to have the same style all over.
I don't see a value to have complete sentences in Error messages that should not happen anyway. Also they would make the logfiles even longer without additional information.
Such simple messages are also common in JDK, for example java.util.Objects.requireNonNullElseGet(T, Supplier<? extends T>)

If you require your suggested changes please put a -1 review.

laeubi:
> The idea & style was copy pasted from org.eclipse.jface.text.ListLineTracker.getLineNumberOfOffset(int). Its good to have the same style all over.

Its not good to copy bad style.

> I don't see a value to have complete sentences in Error messages that should not happen anyway. Also they would make the logfiles even longer without additional information.

I have given already example why I think this is better, if you read the logs you need to also read the source code to knwo what this is about

> Such simple messages are also common in JDK, for example java.util.Objects.requireNonNullElseGet(T, Supplier<? extends T>)

And they also supply variant to give a descriptive message or even throw a custom exception in such cases....

> If you require your suggested changes please put a -1 review.

I gave some hints how it can be imrpoved but I don't require it, but your intend was to improve things in case of failures so next time you see this message you might be happy to have better messages ... or not, at laest all this arguing why the cryptic one is better takes more time than simply chosse a good one.

jukzi:
My only desire here is to get the illegal argument **values** into the logs.  I have no plans to improve error messages that a user should nor see.

akurtakov:
That's ENOUGH!
I strongly SUGGEST that everyone learns to read everyone else's message as an effort to get a better software at the end!
This behavior of "everyone knows better than everyone else should stop RIGHT NOW" . 
Which one is better doesn't really matter here as the discussions tone degrade on daily basis to unacceptable one.

merks:
I sometimes get the feeling that folks are more willing to argue an hour over the tiniest nit than they are willing to spend five  minutes making the code arguably better.  There is something seriously wrong with that behavior and no matter how many times I keep asking for better communication, the poor communication just happens again and again.  Something needs to improve **immediately** or the PMC will step in and take action, and those are actions that people are very likely not to like.

So here I see @jukzi improving something; that's wonderful.  I see @laeubi comes with constructive suggestion for further improvements; that's helpful and also great teamwork.  But then it degrades.  Yes, @laeubi suggestion has grammar errors it and it omits to include the value itself in the one of the messages, but the intent of his suggestion is clear and one would be well advised to listen. 

I'm so fed up with poor communication, especially after yesterday where I had to deal with communication of this form:
> From past experience, "feel free to do as you think best" means go away, waste your own time and I'll bin the results.

where someone reads what I write and believes that I literally I mean the opposite of what I actually said.  It's impossible to communicate when ill intent is assumed.

So I know all too well that it's really a challenge to remain civil in the face of frustration, but it is what we must do.   I'll bet the mind space here sometimes more like "you nitpicked my last changes to death so I'm going to nitpick your latest changes to death", and even when that's not actually the intent, if one believes it to be the intent, one reacts as if the other person has ill intent.  But many of us are super picky and that can be super annoying, but it's also what makes the final results super good.

jukzi:
If somebody wants to further improve the error messages feel free to override this PR. 
I will not merge this PR myself in the current _social_ state.

merks:
@jukzi 

I would like to have a private one-on-one conversation with you.  I'm swamped today with SimRel, but the rest of the week is good. How best to arrange that? I think you know my gmail address.  I am also on IDE WG slack channel...

jukzi:
> I would like to have a private one-on-one conversation with you.

Appreciated. Just call +49 151 18233332 during office hours. Anyone else interested is invited too.

PR #1759, opened 2024-03-18
Clean-up WindowsDefenderConfigurator after subsequent check refinements

HannesWell:
Follow-up for https://github.com/eclipse-platform/eclipse.platform.ui/pull/1453. These clean-ups where hold back to reduce the risk for regressions late in the release cycle.
Some explanations also added in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1832 could also be added to the preference page.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.List;
 import java.util.Optional;
@@ -29,6 +30,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IProduct;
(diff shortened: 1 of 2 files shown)


jukzi:
Hi @HannesWell i just got aware that there is a platform and virusscanner independent possibility to test if a virusscanner is active for a certain directory/filetype:
```
Path path = Path.of("c:/temp/any.class"); // the path should match the directory and the fileextention to test
Files.writeString(path, "X5O!P%@AP[4\\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*");
Files.readAllBytes(path); // throws FileSystemException if any virus scanner is active 
```
probably it would be nice to also inform users of other virusscanners, that they have a problem.

HannesWell:
@BeckerWdf FYI I have slightly updated the text also incorporating some of your doc-changes where I saw fit. I also renamed 'Windows Defender' to 'Microsoft Defender' in texts and labels since it is now branded as such.




> Hi @HannesWell i just got aware that there is a platform and virusscanner independent possibility to test if a virusscanner is active for a certain directory/filetype:
> 
> ```
> Path path = Path.of("c:/temp/any.class"); // the path should match the directory and the fileextention to test
> Files.writeString(path, "X5O!P%@AP[4\\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*");
> Files.readAllBytes(path); // throws FileSystemException if any virus scanner is active 
> ```
> 
> probably it would be nice to also inform users of other virusscanners, that they have a problem.

That's interesting, thank you Jörg for that hint. There is even a [Wikipedia-article](https://en.wikipedia.org/wiki/EICAR_test_file) about it.
Nevertheless I'm not sure how to generalize this. From my experience in the past I can tell, that not all antivirus programs slow down Eclipse (that significantly). I had Bitdefender installed before working on this topic and never noticed such problems.
So it is probably highly depended on the exact software being used.
Furthermore there are also other ways on Windows obtain more general status information:
- `(Get-MpComputerStatus).AntivirusEnabled`
- `(Get-MpComputerStatus).AntispywareEnabled`
- `(Get-MpComputerStatus).BehaviorMonitorEnabled`

IIRC at least some also work for other AV software too.
But as said I don't think an exclusion is necessary in general and it would also be non trivial to implement for arbitrary AV products. Furthermore implementing a disabling logic for all would be practically impossible and just checking the directory can also be insufficient, because just like it is done for MS-Defender one can just exclude a processes by its full path and we can't use the suggested method to test that.

But if you think there is more to do, lets discuss that in a separate issue or PR. :)

PR #1765, opened 2024-03-19
Properly use selected string for replacements in FindReplaceDialog #1754

### LINKED ISSUE #1754: 2024-03: "Find/Replace"-Button has no effect anymore without manually shifting focus back to the editor
To reproduce:
- Type Ctrl+F
- Click the "Replace with"-field
- Click the "Find/Replace"-button
- Nothing happens until you click into the editor again

findreplace

HeikoKlare:
When opening the FindReplaceDialog with an active selection in a text viewer, the input field for the find string is filled with the content of that selection. In addition, the replace functionality of the dialog is activated. Currently, using the replace functionality does not work because it requires a find operation to be executed before, which is currently not done.

With this change, initializing the dialog with the current selection performs the missing find operation before executing a replace operation. The according functionality is refactored and a regression test is added.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1754

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -259,7 +259,7 @@ class FindReplaceDialog extends Dialog {
 		updateCombo(fReplaceField, replaceHistory.get());
 
 		// get find string
-		initFindStringFromSelection();
+		initFindString();
 
 		// set dialog position
 		if (fDialogPositionInit != null)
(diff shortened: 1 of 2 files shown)


HeikoKlare:
@Wittmaxi Can you have a look at this please?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
```diff
@@ -898,35 +898,58 @@ private void storeSettings() {
...
+	 * the newest search history entry.
+	 */
+	private void initFindString() {
+		if (!okToUse(fFindField)) {
+			return;
+		}
+
+		fFindField.removeModifyListener(fFindModifyListener);
+		if (!initFindStringFromSelection()) {
+			initFindStringFromHistory();
```
Wittmaxi, line 915: ~~This is introducing new behavior. Since we select the "history-find-string" by default, simply pressing any beginning letter of a new search word would just overwrite the entire string, meaning that the workflows are exactly the same - I believe that that is an alright change to make. I just wonder if that new behaviour could confuse a user who instinctively presses "backspace" instead of overwriting the word, thus adding one keypress to his user journey?~~

This is not introducing new behaviour and makes total sense - I overlooked the place in the old code where this exact behavior was already present.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
```diff
@@ -898,35 +898,58 @@ private void storeSettings() {
...
+		fFindField.setSelection(new Point(0, fFindField.getText().length()));
+		fFindField.addModifyListener(fFindModifyListener);
+	}
+
 	/**
 	 * Initializes the string to search for and the appropriate text in the Find
 	 * field based on the selection found in the action's target.
 	 */
-	private void initFindStringFromSelection() {
+	private boolean initFindStringFromSelection() {
```
Wittmaxi: I wonder if adding a return type to this function might not be giving too much responsibility to the function?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
```diff
@@ -898,35 +898,58 @@ private void storeSettings() {
...
+		} else {
+			fGlobalRadioButton.setSelection(false);
+			fSelectedRangeRadioButton.setSelection(true);
+		}
+		return true;
+	}
+
+	private void initFindStringFromHistory() {
+		if ("".equals(fFindField.getText())) { //$NON-NLS-1$
+			if (!findHistory.isEmpty())
```
Wittmaxi: In general, we wanted to always add braces even for single-line-scopes.
Wouldn't it make sense to introduce the braces for these blocks now, while we are already doing a refactoring?

Wittmaxi (commented):
I believe these changements make sense. I have a few questions left, but would not mind these changes being merged.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
HeikoKlare: Coupling the query logic with action logic in one method is not necessary here. I have cleaned the code by separating into two methods in 033648c.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
HeikoKlare: Done in 033648c.

HeikoKlare:
Thanks for checking the changes, @Wittmaxi! I have incorporated your improvement proposals.

PR #1766, opened 2024-03-19
Replace usage of actually internal SWT TypedListener

HannesWell:
Use the new API introduced in SWT via https://github.com/eclipse-platform/eclipse.platform.swt/pull/1112

This also fixes the new API-tools warnings about illegal instantiations of `TypedListener`.


HannesWell:
Without re-exported required bundles this would be a much simpler change. :/

PR #1775, opened 2024-03-25
Transfer primary data stack when current one is closed #1774

### LINKED ISSUE #1774: Outdated JavaDoc in FieldEditorPreferencePage

## Suggestion

The documentation of the FieldEditorPreferencePage class mentiones, that subclasses may overwrite the `createLayout()` method to customize the layout. However, the layout is created in the constructor and there is no method with that name in this class, the FieldEditor class or any of its parent classes.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/c40b7a177e4be437a644ba578ace8ed6d0cbaaf0/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/FieldEditorPreferencePage.java#L34-L36

Does anyone know what this is referring to? Because this segment goes back to [5fe99e68c2592c63f5ca9e4b8f3ca5554dddeab9](https://github.com/eclipse-platform/eclipse.platform.ui/blob/5fe99e68c2592c63f5ca9e4b8f3ca5554dddeab9/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/FieldEditorPreferencePage.java), i.e. the initial contribution.

If not, then I would propose to simply remove this part, to avoid any further confusion.

## Community

HeikoKlare:
The "primaryDataStack" is the main editor part stack used for placing newly opened editors. When the shared area containing the `primaryDataStack` is split into multiple stacks and the last editor in the `primaryDataStack`is closed, the `primaryDataStack`is removed from the application model so that the application model does not contain a `primaryDataStack`anymore.

The cleanup functionality that removes a part stack being not the last in the shared are is placed in the `CleanupAddon`. This change adds functionality to this addon that transfers the `primaryDataStack`property from the stack being removed to the first of the remaining part stacks. It also centralizes the functionality to create a primary data stack and to identify an element as such to avoid duplicated functionality.

Note that this chance does _not_ yet solve the problem of the onboarding information being lost within the same session of your workbench. It will only ensure that the `primaryDataStack` does not get lost, such that when restarting the product the onboarding information will be back. The issue of disappearing onboarding information within the same session has to be addressed by changing the way in which the onboarding information is instantiated. I will provide a separate PR for that.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1774

### How to test / reproduce

You can experience the change with the following scenario (also documented in #1774):
1. Start an Eclipse product, e.g., an SDK
image
2. Open two editors in the shared area. Then split the shared area placing the the two editors side by side via drag and drop.
image
3. Close the _left_ editor first, then close the remaining (the originally _right_) editor. You will see that the onboarding information is lost.
image
4. Restart the product. Without this chance, you will see that the onboarding information is still lost (like in the previous screenshot) whereas with this change the onboarding information is present again. This is a result of the `primaryDataStack` still being present. You can also investigate the `workbench.xmi` under `.metadata/.plugins7org.eclipse.e4.workbench` within your workspace. With this change, the `workbench.xmi` will still an entry like the following after executing the described scenario whereas without this chance the entry is missing:
```xml
    <sharedElements xsi:type="advanced:Area" xmi:id="_SYOZIOq3Ee6Ds-nHDao7XQ" elementId="org.eclipse.ui.editorss">
      <children xsi:type="basic:PartStack" xmi:id="_VvWtAeq3Ee6SwsvRjze8_w" elementId="org.eclipse.e4.primaryDataStack">
        <tags>org.eclipse.e4.primaryDataStack</tags>
        <tags>EditorStack</tags>
      </children>
    </sharedElements>
```


HeikoKlare:
Test is "only" failing because of Java 21 JVM being used. With a Java 17 JVM it is working fine. Just adding another test that is executed before the added one and simply initializes and disposes an E4Application also makes the test succeed. This also applies to some other test cases in the `org.eclipse.e4.ui.workbench.addons.swt` plug-in, which when executed standalone start to fail with a Java 21 JVM.

@merks maybe you are interested in taking a look at the proposed fix?

bundles/org.eclipse.e4.ui.workbench.addons.swt/src/org/eclipse/e4/ui/workbench/addons/cleanupaddon/CleanupAddon.java
```diff
@@ -381,4 +384,32 @@ private void subscribeRenderingChanged(
...
 		return modelService.isLastEditorStack(element);
 	}
+
+	private void transferPrimaryDataStackIfRemoved(MElementContainer<?> removedElement) {
+		if (!PRIMARY_DATA_STACK_ID.equals(removedElement.getElementId())) {
+			return;
+		}
+		MPartStack newPrimaryDataStack = findFirstPartStackToBeRendered(removedElement.getParent());
+		if (newPrimaryDataStack != null) {
+			newPrimaryDataStack.setElementId(removedElement.getElementId());
```
merks: I think it's slightly more obvious that you assign this as the PRIMARY_DATA_STACK_ID if you use that constant rather than call a method to return what you established above must be the value of that ID.

bundles/org.eclipse.e4.ui.workbench.addons.swt/src/org/eclipse/e4/ui/workbench/addons/cleanupaddon/CleanupAddon.java
```diff
@@ -381,4 +384,32 @@ private void subscribeRenderingChanged(
...
+			newPrimaryDataStack.getTags().addAll(removedElement.getTags());
```
merks: I wonder, are there specific tags that you expect to be transferred?  Is it ever possible that there might be tags you'd not want transferred?

bundles/org.eclipse.e4.ui.workbench.addons.swt/src/org/eclipse/e4/ui/workbench/addons/cleanupaddon/CleanupAddon.java
HeikoKlare: Makes sense. I have now completely moved the functionality for making a "primary data stack" out of a part stack and for identifying an element as such to a utility, which is used at all places that replicated the according functionality.

bundles/org.eclipse.e4.ui.workbench.addons.swt/src/org/eclipse/e4/ui/workbench/addons/cleanupaddon/CleanupAddon.java
HeikoKlare: I only expect the "EditorStack" and the "primary data stack" tags to be transferred. I think you are right that we should not simply transfer every kind of tag but only those related to the stack being the primary data stack. I have adapted the code accordingly with the introduced utility for adding the relevant tags as mentioned in the comment above (https://github.com/eclipse-platform/eclipse.platform.ui/pull/1775#discussion_r1542729945).

HeikoKlare:
If no one disagrees, I would like to merge this one together with #1778 for M1 to have some productive testing until 2024-06 release.

HeikoKlare:
Failing test is unrelated and documented in #1005.

Issue #1780, opened 2024-03-27
NavigationHistoryEntry and NavigationLocation do not take into account Editor IDs


Greglar4:
In the case where there are multiple different editors whose Editor Inputs are considered the same, undesirable behavior can happen when navigating between them using the go back/forwards Eclipse navigation actions. In particular, the display strings of the NavigationHistoryEntries in the dropdown menus of the go back/forwards buttons can change depending on which editor is currently open. Additionally, the incorrect editor may be opened when actually navigating to various NavigationHistoryEntries when compared to the actual ordered history of which editors were opened by the user.

For a specific example, in our use case we have an Xtext text editor, and a NatTable in a MultiPageEditor that operate on the same underlying file. They both have FileEditorInputs for their EditorInputs, and because they are editing the same underlying file, these EditorInputs are considered the same, though these two editors should be treated as different editors from a navigation history point of view.

This is expanded upon in this [discussion](https://github.com/eclipse-platform/eclipse.platform.ui/discussions/1753).

This is also related to the closed issue [#339](https://github.com/eclipse-platform/eclipse.platform.ui/issues/339), which was resolved by this [pull request](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1150).

A solution would be to also take into account the editor ids of the editor, as was done for the solution of issue #339. I will submit a pull request for this issue shortly.

PR #1781, opened 2024-03-27
Navigation - Issue #1780

Greglar4:
Updates the restoreLocation() method of the NavigationHistoryEntry class to match the editor ID as well as the editor input when opening the editor. Also updates the getEditorPart() method of the NavigationLocation class to return the IEditorPart based on the editor ID as well as the editor input, if both are available.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/NavigationLocation.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/NavigationLocation.java
@@ -25,6 +25,8 @@ public abstract class NavigationLocation implements INavigationLocation {
 
 	private IEditorInput input;
 
+	private String editorId;
+
 	/**
 	 * Constructs a NavigationLocation with its editor part.
 	 *
@@ -32,6 +34,7 @@ public abstract class NavigationLocation implements INavigationLocation {
 	 */
 	protected NavigationLocation(IEditorPart editorPart) {
 		this.page = editorPart.getSite().getPage();
+		this.editorId = editorPart.getSite().getId();
 		this.input = editorPart.getEditorInput();
 	}
 
@@ -43,13 +46,26 @@ public abstract class NavigationLocation implements INavigationLocation {
 	protected IEditorPart getEditorPart() {
 		if (input == null) {
 			return null;
+		} else if (editorId == null) {
+			return page.findEditor(input);
+		} else {
+			IEditorReference[] editorReferences = page.findEditors(input, editorId,
+					IWorkbenchPage.MATCH_ID | IWorkbenchPage.MATCH_INPUT);
+			if (editorReferences.length > 0) {
+				return editorReferences[0].getEditor(false);
+			}
+			return null;
 		}
-		return page.findEditor(input);
 	}
 
 	@Override
 	public Object getInput() {
 		return input;
+	}
+
+	@Override
+	public String getId() {
+		return editorId;
 	}
 
 	@Override
@@ -64,6 +80,11 @@ public abstract class NavigationLocation implements INavigationLocation {
 	@Override
 	public void setInput(Object input) {
 		this.input = (IEditorInput) input;
+	}
+
+	@Override
+	public void setId(String id) {
+		this.editorId = id;
 	}
 
 	/**
--- a/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/GoBackForwardsTest.java
+++ b/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/GoBackForwardsTest.java
@@ -1,0 +1,244 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2020 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Louis Detweiler <greg.detweiler42@gmail.com> - Issue #1780
+ *******************************************************************************/
+package org.eclipse.ui.tests.navigator;
+
+import java.io.IOException;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.INavigationLocation;
+import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.internal.NavigationHistoryAction;
+import org.eclipse.ui.intro.IIntroPart;
+import org.eclipse.ui.part.FileEditorInput;
+import org.eclipse.ui.tests.harness.util.EditorTestHelper;
+import org.eclipse.ui.tests.harness.util.FileTool;
+import org.eclipse.ui.tests.harness.util.FileUtil;
+import org.eclipse.ui.tests.harness.util.UITestCase;
+import org.eclipse.ui.texteditor.AbstractTextEditor;
+import org.eclipse.ui.texteditor.TextSelectionNavigationLocation;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * @since 3.3
+ *
+ */
+public class GoBackForwardsTest extends UITestCase {
+
+	private static final String TEST_NAME = "GoBackForwardsTest";
+
+	public GoBackForwardsTest() {
+		super(TEST_NAME);
+	}
+
+	private static final String PROJECT_NAME = "GoBackForwardsTestProject";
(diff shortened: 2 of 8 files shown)


BeckerWdf:
this PR relates to issue #1780

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/NavigationLocation.java
```diff
@@ -43,8 +46,16 @@ protected NavigationLocation(IEditorPart editorPart) {
 	protected IEditorPart getEditorPart() {
 		if (input == null) {
 			return null;
+		} else if (editorId == null) {
+			return page.findEditor(input);
+		} else {
+			IEditorReference[] editorReferences = page.findEditors(input, editorId,
+					IWorkbenchPage.MATCH_ID | IWorkbenchPage.MATCH_INPUT);
+			if (editorReferences.length > 0) {
+				return editorReferences[0].getEditor(false);
```
BeckerWdf, line 55: Is it correct to always take the first entry if multiple editors are returned?

BeckerWdf (commented):
can you also provide an automated test for this PR?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/NavigationLocation.java
merks: I need to review in detail, but I'm out for a while now, so I just make this quick observation:

Keep in mind that before this logic was changed, it always took the first editor with the same editor input, regardless of the editorId.   The only way to distinguish the editors here is by input + id, so there is no other way to determine which of multiple to choose, and in principle how could it even matter and how could that arbitrary choice be worse that what it's doing now, completely ignored the ID?  Also, I'm not even sure it's possible to create multiple editors for the same input + same id. (Or?)  I think this API is just general for the case that there is no ID matching.

merks (changes requested):
This generally looks correct and an improvement to me.  But, looking very carefully at how this is used throughout the framework, I think it's incomplete.

A new field editorId is added so that makes me wonder how is the existing input field used in the framework?

We see there is an input getter and we see it used here:

image

If I open a *.java file with a Java Editor, a Text Editor, and a Generic Text Editor, I do see this logic get called multiple times.  I'm not sure it does significant harm; they do all share the same underlying Document.   But it would seem more correct to match by ID.

We also see there is an input setter and we see it used here:

image

Here we really should set the ID to avoid loss of information.

So my sense is that we should add a getter and a setter for the editor ID in org.eclipse.ui.INavigationLocation and org.eclipse.ui.NavigationLocation, and we should use them in exactly the places we see the input getter and setter being used.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/NavigationLocation.java
Greglar4: I could provide an automated test for the navigation functionality. Where would be a good place for them to live? There doesn't see to be a test plugin for org.eclipse.ui.workbench, but even if there were I'm not sure if that would be a good place for them, as ideally I would have access to at least two different editors with different IDs but editor inputs which are considered equal.

merks: I see that org.eclipse.ui.tests.navigator depends on org.eclipse.ui.workbench so this is a possible candidate.  It even depends on org.eclipse.jdt.ui.  I'm not sure if two different text editors are a good test, but probably in this case/test there are three available for *.java.

@akurtakov @iloveeclipse 

Do you guys know if there is a more appropriate place to add tests for the navigation buttons as controlled by org.eclipse.ui.INavigationLocation?

Greglar4: It turns out that my install of Eclipse IDE for Eclipse Committers with the platformUi.setup file has several problems, maybe all stemming from the fact that the modular target platform fails because it requires 'org.eclipse.equinox.p2.iu; org.eclipse.license.feature.group 0.0.0' but it could not be found. But in the end, there are errors all throughout my workspace and I can't seem to run any tests. I don't really have the time to figure out how to fix the install, and it would probably be faster for someone with a working environment to supply the tests anyways, so maybe someone else could supply the desired tests?

merks: I'm not sure how to prevent people from using the the Committers package instead of the Eclipse SDK product as is used by this:

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment

When I tired it though, the targlet task worked well for me and has the license feature's update site included:

image

Unfortunately, unless you install using the 4.32 SDK, the API baseline for SWT is broken and result in errors in the workspace:

image

It's better to use the SDK product:

image

Setting up a workspace with the Oomph setup takes literally only a few minutes.  If there are problem, you should report them so they can be addressed.

Note that most everyone on the project works on their personal priorities or on the priorities of the people who fund their work.   So no one is likely to have time to write tests for you.  I can help you help yourself because the IDE WG considerations enabling contribution a significant priority...

Greglar4: Thanks, that worked! I can now run tests. I should be able to write a test for this PR. Crazy how things work when you follow the instructions 🤣

Greglar4: I have written a test for this issue, and confirmed it runs and passes locally running mvn clean verify. Additionally, I ensured that the test fails without the functional code changes of this pull request.

merks:
I see that the new test failed but only on GitHub action on Ubuntu:

image

I need to find time to do a review, but I'm a little swamped with 4.32/2024-06 M1 activities...

Greglar4:
Yeah I saw that too, trying to look into it.

Greglar4:
I ran the test on my local linux install and it passed, so I'm not really sure how to proceed.

merks:
I haven’t forgotten this issue. I’ve just been too busy.  I wonder it there could be timing issues with the test?  I’ll have to run them locally too.

merks:
I see these errors in the IDE when I fetch this:

image

So this should be changed like this:

image

That version should be used
```
	 * @since 3.132
	 */
	String getId();
```

In terms of the previous test failure I suspect it is somehow timing related.  Specifically if I comment out this line it fails:

image

That method does this:

image

So it stops when there are no more events, but maybe on some slow system, this stops early before all the events have happened.  I don't really have a clue.  Maybe a hack like this would help:

image

Or using a listener and using a condition with org.eclipse.ui.tests.harness.util.UITestCase.processEventsUntil(Condition, long) that tests whether the new editor activation as occurred already.

Greglar4:
I initially tried to change all calls to processEvents() with processEventsUntil(condition, 500) with a condition that the active editor had the correct id/selection, however this caused the tests to fail even locally, because these conditions were satisfied on first check, so the while(display.readAndDispatch()) loop was never entered. I then tried processEventsUntil(null, timeout) which basically just calls display.readAndDispatch() for as long as the passed in timeout. I tried timeouts of 500 and 1000 but these still failed on at least one of the OSes. I just now changed it to processEventsUntil(null, 10000) just to see if the failures were timing related by putting in a ridiculously high wait time, and it seems that it has succeeded.

Greglar4:
I created a condition that checks if the editor navigation was added to the navigation history. That seems to have worked the second time, but the first time it timed out when I had the timeout set to 10 seconds. So I increased the timeout, and after that each test took under four seconds total...

tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/GoBackForwardsTest.java
```diff
@@ -0,0 +1,252 @@
...
+		Condition textEditorSelection = currentNavigationHistoryLocationCondition(TEXT_EDITOR_ID, true);
+
+		FileEditorInput editorInput = new FileEditorInput(file);
+
+		openJavaEditor(editorInput);
+
+		if (!processEventsUntil(javaEditorNoSelection, 30000)) {
+			fail("Timeout during navigation.");
+		}
+		;
```
merks: Just one small nit.  Could you please get rid of all these lines with just a semicolon on them?

merks (approved):
It looks good now.

I just updated the branch and will merge after the build completes.

merks:
Oops, my update branch did a merge.  I'll squash that at the end.

merks:
The new test failed like this:

image

Greglar4:
At this point I'm not sure what to do, the code changes fix the problem and the tests pass locally. We could theoretically increase the timeout of the processEventsUntil methods, though I don't see why even on a slow machine it would take more than 30 seconds to open an editor and have the navigation history entry added to the navigation history. Or perhaps we could try adding a junit rule that retries the test several times, as I have seen all of the OSes both fail and pass at least once while developing this test. Or maybe we could just merge the code changes and open an issue for writing a test for navigation.

merks:
I think the new logic is definitely correct and the test failures are kind of inexplicable.  I had some ideas to try for the tests but I’m out of the office until later in the week so it will take some time.

merks:
Superseded by https://github.com/eclipse-platform/eclipse.platform.ui/pull/1821

PR #1787, opened 2024-03-29
Add support for Windows on Arm64 (WoA) platform build.

chirontt:
The following new Equinox plug-ins for WoA are added to the existing `org.eclipse.e4.rcp` feature:
```
org.eclipse.equinox.launcher.win32.win32.aarch64
org.eclipse.equinox.swt.win32.win32.aarch64
```

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -218,6 +218,13 @@
          version="0.0.0"/>
 
    <plugin
+         id="org.eclipse.equinox.launcher.win32.win32.aarch64"
+         os="win32"
+         ws="win32"
+         arch="aarch64"
+         version="0.0.0"/>
+
+   <plugin
          id="org.eclipse.equinox.launcher.win32.win32.x86_64"
          os="win32"
          ws="win32"
@@ -226,6 +233,13 @@
 
    <plugin
          id="org.eclipse.swt"
+         version="0.0.0"/>
+
+   <plugin
+         id="org.eclipse.swt.win32.win32.aarch64"
+         os="win32"
+         ws="win32"
+         arch="aarch64"
          version="0.0.0"/>
 
    <plugin
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,7 @@
                     <plugin id="org.eclipse.equinox.launcher.gtk.linux.ppc64le" />
                     <plugin id="org.eclipse.equinox.launcher.gtk.linux.aarch64" />
                     <plugin id="org.eclipse.equinox.launcher.gtk.linux.x86_64" />
+                    <plugin id="org.eclipse.equinox.launcher.win32.win32.aarch64" />
                     <plugin id="org.eclipse.equinox.launcher.win32.win32.x86_64" />
                   </excludes>
                 </configuration>


HannesWell:
Thanks @chirontt for this. In general it looks good.

But we have to await https://github.com/eclipse-equinox/equinox/pull/559 before this can succeed, I therefore converted this into a draft.

Btw. I think the following list of exclusions should also be updated to contain the new win32.aarch64 fragment:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/3f00c8b2d69ac6f06a803c046a61a9a104ba5790/pom.xml#L83-L92

In case you wonder how one can find all these occasions: You can just search for example for `org.eclipse.equinox.launcher.win32` and check all hits, like it was done in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577#issuecomment-1250597136. :)

> Also, note that there are 2 `pom.xml` files added back to their respective bundle folders, even though they were removed in a previous commit ([b060bc4](https://github.com/eclipse-platform/eclipse.platform.ui/commit/b060bc4176e18101819929e39304578d390abdec)) long ago to go "pomless":

My best guess at the moment is that https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917 is missing.
By it is probably only required to run the build on win32.aarch64. This PRs verification build should succeed without the new poms once https://github.com/eclipse-equinox/equinox/pull/559 is completed.
If you want to test it now, you can checkout https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917 locally (maybe update it) and let the pom.xml in the root of your platform.ui clone point to your checked-out `eclipse-platform-parent/pom.xml`.

chirontt:
> But we have to await [eclipse-equinox/equinox#559](https://github.com/eclipse-equinox/equinox/pull/559) before this can succeed, I therefore converted this into a draft.

Yes, making it a `draft` PR is a good idea. I will add draft PRs to the remaining repos like `eclipse.platform` and `eclipse.platform.releng.aggregator`. Note that for the aggregator repo, I will re-use your existing eclipse-platform/eclipse.platform.releng.aggregator#1917 and add more files to it, if you don't mind.
 
> Btw. I think the following list of exclusions should also be updated to contain the new win32.aarch64 fragment:
> 
> https://github.com/eclipse-platform/eclipse.platform.ui/blob/3f00c8b2d69ac6f06a803c046a61a9a104ba5790/pom.xml#L83-L92

Already done in the PR.

> > Also, note that there are 2 `pom.xml` files added back to their respective bundle folders, even though they were removed in a previous commit ([b060bc4](https://github.com/eclipse-platform/eclipse.platform.ui/commit/b060bc4176e18101819929e39304578d390abdec)) long ago to go "pomless":
> 
> My best guess at the moment is that [eclipse-platform/eclipse.platform.releng.aggregator#1917](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917) is missing.

Nope, I've got the full aggregator for WoA in my [fork](https://github.com/chirontt/eclipse.platform.releng.aggregator), but it still fails to build without these 2 `pom.xml` files.

But I think I've found the culprit: the SWT `win32.aarch64` fragment is missing in some `p2.inf` files; here's the PR for it: eclipse-platform/eclipse.platform.swt#1151, please review it. It's surprising to me that a missing entry in some `p2.inf` file in one repo can severely affect the build of another repo (this one.)

I will remove the 2 `pom.xml` files from this PR as they are now redundant.

HannesWell:
> > But we have to await [eclipse-equinox/equinox#559](https://github.com/eclipse-equinox/equinox/pull/559) before this can succeed, I therefore converted this into a draft.
> 
> Yes, making it a `draft` PR is a good idea. I will add draft PRs to the remaining repos like `eclipse.platform` and `eclipse.platform.releng.aggregator`. Note that for the aggregator repo, I will re-use your existing [eclipse-platform/eclipse.platform.releng.aggregator#1917](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917) and add more files to it, if you don't mind.

I don't mind at all, but only maintainer/committers can change existing PRs from others (and if I'm not mistaken, you aren't a commiter?). But you can mention the files you want to change and I can integrate that and add you as co-author.

> > > Also, note that there are 2 `pom.xml` files added back to their respective bundle folders, even though they were removed in a previous commit ([b060bc4](https://github.com/eclipse-platform/eclipse.platform.ui/commit/b060bc4176e18101819929e39304578d390abdec)) long ago to go "pomless":
> > 
> > 
> > My best guess at the moment is that [eclipse-platform/eclipse.platform.releng.aggregator#1917](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917) is missing.
> 
> Nope, I've got the full aggregator for WoA in my [fork](https://github.com/chirontt/eclipse.platform.releng.aggregator), but it still fails to build without these 2 `pom.xml` files.
> 
> But I think I've found the culprit: the SWT `win32.aarch64` fragment is missing in some `p2.inf` files; here's the PR for it: [eclipse-platform/eclipse.platform.swt#1151](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1151), please review it. It's surprising to me that a missing entry in some `p2.inf` file in one repo can severely affect the build of another repo (this one.)
> 

That's a good catch and the right fix. This configuration of the `o.e.swt` bundle and its fragment is indeed complicated.
The bundle `o.e.swt` is in fact empty and all class files and native binaries are actually contained in the native fragments of each supported platform. And while a fragment requires its host (in this case `o.e.swt`) a host bundle does not automatically require its fragments. So by default requiring `org.eclipse.swt` would not pull the corresponding native fragment. And this is what you've experienced. The `p2.inf` you just changed and the `org.eclipse.swt.fragments.localbuild` mitigate that on a p2-level and add a requirement from the host to the fragment of the current platform. I'm currently also working on a solution on the OSGi level, which makes these p2.inf files obsolete. But in general its not that easy.

chirontt:
> I don't mind at all, but only maintainer/committers can change existing PRs from others (and if I'm not mistaken, you aren't a commiter?). But you can mention the files you want to change and I can integrate that and add you as co-author.

No, I'm not a committer, so I will create a branch in that repo so that you can merge it with your PR.

HannesWell:
As soon as https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917 is submitted this, should be ready too.

HannesWell (approved):
Now that https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917 is completed this can finally be submitted.

Thanks for this one too.

PR #1788, opened 2024-03-30
Add existence check (file protocol) in URLImageDescriptor.getFilePath()

HeikoKlare:
The URLImageDescriptor.getFilePath() method currently returns a path for a given URL using the file protocol even if the file does not exist. This conflicts with the method's contract and breaks consumers (such as the scaled image data retrieval in SWT's DPIUtil).

With this change, the method properly checks for the existence of a file at the given location and returns null in case it does not exist, according to the method's contract. A test case is added to demonstrate the bug and detect regressions.


BeckerWdf (approved):

HeikoKlare:
Failing tests are unrelated and already documented: #1005.

PR #1789, opened 2024-04-02
Add org.eclipse.swt.win32.win32.aarch to org.eclipse.e4.rcp feature

merks:


merks:
@HannesWell 

I did this because the I-Builds don't contains the new SWT Windows aarch fragment so this build

https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/CBIaggregator/

is failing with

>08:22:28 Build failed! Exception was java.lang.IllegalStateException: Unresolved requirement iu org.eclipse.swt.win32.win32.aarch64 [3.126.0.v20240401-0817]
>08:22:29 java.lang.IllegalStateException: Unresolved requirement iu org.eclipse.swt.win32.win32.aarch64 [3.126.0.v20240401-0817]

That's because the host requires it:

image

merks:
The builds fail like this:
```
07:57:24.598 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:4.0.7:package-feature (default-package-feature) on project org.eclipse.e4.rcp: Execution default-package-feature of goal org.eclipse.tycho:tycho-packaging-plugin:4.0.7:package-feature failed: eclipse-plugin artifact with ID "org.eclipse.swt.win32.win32.aarch" and version matching "0.0.0" was not found in the target platform -> [Help 1]
07:57:24.598 [ERROR]
```

@HannesWell 

What's the outlook for the build actually containing the IU required by SWT?

https://github.com/eclipse-platform/eclipse.platform.swt/blob/6bcc2674383e8292df69da42a810fc99f71bfdac/bundles/org.eclipse.swt/META-INF/p2.inf#L38-L41

The maven aggregation will not work with the above requirement until the fragment is available in the 4.32 I-builds.

If the runway is long, we could make the requirement optional (and greedy which is the default), until the build does publish that fragment...

HannesWell:
This is also prepared in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1787, but the intention was to put this on hold until we have a equinox.launcher for win32.aarch64 and can prepare an I-build in order to have the entire SDK (close to be) tested before we publish it.

But if this is required to pass the aggregator I suggest to go forward and continue with this one, #1787 is a super-set of of this.

> What's the outlook for the build actually containing the IU required by SWT?

That totally depends on when this PR (or #1787) is submitted. :)
The o.e.swt.win32.aarch64 fragment is ready, but the e4.rcp feature is the only feature containing the SWT native fragments and I assume the p2-inf requirement is only considered if the win32.aarch64 environment is activated (see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917).

Since we have this chicken-egg problem, the verification build cannot pass So this PR cannot. I suggest to either submit this short before the next I-build starts. Either at midnight or you start one by yourself after you have submitted this.

HannesWell:
OK, too bad: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1957 demonstrate that this change fails the build.
Thinking about it more it is probably because due to the not active win32-aarch64 environment the swt fragment for that env does not resolve and is therefore considered absent. But activating it is not possible as explained before.

> If the runway is long, we could make the requirement optional (and greedy which is the default), until the build does publish that fragment...

So I think we have to go that way.
@merks please have a look at https://github.com/eclipse-platform/eclipse.platform.swt/pull/1158

features/org.eclipse.e4.rcp/feature.xml
```diff
@@ -228,6 +228,13 @@
          id="org.eclipse.swt"
          version="0.0.0"/>
 
+   <plugin
+         id="org.eclipse.swt.win32.win32.aarch"
```
HannesWell, line 232: The ID misses the `64` at the end. 
```suggestion
         id="org.eclipse.swt.win32.win32.aarch64"
```

HannesWell:
Lets close this in favor of the older super-set https://github.com/eclipse-platform/eclipse.platform.ui/pull/1787 that should be ready for submission soon.

PR #1795, opened 2024-04-03
Support plugin_customization.ini for e4.ui.workbench.renderer.swt prefs

### LINKED ISSUE #1794: Predefining settings of org.eclipse.e4.ui.workbench.renderer.swt in plugin_customization.ini not possible
## Steps to reproduce

1. Create a local text file named 'plugin_customization.ini'
2. Add following line: `org.eclipse.e4.ui.workbench.renderers.swt/SHOW_FULL_TEXT_FOR_VIEW_TABS=true`
3. Add following line as program argument to you 'eclipse.ini'
```
-pluginCustomization 
<path/to/your/plugin_customization.ini>
```
4. Start Eclipse and check, if preference 'General > Appearance > Tab icons and titles in view areas' is set
=> Setting is not set although it was specified in the 'plugin_customization.ini' 

As comparison the following parameter can be used:
`org.eclipse.debug.ui/org.eclipse.debug.ui.save_dirty_editors_before_launch=never `
This will set the preference 'Debug > Launching > Save required dirty editors before launching' to 'Never' (Default: 'Prompt')

## Comments

It looks like this is related to how the preferences are read.
Currently the `InstanceScope` for the `org.eclipse.e4.ui.workbench.renderer.swt` preferences is queried explicitly. Preferences set in the 'plugin_customization.ini', which end up in the `DefaultScope`, were never respected.

The use of `org.eclipse.ui.preferences.ScopedPreferenceStore`, which takes the `DefaultScope` into account, is used elsewhere, but lies in a layer above `org.eclipse.e4.ui.workbench.renderer.swt` so cannot be used here. 

I'll provide a pull request which uses `org.eclipse.core.runtime.preferences.IPreferencesService` instead.

soethoff:
Before this change, only the InstanceScope was queried explicitly.

Therefore, any preferences set in 'plugin_customization.ini', which end up in the `DefaultScope`, were never respected.

Ideally, we would use `org.eclipse.ui.preferences.ScopedPreferenceStore`, but that lies in a layer above us, so cannot be used.

Instead, we use the
`org.eclipse.core.runtime.preferences.IPreferencesService` when reading preferences. This searches all relevant scopes in the proper order.

When writing preferences, we write explicitly into `InstanceScope`.

When performing 'Restore Defaults' in the preference page, we take the `DefaultScope` into account.

Fixes #1794.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -19,6 +19,7 @@ package org.eclipse.e4.ui.workbench.renderers.swt;
 
 import java.lang.reflect.Field;
 import java.util.Objects;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.PreferenceChangeEvent;
@@ -45,6 +46,11 @@ import org.eclipse.swt.graphics.Rectangle;
 
 @SuppressWarnings("restriction")
 public class CTabRendering extends CTabFolderRenderer implements ICTabRendering, IPreferenceChangeListener {
+
+	/**
+	 * The preference qualifier.
+	 */
+	private static final String PREF_QUALIFIER_ECLIPSE_E4_UI_WORKBENCH_RENDERERS_SWT = "org.eclipse.e4.ui.workbench.renderers.swt"; //$NON-NLS-1$
 
 	/**
 	 * A named preference for setting CTabFolder's to be rendered with rounded
@@ -169,7 +175,8 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 		super(parent);
 		parentWrapper = new CTabFolderWrapper(parent);
 
-		IEclipsePreferences preferences = getSwtRendererPreferences();
+		IEclipsePreferences preferences = InstanceScope.INSTANCE
+				.getNode(PREF_QUALIFIER_ECLIPSE_E4_UI_WORKBENCH_RENDERERS_SWT);
 		preferences.addPreferenceChangeListener(this);
 		parent.addDisposeListener(e -> preferences.removePreferenceChangeListener(this));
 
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -1283,8 +1290,9 @@ public void setSelectedTabHighlightTop(boolean drawTabHiglightOnTop) {
 	}
 
 	private void cornerRadiusPreferenceChanged() {
-		IEclipsePreferences preferences = getSwtRendererPreferences();
-		boolean useRound = preferences.getBoolean(USE_ROUND_TABS, USE_ROUND_TABS_DEFAULT);
+		boolean useRound = Platform.getPreferencesService().getBoolean(
```
iloveeclipse: Change looks good.

May I ask for a shortcut method to make the code more readable, to avoid `Platform.getPreferencesService().getBoolean(
				PREF_QUALIFIER_ECLIPSE_E4_UI_WORKBENCH_RENDERERS_SWT` be repeated again and again here and in the preference page?
Something like `getSwtRenderersOption(key, deafultvalue)` ?

soethoff: Good point, thanks! 
Extracted the repeating calls as a new method -> Done

BeckerWdf:
How can I test this with a Run Configuration? So I mean I want to start a target IDE from my workspace (that includes this PR). How do I set the preference value in the run configuration?

soethoff:
> How can I test this with a Run Configuration? So I mean I want to start a target IDE from my workspace (that includes this PR). How do I set the preference value in the run configuration?

Create a 'plugin_customization.ini' and add one or more settings you want to set as new default, e.g.:  `org.eclipse.e4.ui.workbench.renderers.swt/SHOW_FULL_TEXT_FOR_VIEW_TABS=true`

In your run configuration add the created 'plugin_customization.ini' as 'Program argument':
```
-pluginCustomization 
<path/to/your/plugin_customization.ini>
```
Note: The option 'Clear the configuration area before launching' should be enabled

BeckerWdf:
> > How can I test this with a Run Configuration? So I mean I want to start a target IDE from my workspace (that includes this PR). How do I set the preference value in the run configuration?
> 
> Create a 'plugin_customization.ini' and add one or more settings you want to set as new default, e.g.: `org.eclipse.e4.ui.workbench.renderers.swt/SHOW_FULL_TEXT_FOR_VIEW_TABS=true`
> 
> In your run configuration add the created 'plugin_customization.ini' as 'Program argument':
> 
> ```
> -pluginCustomization 
> <path/to/your/plugin_customization.ini>
> ```
> 
> Note: The option 'Clear the configuration area before launching' should be enabled

That worked. Thanks. It also worked without the "Clear" option.
From my point of view this PR looks good.

BeckerWdf:
Failed test are know and already reported:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1807
https://github.com/eclipse-platform/eclipse.platform.ui/issues/195

fedejeanne:
> Failed test are know and already reported: #275 #1427 #195

1427 was not correct, the stack traces are different. I edited the comment to avoid confusions

PR #1797, opened 2024-04-04
Add object factories to ISelection interface

laeubi:
Currently creating a IStructuredSelection requires to reference a concrete implementing class, also handling null is explicitly required and only List or Array are supported.

This adds some new object factory methods similar to what users are used to with the Optional.of(...) / List.of(...) and similar and adds some usages of the new methods.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -11,9 +11,11 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Christoph Läubrich - add support for stream() method
+ *                        - add support for factory methods
  *******************************************************************************/
 package org.eclipse.jface.viewers;
 
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
```diff
@@ -71,4 +73,59 @@ public interface IStructuredSelection extends ISelection, Iterable {
...
+
+	/**
+	 * Created a new {@link IStructuredSelection} containing the given object
+	 *
+	 * @param objects the object to contain in the selection
+	 * @return a {@link IStructuredSelection} containing the given elements
+	 */
+	static IStructuredSelection of(Object... objects) {
+		return new StructuredSelection(List.of(objects));
+	}
```
mickaelistria: With jshell, for this code
```
String m(Object o) { return "Single obj";}
String m(Object... o) { return "Multiple objs";}
m(new String[] {})
```
I get this warning:
```
|  Warning:
|  non-varargs call of varargs method with inexact argument type for last parameter;
|    cast to java.lang.Object for a varargs call
|    cast to java.lang.Object[] for a non-varargs call and to suppress this warning
|  m(new String[] {})
|    ^-------------^
$3 ==> "Multiple objs"
```

Isn't there a risk it surfaces in Platform as well for this API?
I fear we might have cases of the selection being made of a single array object which is not to be expanded as object. As we have a constructor for collections already, wouldn't it be safer to skip the varargs constructor?

mickaelistria (commented):
I don't see much added-value here, and while I understand this is an intent to make things simpler, I'm afraid it can introduce some confusion about how to handle `null` or array objects or whether other types are permitted, while the current API makes it very explicit what to expect.
This also gives the impression to end-users that the IStructuredSelection is the only entry point to build a structured selection, while looking at the class hierarchy shows multiple implementations existing (eg TreeSelection), and encourages anyone who needs it to build a new implementation.
So it cannot be a 1-1 replacement and it's usually not more efficient to use it over the existing constructs which give more transparency about what is going to happen and what is possible.

laeubi:
> I'm afraid it can introduce some confusion about how to handle null or array objects or whether other types are permitted, while the current API makes it very explicit what to expect

Why should it be confusing? We can improve the javadoc if there is any uncertainty.

> This also gives the impression to end-users that the IStructuredSelection is the only entry point to build a structured selection while looking at the class hierarchy shows multiple implementations existing (eg TreeSelection), and encourages anyone who needs it to build a new implementation.

I don't think a *user* is encouraged to build an own one and effectively  `TreeSelection` is the only one in whole platform, not even in a testcase there is a single custom implementation so it seems there is rarely a need to implement an own, so I really don't se a problem here, e.g. there are numerous Map/List implementations in the JDK still Map.of(...) are of use for the common case and the same is here with the IStructuredSelection.

> So it cannot be a 1-1 replacement and it's usually not more efficient to use it over the existing constructs which give more transparency about what is going to happen and what is possible.

It is a 1:1 replacement for the 90% case (2156 references to StructuredSelection versus 229 to TreeSelection versus 0 other implementations), still users are not limited to anything.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
laeubi: I think one can simply rename the method to something like ofElements(...) then.

mickaelistria:
Referencing concrete implementation is not always a bad thing. Or are you aware of a particular case where it is an issue here?
I think this change introduce no functional nor technical benefits over the current usage of the StructuredSelection constructors, and brings some (minor) concerns I mentioned earlier. Overall I don't see it as a positive and profitable addition in Platform. I may be missing some rationale to support it.

laeubi:
The benefit is that I (as a user) compared to using the (only) implementation

- I never care about the underlying implementation, but need to remember it (well its rather trivial here but ...) and it easily let me depend on the impl (e.g. if assigned to a local variable JDT always suggest the concrete type never the interface).
- It misses some convenient ways (e.g. it only allow List as the only collection Type) to construct such object, it can not handle null (so one always has to add checks if in doubt) and that's something that can't change without breaking the existing contracts (and getting error because `null` is ambiguous)
- On many places factories on interfaces are common and programmers get used to it (List.of() / Path.of() / and so on..)
- On the other hand this neither limits nor breaks anything if even we one time invent a "better" StructuredSelection2 Impl I not need to change anything in my code (if I use the factory).

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/IStructuredSelection.java
laeubi: This is fixed now

laeubi:
This also makes some trivial performance / memory optimizations possible, e.g. for empty arrays/collections this always return StructuredSelection.EMPTY instead of creating a new object + a copy of the array.

mickaelistria (changes requested):
For the reason mentioned above, more importantly the one that tells that ISelection is a publicly extensible type and not just a facade and that IMO factories on interfaces don't fit in that case as those lead users to false impressions of a single constrained entry point, I cannot give a positive review here.
I suggest we ask project leads to take a decision here.

laeubi:
@eclipse-platform/eclipse-platform-project-leads can you please take actions?

PR #1800, opened 2024-04-08
Use optionally computed content type from CompareConfiguration

### LINKED ISSUE #1747: [RFE] Add support for "editor" content types for compare editor / contentViewers extension
Moved from https://github.com/eclipse/tm4e/issues/730.

Currently textmate (and generic editor) based editors seem not support syntax highlighting in compare editor.

Example: in 4.31 Eclipse, set "Generic Text Editor" as default for XML content type, select two `.project` files in Package Explorer and via right click run "Compare With... -> Each Other" command.

That would show both parts of the editor without syntax highlighting:

image

That is of course not working for any other content type with textmate syntax support, XML here was chosen as it is simple to reproduce.

It would be nice if the syntax highlighting for all supported content types of generic editor also would be supported in compare editor.

During the discussion of this request, Mickael mentioned that:

> Enabling (some) generic editor features in the compare viewer has to be declared per content-type, ideally by the same bundle as the one that defines the content-type association with generic editor. Eg from Wild Web Developer https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/13c104d642e17156c7308ce43a78ccb04edcb55e/org.eclipse.wildwebdeveloper/plugin.xml#L1273C4-L1286C16
> 
> ```
> <extension
>          point="org.eclipse.compare.contentViewers">
>       <contentTypeBinding
>             contentTypeId="org.eclipse.wildwebdeveloper.parent"
>             contentViewerId="org.eclipse.ui.genericeditor.compareViewer">
>       </contentTypeBinding>
>    </extension>
>    <extension
>          point="org.eclipse.compare.contentMergeViewers">
>       <contentTypeBinding
>             contentMergeViewerId="org.eclipse.ui.genericeditor.compareViewer"
>             contentTypeId="org.eclipse.wildwebdeveloper.parent">
>       </contentTypeBinding>
>    </extension>
> ```

and also:

> Note that the generic editor isn't referenced directly here, it's a separate id/object that is assigned to the compare viewer, and there is no declarative mapping between this object and the underlying text editor.
> So it's not trivial to infer the proper compare viewer from the editor association.
> But if you think it's doable, that would indeed be an interesting improvement for Platform.

The problem we have is that editor associations could be manually updated by user, but contentViewers/contentMergeViewers cannot, even if that would technically work (like in generic editor case), because they share same underlying implementation.

So my proposal would be to extend contentViewers/contentMergeViewers extension point and add an attribute to consider "linked" editor content type associations. The code could check if contentViewers/contentMergeViewers contribution refers to such "linked" editor id in extension, and check if editor supports other content types. In case there are more content types supported by same "linked" editor, these would be automatically considered as valid contentTypeBinding for the contentViewers/contentMergeViewers.

This shouldn't be that complicated. From compatibility point of view, since the extra extension attribute would be not set by default, none of existing code would be affected. For the generic text editor that would mean to mark its contentViewers/contentMergeViewers contributions "linked" to "org.eclipse.ui.genericeditor.GenericEditor" and that's it. Any "specialized" editor versions of generic editor (with dedicated editor id) shouldn't be affected either.

Once I have some time, I would try to provide a PR for that idea, but any comments / contributions welcome.

iloveeclipse:
Detected content type id could be retrieved now from the CompareConfiguration with the CONTENT_TYPE key (see https://github.com/eclipse-platform/eclipse.platform/pull/1294).

Use this value (if available) as fallbackContentTypes in ExtensionBasedTextViewerConfiguration.

The allows GenericEditorMergeViewer to know for which content type was it actually created and so use that in cases where the content type can't be derived from the input (like data from git revisions that don't have regular files or buffers associated to the viewer).

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1747

### DIFF
--- a/bundles/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
@@ -15,7 +15,7 @@ Require-Bundle: org.eclipse.ui.workbench.texteditor;bundle-version="3.10.0",
  org.eclipse.ui.ide;bundle-version="3.12.0",
  org.eclipse.core.resources;bundle-version="3.11.0",
  org.eclipse.core.expressions;bundle-version="3.6.0",
- org.eclipse.compare;resolution:=optional
+ org.eclipse.compare;bundle-version="3.11.0";resolution:=optional
 Export-Package: org.eclipse.ui.internal.genericeditor;x-internal:=true,
  org.eclipse.ui.internal.genericeditor.hover;x-internal:=true,
  org.eclipse.ui.internal.genericeditor.markers;x-internal:=true,
--- a/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/ExtensionBasedTextViewerConfiguration.java
+++ b/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/ExtensionBasedTextViewerConfiguration.java
@@ -19,6 +19,9 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.genericeditor;
 
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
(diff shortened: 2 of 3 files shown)


iloveeclipse:
Note: compilation will fail until we have an SDK build with https://github.com/eclipse-platform/eclipse.platform/pull/1294 change merged.

Issue #1583, opened 2024-01-29
Silently erroring VirtualLazyTreeViewerTest and VirtualTableViewerTest


HeikoKlare:
## Problem Description
The test cases for virtual tree and table viewers are flaky on Mac, as already identified in [bug 347491](https://bugs.eclipse.org/bugs/show_bug.cgi?id=347491), because data is not reliably provided for the tree/table items. So far, this has been addressed by different means:
* conditional disablements for some of the tests (using `ViewerTestCase#disableTestsBug347491`)
* additional update mechanisms in some test cases: some of them resulting in tests that seem to work (such as `VirtualTableViewerTest#testSorter()`) while others using this mechanism still fail randomly (such as `VirtualTableViewerTest#testRenameWithSorter`, see #1005)
* accepting that a test case fails randomly by simply returning (successfully) when the viewer is not in a proper state

The last measure can, for example, be seen in `VirtualTableViewerTest#testFilter()`. In case an update of the table does work, it simply returns, so the test succeeds.
https://github.com/eclipse-platform/eclipse.platform.ui/blob/ce26ac01848324629973461f9906e0f6431d8d60/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/VirtualTableViewerTest.java#L135-L137
This "successful abortion" is logged an can actually been seen in different builds, e.g. https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/7692524821/job/20959564152. And this is nothing else than a silent test error.
```
SWT.SetData is not received. Cancelled test testExpandToLevel
SWT.SetData is not received. Cancelled test testRenameWithFilter
SWT.SetData is not received. Cancelled test testFilter
SWT.SetData is not received. Cancelled test testSetFilters
```

Even with all these mechanisms to make tests work or only fail silently but not produce test failures, there are still tests that randomly fail, like can be seen in #1005 and #1183.

## Expectation / Proposal
I find such unreliable tests really problematic, as (1) their failures produce effort to investigate why a build failed and reduce awareness for other, _real_ build problems, and (2) they are not useful at all if they can never really fail but only produce log output that no one considers. So obviously there are two options for how to get rid of these unreliable tests:
* Make them reliable: my question would if there are Mac users who can tell whether virtual table/tree viewers with lazy data providers actually work, so that the problem is only in the tests. Then there should be a way to make the tests work like a productive environment.
[...]

PR #1803, opened 2024-04-09
Fix Random Failing UIEditWorkingSetWizardAuto.testEditPage #275

Wittmaxi:
In my experiments, the test failed in roughly 2% of runs.

The test didn't close all processes and editors before deleting the files it contained.
This PR Makes the Job-Manager join all jobs before finally deleting the project-root.

This fix introduces a significant slowdown of the test (testing takes up to 2 seconds!) but since this is a single test and the test is important and the amount of fails (2%!) is big enough, I believe this tradeoff is very much justified.

I am confident in this fix working after seeing 400 consecutive runs that did not contain any fails (on windows).

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
@@ -13,10 +13,14 @@
  *******************************************************************************/
 package org.eclipse.ui.tests.dialogs;
 
-import java.util.List;
+import static org.eclipse.core.resources.ResourcesPlugin.getWorkspace;
 
+import java.util.List;
+import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.IWorkspaceDescription;
 import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.jface.wizard.IWizardPage;
 import org.eclipse.swt.widgets.Composite;
@@ -56,6 +60,29 @@ public class UIEditWorkingSetWizardAuto extends UIWorkingSetWizardsAuto<WorkingS
 		WorkingSetRegistry registry = WorkbenchPlugin.getDefault().getWorkingSetRegistry();
 		return registry.getDefaultWorkingSetPage();
 	}
+
+	/**
+	 * Blocks the calling thread until autobuild completes.
+	 */
+	private static void waitForBuild() {
+		((Workspace) getWorkspace()).getBuildManager().waitForAutoBuild();
+	}
+
+	/**
+	 * Enables or disables workspace autobuild. Waits for the build to be finished,
+	 * even if the autobuild value did not change and a previous build is still
+	 * running.
+	 */
+	private static void setAutoBuilding(boolean enabled) throws CoreException {
+		IWorkspace workspace = getWorkspace();
+		if (workspace.isAutoBuilding() != enabled) {
+			IWorkspaceDescription description = workspace.getDescription();
+			description.setAutoBuilding(enabled);
+			workspace.setDescription(description);
+		}
+		waitForBuild();
+	}
+
 
 	@Test
 	public void testEditPage() throws Throwable {
@@ -119,4 +146,17 @@ public class UIEditWorkingSetWizardAuto extends UIWorkingSetWizardsAuto<WorkingS
 
 		DialogCheck.assertDialogTexts(getWizardDialog());
 	}
+
+	@Override
+	public void doSetUp() throws Exception {
+		super.doSetUp();
+		setAutoBuilding(false);
+	}
+
+	@Override
+	public void doTearDown() throws Exception {
+		super.doTearDown();
+		ResourcesPlugin.getWorkspace().setDescription(Workspace.defaultWorkspaceDescription());
+	}
+
 }


Wittmaxi:
Random failing tests for this PR are documented here:

https://github.com/eclipse-platform/eclipse.platform.ui/issues/1583
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
```diff
@@ -135,10 +137,18 @@ private void removeAllWorkingSets() {
...
 			ResourcesPlugin.getWorkspace().getRoot().delete(true, null);
 		} catch (CoreException e) {
 			TestPlugin.getDefault().getLog().log(e.getStatus());
 			throw createAssertionError(e);
+		} catch (OperationCanceledException e) {
+			e.printStackTrace();
+			throw new AssertionError(e.getMessage(), e);
+		} catch (InterruptedException e) {
+			e.printStackTrace();
+			throw new AssertionError(e.getMessage(), e);
```
HeikoKlare: I would propose to have the method simply throw the exception, either directly or after doing some logging. So you could just replace the catch blocks with a `throws` declaration on the method signature. The stack trace should be reported then anyway (and otherwise I would not recommend to simply print it but to use some logger for logging it).

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
```diff
@@ -135,10 +137,18 @@ private void removeAllWorkingSets() {
 
 	private void cleanupWorkspace() {
 		try {
+			PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().closeAllEditors(false);
```
HeikoKlare: I wonder why there are editors opens in this test at all. I don't see any editors being opened by this test case. Do I miss something? Otherwise the test that leaves editors open should be responsible for closing them.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
```diff
@@ -135,10 +137,18 @@ private void removeAllWorkingSets() {
...
+			Job.getJobManager().join(ResourcesPlugin.FAMILY_AUTO_BUILD, null);
```
HeikoKlare: The PR description states `This PR Makes the Job-Manager join all jobs before finally deleting the project-root.`. However, this statement does only join the auto build job and no other jobs (like the auto refresh job, which may also conflict with resource deletion). Are you sure that the auto build job is causing the problem in this case? There is, for example `ImportTestUtils.waitForBuild()` that joins on more relevant jobs like refresh and build.

HeikoKlare (commented):
Great to see this test failure being addressed, thank you! I have some questions/comments on the proposed fix.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
Wittmaxi: Addressed! https://github.com/eclipse-platform/eclipse.platform.ui/commit/ded182da305777b5c3fb5cfbedd88cb02449cb1f

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
Wittmaxi: Yes, you are correct, there are no editors opened and thus this does not need to be performed. I removed this https://github.com/eclipse-platform/eclipse.platform.ui/commit/ded182da305777b5c3fb5cfbedd88cb02449cb1f

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
Wittmaxi: Once again, you are correct: I believe that the reason for why this "fix" worked was that I introduced some artificial delay specifically by closing the editors (a bit like what was done in FindReplaceDialogTest for awaiting SWT-events).
I have made sure to close *all* jobs in this commit: https://github.com/eclipse-platform/eclipse.platform.ui/commit/ded182da305777b5c3fb5cfbedd88cb02449cb1f

Wittmaxi:
In 2000 consecutive runs, the error did not occur. If you try this yourself: there is an error occurring on the second run of the test:

grafik

This fail ALWAYS happens, ALWAYS on the second run of the series and NEVER in another place - I do not believe it to be worth our time since it is almost certainly related to having multiple runs back-to-back instead of running the test once (as it is intended to be).

Additionally, the test is now much faster, taking about 0.25s on average

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
HeikoKlare: Note that calling `JobManager.cancel()` gives no guarantee that jobs will really be canceled. Even if they do so, the call is not blocking, so job termination may happen after the `cancel()` call returned. In particular, there will usually still be jobs running after `cancel()` has been called. See this statement in the documentation:
> Currently executing jobs will be asked to cancel but thereis no guarantee that they will do so.

So assuming your analysis of the issue is correct (and I really think it is correct and do not question it), it would be correct to join the relevant jobs (like you did before for the auto-build job). What now probably happens is that the `cancel()` call produces sufficient overhead to ensure that the conflicting job (probably auto build or auto refresh) have finished meanwhile. So at some point in time (when the overhead is not sufficient anymore), the test will fail again. I see two opportunities:
1. Re-introduce the join call, maybe also for the auto refresh job. I think this one is sufficient, still it may happen (in very, very rare cases) that the job is scheduled and started between calling `join()` and cleaning up the workspace.
2. Disable auto build for the test (see `ResourceTestUtil.setAutoBuilding()`): that's probably the cleanest solution but requires slightly more implementation effort.

HeikoKlare:
Unrelated and documented failing test: #1005

Wittmaxi:
Random Failing test documented here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
Wittmaxi: When I try to "join all threads", this is the UI that I get: 
grafik

Thus, I believe that indeed autobuild is the culprit. I will thus try your second solution

Wittmaxi:
I believe the fix to be more adequate now

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
```diff
@@ -133,7 +133,7 @@ private void removeAllWorkingSets() {
 		}
 	}
 
-	private void cleanupWorkspace() {
+	private void cleanupWorkspace() throws OperationCanceledException {
```
HeikoKlare: Is this still necessary?

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
```diff
@@ -57,8 +61,32 @@ private IWorkingSetPage getDefaultEditPage() {
 		return registry.getDefaultWorkingSetPage();
 	}
 
+	/**
+	 * Blocks the calling thread until autobuild completes.
+	 */
+	public static void waitForBuild() {
```
HeikoKlare: These methods (`waitForBuild()` and `setAutoBuilding()`) should be private to ensure that if someone wants to reuse them they have to be moved to some central utility class.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
```diff
@@ -57,8 +61,32 @@ private IWorkingSetPage getDefaultEditPage() {
...
+			description.setAutoBuilding(enabled);
+			workspace.setDescription(description);
+		}
+		waitForBuild();
+	}
+
+
 	@Test
 	public void testEditPage() throws Throwable {
+		setAutoBuilding(false);
```
HeikoKlare: This leaves the environment after the test in an unexpected state, as autobuilding is expected to be turned on. You can either
1. put the complete test in a try/finally block and restore the old autobuilding state in the finally block, or
2. restore the old autobuilding state in a test teardown method.
Usually the latter is the better choice as it avoid to wrap a whole test method in a try/finally block.

HeikoKlare (changes requested):

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
Wittmaxi: https://github.com/eclipse-platform/eclipse.platform.ui/commit/8bb3de749ce6ddc05828bb747dc2c84d2a031889

I opted for the second solution, by Overriding doSetUp() and doTearDown();

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
Wittmaxi: Agreed, done in https://github.com/eclipse-platform/eclipse.platform.ui/commit/8bb3de749ce6ddc05828bb747dc2c84d2a031889

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
Wittmaxi: No, I could completely reduce my PR into one single class-file-change

https://github.com/eclipse-platform/eclipse.platform.ui/commit/8bb3de749ce6ddc05828bb747dc2c84d2a031889

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
HeikoKlare: Alright, now just setting the autobuild propery is unnecessary here since it's already set in `doSetUp()`.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
```diff
@@ -119,4 +146,17 @@ public void testEditPage() throws Throwable {
...
+	@Override
+	public void doSetUp() throws Exception {
+		super.doSetUp();
+		setAutoBuilding(false);
+	}
+
+	@Override
+	public void doTearDown() throws Exception {
+		super.doTearDown();
+		setAutoBuilding(true);
```
fedejeanne: You shouldn't assume `true` was the original value. 

Instead, save the original value in `doSetUp` and apply it again in `doTearDown`

fedejeanne (changes requested):
You should recover the **original** value for auto-building after finishing the test.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
HeikoKlare: Since there will be other tests that do not clean up properly, it is usually better to restore defaults than restoring the state before the test. So a general cleanup (also used by the platform resource tests), is to just restore the default workspace description:
```java
ResourcesPlugin.getWorkspace().setDescription(Workspace.defaultWorkspaceDescription());
```

fedejeanne: This would make the test(s) meddle with each other but I guess since one is restoring the **default** value then it should be OK.

In that case, why not push this all the way to the top to `UITestCase::tearDown`?

HeikoKlare: > This would make the test(s) meddle with each other

It would do the exact opposite.

> In that case, why not push this all the way to the top to `UITestCase::tearDown`?

Because it is out-of-scope of this PR to refactor the complete UI test infrastructure. Having unified cleanups to ensure proper test isolation would of course make sense but should be done (1) as a separate task and (2) should be done properly using mechanisms like JUnit 4 rules or JUnit 5 extensions instead of fostering test class inheritance.

Wittmaxi:
I have now changed the Setup/Teardown to 
1) disable autobuilding for the test
2) reset the workspace description (as Heiko suggested)

```java
@Override
  public void doSetUp() throws Exception {
  super.doSetUp();
  setAutoBuilding(false);
}

@Override
public void doTearDown() throws Exception {
  super.doTearDown();
  ResourcesPlugin.getWorkspace().setDescription(Workspace.defaultWorkspaceDescription());
}
```

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
fedejeanne: > > This would make the test(s) meddle with each other
> 
> It would do the exact opposite.

It's a hack that makes this test "fix whatever other tests might have broken", that's what I meant. But it's ok, it's outside the scope of this PR to hunt down all the offending tests ;-) . BTW do you already know of any such tests (by name) or does this have to be investigated yet?

fedejeanne (approved):

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIEditWorkingSetWizardAuto.java
HeikoKlare: > BTW do you already know of any such tests (by name) or does this have to be investigated yet?

If I knew them, I would have created a ticket or fixed them. Most of the test failures I faced in the last year have one of two reasons:
1. Race conditions
2. Improper test isolation

What can be done systematically to ensure isolation is to encapsulate state changes or potential leaks (like job execution) into proper rules (JUnit 4) or extensions (JUnit 5).

HeikoKlare:
Failing checks because of new issues are only because of an outdated reference build. Recent master build reports the same new issues: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/421/

HeikoKlare (approved):

PR #1791, opened 2024-04-03
Move activation logic for SearchOptions.WHOLE_WORDS to FindReplaceLogic

Wittmaxi:
when "whole words" is selected but a find-string was entered that is not a whole word, the dialog/overlay will disable the option for "whole words". However, the dialog/overlay will keep the selection of the button in it's internal state. By passing the findString to this method, we can make sure that we don't try to perform a whole-words search for a string that isn't a whole word.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192#discussion_r1548564603

Extracted from
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192

PR #1805, opened 2024-04-10
Incremental Search: extract logic for disabling on RegEx-Search #1791

Wittmaxi:
when "incremental search" is selected and "RegEx-Search" is as well, the dialog/overlay will disable the option for "whole words". This PR extracts that functionality into the Find/Replace-Logic.

Extracted from a Review-Comment in #1791

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
@@ -67,6 +67,12 @@ public interface IFindReplaceLogic {
 	public boolean isRegExSearchAvailableAndActive();
 
 	/**
+	 * {@return whether incremental search may be performed by the
+	 * find/replace-logic based on the currently active options}
+	 */
+	public boolean isIncrementalSearchAvailable();
+
+	/**
 	 * Updates the search result after the Text was Modified. Used in combination
 	 * with <code>setIncrementalSearch(true)</code>. This method specifically allows
 	 * for "search-as-you-type"
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
@@ -98,7 +98,8 @@ public class FindReplaceDialogTest {
 
 		private Runnable closeOperation;
 
-		DialogAccess(Accessor findReplaceDialogAccessor) {
+
+		DialogAccess(Accessor findReplaceDialogAccessor, boolean checkInitialConfiguration) {
 			findReplaceLogic= (FindReplaceLogic) findReplaceDialogAccessor.get("findReplaceLogic");
 			findCombo= (Combo) findReplaceDialogAccessor.get("fFindField");
 			forwardRadioButton= (Button) findReplaceDialogAccessor.get("fForwardRadioButton");
@@ -111,7 +112,9 @@ public class FindReplaceDialogTest {
 			replaceFindButton= (Button) findReplaceDialogAccessor.get("fReplaceFindButton");
 			shellRetriever= () -> ((Shell) findReplaceDialogAccessor.get("fActiveShell"));
 			closeOperation= () -> findReplaceDialogAccessor.invoke("close", null);
-			assertInitialConfiguration();
+			if (checkInitialConfiguration) {
+				assertInitialConfiguration();
+			}
 		}
 
 		void restoreInitialConfiguration() {
@@ -148,9 +151,13 @@ public class FindReplaceDialogTest {
 			assertFalse(regExCheckBox.getSelection());
 		}
 
-		void close() {
+		void closeAndRestore() {
 			restoreInitialConfiguration();
 			assertInitialConfiguration();
+			close();
+		}
+
+		void close() {
 			closeOperation.run();
 		}
 
@@ -171,7 +178,7 @@ public class FindReplaceDialogTest {
 		Shell shell= PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell();
 		Accessor dialogAccessor= new Accessor("org.eclipse.ui.texteditor.FindReplaceDialog", getClass().getClassLoader(), new Object[] { shell });
 		dialogAccessor.invoke("create", null);
-		dialog= new DialogAccess(dialogAccessor);
+		dialog= new DialogAccess(dialogAccessor, true);
 	}
 
 	private void openTextViewerAndFindReplaceDialog() {
@@ -190,6 +197,10 @@ public class FindReplaceDialogTest {
 	}
 
 	private void openFindReplaceDialogForTextViewer() {
+		openFindReplaceDialogForTextViewer(true);
+	}
+
+	private void openFindReplaceDialogForTextViewer(boolean checkInitialConfiguration) {
 		Accessor fFindReplaceAction;
 		fFindReplaceAction= new Accessor("org.eclipse.ui.texteditor.FindReplaceAction", getClass().getClassLoader(),
 				new Class[] { ResourceBundle.class, String.class, Shell.class, IFindReplaceTarget.class },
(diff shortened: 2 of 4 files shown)


Wittmaxi:
Random Fail for Ubuntu is reported here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1808
Random Fail for Mac is reported here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/811
Random Fail for Windows is reported here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1427

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -392,6 +393,19 @@ public void testReplaceAndFindAfterInitializingFindWithSelectedString() {
...
 	}
 
+	@Test
+	public void testIncrementalSearchOnlyActiveWhenAllowed() {
+		openTextViewer("text text text");
+		openFindReplaceDialogForTextViewer();
+
+		dialog.incrementalCheckBox.setSelection(true);
+		dialog.regExCheckBox.setSelection(true);
+		dialog.regExCheckBox.notifyListeners(SWT.Selection, null);
```
HeikoKlare: Please use the existing `select()` method instead, which already does the combination of setting the selection and informing listeners.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
```diff
@@ -66,6 +66,14 @@ public interface IFindReplaceLogic {
 	 */
 	public boolean isRegExSearchAvailableAndActive();
 
+	/**
+	 * Returns whether the Find/Replace-Overlay may perform incremental search
+	 * currently.
+	 *
+	 * @return whether incremental search may be performed with the current options.
```
HeikoKlare: This is the logic interface, which is completely independent from UI / overlay. If a method is only a getter (and thus only needs a "returns" documentation), you can also use a simplified Javadoc syntax: `{@return ... }`
```suggestion
	 * @return whether incremental search may be performed based on the currently active options
```

HeikoKlare (commented):
In general, this looks good. I have two minor comments.

HeikoKlare:
And could you please also adapt the commit message, just like for the "whole words" options? (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/1791#issuecomment-2046929929)

Wittmaxi:
@HeikoKlare Addressed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/d654b06728ceb6865a378e7608c0fec4e1ed13f3

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
Wittmaxi: I am not quite sure what you mean with that syntax?
I did adapt your JavaDoc-text change

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: Done in https://github.com/eclipse-platform/eclipse.platform.ui/commit/d654b06728ceb6865a378e7608c0fec4e1ed13f3

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -414,6 +415,18 @@ public void testActivateWholeWordsAndSearchForTwoWords() {
 		assertFalse(dialog.wholeWordCheckBox.getEnabled());
 	}
 
+	@Test
+	public void testIncrementalSearchOnlyActiveWhenAllowed() {
```
fedejeanne: The name of the method mentions "`active`" but it actually checks whether or not the checkbox is **enabled**. Either the name is wrong or the assertions.

Since I don't see any method in `FindReplaceDialog` that can give you the state of `findReplaceLogic.isIncrementalSearchAvailable()` for this test to check it, I assume a better name for the test would be:
```suggestion
	public void testIncrementalSearchOnlyEnabledWhenAllowed() {
```

fedejeanne (changes requested):
Minor improvement in a test name desired

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: https://github.com/eclipse-platform/eclipse.platform.ui/commit/859fef9745d8b5af0d1c8d0a5874a5236329b0cd

Good catch! Although now this becomes inconsistent with `testActivateWholeWordsAndSearchForTwoWords` - I wonder if it makes sense to open a new PR to also change that one (or is it fine changing it in this PR?)

fedejeanne: I'd split that one into 2 (in another PR):
* `testWholeWordsOnlyEnabledWhenSearchingForOneWord`
* `testWholeWordsDisnabledWhenSearchingForRegEx`

The test does more than it says anyway, which is confusing

fedejeanne (approved):

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
HeikoKlare: The javadoc still refers to UI. But unfortunately my proposal was incomplete, sorry for that. This is what I would propose to have the javadoc look like:

```java
/**
  * {@return whether incremental search may be performed based on the currently active options}
  */
```

Then it renders as follows:
image

Wittmaxi: I didn't know this was possible. Thank you for the suggestion!

HeikoKlare (changes requested):
The current state produces a regression, as the dialog can loose its "incremental" option when reopened. To reproduce, activate the "incremental" option, activate and deactive the "regex" option, then close and re-open the dialog. The "incremental" option is unselected but it selection should be recovered from the previous session of the dialog.

This is because the change misses an update of the functionality in  `FindReplaceDialog.setupFindReplaceLogic()`, which still bases the activation of the "incremental" option on its selection _and_ its enablement. The latter has to be removed, just like for the "whole word" option.
https://github.com/eclipse-platform/eclipse.platform.ui/blob/781d89e41f9de47cf7e0d81a0184d34782d69d00/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java#L1267-L1268

fedejeanne:
> The current state produces a regression ...

Now that you mention it, I'm pretty sure one can find some other weird behaviors when playing with "Incremental", "Whole words" and "Regular expressions" and closing + reopening the search dialog :-o

HeikoKlare:
> Now that you mention it, I'm pretty sure one can find some other weird behaviors when playing with "Incremental", "Whole words" and "Regular expressions" and closing + reopening the search dialog :-o

Can you elaborate on that? Is there something specific you have in mind and is that a general negative expectation or related to this PR?

fedejeanne:
> > Now that you mention it, I'm pretty sure one can find some other weird behaviors when playing with "Incremental", "Whole words" and "Regular expressions" and closing + reopening the search dialog :-o
> 
> Can you elaborate on that? Is there something specific you have in mind and is that a general negative expectation or related to this PR?

Sorry for being unclear. I wasn't talking about this PR :-)

Wittmaxi:
@HeikoKlare I have addressed your comment and added a Unit-Test which checks the behavior for "re-opening" the dialog.
A very large amount of improvements in testing-architecture were made in the Find/Replace-Overlay-PR, so once that is moved, we should test the "re-opening"-behavior of the Dialog/Overlay with the other buttons

Wittmaxi:
Windows Random Failing documented here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1807

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -148,12 +151,16 @@ private void assertInitialConfiguration() {
 			assertFalse(regExCheckBox.getSelection());
 		}
 
-		void close() {
+		void closeAndRestore() {
 			restoreInitialConfiguration();
 			assertInitialConfiguration();
 			closeOperation.run();
```
HeikoKlare: Just a nitpicky final comment. 
```suggestion
			close();
```

HeikoKlare: Done in a767ca3

HeikoKlare (approved):

HeikoKlare:
Failing checks because of new issues are only because of an outdated reference build. Recent master build reports the same new issues: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/422

PR #1812, opened 2024-04-11
Fix Regression introduced in #1669

N1k145:


iloveeclipse (changes requested):
While this PR fixes errors, both this and original PR shows both same (wrong) behavior. 
I see some **missing** menu entries and duplicated separators.

"Source", "Refactor", "navigate" and "Search" menus are affected, others may be too.

N1k145:
@iloveeclipse can you please describe your setup? I have all the entries that I am expecting and no double separators. I would like to fix this but I'm currently a bit lost

iloveeclipse:
Screenshot from current SDK (without this PR, PR doesn't fix that): broken Search
image

Try to select different things in different views. Try to open different editors (java vs plain text file). Selection recalculates contributions to menus. During this, something seem to be lost now.
In particular, I see that if compare editor is opened, separators in Search menu seem to be duplicated now (or menu entries missing).

iloveeclipse:
With this PR (same with original one):
image

image

iloveeclipse:
So it looks like neither original fix not this PR fix the problem => I would vote for reverting the https://github.com/eclipse-platform/eclipse.platform.ui/pull/1669 for now, especially because the problem is also visible for other projects now: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1814.

N1k145:
@iloveeclipse I found the issue, for the separators. Sometimes they are still wrapped in a contribution, I added this case. 
Missing Elements I have not seen. 

If this does not fix it for you, I would also agree to revert the initial commit.

iloveeclipse:
> @iloveeclipse I found the issue, for the separators. Sometimes they are still wrapped in a contribution, I added this case. Missing Elements I have not seen.

This introduces **another** regression: Search menu has now no separators at all (should have two around "Text" in the menu below):

image

N1k145:
@iloveeclipse interesting, for me they are still there on Win and GTK

Then I'm also for reverting

iloveeclipse:
> @iloveeclipse interesting, for me they are still there on Win and GTK
> 
> Then I'm also for reverting

=> https://github.com/eclipse-platform/eclipse.platform.ui/pull/1815

iloveeclipse:
After revert I would recommend to carefully test the patch with respect of the dynamic menu updates depending on selected view / editor, by switching between different parts, selecting different elements. So I found most of the regressions related to this PR by switching between Git History view / Package Explorer / Compare editor / Class file Editor / Java editor.

N1k145:
Thanks for the hints, I will spend some more time analyzing this in a few days.

Issue #1819, opened 2024-04-12
Find/Replace-Dialog: When selecting text and opening the Dialog, the search-scope isn't set properly


Wittmaxi:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce
![Search selected lines bug in Eclipse](https://github.com/eclipse-platform/eclipse.platform.ui/assets/2205684/4412d207-16a3-4958-88fa-718ecb58f9bb)

(recorded by @fedejeanne)

I expected: the search scope is set

But got: the search scope isn't set, the selection remains unchanged and the "only search in selected area"-toggle is selected.

## Tested under this environment:
(see gif)

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

PR #1820, opened 2024-04-12
Fix correctly setting Find-Scopes when opening Find/Replace-Dialog #1819

Wittmaxi:
This PR correctly disables SearchOptions.GLOBAL in the FindReplaceLogic object of the Dialog if a multiline-selection was selected while opening the Find/Replace-Dialog.

This PR also provides a unit-test for this behavior.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1819

### DIFF
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.ui.workbench.texteditor.tests;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
@@ -98,6 +100,8 @@ public class FindReplaceDialogTest {
 
 		private Runnable closeOperation;
 
+		public Button replaceAllButton;
+
 
 		DialogAccess(Accessor findReplaceDialogAccessor, boolean checkInitialConfiguration) {
 			findReplaceLogic= (FindReplaceLogic) findReplaceDialogAccessor.get("findReplaceLogic");
@@ -110,6 +114,7 @@ public class FindReplaceDialogTest {
 			incrementalCheckBox= (Button) findReplaceDialogAccessor.get("fIncrementalCheckBox");
 			regExCheckBox= (Button) findReplaceDialogAccessor.get("fIsRegExCheckBox");
 			replaceFindButton= (Button) findReplaceDialogAccessor.get("fReplaceFindButton");
+			replaceAllButton= (Button) findReplaceDialogAccessor.get("fReplaceAllButton");
 			shellRetriever= () -> ((Shell) findReplaceDialogAccessor.get("fActiveShell"));
 			closeOperation= () -> findReplaceDialogAccessor.invoke("close", null);
 			if (checkInitialConfiguration) {
@@ -424,6 +429,19 @@ public class FindReplaceDialogTest {
 		runEventQueue();
 		assertTrue(dialog.wholeWordCheckBox.getSelection());
 		assertFalse(dialog.wholeWordCheckBox.getEnabled());
+	}
+
+	@Test
+	public void testActivateDialogWithSelectionActive() {
+		openTextViewer("text" + System.lineSeparator() + "text" + System.lineSeparator() + "text");
+		fTextViewer.setSelectedRange(4 + System.lineSeparator().length(), 8 + System.lineSeparator().length());
+		openFindReplaceDialogForTextViewer(false);
+
+		assertFalse(dialog.globalRadioButton.getSelection());
+		dialog.findCombo.setText("text");
+		select(dialog.replaceAllButton);
+
+		assertThat(fTextViewer.getDocument().get(), is("text" + System.lineSeparator() + System.lineSeparator()));
 	}
 
 	@Test
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -946,6 +946,7 @@ class FindReplaceDialog extends Dialog {
 		} else {
 			fGlobalRadioButton.setSelection(false);
 			fSelectedRangeRadioButton.setSelection(true);
+			findReplaceLogic.deactivate(SearchOptions.GLOBAL);
 		}
 	}
 


Wittmaxi:
This PR is branched from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1805 since I needed the changes to the Testing Infrastructure I made in that PR in order to test this feature. The diff will look far cleaner, once #1805 is merged

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -426,6 +429,19 @@ public void testActivateWholeWordsAndSearchForTwoWords() {
...
 
+	@Test
+	public void testActivateDialogWithSelectionActive() {
+		openTextViewer("text" + System.lineSeparator() + "text" + System.lineSeparator() + "text");
+		fTextViewer.setSelectedRange(5, 9);
+		openFindReplaceDialogForTextViewer(false);
+
+		assertFalse(dialog.globalRadioButton.getSelection());
+		dialog.findCombo.setText("text");
+		dialog.replaceAllButton.notifyListeners(SWT.Selection, null);
```
HeikoKlare: Please use `select(dialog.replaceAllButton())`

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
```diff
@@ -68,7 +68,7 @@ public interface IFindReplaceLogic {
 
 	/**
 	 * {@return whether incremental search may be performed by the
-	 * find/replace-logic based on the currently active options}
+	 * find/replace-logic based on the currently active options.}
```
HeikoKlare: Please revert this unrelated change

HeikoKlare (commented):
Looks good, thank you!
Just two minor comments.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
Wittmaxi: Addressed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/d2b76269413f3578e8a605d2410049c3801569be

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: Addressed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/d2b76269413f3578e8a605d2410049c3801569be

it is worth noting that select does more than just call the listeners - it actively sets a selection on the replaceAllButton.
Since the button doesn't care about it's "selection", I believe that this is fine.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -426,6 +429,20 @@ public void testActivateWholeWordsAndSearchForTwoWords() {
...
+	@Test
+	public void testActivateDialogWithSelectionActive() {
+		openTextViewer("text" + System.lineSeparator() + "text" + System.lineSeparator() + "text");
+		fTextViewer.setSelectedRange(5, 9);
+		openFindReplaceDialogForTextViewer(false);
+
+		assertFalse(dialog.globalRadioButton.getSelection());
+		dialog.findCombo.setText("text");
+		select(dialog.replaceAllButton);
+		runEventQueue();
```
HeikoKlare: Why should it be necessary to run the event queue here? The test fails on Linux despite this call. I think what is missing in general is to run the event queue once the action to open the find/replace dialog has been invoked (like is introduced in #1192).

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -426,6 +429,20 @@ public void testActivateWholeWordsAndSearchForTwoWords() {
...
+
+		assertEquals(fTextViewer.getDocument().get(), "text" + System.lineSeparator() + System.lineSeparator());
```
HeikoKlare: Please switch expected and actual.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -426,6 +429,20 @@ public void testActivateWholeWordsAndSearchForTwoWords() {
 		assertFalse(dialog.wholeWordCheckBox.getEnabled());
 	}
 
+	@Test
+	public void testActivateDialogWithSelectionActive() {
+		openTextViewer("text" + System.lineSeparator() + "text" + System.lineSeparator() + "text");
+		fTextViewer.setSelectedRange(5, 9);
```
HeikoKlare: The range should properly consider the length of `lineSeparator()`.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
HeikoKlare: Root cause of the test failure will be fixed by #1836.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: I have chosen to use Hamcrest-Matchers instead in order to avoid this confusion. I could imagine this being considered inconsistent as this is the only place I use hamcrest in this test-class. Since the new FindReplaceDialogTest (introduced in the Find/Replace-Overlay-PR) uses Hamcrest, I don't believe this is a problem.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: Will #1836 really fix the test? From my understanding, the busy-indicator won't make this thread wait for performReplaceAll.

What we need for this test is a good way to know when the performReplaceAll-method is done running - no matter which thread it is in.

What is the correct procedure for right now?

HeikoKlare: It will fix the test, because the test fails because of a `NullPointerException` when executing "replace all", as the active shell is still `null` (missing GTK event processing). Doing a `runEventQueue()` after initializing the dialog and before doing the "replace all" would also do the trick, but is rather a hack that will not be necessary after merging #1836.

HeikoKlare (approved):
Great, thank you! Failing test is unrelated and documented in #926.

PR #1821, opened 2024-04-14
Improve the navigation history to take the editor ID into account

### LINKED ISSUE #1780: NavigationHistoryEntry and NavigationLocation do not take into account Editor IDs

merks:
- Update the restoreLocation() method of the NavigationHistoryEntry class to match the editor ID as well as the editor input when opening the editor.
- Update the getEditorPart() method of the NavigationLocation class to return the IEditorPart based on the editor ID as well as the editor input, if both are available.
- Updates the INavigationLocation interface and implementation to have getters and setters for the editor ID associated with the NavigationLocation.
- Whenever the getters and setters for the editor input of an INavigationLocation are used, update the logic to use the getters and setters of the editor ID in a corresponding way.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1780

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -6340,9 +6340,10 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 	 */
 	protected void editorSaved() {
 		IEditorInput input= getEditorInput();
+		String id = getEditorSite().getId();
 		for (INavigationLocation location2 : getSite().getPage().getNavigationHistory().getLocations()) {
 			if (location2 instanceof TextSelectionNavigationLocation) {
-				if(input.equals(location2.getInput())) {
+				if (input.equals(location2.getInput()) && id.equals(location2.getId())) {
 					TextSelectionNavigationLocation location= (TextSelectionNavigationLocation) location2;
 					location.partSaved(this);
 				}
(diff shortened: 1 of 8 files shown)


merks:
@Greglar4 

I cherry picked your commit and modified the test to capture more details about the current state of the workbench when there is an assertion failure, i.e., which editors are open and which navigation locations are in the navigation history.  I have no good theory for why the test fails.  I did notice that some "linking" tests did populate the navigation history before this new test runs.

merks:
@Greglar4 

FYI, one contributing factor to the test problem was that the IFile used for the file input as created not using the IFile APIs but rather directly in the file system such that the IFile was out of sync which causes refresh processing to happen in a less predictable order.  Processing events at the start of the test to clear out events left from a previous test also helped.

Greglar4:
Ah, I see. Thank you for looking into it and getting it sorted out.

merks:
Thanks for your patience and for the test.

PR #1822, opened 2024-04-16
Touching bundles after compiler change

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1979


iloveeclipse:
Mac tests usually take over an hour, all others are passed, good to merge.

PR #1823, opened 2024-04-16
FindReplaceLogicTest: Fix FindBeforeReplace-Bug for general FindReplaceTarget implementaitons

Wittmaxi:
Fix bug where in general, FindReplaceLogicTargets might accidentally perform a replacement even though a prior search did not yield any results.

The TextViewer's implementation fixes this bug intrinsically but in doing so, it violates the specification of the interface it implements.

This PR adds a check that there really was a search-result before performing any replacements.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -352,7 +352,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		if (findString != null && !findString.isEmpty()) {
 
 			try {
-				somethingFound = findNext(findString, isActive(SearchOptions.FORWARD));
+				somethingFound = findNext(findString);
 			} catch (PatternSyntaxException ex) {
 				status = new InvalidRegExStatus(ex);
 			} catch (IllegalStateException ex) {
@@ -510,12 +510,11 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	 * options.
 	 *
 	 * @param findString             the string to search for
-	 * @param forwardSearch          the direction of the search
 	 * @return <code>true</code> if the search string can be found using the given
 	 *         options
 	 *
 	 */
-	private boolean findNext(String findString, boolean forwardSearch) {
+	private boolean findNext(String findString) {
 
 		if (target == null) {
 			return false;
@@ -529,7 +528,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		}
 
 		int findReplacePosition = r.x;
-		if (forwardSearch) {
+		if (isActive(SearchOptions.FORWARD)) {
 			findReplacePosition += r.y;
 		}
 
@@ -542,7 +541,8 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 			return false;
 		}
 
-		if (forwardSearch && index >= findReplacePosition || !forwardSearch && index <= findReplacePosition) {
+		if (isActive(SearchOptions.FORWARD) && index >= findReplacePosition
+				|| !isActive(SearchOptions.FORWARD) && index <= findReplacePosition) {
 			statusLineMessage(""); //$NON-NLS-1$
 		}
 
@@ -565,7 +565,10 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		if (!isFindStringSelected(findString)) {
 			performSearch(findString);
 		}
-		return replaceSelection(replaceString);
+		if (getStatus().wasSuccessful()) {
+			return replaceSelection(replaceString);
+		}
+		return false;
 	}
 
 	private boolean isFindStringSelected(String findString) {
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/FindStatus.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/FindStatus.java
@@ -42,9 +42,7 @@ public class FindStatus implements IFindReplaceStatus {
 
 	@Override
 	public boolean wasSuccessful() {
-		// Also report StatusCode.WRAPPED as unsuccessful, because it implicitly
-		// includes NO_MATCH, as before wrapping no further match was found
-		return false;
+		return messageCode == StatusCode.WRAPPED;
 	}
 
 }
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
@@ -22,6 +22,13 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
 
 import org.junit.After;
 import org.junit.Before;
@@ -29,10 +36,12 @@ import org.junit.Test;
 import org.mockito.Mockito;
 
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Shell;
 
 import org.eclipse.jface.text.Document;
 import org.eclipse.jface.text.IFindReplaceTarget;
+import org.eclipse.jface.text.IFindReplaceTargetExtension3;
 import org.eclipse.jface.text.TextSelection;
 import org.eclipse.jface.text.TextViewer;
 
@@ -42,6 +51,8 @@ import org.eclipse.ui.internal.findandreplace.status.FindStatus.StatusCode;
 import org.eclipse.ui.internal.findandreplace.status.InvalidRegExStatus;
 import org.eclipse.ui.internal.findandreplace.status.NoStatus;
 import org.eclipse.ui.internal.findandreplace.status.ReplaceAllStatus;
+
+import org.eclipse.ui.texteditor.IFindReplaceTargetExtension2;
 
 
 public class FindReplaceLogicTest {
@@ -316,7 +327,7 @@ public class FindReplaceLogicTest {
 		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
 		findReplaceLogic.activate(SearchOptions.FORWARD);
 
-		boolean status = findReplaceLogic.performReplaceAndFind("<replace>", " ");
+		boolean status= findReplaceLogic.performReplaceAndFind("<replace>", " ");
 		assertThat(status, is(true));
 		assertThat(textViewer.getDocument().get(), equalTo("Hello World<replace>!"));
 		assertThat(findReplaceLogic.getTarget().getSelectionText(), equalTo("<replace>"));
@@ -587,6 +598,35 @@ public class FindReplaceLogicTest {
 		expectStatusEmpty(findReplaceLogic);
 		assertThat(findReplaceLogic.getTarget().getSelection().x, not(is(0)));
 		assertThat(findReplaceLogic.getTarget().getSelection().x, not(is(textViewer.getDocument().get().length() - LINE_STRING_LENGTH)));
+	}
+
+	/**
+	 * The TextViewer implementation of IFindReplaceLogic is misleading and not adhering to the
+	 * Interface: IFindReplaceTargetExtension3#replaceSelection will NOT replace the selection if
+	 * nothing was previously found. This does not generally have to be the case and thus we mock a
+	 * target that is setup like this:
+	 *
+	 * The text contained is {@code ~SELECTEDTEXT~abcd} with ~ marking the boundaries of the
+	 * selection. We perform a search for "NOTFOUND" first - the selection stays put since the
+	 * string was not found. At this point, we do not want to perform
+	 * {@code replaceSelection("NOTFOUND")} since an implementation adhering to the specification of
+	 * the function would just overwrite the current selection.
+	 */
+	@Test
+	public void onlySelectAndReplacesIfFindSuccessfulOnCustomTarget() {
+		IFindReplaceTarget mockedTarget= Mockito.mock(IFindReplaceTarget.class, withSettings().extraInterfaces(IFindReplaceTargetExtension3.class, IFindReplaceTargetExtension2.class));
+
+		when(mockedTarget.getSelectionText()).thenReturn("SELECTEDTEXT");
+		when(mockedTarget.getSelection()).thenReturn(new Point(0, "SELECTEDTEXT".length()));
+		when(mockedTarget.isEditable()).thenReturn(true);
+		when(mockedTarget.findAndSelect(anyInt(), anyString(), anyBoolean(), anyBoolean(), anyBoolean())).thenReturn(-1);
+		when(((IFindReplaceTargetExtension2) mockedTarget).validateTargetState()).thenReturn(true);
+
+		IFindReplaceLogic findReplaceLogic= new FindReplaceLogic();
+		findReplaceLogic.updateTarget(mockedTarget, true);
+		findReplaceLogic.performSelectAndReplace("NOTFOUND", "");
+
+		verify((IFindReplaceTargetExtension3) mockedTarget, never()).replaceSelection(anyString(), anyBoolean());
 	}
 
 	private void expectStatusEmpty(IFindReplaceLogic findReplaceLogic) {


tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -474,6 +475,23 @@ public void testWholeWordSearchAvailable() {
...
+		TextViewer textViewer= setupTextViewer("line" + System.lineSeparator() + "line" + System.lineSeparator() + "line");
+		textViewer.setSelectedRange(5, 9);
+		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
+		findReplaceLogic.deactivate(SearchOptions.GLOBAL);
+		findReplaceLogic.performSelectAndReplace("line", "");
+		expectStatusEmpty(findReplaceLogic);
+
+		findReplaceLogic.performSelectAndReplace("line", "");
+		assertThat(textViewer.getTextWidget().getText(), is("line" + System.lineSeparator() + System.lineSeparator()));
+		expectStatusEmpty(findReplaceLogic);
```
HeikoKlare: Can you explain this behavior, @Wittmaxi?
The test initially restricts the find/replace application to a selection containing the second line of the document. Then you replace the second line with empty contents and then assert that doing another replace call also replaces the third line. But actually the third line is not part of the initial selection. So I would expect that the second replace call does nothing. Why is the defined behavior expected/valid?

Wittmaxi: I am not sure if I understood you correctly, I will think out loud - please tell me if that addresses your question;

The function-signature of setSelectedRange is ```public void setSelectedRange(int selectionOffset, int selectionLength)```, so I set a selection that begins at the 5th Character and is nine character long. that could look like this:
```
line\n[line\nline] on linux
line\r[\nline\r\nli]ne on windows
```

the length of the expected selection is 4 + 4 + 1 or 4 + 4 + 2 on windows

Keep in mind that it is not crucial that the ENTIRE line is selected (hence why we can tolerate variable lineSeparator-width): a search scope will automatically expand to be a whole line if only a part of it is selected, as seen here:


With this selection in mind, performing replace once expects to get
```
line\r\n\r\nline
```
and on the second replace, I expect to get
```
line\r\n\r\n
```

On the third replace, I don't expect any changes (which is what this test covers!)

HeikoKlare: Sorry for the confusion, I took the "9" to be the end index and not the length of the selection.

Still, the test fails for me on the first assertion when activated (note that it currently misses an `@Test` annotation and is  thus not executed at all).

Wittmaxi: Indeed, I did not see that the test wasn't activated and assumed the test ran correctly - hence our misunderstanding.
The test fails becaue I did not include the fix to this unit-test into this PR which I had added to the Overlay-PR. I have fixed this, at least on my computer the (now activated) test runs fine.

HeikoKlare: I see. Can you explain why you have removed the assertion after the first replacement?
```java
assertThat(textViewer.getTextWidget().getText(), is("line" + System.lineSeparator() + System.lineSeparator() + "line"));
```
If I readd that assertion, the test fails. The failure shows that the first `performSelectAndReplace()` call replaces the third line instead of the second:
```
Expected: is "line\r\n\r\nline"
     but: was "line\r\nline\r\n"
```

Isn't that a bug?

Wittmaxi: It seems to me that this is indeed a bug. I have traced it down to being caused by a "perform search" in a scope not working properly. Here is a unit-test which reproduces that problem:

```java

	@Test
	public void testFirstSearchInScope() {
		String lineSeparator= System.lineSeparator();
		TextViewer textViewer= setupTextViewer("line" + lineSeparator + "line" + lineSeparator + "line");
		int lineSeparatorLength= lineSeparator.length();
		textViewer.setSelectedRange(4 + lineSeparatorLength, 8 + lineSeparatorLength);
		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
		findReplaceLogic.performSearch("line");
		expectStatusEmpty(findReplaceLogic);
		assertThat(findReplaceLogic.getTarget().getSelection().x, not(is(0)));
}
```

This behavior might be cause because the new search-offset isn't set when I update the searchscope.
In the "real-world", the bug doesn't seem to show:

Wittmaxi: The unit Test I provided for reproducing this bug is missing activation of the search-scope.
This is the new test:

```java
	@Test
	public void testFirstSearchInScope() {
		String lineSeparator= System.lineSeparator();
		TextViewer textViewer= setupTextViewer("line" + lineSeparator + "line" + lineSeparator + "line");
		int lineSeparatorLength= lineSeparator.length();
		textViewer.setSelectedRange(4 + lineSeparatorLength, 8 + lineSeparatorLength);
		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
		findReplaceLogic.activate(SearchOptions.GLOBAL);
		findReplaceLogic.performSearch("line");
		expectStatusEmpty(findReplaceLogic);
		assertThat(findReplaceLogic.getTarget().getSelection().x, not(is(0)));
	}
```

It doesn't fail, so it doesn't reproduce what is happening. I will keep the Unit-test anyway because it tests something important and untested so far

Wittmaxi:
```java
	@Test
	public void testFirstSearchInScope() {
		String lineSeparator= System.lineSeparator();
		TextViewer textViewer= setupTextViewer("line" + lineSeparator + "line" + lineSeparator + "line");
		int lineSeparatorLength= lineSeparator.length();
		textViewer.setSelectedRange(4 + lineSeparatorLength, 8 + lineSeparatorLength);
		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
		findReplaceLogic.deactivate(SearchOptions.GLOBAL);
		findReplaceLogic.performSearch("line");
		expectStatusEmpty(findReplaceLogic);
		assertThat(findReplaceLogic.getTarget().getSelection().x, not(is(0)));
		assertThat(findReplaceLogic.getTarget().getSelection().x, not(is(textViewer.getDocument().get().length() - 4)));
	}
```

This is the test we are looking for

Wittmaxi: I have found the cause of this strange bug:

When initializing the searchScope, the new "search-cursor" is set, like this:

```java
	/**
	 * Tells the dialog to perform searches only in the scope given by the actually
	 * selected lines.
	 */
	private void initializeSearchScope() {
		if (shouldInitIncrementalBaseLocation()) {
			initIncrementalBaseLocation();
		}

		if (target == null || !(target instanceof IFindReplaceTargetExtension)) {
			return;
		}

		IFindReplaceTargetExtension extensionTarget = (IFindReplaceTargetExtension) target;
		IRegion scope;
		Point lineSelection = extensionTarget.getLineSelection();

		scope = new Region(lineSelection.x, lineSelection.y);
		extensionTarget.setScope(scope);

		int scopeOffset = lineSelection.x;
		int scopeLength = lineSelection.y;
		int offset = isActive(SearchOptions.FORWARD) ? scopeOffset : scopeOffset + scopeLength;
		extensionTarget.setSelection(offset, 0);
	}
```

Note that the search-scope is set in dependence of whether forward-search is active or not. In my attempts to fix this bug, I set forward-search AFTER setting global search. This means that the search-cursor was set to the end of the end of the scope.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -597,7 +598,10 @@ public boolean performSelectAndReplace(String findString, String replaceString)
 		if (!isFindStringSelected(findString)) {
 			performSearch(findString);
 		}
-		return replaceSelection(replaceString);
+		if (getStatus() instanceof NoStatus) {
```
Wittmaxi: getStatus().wasSuccessful()

Wittmaxi:
As a follow up -> when this PR is merged, we need to remove the instanceof in the Overlay-PR

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -169,15 +169,16 @@ private void initializeSearchScope() {
...
 		Point lineSelection = extensionTarget.getLineSelection();
-		scope = new Region(lineSelection.x, lineSelection.y);
 
-		int offset = isActive(SearchOptions.FORWARD) ? scope.getOffset() : scope.getOffset() + scope.getLength();
+		scope = new Region(lineSelection.x, lineSelection.y);
+		extensionTarget.setScope(scope);
 
+		int scopeOffset = lineSelection.x;
+		int scopeLength = lineSelection.y;
+		int offset = isActive(SearchOptions.FORWARD) ? scopeOffset : scopeOffset + scopeLength;
```
HeikoKlare: Is there a specific reason why these lines have to be replaced? To me they seem to do the same as the single line statement did before.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -38,12 +39,15 @@
...
 import org.eclipse.ui.internal.findandreplace.status.FindAllStatus;
 import org.eclipse.ui.internal.findandreplace.status.FindStatus;
+import org.eclipse.ui.internal.findandreplace.status.FindStatus.StatusCode;
 import org.eclipse.ui.internal.findandreplace.status.InvalidRegExStatus;
 import org.eclipse.ui.internal.findandreplace.status.NoStatus;
 import org.eclipse.ui.internal.findandreplace.status.ReplaceAllStatus;
 
 
 public class FindReplaceLogicTest {
+	private static final int LINE_STRING_LENGTH= 4;
```
HeikoKlare:
```suggestion
	private static final int LINE_STRING_LENGTH= LINE_STRING.length();
```

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -474,6 +478,43 @@ public void testWholeWordSearchAvailable() {
 		assertThat(findReplaceLogic.isWholeWordSearchAvailable(""), is(false));
 	}
 
+	@Test
+	public void testReplaceInScopeStaysInScope() {
+		String lineSeparator= System.lineSeparator();
```
HeikoKlare: You can use a static import and then directly use `lineSeparator()`.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -542,12 +543,11 @@ private Point replaceSelection(String replaceString, boolean regExReplace) {
 	 * options.
 	 *
 	 * @param findString             the string to search for
-	 * @param forwardSearch          the direction of the search
 	 * @return <code>true</code> if the search string can be found using the given
 	 *         options
 	 *
 	 */
-	private boolean findNext(String findString, boolean forwardSearch) {
+	private boolean findNext(String findString) {
```
HeikoKlare, line 517: Just for my understanding: is this a kind of cleanup or is there a functional reason for this change?

Wittmaxi: The historic findNext method was passed these arguments from the Dialog. This method is only used internally (and is private) and was always used like this: findNext(-, isActive(forwardSearch)). Additionally, the logic part of the method (findNext (string, position)) already discarded this parameter, instead checking for internal state.
The only reason this parameter was used was for generating the right search offset position.

This has a few problems:
- the method signature is conflated to require more parameters than it actually should
- it is possible to generate broken behaviour by calling the method as such: `findNext(findString, !isActive(SearchOptions.FORWARD))`

Functionally, everything stays the same, and since the only reference to this method called the method with an argument polled from internal state anyway, I believe this is a big improvement on the status quo. I do not intend (and could not find) any functional changes

HeikoKlare: Thank you for the explanation! That it totally fine. I justed wanted to know whether I have to consider any behavioral change because of this when reviewing the change or whether it is only a refactoring.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
Wittmaxi: You are right. I thought I would need to rewrite parts of this section, so I pre-emptively refactored them (also to understand better what is happening).

Now, that I understand the error, this refactoring is not really important enough to be part of the PR. I removed it!

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
Wittmaxi: Yes, done in https://github.com/eclipse-platform/eclipse.platform.ui/commit/8e8a527bb19dad39d15c797d7cf57a8d889b579b

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
Wittmaxi: Done in https://github.com/eclipse-platform/eclipse.platform.ui/commit/8e8a527bb19dad39d15c797d7cf57a8d889b579b

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/status/FindStatus.java
```diff
@@ -42,9 +42,7 @@ public boolean isInputValid() {
 
 	@Override
 	public boolean wasSuccessful() {
-		// Also report StatusCode.WRAPPED as unsuccessful, because it implicitly
-		// includes NO_MATCH, as before wrapping no further match was found
-		return false;
+		return messageCode != StatusCode.NO_MATCH;
```
HeikoKlare: Shouldn't `READONLY` also be reported as not successful?
```suggestion
		return messageCode == StatusCode.WRAPPED;
```

Wittmaxi: Yes, it should! Fixed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/80f6de0c42e4cc17c3393545f38b500f1288ed52

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
HeikoKlare: You have removed this added check with one of the previous commits. Was that intended?

Wittmaxi: No, it was not! I have re-introduced a better check in https://github.com/eclipse-platform/eclipse.platform.ui/commit/92fbb3c62eaffa0df833a18128fc5966c8fc66ba

Wittmaxi:
During user-testing I found a bug directly related to this change: when selecting a string after wrapping, the string would not be replaced.
This was fixed by using getStatus().wasSuccessful().
I have added a Unit-Test which tests for this behavior.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -565,7 +565,10 @@ public boolean performSelectAndReplace(String findString, String replaceString)
 		if (!isFindStringSelected(findString)) {
 			performSearch(findString);
 		}
-		return replaceSelection(replaceString);
+		if (getStatus().wasSuccessful()) {
+			return replaceSelection(replaceString);
+		}
```
HeikoKlare, line 570: Even without this change, all tests (even the newly added ones) succeed. Can you add a regression test that validates that the unexpected behavior does not occur anymore?

HeikoKlare: I tried to figure out what kind of unexpected behavior(s) has/have been fixed, but I was unfortunately not able to find out. It would be great if you can split up the PR, such that each PR either
1. only adds tests to ensure that some specific behavior is tested, or
2. implements a change in the behavior along with one or more regression tests for that behavior.

A requirement for the latter is that the tests fail when I revert the change in the behavor.

Wittmaxi: the unexpected bevahior that was fixed is a regression from a previous commit to this PR where I used a check for (status instanceof NoStatus).

After the search wrapped, no replace would be performed since the status after wrap is not nostatus

Wittmaxi:
It looks like the internal behavior of Find/Replace already does what I tried fixing: this is the test-case which I expected to fail and wanted to fix in the addressed code:

```java
	@Test
	public void onlySelectAndReplacesIfFindSuccessful() {
		String setupString= "ABCD" + lineSeparator() + LINE_STRING;
		TextViewer textViewer= setupTextViewer(setupString);
		textViewer.setSelectedRange(0, 4);
		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
		findReplaceLogic.activate(SearchOptions.FORWARD);
		findReplaceLogic.activate(SearchOptions.WRAP);
		findReplaceLogic.performSelectAndReplace("NOTFOUND", "");
		assertThat(textViewer.getTextWidget().getText(), is(setupString));
	}
```
My rationale is: if there is an old selection, I don't want to replace anything because I could accidentally replace an old selection which wasn't overwritten while searching since nothing was found.
Purely from the naming of the methods of the `IFindReplaceTragetExtension3`, this would be a very important thing to do:
```java
	/**
	 * Replaces the selection with <code>replaceString</code>. If
	 * <code>regExReplace</code> is <code>true</code>, <code>replaceString</code> is
	 * a regex replace pattern which will get expanded if the underlying target
	 * supports it. Returns the region of the inserted text; note that the returned
	 * selection covers the expanded pattern in case of regex replace.
	 *
	 * @param replaceString the replace string (or a regex pattern)
	 * @param regExReplace  <code>true</code> if <code>replaceString</code> is a
	 *                      pattern
	 * @return the selection after replacing, i.e. the inserted text
	 */
	private Point replaceSelection(String replaceString, boolean regExReplace) {
		if (target instanceof IFindReplaceTargetExtension3)
			((IFindReplaceTargetExtension3) target).replaceSelection(replaceString, regExReplace);
		else
			target.replaceSelection(replaceString);

		return target.getSelection();
	}
```

However, the internals look a bit different:

`FindReplacDocumentAdapter` keeps in it's state a pattern-Matcher. In order to replace the Selected area, it will retrieve the area to replace from the pattern-matcher and not from the "actual document". Since nothing was found, the pattern-matcher is in an invalid state and will throw an exception.
This "bug-fix" happens out of pure chance because the internal state was not properly managed.

```java

		if (operationCode == REPLACE || operationCode == REPLACE_FIND_NEXT) {
			if (regExSearch) {
				Pattern pattern= fFindReplaceMatcher.pattern();
				String prevMatch= fFindReplaceMatcher.group();
				try {
					replaceText= interpretReplaceEscapes(replaceText, prevMatch);
					Matcher replaceTextMatcher= pattern.matcher(prevMatch);
					replaceText= replaceTextMatcher.replaceFirst(replaceText);
				} catch (IndexOutOfBoundsException ex) {
					throw new PatternSyntaxException(ex.getLocalizedMessage(), replaceText, -1);
				}
			}

			int offset= fFindReplaceMatcher.start(); // <<-- this line (207) fails
			int length= fFindReplaceMatcher.group().length();

			if (fDocument instanceof IRepairableDocumentExtension
					&& ((IRepairableDocumentExtension)fDocument).isLineInformationRepairNeeded(offset, length, replaceText)) {
				String message= TextMessages.getString("FindReplaceDocumentAdapter.incompatibleLineDelimiter"); //$NON-NLS-1$
				throw new PatternSyntaxException(message, replaceText, offset);
			}

			fDocument.replace(offset, length, replaceText);

			if (operationCode == REPLACE) {
				return new Region(offset, replaceText.length());
			}
		}

```

Due to very unfortunate naming, it is possible/likely that other Find/Replace-Targets do not implement their behavior this way and thus DO rely on me checking that there was no match found.

I believe that the best solution is for me to keep the fix and supply the garnish-Tests as a separate PR.
Alternatively, I could provide a very slight refactoring of the internals as a separate PR but that would introduce separate risk and a lot of delay into the release of the Overlay.

About the Tests for `testCanReplaceAfterWrap` It fails again when you replace `performSelectAndReplace` with 
```java
	@Override
	public boolean performSelectAndReplace(String findString, String replaceString) {
		resetStatus();
		if (!isFindStringSelected(findString)) {
			performSearch(findString);
		}
		if (getStatus() instanceof NoStatus) {
		return replaceSelection(replaceString);
		}
		return false;
	}
```

The test is not super relevant but not harmful and definitely a possible workflow which we definitely want to support

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -547,6 +563,74 @@ public void testWholeWordSearchAvailable() {
...
+		IFindReplaceTarget mockedTarget= Mockito.mock(IFindReplaceTarget.class, withSettings().extraInterfaces(IFindReplaceTargetExtension3.class, IFindReplaceTargetExtension2.class));
+
+		when(mockedTarget.getSelectionText()).thenReturn("SELECTEDTEXT");
+		when(mockedTarget.getSelection()).thenReturn(new Point(0, "SELECTEDTEXT".length()));
+		when(mockedTarget.isEditable()).thenReturn(true);
+		when(mockedTarget.findAndSelect(anyInt(), anyString(), anyBoolean(), anyBoolean(), anyBoolean())).thenReturn(-1);
+		when(((IFindReplaceTargetExtension2) mockedTarget).validateTargetState()).thenReturn(true);
+		doThrow(new RuntimeException("The find-string was not found so replacement is invalid at this step"))
+				.when((IFindReplaceTargetExtension3) mockedTarget)
+				.replaceSelection(anyString(), anyBoolean());
```
Wittmaxi: https://stackoverflow.com/questions/9841623/mockito-how-to-verify-method-was-called-on-an-object-created-within-a-method

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -547,6 +558,35 @@ public void testWholeWordSearchAvailable() {
...
+		when(mockedTarget.getSelection()).thenReturn(new Point(0, "SELECTEDTEXT".length()));
+		when(mockedTarget.isEditable()).thenReturn(true);
+		when(mockedTarget.findAndSelect(anyInt(), anyString(), anyBoolean(), anyBoolean(), anyBoolean())).thenReturn(-1);
+		when(((IFindReplaceTargetExtension2) mockedTarget).validateTargetState()).thenReturn(true);
+
+		IFindReplaceLogic findReplaceLogic= new FindReplaceLogic();
+		findReplaceLogic.updateTarget(mockedTarget, true);
+		findReplaceLogic.performSelectAndReplace("NOTFOUND", "");
+
+		verify((IFindReplaceTargetExtension3) mockedTarget, times(0)).replaceSelection(anyString(), anyBoolean());
```
HeikoKlare: Nit: It should be possible to use `none()` as a shortcut for `times(0)` 
```suggestion
		verify((IFindReplaceTargetExtension3) mockedTarget, none()).replaceSelection(anyString(), anyBoolean());
```

tests/org.eclipse.ui.workbench.texteditor.tests/META-INF/MANIFEST.MF
```diff
@@ -18,9 +18,9 @@ Require-Bundle:
...
  org.eclipse.text.tests;bundle-version="[3.5.0,4.0.0)",
- org.eclipse.core.expressions;bundle-version="[3.5.0,4.0.0)"
+ org.eclipse.core.expressions;bundle-version="[3.5.0,4.0.0)",
+ org.mockito.mockito-core
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Eclipse-BundleShape: dir
 Automatic-Module-Name: org.eclipse.ui.workbench.texteditor.tests
-Import-Package: org.mockito,
- org.mockito.stubbing;version="5.5.0"
+Import-Package: org.mockito.stubbing;version="[5.11.0,6.0.0)"
```
HeikoKlare: It look like these dependency changes are not necessary, are they? So please revert them if possible or otherwise please stick to package imports.

HeikoKlare (approved):
Thank you, @Wittmaxi!

Test properly reflects the potential issue in an according implementation of a find/replace target and also properly serves as a regression test for the adaptation made to the `FindReplaceLogic`

Issue #1337, opened 2023-11-29
Failure of ResourceInitialSelectionTest.testSingleSelectionAndOneInitialSelectionWithInitialPattern


fedejeanne:
Failed in [this build](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/7029540124/job/19127333573?pr=1301) of #1301 

```java
org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.testSingleSelectionAndOneInitialSelectionWithInitialPattern -- Time elapsed: 0.092 s <<< ERROR!
org.eclipse.core.internal.resources.ResourceException: Problems encountered while deleting resources.
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:821)
	at org.eclipse.core.internal.resources.Project.delete(Project.java:365)
	at org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.doTearDown(ResourceInitialSelectionTest.java:386)
	at org.eclipse.ui.tests.harness.util.UITestCase.tearDown(UITestCase.java:214)
```

jukzi:
duplicates https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #1824, opened 2024-04-17
Adding functionality to change the delay time to expand the tree while searching. 

ShahzaibIbrahim:
Fixes (https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1337)

Related Commit to https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1246. (This needs to be merged before the mentioned pull request.)

### DIFF
--- a/bundles/org.eclipse.e4.ui.dialogs/src/org/eclipse/e4/ui/dialogs/filteredtree/FilteredTree.java
+++ b/bundles/org.eclipse.e4.ui.dialogs/src/org/eclipse/e4/ui/dialogs/filteredtree/FilteredTree.java
@@ -103,18 +103,38 @@ public class FilteredTree extends Composite {
 	private static final long SOFT_MAX_EXPAND_TIME = 200;
 
 	/**
+	 * Time delay after which the search is triggered, acting as a debounce
+	 * mechanism.
+	 */
+	private final long refreshJobDelayInMillis;
+
+	/**
+	 * Default time for refresh job delay in ms
+	 */
+	private static final long DEFAULT_REFRESH_TIME = 200;
+
+	/**
 	 * Create a new instance of the receiver.
 	 *
-	 * @param parent
-	 *            the parent <code>Composite</code>
-	 * @param treeStyle
-	 *            the style bits for the <code>Tree</code>
-	 * @param filter
-	 *            the filter to be used
+	 * @param parent           the parent <code>Composite</code>
+	 * @param treeStyle        the style bits for the <code>Tree</code>
+	 * @param filter           the filter to be used
+	 * @param refreshDelayTime refresh delay in ms, the time to expand the tree
+	 *                         after debounce
+	 * @since 1.5
+	 */
+	public FilteredTree(Composite parent, int treeStyle, PatternFilter filter, long refreshDelayTime) {
+		super(parent, SWT.NONE);
+		this.refreshJobDelayInMillis = refreshDelayTime;
+		init(treeStyle, filter);
+	}
+
+	/**
+	 * Calls {@link #FilteredTree(Composite, int, PatternFilter, long)} with a
+	 * default refresh time
 	 */
 	public FilteredTree(Composite parent, int treeStyle, PatternFilter filter) {
-		super(parent, SWT.NONE);
-		init(treeStyle, filter);
+		this(parent, treeStyle, filter, DEFAULT_REFRESH_TIME);
 	}
 
 	/**
@@ -129,6 +149,7 @@ public class FilteredTree extends Composite {
 	 */
 	protected FilteredTree(Composite parent) {
 		super(parent, SWT.NONE);
+		this.refreshJobDelayInMillis = DEFAULT_REFRESH_TIME;
 	}
 
 	/**
@@ -540,7 +561,7 @@ public class FilteredTree extends Composite {
 	 * @since 3.5
 	 */
 	protected long getRefreshJobDelay() {
-		return 200;
+		return refreshJobDelayInMillis;
 	}
 
 	/**
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
@@ -148,6 +148,16 @@ public class FilteredTree extends Composite {
 	private static final long SOFT_MAX_EXPAND_TIME = 200;
 
 	/**
+	 * Time for refresh job delay in terms of expansion in long value
+	 */
+	private final long refreshJobDelayInMillis;
+
+	/**
+	 * Default time for refresh job delay in ms
+	 */
+	private static final long DEFAULT_REFRESH_TIME = 200;
+
+	/**
 	 * Create a new instance of the receiver. Subclasses that wish to override the
 	 * default creation behavior may use this constructor, but must ensure that the
 	 * <code>init(composite, int, PatternFilter)</code> method is called in the
@@ -171,9 +181,22 @@ public class FilteredTree extends Composite {
 	public FilteredTree(Composite parent, boolean useNewLook, boolean useFastHashLookup) {
 		super(parent, SWT.NONE);
 		this.parent = parent;
+		this.refreshJobDelayInMillis = DEFAULT_REFRESH_TIME;
 		if (treeViewer != null) {
 			treeViewer.setUseHashlookup(useFastHashLookup);
 		}
+	}
+
+	/**
+	 * Calls
+	 * {@link #FilteredTree(Composite, int, PatternFilter, boolean, boolean, long)}
+	 * with a default refresh time
+	 *
+	 * @since 3.116
+	 */
+	public FilteredTree(Composite parent, int treeStyle, PatternFilter filter, boolean useNewLook,
+			boolean useFastHashLookup) {
+		this(parent, treeStyle, filter, useNewLook, useFastHashLookup, DEFAULT_REFRESH_TIME);
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
```diff
@@ -788,6 +793,16 @@ protected void setFilterText(String filterText) {
...
 
+	/**
+	 * Override the default refresh job delay
+	 *
+	 * @param time
+	 * @since 3.132
+	 */
+	public void setRefreshJobDelay(long time) {
+		this.refreshJobDelay = time;
+	}
```
HeikoKlare: I would propose to provide this as an instantiation parameter (maybe as a contructor parameter or, even better, via some factory method) and make the stored value immutable. The issue with making this value mutable is that you have to consider the cases where this value has been accessed and has a current effect, in particular in terms of a pending delay. So you need to be precise about the effects of setting this value after instantiation. Making the value immutable solves all these problems.

As an additional note: the public API should not expose an information about a "refresh job". The public API could state that there is some delay after which the tree contents are refreshed, but doing that via a job is an implementation detail that should be hidden from the public API.

fedejeanne:
I noticed there is also a `FilteredTree` for E4 (`org.eclipse.e4.ui.dialogs.filteredtree.FilteredTree`). I think these improvements would also make sense in there.

I would pass the timeout as a constructor parameter and make it final. Make sure to let the default (no-args) constructor call the new (1-arg) constructor with the current value of 200 to preserve the current functionality.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
```diff
@@ -176,11 +181,22 @@ public class FilteredTree extends Composite {
...
 		if (treeViewer != null) {
 			treeViewer.setUseHashlookup(useFastHashLookup);
 		}
 	}
 
+	/**
+	 * Calls
+	 * {@link #FilteredTree(Composite, int, PatternFilter, boolean, boolean, long)}
+	 * with a default refresh time
+	 */
```
fedejeanne:
```suggestion
	/**
	 * Calls
	 * {@link #FilteredTree(Composite, int, PatternFilter, boolean, boolean, long)}
	 * with a default refresh time.
	 *
	 * @since 3.116
	 */
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
```diff
@@ -265,6 +285,7 @@ protected FilteredTree(Composite parent) {
 	@Deprecated
 	protected FilteredTree(Composite parent, boolean useNewLook) {
 		super(parent, SWT.NONE);
+		this.refreshJobDelay = DEFAULT_REFRESH_TIME;
 		this.parent = parent;
```
fedejeanne: Nit-pick:

```suggestion
		this(parent);
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
```diff
@@ -288,6 +309,7 @@ protected FilteredTree(Composite parent, boolean useNewLook) {
 	@Deprecated
 	public FilteredTree(Composite parent, int treeStyle, PatternFilter filter) {
 		super(parent, SWT.NONE);
+		this.refreshJobDelay = DEFAULT_REFRESH_TIME;
 		this.parent = parent;
```
fedejeanne: Nit-pick:
```suggestion
		this(parent);
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
```diff
@@ -315,6 +337,7 @@ public FilteredTree(Composite parent, int treeStyle, PatternFilter filter) {
 	public FilteredTree(Composite parent, int treeStyle, PatternFilter filter, boolean useNewLook) {
 		super(parent, SWT.NONE);
 		this.parent = parent;
+		this.refreshJobDelay = DEFAULT_REFRESH_TIME;
 		init(treeStyle, filter);
```
fedejeanne: Nit-pick:
```suggestion
		this(parent, treeStyle, filter);
```

fedejeanne (changes requested):
Looks good in general, I only have one minor change request about preserving the `@since` information in a constructor and some nit-picking suggestions to reduce code duplication in other constructors (I think it improves readability too when one calls other constructors).

Please also squash commits.

HeikoKlare (changes requested):
Recent changes do not reflect my initial comment (https://github.com/eclipse-platform/eclipse.platform.ui/pull/1824#discussion_r1570142351):
> As an additional note: the public API should not expose an information about a "refresh job". The public API could state that there is some delay after which the tree contents are refreshed, but doing that via a job is an implementation detail that should be hidden from the public API.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
```diff
@@ -788,6 +811,7 @@ protected void setFilterText(String filterText) {
 		}
 	}
 
+
```
fedejeanne: Remove this new empty line

fedejeanne (approved):
Ready to go on my side (after removing the last unnecessary empty line).

Please address Heiko's comment though.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
fedejeanne: Has this been addressed? I think you meant that the name of the setter should not have the word `Job` in it, right? But the setter is gone, the new parameters do not have the word `Job` in it and the new field (and its JavaDoc) are `private`, so it should be fine, right?

ShahzaibIbrahim: @fedejeanne yes, I have removed the setter and used constructor to initialize the final value. Also updated the JavaDoc without the word job in it.

fedejeanne: Sorry, I was unclear :-)
I meant it as a question to @HeikoKlare as in "are you happy with the current state?" If yes, then the "Request for changes" can be withdrawn and you (@ShahzaibIbrahim) can squash your commits so I can merge this for you ;-)

bundles/org.eclipse.e4.ui.dialogs/src/org/eclipse/e4/ui/dialogs/filteredtree/FilteredTree.java
```diff
@@ -102,21 +102,40 @@ public class FilteredTree extends Composite {
 	 */
 	private static final long SOFT_MAX_EXPAND_TIME = 200;
 
+	/**
+	 * Time for refresh job delay in terms of expansion in long value
+	 */
+	private final long refreshJobDelay;
```
HeikoKlare: I do not understand what "in terms of expansion in long value" means.

In addition, it would be good to have the variable reflect the unit, so either call it something like `refreshJobDelayInMillis` or, even better, use a `Duration` instance.

bundles/org.eclipse.e4.ui.dialogs/src/org/eclipse/e4/ui/dialogs/filteredtree/FilteredTree.java
```diff
@@ -102,21 +102,40 @@ public class FilteredTree extends Composite {
...
-	 * @param filter
-	 *            the filter to be used
-	 */
-	public FilteredTree(Composite parent, int treeStyle, PatternFilter filter) {
+	 * @param parent               the parent <code>Composite</code>
+	 * @param treeStyle            the style bits for the <code>Tree</code>
+	 * @param filter               the filter to be used
+	 * @param refreshDelayInMillis refresh delay in ms, the time to expand the tree
+	 *                             after debounce
+	 * @since 1.4
```
HeikoKlare: This tag is not correct, is it? In 1.4.0, this constructor did not exist. Shouldn't the minor version of the bundle be incremented when adding a constructor, i.e., new public API?

HeikoKlare (commented):
I have two further comments. One is only about clarity, the other is about incorrect versioning, if I am not mistaken.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/dialogs/FilteredTree.java
HeikoKlare: Yes, my original comment has been addressed, but unfortunately I don't think this is ready to be merged yet (see my recent comments).

bundles/org.eclipse.e4.ui.dialogs/src/org/eclipse/e4/ui/dialogs/filteredtree/FilteredTree.java
ShahzaibIbrahim: @HeikoKlare should the minor version be 5 for this? If yes, then who update it in Manifest.mf file?

HeikoKlare: Yes, I would expect the minor version to be incremented to 5. You can simply replace the version `1.4.300.qualifier` in the Manifest with `1.5.0.qualifier`

HeikoKlare:
Please also adapt the commit message according to the information in the [CONTRIBUTING.MD](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations), i.e., with a one-line title and and a short description of what the commit does. In particular, when referring to an issue in the commit message it should be placed in the "official" GitHub repositories (and not, e.g., in https://github.com/vi-eclipse/ like in the current commit message).

HeikoKlare (commented):
Functionally, this looks fine now, thanks!

Sorry for being nitpick about the commit message, but it is currently rather confusing: the header states something about the move refactoring, but the commit is about making the refresh delay for the FilteredTree configurable, so that's what the commit message should state. It does not really matter for this commit that the (first) consumer of this functionality will be the move refactoring of JDT. In addition, the issue reference in the commit message is wrong. It points to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1337. You can keep the link to the JDT issue in the commit message description, but the short link `#1337` in the commit message header will not point to a JDT issue but to platform UI. So please remove the `#1337` reference from the commit message header.

HeikoKlare:
I am sorry, but there is still a reference to an unrelated issue (`#1337` of platform.ui instead of jdt.ui) in the commit message.

ShahzaibIbrahim:
> I am sorry, but there is still a reference to an unrelated issue (`#1337` of platform.ui instead of jdt.ui) in the commit message.

Sorry for that. The issue was linked by itself. All I wanted was just plain-text issue reference that I added below.

HeikoKlare (approved):
Thank you for the contribution.

HeikoKlare:
Failing test is unrelated: #1005

PR #1834, opened 2024-04-21
Correct setting search scope when start of line is selected #1833

### LINKED ISSUE #1833: Selected lines option in "Find/Replace" not working when cursor is at beginning of line

## Steps to reproduce

Open any file and move the cursor to the beginning of a line. Open the find/replace dialog by pressing Ctrl+F and choose "selected line". You will see that the current line is not selected. If you move the cursor to any other position in the same line, the line is properly selected.

missing_line_selection

I expect the line containing the cursor to be selected, even if the cursor is at the beginning of the line.

## Cause

The reason is this faulty condition. It is supposed to check whether a multi-line selection ends at the beginning of a new line (and then excludes that last line), but does not consider that the selection is of length 0, i.e., no preceding line is selected at all.
https://github.com/eclipse-platform/eclipse.platform.ui/blob/316aec2cd102e71dd9284ea91445af9c27aa3f32/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewer.java#L894

The behavior is also reproducible with older SDK version. E.g., I have tested a 2020-03 release.
The faulty line of code exists since 2003, so the bug probably exists since then.

## Community

HeikoKlare:
When activating the "selected lines" option in the find/replace dialog with the cursor being placed at the beginning (i.e., index 0) of a line, no search scope is initialized. For all other potential cursor positions in a line, the search scope is properly initialized. The reason is a faulty check that prevents the last line of a multi-line selection to be included when the selection ends at the beginning of that last line.

This change corrects the behavior to properly initialize a search scope also the cursor is at the beginning of a line. It improves the according code and adds several test cases for the search scope initialization behavior, including one that serves as a regression test for the fixed issue.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1833

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewer.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewer.java
@@ -879,29 +879,40 @@ public class TextViewer extends Viewer implements
 		@Override
 		public Point getLineSelection() {
 			Point point= TextViewer.this.getSelectedRange();
+			int originalSelectionBeginIndex= point.x;
+			int originalSelectionLength= point.y;
 
 			try {
 				IDocument document= TextViewer.this.getDocument();
 
-				// beginning of line
-				int line= document.getLineOfOffset(point.x);
-				int offset= document.getLineOffset(line);
+				int firstSelectedLineNumber= document.getLineOfOffset(originalSelectionBeginIndex);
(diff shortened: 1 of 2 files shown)


HeikoKlare:
Failing MacOS build because of missing Java 8 is unrelated. A previous build of the same changes (when infrastructure still worked for MacOS) succeeded on all platforms: https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/8774565155

PR #1836, opened 2024-04-21
Move busy indicator handling for find/replace from business logic to UI

HeikoKlare:
The IFindReplaceLogic and its implementation currently accept a Display instance in the `performReplaceAll()` and `performSelectAll()` methods. This is used to show a busy indicator in case the operation takes too long. Showing a busy indicator is, however, the responsibility of the UI.

This change removes the Display instance from the logic interface and moves the responsibility of showing a busy indicator for the two operations to the UI consumer, i.e., the FindReplaceDialog class. It also ensures that in case the active shell of the dialog is null (e.g., because setting the shell has not been processed by the GTK event queue), no exception is thrown.


Wittmaxi:
Looks good to me! Will fix a few issues in my unit tests for the Dialog

Issue #1530, opened 2024-01-20
Support New > Recently Used


laeubi:
Currently if one creates a new file the user is presented by some static items depending on the perspective for example:

![grafik](https://github.com/eclipse-platform/eclipse.platform.ui/assets/1331477/d9374016-866a-4723-bdb9-3a132b38d8b7)

but there is also an `Other` category where one can select more.

It would be good that before `Other` the menu would simply show the last 5 recently used items (if not already in the menu) so if a user regular uses another type it becomes a shortcut automatically.

Dinesh0723:
@laeubi, even after the restart of eclipse application recently used shortcuts should appear?

Dinesh0723:
@laeubi, For example, I am in the Plugin perspective. If I am creating a new Java project, this project will appear as a shortcut for the Plugin perspective (since the plugin perspective does not contain Java project as a shortcut by default). Then, if I switch to another perspective, let's say, a perspective other than Java (for example), should the Java project still appear as a shortcut before others in this use case?

laeubi:
Yes the best would be to save it permanently.

laeubi:
ANd yes I think it should just always be shown, maybe filtered if it is already part of the menu.

PR #1837, opened 2024-04-22
Support - New Recently Used Submenu

Dinesh0723:
Introduced support for recently used submenus.

Fixes:https://github.com/eclipse-platform/eclipse.platform.ui/issues/1530

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/actions/BaseNewWizardMenu.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/actions/BaseNewWizardMenu.java
@@ -166,7 +166,10 @@ public class BaseNewWizardMenu extends CompoundContributionItem {
 	/*
 	 * Returns the action for the given wizard id, or null if not found.
 	 */
-	private IAction getAction(String id) {
+	/**
+	 * @since 3.132
+	 */
+	protected IAction getAction(String id) {
 		// Keep a cache, rather than creating a new action each time,
 		// so that image caching in ActionContributionItem works.
 		IAction action = actions.get(id);
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Locale;
+import java.util.Set;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.IExtensionPoint;
@@ -53,6 +54,7 @@ import org.eclipse.ui.ISharedImages;
 import org.eclipse.ui.IWorkingSetManager;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.internal.decorators.DecoratorManager;
+import org.eclipse.ui.internal.dialogs.DynamicMenuSelection;
 import org.eclipse.ui.internal.dialogs.WorkbenchPreferenceManager;
 import org.eclipse.ui.internal.help.CommandHelpServiceImpl;
 import org.eclipse.ui.internal.help.HelpServiceImpl;
@@ -80,6 +82,7 @@ import org.eclipse.ui.plugin.AbstractUIPlugin;
 import org.eclipse.ui.testing.TestableObject;
 import org.eclipse.ui.views.IViewRegistry;
 import org.eclipse.ui.wizards.IWizardRegistry;
+import org.eclipse.ui.workbench.DynamicMenuItem;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleEvent;
@@ -202,6 +205,8 @@ public class WorkbenchPlugin extends AbstractUIPlugin {
 
 	private ICommandHelpService commandHelpService;
 
+	private DynamicMenuItem dynamicMenuItem;
+
 	/**
 	 * Create an instance of the WorkbenchPlugin. The workbench plugin is
 	 * effectively the "application" for the workbench UI. The entire UI operates as
@@ -210,6 +215,7 @@ public class WorkbenchPlugin extends AbstractUIPlugin {
 	public WorkbenchPlugin() {
 		super();
 		inst = this;
+		dynamicMenuItem = new DynamicMenuItem();
 	}
 
 	/**
@@ -766,6 +772,10 @@ public class WorkbenchPlugin extends AbstractUIPlugin {
 			// to be loaded.s
 			if (uiBundle != null)
 				uiBundle.start(Bundle.START_TRANSIENT);
+
+			// For dynamic menu items
+			Set<String> selectedFromOther = DynamicMenuSelection.getInstance().getSelectedFromOther();
+			selectedFromOther.addAll(dynamicMenuItem.getDynamicMenuItems());
 		} catch (BundleException e) {
 			WorkbenchPlugin.log("Unable to load UI activator", e); //$NON-NLS-1$
 		}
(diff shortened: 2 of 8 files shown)


laeubi:
@Dinesh0723 many thanks can you share some screenshots?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
```diff
@@ -766,6 +772,10 @@ public void start(BundleContext context) throws Exception {
 			// to be loaded.s
 			if (uiBundle != null)
 				uiBundle.start(Bundle.START_TRANSIENT);
+
+			// For dynamic menu items
+			Set<String> selectedFromOther = DynamicMenuSelection.getInstance().getSelectedFromOther();
+			selectedFromOther.addAll(dynamicMenuItem.getDynamicMenuItems());
```
fedejeanne, line 778: Don't do this, this is a bad practice. If you want to add several items to that list then add a proper method in `DynamicMenuSelection` and document it.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/actions/BaseNewWizardMenu.java
```diff
@@ -166,7 +166,10 @@ public void dispose() {
 	/*
 	 * Returns the action for the given wizard id, or null if not found.
 	 */
-	private IAction getAction(String id) {
+	/**
+	 * @since 3.132
+	 */
```
mickaelistria, line 171: Please merged with already existing Javadoc block.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/DynamicMenuSelection.java
```diff
@@ -0,0 +1,55 @@
...
+			iterator.remove();
+		}
+	}
+
+	/**
+	 * @return Returns the selectedFromOther.
+	 */
+	public Set<String> getSelectedFromOther() {
+		return selectedFromOther;
+	}
```
fedejeanne, line 53: Please return an immutable set to avoid calls like the one you used in `WorkbenchPlugin.start()`. Adding stuff to this set should be done from within this class, otherwise it becomes really difficult to track who adds what.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/workbench/DynamicMenuItem.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
+
+public class DynamicMenuItem {
```
fedejeanne, line 26: Please add some JavaDoc to the class and also to the public methods.

mickaelistria, line 26: If this is to become a new API class, then it requires proper Javadoc and `@since ...` markers.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/workbench/DynamicMenuItem.java
```diff
@@ -0,0 +1,55 @@
...
+		if (jsn != null) {
+			Gson gsn = new Gson();
+			Type typ = new TypeToken<List<String>>() {
+			}.getType();
+			return gsn.fromJson(jsn, typ);
+		}
+		return new ArrayList<>();
+	}
+
+	public void setDynamicMenuItems(Set<String> set) {
```
fedejeanne, line 43: Rename this parameter to reflect what it represents and not what it is. One can see this is a set by looking at the type but one doesn't know what the set contains (I bet `items` would be more suitable)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/workbench/DynamicMenuItem.java
```diff
@@ -0,0 +1,55 @@
...
+public class DynamicMenuItem {
+
+	private static final String PLUGIN_ID = "org.eclipse.ui.workbench"; //$NON-NLS-1$
+	private static final String DYNAMIC_MENU_ITEMS = "dynamicMenuItems"; //$NON-NLS-1$
+
+	public List<String> getDynamicMenuItems() {
+		IEclipsePreferences pref = InstanceScope.INSTANCE.getNode(PLUGIN_ID);
+		String jsn = pref.get(DYNAMIC_MENU_ITEMS, null);
+		if (jsn != null) {
+			Gson gsn = new Gson();
```
mickaelistria, line 35: GSon is a relatively heavy library to use for this case. Can you try avoiding it, for example by storing/reading the list as comma-separated ids and use `pref.get(DYNAMIC_MENU_ITEM, "").split(",")` ?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/workbench/DynamicMenuItem.java
mickaelistria: But it would be even better if this could remain internal.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/workbench/DynamicMenuItem.java
```diff
@@ -0,0 +1,55 @@
...
+			Type typ = new TypeToken<List<String>>() {
+			}.getType();
+			return gsn.fromJson(jsn, typ);
+		}
```
fedejeanne, line 39: Rename: 
* `jsn` -> `dynamicItemsJson`
* `gsn` -> `gson` (or skip it altogether and do `new GSon().fromJson(...)`)
* `typ` -> `type` (or even `emptyType` if this is actually "empty"? )

In general, try to avoid using just abbreviations for the names of the variables/fields, they are a legacy practice and they make the code harder to read.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/workbench/DynamicMenuItem.java
mickaelistria: Also, DynamicMenuItem is not a great name, especially if it's meant to serve only the newWizard case. I think it would be better to have this nested in the NewWizardNewPage or something, with more narrowly-scope names for preferences.
So it could be invoked with something like `NewWizardNewPage.getItemsFromPreferences()` or something of that sort that reads more efficient.

fedejeanne (changes requested):
Thank you @Dinesh0723 for this contribution. 

I just have some minor changes based on some coding practices, could you please incorporate them?

To be clear, I did **not** test the changes, my observations are merely about the code itself and not about the functionality.

mickaelistria (changes requested):

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/workbench/DynamicMenuItem.java
lathapatil: @mickaelistria This cannot be nested in` NewWizardNewPage` class as it is having default / Package visibility . Can I use `NewWizard `class instead ? (Need to access from `WorkbenchPlugin.java` )

lathapatil:
@mickaelistria  , @fedejeanne  Changes requested are addressed in PR https://github.com/eclipse-platform/eclipse.platform.ui/pull/1922

> @Dinesh0723 many thanks can you share some screenshots?

@laeubi Shared here https://github.com/eclipse-platform/eclipse.platform.ui/pull/1922#issuecomment-2145362916

PR #1840, opened 2024-04-22
Extract testing framework from Find/Replace overlay PR #1192

Wittmaxi:
This commit extracts the testing framework for the find/Replace overlay. It introduces an abstract test class which abstractly tests interfaces for find/replace. For that, this commit provides an abstract "interface access" class which allows for abstract access to the underlying nature of find/replace interfaces.

This commit makes most sense when viewed together with the unit tests for the find/replace overlay. I have extracted this refactoring because a lot of small pre-PRs introduce new unit tests. Extracting these tests means that I do not need to port them to a new infrastructure later on.

Extracted from #1192

### DIFF
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/DialogAccess.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/DialogAccess.java
@@ -1,0 +1,345 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.ui.workbench.texteditor.tests;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.function.Supplier;
+
+import org.junit.Rule;
+import org.junit.rules.TestName;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Combo;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Shell;
+
+import org.eclipse.text.tests.Accessor;
+
+import org.eclipse.jface.util.Util;
+
+import org.eclipse.jface.text.IFindReplaceTarget;
+import org.eclipse.jface.text.IFindReplaceTargetExtension;
+
+import org.eclipse.ui.internal.findandreplace.FindReplaceLogic;
+import org.eclipse.ui.internal.findandreplace.IFindReplaceLogic;
+import org.eclipse.ui.internal.findandreplace.SearchOptions;
+
+class DialogAccess implements IFindReplaceUIAccess {
+
+	FindReplaceLogic findReplaceLogic;
+
+	Combo findCombo;
+
+	Combo replaceCombo;
+
+	Button forwardRadioButton;
+
+	Button globalRadioButton;
+
+	Button searchInRangeRadioButton;
+
+	Button caseCheckBox;
+
+	Button wrapCheckBox;
+
+	Button wholeWordCheckBox;
+
+	Button incrementalCheckBox;
+
+	Button regExCheckBox;
+
+	Button findButton;
+
+	Button replaceButton;
+
+	Button replaceFindButton;
+
+	Button replaceAllButton;
+
+	private Supplier<Shell> shellRetriever;
+
+	private Runnable closeOperation;
+
+	Accessor dialogAccessor;
+
+	DialogAccess(Accessor findReplaceDialogAccessor) {
+		dialogAccessor= findReplaceDialogAccessor;
+		findReplaceLogic= (FindReplaceLogic) findReplaceDialogAccessor.get("findReplaceLogic");
+		findCombo= (Combo) findReplaceDialogAccessor.get("fFindField");
+		replaceCombo= (Combo) findReplaceDialogAccessor.get("fReplaceField");
+		forwardRadioButton= (Button) findReplaceDialogAccessor.get("fForwardRadioButton");
+		globalRadioButton= (Button) findReplaceDialogAccessor.get("fGlobalRadioButton");
+		searchInRangeRadioButton= (Button) findReplaceDialogAccessor.get("fSelectedRangeRadioButton");
+		caseCheckBox= (Button) findReplaceDialogAccessor.get("fCaseCheckBox");
+		wrapCheckBox= (Button) findReplaceDialogAccessor.get("fWrapCheckBox");
+		wholeWordCheckBox= (Button) findReplaceDialogAccessor.get("fWholeWordCheckBox");
+		incrementalCheckBox= (Button) findReplaceDialogAccessor.get("fIncrementalCheckBox");
+		regExCheckBox= (Button) findReplaceDialogAccessor.get("fIsRegExCheckBox");
+		shellRetriever= () -> ((Shell) findReplaceDialogAccessor.get("fActiveShell"));
+		closeOperation= () -> findReplaceDialogAccessor.invoke("close", null);
+		findButton= (Button) findReplaceDialogAccessor.get("fFindNextButton");
+		replaceButton= (Button) findReplaceDialogAccessor.get("fReplaceSelectionButton");
+		replaceFindButton= (Button) findReplaceDialogAccessor.get("fReplaceFindButton");
+		replaceAllButton= (Button) findReplaceDialogAccessor.get("fReplaceAllButton");
+	}
+
+	void restoreInitialConfiguration() {
+		findCombo.setText("");
+		select(SearchOptions.FORWARD);
+		select(SearchOptions.GLOBAL);
+		select(SearchOptions.WRAP);
+		unselect(SearchOptions.INCREMENTAL);
+		unselect(SearchOptions.REGEX);
+		unselect(SearchOptions.CASE_SENSITIVE);
+		unselect(SearchOptions.WHOLE_WORD);
+	}
+
+	@Override
+	public Button getButtonForSearchOption(SearchOptions option) {
+		switch (option) {
+			case CASE_SENSITIVE:
+				return caseCheckBox;
+			case FORWARD:
+				return forwardRadioButton;
+			case GLOBAL:
+				return globalRadioButton;
+			case INCREMENTAL:
+				return incrementalCheckBox;
+			case REGEX:
+				return regExCheckBox;
+			case WHOLE_WORD:
+				return wholeWordCheckBox;
+			case WRAP:
+				return wrapCheckBox;
+			default:
+				return null;
+		}
+	}
+
+	@Override
+	public void select(SearchOptions option) {
+		Button button= getButtonForSearchOption(option);
+		button.setSelection(true);
+		button.notifyListeners(SWT.Selection, null);
+	}
+
+	@Override
+	public void unselect(SearchOptions option) {
+		Button button= getButtonForSearchOption(option);
+		if (option == SearchOptions.GLOBAL) {
+			button= searchInRangeRadioButton;
+			button.setSelection(true);
+			globalRadioButton.setSelection(false);
+		} else {
+			button.setSelection(false);
+		}
+		button.notifyListeners(SWT.Selection, null);
+	}
+
+
+	@Override
+	public void closeAndRestore() {
+		restoreInitialConfiguration();
+		assertInitialConfiguration();
+		closeOperation.run();
+	}
+
+	@Override
+	public void close() {
+		closeOperation.run();
+	}
+
+	@Rule
+	public TestName name= new TestName();
+
+	@Override
+	public void ensureHasFocusOnGTK() {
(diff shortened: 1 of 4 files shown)


Wittmaxi:
extracted from #1192

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/IFindReplaceUIAccess.java
```diff
@@ -0,0 +1,59 @@
...
+
+	void pressReplaceAll();
+
+	/**
+	 * Some UIs require a "find" before allowing a Replace. This method handles that, ensuring that
+	 * if "findstring" is contained in the target's text, it will definitely be replaced.
+	 *
+	 * To keep parity between tests, this method should only be called if no text is currently
+	 * selected on the target.
+	 */
```
HeikoKlare: There seems to be some important part in this specification, but the fact that different UIs may have different requirements for performing a replace is an implementation detail of the class realizing this interface. The specification should only state pre- and postconditions, i.e., what the caller of the method has to fulfill (no text must be selected on the target) and what the caller can expect (replacement of the specified find string in the target's text, if existing).

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/IFindReplaceUIAccess.java
```diff
@@ -0,0 +1,59 @@
...
+
+	Widget getButtonForSearchOption(SearchOptions option);
+
+	Set<SearchOptions> getEnabledOptions();
+
+	Set<SearchOptions> getSelectedOptions();
+
+	IFindReplaceLogic getFindReplaceLogic();
+
+	void pressReplaceAll();
```
HeikoKlare: There are the methods `pressReplaceAll()` and `performReplace()`. I would prefer to have a uniform naming, either referring to a button press or to performing some action (by doing a button press).

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/DialogAccess.java
```diff
@@ -0,0 +1,255 @@
...
+			button.setSelection(false);
+		}
+		button.notifyListeners(SWT.Selection, null);
+	}
+
+
+	@Override
+	public void close() {
+		restoreInitialConfiguration();
+		closeOperation.run();
```
HeikoKlare: In the previous implementation, there was an assertion of the initial configuration after restoring it. This needs to be re-integrated.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -0,0 +1,314 @@
...
+		assertEnabled(SearchOptions.WHOLE_WORD);
+		assertSelected(SearchOptions.WHOLE_WORD);
+
+		dialog.select(SearchOptions.REGEX);
+		assertDisabled(SearchOptions.WHOLE_WORD);
+		assertSelected(SearchOptions.WHOLE_WORD);
+	}
+
+
+	public abstract void assertInitialConfiguration();
```
HeikoKlare: This should be part of the `IFindReplaceUIAccess` class, as that class is also responsible for restoring the initial configuration and is thus able to properly evaluate it. This will also allow to assert the initial configuration on closing the UI, as it was the case before this PR.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -0,0 +1,314 @@
...
+			}
+			try {
+				Thread.sleep(50);
+			} catch (InterruptedException e) {
+				// do nothing
+			}
+		}
+	}
+
+	protected AccessType dialog;
```
HeikoKlare: Please avoid protected fields and provide a protected getter instead (to avoid that a subclass may override the value).

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -0,0 +1,314 @@
...
+ * Tests the FindReplaceDialog.
+ *
+ * @since 3.1
+ */
+public abstract class FindReplaceUITest<AccessType extends IFindReplaceUIAccess> {
+
+	@Rule
+	public TestName testName= new TestName();
+
+	protected TextViewer fTextViewer;
```
HeikoKlare: Please avoid protected fields and provide a protected getter instead (to avoid that a subclass may override the value).

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -0,0 +1,314 @@
...
+import org.eclipse.jface.text.TextViewer;
+
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.internal.findandreplace.SearchOptions;
+
+/**
+ * Tests the FindReplaceDialog.
+ *
+ * @since 3.1
+ */
```
HeikoKlare: The documentation is incorrect, both the specification text as well as the since tag. Please remove completely.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -0,0 +1,314 @@
...
+public abstract class FindReplaceUITest<AccessType extends IFindReplaceUIAccess> {
```
HeikoKlare, line 36: Please remove unnecessary empty lines between the methods in this class.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/IFindReplaceUIAccess.java
```diff
@@ -0,0 +1,59 @@
...
+
+	void simulateEnterInFindInputField(boolean shiftPressed);
+
+	void simulateKeyPressInFindInputField(int keyCode, boolean shiftPressed);
+
+	String getFindText();
+
+	String getReplaceText();
+
+	void setFindText(String text, boolean grabFocus);
```
HeikoKlare: Would be better to remove the `grabFocus` parameter and set focus on the input field separately. It is only used at a single place, at which the find combo can be accessed directly (via `dialog.getFindCombo()`) to set the focus.

HeikoKlare (commented):
In general, this looks good as a preparatory step to structure tests for the find/replace overlay PR. There is further potential for design improvements, but that can be done in follow-ups.

Still, there are minor things I would like to see addressed before merging, and tests are slightly less expressive than before this PR as some assertions have been removed (see comments). They should be readded. In addition, please take a look at the removed/added tests, as there seem to be tests that are removed by this PR, whereas (interestingly) there are four tests added.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: In order to extract this, I have also extracted the following methods into the Access-Interface: 
```java
        void assertUnselected(SearchOptions option);
	void assertSelected(SearchOptions option);
	void assertDisabled(SearchOptions option);
	void assertEnabled(SearchOptions option);
```

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: I have now made it so that two methods are separated by one single line in this class

Wittmaxi:
In addition to your comments, I noticed that a lot of tests were lost during the merging of the two frameworks. I have gone through all of the recently merged PRs and I have adapted the tests

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/IFindReplaceUIAccess.java
Wittmaxi: Done in https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: Addressed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: Done in https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: textviewer and dialog are now private in https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: Moved in https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/DialogAccess.java
Wittmaxi:
```java
	@Override
	public void closeAndRestore() {
		restoreInitialConfiguration();
		assertInitialConfiguration();
		closeOperation.run();
	}

	@Override
	public void close() {
		closeOperation.run();
	}
```

Dialog access now implements one pure "close" method which does not reset the initial configuration. This is needed to test persistence of  properties between sessions.
The old method now restores the initial configuration and tests that that configuration was correctly set.
https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/IFindReplaceUIAccess.java
Wittmaxi: Since "single replace" does more than just press a button and overally, what really matters is that the action is performed, I have renamed both to "perform..."

https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/IFindReplaceUIAccess.java
Wittmaxi: changed to 
```java
	/**
	 * This method replaces all occurences of the find text in the arget with the replace text. No
	 * text must be selected on the target before calling.
	 */
```

https://github.com/eclipse-platform/eclipse.platform.ui/commit/49851d64fd5a67634b90f951a0aeddf3da74bfaa

HeikoKlare (approved):
I now made an in-depth comparison of the old and the new tests and found some changes that have made to how the tests have been set up. Since we should separate this refactoring (and addition of tests) from actual changes made to tests, I have slightly adapted this PR to only be a refactoring and not apply changes to how individual tests work. In case they are needed, we may apply them afterwards in separate PRs dedicated to those changes.

All builds succeeded and all refactored tests have been executed, so changes are fine.
I will push missing copyright headers to the files and then merge the PR.

Issue #1523, opened 2024-01-19
Copy and paste of a folder currently requires moving the cursor to the top folder


vogella:
I use Eclipse for text editing and this task requires frequently to create a copy of a folder.

To copy a folder I have to move to the top-level folder to paste it. This is not common in other tools.

Example:

FolderTop
-> Subfolder1
-> Lots of other folders
-> Subfolders99

Mouse click on Subfolder99, Ctrl+C on it, Ctrl+V-> nothing happens

To move Subfolder99 the following process is required: Mouse click on Subfolder99, Ctrl+C on it, Scroll to toplevel folder, mouse click on FolderTop and Ctrl+V.

Scrolling can be quite time-consuming if you have lots of folders, as for example here:

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/139910/bfb0b6f2-2f20-4b82-9258-340859d2b482)

Other tools like VsCode support this, in this screenshot I copyied the lib folder.

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/139910/9a45c144-dd2e-433a-898e-1359f2fc045a)

merks:
I suppose one could specialize "copying to self" to be "copy to parent".  One could also click the left arrow key before pasting.

alexanderKononov-sap:
Hi all, 🙋‍♂️ I am new here, I'm from SAP ADT team and would like to start contributing to eclipse too. Lack of this feature was bugging me too. I'd like to try and fix it if it's ok with everyone :)

vogella:
That would be great @alexanderKononov-sap

BeckerWdf:
Thanks @alexanderKononov-sap for providing a fix for this.

PR #1842, opened 2024-04-23
Copy paste into same folder

alexanderKononov-sap:
Fix for #1523

### DIFF
--- a/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
+++ b/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
@@ -14,6 +14,8 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.navigator.resources.actions;
 
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
 import org.eclipse.core.resources.IContainer;
@@ -80,10 +82,17 @@ import org.eclipse.ui.part.ResourceTransfer;
 	 * Returns the actual target of the paste action. Returns null
 	 * if no valid target is selected.
 	 *
+	 * @param clipboardContent current content of the clipboard
+	 *
 	 * @return the actual target of the paste action
 	 */
-	private IResource getTarget() {
+	private IResource getTarget(IResource[] clipboardContent) {
 		List<? extends IResource> selectedResources = getSelectedResources();
+
+		// selection is copied to itself => copy to parent
+		if (clipboardContent != null && areEqualsUnordered(selectedResources, Arrays.asList(clipboardContent))) {
+			return selectedResources.get(0).getParent();
+		}
 
 		for (IResource resource : selectedResources) {
 			if (resource instanceof IProject && !((IProject) resource).isOpen()) {
@@ -97,6 +106,18 @@ import org.eclipse.ui.part.ResourceTransfer;
 			}
 		}
 		return null;
+	}
+
+	/**
+	 * @param a The first collection
+	 * @param b The second collection
+	 * @return true if both collections aren't null and they contain the exact same
+	 *         items, regardless of their order.
+	 */
+	private boolean areEqualsUnordered(Collection<? extends IResource> a, Collection<? extends IResource> b) {
+		return b != null && a != null && !a.isEmpty() // they are not empty...
+				&& a.size() == b.size() // ... and they have the same size
+				&& a.containsAll(b); // ... and all elements of A are in B
 	}
 
 	/**
@@ -133,7 +154,7 @@ import org.eclipse.ui.part.ResourceTransfer;
 				}
 			} else {
 				// enablement should ensure that we always have access to a container
-				IContainer container = getContainer();
+				IContainer container = getContainer(resourceData);
 				CopyFilesAndFoldersOperation operation = new CopyFilesAndFoldersOperation(shell);
 				operation.copyResources(resourceData, container);
 			}
@@ -146,7 +167,7 @@ import org.eclipse.ui.part.ResourceTransfer;
 
 		if (fileData != null) {
 			// enablement should ensure that we always have access to a container
-			IContainer container = getContainer();
+			IContainer container = getContainer(null);
 			CopyFilesAndFoldersOperation operation = new CopyFilesAndFoldersOperation(shell);
 			operation.copyFiles(fileData, container);
 		}
@@ -154,9 +175,17 @@ import org.eclipse.ui.part.ResourceTransfer;
 
 	/**
 	 * Returns the container to hold the pasted resources.
+	 *
+	 * @param clipboardContent current content of the clipboard
 	 */
-	private IContainer getContainer() {
+	private IContainer getContainer(IResource[] clipboardContent) {
 		List<? extends IResource> selection = getSelectedResources();
+
+		// selection is copied to itself => copy to parent
+		if (clipboardContent != null && areEqualsUnordered(selection, Arrays.asList(clipboardContent))) {
+			return selection.get(0).getParent();
+		}
+
 		if (selection.get(0) instanceof IFile) {
 			return ((IFile) selection.get(0)).getParent();
 		}
@@ -205,7 +234,7 @@ import org.eclipse.ui.part.ResourceTransfer;
 			return false;
 		}
 
-		IResource targetResource = getTarget();
+		IResource targetResource = getTarget(resourceData);
 		// targetResource is null if no valid target is selected (e.g., open project)
 		// or selection is empty
 		if (targetResource == null) {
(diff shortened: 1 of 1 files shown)


BeckerWdf:
@alexanderKononov-sap: Thanks for this PR and for signing the ECA.

BeckerWdf:
I manually tested this in the project explorer. It looks good to me.
@vogella: Maybe you can test with your use-case as well.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
```diff
@@ -99,6 +107,22 @@ private IResource getTarget() {
...
+	private boolean selectionIsSameAsClipboardContent(List<? extends IResource> selectedResources,
+			List<? extends IResource> clipboardContent) {
+		if (clipboardContent != null && selectedResources != null && !selectedResources.isEmpty()) {
+			if (selectedResources.size() != clipboardContent.size()) {
+				return false;
+			}
+			return selectedResources.containsAll(clipboardContent);
+		}
+		return false;
+	}
```
fedejeanne: This method is an "equals" but it allows for different order of the items, so I think this more generic version is more appropriate:

```suggestion
	/**
	 * @param a The first collection
	 * @param b The second collection
	 * @return true if both collections aren't null and they contain the exact same
	 *         items, regardless of their order.
	 */
	private boolean areEqualsUnordered(Collection<? extends IResource> a, Collection<? extends IResource> b) {
		return b != null && a != null && !a.isEmpty() // they are not empty...
				&& a.size() == b.size() // ... and they have the same size
				&& a.containsAll(b); // ... and all elements of A are in B
	}
```

If performance ever becomes an issue (please test for it) then you could change the implementation so it first sorts both lists (or a copy of the list if you don't want to change the original lists) and then calls `equals` on them.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
```diff
@@ -146,17 +170,25 @@ public void run() {
...
 	 * Returns the container to hold the pasted resources.
+	 *
+	 * @param clipboardContent current content of the clipboard
 	 */
-	private IContainer getContainer() {
+	private IContainer getContainer(IResource[] clipboardContent) {
 		List<? extends IResource> selection = getSelectedResources();
+
+		// selection is copied to itself => copy to parent
+		if (clipboardContent != null && selectionIsSameAsClipboardContent(selection, Arrays.asList(clipboardContent))) {
```
fedejeanne:
```suggestion
		if (clipboardContent != null && areEqualsUnordered(selection, Arrays.asList(clipboardContent))) {
```

fedejeanne (commented):
Hello @alexanderKononov-sap , thank you for contributing.

The code looks good, I just have a minor remark on the name and the implementation of the new method.

DISCLAIMER: I haven't tested the code yet, I just focused on the coding practices :-)

vogella:
@alexanderKononov-sap thanks, code works fine. I tested in project explorer.

Please combine your two commits and write a good commit message, something like:

-----------
Enhance copy of a folder in project explorer.  Previous to fix fix you had to move to the top-level folder to paste a copied folder. This is not common in other tools.

Example:

FolderTop
-> Subfolder1
-> Subfolders99

Copy Subfolders99. Before this fix you had to move FolderTop to paste it, now you do not have to move the cursor.

Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1523

-----------

The main source of documentation should be the Git history.

vogella:
After this change and addressing the comment of @fedejeanne I plan to merge this change. Thanks again @alexanderKononov-sap for fixing this super ugly UX issue in the Eclipse IDE.

fedejeanne (changes requested):
I just tested the new functionality and found an error.

Here's a screenshot of a simple java project before this PR (left) and after this PR (right, in red)

image

You can see that the _Project Explorer_ has a nested package called `myproject.copy2` under the package `myproject` after applying this PR, which is wrong. The _Package Explorer_ doesn't have this issue though.

**Steps to reproduce**
* Open the _Project Explorer_
* Copy and paste the package `myproject` into itself

~I also noticed while using the _Project explorer_ that copying + pasting using the keyboard doesn't always react instantly, it takes some tries to get it to react. I haven't been able to reproduce this always, it just happened some times. Maybe a performance issue with the new methods?~

Please analyze and fix.

EDIT: the issue with copying and pasting with the keyboard is probably my keyboard/environment, I'm experiencing the same issue in other apps. You can ignore it :-)

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
alexanderKononov-sap: Thanks for the feedback, I'll change that

alexanderKononov-sap:
> I just tested the new functionality and found an error.
> 
> Here's a screenshot of a simple java project before this PR (left) and after this PR (right, in red)
> 
> image
> 
> You can see that the _Project Explorer_ has a nested package called `myproject.copy2` under the package `myproject` after applying this PR, which is wrong. The _Package Explorer_ doesn't have this issue though.
> 
> **Steps to reproduce**
> 
> * Open the _Project Explorer_
> * Copy and paste the package `myproject` into itself
> 
> ~I also noticed while using the _Project explorer_ that copying + pasting using the keyboard doesn't always react instantly, it takes some tries to get it to react. I haven't been able to reproduce this always, it just happened some times. Maybe a performance issue with the new methods?~
> 
> Please analyze and fix.
> 
> EDIT: the issue with copying and pasting with the keyboard is probably my keyboard/environment, I'm experiencing the same issue in other apps. You can ignore it :-)

Hi @fedejeanne! Thank you for testing it!
Somehow I do not understand what is happening there...

I am able to reproduce the issue too, but I cannot explain it, maybe due to my lack of deep knowledge of Eclipse :)

I was under the impression that Java project tree is completely handled in JDT (even if it is shown in *project explorer*, it still behaves like *package explorer*). That is why I also did not test it that much. And I am still under that impression, although confused.

The thing is, I do see the difference in behavior, but when I test it, my new coding does not seem to be used! I set breakpoints in all the methods and nothing seems to be used. Can it be that JDT also made some changes and this is because of that?

Help! :)

--EDIT: just tested it again in target without my new changes and it behaves the same. So I am guessing that you have also tested in the IDE, not in target? Apparently there were some JDT changes in between... can you confirm?

iloveeclipse:
Have you seen that in Package explorer we have `org.eclipse.jdt.internal.ui.refactoring.reorg.PasteAction` and in Project explorer `org.eclipse.ui.internal.navigator.resources.actions.PasteAction`?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
```diff
@@ -80,11 +81,18 @@ public PasteAction(Shell shell, Clipboard clipboard) {
...
+	 * @param clipboardContent current content of the clipboard
+	 *
 	 * @return the actual target of the paste action
 	 */
-	private IResource getTarget() {
+	private IResource getTarget(IResource[] clipboardContent) {
 		List<? extends IResource> selectedResources = getSelectedResources();
 
+		// selection is copied to itself => copy to parent
+		if (clipboardContent != null && selectionIsSameAsClipboardContent(selectedResources, Arrays.asList(clipboardContent))) {
```
fedejeanne:
```suggestion
		if (clipboardContent != null && areEqualsUnordered(selectedResources, Arrays.asList(clipboardContent))) {
```

fedejeanne (changes requested):
> --EDIT: just tested it again in target without my new changes and it behaves the same. So I am guessing that you have also tested in the IDE, not in target? Apparently there were some JDT changes in between... can you confirm?

Sorry for the late response. You're right, I just tested in **I20240425-1800** and the problem is reproducible, so it was not introduced by this PR:

image

You forgot to change one call to `selectionIsSameAsClipboardContent` though. Also, I think some unwanted commits slipped in, please check that too and after that this PR is ✔️ .

You can ping me after the freeze period is finished and I can merge it.

alexanderKononov-sap:
Sorry for the late reaction too, I am on vacation :) I hope it's fine now?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
BeckerWdf: @alexanderKononov-sap: Did you have a look on performance. Was the proposal "first sort and then call equals" not necesarry?

fedejeanne (approved):
> Sorry for the late reaction too, I am on vacation :) I hope it's fine now?

It's better, thank you. 

Just out of curiosity, have you looked into the other `PasteAction`s as commented before? https://github.com/eclipse-platform/eclipse.platform.ui/pull/1842#issuecomment-2075307072 . Is the new functionality also necessary there?

This PR is good to go. I checked the performance by copying a package with 100 classes into itself and I saw no performance degradation.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/PasteAction.java
fedejeanne: @BeckerWdf performance is still good, I checked by copying a folder with 100 classes into itself

jukzi:
@fedejeanne the used commit message violate best practice, please update commit message next time to have a single line summary and link to the issue
image

PR #1845, opened 2024-04-23
Use diamond operator wherever applicable

HeikoKlare:
Applies the cleanup rule for using the diamond operator where applicable in all projects to be conform with defined save actions.

Cleanup is explicitly _not_ applied to generated EMF code.


akurtakov:
MacOS not finding Java 8 anymore is unrelated.

akurtakov (approved):

PR #1846, opened 2024-04-24
FindReplaceLogic.updateTarget should call endSession on current target

merks:


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -588,12 +588,11 @@ public void updateTarget(IFindReplaceTarget newTarget, boolean canEditTarget) {
 		this.isTargetEditable = canEditTarget;
 
 		if (this.target != newTarget) {
-			if (newTarget != null && newTarget instanceof IFindReplaceTargetExtension)
-				((IFindReplaceTargetExtension) newTarget).endSession();
+			if (this.target instanceof IFindReplaceTargetExtension)
```
merks, line 591: We should end the session for the current target not for the new target.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -588,12 +588,11 @@ public void updateTarget(IFindReplaceTarget newTarget, boolean canEditTarget) {
...
+				((IFindReplaceTargetExtension) this.target).endSession();
 
 			this.target = newTarget;
-			if (newTarget != null)
-				isTargetSupportingRegEx = newTarget instanceof IFindReplaceTargetExtension3;
+			isTargetSupportingRegEx = newTarget instanceof IFindReplaceTargetExtension3;
```
merks, line 595: The null guard seems pointless.

merks:
@Wittmaxi 

FYI, I noticed that find in Oomph's setup editors was broken, often requiring me to close the editor to get it working again.   I tested against EMF's editors (e.g., the Ecore editor) and it had the same problem.  I finally tracked it down to this lifecycle problem.

EMF add this control:

image

And when the editor no longer has focus, the session should end and the control should be removed:

image

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
HeikoKlare: Thank you for finding this, @merks! This is a regression from #1132. Your proposal properly restored the previous behavior.

HeikoKlare (approved):
Thank you for the fix! It properly corrects a recently introduced regression. Thus, I propose to merge this once CI has finished. Then it should also make its way into the M2 build.

merks:
Windows/Mac failures unrelated.

PR #1847, opened 2024-04-24
Add missing override and deprecated annotations

HeikoKlare:
Applies the cleanup rules for adding missing `@override` and `@deprecated` annotations in all projects to be conform with defined save actions.
Also removes unnecessary "non-Javadoc" at overridden methods.


akurtakov (approved):
Thanks!

PR #1848, opened 2024-04-24
Use primitive types instead of wrapper types

HeikoKlare:
Applies the cleanup rule for using primitive instead of wrapper types in all projects to be conform with defined save actions.


akurtakov:
Thanks!

PR #1849, opened 2024-04-24
Unify access to static members

HeikoKlare:
Applies the following cleanup rules concerning static member access to all projects in order to be conform with defined save actions:
* change non-static accesses to static members using declaring type
* change indirect accesses to static members to direct accesses


akurtakov:
Thanks!

Issue #1827, opened 2024-04-18
Random failing SmartImportTests.testConfigurationFullBuildAfterImportedProjects 


HeikoKlare:
Random failing `SmartImportTests.testConfigurationFullBuildAfterImportedProjects`, e.g., in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1823/checks?check_run_id=23971305789

```
Expected 1 full-build trigger expected:<1> but was:<0>
java.lang.AssertionError: Expected 1 full-build trigger expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.eclipse.ui.tests.datatransfer.ImportTestUtils$TestBuilder.assertFullBuildWasDone(ImportTestUtils.java:96)
	at org.eclipse.ui.tests.datatransfer.SmartImportTests.testConfigurationFullBuildAfterImportedProjects(SmartImportTests.java:326)
```

HeikoKlare:
@fedejeanne maybe you can have a look since you recently worked on `SmartImportTests` (#1810)?

fedejeanne:
Failed on [this check](https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/8785263563/job/24105356246#step:7:20597) of #1841 (the PR is not committed yet so the names of the methods may change in the future). The important thing here is that the test that explicitly triggers a **full** build ends up failing because there were 0 full builds and 1 auto-build.

```java
java.lang.AssertionError: [Auto-build triggers expected:<0> but was:<1>] expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.eclipse.ui.tests.datatransfer.ImportTestUtils$TestBuilder.assertAll(ImportTestUtils.java:123)
	at org.eclipse.ui.tests.datatransfer.ImportTestUtils$TestBuilder.assertBuildsWereDone(ImportTestUtils.java:103)
	at org.eclipse.ui.tests.datatransfer.ImportTestUtils$TestBuilder.assertFullBuildWasDone(ImportTestUtils.java:95)
[...]

PR #1853, opened 2024-04-25
Convert functional interface instances into lambda expressions

HeikoKlare:
Apply cleanup to convert functional interface instances into simple lambda expression in all projects to align with defined save actions. Store lambda expression result in local variable where reasonable.


fedejeanne (dismissed):
LGTM

fedejeanne:
Failing tests are unrelated and already documented in #275 , #811 and https://github.com/eclipse-platform/eclipse.platform.ui/issues/1827#issuecomment-2078992830

PR #1861, opened 2024-04-29
VirtualLazyTreeViewerTest: avoid silent failures and force active shell

HeikoKlare:
Several tests in VirtualLazyTreeViewerTest silently fail because the tree data is not set and the test then simply returns as succeeding. This change does the following:
* Force shell activation to avoid the silent test failures
* Make that the variable identifying data to be set volatile to be consistent across threads
* Make the tests fail explicitly rather than silently

In case tests start to fail explicitly despite the changes, I propose to disable them completely (which does make a difference to the current state, see #1583).

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1583


HeikoKlare:
Failing test is unrelated and already documented: #370

Issue #1737, opened 2024-03-07
Random failing MMenuItemTest.testElementHierarchyInContext_HandledItem


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20240306-1800/testresults/html/org.eclipse.e4.ui.tests_ep432I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html
```
java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:87)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.junit.Assert.assertTrue(Assert.java:53)
at org.eclipse.e4.ui.tests.workbench.MMenuItemTest.testElementHierarchyInContext_HandledItem(MMenuItemTest.java:848)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at org.eclipse.e4.ui.tests.rules.WorkbenchContextRule$1.evaluate(WorkbenchContextRule.java:62)
```

fedejeanne:
It failed again in #3548

fedejeanne:
Failed again in https://github.com/eclipse-platform/eclipse.platform.ui/pull/4121

PR #1862, opened 2024-04-29
VirtualTableViewerTest: improve table update + force active shell #1005

### LINKED ISSUE #1005: Random failing VirtualTableViewerTest.testRenameWithSorter and VirtualTableViewerTest.testContains

HeikoKlare:
Several test cases of VirtualTableViewerTest randomly fail. This includes explicit failures of testRenameWithSorter() and testContains() and silent failures of other test cases that simply return successfully even if the test actually failed. The reason is a missing processing of an update of the table data, which is most likely caused by the shell not having focus.

With this change, the test execution ensures that in cases where an update of the table data is expected, the shell is forced active. In addition, early returns producing silent failures are replaced by assertions leading to explicit failures in case the condition is still not met.

In case tests that previously failed silently now start to fail explicitly despite the changes, I propose to disable them completely (which does make a difference to the current state, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/1583).

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1583

May fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/1005

### DIFF
--- a/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/VirtualLazyTableViewerTest.java
+++ b/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/VirtualLazyTableViewerTest.java
@@ -50,7 +50,6 @@ public class VirtualLazyTableViewerTest extends VirtualTableViewerTest {
 	public void setUp() {
 		updatedElements = new ArrayList<>();
 		super.setUp();
-		processEvents();
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


HeikoKlare:
Failing test is unrelated and already documented: #1737

PR #1863, opened 2024-04-29
added unit tests findReplaceLogic#selectAndReplace

Wittmaxi:
These tests test findReplaceLogic#selectAndReplace, in particular what happens when the search before replace results in edge-cases, like "no result was found" or "the result was only found after wrapping"

extracted from #1823

### DIFF
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
@@ -629,6 +629,35 @@ public class FindReplaceLogicTest {
 		verify((IFindReplaceTargetExtension3) mockedTarget, never()).replaceSelection(anyString(), anyBoolean());
 	}
 
+	@Test
+	public void testCanReplaceAfterWrap() {
+		TextViewer textViewer= setupTextViewer(LINE_STRING + lineSeparator() + LINE_STRING);
+		textViewer.setSelectedRange(LINE_STRING_LENGTH + lineSeparator().length(), 0);
+		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
+		findReplaceLogic.activate(SearchOptions.FORWARD);
+		findReplaceLogic.activate(SearchOptions.WRAP);
+		findReplaceLogic.performSelectAndReplace(LINE_STRING, "");
+		assertThat(textViewer.getTextWidget().getText(), is(LINE_STRING + lineSeparator()));
+		findReplaceLogic.performSelectAndReplace(LINE_STRING, "");
+		assertThat(textViewer.getTextWidget().getText(), is(lineSeparator()));
+	}
+
+	@Test
+	public void testDontSelectAndReplaceIfFindNotSuccessful() {
+		String setupString= "ABCD" + lineSeparator() + LINE_STRING;
+		TextViewer textViewer= setupTextViewer(setupString);
+		textViewer.setSelectedRange(0, 4);
+		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
+		findReplaceLogic.activate(SearchOptions.FORWARD);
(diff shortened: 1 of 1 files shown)


tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -547,6 +551,32 @@ public void testWholeWordSearchAvailable() {
...
+		findReplaceLogic.activate(SearchOptions.WRAP);
+		findReplaceLogic.performSelectAndReplace(LINE_STRING, "");
+		assertThat(textViewer.getTextWidget().getText(), is(LINE_STRING + lineSeparator()));
+		findReplaceLogic.performSelectAndReplace(LINE_STRING, "");
+		assertThat(textViewer.getTextWidget().getText(), is(lineSeparator()));
+	}
+
+	@Test
+	public void onlySelectAndReplacesIfFindSuccessful() {
+		String setupString= "ABCD" + lineSeparator() + LINE_STRING;
```
HeikoKlare, line 647: Should't this string contain the substring "NOTFOUND"? Otherwise the replace operation cannot replace anything anyway, no matter whether a specific range is selected or not.

Wittmaxi: If this string contained the subtring "NOTFOUND", then the substring "NOTFOUND" would be found, meaning that there is a selection that can be set for which the replacement would be valid.

The point of this test is that we want the replace operation to recognize itself when it cannot replace anything (e.g. because the search-string could not be found) and then leaves the current selection unaffected. In case the find/replace-operation cannot replace anything by itself, we don't want it to call the internal "replace selection"-method since that would (... if implemented as specified ...) overwrite a selection which doesn't contain the search-string

HeikoKlare: Thank you for the clarification. I accidentally though the selection would introduce a search scoped, but actually the "global" option is not deactivated.

Still I do not understand how this test is supposed to validate that the selection did not change. Currently, it only validates that when performing a search and replace operation on a string that does not contain the string to replace, then the content does not change. But I do not find any assertion that validated that the selection does not change.

Wittmaxi: The Test does not validate that the selection doesn't change, this test tests that the text highlighted by the selection does not change.
We assert that by ensuring that both before and after replace, the entire text inside of the target is exactly the same (since we do not expect anything to be replaced!)

HeikoKlare: > this test tests that the text highlighted by the selection does not change.

I am somehow stuck here. Can you elaborate how the test ensures this? The only assertion of this test validates that the text of the target does not change, but that does not validate anything with respect to the selection/highlighting, does it?

Wittmaxi: What I care about is that the text has not changed at all (there is nothing to replace).
I believe the misunderstanding stems from me not checking for a selection being set after the replace.
I think it would make sense to additionally check whether the selection stayed in place, I will add that check

HeikoKlare: > I believe the misunderstanding stems from me not checking for a selection being set after the replace.

Exactly, because I do not see why it should be necessary to set a selection in the test setup when this has no effect to the rest of the test (i.e., affecting the test outcome). Thanks for adding the selection checks.

HeikoKlare (approved):
Thank you for adding the tests. They look good now.

PR #1867, opened 2024-05-02
FindReplaceLogic add Tests for scoping-behavior

Wittmaxi:
These new unit-tests test for the behavior of FindReplaceLogic when applied to a search-scope. In particular, this commit tests that the search neaver accidentally leaves the scope and that the searching will automatically jump to the scope.


Wittmaxi:
Test failures are not related

HeikoKlare (approved):
Looks good. Thank you for adding these test!

Failing tests are unrelated and documented: #294 #275

PR #1869, opened 2024-05-02
Custom window size not considered for second opened window

### LINKED ISSUE #1868: Custom window size not considered for second opened window
We customize application by extending `WorkbenchWindowAdvisor` and there by overriding `preWindowOpen()` and in particular, setting custom window size for new windows (similar to what is done by `org.eclipse.equinox.internal.p2.ui.admin.rcp.ApplicationWorkbenchWindowAdvisor`):

```
	@Override
	public void preWindowOpen() {
		IWorkbenchWindowConfigurer configurer = getWindowConfigurer();
		configurer.setInitialSize(new Point(1280, 1024));
```

This works fine for the very first window opened by the application on startup.
However, windows opened from already opened window always have the **initial** size `1024, 768` which is hard coded in

https://github.com/eclipse-platform/eclipse.platform.ui/blob/9fd8c4a4d0dabbd767923f8374c02c92f8f34397/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/WorkbenchWindowConfigurer.java#L128

This happens because the code that sets the size for new windows created from existing one is called **before**  `preWindowOpen()` is called:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/9fd8c4a4d0dabbd767923f8374c02c92f8f34397/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/Workbench.java#L1479-L1482

The actual problem we see is that, **depending on timing**, the code that correctly updates the initial size for new windows, runs into troubles on GTK

https://github.com/eclipse-platform/eclipse.platform.ui/blob/9fd8c4a4d0dabbd767923f8374c02c92f8f34397/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/WorkbenchWindow.java#L976-L983

The last call here `shell.setBounds(bounds);` is the one that makes new window **randomly** (timing dependent) misbehave (wrong menu layout and missing repaints). I've debugged it and see that after `WorkbenchWindow.configureShell()` is done, GTK seem to resize the not yet shown shell **in parallel** to the following `WorkbenchWindow.setup()` code, which creates lot of async tasks that update the shell menu (see `menuUpdater` and `menuChangeManager` in same method). They run multiple times during window opening and shell resize, and I assume this is the place where we have race conditions with GTK shell init code. It looks like processing GTK display events between `fireWindowOpening()` and `configureShell()` "fixes" problems with created window.

However, it seem to cause troubles in case where `E4Workbench.NO_SAVED_MODEL_FOUND == null` (new window is **restored** by model), because UI tasks are dispatched before page is configured for the window (which by default happens right after `configureShell()`).

I would avoid risc to process events while window is being opened before it is fully constructed, so instead, I propose to fix the initial window size set with wrong values.

To fix that, the initial window size should be taken from already configured configurer of the current window, not from not yet properly initialized configurer of the window being opened.

PS. 
Additionally, this bug results in a problem that **after changing window size & restart**, new windows are **always** created with the wrong size because code in `configureShell()` doesn't get into the `if` block above and so doesn't update the window size properly.

I will provide a patch.

iloveeclipse:
If a new window should be created, read initial window size from the window configurer of already existing window (if any), and not from the not yet properly initialized window configurer of the "to be created" window.

This avoids extra shell.setBounds() call during WorkbenchWindow.setup() that may cause troubles on not yet opened shell instance depending on the timing.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1868

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -1477,7 +1477,14 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 				windowWhileInit = getActiveWorkbenchWindow();
 
 				if (newWindow) {
-					Point size = result.getWindowConfigurer().getInitialSize();
+					WorkbenchWindowConfigurer windowConfigurer;
+					WorkbenchWindow existingWindow = (WorkbenchWindow) windowWhileInit;
+					if (existingWindow != null) {
+						windowConfigurer = existingWindow.getWindowConfigurer();
+					} else {
+						windowConfigurer = result.getWindowConfigurer();
+					}
+					Point size = windowConfigurer.getInitialSize();
 					window.setWidth(size.x);
 					window.setHeight(size.y);
 


iloveeclipse:
Failed tests are unrelated (and do not fail on other platforms)
- `testEditPage (org.eclipse.ui.tests.dialogs.UIEditWorkingSetWizardAuto)`
- `testCancellingWhenRunning (org.eclipse.jface.text.tests.reconciler.FastAbstractReconcilerTest)`

Issue #1870, opened 2024-05-03
The label for internet exporer is missing


lak-proddev:
The label of Internet Explorer in the browser list on the preference page is wrong. It's due to the regression of the old commit

commit https://github.com/eclipse-platform/eclipse.platform.ui/commit/21add1aaf9e11ce8c092f6d981d2775cfb2eefb8
Deleted both Camino and Internet Explorer from Plugin.properties file and Plugin.xml file.

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/39690425/74101825-2368-40e3-bab6-40222a5953bc)

Bananeweizen:
I still experience the missing label, both in 2024-06, as well as in 2024-09 M1. Might this need some additional plugin version bumping or similar?

elsazac:
I made this commit https://github.com/eclipse-platform/eclipse.platform.ui/commit/21add1aaf9e11ce8c092f6d981d2775cfb2eefb8 . As per [link](https://support.microsoft.com/en-us/windows/internet-explorer-help-23360e49-9cd3-4dda-ba52-705336cc0de2) , support for IE was ended in 2022. Not sure why this is still required.

HeikoKlare:
IE is still the default browser used by Eclipse (SWT) and thus also the Platform. The Edge adaptation is not ready for productive use yet (even though currently some efforts are put into changing that), so we have to keep IE for now.

The removal of the label has been reverted here: 
- #1871

In my SDK setups (and in downloaded EPP products) I have this version of the `org.eclipse.ui.browser` bundle: 3.8.300.v20240524-2010
In that bundle, the correct plugin.properties containing the label are embedded.

When I start a fresh product (2024-09 RC1 for Java Developers), the label is properly shown:
![image](https://github.com/user-attachments/assets/f13c3a14-61e3-40b0-809e-6fcd7dc615ad)

Still, my development SDK had the broken label as I had set it up while the label was removed. This will never refresh on updates, but is initialized only once during setup. That may be the reason why you still see the broken label, @Bananeweizen. Can you confirm whether you can still reproduce the issue?

Bananeweizen:
@HeikoKlare that explains it. I was also wondering, as I did not immediately see a good reason in code/versioning/similar why that fix wouldn't change anything.
I can't reproduce the issue in other IDEs that I set up now. Thanks a lot.

PR #1871, opened 2024-05-03
The label for internet exporer is missing #1870

lak-proddev:
This commit will correct the label of Internet Explorer in the browser list on the preference page. It's due to the regression of the old commit

commit 21add1aaf9e11ce8c092f6d981d2775cfb2eefb8
Deleted both Camino and Internet Explorer from Plugin.properties file and Plugin.xml file.

### DIFF
--- a/bundles/org.eclipse.ui.browser/plugin.properties
+++ b/bundles/org.eclipse.ui.browser/plugin.properties
@@ -30,7 +30,7 @@ commandParameter.openBrowser.browserId.name=Browser Id
 commandParameter.openBrowser.name.name=Browser Name
 commandParameter.openBrowser.tooltip.name=Browser Tooltip
 
-
+browserInternetExplorer=Internet Explorer
 browserMicrosoftEdge=Microsoft Edge
 browserOpera=Opera
 browserChrome=Chrome


BeckerWdf:
Fix for #1870

BeckerWdf (commented):
We don't need the
Change-Id: I9f08c326dd89d9fd121759251610af2de603b831
line on github

fedejeanne:
@lak-proddev thank you for the contribution!
Could you please reword the commit and add a description of what it does? You don't need then "Change-Id", as already mentioned above. Same with the description of this PR, please.

Thank you!

fedejeanne:
Nitpick: I would have liked a commit description that says what the commit **does** and not what the status of the code **is** _i.e._ "**Add** missing label for Internet explorer" (notice the use of the verb) instead of "Label **is** missing".

I'll make sure to be more specific next time :-)

Thank you for your contribution, @lak-proddev!

lak-proddev:
> Nitpick: I would have liked a commit description that says what the commit **does** and not what the status of the code **is** _i.e._ "**Add** missing label for Internet explorer" (notice the use of the verb) instead of "Label **is** missing".

Agree

jukzi:
Note that when a commiter submits a commit github asks for the actual commit message that should be used. In this case i edited it on the fly - see 8ba3829122a77238ae15f1bcc0aa7d2592253022

PR #1873, opened 2024-05-05
[Generic Editor] Derive the file name from storage instead from the editor name # 1872

### LINKED ISSUE #1872: [Generic Editor] Content type determination does not work when opening files from history

## Steps to reproduce

* Install [Eclipse Wild Web Developer](https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper)
* Create a Typescript file (`.ts`)
* Write some code (e.g. functions, comments)
* Save the file
* Make further changes to the file in order to create entries in the local history
* Open the history: _Navigate > Show In > History_
* Open an entry from the history (not the last one)

I expected: The file is recognized as a Typescript file and syntax highlighting is applied.

But got: The Generic Editor displays the content as a plain text file without syntax highlighting.

When opening the latest (newest) entry, everything works as expected. This bug affects entries from the **local history** and also the **history of Git commits** (and possibly other team providers).

* Eclipse IDE/Platform version (as shown in *Help > About*): 2024-06 (4.32)

## Unsuccessful content type determination

There exist multiple strategies how the content type for the current editor is determined. In this case, the content type is derived from the name of the editor input:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/53689d198b5add99202778470211c70ddcd17983/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/ExtensionBasedTextViewerConfiguration.java#L177

The string consists of the **name** (filename), a **space** and a **revision identifier** (see https://github.com/eclipse-platform/eclipse.platform/blob/74bd1ca02d037a69d94e350ea3016c3fd893eb31/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/FileRevisionEditorInput.java#L172), e.g.

* Git: `"test.ts 826d3d5"`
* Local history: `"test.ts "` (with trailing space, because the revision identifier is empty)

From these strings the content type cannot be determined.

oleosterhagen:
Fixes #1872 

When possible, derive the file name from the underlying storage. This ensures that the file is not appended with a revision identifier and content type determination works when opening an editor from the History View.

### DIFF
--- a/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/ExtensionBasedTextViewerConfiguration.java
+++ b/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/ExtensionBasedTextViewerConfiguration.java
@@ -37,6 +37,8 @@ import java.util.stream.Collectors;
 
 import org.eclipse.core.filebuffers.FileBuffers;
 import org.eclipse.core.filebuffers.ITextFileBuffer;
+import org.eclipse.core.resources.IStorage;
+import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IPath;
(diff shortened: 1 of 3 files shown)


iloveeclipse (approved):
LGTM

PR #1874, opened 2024-05-06
NextTab action: allow Cycling when reaching the end

Wittmaxi:
Specifically for CTabFolders, this commit detects when the NextTab action has reached the last item and will cycle to the first item instead of doing nothing. It will load the item which is displayed in that page.

(achieved by pressing Ctrl + pgdown)

This fix only works for CTabFolders: the "traverse"-Method only returns whether it performed something or nothing. However, it also returns true if it effectively performed nothing but actually couldn't because it was on the last item.
For CTabFolders, I can check whether we are on the last item and manually jump to the second-to last item in the row. By jumping to the second-to-last item, I can still perform the "normal action" performed by SWT for traverse, which means that all Components are loaded correctly and the chevron is shown correctly.
I could not find a good general API which would allow to tell in general which tab I am on.
However, all of the "normal" views use a CTabFolder, so the loss of generality is minimal

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/handlers/TraversePageHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/handlers/TraversePageHandler.java
@@ -16,6 +16,8 @@ package org.eclipse.ui.internal.handlers;
 import java.lang.reflect.Method;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.CTabFolder;
+import org.eclipse.swt.custom.CTabItem;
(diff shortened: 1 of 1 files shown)


fedejeanne:
@Wittmaxi, thank you for contributing this.
I haven't really looked at the code yet so my comment is only a general one: is this a new feature or was the behavior "wrong" before this PR?

If I navigate through the tabs before applying this PR then the chevron is opened and I can select the next tab (it takes more clicks, but it's doable), which might be ok for others. This mean that this PR would be changing some normal (maybe even desired?) behavior. 

With that in mind, I have only one ask for now and that is: please explain the current behavior and how this PR changes it **in the description of this PR** so the community can give its opinion. Change the PR to "ready for review" once this is done and see if someone comments on that.

I'll take a look at the code later today.

_(FWIW I like the functionality that this PR brings)_

PR #1875, opened 2024-05-06
Move specific FindReplaceDialog Tests into the separated test-class

Wittmaxi:
Some tests in FindReplaceUITest.java are not general to all find/replace interfaces and are specific for the find/replace dialog. This commit moves these tests into the dedicated test class and provides the necessary interface in FindReplaceUITest

### DIFF
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
@@ -14,6 +14,8 @@
 package org.eclipse.ui.workbench.texteditor.tests;
 
 import static org.eclipse.ui.workbench.texteditor.tests.FindReplaceTestUtil.runEventQueue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeFalse;
@@ -37,7 +39,6 @@ import org.eclipse.jface.text.TextViewer;
 import org.eclipse.ui.internal.findandreplace.SearchOptions;
 
 public class FindReplaceDialogTest extends FindReplaceUITest<DialogAccess> {
-
 	@Override
 	public DialogAccess openUIFromTextViewer(TextViewer viewer) {
 		TextViewer textViewer= getTextViewer();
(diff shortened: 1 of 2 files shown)


tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -263,9 +249,6 @@ public void testReplaceAllInSelection() {
 		fTextViewer.setSelection(new TextSelection(4 + System.lineSeparator().length(), 8 + System.lineSeparator().length()));
 		initializeFindReplaceUIForTextViewer();
 
-		dialog.unselect(SearchOptions.GLOBAL);
-		dialog.unselect(SearchOptions.WHOLE_WORD);
```
HeikoKlare: Why do you remove this?

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -189,45 +189,31 @@ public void testChangeInputForIncrementalSearch() {
 	@Test
 	public void testFindWithWholeWordEnabledWithMultipleWords() {
 		initializeTextViewerWithFindReplaceUI("two words two");
-		dialog.setFindText("two");
+		dialog.select(SearchOptions.INCREMENTAL);
```
HeikoKlare, line 192: Is this change necessary? If yes, it may be worth to be provided in a separate PR with an according explanation.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
```diff
@@ -49,18 +49,18 @@ protected final void initializeTextViewerWithFindReplaceUI(String content) {
 		initializeFindReplaceUIForTextViewer();
 	}
 
-	private void openTextViewer(String content) {
+	public void openTextViewer(String content) {
```
HeikoKlare: Please only make these three methods `protected` instead of `public`.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corporation and others.
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
```
HeikoKlare: Please don't change the existing copyright information (in particular do not replace the company name).

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: I want to keep the changes in the Overlay-PR to a minimum. Since the Overlay does NOT have an incremental-option (all searches are incremental unless that is not possible), I need to add this option so both Interfaces perform the same operations.

As discussed, I will add a Test which is exclusive to the Dialog, which tests the same behavior but without Incremental

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceUITest.java
Wittmaxi: The FindReplaceOverlayTest still works correctly with those lines in. I will add them back

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: This was not an intended change and stems from a blind Copy-Paste

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
```diff
@@ -146,4 +147,87 @@ public void testShiftEnterReversesSearchDirectionDialogSpecific() {
...
+		initializeTextViewerWithFindReplaceUI("two words two");
+		DialogAccess dialog = getDialog();
+		dialog.setFindText("two");
+		dialog.select(SearchOptions.WHOLE_WORD);
+		dialog.select(SearchOptions.WRAP);
+		IFindReplaceTarget target= dialog.getTarget();
+
+		dialog.simulateEnterInFindInputField(false);
+		assertEquals(0, (target.getSelection()).x);
+		assertEquals(3, (target.getSelection()).y);
```
HeikoKlare, line 221: Is there any reason for changing the behavior of the test? This is the code of the original test
```suggestion
		assertEquals(0, (target.getSelection()).x);
		assertEquals(0, (target.getSelection()).y);
```

Wittmaxi: @HeikoKlare The proposed change would (,... and the old test did) simply fail, and that makes sense!

We start in a non-incremental position, set a find-text, press enter and expect the first occurrence of the text to be selected.

The reason there was some back-and-forth with this test is that at some point, I tested that "nothing is selected before pressing enter", which is not relevant to testing the whole-word behavior of the Dialog.

I COULD add a new test which tests for "what happens to the selection when I open the Dialog and do nothing" but I doubt that that is relevant enough, unless we spot bugs in the future

HeikoKlare: Thank you. The reason for the change seems sound to me, so it's acceptable to have it in this PR. I am just hestitant to combine a refactoring with a functional change, at least in case both are actually independent from each other (as in this case where one does not require the other).

> The proposed change would (,... and the old test did) simply fail, and that makes sense!

That's not correct, as the existing test obviously succeeds 🙂

HeikoKlare (approved):
Thank you!

HeikoKlare:
Failing test is unrelated and documented: #275

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/workbench/texteditor/tests/FindReplaceDialogTest.java
Wittmaxi: With proposed change I meant the proposed change to 

```java
		assertEquals(0, (target.getSelection()).x);
		assertEquals(0, (target.getSelection()).y);```

HeikoKlare: Yes, that proposed change leads to the exact behavior of the existing test, which does obviously not fail. The PR introduced a slight modification of the test behavior, but as mentioned, it's fine.

PR #1876, opened 2024-05-07
Mark COPY_LINE_UP and COPY_LINE_DOWN as state dependent

rubenporras:
We found that the "copy lines up" and "copy lines down" commands are disabled in our editors, and we found that this is because the editor is not modifiable when it is created. The commands do not become enabled when the editor does become modifiable, which happens once the file being edited is synchronized.

The solution is to mark this commands state dependent, as other similar commands such as move line up/down or the different cut line variants (and others just see the lines which are close to my additions).

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -5976,10 +5976,12 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 		markAsStateDependentAction(ITextEditorActionConstants.DELETE_LINE_TO_END, true);
 		markAsStateDependentAction(ITextEditorActionConstants.MOVE_LINE_UP, true);
 		markAsStateDependentAction(ITextEditorActionConstants.MOVE_LINE_DOWN, true);
+		markAsStateDependentAction(ITextEditorActionConstants.COPY_LINE_UP, true);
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@rubenporras : could you please describe the problem you are trying to fix?

rubenporras:
We found that the "copy lines up" and "copy lines down" commands are disabled in our editors, and we found that this is because the editor is not modifiable when it is created. The commands do not become enabled when the editor does become modifiable, which happens once the file being edited is synchronized.

The solution is to mark this commands state dependent, as other similar commands such as move line up/down or the different cut line variants (and others just see the lines which are close to my additions).

iloveeclipse:
@rubenporras : thanks. 
Could you please amend your commit and provide a proper commit message with that information. 
See https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations
After that please rebase it on latest master state and force push your commit to the github.

fedejeanne (changes requested):
> @rubenporras : thanks. Could you please amend your commit and provide a proper commit message with that information. See https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations After that please rebase it on latest master state and force push your commit to the github.

@rubenporras a proper commit message is not a copy-paste of the description of this PR, it's rather something that can be quickly understood while reading the history. Something like:

```
Make COPY_LINE_UP/DOWN and JOIN_LINES commands state dependent

This will (de)activate them accordingly together with the editor.
```

Please reword it

fedejeanne (approved):

rubenporras:
Thanks for the review. Could we merge it?

PR #1890, opened 2024-05-20
Handle the case when application has no focus in ApplicationPartService

laeubi:
Currently when the application has no focus there is also no active window child and in this case several actions still fail even though a part is available for perform actions.

This now handles the case of an application without focus with an empty optional to allow fall back to the part context. If that context has still no suitable part service implementation as a last resort the containing window context is used.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.workbench/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.workbench;singleton:=true
-Bundle-Version: 1.16.0.qualifier
+Bundle-Version: 1.16.100.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/ApplicationPartServiceImpl.java
+++ b/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/ApplicationPartServiceImpl.java
@@ -27,6 +27,7 @@ import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.e4.ui.model.application.ui.advanced.MPerspective;
 import org.eclipse.e4.ui.model.application.ui.advanced.MPlaceholder;
 import org.eclipse.e4.ui.model.application.ui.basic.MPart;
+import org.eclipse.e4.ui.model.application.ui.basic.MWindow;
 import org.eclipse.e4.ui.workbench.UIEvents;
 import org.eclipse.e4.ui.workbench.modeling.EModelService;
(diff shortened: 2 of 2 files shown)


tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
```diff
@@ -134,11 +137,24 @@ private void removeAllWorkingSets() {
...
+			}
+			try {
+				root.refreshLocal(IResource.DEPTH_INFINITE, null);
+				if (root.exists()) {
+					root.delete(true, null);
+				}
+			} catch (CoreException e1) {
+				TestPlugin.getDefault().getLog().log(e.getStatus());
+				throw createAssertionError(e);
+			}
```
fedejeanne: Are these changes related to this PR? Judging by the description of the PR, it doesn't look like they do.

fedejeanne (dismissed):
@laeubi thank you for contributing this. I only have a minor concern regarding the changes in the test class.

NOTE: I only skimmed through the the other changes but I haven't tested them. I could test them do it if you briefly describe how to do it.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dialogs/UIWorkingSetWizardsAuto.java
laeubi: The test seem to randomly fail so I added another commit here to see if it can be fixed.

fedejeanne: But the failing test doesn't have anything to do with `ApplicationPartServiceImpl` right?

laeubi: Only that it failed while building the PR once...

HeikoKlare: I would propose to keep this separate and have another PR that is also linked to the according issue: #275

fedejeanne: Ah, thank you @HeikoKlare , I was looking for the right issue to link here :-)

@laeubi please propose another PR for these changes.

fedejeanne (commented):
Looks better. I'm still missing a brief "How to test" in this PR and a description of the original bug/behavior that caused this PR.

laeubi:
Sadly the case is quite tricky to test, you need a custom RCP application, you need an external trigger (e.g. Network communication) ...

laeubi:
One additional test failure reported here:

- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2649

but it suceeds now.

The other test failure is already known:

- https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #1894, opened 2024-05-22
Sticky scrolling in Eclipse

### LINKED ISSUE #1719: Sticky Option for breadcrumb / additional to breadcrumb
Originally provided it as a feature-request within the SAP-Ecospace.  
As I suggested there, this is a feature, which will provide a better usability no matter what language is used. So moved it here and hope to get that feature.

Please describe your improvement request

Provide different option for the breadcrumb / add a feature

What is the opportunity/problem the request will address?

In long statements it's difficult to see, which are dependencies, and therefore it's difficult to add the logic to the correct place. Especial when the source-code gets long

What is the expected benefit?

Will give developers a better option to navigate and see what their focus is

Found it as an option within VSCode and think this option would also be a great benefit for ABAP. Maybe it's a feature for the eclipse foundation.

Have a look here

https://devblogs.microsoft.com/visualstudio/sticky-scroll-now-in-preview/

Christopher-Hermann:
**Description**
With this change, sticky scrolling is introduced into Eclipse text editors, as it is available for many other modern IDEs. Sticky scrolling will keep certain source code lines visible and in a fixed position on the screen as the user scrolls down the page. This technique improves user experience by keeping information within reach at all times.

The feature can be enabled via the TextEditor settings.

Provides feature for https://github.com/eclipse-platform/eclipse.platform.ui/issues/1719 and https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1364

**Current state**
There are some TODOs left, however the general functionality should work. I'm already creating the PR to get early feedback and maybe there are some volunteer beta testers.

**Testing**
Enable the feature in the preferences: General > Editors > Text Editors 
Check "Enable sticky scrolling" 

The sticky lines should be visible in all source/text editors and should work for every language since it is working on line indentation.
stickyScrolling


bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractDecoratedTextEditor.java
BeckerWdf, line 1: a lot of whitespace changes mix with "real" changes.
Can you separate the whitespace changes into a separate commit?

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractDecoratedTextEditorPreferenceConstants.java
BeckerWdf, line 1: Same whitespace change issues here.

AObuchow:
Just wanted to say this looks _awesome_ & a great modernization effort for the Eclipse IDE :) Great work

mickaelistria:
That's a very interesting proposal and the quality already seems good!
What defines which lines are to remain visible? Identation only?

iloveeclipse:
> What defines which lines are to remain visible? Identation only?

This seem to produce unexpected effects with code formatted like this:
```
    private void writeObject(java.io.ObjectOutputStream s)
        throws java.io.IOException
    {
```
see
image

iloveeclipse:
In general I believe adding first type in the file wastes space / just duplicates already shown information, as the first type is already shown in the editor tab title.

On Linux I see the right side of the "sticky" widget is shown one pixel over the scrollbar, as seen in the screenshot above.

Beside this, I would wish a bit different background for the sticky part, as it is hard to differentiate from actual code (also seen in the screenshot above).

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractDecoratedTextEditor.java
```diff
@@ -251,9 +255,11 @@ public void gotoMarker(IMarker marker) {
 	/**
 	 * The overview ruler of this editor.
 	 *
-	 * <p>This field should not be referenced by subclasses. It is <code>protected</code> for API
+	 * <p>
```
iloveeclipse: There is a lot of unrelated javadoc format changes. Please undo, it is hard to find code for a review

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractDecoratedTextEditor.java
```diff
@@ -330,6 +344,12 @@ public void gotoMarker(IMarker marker) {
 	 * @since 3.5
 	 */
 	private FocusedInformationPresenter fInformationPresenter;
+	/**
+	 * The handler for sticky scrolling
+	 * 
+	 * @since 4.3 //TODO ?
```
iloveeclipse: So not I've found what I was looking for. This must be bundle major.minor version, which brings us to the next issue: there is new API and that means, bundle version must be updated from `3.17` to `3.18` and this is what should be added here instead of `4.3`.

But the **private** fields are never APi, so at this place it is wrong. It should be at `AbstractDecoratedTextEditorPreferenceConstants.EDITOR_STICKY_SCROLLING_ENABLED`

BeckerWdf:
> That's a very interesting proposal and the quality already seems good! What defines which lines are to remain visible? Identation only?

I think yes. How do other IDEs solve this?

Christopher-Hermann:
> > What defines which lines are to remain visible? Identation only?
> 
> This seem to produce unexpected effects with code formatted like this:
> 
> ```
>     private void writeObject(java.io.ObjectOutputStream s)
>         throws java.io.IOException
>     {
> ```

> > That's a very interesting proposal and the quality already seems good! What defines which lines are to remain visible? Identation only?
> 
> I think yes. How do other IDEs solve this?

Yes, currently it is working with indentation only. At least VS Code does the exact same thing, see the EnumMap example: 
Bildschirmfoto 2024-05-23 um 08 21 46

But of course we can try to find a better solution. 
On the first look I thought about using the code folding information. But this has some drawbacks as for example Java doesn't provide information for if else statements. Furthermore, this won't work for language that doesn't provide code folding at all.

mickaelistria:
> On the first look I thought about using the code folding information

That's a good idea to get started from.
As we could already notice, indentation is sometimes not sufficient. While folding may not be perfect, its internals are a good inspiration: it does rely on document annotations (which are set by JDT when reconciling/parsing the Java file). A similar strategy allowing document annotations would be interesting here, because the problem "what to stick in" cannot be fully fixed with plain text logic, it seems language-specific and even code-style-specific.
So one possible approach could be to define a new "sticky" document annotation type in Platform and show in the sticky area the lines that have this sticky annotation. Then, language analyzers such as JDT can mark some lines as "sticky" according to language-specific logic. If no line is marked as sticky, indentation can be used as fall-back, with its limits.
What do you think?

Christopher-Hermann:
> So one possible approach could be to define a new "sticky" document annotation type in Platform and show in the sticky area the lines that have this sticky annotation. Then, language analyzers such as JDT can mark some lines as "sticky" according to language-specific logic. If no line is marked as sticky, indentation can be used as fall-back, with its limits. What do you think?

Yes, that's a good idea. I will finish the work on the layout and on the fallback logic based on indentation. Afterwards we can have a look how we can implement the annotation logic.

Madjosz:
grafik
Eclipse Theia seems to work better in this regard then than VS code as it sticks the correct lines with class and function names in 1TBS as well as in Allman style. The sticking of the `if` brace seems to be an artifact and only happens after restart. @Christopher-Hermann are you sure this behavior keeps appearing after the file is fully loaded and parsed? I see the same behaviour in Theia during loading of the file.

Regarding redundant `class` name stick one could argue that you can have multiple non-public top level classes where you want the stick.

Having sticky scroll use the folding functionality sounds like a good approach. Having conditionals not stick yet sounds more like a missing feature on the folding side to me.

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractDecoratedTextEditor.java
Christopher-Hermann: Done

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractDecoratedTextEditor.java
Christopher-Hermann: Done

Wittmaxi:
This is a very important feature that eclipse is lacking (and I get asked about a lot!).
I am very glad that somebody is working on it - I am willing to test it, if you need help with that :)

Christopher-Hermann:
> This is a very important feature that eclipse is lacking (and I get asked about a lot!). I am very glad that somebody is working on it - I am willing to test it, if you need help with that :)

Christopher-Hermann:
> On Linux I see the right side of the "sticky" widget is shown one pixel over the scrollbar, as seen in the screenshot above.

It seems that the layout of the scroll bar is different on the different operating systems. It would be nice to get a screenshot from a Windows user as well, so that I can address the issue.

On MacOS the scroll bar is transparent with the latest client:
ScrollBar

> Beside this, I would wish a bit different background for the sticky part, as it is hard to differentiate from actual code (also seen in the screenshot above).

Yes you are right, it's hard to see. Any color suggestions? I would like to reuse a existing color, but I cannot find a matching one.
Bildschirmfoto 2024-05-28 um 13 04 18

Christopher-Hermann:
> Having sticky scroll use the folding functionality sounds like a good approach. Having conditionals not stick yet sounds more like a missing feature on the folding side to me.

Yes, most probably you are right and this could also be improved in the code folding in general.

However, I propose merging this change as a "beta feature" after addressing layout questions and providing unit tests. Since the feature is disabled by default it should not have any impacts on the daily work. And having it in the official eclipse release will increase the amount of tester significantly. Even I don't use my target IDE in my daily work, so it's hard to test every single editor feature, like quick fixes, etc.

The sticky scrolling based on annotations/code folding can then be implemented on a separate PR.

Madjosz:
> It would be nice to get a screenshot from a Windows user as well, so that I can address the issue.

This is how it currently looks in Windows 11 light theme without OS scaling and dark theme with 125% OS scaling:
Sticky Scroll in Windows 11
The line number and code alignment is pixel perfect. What I don't like is the yellow background, that is too much.

Overall a very nice outcome, thank you for your efforts!

Christopher-Hermann:
> What I don't like is the yellow background, that is too much.

Yes, I played around a litte bit with the colors, but I don't like this as well. I removed the coloring and incorporated a second vertical separator beneath the sticky lines, which improved visibility slightly. However, there is still room for improvement.
image


An other option would be to color the line numbers:
Bildschirmfoto 2024-05-29 um 16 19 58

Christopher-Hermann:
The feature/pull request is now ready for review. The sticky lines calculation is fully reliant on the indentation of the source code. I propose we merge it as a "beta" feature to gather preliminary feedback, particularly from users coding in languages other than Java, which was my exclusive testing focus.

For future enhancements, the sticky scrolling functionality can be refined by introducing a sticky lines provider capable of interpreting code folding or working from a new annotation as suggested in a comment above. To allow for broader use across different languages, we could implement this with extension points.

The "final" state of the UI now contains two horizontal separators to make the separation between the sticky lines and the actual source code easier to see. Ideas for improvement are welcome:
Bildschirmfoto 2024-06-10 um 13 06 44

BeckerWdf:
> I propose we merge it as a "beta" feature to gather preliminary feedback, particularly from users coding in languages other than Java, which was my exclusive testing focus.

That's a good idea.

mickaelistria (changes requested):
I'm all in favor of merging ASAP, this is an interesting, opt-in feature with only internal code and just a couple of new symbols in API, and we're in M1 which is the best time for such addition without adding major risk on the release.
Please just add the copyright headers to new file before we merge.

tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyLinesProviderTest.java
```diff
@@ -0,0 +1,217 @@
+package org.eclipse.ui.internal.texteditor.stickyscroll;
```
iloveeclipse, line 14: Please add header comments to all new files

Wittmaxi:
As said above, I really like this feature and I absolutely want to use it!
I have tested your feature and will just dump all of my findings - no matter how big or little.
To me, 1), 2) and 3) are (almost) blockers, the rest are optional but should be taken care of (maybe as a follow-up).

1) Have you looked into this change in dark mode?
grafik

The double line in dark mode is really annoying to me, I would likely not use this feature as it is right now because of the annoying lines.

2) When enabling/disabling breadcrumbs, the sticky scrolling-window is not updated correctly


3) when doing a find/replace, the sticky lines don't update


4) (optional) Line annotations aren't shown in the sticky lines
grafik

5) Why is there a cap on the amount of sticky lines? (I agree that >10 is likely not practical)
grafik

6) in a big file, the amount of Sticky lines should automatically decrease so that you can always still see a part of the editor. Otherwise, it can be very confusing since you cannot scroll in or correctly interact with a the editor


7) Sticky scroll sometimes "skips a line" - it's not so easy to describe this and it's not that big of a deal, but it somehow feels unnatural that in some scenarios, two lines or more are stuck at once: 


(I know, this is quite an extreme and construed example)

8) (optional) Sticky scrolling in text files is weird:
grafik

9) I cannot scroll the editor text while inside of the sticky-lines area (and I would intuitively like to, especially when I have stuck 10 lines)

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/editors/text/TextEditorMessages.properties
```diff
@@ -69,6 +69,9 @@ TextEditorDefaultsPreferencePage_showWhitespaceCharactersLinkText= (<a>configure
 TextEditorDefaultsPreferencePage_showWhitespaceCharactersDialogInvalidInput=''{0}'' is not a valid input.
 TextEditorDefaultsPreferencePage_showWhitespaceCharactersDialogTitle=Show Whitespace Characters
 TextEditorDefaultsPreferencePage_space=Space ( \u00b7 )
+TextEditorDefaultsPreferencePage_stickyScrollingEnabled=Enable sticky scrolling
```
Wittmaxi:
```suggestion
TextEditorDefaultsPreferencePage_stickyScrollingEnabled=Enable stic&ky scrolling
```

k is still available for mnemonics

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyLinesProvider.java
```diff
@@ -0,0 +1,174 @@
...
+
+import org.eclipse.jface.text.ITextViewerExtension5;
+import org.eclipse.jface.text.source.ISourceViewer;
+
+/**
+ * This class provides sticky lines for the given source code in the source viewer. The
+ * implementation is completely based on indentation and therefore should work by default for
+ * several languages.
+ */
+public class StickyLinesProvider {
```
Wittmaxi, line 30: For the Future: how about we extract an IStickyLinesProvider and ask each editor to provide their own sticky lines provider? This fixes my comment about "normal text files" having weird behavior

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
```diff
@@ -0,0 +1,482 @@
...
+		try {
+			int offset= sourceViewer.getDocument().getLineOffset(clickedStickyLine.lineNumber());
+			sourceViewer.setSelectedRange(offset, 0);
+			enusreSourceViewerLineVisible(clickedStickyLine.lineNumber());
+		} catch (BadLocationException e) {
+			//Do not navigate
+		}
+	}
+
+	private void enusreSourceViewerLineVisible(int line) {
```
Wittmaxi:
```suggestion
	private void ensureSourceViewerLineVisible(int line) {
```

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
```diff
@@ -0,0 +1,482 @@
...
+	 * @param stickyLines The sticky lines to show
+	 */
+	public void setStickyLines(List<StickyLine> stickyLines) {
+		if (!stickyLines.equals(this.stickyLines)) {
+			this.stickyLines= stickyLines;
+			updateStickyScrollingControls();
+		}
+	}
+
+	public void setSettings(StickyScrollingControlSettings settings) {
```
Wittmaxi:
```suggestion
	public void applySettings(StickyScrollingControlSettings settings) {
```

But that's just preference ;)

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyLinesProvider.java
```diff
@@ -0,0 +1,174 @@
...
+			}
+		}
+	}
+
+	private void calculateStickyLinesUnderStickyLineControl(LinkedList<StickyLine> stickyLines, ISourceViewer sourceViewer, int startLine) {
+		int firstBelowControl= startLine + stickyLines.size();
+		StyledText textWidget= sourceViewer.getTextWidget();
+		int lineCount= textWidget.getLineCount();
+
+		for (int i= startLine; i < firstBelowControl && i < lineCount; i++) {
```
Wittmaxi, line 90: Is this a safe thing to do? For larger files, this can potentially become a problem.

Scrolling with Sticky Scrolling in a huge file is already not super smooth for me:

vs


A possible solution is to wait until scrolling is "done" to perform this analysis - and maybe do it in a separate thread (eclipse already has enough performance-issues in the UI-thread ;_;)

I don't believe this is too big of a deal since this feature can be opted out of easily

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyLinesProvider.java
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	private int mapLineNumberToSourceViewerLine(int lineNumber, ISourceViewer sourceViewer) {
+		if (sourceViewer instanceof ITextViewerExtension5 extension) {
+			return extension.widgetLine2ModelLine(lineNumber);
+		}
+		return lineNumber;
+	}
+
+	private int getStartIndetation(int startFromLine, StyledText styledText) {
```
Wittmaxi:
```suggestion
	private int getStartIndentation(int startFromLine, StyledText styledText) {
```

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
```diff
@@ -0,0 +1,482 @@
...
+		if (!newStickyLineText.equals(stickyLineText.getText())
+				|| !newStickyLineNumber.equals(stickyLineNumber.getText())) {
+			stickyLineText.setText(stickyLineTextJoiner.toString());
+			stickyLineNumber.setText(stickyLineNumberJoiner.toString());
+			styleStickyLines();
+			layoutStickyLines();
+		}
+	}
+
+	private String fillLineNumberWithLeadingSpaces(int lineNumber) {
```
Wittmaxi, line 204: Isn't there a way to use an [alignment option of SWT](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2Fcustom%2FStyledText.html) when displaying the line numbers instead of padding the line number with spaces?

Wittmaxi (approved):
The code looks good! I have added a few-nitpicks and there is one potential performance issue that I would like to hear your thoughts on :)

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
Christopher-Hermann: I tried to implement right alignment but ended up mirroring the logic used in the line number ruler, as it also lacks specified alignment. (see [LineNumberColumn.java](https://github.com/eclipse-platform/eclipse.platform.ui/blob/385b30bb74bce1bc048584281a7b4fdedc028456/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java#L546C2-L546C11)).

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyLinesProvider.java
Christopher-Hermann: Yes, I also though about that. But I wanted to wait until we have two implementations to have a better feeling what is needed before creating a new API.

Christopher-Hermann:
Thanks for your valuable review and input. This helps a lot.

> 1. Have you looked into this change in dark mode?
> 2. When enabling/disabling breadcrumbs, the sticky scrolling-window is not updated correctly

On MacOS it's working. I need to check how I can fix and test this on windows.


> 4. (optional) Line annotations aren't shown in the sticky lines

You mean the blue coloring, right? This does only makes sense for the last sticky line, since all other sticky lines will be related to another source code part. Not sure if it is worth to invest here.

> 5. Why is there a cap on the amount of sticky lines? (I agree that >10 is likely not practical)

Yes, it seemed to be unnecessary to have more than 10 lines. Of course we could remove this limitation, but on the other hand this goes in the same direction as your 6. comment.

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyLinesProvider.java
Christopher-Hermann: The for loop does only check up to the maximum amount of sticky lines. So this should not be the bottle neck. The performance problem is caused by the fast scrolling  as the sticky lines are recalculated for every moved pixel. 

Waiting until the scrolling is done is not a proper solution, I think this will look buggy / laggy.

I will try to either move the calculation in another thread or skip some calculations when the editor is scrolled really fast.

mickaelistria (approved):

mickaelistria:
I'm merging this right now because the current code and feature quality are already good and it's the right time to push such experimental features in.
@Wittmaxi Can you please open dedicated issues for the point you've mentioned above that still require some improvements?

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyLinesProvider.java
Christopher-Hermann: See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1949

jukzi:
I need to get used to it, but looks promising. Just annoying that it feels like the row count is sometimes +-1 line off
Always for (empty) lines after javadoc before a method:
image

In this example it is unsure to which blockstart the block belongs, depends how much pixels exactly i scroll:
image
image

PR #1921, opened 2024-06-03
Find/Replace overlay: consistent handling of incremental base location (#1913 #1914)

### LINKED ISSUE #1913: [find/replace logic] cannot easily reset the Incremental Base-Location
[comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192#issuecomment-2117826351)

FindReplaceLogic internally keeps track of it's "incremental base location" - when starting a search from a fresh environment, it needs to be set to the current cursor location. However, FindReplaceLogic does not have logic for this, instead only relying on updating the base location when Incremental search is enabled or disabled.
FindReplaceLogic should expose a method for updating an incremental base location instead of relying on a trick to update the Incremental Base Location.

### LINKED ISSUE #1914: [find/replace overlay] incremental base position resets randomly when using "search in selection"
[comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192#pullrequestreview-2085146414) by @HeikoKlare 

When activating and deactivating the "selected lines" option, the incremental base position seems to be reset (or not set properly).
grafik

### LINKED ISSUE #1915: [find/replace overlay] disabling RegEx resets to the wrong base-location
[comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192#pullrequestreview-2090057293) by @HeikoKlare 

disabling the regex search option leads to some unexpected search match being highlighted. If I am not mistaken, it is the one after the incremental base position before enabling the regex option:
grafik

This should be taken into account (and may even implicitly be addressed) when refactoring the way in which the incremental base position is set from outside the FindReplaceLogic. If I understand correctly, the position is reset by disabling and enabling the incremental search option, but that is only a side effect (that does not always apply, like when the regex option is enabled). Probably, resetting the incremental base position when the logic is not (or even cannot) be aware of that necessity should be done in a more explicit way.

Wittmaxi:
Increment the incremental base location consistently, remove method for
"performIncrementalSearch" as it only called "performSearch" and lead to
semantic confusion while calling "performSearch".
- "SearchOptions.REGEX" will now correctly update the incremental base
location once it is disabled.
- Added a method for updating the incremental base location manually
  from outside

fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1914
fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1913
fixes #1915

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -66,8 +66,8 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 			break;
 		case FORWARD:
 		case INCREMENTAL:
-			if (shouldInitIncrementalBaseLocation()) {
-				initIncrementalBaseLocation();
+			if (shouldResetIncrementalBaseLocation()) {
+				resetIncrementalBaseLocation();
 			}
 			break;
 		// $CASES-OMITTED$
@@ -82,12 +82,16 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 			return;
 		}
 
+		if (searchOption == SearchOptions.REGEX) {
+			resetIncrementalBaseLocation();
+		}
+
 		if (searchOption == SearchOptions.GLOBAL) {
 			initializeSearchScope();
 		}
 
-		if (searchOption == SearchOptions.FORWARD && shouldInitIncrementalBaseLocation()) {
-			initIncrementalBaseLocation();
+		if (searchOption == SearchOptions.FORWARD && shouldResetIncrementalBaseLocation()) {
+			resetIncrementalBaseLocation();
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -359,6 +352,7 @@ private boolean performSearch(boolean mustInitIncrementalBaseLocation, String fi
 				// we don't keep state in this dialog
```
CodeLtDave, line 352: is this unhandled catch intended?

Wittmaxi: This code is directly copied from https://github.com/eclipse-platform/eclipse.platform.ui/blob/01c5b676196b84ed721cfa43267079374abf4498/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java#L1544

As stated in the code-comment, the dialog does not keep state + the code compiles just as well without this catch-clause.
Even if this was problematic, changing this here is not in the scope of this PR. If you believe this is a big enough poblem, please create an issue for it.

HeikoKlare (commented):
I have tested the changes and made two findings:

1. Activating "selection in area" does not work properly anymore
find_replace_selectinline

2. When deactivating a search option that disables incremental search (in particular regex search) and the currently selected word still matches the search specification, still a new search is triggered, moving the selection to the next match. This behavior already existed before this PR but since this PR affects setting the incremental base position, it should probably be considered here.
find_replace_disableregex

HeikoKlare:
Closing this as it has been split up into and replaced by #2104, #2106, #2108
Mentioned issues have also been addressed by those PRs.

PR #1922, opened 2024-06-03
Show recently used "File > New" wizards

### LINKED ISSUE #1530: Support New > Recently Used

lathapatil:
Changes requested in the PR
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1837 are addressed here

Fixes : #1530

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/NewWizardMenu.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/NewWizardMenu.java
@@ -15,8 +15,12 @@
 package org.eclipse.ui.actions;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 
 import org.eclipse.jface.action.ActionContributionItem;
 import org.eclipse.jface.action.IAction;
@@ -28,6 +32,7 @@ import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.activities.WorkbenchActivityHelper;
 import org.eclipse.ui.internal.WorkbenchPlugin;
 import org.eclipse.ui.internal.actions.NewWizardShortcutAction;
+import org.eclipse.ui.internal.dialogs.RecentNewWizardSelection;
 import org.eclipse.ui.internal.dialogs.WorkbenchWizardElement;
 import org.eclipse.ui.internal.registry.WizardsRegistryReader;
 import org.eclipse.ui.wizards.IWizardCategory;
@@ -186,9 +191,28 @@ public class NewWizardMenu extends BaseNewWizardMenu {
 			list.add(new ActionContributionItem(newExampleAction));
 			list.add(new Separator());
 		}
+		// To add shortcuts from OTHER... wizard regardless of perspective
+		Collection<IContributionItem> actions = new LinkedList<>();
+		if (!RecentNewWizardSelection.getInstance().getMenuIds().isEmpty()) {
+			for (String selectedItem : RecentNewWizardSelection.getInstance()
+					.getMenuIds()) {
+				IAction action = getAction(selectedItem);
+				actions.add(new ActionContributionItem(action));
+			}
+			subMenuShortcutCheck(list, actions);
+		}
 		list.add(new ActionContributionItem(getShowDialogAction()));
 	}
 
+	private void subMenuShortcutCheck(List<IContributionItem> list, Collection<IContributionItem> otherItems) {
+		Set<IContributionItem> existingShortcutsInPerspective = new HashSet<>(list);
+		for (IContributionItem item : otherItems) {
+			if (!existingShortcutsInPerspective.contains(item)) {
+				list.add(item);
+				existingShortcutsInPerspective.add(item);
+			}
+		}
+	}
 	private boolean isNewProjectWizardAction(IAction action) {
 		if (action instanceof NewWizardShortcutAction) {
 			IWizardDescriptor wizardDescriptor= ((NewWizardShortcutAction) action).getWizardDescriptor();
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/actions/BaseNewWizardMenu.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/actions/BaseNewWizardMenu.java
@@ -163,10 +163,12 @@ public class BaseNewWizardMenu extends CompoundContributionItem {
 		}
 	}
 
-	/*
+	/**
 	 * Returns the action for the given wizard id, or null if not found.
+	 *
+	 * @since 3.134
 	 */
-	private IAction getAction(String id) {
+	public IAction getAction(String id) {
 		// Keep a cache, rather than creating a new action each time,
 		// so that image caching in ActionContributionItem works.
 		IAction action = actions.get(id);
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
@@ -53,6 +53,8 @@ import org.eclipse.ui.ISharedImages;
 import org.eclipse.ui.IWorkingSetManager;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.internal.decorators.DecoratorManager;
+import org.eclipse.ui.internal.dialogs.NewWizard;
+import org.eclipse.ui.internal.dialogs.NewWizard.RecentNewWizardsPreferenceManager;
 import org.eclipse.ui.internal.dialogs.WorkbenchPreferenceManager;
 import org.eclipse.ui.internal.help.CommandHelpServiceImpl;
 import org.eclipse.ui.internal.help.HelpServiceImpl;
@@ -201,6 +203,8 @@ public class WorkbenchPlugin extends AbstractUIPlugin {
 	private EHelpService helpService;
 
 	private ICommandHelpService commandHelpService;
+
+	private NewWizard.RecentNewWizardsPreferenceManager recentNewWizardsPreferenceManager;
 
 	/**
 	 * Create an instance of the WorkbenchPlugin. The workbench plugin is
@@ -766,6 +770,10 @@ public class WorkbenchPlugin extends AbstractUIPlugin {
 			// to be loaded.s
 			if (uiBundle != null)
 				uiBundle.start(Bundle.START_TRANSIENT);
+
+			recentNewWizardsPreferenceManager = new RecentNewWizardsPreferenceManager();
+			recentNewWizardsPreferenceManager.populate();
+
 		} catch (BundleException e) {
 			WorkbenchPlugin.log("Unable to load UI activator", e); //$NON-NLS-1$
 		}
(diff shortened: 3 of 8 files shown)


iloveeclipse:
@lathapatil : could you please provide some screenshot showing the change?

lathapatil:
@iloveeclipse Yes, PFB for recently used 5 new shortcuts

recentlyUsedNewWizards

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/NewWizardMenu.java
```diff
@@ -186,9 +191,28 @@ protected void addItems(List<IContributionItem> list) {
 			list.add(new ActionContributionItem(newExampleAction));
 			list.add(new Separator());
 		}
+		// To add shortcuts from OTHER... wizard regardless of perspective
+		Collection<IContributionItem> otherItems = new LinkedList<>();
+		if (!RecentNewWizardSelection.getInstance().getSelectedFromOther().isEmpty()) {
```
iloveeclipse: Method name `getSelectedFromOther()` and derived variable names are unfortunate, I had no idea how to understand these names.
Why not just `RecentNewWizardSelection.getInstance().getItems()` ?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/actions/BaseNewWizardMenu.java
```diff
@@ -163,10 +163,12 @@ public void dispose() {
 		}
 	}
 
-	/*
+	/**
 	 * Returns the action for the given wizard id, or null if not found.
+	 *
+	 * @since 3.132
```
iloveeclipse: 1) Please make the method **package protected**, as we don't expect anyone to overwrite it.
2) Please remove "since" comment after making the method package protected, as it would not be new API anymore.
3) **If** this would be changed to "public" as proposed, this would be wrong version, API tooling shows an error in the IDE. It should be `3.134` and MANIFESt.MF version should be changed from `3.133.100` to `3.134.0`, because new API method would be added.

@laeubi, @HannesWell : looks like API tooling validation on jenkins stopped to work (second time I see it after https://github.com/eclipse-platform/eclipse.platform.ui/issues/2430). 
Could you please check whether is that tycho, PDE or jenkins/maven configuration issue?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
```diff
@@ -210,6 +217,7 @@ public class WorkbenchPlugin extends AbstractUIPlugin {
 	public WorkbenchPlugin() {
 		super();
 		inst = this;
+		recentNewWizardsPreferenceManager = new RecentNewWizardsPreferenceManager();
```
iloveeclipse: Never ever do anything in bundle constructor! Any bundle initialization code should be added in the `start()` method.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
```diff
@@ -766,6 +774,10 @@ public void start(BundleContext context) throws Exception {
 			// to be loaded.s
 			if (uiBundle != null)
 				uiBundle.start(Bundle.START_TRANSIENT);
+
+			List<String> recentlyUsedNewWizards = recentNewWizardsPreferenceManager.getMenuShortcutsFromPreferences();
```
iloveeclipse: This code doesn't belong to the bundle activator, it is an internal implementation detail of the RecentNewWizardsPreferenceManager and should be done there.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchPlugin.java
```diff
@@ -1048,6 +1060,9 @@ public void stop(BundleContext context) throws Exception {
 			testableTracker.close();
 			testableTracker = null;
 		}
+		// Store recently used new page shortcuts to preferences
+		Set<String> selectedFromOther = RecentNewWizardSelection.getInstance().getSelectedFromOther();
```
iloveeclipse: This code doesn't belong to the bundle activator, it is an internal implementation detail of the RecentNewWizardsPreferenceManager and should be done there. What yuo could do is to call RecentNewWizardsPreferenceManager.shutdown() here.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/NewWizard.java
```diff
@@ -169,4 +175,37 @@ public boolean canFinish() {
 		return super.canFinish();
 	}
 
+	/**
+	 * A <code>RecentNewWizardsPreferenceManager</code> is used to store and fetch
+	 * the most recent new wizards used or created from the "Other" shortcut, which
+	 * is part of the menu manager with New Wizard actions, from preferences.
+	 *
+	 * @since 3.5
```
iloveeclipse: please remove "since" as the package is not API

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/NewWizard.java
```diff
@@ -169,4 +175,37 @@ public boolean canFinish() {
...
+		private static final String RECENT_NEW_MENU_ITEMS = "recentNewMenuItems"; //$NON-NLS-1$
+		private static final String SEPARATOR = ","; //$NON-NLS-1$
+
+		public List<String> getMenuShortcutsFromPreferences() {
+			IEclipsePreferences pref = InstanceScope.INSTANCE.getNode(PLUGIN_ID);
+			String preferences = pref.get(RECENT_NEW_MENU_ITEMS, null);
+			if (preferences != null && !preferences.trim().isEmpty()) {
+				return Arrays.asList(preferences.split(SEPARATOR));
+			}
+			return new ArrayList<>();
```
iloveeclipse: This returns modifiable list, the line above not modifiable. Please use `List.of()` in both cases to be consistent

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/NewWizard.java
```diff
@@ -169,4 +175,37 @@ public boolean canFinish() {
...
+		}
+
+		public void setMenuShortcutsToPreferences(Set<String> items) {
+			IEclipsePreferences pref = InstanceScope.INSTANCE.getNode(PLUGIN_ID);
+			pref.put(RECENT_NEW_MENU_ITEMS, String.join(SEPARATOR, items));
+			try {
+				pref.flush();
+			} catch (org.osgi.service.prefs.BackingStoreException e) {
+				e.printStackTrace();
```
iloveeclipse: Please never ever use `e.printStackTrace();` in bundle code, we have **logging** API. As you are in the workbench, you could simply do `WorkbenchPlugin.log(e);`

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/NewWizardNewPage.java
```diff
@@ -689,4 +689,11 @@ public IWorkbenchWizard createWizard() throws CoreException {
 
 		updateDescription(selectedObject);
 	}
+
+	/**
+	 * @return Returns the selectedElement.
```
iloveeclipse: Comment is meaningless. Instead, one could provide information that the selection could be null

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/NewWizardSelectionPage.java
```diff
@@ -94,6 +94,7 @@ public void createControl(Composite parent) {
 	 * they will persist into the next invocation of this wizard page
 	 */
 	protected void saveWidgetValues() {
+		RecentNewWizardSelection.getInstance().addItem(newResourcePage.getSelectedElement().getId());
```
iloveeclipse: `newResourcePage.getSelectedElement()` may return null, so please add check before.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/RecentNewWizardSelection.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
+ * A <code>RecentNewWizardSelection</code> is used to manage the five most
+ * recent new wizards used or created from the "Other" shortcut, which is part
+ * of the menu manager with New Wizard actions. The recently used new wizards
+ * will appear before the "Other" shortcut.
+ *
+ * @since 3.5
```
iloveeclipse: please remove "since", this package is not API

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/RecentNewWizardSelection.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class RecentNewWizardSelection {
+	private Set<String> selectedFromOther = new LinkedHashSet<>();
+	private static RecentNewWizardSelection instance;
+	private static final int MAX_MENU_SIZE = 5;
```
iloveeclipse, line 31: If you intended to have 5 recent items: currently only **four** are kept.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/RecentNewWizardSelection.java
```diff
@@ -0,0 +1,69 @@
...
+		}
+	}
+
+	/**
+	 * Returns the set of recently used new wizard menu shortcut IDs.
+	 *
+	 * @return the set of recently used new wizard menu shortcut IDs
+	 */
+
+	public Set<String> getSelectedFromOther() {
```
iloveeclipse: I would propose to rename `getSelectedFromOther` to `getItems()` or `getMenuIds()` or `getIds()`

iloveeclipse (changes requested):

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/RecentNewWizardSelection.java
lathapatil: > If you intended to have 5 recent items: currently only **four** are kept.

Could you please share the scenario you tested .
 It do not add the shortcuts if it is already present as part of fixed shortcuts (highlighted in the plugin perspective  image) of current perspective . The last used shortcuts change if you switch the perspective .
 
 Java perspective 
 
image

When I switch to plugin perspective ..
 
image

PR #1923, opened 2024-06-03
Find/Replace dialog: remove ring-tone #1908

### LINKED ISSUE #1908: [find/replace dialog] remove Ringtone
which could not be addressed into the current proposal:
[comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192#issuecomment-2033690701) by @HeikoKlare and @vogella 

In the PR for the Overlay, we decided to remove the Ringtone notifying of errors while searching, because it was perceived as annoying.
We should also remove it from the Dialog.

Wittmaxi:
Removes the beep which is issued after some unsuccessful operations in
the find/replace dialog.

fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1908

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor; singleton:=true
-Bundle-Version: 3.17.400.qualifier
+Bundle-Version: 3.17.500.qualifier
 Bundle-Activator: org.eclipse.ui.internal.texteditor.TextEditorPlugin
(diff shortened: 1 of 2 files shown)


HeikoKlare (commented):
I absolutely like the proposal to remove the beep functionality (and I think particularly @vogella will do as well 🙂).

Just two comments on the proposed changes:
1. The removal seems incomplete to me, as the `allowBeep` parameter in `evaluateFindReplaceStatus` is now unused. In my opinion, this should be cleaned up throughout the call hierarchy.
2. This change does not seem to depend on #1192 but currently contains the commit of that PR. You could rebase the single commit relevant for this PR onto master instead of the other PR.

Wittmaxi:
@HeikoKlare I believe that the issues you just mentionned are addressed

HeikoKlare (approved):
Looks good now, thank you! I have added a version bump for 4.33 stream to see if builds run fine. In case this is merged after some other PR affecting the same project (like #1192), the version bump will be automatically removed via rebase.

HeikoKlare:

PR #1924, opened 2024-06-03
Don't use IFileInfo.toString() in UI, use IFileStore.toString()

### LINKED ISSUE #1404: "Project deleted" dialog shows redundant / meaningless information
Noticed after the merge of https://github.com/eclipse-platform/eclipse.platform/pull/1329.

Steps:
- create new project (any kind, I've used Java)
- exit Eclipse
- delete (or move) project on file system
- start Eclipse
- Hit "F5" for refresh in "Package Explorer" (nothing should be selected to refresh entire workspace)
- Dialog is shown: 
image
- Dialog was shown before https://github.com/eclipse-platform/eclipse.platform/pull/1329: 
image

In both cases information inside brackets is redundant and makes no sense.
The original intent was to show **project path** but that was broken via 92d92ae5e9dbd517d0951c1da7cfbbd76eb9b275.

I have a fix, the actual dialog should look like:
image

iloveeclipse:
The `IFileInfo.toString()` isn't intended to be shown to user, it has no
information interesting for users.

The `IFileStore.toString()` should be used in cases if information about
deleted resources need to be shown in UI.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1404

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/RefreshAction.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/RefreshAction.java
@@ -16,11 +16,13 @@
 package org.eclipse.ui.actions;
 
 import java.lang.reflect.InvocationTargetException;
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
 import org.eclipse.core.filesystem.IFileInfo;
+import org.eclipse.core.filesystem.IFileStore;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceRuleFactory;
@@ -113,11 +115,13 @@ public class RefreshAction extends WorkspaceAction {
 		if (!project.exists()) {
 			return;
 		}
-		IFileInfo location = IDEResourceInfoUtils.getFileInfo(project.getLocationURI());
+		URI locationURI = project.getLocationURI();
+		IFileInfo location = IDEResourceInfoUtils.getFileInfo(locationURI);
 		if (!location.exists()) {
+			String displayedProjectPath = toDisplayPath(locationURI);
 			String message = NLS.bind(
 					IDEWorkbenchMessages.RefreshAction_locationDeletedMessage,
-					project.getName(), location.toString());
+					project.getName(), displayedProjectPath);
 
 			final MessageDialog dialog = new MessageDialog(getShell(),
 					IDEWorkbenchMessages.RefreshAction_dialogTitle, // dialog
@@ -141,6 +145,11 @@ public class RefreshAction extends WorkspaceAction {
 				project.delete(true, true, null);
 			}
 		}
+	}
+
+	private static String toDisplayPath(URI locationURI) {
+		IFileStore fileStore = IDEResourceInfoUtils.getFileStore(locationURI);
+		return fileStore != null ? fileStore.toString() : locationURI.toString();
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/RefreshAction.java
```diff
@@ -113,11 +115,14 @@ void checkLocationDeleted(IProject project) throws CoreException {
...
-		IFileInfo location = IDEResourceInfoUtils.getFileInfo(project.getLocationURI());
+		URI locationURI = project.getLocationURI();
+		IFileInfo location = IDEResourceInfoUtils.getFileInfo(locationURI);
 		if (!location.exists()) {
+			IFileStore fileStore = IDEResourceInfoUtils.getFileStore(locationURI);
+			String projectPath = fileStore != null ? fileStore.toString() : location.getName();
 			String message = NLS.bind(
 					IDEWorkbenchMessages.RefreshAction_locationDeletedMessage,
-					project.getName(), location.toString());
+					project.getName(), projectPath);
```
Wittmaxi: I don't see a "quick" way to separate this UI code from the logic, but at least it would make sense to extract this into a method, like this:

```java
if (!location.exists()) {
    String displayedProjectPath = generateDisplayedPath(project, location, filestore);
    String message = NLS.bind(
					IDEWorkbenchMessages.RefreshAction_locationDeletedMessage,
					project.getName(), projectPath);
```

What do you think?

iloveeclipse: Sure, will change.

jukzi:
@iloveeclipse  could you please also fix the Error message and title: it is not a "location" that has been deleted but "Project has been deleted from file system"
image
Did you also check that Locattion.toString() is not used on other places as well?

iloveeclipse:
> it is not a "location" that has been deleted but "Project has been deleted from file system"

Actually the message is correct, as the project is still there in the metadata, but its location isn't.

jukzi:
> Actually the message is correct, as the project is still there in the metadata, but its location isn't.

Its just the wording: I mean it is in general never possible to create or delete a **location**. One can delete a file or directory but a location exits independently from its content.

iloveeclipse:
> Its just the wording: I mean it is in general never possible to create or delete a **location**. One can delete a file or directory but a location exits independently from its content.

Not sure if our understanding is same. The "location" is the "address" (path) of the files/folders which are deleted now. That location describes something that is deleted. So one can say "location is deleted" if the underlined file/folder is deleted, at least that wording used here wasn't questioned for years so far.

Anyway, if you have a better proposal, please provide an example how it should look like.

jukzi:
> Anyway, if you have a better proposal, please provide an example how it should look like.

maybe just skip the word "location" like:

Headline: `Deleted Project`
msg: `Project $1 has been deleted from '$2'. Delete it from workspace?`

iloveeclipse:
What about this:
image

jukzi:
image
The Headline is very understandable now, thanks. But could you please change "Location" to "Directory"? it is a directory, isn't it?

merks:
Is it a directory, a folder, or a project?  In the Window Explorer there is a context menu New -> Folder so if we're going to split hairs...

jukzi:
At least in german i know Microsoft has changed the wording sometimes and on linux they are releated concepts, so there might be not a single optimal word. But the used org.eclipse.core.filesystem.IFileInfo.exists() calls it "directory" so that seem to be more appropriate then "location".

iloveeclipse:
OK, last bid:
image

jukzi:
> OK, last bid:

thank you, that looks fine.

iloveeclipse:
The one test fail on Jenkins is unstable, created https://github.com/eclipse-platform/eclipse.platform.ui/issues/1935.

PR #1927, opened 2024-06-04
fix warning about Field.isAccessible()

jukzi:
'The method isAccessible() from the type AccessibleObject is deprecated since version 9'

tested with
 WorkbenchThemeChangedHandlerTest,
 org.eclipse.ui.tests.commands.HelpContextIdTest.testHelpContextId()

### DIFF
--- a/bundles/org.eclipse.core.commands/src/org/eclipse/core/internal/commands/util/Util.java
+++ b/bundles/org.eclipse.core.commands/src/org/eclipse/core/internal/commands/util/Util.java
@@ -255,14 +255,16 @@ public final class Util {
 
 		String contextId = null;
 		if (method != null) {
-			boolean accessible = method.isAccessible();
-			method.setAccessible(true);
-			try {
-				contextId = (String) method.invoke(command);
-			} catch (Exception e) {
-				// do nothing
+			boolean accessible = method.canAccess(command);
+			if (method.trySetAccessible()) {
+				try {
+					contextId = (String) method.invoke(command);
+				} catch (Exception ignored) {
+					// do nothing
+				} finally {
+					method.setAccessible(accessible);
+				}
 			}
-			method.setAccessible(accessible);
 		}
 		return contextId;
 	}
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -1254,14 +1254,15 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 		protected Object getFieldValue(Field field) {
 			Object value = null;
 			if (field != null) {
-				boolean accessible = field.isAccessible();
-				try {
-					field.setAccessible(true);
-					value = field.get(instance);
-				} catch (Exception exc) {
-					// do nothing
-				} finally {
-					field.setAccessible(accessible);
+				boolean accessible = field.canAccess(instance);
+				if (field.trySetAccessible()) {
+					try {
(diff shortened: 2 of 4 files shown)


bundles/org.eclipse.core.commands/src/org/eclipse/core/internal/commands/util/Util.java
```diff
@@ -255,10 +255,11 @@ public static final String getHelpContextId(Command command) {
...
-			method.setAccessible(true);
+			boolean accessible = method.canAccess(command);
 			try {
-				contextId = (String) method.invoke(command);
+				if (method.trySetAccessible()) {
+					contextId = (String) method.invoke(command);
+				}
 			} catch (Exception e) {
 				// do nothing
 			}
```
sratz, line 267: wouldn't the finally block also need be adjusted?

sratz: Also: `Command#getHelpContextId()` says clients shally use `CommandManager#getHelpContextId(Command)` instead. Can't we call that here?

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -1254,10 +1254,11 @@ public ReflectionSupport(T instance) {
...
+				boolean accessible = field.canAccess(instance);
 				try {
-					field.setAccessible(true);
-					value = field.get(instance);
+					if (field.trySetAccessible()) {
+						value = field.get(instance);
+					}
 				} catch (Exception exc) {
 					// do nothing
 				} finally {
```
sratz: finally block is not symmetric anymore. I think the finally can be moved inside the `if(try)` block.

bundles/org.eclipse.core.commands/src/org/eclipse/core/internal/commands/util/Util.java
jukzi: there is no final block here, `CommandManager does a lot more then just calling the method. I don't want to break anything.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
jukzi: i don't get what you mean with "symmetric" here or how moving the try block should help, but OK, won't hurt either.

bundles/org.eclipse.core.commands/src/org/eclipse/core/internal/commands/util/Util.java
sratz: ah, right, there is no finally block here. but the same line of thought applies:

The condition is changed from `isAccessible()` to `canAccess()`, and now the defensive `if(tryAccess())` is used, but in line 266 `setAccessible()` is still called unconditionally which can then still fail.

I propose to also move that 'undo/cleanup' line inside the `if(try)` block, since otherwise there is nothing to undo and it could needlessly raise an exception.

jukzi: OK

jukzi:
unrelated random fails:
`12:22:53.295 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.8:verify (verify) on project org.eclipse.ui.ide: Execute ApiApplication failed: InvocationTargetException: Component 'Eclipse IDE UI' in the baseline 'workspace' is disposed: Component was disposed here -> [Help 1]`
https://github.com/eclipse-pde/eclipse.pde/issues/553

PR #1928, opened 2024-06-04
fix warning about java.lang.Class.newInstance()

jukzi:
'The method newInstance() from the type Class<capture#3-of ?> is deprecated since version 9'

tested by
org.eclipse.e4.ui.tests.css.core.parser.ValueTest

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/sac/SACParserFactoryImpl.java
+++ b/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/sac/SACParserFactoryImpl.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.e4.ui.css.core.impl.sac;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.HashMap;
 import java.util.Map;
 import org.eclipse.e4.ui.css.core.SACConstants;
@@ -49,7 +50,11 @@ public class SACParserFactoryImpl extends SACParserFactory {
 		String classNameParser = parsers.get(name);
 		if (classNameParser != null) {
 			Class<?> classParser = super.getClass().getClassLoader().loadClass(classNameParser);
-			return (Parser) classParser.newInstance();
+			try {
+				return (Parser) classParser.getDeclaredConstructor().newInstance();
+			} catch (InvocationTargetException | NoSuchMethodException e) {
+				throw (InstantiationException) new InstantiationException(classNameParser).initCause(e);
+			}
 		}
 		throw new IllegalAccessException("SAC parser with name=" + name
 				+ " was not registered into SAC parser factory.");
--- a/bundles/org.eclipse.e4.ui.css.core/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.css.core/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-SymbolicName: org.eclipse.e4.ui.css.core;singleton:=true
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
-Bundle-Version: 0.14.400.qualifier
+Bundle-Version: 0.14.500.qualifier
 Export-Package: org.eclipse.e4.ui.css.core;x-internal:=true,
  org.eclipse.e4.ui.css.core.css2;x-friends:="org.eclipse.e4.ui.css.swt.theme,org.eclipse.e4.ui.css.swt,org.eclipse.e4.ui.css.jface",
  org.eclipse.e4.ui.css.core.dom;x-friends:="org.eclipse.e4.ui.css.swt,org.eclipse.ui.views.properties.tabbed,org.eclipse.ui.forms",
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/sac/SACParserFactoryImpl.java
```diff
@@ -49,7 +50,12 @@ public Parser makeParser(String name) throws ClassNotFoundException, IllegalAcce
 		String classNameParser = parsers.get(name);
 		if (classNameParser != null) {
 			Class<?> classParser = super.getClass().getClassLoader().loadClass(classNameParser);
-			return (Parser) classParser.newInstance();
+			try {
+				return (Parser) classParser.getConstructor().newInstance();
```
sratz: javadoc says to use `getDeclaredConstructor()`

bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/sac/SACParserFactoryImpl.java
```diff
@@ -49,7 +50,12 @@ public Parser makeParser(String name) throws ClassNotFoundException, IllegalAcce
...
+			} catch (InstantiationException | IllegalArgumentException | InvocationTargetException | SecurityException
+					| NoSuchMethodException e) {
+				throw new ClassNotFoundException("Constructor not available", e);
```
sratz: _Constructor not available_ is not the correct message in all those potential exception cases.

In fact, it feels strange that `ISACParserFactory` lists all those exceptions individually.

```
	public abstract Parser makeParser(String name)
			throws ClassNotFoundException, IllegalAccessException,
			InstantiationException, NullPointerException, ClassCastException;
```

We already have so many, can't we just add the bunch of additional ones there too? They were thrown anyways unchecked by the deprecated `newInstance()` method. How many callers would need to be adapted?

Could also be reduced to just `InvocationTargetException` + `ReflectiveOperationException`.

bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/sac/SACParserFactoryImpl.java
jukzi: The difference is if the constructor has to be public. As the Constructor is public anyway it does not make a difference -but OK, we can use getDeclaredConstructor

bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/sac/SACParserFactoryImpl.java
jukzi: The affected packages are not named "internal" but marked x-internal, so we can just change the declaration

jukzi: oh not that simple - the list comes from 3rd party lib org.w3c.css.sac.helpers.ParserFactory.makeParser()

bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/sac/SACParserFactoryImpl.java
```diff
@@ -49,7 +51,12 @@ public Parser makeParser(String name) throws ClassNotFoundException, IllegalAcce
 		String classNameParser = parsers.get(name);
 		if (classNameParser != null) {
 			Class<?> classParser = super.getClass().getClassLoader().loadClass(classNameParser);
-			return (Parser) classParser.newInstance();
+			try {
+				return (Parser) classParser.getDeclaredConstructor().newInstance();
+			} catch (InstantiationException | IllegalArgumentException | InvocationTargetException | SecurityException
+					| NoSuchMethodException e) {
+				throw new ParserNotFoundException(e);
```
sratz: what about using `InstantiationException` to imiate the old behavior of the deprecated method?

```
		Constructor<Parser> constructor;
		try {
			constructor = classParser.getDeclaredConstructor();
		} catch (NoSuchMethodException e) {
			InstantiationException ie = new InstantiationException(classParser.getName());
			ie.initCause(e);
			throw ie;
		}
		try {
			return constructor.newInstance();
		} catch (InvocationTargetException e) {
			throw new IllegalStateException(e);
		}`
```

jukzi: that initCause() is a nice trick! an extra clause for the InvocationTargetException  seems to be overengeneered for an exceptional case that should not happen anyway. Finally org.eclipse.e4.ui.css.core.impl.dom.parsers.AbstractCSSParser.getParser() converts all Exception to ParserNotFoundException anyway.

jukzi:
Only random fail on Mac: https://github.com/eclipse-pde/eclipse.pde/issues/1289

PR #1938, opened 2024-06-07
Version bump org.eclipse.text.tests for 4.33 stream

vogella:


vogella:
For https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192

vogella:
As baseline check is ok on Linux / Mac and I see an failure which looks unrelated to me on Windows I merge.

PR #1947, opened 2024-06-10
Proposal: Opening/Closing Mechanism for Zip Files

Michael5601:
# Proposal: Opening/Closing Mechanism for Zip Files

## Background
The Eclipse IDE has no built in functionality to open Zip Files and read or manipulate their content. Because of this, other operations like searching inside of Zip Files or comparing two Zip Files were also not possible.

## Description
This PR introduces UI support for accesing the opening/closing mechanism for Zip Files over the UI-menu. It is accessed by right-clicking the Zip File that should be opened and then clicking on "Open Zip File". Closing is accessed the same way but when right-clicking an opened Zip File.

Please see [Plaform PR](https://github.com/eclipse-platform/eclipse.platform/pull/1413) for the PR on the repository **eclipse.platform** for the platform implementation and further information.

Co-Authored-By: CodeLtDave [david.erdoes@vector.com](mailto:david.erdoes@vector.com)


iloveeclipse (changes requested):
> Co-Authored-By: CodeLtDave [david.erdoes@vector.com](mailto:david.erdoes@vector.com)

Here and on other PR: **ALL** contributors must sign ECA. 

If `david.erdoes@vector.com` has not signed ECA, we can't  merge this PR.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/CloseZipFileHandler.java
```diff
@@ -0,0 +1,59 @@
...
+
+		Object element = ((IStructuredSelection) selection).getFirstElement();
+
+		if (!(element instanceof IFolder)) {
+			return null;
+		}
+		try {
+			ZipFileTransformer.closeZipFile((IFolder) element);
+		} catch (Exception e) {
+			MessageDialog.openError(shell, "Error", "Error opening zip file"); //$NON-NLS-1$ //$NON-NLS-2$
```
jukzi: log this error instead of error dialog/std.err

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
```diff
@@ -0,0 +1,79 @@
...
+			return null;
+		}
+			try {
+				dialog.run(true, false, new IRunnableWithProgress() {
+					@Override
+					public void run(IProgressMonitor monitor) throws InterruptedException {
+						monitor.beginTask("Opening Zip File", 5); //$NON-NLS-1$
+						try {
+							ZipFileTransformer.openZipFile((IFile) element, monitor, true);
+						} catch (URISyntaxException | CoreException e) {
```
jukzi: The exception stacktrace is lost here. Make sure full exception can somehow be seen in log

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
```diff
@@ -0,0 +1,79 @@
...
+							throw new InterruptedException(e.getMessage());
+						}
+						monitor.worked(1);
+					}
+				});
+			} catch (InvocationTargetException e) {
+				e.printStackTrace();
```
jukzi: do not use std.err

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
```diff
@@ -0,0 +1,79 @@
...
+			} catch (InterruptedException e) {
```
jukzi: typically InterruptedException is handled silently

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/ZipFileSystemContributor.java
```diff
@@ -0,0 +1,84 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
```
jukzi, line 2: 2024 Vector?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/ZipFileSystemContributor.java
```diff
@@ -0,0 +1,84 @@
...
+		pathBuf.append(pathString);
+		try {
+			//scheme, host, path, query, fragment
+			return new URI("zip", null, "/", pathBuf.toString(), null); //$NON-NLS-1$ //$NON-NLS-2$
+		} catch (URISyntaxException e) {
+			return null;
+		}
+	}
+
+	/* (non-Javadoc)
```
jukzi: remove this javadoc. javadoc is automatically inherited for overriden methods

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/ZipFileSystemContributor.java
```diff
@@ -0,0 +1,84 @@
...
+ * @since 3.23
+ */
+public class ZipFileSystemContributor extends FileSystemContributor {
+
+	public ZipFileSystemContributor() {
+		super();
+	}
+
+	@Override
+	public URI getURI(String pathString) {
```
jukzi, line 37: this method feels confusing complicated. please give an example in java comment what the resulting URI will look like.

jukzi:
Please post some screenshots about the workflow.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/ZipFileSystemContributor.java
Michael5601: This method is copied unchanged from the example project. We thought this would be the best place for it but did not change it, thus the Copyright is still the old one.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/CloseZipFileHandler.java
Michael5601: thanks for the input. I commited the changed under 297459a371fb37bc1c023abc6ec0da099d6688cc. But I wonder if the user should get a response that closing the zip file did not work. This was the reason behind the MessageDialog even though the message is wrong and should be "Error closing zip file".

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
Michael5601: This is the same as in the [comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1947#discussion_r1633350494) above. I want the user to be informed about the failed opening of the zip file.

Michael5601:
I want to clarify the workflow with the following screenshots:

At first the Zip File is closed and it is not possible to view/modify or search its contents. 
ClosedZip

By right-clicking and selection "Open Zip File" the Zip File can be opened.
Menu

Opening the Zip File allows using it like a normal folder with all operations supported.
OpenedZip

Opening a resource inside of the zip opens the editor for the resource.
OpenedResource

Opened Zip Files can be closed by right-clicking and selecting "Close Zip File".
Menu2

After closing the zip file we have the same state as in the beginning.
ClosedZip

For other archive types than "zip" the button will not show in the menu when right-clicking the archive.
I hope these screenshots could help to understand the workflow.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
Michael5601: In this 297459a371fb37bc1c023abc6ec0da099d6688cc I removed the usage of std.err.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
jukzi: That makes more sense in the Open Action then during close, but please do not misuse InterruptedException for that

jukzi:
> I hope these screenshots could help to understand the workflow.

Makes sense. I would have expected open/close to be more at the top, as the normal "Open" action is:
image

jukzi:
Just another idea: explicit open/closing zipfiles could be avoided by adding a node "Archives" like JDT already does for "Referenced Libraries" which makes it possible to explore the zip files contents too.
image

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
Michael5601: Thank you for your input last week. I removed the ProgressDialog in 297459a371fb37bc1c023abc6ec0da099d6688cc and thus no InterruptedException is thrown anymore.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/ZipFileSystemContributor.java
Michael5601: You are right. I removed the javadoc.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/ide/fileSystem/zip/OpenZipFileHandler.java
Michael5601: In 21ca9345a09d9a8bdcad30bd5712051ed5172a25 I added a log before the MessageDialog to preserve the full exception.

Michael5601:
> Just another idea: explicit open/closing zipfiles could be avoided by adding a node "Archives" like JDT already does for "Referenced Libraries" which makes it possible to explore the zip files contents too. image

This was our initial approach to the topic, though I don't recall the exact reasons we decided against it, given the time that has passed. The main issue, I believe, was that this window relies on JDT, while we wanted to exclusively use platform content. Additionally, the search functionality only targets content within the local workspace. Since our project aimed to improve the file search at first, we quickly abandoned the JDT approach. If you know a specific way how this could work with JDT we can also examine it further.

PR #1949, opened 2024-06-10
[Sticky Scrolling] Throttle sticky scrolling calculation

Christopher-Hermann:
The calculation of the sticky-scrolling lines is now throttled, executing only every 100 ms, to prevent potential performance bottlenecks.

Follow up to #1894


mickaelistria:
Thank you!

Issue #1956, opened 2024-06-12
Inactive selected table/tree rows background color wrong, selected icon missing, foreground colors missing on Linux


iloveeclipse:
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1690.

1) Open two tree views next to each other, like Package and Project explorers. 
2) Select some element in one of them.
3) Select an element in other one.
4) Selection in previous one has background color that is almost indistinguishable from text color, so content of selection is impossible to read anymore.

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/964108/3dc2f04e-8c1d-40ed-ac82-011772408f97)

**This must be fixed, as it makes IDE usage impossible!**

Christopher-Hermann:
@iloveeclipse can you provide me a screenshot of the coloring without my changes?

iloveeclipse:
Once [I can start debugger](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2118#issuecomment-2162323939) I will do that :-)

iloveeclipse:
Actually not only that is broken. Icons on selected rows aren't drawn anymore.
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/964108/0bb2f313-19b3-4195-8e96-5f5fa666da7d)

Christopher-Hermann:
> Actually not only that is broken. Icons on selected rows aren't drawn anymore. 

Mh strange, is this really related to this change? At least on your other screenshots the icons are drawn. And on MacOS the icons in the plugin.xml are drawn as expected.
If it is related, I would revert the change if possible. I think it will take a while to check and fix the issues on other platforms.

iloveeclipse:
> Once [I can start debugger](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2118#issuecomment-2162323939) I will do that :-)

So I've fixed my stupid error and can start debugger now.
This is how the IDE looks like **before** the change in question (note also NOT missing icon in Plugin editor):

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/964108/621d8c88-39c2-4d27-9fe9-a567944bc4c2)

iloveeclipse:
Same **without** CSS themes, OK before https://github.com/eclipse-platform/eclipse.platform.ui/pull/1690 :

[...]

Issue #1955, opened 2024-06-12
plenty test failure in ColumnViewerSelectionColorListener.handleEvent


jukzi:
Hundrets of fails in I Build like

```
Argument cannot be null
java.lang.IllegalArgumentException: Argument cannot be null
at org.eclipse.swt.SWT.error(SWT.java:4903)
at org.eclipse.swt.SWT.error(SWT.java:4837)
at org.eclipse.swt.SWT.error(SWT.java:4808)
at org.eclipse.swt.graphics.GC.setBackground(GC.java:3329)
at org.eclipse.jface.viewers.internal.ColumnViewerSelectionColorListener.handleEvent(ColumnViewerSelectionColorListener.java:69)
at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4660)
at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1622)
at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1645)
at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1630)
at org.eclipse.swt.widgets.Table.drawInteriorWithFrame_inView(Table.java:1056)
at org.eclipse.swt.widgets.Display.windowProc(Display.java:6669)
at org.eclipse.swt.internal.cocoa.OS.objc_msgSendSuper(Native Method)
at org.eclipse.swt.widgets.Widget.drawRect(Widget.java:812)
at org.eclipse.swt.widgets.Table.drawRect(Table.java:1231)
at org.eclipse.swt.widgets.Display.windowProc(Display.java:6254)
at org.eclipse.swt.internal.cocoa.OS.objc_msgSendSuper(Native Method)
at org.eclipse.swt.widgets.Display.applicationNextEventMatchingMask(Display.java:5578)
at org.eclipse.swt.widgets.Display.applicationProc(Display.java:5979)
at org.eclipse.swt.internal.cocoa.OS.objc_msgSend(Native Method)
at org.eclipse.swt.internal.cocoa.NSApplication.nextEventMatchingMask(NSApplication.java:92)
at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3997)
at org.eclipse.jface.tests.viewers.ViewerTestCase.processEvents(ViewerTestCase.java:104)
[...]

PR #1957, opened 2024-06-12
Revert highlighting fix

fedejeanne:
Revert #1690 

Due to #1955 , #1956 , https://github.com/eclipse-platform/eclipse.platform.swt/issues/1275


iloveeclipse:
You cannot revert bundle versions for published bundles!

iloveeclipse:
We have IBuild there with these versions.
The problem is that all new builds seeing IBuild versions will always pick the highest (broken) version of bundles.

@merks : WDYT, should we mark it unstable and revert versions, or better don't try to revert versions?

akurtakov:
Please don't revert versions, everyone that already updated to latest I-build (I am aware and will redownload myself but others might not be) will not get updates.

fedejeanne:
> Please don't revert versions, everyone that already updated to latest I-build (I am aware and will redownload myself but others might not be) will not get updates.

👍 
Done in 22dcffa4f1f6f97d2029c3b88e943e9ec40d94c3

merks:
Yes, please don't revert versions.  These things have already escape and I are my pool.   Also, I think we can assume this will eventually be fixed properly and hence all the bundle will be incremented at some point in the not-to-distant future.

fedejeanne:
> Yes, please don't revert versions. 

@merks versions should be as published now --> 22dcffa4f1f6f97d2029c3b88e943e9ec40d94c3

>  I think we can assume this will eventually be fixed properly and hence all the bundle will be incremented at some point in the not-to-distant future.

I'd say it should be possible to wait until later today to apply this PR and give @BeckerWdf / @Christopher-Hermann some time to find a better fix.

@iloveeclipse would that be ok for you? Since you are in Linux, you are seeing the problems in your IDE already. Colors can be changed though (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/1690#issuecomment-2073052771), so it should be endurable for a couple of hours (?)

fedejeanne:
> Please don't revert versions, everyone that already updated to latest I-build (I am aware and will redownload myself but others might not be) will not get updates.

Hm, on a second thought, this means that not only should I not revert (decrease) versions but I should actually **increase** versions, right @akurtakov ? I mean those who already updated their IDE and have #1690 in it should be able to get an update with #1957 (this PR) tomorrow. 

I'll increase versions in a separate commit in a minute.

merks:
Yes, you should keep the versions as they are.  They should end up with a new qualifier in a new build via the other revert commit in the projects.

mickaelistria:
> this means that not only should I not revert (decrease) versions but I should actually increase versions

No need to modify the x.y.z, the version qualifier will be increased at build-time because as it will be updated with the commit merge date, which would be newer.

merks:
I'm okay with not doing a revert at all if this can be addressed in the next few hours.  Yesterday when was testing the removal of a feature, the build itself took like 3.5 hours and then the testing seemed to take another 4, so there is along delay between when you might commit a revert (without reverting versions) and a build from that being available.

merks:
> > Please don't revert versions, everyone that already updated to latest I-build (I am aware and will redownload myself but others might not be) will not get updates.
> 
> Hm, on a second thought, this means that not only should I not revert (decrease) versions but I should actually **increase** versions, right @akurtakov ? I mean those who already updated their IDE and have #1690 in it should be able to get an update with #1957 (this PR) tomorrow.
> 
> I'll increase versions in a separate commit in a minute.

Just in case the rapid responses are not 100% clear, the idea is to revert everything **except the version changes** and as @mickaelistria correctly points out, this will result in new qualifier versions so that we can update to the reverted versions.

Hopefully the problem is just fixed very quickly...

---

I'm out for a few hours, otherwise I could test that the installer works with whatever fixes are being applied...

iloveeclipse:
On Linux (and probably on other OSes) we have more problems that needs a fix: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1956 (wrong selection color, missing selection icon, missing foreground coloring on trees).

@Christopher-Hermann : if you could fix all of them today till evening, I can validate your fixes. 
Otherwise I would like to merge this revert for the next IBuild this night.

Christopher-Hermann:
> @Christopher-Hermann : if you could fix all of them today till evening, I can validate your fixes. Otherwise I would like to merge this revert for the next IBuild this night.

Let's merge this revert. I will not mange to fix the issues by today.

akurtakov:
FWIW, I strongly recommend testing colors issues/changes with both theming engine enabled and disabled, IMO it does a lot more than needed trying to workaround issues on lower lever thus breaking UI when an issue underneath is fixed.

iloveeclipse:
> I strongly recommend testing colors issues/changes with both theming engine enabled and disabled

Exact. This is what I did on https://github.com/eclipse-platform/eclipse.platform.ui/issues/1956 and the problem appears in **both** modes.

iloveeclipse:
I've quickly tested this PR and it fixes all the issues I've reported in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1956. So let's merge.

iloveeclipse:
Should I trigger another IBuild now to see the real test results without JFace related errors?

merks:
Yes please.

iloveeclipse:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/15/ started

PR #1961, opened 2024-06-13
[Sticky Scrolling] Dispatch scrolling to source viewer

Christopher-Hermann:
Scrolling on the sticky lines control is dispatched to the linked source viewer. With this change the source viewer's scrolling behavior remains consistent, regardless of whether sticky scrolling is enalbed or not.

Follow up for #1894


BeckerWdf:
I just manually tested on macOS. Scrolling now is the same regardless of the position of the mouse pointer.
Even when it's in the sticky lines the editor scrolls. So this looks good to me.
If nobody objects I will merge this today.

BeckerWdf:
Failures on windows and macOS seem unrelated to me.

PR #1969, opened 2024-06-14
NPE in StickyScrollControl.getStickyLineStyleRanges #1964

Christopher-Hermann:
When the sticky lines are styled, the linked text widget could already be disposed or null. In this case, the sticky lines must not be styled.

Fixes #1964

PR #1967, opened 2024-06-14
[Sticky Scrolling] Fix separator color on windows

Christopher-Hermann:
The separator in the sticky lines controls are drawn in a bad color on windows dark theme. With this change, the separator color is aligned with the theme.

Current coloring on windows:
Bildschirmfoto 2024-06-14 um 10 42 22

How it should look like:
Bildschirmfoto 2024-06-14 um 10 43 20

This is a follow up for eclipse-platform/eclipse.platform.ui#1894

PR #1970, opened 2024-06-14
[Sticky Scrolling] Fix instable test

Christopher-Hermann:
Asserting on magic numbers for the control width is very error prone. To avoid this problem, it's asserted that the control is resized.

This solves test failures on #1967 and #1969


BeckerWdf (approved):

BeckerWdf:
Thanks for fixing

Issue #1866, opened 2024-04-29
After failed file creation due to "A resource exists with a different case", wizard still tries to open non-existant File


jwflicker:
- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK]Eclipse IDE for Java Developers (includes Incubating components)

## Steps to reproduce
From a fresh installation and clean workspace:

1. From Package Explorer (or any explorer)
2. Crdate a new general project using menu New > Project and select General > Project.
3. Select that general project and from its context menu select New -> File
4. Create file "abc.txt" in that project.
5. Again, select that general project and from its context menu select New -> File
6. Create file "ABC.txt" in that project. (Same file name, different casing)
7. You will see an error dialog with message: "A resource exists with a different case: '/project/abc.txt'"
8. After pression OK on the error dialog, the IDE will then still attempt to open that non-existant file in an Editor.
9. Depending on the file extension, it may  open the existing file with matching name in a System Editor, other times it will try to open the non-existant file in an internal editor so that the editor just displays the error message: `Resource '/project/ABC.txt' does not exist`

I expected: that it would not try to open an editor to the file name that could not be created due to a an existing file with similar name but in a different case.

But got: Sometimes an extenral system editor open on the 'wrong' file, sometimes an internal editor w/ the Resource not found eror.

## Here is some relevant source code
BasicNewFileResourceWizard:
```
	@Override
	public boolean performFinish() {
		IFile file = mainPage.createNewFile();
		if (file == null) {
			return false;
		}
```
 mainPage.createNewFile() will trigger the display of the error message dialog showing "A resource exists with a different case", but it will return a non-null File object whose .exists() method would return false, but is not checked before invoking `IDE.openEditor(page, file, true);` just below

within WizardNewFileCreationPage.createNewFile() method, the error is caught and thrown here... but nothing is done to cause a return of null:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFileCreationPage.java#L409

## Tested under this environment:
- Version: 2024-03 (4.31.0)
- Build id: 20240307-1437
- OS: Mac OS X, v.14.4.1, x86_64 / cocoa
- Java vendor: Eclipse Adoptium
- Java runtime version: 17.0.10+7
- Java version: 17.0.10

## Community

[...]

PR #1984, opened 2024-06-18
case-sensitive file existence check

andrew-tram:
Check to address - https://github.com/eclipse-platform/eclipse.platform.ui/issues/1866

### DIFF
--- a/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.ide; singleton:=true
-Bundle-Version: 3.22.300.qualifier
+Bundle-Version: 3.22.301.qualifier
 Bundle-Activator: org.eclipse.ui.internal.ide.IDEWorkbenchPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %Plugin.providerName
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFileCreationPage.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFileCreationPage.java
@@ -781,25 +781,31 @@ public class WizardNewFileCreationPage extends WizardPage implements Listener {
 	    IPath newFilePath = getContainerFullPath().append(getFileName());
 	    IFile newFileHandle = createFileHandle(newFilePath);
 
+	    boolean isLinux = System.getProperty("os.name").toLowerCase().contains("linux");  //$NON-NLS-1$//$NON-NLS-2$
+
 	    try {
 	        IContainer parentContainer = newFileHandle.getParent();
 	        IResource[] members = parentContainer.members();
 	        for (IResource member : members) {
 	            if (member.getType() == IResource.FILE) {
-	                IPath memberPath = member.getFullPath();
-	                if (memberPath.equals(newFilePath)) {
+	                IFile existingFile = (IFile) member;
+	                if (existingFile.getFullPath().equals(newFilePath)) {
 	                    continue; // Skip if it's the same path
 	                }
 
-	                // Check each segment for case differences
-	                if (isCaseDifferent(memberPath, newFilePath)) {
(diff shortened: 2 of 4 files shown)


jukzi (changes requested):
A filesystem may be case sensitive. Mostly depends on OS

andrew-tram:
@jukzi , good call! I adjusted the checks to account for OS

jukzi (dismissed):
please do not explicitly check for OS but use any existing function to compare files for equality.

andrew-tram:
Understood, hopefully this is closer...

```
	    // Use workspace and resource APIs to handle file comparison
	    IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
	    IPath newFilePath = getContainerFullPath().append(getFileName());
	    IFile newFileHandle = createFileHandle(newFilePath);

	    IFile[] files = root.findFilesForLocationURI(newFileHandle.getLocationURI());
	    for (IFile file : files) {
	        if (file.exists() && file.getName().equalsIgnoreCase(newFileHandle.getName()) &&
	                !file.getName().equals(newFileHandle.getName())) {
	            setErrorMessage(NLS.bind(IDEWorkbenchMessages.ResourceGroup_nameExistsDifferentCase, file.getName()));
	            return false;
	        }
	    }
```

jukzi:
You should not need to compare Strings yourself.

merks:
I’d expect to see IFile or IPath to be compared which know whether to do case-sensitive comparisons.

andrew-tram:
I believe this is closer to what yall are suggesting...

```
// Use workspace and resource APIs to handle file comparison
IPath newFilePath = getContainerFullPath().append(getFileName());
IFile newFileHandle = createFileHandle(newFilePath);

try {
	IContainer parentContainer = newFileHandle.getParent();
	IResource[] members = parentContainer.members();
	for (IResource member : members) {
		if (member.getType() == IResource.FILE) {
			IPath memberPath = member.getFullPath();
			if (memberPath.equals(newFilePath)) {
				continue; // Skip if it's the same path
			}
			if (memberPath.lastSegment().equalsIgnoreCase(newFileHandle.getName()) &&
					!memberPath.lastSegment().equals(newFileHandle.getName())) {
				setErrorMessage(NLS.bind(IDEWorkbenchMessages.ResourceGroup_nameExistsDifferentCase, member.getName()));
				return false;
			}
		}
	}
} catch (CoreException e) {
	setErrorMessage("Error while validating file name: " + e.getMessage());
	return false;
}
```

Will externalize the new string if all is good.

merks:
I think you should test it.   I think memberPath.equals(newFilePath) will return true for two paths with different cases on a case-insensitive file system.  I think you if want to check if two paths have different case, you should compare the strings.  So if the paths are equal but the string representation are not equal, then what's specified is different from what's present.  And that problem can be the case for the any segment of the path (project/folder), not just the final segment.

andrew-tram:
Ive tested this on Windows 11 and Mac 14.4.1 (default HFS+ or APFS in case-insensitive mode). The expected results look good where an error displays saying the file already exists with the same name, different case.

Trying Linux now.

andrew-tram:
So, I'm not entirely sure how it's feasible to do this without checking for at least if it's Linux.

jwflicker:
If this approach becomes a challenge, a smaller, targeted fix might be to not selectAndReveal the file and open it an editor if what is returned from `mainPage.createNewFile()` doesn't exist.

[BasicNewFileResourceWizard](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/wizards/newresource/BasicNewFileResourceWizard.java#L96C17-L96C30):
```
	@Override
	public boolean performFinish() {
		IFile file = mainPage.createNewFile();
		if (file == null) {
			return false;
		}
```
 mainPage.createNewFile() will trigger the display of the error message dialog showing "A resource exists with a different case", but it will return a non-null File object whose .exists() method would return false, but is not checked before invoking `IDE.openEditor(page, file, true);` just below
 
So perhaps just changing `if (file == null) {` to `if (file == null || !file.exists()) {` would dullen the bulk of the annoyance of  https://github.com/eclipse-platform/eclipse.platform.ui/issues/1866

jukzi:
You are supposed to find an existing method of signature
`boolean equals(IFile, IFile)` or `boolean equals(IPath, IPath)` which does the OS check for you. I dont know its place or name by heart but am sure you can find many of them.
Even java.nio.Path.equals(Path) probably does it if the Path is constructed using the appropriate FileSystem.

PR #1990, opened 2024-06-22
Find/Replace overlay: add a search- and replace- history

### LINKED ISSUE #1907: [Find/Replace Overlay] The search-history cannot be accessed from the Overlay
which could not be addressed into the current proposal. [comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192#issuecomment-1777142848) by @jukzi 

### Currently, the search-history cannot be accessed from the Overlay.
In the find/replace dialog, there is a dropdown for each the search-bar and the replace-bar, as such:
grafik

### (Help wanted!) A few ideas on how to implement this into the Overlay:
-) Navigate history by pressing up/down in each Bar (VSCode does this! I am strongly in favor of this implementation!) 
grafik
-) Add up/down buttons in the toolbar (I think this can be okay-ish)
-) Add a drop-down button to each bar (I am not keen on this idea as it would impact the flat look of the Overlay)

Wittmaxi:
Adds a history feature for the find and replace bar of the overlay.
The history is displayed using a menu filled with entries from the
HistoryStore.

fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1907

grafik
grafik


Wittmaxi:
@jukzi Can you please review this feature and tell me what you think of it? :)

jukzi:
next week

Wittmaxi:
@jukzi yes, no rush :)

iloveeclipse:
Instead of the menu (or additionally to it), could we have ctrl+space for the list of proposals, filled with the history?

Wittmaxi:
@iloveeclipse I think that could be a good idea! I was also thinking about allowing moving up/down the history using the arrow keys like vscode does. Ctrl+Space fits the "content assist" idiom and is also a very good idea!

HannesWell:
Thank you for this. The screenshot already look promising, 🚀 

> Instead of the menu (or additionally to it), could we have ctrl+space for the list of proposals, filled with the history?

That's also an interesting proposal. But in order to make users aware of it, it should for example be mentioned in the empty-block background text (similar to the screenshot of VS code in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1907).
I'm not 100% sure if a content-assist behavior fits as trigger to show a search history, but maybe that's just based on what I'm used to, but in the end there is no right or wrong.

What also came into my mind when I read Andrey's proposal: The history could automatically drop-down as soon as the text-block gets the focus. Additionally the already typed text could then be used as 'filter' for the history and remove not matching items.
But I'm not sure if this is too much noise if one already knows what to type?

iloveeclipse:
> order to make users aware of it, it should for example be mentioned in the empty-block background text

Not necessarily, Eclipse has a concept of "light bulb" field decoration if there is something that provides extra functionality.

Wittmaxi:
@HannesWell @iloveeclipse @jukzi before coming up with many new features, I would like to understand the problem at hand a little bit better. I never use the search history myself (even in the Dialog) and I couldn't come up with a workflow where the search history could be "really" useful. Can you please tell me a bit about how the search history is really used?

>  The history could automatically drop-down as soon as the text-block gets the focus.

I never use the search history, for me, this would just be (very) annoying.

jukzi:
> Can you please tell me a bit about how the search history is really used

example: i search a method in a .java file, but i am too lazy to type each letter of the method name. Therefore i copy paste a Stacktrace line "p.C.method ()" from VisualVm . Since the pasted line contains more then only the method name i have to delete  the classname prefix "p.C" and the " ()" suffix within the search dialog. As soon as i have done that i will find a method and all its calls in the active editor. Now i continue with editing, searching other things and so on. Later i want to search the same method again - but still i don't know its name. I used to find its name in the search history. Now i only find it in the copy/paste history from windows but there it still contains prefix and suffix.

jukzi:
I have  tested it. Optically it is not beautiful but at least it works. Here are my nitpicks:
* i expect that drop down to be below and aligned to the text field  - as it was in the old search dialog:
image
old:
image

* i expect that active UI components i.e. the drop down to have the highest possible contrast i.e. Black on white - as it was in the old search dialog
* i expect that up and down keys can be used to scroll through the list - as it was in the old search dialog
* i expect that i can close the list when i click on on the arrow (that is used to open the list) a second time - as it was in the old search dialog
* compare to the old search dialog the list wastes much space in the front of the items
image

In summary i expected that the drop down would be exactly as in the old dialog, because that worked well and i dont know how it could be made better.

iloveeclipse:
> I never use the search history myself (even in the Dialog) and I couldn't come up with a workflow where the search history could be "really" useful. Can you please tell me a bit about how the search history is really used?

- Trying to fix ~2000 test fails in not my own code. 
- Searching / replacing different constants / code with other constants / code, also using regex.
- Mass search/replace doesn't work as code to replace relies on context
- Gazillion of files opened, same patterns with different variations need to be replaced with other patterns with different variations.
- Having search *and* replace field history allows me to search in same file for different occurrences from the history and replace it with different occurrences from the history.
- Search (and replace) patterns are not trivial text and just typing it would be a constant pain, so history entries for *both* is essential.

Wittmaxi:
@iloveeclipse @jukzi thank you very much for your user-stories! That helps me a lot understanding your concerns for the usability of the history.

It is now possible to navigate through the history using the arrow keys (arrow up/arrow down)


I will move on to address @jukzi s concerns.
I have implemented the search history using a menu, but it might be better to implement it as a shell overlaid on the Find/Replace overlay - what do you think?
In particular, I cannot use a Combo (like Find/Replace Dialog does) because it does not fit into the Overlay aesthetically and does also not support all required Operations (for example: it has no concept of a "message" which is the preview text shown when the text bar is empty)

Wittmaxi:
@jukzi what do you think of this version?

jukzi:
it *looks* good!

- up/down arrows still not working
- error: when i press down button again eclipse is send to background!

jukzi:
another minor: in the bottom left of the eclipse window the progress shortly says "String xyz not found"
image

1. The word "String" is something a user does not need to understand. Better: "'xyz' not found"
2. "xyz NOT FOUND" is a status not a progress, that should not disappear within a tens of a second


--
up/down key work if and only if the replace field is visible.

jukzi:
error loged:
```
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:500)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:419)
	at org.eclipse.swt.widgets.Control.isFocusControl(Control.java:1859)
	at org.eclipse.ui.texteditor.FindReplaceOverlay.lambda$0(FindReplaceOverlay.java:195)
	at org.eclipse.swt.events.KeyListener$1.keyPressed(KeyListener.java:66)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:184)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4300)

```

Wittmaxi:
Thank you for the report, I fixed the error here: https://github.com/eclipse-platform/eclipse.platform.ui/commit/0b1b6224ab8ea42a215af25fb104a884317e0aca

Can you please tell me what exactly you do to send eclipse to the background? I could not reproduce that.

Changing the status text is out of scope for this issue. I don't really think it's worth changing (and it is the "old" behavior), but if you think it is, we can create an Issue for it.

jukzi:
> Can you please tell me what exactly you do to send eclipse to the background? I could not reproduce that.

i am on WIndows. i have multiple programs open. I have replace line closed. I click on down-> history opens. I click on down again .> another program comes in front (feels like closing to the dropdown sends eclipse to background).

Wittmaxi:
@jukzi yes, I can reproduce this! Thanks!
I will look into this

jukzi:
now eclispe does not move to background anymore but again drop down can not be closed by second click.

jukzi:
* i can not scroll up/down the drop down with the keys when i use TAB+ENTER to open the drop down (instead the drop down closes)
* right click on an entry in the drop down opens the context menu of the editor in the background (i was looking if i could remove a search entry from history)

Wittmaxi:
@jukzi I have addressed all your concerns.
Allowing "right click" to remove the menu item is an interesting proposal but out of scope for now. If you want that option added, please create an issue for it :)

Wittmaxi:
@HeikoKlare I have moved quite some files into a separate folder as part of this PR. Should we extract those changes into another PR?

HeikoKlare:
> @HeikoKlare I have moved quite some files into a separate folder as part of this PR. Should we extract those changes into another PR?

Yes, it would be good to have the refactoring separated from this feature implementation. Currently, the refactoring does not seem to be correct, as for example the `AccessibleToolBar` class is now duplicated instead of being moved.

jukzi:
> @jukzi I have addressed all your concerns.
> Allowing "right click" to remove the menu item is an interesting proposal but out of scope for now. If you want that option added, please create an issue for it :)

Starting with an empty history the drop down looks wrong ("big void") and gives an exception when selecting the void:
image

```
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at org.eclipse.ui.internal.findandreplace.overlay.HistoryTextWrapper.lambda$2(HistoryTextWrapper.java:88)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:265)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4300)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1174)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1198)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1183)
	at org.eclipse.swt.widgets.Widget.notifyListeners(Widget.java:853)
	at org.eclipse.ui.internal.findandreplace.overlay.SearchHistoryMenu.lambda$0(SearchHistoryMenu.java:67)
	at org.eclipse.swt.events.MouseListener$2.mouseDown(MouseListener.java:96)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:209)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4300)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1174)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4098)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3688)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:639)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
```

The void even remains after adding a history:
image

On the right side (place of a v-scrollbar) there is a void too, i guess it should be a disabled scrollbar.
image

jukzi:
very minor: it feels unusual that "search" and "replace" text field have different width - and so does the dropdown.
May be it would be worth to give it a try to have the both drop down buttons aligned, i.e. at the right (or left) most position.
image
Or to make the replace text field as small as the search text field.

Wittmaxi:
@jukzi Thanks for the findings :)

> Starting with an empty history the drop down looks wrong ("big void") and gives an exception when selecting the void:

The Find/Replace Dialog shows a "big void" too when no search options are shown 
grafik

I am not really sure what I should do instead. Do you have an idea?
Until we have a better idea, I can reduce the size of the big void:
grafik

In case you are curious, this is how I set the size of the search-history dialog:
```java
getShell().setBounds(location.x,
  location.y, width,
  Math.min(
     table.getItemHeight() * 7,
     table.getItemHeight() * (table.getItemCount() + 1)
   ));
```

Wittmaxi:
> gives an exception when selecting the void:

Clicking the big void should not throw an exception! I have reduced it's size and fixed the exception in https://github.com/eclipse-platform/eclipse.platform.ui/commit/42e8b8aa29339dbf472787093add12b03341cef2

Wittmaxi:
grafik

I have also fixed the spacing issue with the empty dropdown

Wittmaxi:
> minor: it feels unusual that "search" and "replace" text field have different width - and so does the dropdown.
May be it would be worth to give it a try to have the both drop down buttons aligned, i.e. at the right (or left) most position.

I am not quite sure how we could align those buttons;
The dropdowns are directly related to the text-bar so they should absolutely be next to it. And I think it would feel weird to reduce the size of the replace-bar.

Since this issue is unrelated to this PR, we can either create a new issue and discuss it there or accept the size difference (which is what I would vote for)

HeikoKlare:
I am not sure whether that might feel unnatural, but another option to have the drop-down arrows for the history horizontally aligned, they could be placed before (i.e., left of) the input fields instead of after (i.e., right of) the input fields.
As a minor benefit that would also reduce the required mouse movement, as you usually move the mouse to the text to select an item in the drop-down menu, and since the text is left-aligned, opening the drop-down with an arrow to the left reduces the distance between arrow and entry text.

image

Wittmaxi:
@HeikoKlare I don't really know how to feel about the proposed fix. I don't think I have seen a spinner with the dropdown-button on the left before (I even searched on google!). Since in my eyes, there is no problem (the dropdowns are already aligned to the left, just the buttons aren't!), I don't see why we would opt for this unintuitive solution

However, there might be user stories or expectations, which require the buttons to be aligned vertically (I'd be very happy to hear more about them, so I can understand them better). In that case, your solution would definitely bring a very good solution to the table.

jukzi:
It is ok to have a big void for an empty history - as it was in old dialog. If there are items in the history there should be no void beneeth them
image
 i still see a void on the bottom and still get an exception clicking on it
image
```
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at org.eclipse.ui.internal.findandreplace.overlay.HistoryTextWrapper.lambda$2(HistoryTextWrapper.java:88)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:265)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4300)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1174)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1198)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1183)
	at org.eclipse.swt.widgets.Widget.notifyListeners(Widget.java:853)
	at org.eclipse.ui.internal.findandreplace.overlay.SearchHistoryMenu.lambda$0(SearchHistoryMenu.java:67)
	at org.eclipse.swt.events.MouseListener$2.mouseDown(MouseListener.java:96)
```

Wittmaxi:
> It is ok to have a big void for an empty history - as it was in old dialog. If there are items in the history there should be no void beneeth them

Now I understand you! Sorry for the confusion, that is something I can easily change :)
But first, let me understand the exception you are receiving and fix it

Wittmaxi:
@jukzi 
The event isn't triggered (anymore) for me. Did you do anything special I didn't think to do?

jukzi:
exception is gone. maybe i was working on an old version. sorry.

Wittmaxi:
@jukzi it's all good :)

grafik

I have made it so that the dropdown shrinks as much as possible. If no items are shown, it will not have a "big void" but a "small void" with info text:
grafik

jukzi:
ok, then i suggest to merge this, to get feedback from others.

Wittmaxi:
@jukzi sounds good, Thank you a lot for the valuable feedback and the help iterating over this feature! @HeikoKlare

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayFirstTimePopup.java
```diff
@@ -41,6 +41,11 @@
  * shown, informing the user about the new functionality. This class will track
  * whether the popup was already shown and will only show the Overlay on the
  * first time the popup was shown.
+<<<<<<< Upstream, based on 4674cdf92b73b519d7f1fc85ba15112cfa2d5a2f
```
jukzi: merge conflict

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/HistoryStore.java
```diff
@@ -109,4 +109,8 @@ private void writeHistory() {
 		settingsManager.put(sectionName, names);
 	}
 
+	public List<String> asList() {
+		return new ArrayList<>(history);
```
jukzi: What is the intention of this method? Immutable copy? then use List.copyOf()

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -88,28 +91,36 @@ public class FindReplaceOverlay extends Dialog {
...
 	private Composite searchBarContainer;
-	private Text searchBar;
+	private HistoryTextWrapper searchBar;
 	private AccessibleToolBar searchTools;
 	private ToolItem searchInSelectionButton;
 	private ToolItem wholeWordSearchButton;
 	private ToolItem caseSensitiveSearchButton;
 	private ToolItem regexSearchButton;
+
+	@SuppressWarnings("unused")
```
jukzi: general advice about @SuppressWarnings: As far as i understand those buttons are used in junit test via reflection. Either document that reason, or much better: do not use reflection. Instead for example offer and use getters. That makes code easier to maintain.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/HistoryStore.java
Wittmaxi: The intention of this method is that I want to have access to methods implemented by List (but really only indexOf and size()).
Now that you question this method, I start having doubts myself whether it's a good idea to have it.
An alternative would be implementing size() and indexOf() in HistoryStore. What do you think?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: This is a tricky one and has caused us quite some headaches before.

1) We don't want to expose the buttons to the outside world (at that point, we could just as well make them public)
2) The old tests used reflection - which is a bummer, but "works". We ended up using that method as well, knowing that it has problems
3) The bigger problem is that the tests are not in the same package as the Overlay: we cannot even make the members package-private since the test still wouldn't know about them
4) We tried looking into Reddeer and SWTBot for solving this issue. Unfortunately, neither are in the dependencies of Eclipse and we didn't want to introduce a new dependency (+ there were some other problems if I remember correctly)

I can add a comment to explain why I suppress warnings. I would opt for adding a single comment on the top one, I think that there's no need to repeat myself

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/HistoryStore.java
jukzi: sounds good - avoids the copy

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: I have added a comment to explain what is happening and referred to this comment for future reference: https://github.com/eclipse-platform/eclipse.platform.ui/commit/060b9941fea88d89e98acb6b8adea55cde83bbe7

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/HistoryStore.java
Wittmaxi: I agree - the code became much clearer! https://github.com/eclipse-platform/eclipse.platform.ui/commit/060b9941fea88d89e98acb6b8adea55cde83bbe7

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayFirstTimePopup.java
Wittmaxi: Sorry, I was a bit too fast there ;-)
https://github.com/eclipse-platform/eclipse.platform.ui/commit/060b9941fea88d89e98acb6b8adea55cde83bbe7

Wittmaxi:
I have added a forgotten copyright-header and removed some clutter in the diff. I have removed a @since-tag in the FindReplaceOverlayImages-class

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
jukzi: You can make them package private and use a testcode in the same package. But since the class is "internal" only, so it is not a problem to make public methods.
On the other hand you probably just don't strongly need those Buttons as members but can find the widgets from the  swt tree.

Wittmaxi:
@jukzi @HeikoKlare I believe that we should discuss the "unused fields" later and prioritize getting feedback on the search history. I have created this issue for the unused fields https://github.com/eclipse-platform/eclipse.platform.ui/issues/2060 and I have reverted the changes that made them unused.

I have also fixed the unit tests (at least locally) and removed a file which I forgot to remove while merging

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -102,14 +105,15 @@ public class FindReplaceOverlay extends Dialog {
...
-	private Text replaceBar;
+	private HistoryTextWrapper replaceBar;
 	private AccessibleToolBar replaceTools;
 	private ToolItem replaceButton;
 	private ToolItem replaceAllButton;
 
 	private Color backgroundToUse;
 	private Color normalTextForegroundColor;
 	private boolean positionAtTop = true;
+	private static final int HISTORY_SIZE = 15;
```
HeikoKlare: Nitpicky: constants should usually be placed on top of the file (above instance fields).

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -410,7 +416,7 @@ private void applyOverlayColors(Color color, boolean tryToColorReplaceBar) {
 	private void unbindListeners() {
 		getShell().removeShellListener(overlayDeactivationListener);
 		if (targetPart != null && targetPart instanceof StatusTextEditor textEditor) {
-			Control targetWidget = textEditor.getAdapter(ITextViewer.class).getTextWidget();
+			Control targetWidget = Adapters.adapt(textEditor, ITextViewer.class).getTextWidget();
```
HeikoKlare: Are the `textEditor.getAdapter(...`->`Adapters.adapt(textEditor, ...)` changes intended or by accident?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -472,17 +478,20 @@ private void retrieveBackgroundColor() {
 		textBarForRetrievingTheRightColor.dispose();
 	}
 
+
 	private void createSearchTools() {
 		searchTools = new AccessibleToolBar(searchContainer);
 		GridDataFactory.fillDefaults().grab(false, true).align(GridData.END, GridData.END).applyTo(searchTools);
 
+		@SuppressWarnings("unused")
+		ToolItem separator = searchTools.createToolItem(SWT.SEPARATOR);
```
HeikoKlare: Why store the item in a variable if it unused anyway? I think you can simply remove the variable along with the warning suppression.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+			Table table = (Table) f.widget;
+			TableItem[] selection = table.getSelection();
+			if (selection.length == 0) {
+				return;
+			}
+			String text = selection[0].getText();
+			if (text != null) {
+				textBar.setText(text);
+			}
+		}));
```
HeikoKlare: OO design: It would be cleaner to have the `SearchHistoryMenu` accept some `SelectedItemChangedListener`, being a functional interface for a consumer of `TableItem` (the one that changed). The current implementation assumes that the the `SearchHistoryMenu` attaches the given listener to a table, which introduces a dependency to implementation details of the `SearchHistoryMenu`.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+ * stored to history when requested by the Client code, the history is stored in
+ * a {@code HistoryStore} provided by the client. The history bar behaves like a
+ * normal {@code Text}.
+ */
+public class HistoryTextWrapper extends Composite {
+	private Text textBar;
+	private AccessibleToolBar tools;
+	private ToolItem dropDown;
+	private SearchHistoryMenu menu;
+	private HistoryStore history;
```
HeikoKlare: nitpicky: all these fields (except for `menu`) can be `final`

HeikoKlare (commented):
In general, the code changes look good and sound. When testing, the UI also looks good and works quite well. I have rather nitpicky comments, but at least one issue (exception) that needs to be addressed:

- Clicking "down" key when the history is empty results in an exception:
```
java.lang.ArithmeticException: / by zero
	at org.eclipse.ui.internal.findandreplace.overlay.HistoryTextWrapper.navigateInHistory(HistoryTextWrapper.java:119)
	at org.eclipse.ui.internal.findandreplace.overlay.HistoryTextWrapper.lambda$1(HistoryTextWrapper.java:75)
```

- There seems to be an issue with the background color of a separator:
image

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: This is an accident, very likely stemming from a manual merge

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: I agree!

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
Wittmaxi: Yes, they can!
Fixed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1990/files#diff-5f8b29df540e278eb662337e8f6648429250ca274cd72b5e844ff86f13559f43R45

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
Wittmaxi: I agree! I have created a `bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryItemSelectedListener.java` which abstracts from this to the outside user

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: Removed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/c7042b7e8f9f3aec27f4d038fa10befb832d8214

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: Changed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/c7042b7e8f9f3aec27f4d038fa10befb832d8214

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: I agree, changed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/c7042b7e8f9f3aec27f4d038fa10befb832d8214

Wittmaxi:
I have fixed the divide by zero and the coloring of the separators now works correctly

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
```diff
@@ -0,0 +1,139 @@
...
+			}
+		}
+	};
+	private Point location;
+	private int width;
+	private Table table;
+	private TableColumn column;
+
+	public SearchHistoryMenu(Shell parent, HistoryStore history,
+			HistoryItemSelectedListener menuItemSelectionListener) {
```
HeikoKlare: It would be sufficient to use a general `Consumer<String>` instead of introducing a new (functional) interface here. If not, the listener should at least be declared as a functional interface, so that a lambda instead of an anonymous class can be passed as an argument to this constructor.

HeikoKlare (commented):
Thank you for addressing my concerns! Regarding the code, I only have a minor comment left. For the images, we also need to have a PR with the svg to the platform.images repo.

I have two finding with respect to the UI / layout:

1. The text boxes are not properly centered (vertically) anymore. They are off to the top.

Before this PR:
image

After this PR:
image

2. On MacOS, the drop down menu is misplaced (off to the top):
image

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
```diff
@@ -0,0 +1,208 @@
...
+			public void execute(String selectedHistoryItem) {
+				if (selectedHistoryItem != null) {
+					textBar.setText(selectedHistoryItem);
+				}
+			}
+
+		});
+
+		Point barPosition = textBar.toDisplay(0, 0);
+		menu.setPosition(barPosition.x, barPosition.y + dropDown.getWidth(), textBar.getSize().x + dropDown.getWidth());
```
HeikoKlare: I have investigated the MacOS placement issue and found the reason here:
This calculation does not work on MacOS, because the drop down's width returns a default value rather than the actual width. I am not sure whether that implementation is correct/expected, but since the implementation has always been like this, it cannot be changed anyway. Since the bounds contain the correct values, we can adapt the code as follows:
```java
Rectangle dropDownBounds = dropDown.getBounds();
menu.setPosition(barPosition.x, barPosition.y + dropDownBounds.height,
	textBar.getSize().x + dropDownBounds.width);
```

Without the change, it looks like this:
image

With the change, it looks like this:
image

HeikoKlare:
@Wittmaxi I have pushed my proposed changed onto your branch: https://github.com/eclipse-platform/eclipse.platform.ui/compare/c7042b7e8f9f3aec27f4d038fa10befb832d8214..9ff464d8c7f049d28d2fd8e12ff026051b2420f1

Maybe you can have a look whether that is fine for you, because I would like to merge this PR soon, so that we finally have that feature available and more easily find potential problems in productive use. If the changes are not okay, just let me know and I will revert them.

Wittmaxi:
@HeikoKlare the changes look sound, especially the use of Consumer<String> is much better than a custom object. I can't comment on the mac changes. If they work, that's fine.

thank you :)

HeikoKlare:
Thank you for the fast feedback! :-)
With the latest change, the layout on MacOS is as expected (i.e., just like on Windows and Linux). Here are screenshots of the final version from all three OS:

### Windows
image

### MacOS
image

### Linux
image

HeikoKlare:
Test failures are unrelated: #294 #1941

Let's merge this and see how the histories behave in productive use. Thank you for contributing this, @Wittmaxi!

HeikoKlare:
Can you please share the icon sources in the images repo, @Wittmaxi? It's not urgent, we should just not forget about that. I have created a tracking issue for that: https://github.com/eclipse-platform/eclipse.platform.images/issues/83

And I just saw that the "scrolling hints" you added to the search and replace bar hints were removed in some of the recent commits. What that intended or by accident, so that we should apply that change in a follow-up?
```
FindReplaceOverlay_searchBar_message=Find (\u2195 for history)
FindReplaceOverlay_replaceBar_message=Replace (\u2195 for history)
```

Wittmaxi:
@HeikoKlare That's not intended and should be re-added. Thanks for the catch :)

I can add an issue for that, I am not at the computer currently

HeikoKlare:
> I can add an issue for that, I am not at the computer currently

I have just created a quick follow-up PR: #2091
No need to have an issue for that.

HannesWell:
Thank you for the adding the history to the new overlays. In general this works like charm.

The only thing that made me wonder when using this is that I have to press the `UP` key to navigate _downwards_ the list of elements in the history. I found it confusing because when I open the full list its also a drop-_down_ menu where the older elements are further down the list.
Also when I open the overlay and want to get the last search back, then I have to press the `UP` key.
Actually I expected the reverse key-binding: Use DOWN to get to older elements and `UP` to get newer ones. And also that I have to press `DOWN` to get my last search back in.

HeikoKlare:
Thank you for the feedback, @HannesWell! Great to hear that, in general, the functionality works as expected.

I fully agree with your expectation regarding behavior of pressing the up/down keys. I am not sure whether this was ever intended. The only potential reason I found is the behavior of VS code, which moves "downwards" in the history by pressing the "up" button. In this case, I think we should deviate from the VS code behavior because (1) we also have a drop-down menu with an inverse order making the behavior inconsistent and (2) Eclipse users are used to the navigation direction because of the existing find/replace dialog. I have created a PR to invert the navigation direction: #2136

PR #2003, opened 2024-06-26
Refresh option only available for projects eclipse-platform#1904

DaveCarpeneto:
Returns functionality whereby open projects _and_ non-project resources get the 'refresh' contextual menu (option was removed for non-project resources via the fix @ https://github.com/eclipse-platform/eclipse.platform/issues/876 ).

### DIFF
--- a/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/ResourceMgmtActionProvider.java
+++ b/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/ResourceMgmtActionProvider.java
@@ -145,8 +145,10 @@ public class ResourceMgmtActionProvider extends CommonActionProvider {
 			buildAction.selectionChanged(selection);
 			menu.appendToGroup(ICommonMenuConstants.GROUP_BUILD, buildAction);
 		}
-		// To refresh, even if one project is open
-		if (hasOpenProjects) {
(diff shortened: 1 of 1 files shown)


mickaelistria:
What about making the condition `if (projects.stream().anyMatch(IProject::isAccessible))` ?

DaveCarpeneto:
> What about making the condition `if (projects.stream().anyMatch(IProject::isAccessible))` ?

Given the boolean that "does the job" exists already, and this code runs on the main thread: what I put here should be quicker, especially if there's a lot of resources selected. However if you'd rather I went with something like the above please do let me know & I'll try again :-)

mickaelistria:
You're probably right. Let's merge it and revise it later if there are some issues.

mickaelistria:
Thank you

DaveCarpeneto:
> Thank you

Thank **you** ;-) 

Assuming this change stays in, will this be in 2024-09, or will it be in a subsequent release? Asking since I've got a variety of people over here crying about the missing functionality, and timelines seem to help when this happens ...

merks:
Yes, everything committed today will be in a nightly build with in a day:

https://download.eclipse.org/eclipse/updates/4.33-I-builds/

in a SimRel milestone build for M1 and in the final release:

https://download.eclipse.org/releases/milestone/
https://download.eclipse.org/releases/2024-09/

according to this schedule:

https://github.com/eclipse-simrel/.github/blob/main/wiki/SimRel/2024-09.md

PR #2008, opened 2024-06-28
Use consistent key bindings in FindReplaceOverlay on MacOS #2007

### LINKED ISSUE #2007: Find/Replace Overlay: unexpected key bindings on MacOS

## Steps to reproduce

The find/replace overlay can be opened via a keyboard shortcut and also supports a bunch of shortcuts to directly access the find or replace input field, (de-)activate options like regex search or case-sensitive search etc.
On Windows, all these shortcuts are accessible via CTRL+CHAR, while CHAR is some char key like "f", "r" or "w".
On MacOS, however, opening the overlay is done via CMD+F while all the other shortcuts within the overlay are done via CTRL+CHAR (including moving focus to the find input field via CTRL+F). This is because all bindings within the overlay use CTRL as a modifier while the opening action is bound to M1, i.e., the system-dependendent "modifier 1".

From what I experienced with MacOS so far, I would expect all the key bindings to use CMD as a modifier, like it is done for opening the overlay. This can achieved by using `SWT.MOD1` instead of `SWT.CTRL` within the dialog, which still maps to CTRL on Windows.

* OS & version: MacOS Sonoma 14.5
* Eclipse IDE/Platform version (as shown in *Help > About*): I20240627-1800

## Community

HeikoKlare:
Currently, the opening action for the find/replace overlay uses a key binding with the OS-dependent "modifier 1" (mapping to CTRL on Windows and CMD on MacOS), while all key bindings within the overlay (such as activating search options) use key bindings with CTRL as a modifier.

In order to unify the key bindings, this change adapts the key bindings within the overlay to also use the OS-dependent "modifier 1", such that on MacOS the CMD modifier is used. The change also adapts the tooltips to (1) reflect the changed keys and (2) represent the MacOS modifier keys via icons like at all other places.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2007

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -13,6 +13,10 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.findandreplace.overlay;
 
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 import org.osgi.framework.FrameworkUtil;
 
 import org.eclipse.swt.SWT;
@@ -44,6 +48,7 @@ import org.eclipse.swt.widgets.Text;
 import org.eclipse.swt.widgets.ToolItem;
 import org.eclipse.swt.widgets.Widget;
 
+import org.eclipse.jface.bindings.keys.KeyStroke;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.IDialogSettings;
 import org.eclipse.jface.layout.GridDataFactory;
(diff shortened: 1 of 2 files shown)


HeikoKlare:
@BeckerWdf Can you confirm that these changed key bindings make sense on MacOS?

### Example
#### Current
Open find/replace overlay: CMD+F
Move focus to find input field within find/replace overlay: CTRL+F
Activate whole word search: CTRL+W

#### New
Open find/replace overlay: CMD+F
Move focus to find input field within find/replace overlay: CMD+F
Activate whole word search: CMD+W

BeckerWdf:
> Activate whole word search: CMD+W

With your patch it's CMD-Shift-W. But the tooltip still says CTRL-W

BeckerWdf:
and btw:
in the UI CMD should be rendered with the CMD-Icon like e.g. in the menu entries:
image

Can these bindings be changes on the "Keys" preference page?

HeikoKlare:
You are right, the tooltip has to be adapted as well, thanks! I think we should try to adapt the tooltip to use the proper modifier icon like in the menus. I will check that.

But do you agree with me that, in general, CMD instead of CTRL should be used on MacOS for these shortcuts?

> Can these bindings be changes on the "Keys" preference page?

I don't think so, they are currently hard coded.

BeckerWdf:
> But do you agree with me that, in general, CMD instead of CTRL should be used on MacOS for these shortcuts?

Yes.

> I don't think so, they are currently hard coded.

I think this should also be changed.

Wittmaxi:
@BeckerWdf I am not sure if that is sensible - the keybindings are not actions like the actions which are targetted by the settings of the "keys" page. Maybe the keybindings should be editable somewhere else?

BeckerWdf:
> @BeckerWdf I am not sure if that is sensible - the keybindings are not actions like the actions which are targetted by the settings of the "keys" page. 
Is it a must to be an action to be on the "keys" pref page?

Maybe the keybindings should be editable somewhere else?
> Pls. don't a second possibility to configure shortcuts.

Wittmaxi:
> Pls. don't a second possibility to configure shortcuts.

I understand! Then let's use the "keys" page.

> Is it a must to be an action to be on the "keys" pref page?

I am not really sure - from my understanding yes, but my understanding is extremely limited! Maybe somebody can shed light on this?

HeikoKlare:
We have extracted the custom binding of keys into a separate issue: #2015.

I have adapted this PR so that tooltips show the OS-dependent shortcuts and also use modifier icons on MacOS to be consistent with the presentation of shortcuts at other places (in particular, menus).

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -70,6 +74,8 @@
  * @since 3.17
  */
 public class FindReplaceOverlay extends Dialog {
+	private static final String MOD1_KEY_STRING = OS.isMac() ? "⌘" : "Ctrl+"; //$NON-NLS-1$//$NON-NLS-2$
```
BeckerWdf: There's a util that does that for you:
See: org.eclipse.jface.bindings.key.KeySequence#org.eclipse.jface.bindings.key

This is also used for formatting they shortcuts in the empty text editor's onboarding implementation.
See: ModeledPageLayout#addEditorOnboardingCommandId

BeckerWdf:
I have some improvement suggestion regarding the formatting of the keybindings. It's up to you if you want to consider this or not.

HeikoKlare:
> I have some improvement suggestion regarding the formatting of the keybindings. It's up to you if you want to consider this or not.

Improvement suggestions are always appreciated 🙂 I will gladly adopt them.
@BeckerWdf Can you share those suggestions?

BeckerWdf:
>  @BeckerWdf Can you share those suggestions?

Ups....

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: That looks great, thank you! I will try to adapt this PR using that functionality.

BeckerWdf (approved):
Works fine and looks good

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -464,75 +454,105 @@ private void createSearchTools() {
 		@SuppressWarnings("unused")
 		ToolItem separator = searchTools.createToolItem(SWT.SEPARATOR);
 
+		Runnable searchUpOperation = () -> performSearch(false);
 		searchUpButton = new AccessibleToolItemBuilder(searchTools).withStyleBits(SWT.PUSH)
 				.withImage(FindReplaceOverlayImages.get(FindReplaceOverlayImages.KEY_FIND_PREV))
-				.withToolTipText(FindReplaceMessages.FindReplaceOverlay_upSearchButton_toolTip)
-				.withSelectionListener(SelectionListener.widgetSelectedAdapter(e -> performSearch(false))).build();
+				.withToolTipText(MessageFormat.format(FindReplaceMessages.FindReplaceOverlay_upSearchButton_toolTip,
```
Wittmaxi: Nit: How about extracting the formatting of tooltips into a "addShortcutToToolTipText" method?

Wittmaxi (approved):
Very Sensible way of extracting keyboard shortcuts from the handler code and at the same time address the formatting of the shortcuts! Thanks 😀

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: Maybe even add a "withShortcutHint" method to the ToolItemBuilder, which takes a shortcut and adds it to the ToolTip?

HeikoKlare: Good point! I have added a helper method in ca7b586c04e8743cf7e9e833fa79aceb5a9a676e.
Integrating the shortcut into the tool item (or its builder) would be even better. I am already working on a follow-up cleanup that better integrates shortcuts with operations, as well as the operations assigned to shortcuts and "usual" item clicks, which will cover that proposal in some way. Currently, there is quite some manual effort to initialize everything consistently. But that should not prevent us from having this improvement merged and is also easier to review in separation.

Wittmaxi (approved):

PR #2009, opened 2024-06-29
Ignore additional parameter

laeubi:
Currently if one passes more parameters as the command accepts (e.g. a parameter was removed from the command), this suddenly leads to commands no longer being found as null is returned.

Instead now ignore additional parameters and simply continue the loop over the map.


iloveeclipse:
But that would allow commands to start with unexpected number of commands?

laeubi:
> But that would allow commands to start with unexpected number of commands?

Can you explain what exactly is meant here? This method constructs a `ParameterizedCommand` that only contains parameters that are declared by the command, the passed in `Map`  is no longer used.

laeubi:
Only API tools failure in linux GH action.

PR #2013, opened 2024-07-01
Find/Replace overlay: move components into internal package

Wittmaxi:
moves helper classes for the find/replace overlay into an internal package.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
@@ -11,6 +11,7 @@ Export-Package:
  org.eclipse.ui.contentassist,
  org.eclipse.ui.internal.findandreplace;x-friends:="org.eclipse.ui.workbench.texteditor.tests",
  org.eclipse.ui.internal.findandreplace.status;x-friends:="org.eclipse.ui.workbench.texteditor.tests",
+ org.eclipse.ui.internal.findandreplace.overlay;x-friends:="org.eclipse.ui.workbench.texteditor.tests",
  org.eclipse.ui.internal.texteditor;texteditor=split;mandatory:=texteditor;x-friends:="org.eclipse.ui.editors",
  org.eclipse.ui.internal.texteditor.codemining;x-internal:=true,
  org.eclipse.ui.internal.texteditor.quickdiff;x-internal:=true,
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
@@ -11,7 +11,7 @@
  * Contributors:
  *     Vector Informatik GmbH - initial API and implementation
  *******************************************************************************/
-package org.eclipse.ui.texteditor;
+package org.eclipse.ui.internal.findandreplace.overlay;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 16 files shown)


Wittmaxi:
Extracted from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1990

HeikoKlare (commented):
Wouldn't it make sense to also move the `FindReplaceOverlay` to the `org.eclipse.ui.internal.findandreplace.overlay` package? Then the other classes could remain package protected, as they are only supposed to be used by the overlay class (and not any other class in the bundle).

Wittmaxi:
@HeikoKlare 
grafik

FindReplaceOverlay uses package-private methods of StatusTextEditor and for now, I don't want to touch StatusTextEditor. Also, since the Dialog is also in ui.texteditor, I believe it is best if (for now!) we keep the Overlay in ui.texteditor

laeubi:
@Wittmaxi by the way a good way to decouple things is usually using the [Eclipse Adapters](https://www.eclipse.org/articles/Article-Adapters/) (e.g. implements `IAdaptable`) that way one often can get access to some things (e.g. a `SourceViewer`) by doing

`Adapters.adapt(whatever, SourceViewer.class)`

without need to call any methods directly on the target.

HeikoKlare:
I see. Then, however, I am not sure about the value of the proposed refactoring. What's the pain point in keeping the classes where they currently are? In particular, placing the overlay class in a different package than its first-time usage popup and it's images provider class does not look sound to me.

Wittmaxi:
@HeikoKlare I am fine with moving these specific classes back. The reason I want to refactor these classes is that I am expecting many helper classes for the Overlay (For example in the PR adding a search-history) which are not really related to the texteditor-ui.
I understand that by putting the overlay away from it's helpers we reduce cohesion. I will try to use @laeubi 's advice using the adapters to call methods of the SourceViewer

HeikoKlare:
> The reason I want to refactor these classes is that I am expecting many helper classes for the Overlay (For example in the PR adding a search-history) which are not really related to the texteditor-ui.

Thanks for the explanation. Now I understand.

Following the advice from @laeubi sounds good! The following should probably already do the trick:
```java
   Control targetWidget = textEditor.getAdapter(ITextViewer.class).getTextWidget();
```

Wittmaxi:
@HeikoKlare yes! This works :)

HeikoKlare:
> ```java
>   Control targetWidget = textEditor.getAdapter(ITextViewer.class).getTextWidget();
> ```

Access should of course be made via `Adapters`. Sorry for the "inadequate" proposal:
```java
   Control targetWidget = Adapters.adapt(textEditor, ITextViewer.class).getTextWidget();
```

Wittmaxi:
@HeikoKlare what is the difference between your proposals and why use the second one instead of the first one? The first one already works perfectly!

HeikoKlare:
Maybe in this case it is even fine to call `getAdapter()` directly. The Javadoc states:
image

I guess the reason is that there may be more adapters than delivered by `getAdapters()` (as there is an adapter manager that is also queried by `Adapters.adapt()`). In addition, you can call `Adapters.adapt()` for objects of which you do not know whether they implement `IAdaptable` or not. But none of these cases applies here.
@laeubi do you see a reason for using `Adapters.adapt()` instead of directly calling `IAdaptable.getAdapter()` in this case?

laeubi:
> @laeubi do you see a reason for using Adapters.adapt() instead of directly calling IAdaptable.getAdapter() in this case?

`Adapters.adapt() ` is just more general form and you don't need to bother if it implements `IAdaptable`, is `null` and so on ... but its not forbidden to call `IAdaptable` directly. Beside that there is a convenient `Adapters#of` that even returns an `Optional` ...

HeikoKlare (approved):

PR #2014, opened 2024-07-01
Editor preferences: search for find/replace overlay preferences

### LINKED ISSUE #1953: find/replace overlay: The `Use find/replace overlay` preference is hard to locate because keyword definitions are absent
The preference for the find/replace overlay is here:

image

But that preference is hard to find and cannot be found via a keyword search:

image

The keywords are associated here:

image

And defined here:

image

I'm not sure if we need a new keyword definition or if it should be added to the general one?

Here is how that is defined:

image

Probably `find replace overlay` should be added.

See also https://github.com/eclipse-platform/eclipse.platform.ui/issues/1952

Wittmaxi:
Add preference keywords for the preferences of the find/replace overlay

fixes #1953

### How to test
search for the find/replace overlay in the preferences:

### DIFF
--- a/bundles/org.eclipse.ui.editors/plugin.properties
+++ b/bundles/org.eclipse.ui.editors/plugin.properties
@@ -118,7 +118,7 @@ command.showRulerAnnotationInformation.description = Displays annotation informa
 conversionActionSet.label= Convert Line Delimiters
 conversionSubMenu.label= Con&vert Line Delimiters To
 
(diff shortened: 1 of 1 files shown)


Wittmaxi:
@merks this should fix your issue! I don't think it's a good idea to create a new category for search since the different search modes have different parametrizations. Overall, search in Eclipse is spread out into many different solutions which are really hard to combine

HeikoKlare (approved):
Looks good, works as intended. Thank you!

PR #2016, opened 2024-07-01
Find/Replace overlay: reorder buttons

### LINKED ISSUE #1998: Find/Replace overlay: reorder regexp/whole word button
The regexp button disables the "whole word" button. that is counter intuitive, especially when navigating with TAB, when the "whole word" button is in front. i guess it would be better understandable if "whole word" button is after (right of) the regexp button

Wittmaxi:
reorder buttons such that the "whole words" button is not in front,
making the taborder more intuitive in the case that whole word search is
not supported in the current state.

fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1998

### how to test
Ensure that the whole words button is not first in line
grafik


Wittmaxi:
@jukzi please check

HeikoKlare (approved):
Looks good. Trivial change to fix #1998.

PR #2017, opened 2024-07-01
Position find/replace overlay correctly when unmaximizing the window

Wittmaxi:
### how to test

1) maximize window
2) open overlay
3) double-click window title (unmaximize)
4) see that overlay position is still good


HeikoKlare (approved):
Works as intended 👍

PR #2030, opened 2024-07-04
Properly update find/replace overlay size and displayed widgets #2006

### LINKED ISSUE #2006: Find/Replace overlay: Buttons sometimes missing
After resize the Buttons are sometimes missing even if the editor is super big after resizing it:
image
to reproduce:  shrink editor until buttons missing.
then double click the editor title to enlarge it -> buttons still missing

HeikoKlare:
Currently, the widgets in the find/replace overlay (search options, replace expand bar) are not always properly updated on size changes of the target editor. This is complicated by the implementation mixing up the calculation of the overlay size and the calculations to determine which widgets are displayed, which also leads to the missing updates the UI upon certain events requiring the resizing of the overlay.

With this change, the calculations for the overlay size and the determination of the displayed widgets are separated. This also makes a template string for a "compromise text length" obsolete. It also removes the "size gap" between the minimal width of the overlay with the replace expansion bar, taking up 70% of the editor width, and the size of the width of the overlay without the replace expansion bar, taking up 95% of the editor width.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2006

### Before
find_replace_maximize_before

### After
find_replace_maximize_after


HeikoKlare:
@Wittmaxi maybe you want to have a look at this, as it slightly changes the resizing behavior of the overlay.

BeckerWdf:
I could not see this issue (before this change) on macOS. Is this only visible on windows?

HeikoKlare:
> I could not see this issue (before this change) on macOS. Is this only visible on windows?

Indeed, I cannot reproduce the problem on MacOS. But I see it on Windows and Linux.

Wittmaxi:
This looks like a proper fix for the issue, the refactoring-on-the-fly is helpful for the code quality. I also could not find any bugs while testing.

Issue #1991, opened 2024-06-25
Find/Replace overlay should have an X for closing


vogella:
I think we need an faster way to close the Find/Replace overlay. Current process seems to be:

1.) Move Cursor inside the Find Overlay 
2.) Press ESC

VsCode has a button for this, which I think is very intuitive. I was searching for such a button before realizing that most likely I have to click into it and use the escape button to close it.

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/139910/4e618b02-b66e-4aba-ae5a-2de298855eb8)

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/139910/3bd9b999-ad21-4429-9900-71e5328a00d4)

@Wittmaxi WDYT?

HannesWell:
You can also close it by clicking Ctrl+F again (or twice if doesn't have the focus).

Wittmaxi:
@vogella I have been trying to think for the last two days about how I could implement this in the Overlay.
I never thought this would be an issue, but I see why it is now that you point it out.

I will try to come up with a good solution and maybe we iterate a few times

laeubi:
> (or twice if doesn't have the focus).

That's actually something I found feels "buggy" it should not matter if the control has focus or not, it should simply listen to ESC + CTRL+F on the main editor and handle accordingly.

Wittmaxi:
@laeubi that is by design and really important to my use!
The idea is that I want to "quickly" get into the overlay while editing, no matter whether the Overlay is open already or not.

HeikoKlare:
> it should simply listen to ESC + CTRL+F on the main editor and handle accordingly.

I agree that pressing Escape when being in the editor should, in my opinion, also close the overlay. Maybe that would even render an additional close button obsolete, as it makes it easy and fast enough to close the overlay?

laeubi:
>  that is by design and really important to my use!
> The idea is that I want to "quickly" get into the overlay while editing, no matter whether the Overlay is open already or not.

I found it rather "surprising" and the first time I opend the new overlay that ways exactly my problem "how to close that thing?!?" because it now hangs over my code and I have no idea how to close it, e.g. I searched some text, found it, edit something in the document and then the search has no focus (what one can not really "see" in contrast to a dialog where the title-bar usually indicate this) and I needed a while to find out how to close. Performing a shortcut twice to close a dialog is something I never seen before.

[...]

PR #2032, opened 2024-07-04
Add close button to find/replace overlay #1991

HeikoKlare:
The find/replace overlay can currently only be closed via pressing Escape or using the shortcut CTRL+F when the overlay has input focus. This change adds a close button to the overlay for providing an easy and comprehensible way of closing the overlay.

In order to be compatible with different themes on all OS, we can, unfortunately, not reuse some shared close icons (i.e., one already provided in the platform.ui plugins). The icon adopts the color scheme (i.e., the same shade of gray) of the existing search options buttons.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1991

### How it looks like (Windows)
image
image

### Alternatives
As an alternative, this is how it looks like with a preceding separator. 
image
That would clearly separate the single closing operation from the preceding search operations, but in my opinion it bloats up the overlay providing a significant benefit. Still it would be great to have our opinions (on the alternative as well as the proposal in general).

### PR for images repository
The images (including the orignal SVG) are provided here: https://github.com/eclipse-platform/eclipse.platform.images/pull/80

### Relation to other PRs
If this is merged before #2030, that PR will probably need some adaptation for the considerations of the toolbars on size calculation.


HannesWell (approved):
Looks good to me. 

> That would clearly separate the single closing operation from the preceding search operations, but in my opinion it bloats up the overlay providing a significant benefit.

Agree on that. On first sight it would be nice to separate it, but since all buttons on the right of the first separator are some kind of control buttons, I think it is fine to group them all together.

Wittmaxi:
Ia agree that we shouldn't separate the closing-button, a separator would just add more to visually process and does not add any clear benefit.

Code looks sound!

Wittmaxi:
I have documented the second part of the issue (which was only documented in the trace of comments below)
here https://github.com/eclipse-platform/eclipse.platform.ui/issues/2033

Wittmaxi:
I have documented the second part of the issue (which was only documented in the trace of comments below)
here https://github.com/eclipse-platform/eclipse.platform.ui/issues/2033

HeikoKlare:
Thank you for your quick feedback!
I just found noticed that by adding the close button to the existing search tools bar, it will be hidden when the editor becomes too small. That's not intended, so I will update the PR with the close button placed in a different toolbar that is not hidden when space is limited.

HeikoKlare:
I have updated the PR so that the close button will always stay visible (even in the most reduced configuration of the overlay).
find_replace_close

vogella:
Look great, thanks

BeckerWdf (approved):
Looks good and manually test on macOS dark theme is fine.

vogella:
This new close icon can also be used to update the ugly notification close API. See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2035

HeikoKlare:
I consider your positive feedback as a general approval, so I will merge this PR now. In case there are things to change/optimize, we can still do that as a follow-up, but I guess it's good to have an easy way to close the overlay provided as soon as possible.

The failing test is a known random failure (#926) and the freeze period check is just because of M1 promotion, which does not forbid merges anymore.

PR #2036, opened 2024-07-05
Hide find/replace overlay when target editor is too small #1989

### LINKED ISSUE #1989: Find/Replace overlay isn't hidden when the editor is too small
grafik

HeikoKlare:
The find/replace overlay is still visible when it does not vertically fit into the target editor.
With this change, the overlay is hidden as long the target editor has insufficient height. It also streamlines the whole calculation of the overlay positioning to be properly reusable when determining its visibility.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1989

### How it looks like
find_replace_hide

### Discussion points
* Instead of completely hiding the overlay, we could think about keeping some indicator for the user that the overlay is open. That would, however, complicate the functionality and I would expect such a small editor to be a temporary situation while moving/scaling the workbench or part stacks, so the indicator will mostly be unnecessary. I see most benefit of such an indicator when opening the overlay via CTRL+F in a too small editor, where a user now might wonder why the overlay it not shown at all.
* When having the replace tools open, we could also first hide the replace tools instead of completely hiding the overlay. It could however be perceived unexpected when the replace tools close without explicitly triggering it.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -751,7 +751,7 @@ public class FindReplaceOverlay extends Dialog {
 		replaceTools.setVisible(enable);
 	}
 
-	private int getIdealDialogWidth(Rectangle targetBounds) {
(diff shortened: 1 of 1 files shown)


BeckerWdf:
That would, however, complicate the functionality and I would expect such a small editor to be a temporary situation while moving/scaling the workbench or part stacks, so the indicator will mostly be unnecessary.

I would expect that too.

>It could however be perceived unexpected when the replace tools close without explicitly triggering it.

I also agree here.

PR #2037, opened 2024-07-05
Update Notification API to use modern looking X

vogella:
The find/replace overlay introduces a new close icon similar to the close icon used for the tab folder from
https://bugs.eclipse.org/bugs/show_bug.cgi?id=575395

This pull request updates the notification close icon which still looks like the old tab icon before the change in
https://bugs.eclipse.org/bugs/show_bug.cgi?id=575395  and does not fit anymore in the overall look and feel.

A follow-up PR could also update the active folder but find a new "reddisch" color would require more testing on the different OS, while with the regular icon we can build on the testing of https://github.com/eclipse-platform/eclipse.platform.ui/issues/1991

Fixes: #2035

PR #2038, opened 2024-07-05
Version bump org.eclipse.jface.notifications for 4.33 stream

vogella:


HeikoKlare:
@vogella I close this one as you have already committed and merged the version bump with #2037, so rebasing this PR on master will result in an empty commit and close the PR automatically anyway.

PR #2044, opened 2024-07-06
Find/Replace Overlay: close when escape is pressed in the editor

### LINKED ISSUE #2033: Find/Replace Overlay: Close by escape even when in editor
The overlay should react to escape-presses even when it is not focused.

Wittmaxi:
When escape is pressed inside of the editor which is targeted by the find/replace overlay, the overly now closes.

fixes #2033

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -302,6 +302,11 @@ public class FindReplaceOverlay extends Dialog {
 			targetWidget.forceFocus();
 		}
 	};
+	private KeyListener closeOnTargetEscapeListener = KeyListener.keyPressedAdapter(c -> {
+		if (c.keyCode == SWT.ESC) {
+			this.close();
+		}
+	});
 
 	private boolean isPartCurrentlyDisplayedInPartSash() {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -303,6 +303,11 @@ private void focusTargetWidget() {
 			targetWidget.forceFocus();
 		}
 	};
+	private KeyListener closeOnEditorEscapeListener = KeyListener.keyPressedAdapter(c -> {
```
HeikoKlare: very nitpicky: you could generalize this to `closeOnTargetWidgetEscapeListener` or the like, since it is (at least here and in the future hopefully everywhere in the code) not necessarily an editor.

HeikoKlare (approved):
Works as expected and, in my opinion, _significantly_ improves user experience 👍 

The implementation looks sound. I was just thinking whether there might be situations in which adding a key listener to the target widget/editor could lead to some kind of key-handling conflicts, but since it's just a listener, I don't see a risk for that to happen.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: I agree with you, this behavior should still be true when targetting the outline for example! Addressed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/03eaebeb9785bc814ade730a274c6e6708de6c57

PR #2047, opened 2024-07-06
Find/Replace overlay: hide if target editor is too shallow

### LINKED ISSUE #1989: Find/Replace overlay isn't hidden when the editor is too small

Wittmaxi:
Hide the overlay if the targeted overlay is too shallow to fully contain the overlay, show the overlay again as soon as the targetted overlay is high enough again.

fixes #1989


Wittmaxi:
This behavior might be confusing since the overlay goes away completely until the editor is big enough again.
I don't think this will be a big issue since people who use the overlay will most likely try to make the editor big enough to use it - and it's better to hide the overlay than have the overlay obstruct the view over other buttons from different views which we don't target

HeikoKlare:
I guess you did not see #2036, did you? 😉

Wittmaxi:
@HeikoKlare I did not, your code looks far better, has a reasonable refactoring and has less weird bugs. I will close my PR (for now!)

PR #2049, opened 2024-07-06
Add Content assist to the Find/Replace overlay in RegEx Mode

Wittmaxi:
Missing feature which exists in the Dialog but not (yet) in the overlay


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -360,7 +365,8 @@ public int open() {
 
 		getShell().layout();
 		updatePlacementAndVisibility();
-
+		updateContentAssistAvailability();
+		searchBar.forceFocus();
```
HeikoKlare: Duplicate `forceFocus()` should not be necessary in the `open()` method.

HeikoKlare (commented):
In general, this looks good and is a great enhancement for regex search in the find/replace overlay. There is just one functional problem: the enablement is only updated when activating/reactivating the regex option, but this only affects the replace bar if it is open at that point in time. Thus, when opening the replace bar, content assist will always be enabled, because it's enablement is not initialized.

In the following screenshot, I have just opened the overlay and then opened the replace bar. Even though the regex option is disabled, content assist is activated for the replace input field:
image

Issue #1141, opened 2023-09-19
[Feature Request] Surround a selection with brackets and quotes in all kind of text editors


sergeniously:
`({[<"'Hello'">]})`

The greeting above was texted in VSCode's default text editor by typing the word **Hello**, selecting it and sequentially pressing each opening bracket and quote. Is not it amazing?

In all modern and self-respecting IDEs (aka VSCode, CLion, etc) there is a feature that surrounds a selected text with brackets or quotes when you type an opening bracket or quote: `(, {, [, <, ", '`. It is an extremely useful feature for those people who value their time.

I would say it is the only thing I'm missing in Eclipse since Eclipse CDT has a [Bug 431534](https://bugs.eclipse.org/bugs/show_bug.cgi?id=431534) in its `Surround with...` feature which prepends unwanted spaces for `${word_selection}` and so that cannot be considered as an applicable feature.

I've come here with the feature request from **eclipse-cdt** because I think it would be better implement it in the general TextEditor base class so that every Eclipse plugin will automatically support the feature (which could be optionally enabled/disabled in General / Text Editors preferences).

Furthermore, this feature together with another excellent and unique Eclipse feature of selecting text between brackets by double-click could make Eclipse the currently best IDE in the World!

Thank you in advance.

merks:
@sergeniously 

Probably you have plans to contribute.  All modern self-respecting developers contribute to open source:

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md

sergeniously:
@merks 
I would be glad to try to contribute if you at least guided me to source files I have to consider to implement the feature.

merks:
If you create an development environment with all the source (as in the link), then you can determine something similar in the editor as it currently functions, and search the source to find out how that's implemented, and, using the debugger to set breakpoints, to see how it's really working at runtime.  Unfortunately few are likely to have much time to offer detailed guidance; I'm personally not familiar with the editor implementation to provide such guidance.  Perhaps some other other contributor will see this exchange and offer better suggestions.

sratz:
`org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitEditor.BracketInserter` might be a good starting point

codeofandrin:
How is this basic feature not implemented yet ^^

Phillipus:
> How is this basic feature not implemented yet ^^

Because you haven't implemented it.

mickaelistria:
[...]

PR #2051, opened 2024-07-08
Typing open bracket should surround the selected text with brackets

### LINKED ISSUE #865: Typing open bracket should surround the selected text element with brackets

lathapatil:
Surround the selected text with brackets in all editors when an opening bracket is inserted. Brackets included are (,<,[,{,",',`

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/865
#1141 

Linked to #1718

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
@@ -1,0 +1,66 @@
+/*******************************************************************************
+ * Copyright (c) ETAS GmbH 2024, all rights reserved.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     ETAS GmbH - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jface.text;
+
+import java.util.Map;
+
+import org.eclipse.swt.SWT;
+
+import org.eclipse.jface.text.source.ISourceViewer;
+
+/**
+ * @since 3.26 This strategy supports surrounding the selected text with similar opening and closing
+ *        brackets when the text is selected and an opening bracket is inserted.
+ */
+public class SurroundWithBracketsStrategy implements IAutoEditStrategy {
+
+	private ISourceViewer sourceViewer;
+
+	@SuppressWarnings("nls")
+	private final Map<String, String> bracketsMap= Map.of("(", ")", "[", "]", "{", "}", "<", ">", "\"", "\"", "'", "'", "`", "`");
+
+	public SurroundWithBracketsStrategy(ISourceViewer sourceViewer) {
+		this.sourceViewer= sourceViewer;
+	}
+
+	@Override
+	public void customizeDocumentCommand(IDocument document, DocumentCommand command) {
+		if (bracketsMap.containsKey(command.text)) {
+			try {
+				ITextSelection selection= (ITextSelection) sourceViewer.getSelectionProvider().getSelection();
+				if (selection != null && selection.getLength() > 0) {
+					String selectedText= document.get(selection.getOffset(), selection.getLength());
+					String closingBracket= bracketsMap.get(command.text);
+					command.text= command.text + selectedText + closingBracket;
+					command.offset= selection.getOffset();
+					command.length= selection.getLength();
+
+					// Set the caret offset after the opening bracket but before the closing bracket
+					command.caretOffset= command.offset + command.text.length() - closingBracket.length();
+					command.shiftsCaret= false;
+
+					// Run this in a UI thread asynchronously to ensure the selection is updated correctly
+					sourceViewer.getTextWidget().getDisplay().asyncExec(new Runnable() {
+						@Override
+						public void run() {
+							sourceViewer.setSelectedRange(command.offset + 1, selectedText.length());
+						}
+					});
+				}
+			} catch (BadLocationException e) {
+				SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+			}
+		}
+	}
+}
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -14,10 +14,12 @@
  *     Tom Hofmann (Perspectix AG) - bug 297572
  *     Sergey Prigogin (Google) - bug 441448
  *     Angelo Zerr <angelo.zerr@gmail.com> - [CodeMining] Add CodeMining support in SourceViewer - Bug 527515
+ *     Latha Patil (ETAS GmbH) - Issue 865 - Surround the selected text with brackets on inserting a opening bracket
  *******************************************************************************/
 package org.eclipse.jface.text.source;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Stack;
@@ -57,6 +59,7 @@ import org.eclipse.jface.text.ITextViewerExtension2;
 import org.eclipse.jface.text.ITextViewerLifecycle;
 import org.eclipse.jface.text.Position;
 import org.eclipse.jface.text.Region;
+import org.eclipse.jface.text.SurroundWithBracketsStrategy;
 import org.eclipse.jface.text.TextViewer;
 import org.eclipse.jface.text.codemining.ICodeMiningProvider;
 import org.eclipse.jface.text.contentassist.IContentAssistant;
@@ -543,7 +546,12 @@ public class SourceViewer extends TextViewer
 		String[] types= configuration.getConfiguredContentTypes(this);
 		for (String t : types) {
 
-			doSetAutoEditStrategies(configuration.getAutoEditStrategies(this, t), t);
+			IAutoEditStrategy[] autoEditStrategies= configuration.getAutoEditStrategies(this, t);
+			List<IAutoEditStrategy> autoEditStrategiesList= new ArrayList<>(Arrays.asList(autoEditStrategies));
+			autoEditStrategiesList.add(new SurroundWithBracketsStrategy(this));
+			IAutoEditStrategy[] newStrategies= autoEditStrategiesList.toArray(new IAutoEditStrategy[0]);
+
+			doSetAutoEditStrategies(newStrategies, t);
 			setTextDoubleClickStrategy(configuration.getDoubleClickStrategy(this, t), t);
 
 			int[] stateMasks= configuration.getConfiguredTextHoverStateMasks(this, t);
--- a/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/TextViewerTest.java
+++ b/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/TextViewerTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2019 Google, Inc and others.
+ * Copyright (c) 2014, 2024 Google, Inc and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -12,6 +12,7 @@
  * 	   Sergey Prigogin (Google) - initial API and implementation
  * 	   Mickael Istria (Red Hat Inc.) - [Bug 544708] Ctrl+Home
  * 	   Paul Pazderski - [Bug 545530] Test for TextViewer's default IDocumentAdapter implementation.
+ * 	   Latha Patil (ETAS GmbH) - Issue 865 - Test for Surround the selected text with brackets
  *******************************************************************************/
 package org.eclipse.jface.text.tests;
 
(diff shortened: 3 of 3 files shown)


bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * @since 3.26 This strategy supports surrounding the selected text with similar opening and closing
+ *        brackets when the text is selected and an opening bracket is inserted.
+ */
+public class SurroundWithBracketsStrategy implements IAutoEditStrategy {
+
+	private ISourceViewer sourceViewer;
+
+	@SuppressWarnings("nls")
+	private final static Map<String, String> bracketsMap= new HashMap<>(Map.of("(", ")", "[", "]", "{", "}", "<", ">", "\"", "\"", "'", "'", "`", "`"));
```
laeubi: Must it really be a modifiable (static) map here?!?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
```diff
@@ -543,7 +545,12 @@ public void configure(SourceViewerConfiguration configuration) {
 		String[] types= configuration.getConfiguredContentTypes(this);
 		for (String t : types) {
 
-			doSetAutoEditStrategies(configuration.getAutoEditStrategies(this, t), t);
+			IAutoEditStrategy[] autoEditStrategies= configuration.getAutoEditStrategies(this, t);
+			IAutoEditStrategy[] newStrategies= new IAutoEditStrategy[autoEditStrategies.length + 1];
+			System.arraycopy(autoEditStrategies, 0, newStrategies, 0, autoEditStrategies.length);
```
laeubi: Use a List/Stream or Arrays#copyOf might give better results

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+					command.text= command.text + selectedText + closingBracket;
+					command.offset= selection.getOffset();
+					command.length= selection.getLength();
+
+					// Set the caret offset after the opening bracket but before the closing bracket
+					command.caretOffset= command.offset + command.text.length() - closingBracket.length();
+					command.shiftsCaret= false;
+
+					// Run this in a UI thread to ensure the selection is updated correctly
+					sourceViewer.getTextWidget().getDisplay().asyncExec(new Runnable() {
```
laeubi, line 54: Use `Display#execute` to ensure it is run in the UI-Thread.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+						@Override
+						public void run() {
+							sourceViewer.setSelectedRange(command.offset + 1, selectedText.length());
+						}
+					});
+				}
+			} catch (BadLocationException e) {
+				SWT.error(SWT.ERROR_INVALID_ARGUMENT);
```
laeubi, line 62: Should we really raise an SWT error here?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
mickaelistria: What's wrong with the current approach?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
mickaelistria: +1
Using a `Map.of()` construct here would probably fit better.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
laeubi: The comment says to "Run this in a UI thread to ensure the selection is updated correctly" but in fact it schedules an async execution in the UI thread, this might be okay / desired but the comment does not reflect that.

In general such "patterns" have several problems for example if you are already in an UI thread, this will possibly execute an arbitrary number of other UI actions that possibly alter your current state. `Display#execute` on the other hand will execute your command in-place if you are already on an UI thread.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
lathapatil: @laeubi  I referred  `DefaultDocumentAdapter`  class  raising SWT error in methods `getTextRange` and `replaceTextRange` for `BadLocationException`

I even see catch block is empty in many other cases .
What do you Suggest ?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
lathapatil: @laeubi 
If I do not run this in async execution, the selected text is unselected as soon as the text is surrounded by braces, making the selection unavailable for further operations of surrounding with braces.

Maybe I will update the comment to: 'Run this in a UI thread asynchronously to ensure the selection is updated correctly.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
lathapatil: Can it be changed to 

`private final Map<String, String> bracketsMap= Collections.unmodifiableMap(Map.of("(", ")", "[", "]", "{", "}", "<", ">", "\"", "\"", "'", "'", "`", "`"));`

merks: Map.of's result is already unmodifiable.

lathapatil:
@mickaelistria @laeubi is there an issue with this approach or this issue fix is not required now?

mickaelistria:
I have some concern that changing selection that way is likely to conflict with some other auto editor strategies that would run after this one for the same event and alter the command changing the size of the inserted text, or that would implement a similar strategy to also change the selection. But both are minor and I think the best way to figure it out is to merge.
Another thing I noticed, which is not related and is deeper in the editor, is that this very convenient feature wouldn't work for multiple or block text selection. But this seems out of the scope for this change.

So IMO, please just rebase to get the MANIFEST.MF compatible with current master branch and we can merge.

akurtakov:
Content of this PR is plain wrong as it contains many other unrelated changes.

lathapatil:
> Content of this PR is plain wrong as it contains many other unrelated changes.

Its my mistake. I am checking on it.

lathapatil:
> this very convenient feature wouldn't work for multiple or block text selection

@mickaelistria  It works in generic editor for all cases but works in Java editor only if text blocks surrounds with { and `
Is it expected or is further analysis required to fix this in java editor as well?

https://github.com/user-attachments/assets/4604c702-c1fc-4eb8-9fe2-3804008bad7c

mickaelistria:
The Java editor has some extra configuration here and there to tweak the edition experience.
I don't mind merging it if it works for the generic editor and most text editors. The JDT part will probably have to be investigated in JDT directly anyway, without need to fix anything here.
I'm sorry we missed to merge this change earlier, since we're in RC phase, it's most likely better to wait until next stream starts (so we have time to get feedback and improve it if necessary).

mickaelistria:
Now is a good time to merge. Thank you!
Please add a note to the noteworthy documentation about this new feature.

HannesWell:
Thank you for this PR!
@mickaelistria the next time please make sure that such long standing PRs are rebased so that the version-bump bot can do it's duty. This is now missing, but I have created https://github.com/eclipse-platform/eclipse.platform.ui/pull/2598, to apply it.

mickaelistria:
Oops, sorry and thank you for addressing it!

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.eclipse.swt.SWT;
+
+import org.eclipse.jface.text.source.ISourceViewer;
+
+/**
+ * @since 3.26 This strategy supports surrounding the selected text with similar opening and closing
+ *        brackets when the text is selected and an opening bracket is inserted.
+ */
+public class SurroundWithBracketsStrategy implements IAutoEditStrategy {
```
szarnekow, line 26: Does this work with block selections, too? E.g., when the user has multiple selections in the current document?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+	@SuppressWarnings("nls")
+	private final Map<String, String> bracketsMap= Map.of("(", ")", "[", "]", "{", "}", "<", ">", "\"", "\"", "'", "'", "`", "`");
+
+	public SurroundWithBracketsStrategy(ISourceViewer sourceViewer) {
+		this.sourceViewer= sourceViewer;
+	}
+
+	@Override
+	public void customizeDocumentCommand(IDocument document, DocumentCommand command) {
+		if (bracketsMap.containsKey(command.text)) {
```
szarnekow, line 39: I see quite a bunch of null-checks when command.text is accessed in other context.
Wouldn't this throw an NPE since bracketsMap is null-hostile cause it's created via Map.of()?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+			try {
+				ITextSelection selection= (ITextSelection) sourceViewer.getSelectionProvider().getSelection();
```
szarnekow, line 41: DocumentCommand has a package visible field `fSelection` - is that the same or something else then the selection obtained here?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
```diff
@@ -543,7 +546,12 @@ public void configure(SourceViewerConfiguration configuration) {
 		String[] types= configuration.getConfiguredContentTypes(this);
 		for (String t : types) {
 
-			doSetAutoEditStrategies(configuration.getAutoEditStrategies(this, t), t);
+			IAutoEditStrategy[] autoEditStrategies= configuration.getAutoEditStrategies(this, t);
+			List<IAutoEditStrategy> autoEditStrategiesList= new ArrayList<>(Arrays.asList(autoEditStrategies));
+			autoEditStrategiesList.add(new SurroundWithBracketsStrategy(this));
```
szarnekow, line 551: Adding the strategy here makes it incredibly hard to disable. Why did you opt for this solution instead of returning it from `org.eclipse.jface.text.source.SourceViewerConfiguration.getAutoEditStrategies(ISourceViewer, String)` as a default?

szarnekow:
@mickaelistria you suggest that JDT can tweak this, but the way it is configured doesn't allow subclasses to override the behavior. What did you have in mind when you gave green light to merge?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
mickaelistria: That is a good suggestion.
@lathapatil can you please consider the suggested change in a next PR ?

jukzi:
please fix https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1843

iloveeclipse:
> please fix [eclipse-jdt/eclipse.jdt.ui#1843](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1843)

Please also consider comments from @szarnekow and **do not fix** JDT UI, but improve platform code - the change here shouldn't break clients!

iloveeclipse:
Just discovered that this change also broke source hover functionality in Java editor (Shift + Hover over method to see implementation) => https://github.com/eclipse-platform/eclipse.platform.ui/issues/2600

lathapatil:
I will check the root cause and fix it ASAP . Thanks for reviewing .

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
lathapatil: @szarnekow please refer video in this [comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2051#issuecomment-2426175808). 

Are you also asking for the same thing?

mickaelistria: No, it's a different thing (which is likely unsupported yet). For example, in a text `abababa` , try to search for `a` and click "Select All", then you get a multi-selection. In such case, the expectation would be that if someone hits `(` the resulting text becomes `(a)b(a)b(a)b(a)`.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
lathapatil: A null check will be added.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/SurroundWithBracketsStrategy.java
lathapatil: It works for both the `ITextSelection` obtained from the `sourceViewer` and the `fSelection` from the `DocumentCommand`

sergeniously:
Where is the feature? The option is lost in the last Eclipse builds. What happened?

mickaelistria:
The feature seems available by default in most text ediors; but the Java Editor seems to override it so it's not available for Java files. An issue should be opened to eclipse-jdt (if not already done) to make this work there too.

sergeniously:
I use CDT and there are no editors that supports this features. But a few updates ago it'd been working. I'm pretty sure there was an option in `General/Text Editors` preferences. And now it's gone. What am I doing wrong?

eclipse1
eclipse2

PR #2052, opened 2024-07-08
Find/Replace overlay: immediate replace after selected text #2011

### LINKED ISSUE #2011: Find/Replace overlay: Immediate Replace after searching selected text not working

## Steps to reproduce

When selecting a text in a file and opening the Find/Replace overlay the selected text is already entered into the search box.
If one subsequently opens the `Replace` part, types in a replacement and then hits Enter or presses the Replace button nothing happens.
I would expect that the previously selected text is replaced and if I continue pressing enter all other occurrences of that text.
If I press Search (or hit Enter in the Search text field) the next occurrence of the text is found and replace works as expected.
Therefore I assume replace is not working because the overlay is not yet in a 'Match-found' state.

In the Search/Find context it would probably be a bit odd to find the selected as first result, it would mean that one has to press Enter/Search twice to find the next occurrence. Therefore it is fine to start the search there after the selected text.
But in the context of replace I think it is reasonable to start the search before the selected text (or after if one replaces backwards, although that seems not to be possible).

Replace all is working as expected.

* OS & version: Windows 10
* Eclipse IDE/Platform version (as shown in *Help > About*): I20240629-1800

## Community

@Wittmaxi could you please have a look at this?

Wittmaxi:
Work around for a bug of the Editor's implementation of the
FindReplaceTargetExtension3; when initializing the search with an
already selected string, force the overlay to search for it again so
that the Editor can then later replace it if required.

fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2011

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -944,6 +944,7 @@ public class FindReplaceOverlay extends Dialog {
 				selectionText = FindReplaceDocumentAdapter.escapeForRegExPattern(selectionText);
 			}
 			searchBar.setText(selectionText);
+			findReplaceLogic.findAndSelect(findReplaceLogic.getTarget().getSelection().x, searchBar.getText());
 		}
 		searchBar.setSelection(0, searchBar.getText().length());
 	}
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceUITest.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceUITest.java
@@ -353,6 +353,18 @@ public abstract class FindReplaceUITest<AccessType extends IFindReplaceUIAccess>
 		assertThat(fTextViewer.getDocument().get(), is("test ;"));
 	}
 
+	@Test
+	public void testReplaceIfSelectedOnStart() {
+		openTextViewer("abcdefg");
+		fTextViewer.setSelection(new TextSelection(2, 2));
+		initializeFindReplaceUIForTextViewer();
+
+		dialog.setReplaceText("aa");
+		dialog.performReplace();
+
+		assertThat(fTextViewer.getDocument().get(), is("abaaefg"));
+	}
+
 	protected AccessType getDialog() {
 		return dialog;
 	}


HeikoKlare (commented):
Functionality is sound and aligns with the `FindReplaceOverlay`. Test is a proper regression test. Thank you!

Just one thing: I do not understand the commit message. Where is the bug in the implementation of the `IFindReplaceTargetExtension3`? And if that was the case, why not fix that implementation?

Wittmaxi:
@HeikoKlare the Bug we need to work around is the one we already encountered when implementing `FindReplaceLogicTest#onlySelectAndReplacesIfFindSuccessfulOnCustomTarget`.
```java
	/**
	 * The TextViewer implementation of IFindReplaceLogic is misleading and not adhering to the
	 * Interface: IFindReplaceTargetExtension3#replaceSelection will NOT replace the selection if
	 * nothing was previously found. This does not generally have to be the case and thus we mock a
	 * target that is setup like this:
	 *
	 * The text contained is {@code ~SELECTEDTEXT~abcd} with ~ marking the boundaries of the
	 * selection. We perform a search for "NOTFOUND" first - the selection stays put since the
	 * string was not found. At this point, we do not want to perform
	 * {@code replaceSelection("NOTFOUND")} since an implementation adhering to the specification of
	 * the function would just overwrite the current selection.
	 */
```

In short: the editor will only replace a selection if it "searched" for it before, even though we would expect it to "blindly replace" the selection.

As for fixing the implementation, I am not sure if we want to get into that; we would need to write tests for the Implementation of FindReplaceTargetExtension3 of the Editor, learn and debug the code and hope that no user depends on this behavior. What do you think?

HeikoKlare:
I see, still I find it confusing to have a commit message stating that there is a bug that we work around without stating explicitly what and where the bug is and that it is accepted to (currently) not address it.
So something like: "The TextViewer implementation of IFindReplaceTargetExtension3.replaceSelection() does not perform a replace operation if not preceeding find operation was called. This address this accepted deviation from the method's specification, this change ...".

Taking a look at that bug led me to a related bug in the current implementation of the overlay: when having the regex option enabled, the string selected in the editor is not properly transformed into a regex when inserting it into the overlay (like done in the existing dialog). This PR will make the situation worse, as a find operation is executed based on that string, which can change the selection in the editor:
find_replace_regex_init

Wittmaxi:
@HeikoKlare we can fix that bug first and then rebase this PR on that PR - what do you think?

HeikoKlare:
> we can fix that bug first and then rebase this PR on that PR - what do you think?

Yes, makes sense 👍

Wittmaxi:
@HeikoKlare I rebased this PR on #2056 as discussed.

Wittmaxi:
I have fixed the merging conflicts.
Since this change mimics the behavior in the Dialog and the faulty behavior with the RegEx is fixed in merged, I believe we are in a good state.
I have changed the commit message to better reflect our thought process in regards to the faulty behavior of the TextViewer's implementation of IFindReplaceTargetExtension3

HeikoKlare (approved):
Works as expected, thank you!
I've rebased on master and resolved newly introduced conflicts. If checks succeed, I consider this ready to be merged.

HeikoKlare:
Failing test is unrelated: #1935

HannesWell:
Thank you for this fix!
Works like a charm now.

Issue #1808, opened 2024-04-10
Random Failing testProblemHover in genericeditor.tests.HoverTest


Wittmaxi:
### Random Failing testProblemHover (org.eclipse.ui.genericeditor.tests.HoverTest) 
Fails on Ubuntu

```null
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.ui.genericeditor.tests.HoverTest.testProblemHover(HoverTest.java:130)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
[...]

PR #2056, opened 2024-07-08
Find/Replace overlay: escape RegEx on init

### LINKED ISSUE #2053: Find/Replace Overlay: doesn't escape RegEx selection when opening
Extracted from @HeikoKlare (https://github.com/eclipse-platform/eclipse.platform.ui/pull/2052#issuecomment-2213498108)

Taking a look at that bug led me to a related bug in the current implementation of the overlay: when having the regex option enabled, the string selected in the editor is not properly transformed into a regex when inserting it into the overlay (like done in the existing dialog). This PR will make the situation worse, as a find operation is executed based on that string, which can change the selection in the editor:
grafik

Wittmaxi:
When initializing the search text from the selection in the editor, escape the text if it should be interpreted as RegEx

fixes #2053

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -51,6 +51,7 @@ import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.jface.resource.JFaceColors;
 import org.eclipse.jface.window.Window;
 
+import org.eclipse.jface.text.FindReplaceDocumentAdapter;
 import org.eclipse.jface.text.IFindReplaceTarget;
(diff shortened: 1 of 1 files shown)


Wittmaxi:
Random fails: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1808 https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

HeikoKlare (commented):
Looks good, thank you!

Would it be possible to add a simple regression test for this? There is already `testActivateDialogWithSelectionActive()` which can be used for another test case covering the regex scenario.

Wittmaxi:
@HeikoKlare I began with adapting that test method and opted to create a new test method instead. The above method tests that the replace operation really only happens in the boundaries (i.e. tests that the scope was set correctly) where as here we only care about whether a search scope was set at all

HeikoKlare (approved):
I will provide an according test case, but I will first submit some adaptation for the test classes overall to overcome limitations preventing easy provision of a test case (see #2061). Let's get this improvement merged and provide the test via a follow-up PR.

PR #2058, opened 2024-07-08
Find/Replace Overlay: store overlay options

### LINKED ISSUE #2055: Find/Replace overlay: remember parameters from earlier
store search options and load on opening - also helps synchronise search options between different editors

Wittmaxi:
Stores the search options of the overlay and actively reloads them, even when switching between overlays.

fixes #2055


Wittmaxi:
@HeikoKlare We had discussed that this PR is not yet ready because we were not quite sure how storing overlay options should act.
Can you please update me if you had any new insights into this functionality? If not, we should move this PR from the "medium" prio to the "unclear" prio until we have a better idea of what we want!

HeikoKlare:
I've answered in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2055#issuecomment-2307361524, as the discussion rather belongs to the issue than the solution (as the latter might change while the issue remains the same).

Wittmaxi:
converted to a draft: the expected behavior needs to be detailed out

PR #2063, opened 2024-07-09
Migrate ProjectionDocumentTest to JUnit5

vogella:


tests/org.eclipse.text.tests/projection/org/eclipse/text/tests/ProjectionDocumentTest.java
```diff
@@ -680,18 +679,18 @@ public void test9_7() {
...
 			fSlaveDocument.addMasterDocumentRange(startOffset, endOffset - startOffset);
 
 			assertSlaveContents(getOriginalMasterContents().substring(80, 140));
 
 			fMasterDocument.replace(endOffset, 1, "x");
 			assertLineInformationConsistency(fSlaveDocument);
 		} catch (BadLocationException e) {
-			assertTrue(false);
+			fail("BadLocationException occurred: " + e.getMessage());
 		}
```
HeikoKlare: nitpicky: the try-catch could be removed in favor of having the test method throwing the exception, such that more information than only the message is provided to the test executor.

HeikoKlare (commented):
The test class adaptations look good, thank you! Only a nitpicky comment (which I do not require to be changed in this PR).

Just note that you need to add the JUnit 5 dependencies to the MANIFEST and migrate the executing test suite (`ProjectionTestSuite`) to JUnit 5 as well.

vogella:
Thanks for the fast feedback @HeikoKlare, updated

HeikoKlare:
I just saw that the project contains two source folders and the project's "main" test suite (`EclipseTextTestSuite`) is placed in the one I missed before. That one needs to be migrated to JUnit 5 as well, as it is still based on JUnit 4 and can thus not execute the contained `ProjectionTestSuite` that has been migrated.

That's the reason why the bot currently reports 161 test cases (exactly the number of tests executed by `ProjectionTestSuite`:
```
This pull request removes 577 and adds 416 tests. Note that renamed tests count towards both.
```

The failing Jenkins build should hopefully be fixed in the next run. I have reported the issue (which happened to most PRs starting from yesterday afternoon on) to the helpdesk and they restarted the server: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4783

PR #2070, opened 2024-07-10
Fix HTMLPrinter.insertPageProlog(StringBuffer, int, RGB, RGB, String)

merks:
- The implementation uses the fgRGB where it should be using the bgRGB instead.


BeckerWdf:
good catch

merks:
Black text on a black ground is really the ultimate darkest dark theme.  😱 

image

PR #2073, opened 2024-07-11
[Sticky Scrolling] Limit visible sticky lines to text widget height

Christopher-Hermann:
When the editor is very small, the sticky lines should not be displayed or the amount of sticky lines should be reduced so that the original source code can be seen.

With this change the sticky lines are automatically reduced on resize of the editor:

### Testing
1. Open a editor
2. Enable sticky scrolling
3. Scroll down so that some sticky lines are shown
4. Reduce the size of the source code editor

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
@@ -77,6 +77,15 @@ import org.eclipse.ui.internal.texteditor.LineNumberColumn;
  */
 public class StickyScrollingControl {
 
+	/**
+	 * This threshold represents the minimum number of source lines that must remain visible in the
+	 * editor. If the StickyScrollingControl's size would result in fewer than this number of lines
+	 * being visible, the height of the StickyScrollingControl will be reduced to ensure visibility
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
```diff
@@ -77,6 +77,12 @@
  */
 public class StickyScrollingControl {
 
+	/**
+	 * The amount of visible source lines that should be shown as minimum. The sticky lines control
```
BeckerWdf: you mean the amount of lines *in the editor*? Would be good to make this explicit in the java doc and manybe in the name of the constant

BeckerWdf (changes requested):

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
Christopher-Hermann: Done

BeckerWdf (approved):

BeckerWdf:
Failed test is documented in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1935

PR #2074, opened 2024-07-11
Find/replace: always select text when search input receives focus #2012

### LINKED ISSUE #2012: Find/Replace overlay: Search text not selected when switching to open overlay

## Steps to reproduce
If the search overlay is already open and I switch the cursor focus from the Editor back to the overlay using Ctrl+F the curer is placed in the beginning of the search text field. But if the text field is not empty, because it either contains the text of a preavuius Search or because the text selected in the editor when pressing Ctrl+F was added, then it would be often convenient if the entire text in the search field were full selected.
If the overlay is closed and opened with an initial search text (i.e. the text selected in the Editor while pressing Ctrl+F), then the behavior is already as desired and the search text is completely selected.

## Community

HeikoKlare:
When the FindReplaceOverlay is opened, the search input field is initialized with the string currently selected in the target editor and this string is selected in the search input field. In consequence, starting to type after opening the dialog replaces the search string. When there is no selection in the target editor, the current search string is kept, but it is not selected. So starting to type will insert further text before the current search string instead of replacing it. This is inconsistent with the existing FindReplaceDialog.

This change improves the behavior to always select the search string when the input field receives focus. According test cases are added.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2012

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -332,11 +332,11 @@ public class FindReplaceOverlay extends Dialog {
 		overlayOpen = true;
 		applyOverlayColors(backgroundToUse, true);
 		updateFromTargetSelection();
+		searchBar.forceFocus();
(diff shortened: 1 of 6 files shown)


Wittmaxi (approved):
Changes address the issue adequately and the tests are sensible and helpful in preventing future regression. My manual testing was smooth

PR #2076, opened 2024-07-12
Find/Replace: keep overlay visible in all kinds of editors #2075

### LINKED ISSUE #2075: Find/Replace Overlay: Lost in Manifest editor

## Steps to reproduce

* Open any `MANIFEST.MF` file.
* Press CTRL+F / CMD+F to open the find/replace overlay.
* Click into the Manifest editor.

I expected: the overlay to still be present and the focus to be in the Manifest editor

But got: the overlay is hidden and cannot be reopened without closing and reopening the editor

find_replace_manifest_lost

* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): I20240711-1800

## Community

HeikoKlare:
The FindReplaceOverlay is sometimes hidden unintendedly when switching focus to the underlying editor, which is not a standard text / source code editor, such as a Manifest editor. The reason is a faulty identification of whether the target part is displayed.

This change corrects the visibility determination of the overlay by comparing the FindReplaceTargets of active and target part instead of comparing the part itself.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2075

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -280,18 +280,16 @@ public class FindReplaceOverlay extends Dialog {
 
 	private boolean isPartCurrentlyDisplayedInPartSash() {
 		IWorkbenchPage activePage = PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage();
-
-		// Check if the targetPart is currently displayed on the active page
-		boolean isPartDisplayed = false;
-
-		if (activePage != null) {
-			IWorkbenchPart activePart = activePage.getActivePart();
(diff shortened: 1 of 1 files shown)


Wittmaxi:
The reason the Overlay hides when the target part is not active is because my knowledge about viewparts is/was very limited when implementing this mechanism. To me, it might make more sense to just always display the overlays as long as we can. What do you think @HeikoKlare? Is this even easily feasible?

HeikoKlare:
> The reason the Overlay hides when the target part is not active is because my knowledge about viewparts is/was very limited when implementing this mechanism. To me, it might make more sense to just always display the overlays as long as we can. What do you think @HeikoKlare? Is this even easily feasible?

That makes sense and yes, that seems to be possible: #2085 
When working on that PR, I found that this one does not even work correctly, because it does not consider MultiPageEditorParts correctly, leading to the overlay not being hidden when moving to other pages of a MultiPageEditorPart (such as different pages of the Manifest editor).

HeikoKlare:
Closing this in favor of:
- #2085

PR #2077, opened 2024-07-12
[Sticky Scrolling] Don't scroll on caret movement to the end of editor

Christopher-Hermann:
When the caret is moved to the end of the editor, the sticky scrolling does not need to adapt the visible source lines. This is for example done at the select all command (STRG+A). 
Without sticky scrolling enable, the complete source code gets selected and the caret is moved to the end without scrolling. When sticky scrolling is enabled, the editor should behave in the same way.

### Testing
1. Open a editor and enable sticky scrolling
2. Select all via STRG+A, the editor should not scroll down to the end of the file


BeckerWdf (approved):

BeckerWdf:
can this somehow be validated with an automated test?

BeckerWdf:
Manually tested on macOS. Fix works fine.

Christopher-Hermann:
> can this somehow be validated with an automated test?

Yes good point, I added some tests. Actually I though it would be much harder to test this feature.

PR #2078, opened 2024-07-12
Find/replace overlay: improve operation and shortcut association

HeikoKlare:
In the FindReplaceOverlay, the assignment of actions to buttons and checkboxes as well as their shortcuts, and the correlation of shortcuts with buttons is not supported by proper OO design by needs to be done manually. This leads to higher code complexity in the FindReplaceOverlay and makes it prone to error when doing changes in that class.

This change extracts the actions assigned to buttons into a separate class that combines these actions with associated shortcuts. They are then automatically registered for the buttons they are assigned to.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
@@ -20,8 +20,7 @@ import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
-import org.eclipse.swt.widgets.ToolBar;
-import org.eclipse.swt.widgets.ToolItem;
+import org.eclipse.swt.widgets.Control;
 
 import org.eclipse.jface.layout.GridLayoutFactory;
 
@@ -35,11 +34,14 @@ import org.eclipse.jface.layout.GridLayoutFactory;
  */
 class AccessibleToolBar extends Composite {
 
-	private List<ToolBar> toolBars = new ArrayList<>();
+	private final List<AccessibleToolItem> accessibleToolItems = new ArrayList<>();
+
+	private final GridLayout layout;
 
 	public AccessibleToolBar(Composite parent) {
 		super(parent, SWT.NONE);
-		GridLayoutFactory.fillDefaults().numColumns(0).spacing(0, 0).margins(0, 0).applyTo(this);
+		this.layout = GridLayoutFactory.fillDefaults().numColumns(0).spacing(0, 0).margins(0, 0).create();
+		this.setLayout(layout);
 	}
 
 	/**
@@ -49,33 +51,26 @@ class AccessibleToolBar extends Composite {
 	 * @param styleBits the StyleBits to apply to the created ToolItem
 	 * @return a newly created ToolItem
 	 */
-	public ToolItem createToolItem(int styleBits) {
-		ToolBar parent = new ToolBar(this, SWT.FLAT | SWT.HORIZONTAL);
-		ToolItem toolItem = new ToolItem(parent, styleBits);
-
-		addToolItemTraverseListener(parent, toolItem);
-
-		((GridLayout) getLayout()).numColumns++;
-
-		toolBars.add(parent);
+	public AccessibleToolItem createToolItem(int styleBits) {
+		AccessibleToolItem toolItem = new AccessibleToolItem(this, styleBits);
+		accessibleToolItems.add(toolItem);
+		layout.numColumns++;
 		return toolItem;
-	}
-
-	private void addToolItemTraverseListener(ToolBar parent, ToolItem result) {
-		parent.addTraverseListener(e -> {
-			if (e.keyCode == SWT.CR || e.keyCode == SWT.KEYPAD_CR) {
-				result.setSelection(!result.getSelection());
-				e.doit = false;
-			}
-		});
 	}
 
 	@Override
 	public void setBackground(Color color) {
 		super.setBackground(color);
-		for (ToolBar bar : toolBars) { // some ToolItems (like SWT.SEPARATOR) don't easily inherit the color from the
-										// parent control.
-			bar.setBackground(color);
+		// some ToolItems (like SWT.SEPARATOR) don't easily inherit the color from the
+		// parent control
+		for (AccessibleToolItem item : accessibleToolItems) {
+			item.setBackground(color);
+		}
+	}
+
+	void registerActionShortcutsAtControl(Control control) {
+		for (AccessibleToolItem item : accessibleToolItems) {
+			item.registerActionShortcutsAtControl(control);
 		}
 	}
 
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItem.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItem.java
@@ -1,0 +1,81 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.ui.internal.findandreplace.overlay;
+
+import java.util.List;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.SelectionListener;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.ToolBar;
+import org.eclipse.swt.widgets.ToolItem;
+
+import org.eclipse.jface.bindings.keys.KeyStroke;
+
+class AccessibleToolItem {
+	private final ToolItem toolItem;
+
+	private FindReplaceOverlayAction action = new FindReplaceOverlayAction(null);
+
+	AccessibleToolItem(Composite parent, int styleBits) {
+		ToolBar toolbar = new ToolBar(parent, SWT.FLAT | SWT.HORIZONTAL);
+		toolItem = new ToolItem(toolbar, styleBits);
+		addToolItemTraverseListener(toolbar);
+	}
+
+	private void addToolItemTraverseListener(ToolBar parent) {
+		parent.addTraverseListener(e -> {
+			if (e.keyCode == SWT.CR || e.keyCode == SWT.KEYPAD_CR) {
+				action.execute();
+				e.doit = false;
+			}
+		});
+	}
+
+	ToolItem getToolItem() {
+		return toolItem;
+	}
+
+	void setBackground(Color color) {
+		toolItem.getParent().setBackground(color);
+	}
+
+	void setImage(Image image) {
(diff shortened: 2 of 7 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
```diff
@@ -35,11 +34,14 @@
...
 
-	private List<ToolBar> toolBars = new ArrayList<>();
+	private final List<AccessibleToolItem> accessibleToolItems = new ArrayList<>();
+
+	private final GridLayout layout;
 
 	public AccessibleToolBar(Composite parent) {
 		super(parent, SWT.NONE);
-		GridLayoutFactory.fillDefaults().numColumns(0).spacing(0, 0).margins(0, 0).applyTo(this);
+		this.layout = GridLayoutFactory.fillDefaults().numColumns(0).spacing(0, 0).margins(0, 0).create();
```
Wittmaxi, line 43: Why is this an improvement of the code?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
```diff
@@ -49,34 +51,25 @@ public AccessibleToolBar(Composite parent) {
...
 	@Override
 	public void setBackground(Color color) {
 		super.setBackground(color);
-		for (ToolBar bar : toolBars) { // some ToolItems (like SWT.SEPARATOR) don't easily inherit the color from the
-										// parent control.
-			bar.setBackground(color);
+		// some ToolItems (like SWT.SEPARATOR) don't easily inherit the color from the
+		// parent control
+		for (AccessibleToolItem item : accessibleToolItems) {
+			item.setBackground(color);
```
Wittmaxi, line 67:
```suggestion
		accessibleToolItems.forEach(item -> item.setBackground(color));
```

in analogy to https://github.com/eclipse-platform/eclipse.platform.ui/pull/2078/files#diff-039028c8c1559043f83229f2942af583d01c16a791712455f9a3ca7abf596997R71

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayAction.java
```diff
@@ -0,0 +1,50 @@
...
+
+	boolean executeIfMatching(KeyStroke keystroke) {
+		if (shortcuts.stream().anyMatch(keystroke::equals)) {
+			execute();
+			return true;
+		}
+		return false;
+	}
+
+	String addShortcutHintToTooltipText(String originalTooltipText) {
```
Wittmaxi, line 43: This doesn't feel like a Responsibility of the `FindReplaceOverlayAction` to me (shortcutHints are closer to UI than to the functionality which this class is providing). In lack of a better idea: how about moving it to the utils?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -420,13 +412,20 @@ private Control createDialog(final Composite parent) {
...
 		createCloseTools();
+		initializeSearchShortcutHandlers();
 
 		container.layout();
 
 		applyDialogFont(container);
 		return container;
 	}
 
+	private void initializeSearchShortcutHandlers() {
```
Wittmaxi, line 453: I would expect this method to be close in the code to this method https://github.com/eclipse-platform/eclipse.platform.ui/pull/2078/files#diff-05a02220b3b1c63bb5dac5a281641fc7a61fc1bb0d83ddfa3d65d0608f7b1c4aR695, increasing code cohesion.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItemBuilder.java
```diff
@@ -49,26 +53,26 @@ public AccessibleToolItemBuilder withToolTipText(String newToolTipText) {
...
+		return this;
+	}
 
-		if (toolTipText != null) {
-			toolItem.setToolTipText(toolTipText);
-		}
+	public ToolItem build() {
+		Objects.requireNonNull(image, "image for tool item must not be null"); //$NON-NLS-1$
+		Objects.requireNonNull(operation, "operation for tool item must not be null"); //$NON-NLS-1$
+		Objects.requireNonNull(toolTipText, "tooltip for tool item must not be null"); //$NON-NLS-1$
```
Wittmaxi: Is there any particular reason to add these constraints?
An `AccessibleToolItemBuilder` does not inherently have an operation or a toolTipText attached and some ToolItems can, for example, not carry an image but rather have a text inside of them (https://archive.eclipse.org/eclipse/downloads/documentation/2.0/html/plugins/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/widgets/ToolItem.html#setText(java.lang.String))

If we want to enforce these constraints, I am in favor of renaming our ToolItems to "AccessibleActionToolItem" or something better

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItem.java
```diff
@@ -0,0 +1,78 @@
...
+		toolItem.setImage(image);
+	}
+
+	void setToolTipText(String text) {
+		toolItem.setToolTipText(action.addShortcutHintToTooltipText(text));
+	}
+
+	void setOperation(Runnable operation, List<KeyStroke> shortcuts) {
+		boolean isCheckbox = (toolItem.getStyle() & SWT.CHECK) != 0;
+		if (isCheckbox) {
```
Wittmaxi, line 64: This method is misleading because a user might specify an operation on a checkbox and wonder why the operation isn't being ran on click of the checkbox.
How about always running the operation and aditionally toggle the selection if the ToolItem is a checkbox?

Wittmaxi:
In general, this is a sound change which greatly improves the quality of the code of the FindReplaceOverlay.
I have noted a few questions above, except for that I strongly support this change.

Thank you for your help :)

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItem.java
Wittmaxi: Nevermind, this works as intrended

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItemBuilder.java
HeikoKlare: The only reason was to avoid that, by accident, one of the item initializations misses to set one of these properties. Since we don't have tests for tool items being properly initialized, this still gives you instant feedback about non properly initialized items.

But I agree that if we want to enforce this, the builder should simply require calling all the `with` methods to a static correctness guarantee instead of performing runtime checks. I have removed them in 38fa391.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: Can you explain why? The method is placed right after it's (single) caller (and should actually be placed after the other methods being called from the caller, but I did not want to reorder all of them in this PR).

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
HeikoKlare: Storing the layout improves static correctness, as you do not need to call `getLayout()` and optimistically cast it to `GridLayout` anymore but have the guarantee that it has the correct type. In case someone changes the code, it will fail statically (instead of having a `ClassCastException` at runtime).

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
HeikoKlare: Thanks! I actually changed the other code instead of this one. I am not a fan of the `Iterable.forEach()` as it unnecessarily complicates stack traces and debugging. But that's kind of personal preference.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayAction.java
HeikoKlare: This class is (almost) only about UI. Shortcuts are a UI concept and this class attached them to an operation without adding any pure business logic. When moving this method to a utility class, you need to pass the shortcuts to that utility method, which would require this action to expose the shortcuts and require from every consumer to properly extract the shortcuts and pass them to that utility. That would render this class almost useless.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
Wittmaxi: Thank you for the insight, I will keep this in mind in the future :)

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
Wittmaxi: Both methods do pretty much the same, except that one is working on the search tools, the other one on the RegEx tools. This is a matter of taste and rather a symptom of this class being so huge than a blocker. I will close this comment for now

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayAction.java
Wittmaxi: The reasoning that shortcuts are UI concepts is very sound. When I read "Action", I instinctively thought of "something like a Runnable", which is why I commented.
Now that I undersatnd this better, I am totally fine with this method

Wittmaxi (approved):

Wittmaxi:
From my side, this can be merged! Added benefit of this PR is that it makes fixing this issue really easy: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2080

Wittmaxi:
Random fail documented here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: I agree that the class is quite huge and should be stripped down 🙂 
Apart from that, method placement is a rather subjective topic. Everyone has their own preference how methods should be ordered. I try to follow clean code ideas of having methods placed in a book-like reading order (not necessarily because that's a great idea but because it is some comprehensible rule to follow), but that also has its limitations.

PR #2045, opened 2024-07-06
Find/Replace Overlay: add help context to Overlay

Wittmaxi:
Add the find/replace help context to the overlay which was also used for the find/replace dialog. In particular, display help for find/replace with regex patterns.

fixes #1994

PR #2079, opened 2024-07-13
IAbstractTextEditorHelpContextIds.FIND_REPLACE_OVERLAY is @since 3.18

merks:
- New API requires a new minor version

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor; singleton:=true
-Bundle-Version: 3.17.500.qualifier
+Bundle-Version: 3.18.0.qualifier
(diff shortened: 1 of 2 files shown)


merks:
@HeikoKlare  @Wittmaxi  @iloveeclipse 

I'm not sure how this slipped past and why the builds don't fail.  I had an API error about this in my workspace.  It can't be correct that something new is `@since 3.17` while the current version is current 3.17.500.

HeikoKlare:
Thank you for the correction, @merks!

It seems like the API error introduced in #2045 was not properly processed by the API tooling and instead "only" considered as a Maven issue, for which we have no quality gate defined, such that all PR checks succeed (https://github.com/eclipse-platform/eclipse.platform.ui/runs/27238010348):
```
<pre><code>1 API problems can&#x27;t be mapped to the compiler log!</code></pre>
````

See also recent Jenkins logs for `master` builds (such as https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/543/maven-warnings/moduleName.678174556/):
Bildschirmfoto 2024-07-14 um 21 28 07

Sorry that I missed that the PR introduced new Maven issues. I will have a better look in future, even if all GitHub checks run green.

HeikoKlare (approved):

laeubi:
Just wanted to mention that you can see it in the checks overview as well:

grafik

PR #2083, opened 2024-07-15
[Sticky Scrolling] Fix line number layout in sticky scrolling control

Christopher-Hermann:
In some cases the line number layout was to small so that the line numbers was not displayed correct.

Before:
Bildschirmfoto 2024-07-15 um 13 44 32

After:
Bildschirmfoto 2024-07-15 um 13 45 39

### Testing
I did not find out when the issue exactly occurred. For me I was able to reproduce the error by:

1. Open a plugin.xml in text editor
2. Scroll down until a sticky line is visible


marcushoepfner (commented):
LGTM. But one question, did you check whether `requestLayout()` works as well?
I have the impression it should be favored over `layout()`.

Christopher-Hermann:
> LGTM. But one question, did you check whether `requestLayout()` works as well? I have the impression it should be favored over `layout()`.

I just tried it, requestLayout() doesn't work.

marcushoepfner (approved):

PR #2084, opened 2024-07-15
Convert shared DnD icons from bmp to png and add 2x zoom variants

ptziegler:
This converts all shared images with the prefix IMG_OBJS_DND from bmp to png (as bmp is slowly being deprecated by Linux) and adds a 200% zoom variant, in order to improve the quality on high DPI screens.


HeikoKlare:
Thank you @ptziegler! Have you created the icons from SVGs, which you could share as well? We store the SVGs as sources in a separate images repository, so that, e.g., higher resolution versions (4x, 8x) could be generated once required or the SVGs may directly be used for rendering icons on-the-fly in the future: https://github.com/eclipse-platform/eclipse.platform.images

ptziegler:
@HeikoKlare 

No. The 2x versions were created by hand. But I suppose it makes sense to also create an SVG to support other levels. I'll try to have a look and update this PR with the new icons, if necessary.

ptziegler:
Now that I think about it... afaik, fast views have been removed a long time ago. Would it make sense to mark those images as deprecated/for removal?

i.e. `IMG_OBJS_DND_TOFASTVIEW_MASK` and `IMG_OBJS_DND_TOFASTVIEW_SOURCE`

ptziegler:
The svg files are contributed via https://github.com/eclipse-platform/eclipse.platform.images/pull/84

HeikoKlare:
Thank you for transforming the icons into SVGs, @ptziegler! In general, they look really good. Only those with round contents look quite edgy in the generated PNGs. How did you generate the PNGs out of the SVGs?

For example, a 64x64 export of the `invalid_source` via Inkscape gives me this:
invalid_source
The one in this PR looks like this:
invalid_source@2x

Edit: is that because the images need to be monochrome, i.e., only contain black and white pixels?

ptziegler:
@HeikoKlare 

Initially, I also used Inkscape to generate those pngs. But for the sake of reproducability, I switched to the [org.eclipse.images.renderer](https://github.com/eclipse-platform/eclipse.platform.images) plugin, which seem to be producing those jagged circles.

I don't think there is any noticable difference in the other images so if you want, I can regenerate `invalid_source` and push again.

HeikoKlare:
Thank you for the explanation! I just tried out the different images and found that the anti-aliased circles produced by Inkscape do actually not work well when used in cursors with the provided mask data.

This is how usage in a cursor looks with the original BMP:
dnd_icon_original

This is how it looks with the icon in this PR:
dnd_icon_edgy

This is how it looks with the one produced by Inkscape:
dnd_icon_mask

To be honest, the original one looks best here. Since the usage of additional mask images seems rather outdated to me, I would propose to provide the PNGs as new shared icons (with a new symbolic name) without the need to provide an additional mask, and mark the old ones as deprecated. This is how the invalid source icon would look like:


I you really need GEF Classic to be compatible with older Eclipse versions (as you mentioned in https://github.com/eclipse/gef-classic/pull/479#issuecomment-2234072603), you could have the cursor creation code test for the availability of the new icon and only then use it with the cursor constructor not accepting a mask, and in case it is not there still access the old one with the mask via fallback logic.

ptziegler:
> I just tried out the different images and found that the anti-aliased circles produced by Inkscape do actually not work well when used in cursors with the provided mask data.

I noticed something similar, which is why I set antialias to `NONE` when exporting it via Inkscape.

ptziegler:
@HeikoKlare For comparison, those are the same images exported with Inkscape when Antialiasing is set to `CAIRO_ANTIALIAS_NONE`. The first image is identical to the current bmp and the second a faithful variant.

invalid_source
invalid_source@2x

> Since the usage of additional mask images seems rather outdated to me, I would propose to provide the PNGs as new shared icons (with a new symbolic name) without the need to provide an additional mask, and mark the old ones as deprecated.

> I you really need GEF Classic to be compatible with older Eclipse versions [...], you could have the cursor creation code test for the availability of the new icon and only then use it with the cursor constructor not accepting a mask,

I really don't like the idea of creating a new entry for effectively the same image and then do a case-distinction within GEF. Especially because other downstream projects won't benefit from those new images, unless they also add the same checks.

> To be honest, the original one looks best here. Since the usage of additional mask images seems rather outdated to me, I would propose to provide the PNGs as new shared icons (with a new symbolic name) without the need to provide an additional mask, and mark the old ones as deprecated.

That I can agree on. What about the following proposal?

I export all SVGs again, with antialiasing disabled. The resulting images should then be an (almost) exact representation of the current BMP files. For the "source" files, I replace the white background with transparency.

Then it shouldn't matter whether applications use the mask, as the background is is already transparent, effectively rendering it redundant.

I can then mark the masks as deprecated, so that they can be removed at a later stage. In our case, we would continue to use the masks to remain backwards compatible, but new code only requires the source images.

Once the masks are removed, one can then reconsider whether it's worth to redo the images _with_ antialiasing enabled, to get smoother curves.

HeikoKlare:
> I export all SVGs again, with antialiasing disabled. The resulting images should then be an (almost) exact representation of the current BMP files. For the "source" files, I replace the white background with transparency.
> 
> Then it shouldn't matter whether applications use the mask, as the background is is already transparent, effectively rendering it redundant.
> 
> I can then mark the masks as deprecated, so that they can be removed at a later stage. In our case, we would continue to use the masks to remain backwards compatible, but new code only requires the source images.
> 
> Once the masks are removed, one can then reconsider whether it's worth to redo the images _with_ antialiasing enabled, to get smoother curves.

I have to admit that I am not in favor if this process, because:
1. It prevents us from having higher quality images for multiple years.
2. It repeatedly changes existing resources and requires consumer adaptations, but instead of providing most benefits from an adaptations, the consumers have to wait two or more years until we remove the deprecated masks and regenerate the images. I consider this worse than deprecating something existing and providing something new (in its final state) that replaces it.
3. There has to be someone keeping track of this process, not only removing deprecated stuff in two years but also regenerating the PNGs. I am afraid that this will never happen.

> I really don't like the idea of creating a new entry for effectively the same image and then do a case-distinction within GEF. Especially because other downstream projects won't benefit from those new images, unless they also add the same checks.

I understand that point but that's only correct for downstream projects that require downwards compatibility with older Eclipse platform versions. In general, if you want to benefit from new features in your dependency, you have to bump your version dependency. We already have strict downwards compatibility requirements for Eclipse and I am not in favor of introducing upwards compatibility constraints so that new features are designed in a way that a dependency of a downstream project to an older version does also profit from newer features. That would prevent progress (as mentioned above).

#

Let's try it with another proposal:
- Update the existing images as you propose (PNGs looking like existing BMPs) and mark the images as well as masks deprecated. Whether the PNGs have transparency information in them or not does not matter.
- Add the new, improved PNGs with transparency information and proper anti aliasing and have the deprecation information for the existing icons refer to the new ones.

Benefits:
- We provide proper new images without the necessity to deal with masks and can simply remove the old ones in two years or so.
- Consumers that do not require downwards compatibility can directly use the "best" possible images and do not have to wait two or more years until the masks are removed.
- Consumers have at least two years to decide when they want to move over to the new images (for reasons of downwards compatibility)

As a consquence, there would be no necessity for GEF Classic to change to the new images and break downwards compatibility for now, but you would still benefit from higher resolution images. What do you think?

ptziegler:
@HeikoKlare That's fine by me. In reference to my other PR over at the eclipse.platform.images PR, a quick side-by-side comparising of all icons. The original ones and the generated one are almost identical, with the only difference being some small changes in the "no-drop" image. But I believe that is almost unnoticeable.

image

I've marked the old constants as deprecated and for-removal and linked everything to the new constants. Those new constants re-use the names of the old constants, just without the "_SOURCE" and "_MASK" suffix.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/ISharedImages.java
```diff
@@ -570,89 +570,143 @@ public interface ISharedImages {
 	 *
 	 * @see Cursor#Cursor(Device, ImageData, ImageData, int, int)
 	 * @since 3.5
+	 * @deprecated Dropped in favor of {@link #IMG_OBJS_DND_LEFT}. The replacement
+	 *             icon does <b>not</b> require a mask.
+	 * @noreference Marked for deletion since never versions are available; see
```
merks: never -> newer

ptziegler: Whoops. Updated.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/ISharedImages.java
```diff
@@ -678,16 +741,98 @@ public interface ISharedImages {
...
 	String IMG_OBJS_DND_TOFASTVIEW_MASK = "IMG_OBJS_DND_TOFASTVIEW_MASK"; //$NON-NLS-1$
 
+	// The following set of constants represent the image pairs that are used
+	// to construct cursors for drag and drop operations within the workbench
+	// Each cursor is represented by two images; the 'source' and the 'mask'
+	// These need to be combined using the following code snippet:
+	// source = getImageDescriptor(sourceId);
+	// mask = getImageDescriptor(maskId);
+	// cursor = new Cursor(display, source.getImageData(), mask.getImageData(), 16,
+	// 16);
```
HeikoKlare: This seems to be an unintended copy. The following constants represent icons that work standalone without a source/mask combination.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchImages.java
```diff
@@ -257,22 +257,30 @@ private static final void declareImages() {
...
-		declareImage(ISharedImages.IMG_OBJS_DND_BOTTOM_MASK, PATH_POINTER + "bottom_mask.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_INVALID_SOURCE, PATH_POINTER + "invalid_source.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_INVALID_MASK, PATH_POINTER + "invalid_mask.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_STACK_SOURCE, PATH_POINTER + "stack_source.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_STACK_MASK, PATH_POINTER + "stack_mask.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_OFFSCREEN_SOURCE, PATH_POINTER + "offscreen_source.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_OFFSCREEN_MASK, PATH_POINTER + "offscreen_mask.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_TOFASTVIEW_SOURCE, PATH_POINTER + "tofastview_source.bmp", true); //$NON-NLS-1$
-		declareImage(ISharedImages.IMG_OBJS_DND_TOFASTVIEW_MASK, PATH_POINTER + "tofastview_mask.bmp", true); //$NON-NLS-1$
+		declareImage(ISharedImages.IMG_OBJS_DND_LEFT_SOURCE, PATH_POINTER + "left_source.png", true); //$NON-NLS-1$
```
HeikoKlare, line 261: A `@SuppressWarnings("removal")` will be required for the `declareImages()` method, as otherwise new warnings will be introduced.

HeikoKlare (commented):
Thank you for the recent update, @ptziegler! The changes look almost fine now. I only have two comments left, which are easy to resolve.

Also thank you for being so patient in the discussion and in finding an agreement. I think/hope that we now found a good solution for both side, the Platform and GEF Classic.

vogella:
Thanks for this work @ptziegler

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchImages.java
ptziegler: Done

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/ISharedImages.java
ptziegler: That is true. I initially wanted to emphasize that the hotspot of those cursors is the center of the image and thus needs to be calculated based on the current device zoom. I simply forgot to push those changes but have done so, now.

HeikoKlare (approved):
Thank you, @ptziegler!
Merging this together with the according images repo PR: https://github.com/eclipse-platform/eclipse.platform.images/pull/84

PR #2085, opened 2024-07-16
Find/Replace: always show overlay when target is visible #2075 #2046

### LINKED ISSUE #2046: Find/Replace Overlay: when closing another view, the overlay completely closes itself
1) open find/replace overlay
2) close another unrelated view
3) give focus back to the editor -> see find/replace overlay gone

Expected: the overlay is hidden when leaving the editor, but we want it to re-appear as soon as we enter the editor again, even if another (unrelated) view was closed.

### LINKED ISSUE #2075: Find/Replace Overlay: Lost in Manifest editor

HeikoKlare:
The FindReplaceOverlay is currently only supposed to be shown when the target editor is visible. Sometimes it is even hidden unintendedly when switching focus to the underlying editor, which is not a single editor but encapsulated in a MultiPageEditorPart, which has different pages. Switching between the pages of that multi-page part, even to pages without a text editor, does not recalculate the visibility of the overlay. The reason for this behavior is a faulty identification of whether the target part is displayed.

This change reimplements the visibility handling for the FindReplaceOverlay based on the target part's visibility. It's goal is to show the overlay if, and only if, the target part is visible, no matter whether it is encapsulated into a multi-editor part or not.
- It encapsulates all target-visility-related functionality into a nested class. This class tracks and exposes the visibility state.
- It adds an IPageChangedListener to get notified about page changes in MultiPageEditorParts.
- It reduces the recalculation of the visibility state to part/page change cases in which the target part is involved.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2075

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2046

Replaces #2076 

### How it looks like

find_replace_target_visibility_improvement

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -16,6 +16,7 @@ package org.eclipse.ui.internal.findandreplace.overlay;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Supplier;
 
 import org.osgi.framework.FrameworkUtil;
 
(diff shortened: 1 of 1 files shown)


Wittmaxi (approved):
Very much appreciated change, very good refactoring. In particular, It is good that the method `isPartCurrentlyDisplayedInPartSash()` is being replaced, which was not implemented well in the first place.

raducoravu:
+1

PR #2086, opened 2024-07-16
Improve and add assertions for random failing HoverTest #926 #1808

HeikoKlare:
This contributes to fixing the randomly failing HoverTests:
- Removes unnecessary focus enforcing: The code checks for the text editor having focus and afterwards uses different methods to force focus to the widget's shell again. This is unnecessary and is only prone to introduce further problems.
- Introduces further assertions: in order to better locate the cause of the failing tests, assertions are added for the conditions, on which the execution waits, to be actually fulfilled.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1808

### DIFF
--- a/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/HoverTest.java
+++ b/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/HoverTest.java
@@ -65,6 +65,8 @@ import org.eclipse.ui.texteditor.AbstractTextEditor;
  */
 public class HoverTest extends AbstratGenericEditorTest {
 
+	private static final int MAXIMUM_HOVER_RETRY_COUNT = 5;
+	
 	@Rule
 	public TestName testName= new TestName();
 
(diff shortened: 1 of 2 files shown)


HeikoKlare:
With these changes, the test randomly fails as follows:
```
hover data not found
java.lang.AssertionError: hover data not found
	at org.eclipse.ui.genericeditor.tests.HoverTest.triggerCompletionAndRetrieveInformationControlManager(HoverTest.java:259)
	at org.eclipse.ui.genericeditor.tests.HoverTest.testEnabledWhenHover(HoverTest.java:97)
```

tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/HoverTest.java
```diff
@@ -227,38 +229,44 @@ private Object getHoverData(AbstractInformationControlManager manager) {
...
+				@Override
+				protected boolean condition() {
+					return getHoverData(textHoverManager) != null;
+				}
+			}.waitForCondition(hoverEvent.display, 6000);
+		}
+		assertTrue("hover data not found", foundHoverData);
 		return textHoverManager;
 	}
+
```
Wittmaxi: nit: this line can be removed form the diff

Wittmaxi (approved):
In general, I have seen SWT tests fail because some control wasn't in focus (happens in the FInd/Replace tests if you click into the window while something is being tested, for example). I think these assertions can at least notify us if something like that is happenign and help us further diagnose what is happening.

PR #2091, opened 2024-07-16
Add hints for history navigation to FindReplaceOverlay

HeikoKlare:
In the FindReplaceOverlay, the search and replace histories can be navigated via pressing arrow up and down keys. This adds an according hint to the hint messages of the input fields.

Before:
image

After:
image

This should already have been introduced with #1990 but was lost during some cleanup / conflict resolution. So this PR complements #1990. Also see https://github.com/eclipse-platform/eclipse.platform.ui/pull/1990#issuecomment-2230767657


HeikoKlare:
Also looks as expected on Linux (Ubuntu 24.04):
image

And looks as expected on MacOS:
image

PR #2095, opened 2024-07-16
Find/replace overlay: fix skipping search history entry #2094

### LINKED ISSUE #2094: Find/Replace overlay: navigating through search history skips entry
When navigating through the search history in the FindReplaceDialog via pressing key down, the first entry in the list is always skipped:
find_replace_history_arrow_skips_first

HeikoKlare:
When navigating through the search history in the FindReplaceOverlay via pressing the down arrow, the first entry in the list is always skipped. This is caused by a faulty overflow calculation, which is corrected by this change.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2094


Wittmaxi:
Not something that should be done in this PR, but this fix begs the question: should we test the Search history somehow?

HeikoKlare:
> should we test the Search history somehow?

Yes, makes sense. We can have unit tests for the business logic (`HistoryStore`) as well as integration tests for the UI, such as simulating key up/down events to navigate through the history for both evaluating that the business logic is integrated correctly and that the UI interaction works as expected.

PR #2096, opened 2024-07-16
Find/replace overlay: reassign to changed shell asynchronously #1945

HeikoKlare:
The synchronous execution of reassigning the FindReplaceOverlay to a shell sporadically makes the application hang on Linux. This makes the execution asynchronous and avoid repeated executions of the reassignment.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1945

Follow-up to #1946

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -883,11 +883,13 @@ public class FindReplaceOverlay extends Dialog {
 			return;
 		}
 		if (isInvalidTargetShell()) {
-			getShell().getDisplay().syncExec(() -> {
-				close();
-				setParentShell(targetPart.getSite().getShell());
-				open();
-				targetPart.setFocus();
+			getShell().getDisplay().asyncExec(() -> {
+				if (isInvalidTargetShell()) {
+					close();
+					setParentShell(targetPart.getSite().getShell());
+					open();
+					targetPart.setFocus();
+				}
 			});
 			return;
 		}


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -883,11 +883,13 @@ private void updatePlacementAndVisibility() {
 			return;
 		}
 		if (isInvalidTargetShell()) {
-			getShell().getDisplay().syncExec(() -> {
-				close();
-				setParentShell(targetPart.getSite().getShell());
-				open();
-				targetPart.setFocus();
+			getShell().getDisplay().asyncExec(() -> {
```
Wittmaxi, line 886: This was the code leading to a bad bug on windows https://github.com/eclipse-platform/eclipse.platform.ui/pull/1946#discussion_r1679612994
The difference with the if being inside of the asyncexec block ***might*** be enough to not have the overlay be called multiple times, but somehow I doubt we really want to always create a new asynchronous job every time some movement was made

HeikoKlare: I am not sure whether I get your points correctly.

The bug we had before was a flickering because the close/open combination was executed multiple times. This is properly prevented by re-checking the necessity for doing this close/open within the asynchronous execution block. On Windows, I looks fine:
find_replace_async_new_parent_shell

> but somehow I doubt we really want to always create a new asynchronous job every time some movement was made

Why do you assume that the execution is done "very time some movement was made"? The asynchronous execution is only initiated when the target of an open find/replace overlay is moved to another shell, which is a rather seldom user interaction (the condition is checked inside and outside of the execution block). I have debugged the behavior again and found that in this scenario the asynchronous execution is spawned either one or two times, such that at most one execution is "unecessary" (i.e., doing nothing as the condition evaluates to false) on every occurence of that scenario. I find this acceptable compared to some more complex tracking if whether the asynchronous execution was already initiated to avoid doing it a second time. What do you think, @Wittmaxi?

Wittmaxi: I missed the outer if-clause, sorry! Thank you for clearing this up, I am in favor of merging this PR now.

PR #2104, opened 2024-07-18
Find/Replace Logic: Allow reset of Incremental Base Position

### LINKED ISSUE #1913: [find/replace logic] cannot easily reset the Incremental Base-Location

Wittmaxi:
Allow the client to reset the base position for incremental search without enabling/disabling the Incremental search.

fixes #1913

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -67,7 +67,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		case FORWARD:
 		case INCREMENTAL:
 			if (shouldInitIncrementalBaseLocation()) {
-				initIncrementalBaseLocation();
+				resetIncrementalBaseLocation();
 			}
 			break;
 		// $CASES-OMITTED$
@@ -87,7 +87,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		}
 
 		if (searchOption == SearchOptions.FORWARD && shouldInitIncrementalBaseLocation()) {
-			initIncrementalBaseLocation();
+			resetIncrementalBaseLocation();
 		}
 	}
 
@@ -136,12 +136,8 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		return isActive(SearchOptions.REGEX) && isTargetSupportingRegEx;
 	}
 
-
-	/**
-	 * Initializes the anchor used as starting point for incremental searching.
-	 *
-	 */
-	private void initIncrementalBaseLocation() {
+	@Override
+	public void resetIncrementalBaseLocation() {
 		if (target != null && isActive(SearchOptions.INCREMENTAL) && !isRegExSearchAvailableAndActive()) {
 			incrementalBaseLocation = target.getSelection();
 		} else {
@@ -159,7 +155,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	 */
 	private void initializeSearchScope() {
 		if (shouldInitIncrementalBaseLocation()) {
-			initIncrementalBaseLocation();
+			resetIncrementalBaseLocation();
 		}
 
 		if (target == null || !(target instanceof IFindReplaceTargetExtension)) {
@@ -343,7 +339,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	 */
 	private boolean performSearch(boolean mustInitIncrementalBaseLocation, String findString) {
 		if (mustInitIncrementalBaseLocation) {
-			initIncrementalBaseLocation();
+			resetIncrementalBaseLocation();
 		}
 		resetStatus();
 
@@ -604,7 +600,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 			}
 		}
 
-		initIncrementalBaseLocation();
+		resetIncrementalBaseLocation();
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


Wittmaxi:
@HeikoKlare I am not sure what is failing here. Can you please have a look?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -141,7 +141,8 @@ public boolean isRegExSearchAvailableAndActive() {
 	 * Initializes the anchor used as starting point for incremental searching.
 	 *
 	 */
-	private void initIncrementalBaseLocation() {
+	@Override
+	public void resetIncrementalBaseLocation() {
```
HeikoKlare, line 140: The comment of this method is obsolete, now that it is implementing an already documented interface method.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
```diff
@@ -174,4 +174,10 @@ public interface IFindReplaceLogic {
 	 */
 	public boolean isWholeWordSearchAvailable(String findString);
 
+	/**
+	 * Initializes the anchor used as starting point for incremental searching.
+	 * Subsequent incremental searches will start from the current selection.
```
HeikoKlare: Maybe you could state what "the current selection" is. Precisely, this is the current selection in the target, isn't it? I would also appreciate a short explanation about how range selections instead of single positions are or should be treated. My current understanding is: forward search operations uses the beginning of the selection, backward search operations use the end of the selection, is that correct?

This allows us in the future to better decide whether some behavior is correct/intended or not and how tests according to the spec need to be written properly.

HeikoKlare (commented):
The change is almost good to go. It provides a dedicated way of resetting the incremental base position, which removes a workaround in the find/replace overlay. I only have two minor comments regarding documentation.

Just one further thought on the design of the FindReplaceLogic: In the long term, I hope that we find a design which does not require the consumer of the FindReplaceLogic to think about when to reset the incremental base position, but that the logic provides a proper level of abstraction in it's functionality so that resetting the incremental base position can be hidden within it. But let's see if that will be possible and reasonable, or if that would require to increase the level of abstraction of the FindReplaceLogic too far. With this change, we can first make the behavior of incremental base position updates consistent and provide regression proper tests before maybe improving the design again.

HeikoKlare:
> @HeikoKlare I am not sure what is failing here. Can you please have a look?

The failure was because of this issue: https://github.com/eclipse-pde/eclipse.pde/issues/1310

Note that this kind of failure makes the build for that OS fail immediately, so you should restart the job as you might miss broken tests or other failures specific to that OS otherwise.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
Wittmaxi: Removed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/f284814f4e9a8c2034555608d15c4837b0841875

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
```diff
@@ -174,4 +174,26 @@ public interface IFindReplaceLogic {
...
+	 * When handling range selections:
+	 * <ul>
+	 * <li>Forward search operations will use the beginning of the selection as the
+	 * starting point.</li>
+	 * <li>Backward search operations will use the end of the selection as the
+	 * starting point.</li>
+	 * </ul>
+	 * </p>
+	 */
+	void resetIncrementalBaseLocation();
```
Wittmaxi, line 197: @HeikoKlare I have made the specification more precise. In particular, this improved specification makes https://github.com/eclipse-platform/eclipse.platform.ui/pull/2109 wrong. It is very good that we specified this more clearly

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
Wittmaxi: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2104#discussion_r1684201267

improved specification!

Wittmaxi:
@HeikoKlare once the tests run through and you give me your OK, I will merge this

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
HeikoKlare: Perfect, thank you for the improvement!

HeikoKlare (approved):
Thank you for the final improvement! This is good to go from my side.

Feel free to merge once the builds/checks have succeeded.

PR #2106, opened 2024-07-18
Perform incremental search by calling performSearch by default instead

### LINKED ISSUE #2105: FindReplaceLogic: performIncrementalSearch breaks contract
The contract for `performIncrementalSearch` is documented as follows:

```java
	/**
	 * Updates the search result after the Text was Modified. Used in combination
	 * with <code>setIncrementalSearch(true)</code>. This method specifically allows
	 * for "search-as-you-type"
	 *
	 * "Search-as-you-type" is not compatible with RegEx-search. This will
	 * initialize the base-location for search (if not initialized already) but will
	 * not update it, meaning that incrementally searching the same string twice in
	 * a row will always yield the same result, unless the Base location was
	 * modified (eg., by performing "find next")
	 *
	 * @param searchString the String that is to be searched
	 */
```

However, the following test fails:

```java
	@Test
	public void testIncrementalSearchContract() {
		String setupString= "test\ntest\ntest";
		TextViewer textViewer= setupTextViewer(setupString);
		textViewer.setSelectedRange(0, 0);
		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
		findReplaceLogic.activate(SearchOptions.FORWARD);
		findReplaceLogic.activate(SearchOptions.WRAP);
		findReplaceLogic.performIncrementalSearch("test");
		assertThat(textViewer.getSelectedRange(), is(new Point(0, 4)));
	}
```

It can be made to work by additionally enabling SearchOptions.INCREMENTAL.

Internally, the method checks whether INCREMENTAL is checked and does nothing if it isn't. This breaks contract!

Wittmaxi:
The method performIncrementalSearch broke it's contract by only
performing incremental search when incremental search was enabled (even
though it didn't specify that as precondition). Now, performSearch with
the right search option is used to perform incremental search.

fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2105

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -325,24 +325,12 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	}
 
 	@Override
-	public boolean performSearch(String searchString) {
-		return performSearch(shouldInitIncrementalBaseLocation(), searchString);
-	}
-
-	/**
-	 * Locates the user's findString in the text of the target.
-	 *
-	 * @param mustInitIncrementalBaseLocation <code>true</code> if base location
-	 *                                        must be initialized
-	 * @param findString                      the String to search for
-	 * @return Whether the string was found in the target
-	 */
-	private boolean performSearch(boolean mustInitIncrementalBaseLocation, String findString) {
-		if (mustInitIncrementalBaseLocation) {
-			resetIncrementalBaseLocation();
-		}
+	public boolean performSearch(String findString) {
 		resetStatus();
 
+		if (isActive(SearchOptions.INCREMENTAL) && !isIncrementalSearchAvailable()) {
+			return false; // Do nothing if search options are not compatible
+		}
 		boolean somethingFound = false;
 
 		if (findString != null && !findString.isEmpty()) {
@@ -650,26 +638,6 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	 */
 	private void statusLineMessage(String message) {
 		statusLineMessage(false, message);
-	}
-
-	@Override
-	public void performIncrementalSearch(String searchString) {
-		resetStatus();
-
-		if (isActive(SearchOptions.INCREMENTAL) && isIncrementalSearchAvailable()) {
-			if (searchString.equals("") && target != null) { //$NON-NLS-1$
-				// empty selection at base location
-				int offset = incrementalBaseLocation.x;
-
-				if (isActive(SearchOptions.FORWARD)) {
-					offset = offset + incrementalBaseLocation.y;
-				}
-
-				findAndSelect(offset, ""); //$NON-NLS-1$
-			} else {
-				performSearch(false, searchString);
-			}
-		}
 	}
 
 	@Override
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
@@ -71,21 +71,6 @@ public interface IFindReplaceLogic {
 	public boolean isIncrementalSearchAvailable();
 
 	/**
-	 * Updates the search result after the Text was Modified. Used in combination
-	 * with <code>setIncrementalSearch(true)</code>. This method specifically allows
-	 * for "search-as-you-type"
-	 *
-	 * "Search-as-you-type" is not compatible with RegEx-search. This will
-	 * initialize the base-location for search (if not initialized already) but will
-	 * not update it, meaning that incrementally searching the same string twice in
-	 * a row will always yield the same result, unless the Base location was
-	 * modified (eg., by performing "find next")
-	 *
-	 * @param searchString the String that is to be searched
-	 */
-	public void performIncrementalSearch(String searchString);
-
-	/**
 	 * Replaces all occurrences of the user's findString with the replace string.
 	 * Indicate to the user the number of replacements that occur.
 	 *
@@ -102,7 +87,11 @@ public interface IFindReplaceLogic {
 	public void performSelectAll(String findString);
 
 	/**
-	 * Locates the user's findString in the target
+	 * Locates the user's findString in the target. If incremental search is
+	 * activated, the search will be performed starting from an incremental search
+	 * position, which can be reset using {@link #resetIncrementalBaseLocation()}.
+	 * If incremental search is activated and RegEx search is activated, nothing
+	 * happens.
 	 *
 	 * @param searchString the String to search for
 	 * @return Whether the string was found in the target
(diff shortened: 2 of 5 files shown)


Wittmaxi:
Failing test documented here https://github.com/eclipse-platform/eclipse.platform.ui/issues/2110

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
```diff
@@ -70,26 +70,11 @@ public interface IFindReplaceLogic {
...
-	 * @param searchString the String that is to be searched
-	 */
-	public void performIncrementalSearch(String searchString);
-
 	/**
 	 * Replaces all occurrences of the user's findString with the replace string.
 	 * Indicate to the user the number of replacements that occur.
 	 *
-	 * @param findString    The string that will be replaced
+	 * @param findString    The string that will be replacedÖ
```
HeikoKlare: Unintended change

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -330,24 +330,12 @@ private boolean replaceSelection(String replaceString) {
...
-	 * @return Whether the string was found in the target
-	 */
-	private boolean performSearch(boolean mustInitIncrementalBaseLocation, String findString) {
-		if (mustInitIncrementalBaseLocation) {
-			resetIncrementalBaseLocation();
-		}
+	public boolean performSearch(String findString) {
 		resetStatus();
 
+		if (isActive(SearchOptions.INCREMENTAL) && !isIncrementalSearchAvailable()) {
```
HeikoKlare, line 331: To me, it looks as if this should do basically the same than `shouldInitIncrementalBaseLocation()`, only that the other method has a simpler check for regex option being activated. I think this implementation is "more correct", so maybe they should/could be combined to one correct version?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
```diff
@@ -292,9 +294,12 @@ public void widgetSelected(SelectionEvent e) {
 						setupFindReplaceLogic();
 						boolean eventRequiresInverseSearchDirection = (e.stateMask & SWT.MODIFIER_MASK) == SWT.SHIFT;
 						boolean forwardSearchActivated = findReplaceLogic.isActive(SearchOptions.FORWARD);
+						boolean incrementalSearchActivated = findReplaceLogic.isActive(SearchOptions.INCREMENTAL);
 						activateInFindReplaceLogicIf(SearchOptions.FORWARD,
 								eventRequiresInverseSearchDirection != forwardSearchActivated);
+						findReplaceLogic.deactivate(SearchOptions.INCREMENTAL);
 						boolean somethingFound = findReplaceLogic.performSearch(getFindString());
+						activateInFindReplaceLogicIf(SearchOptions.INCREMENTAL, incrementalSearchActivated);
```
HeikoKlare, line 302: Why is it necessary to deactive incremental search while performing an explicit search operation? To me this looks like a workaround, because the logic does not behave as xpected.

HeikoKlare (commented):
I have two functional remarks on the proposed change (see detailed comments).

And one very very nitpicky hint w.r.t. language in the commit message:
"remove contract breaking performIncrementalSearch" can be read as "remove the contract that breaks performIncrementalSearch". Thus, compound adjective are written with hyphens, i.e., "remove contract-breaking performIncrementalSearch". But don't take this too serios. I just always have the "hate free speech" vs. "hate-free speech" example in mind 😉

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
Wittmaxi: The specification dictates the inverse (i.e. the old implementation was wrong): if `INCREMENTAL` is active, then every search should be incremental unless incremental is impossible (we didn't specify any exceptions, except for that case)

In this case, the new behavior makes most sense: we temporarily disable the incremental mode in the Logic (i.e. we want to perform a normal search) and re-enable it. Of course, we could add a helper method for that in the Logic but that would only render the interface less coherent (because suddenly we introduce a case where a search option can sometimes be disregarded by the Logic)

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
Wittmaxi: No, this isn't quite the same. The check here checks:

isActive(Incremental) AND is unavailable

the check in shouldInitIncrementalBaseLocation checks for

isActive(Incremental) AND is available

regardless, shouldInitIncrementalBaseLocation changes anyway in this PR: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2108. With that change, I don't think your comment still applies

HeikoKlare: I see, thank you for the explanation.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
HeikoKlare: Okay, I think there is/was a slight misunderstanding about the meaning of "incremental" on my side. So actually, the "incremental" option in the UI does not totally activate incremental search (i.e., the option in the logic), but it activates some "incremental search-as-you-type". With that in mind, the change makes sense. I have to admit that for me it is still hard to comprehend how the interaction between incremental and non-incremental search operations works or is supposed to work (as incremental mode only seems to be used for search-as-you-type). Maybe we can once find an even better design to reflect the behaviors. For now, having the logic in either incremental or non-incremental mode and only a single `performSearch` operation is good to make it easier to understand and use. So if you are confident that this change is semantics-preserving, it is good to go from my side.

Wittmaxi: >  So actually, the "incremental" option in the UI does not totally activate incremental search (i.e., the option in the logic), but it activates some "incremental search-as-you-type"

UI-Wise what happens is that indeed every input to the Editor will perform a "search as you type".
However yes, when the user presses the "search" button of this dialog, he expects to get to the next search result. The logic should not care about this as this is a pure UI decision (some UIs might, for example, not do a "search as you type" and instead use a search button to trigger the incremental search).

An alternative would be to not store Incremental search as a search option but to instead have another method which just runs the incremental search. But then, the Dialog would need to store that it is set to run Incremental search itself, which would split the concern of storing parametrization over two classes.

Wittmaxi:
@HeikoKlare once checks ran, I will merge unless you have further concerns.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
HeikoKlare: > However yes, when the user presses the "search" button of this dialog, he expects to get to the next search result.

Exactly. So the "incremental" option in the dialog is actually not an option for the search overall but activates search-as-you-type for the input box (including incremental search execution when doing search-as-you-type). That's might be confusing (but is not related to your changes but has just always been like this).

PR #2108, opened 2024-07-18
Always update incremental search base location, even if RegEx is enabled

### LINKED ISSUE #1915: [find/replace overlay] disabling RegEx resets to the wrong base-location

Wittmaxi:
Update the incremental Base location even if RegEx is enabled and
incremental search is disabled. Once RegEx is disabled, the incremental
base location is thus still valid.

fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1915

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -138,7 +138,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 
 	@Override
 	public void resetIncrementalBaseLocation() {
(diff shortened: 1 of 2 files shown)


Wittmaxi:
Again, this test fails: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2110

Maybe this isn't a random fail, I will investigate this

PR #2111, opened 2024-07-18
Height reduction of highlight

mvm-sap:
Reduced the thickness of the blue line (highlight) for selected tabs.
For the full details of the change including screenshots see issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114

image

image

See the highlight for the selected Properties tab

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -621,10 +621,12 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 			if (onBottom) {
 				highlightOnTop = !highlightOnTop;
 			}
-			int verticalOffset = highlightOnTop ? 0 : bounds.height - 2;
+			int highlightHeight = 2;
+			int verticalOffset = highlightOnTop ? 0 : bounds.height - (highlightHeight - 1);
 			int horizontalOffset = itemIndex == 0 || cornerSize == SQUARE_CORNER ? 0 : 1;
 			int widthAdjustment = cornerSize == SQUARE_CORNER ? 0 : 1;
-			gc.fillRectangle(bounds.x + horizontalOffset, bounds.y + verticalOffset, bounds.width - widthAdjustment, 3);
+			gc.fillRectangle(bounds.x + horizontalOffset, bounds.y + verticalOffset, bounds.width - widthAdjustment,
+					highlightHeight);
 		}
 
 		if (backgroundPattern != null) {


iloveeclipse:
As usually, screenshot before/after would be nice.

HeikoKlare:
This change even makes sense without the overall theme changes, doesn't it? So would it be possible to remove the theme change commit from this PR and make it only consist of the hight reduction? We can probably can faster agree on that and merge than on the overall theme changes.

laeubi:
And can we please add the relevant screenshots here? Its really hard to guess/follow what will _exactly_ be affected here.

Also as far as I can see this will affect **all user** of CTab and is not only a styling change.

HeikoKlare:
Just tried it out. To get a first impression, find some screenshots below. Looks better in my opinion. The old marker was pretty bold and made the overall header not look layouted correctly (the text looked as if it should be placed more to the top when the marker popped up). Only thing I see is a slight vertical misplacement of the marker (there is 1 pixel white line in the tab header below the marker line). This is better visible when using 200% monitor scaling.

Before 100%:
image

After 100%:
image

Before 200%:
image

After 200%:
image

thomasritter:
@HeikoKlare Fully agree with you. We always wanted to do this separately. Regarding the misalignment. Did not notice this before. Looks like we have to push it down by one pixel. What do you think?

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -624,7 +624,7 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
 			int verticalOffset = highlightOnTop ? 0 : bounds.height - 2;
```
HeikoKlare: I think the cause for the misalignment is the combination with this magic number used for the vertical placement. Changing it to `1` seems to solve the issue.

I propose to put the highlight thickness into a separate variable and use it here for the offset calculation as well as for the line thickness. That improves modifiability and comprehensibility (like always when facing magic numbers 🙂 ).

HeikoKlare:
Thank you for the update! To me, this looks good now. @mvm-sap could you please squash the commits to a single one.

I propose to merge this soon to receive feedback of early adopters that do not take notice of and investigate each PR. Thus, if no one objects, I will merge until end of today.

Here is how it looks now, also incorporating the new theme that has just been merged. I see even more value in this change with the new themes, as now every view/editor stack shows a highlight at the selected tab, which makes the existing highlights look even more bulky.

Before:
image

After:
image

thomasritter:
Looks good to me as well. Yes, once we start using the theme we noticed that the underline is too thick. Looking forward to see how it feels with the smaller one.

HeikoKlare (approved):
Thank you for updating the PR! Looks good now. I've just tested again. Once the builds are done, let's merge it, so it might still be part of M2.

vogella:
Mac failure seems unrelated

HeikoKlare:
> Mac failure seems unrelated

@vogella fyi: this kind of failure is documented here: https://github.com/eclipse-pde/eclipse.pde/issues/1310
The problem is that these kinds of failures abort a build, thus you will not know whether the build would have also failed for other reasons later on. I don't see a problem w.r.t. this PR, but just wanted to share information about this kind of failure, so in case there are PRs with changes that have a chance to fail OS-specifically, it might make sense to rerun such builds.

thomasritter:
I played around with the newest iBuild, yesterday. For me this is a change which needs time to get used to. So, I do not want to give an opinion on it, yet. However, while I personally prefer minimalist designs. I am currently not sure if this is pushing it too far to the minimalist side. In the end, form should follow functions and not the other way around.

@HannesWell already posted some thoughts that the new tab design makes them less visible (https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114#issuecomment-2253186447). The new tab design is the standard in modern UIs and modern means more minimalist, more focus on whitespacing. But his comment motivated me to look at existing designs for tabs, again. Here a collection of examples which might be helpful for you.

**Eclipse new light theme before this change (Line-weight: thick, Font-weight: light)**
image

**Eclipse new light theme with this change (Line-weight: thin, Font-weight: light)**
image

**Intellij light theme (Line-weight: thick, Font-weight: light)**
Screenshot 2024-07-29 at 16 53 07

**VS Code light theme (Line-weight: thin, Font-weight: medium -> not selected tabs are grey)**
Screenshot 2024-07-29 at 16 52 48

**Microsoft Teams (Line-weight: thick, Font-weight: thick)**
Screenshot 2024-07-29 at 16 51 56

**Microsoft Outlook (Line-weight: thick, Font-weight: thick)**
Screenshot 2024-07-29 at 16 51 32

**Some thoughts**
The current tab design is definitely the right direction. Following this direction more tweaks are possible:

1) Keep the old thicker line thickness but round off the edges. Intellij und Microsoft Office products use that trick. The line is thicker but feels a little bit lighter. Here a quick mock up.

Image 30 07 24 at 15 40

2) The new light theme removed a lot of visual clutter including a lot of vertical and horizontal lines. For the view tabs section it might be better to add the line back. It helps separating the view tab section from the view content section. Have a look at the Intellij screenshot to see how this looks like. With the additional line the thin blue line feels a lot stronger. I also mocked an Eclipse version below.

Image 30 07 24 at 15 39

What are your thoughts?

BeckerWdf:
> **Some thoughts** The current tab design is definitely the right direction. Following this direction more tweaks are possible:
> 
> 1. Keep the old thicker line thickness but round off the edges. Intellij und Microsoft Office products use that trick. The line is thicker but feels a little bit lighter. 

> 2. The new light theme removed a lot of visual clutter including a lot of vertical and horizontal lines. For the view tabs section it might be better to add the line back. It helps separating the view tab section from the view content section. Have a look at the Intellij screenshot to see how this looks like. With the additional line the thin blue line feels a lot stronger.
> What are your thoughts?

I personally have the feeling that proposal 2 is a good one. It not only makes the "blue" line feeling thicker but also helps in separating the view tabs from the view content.

HannesWell:
> > 2. The new light theme removed a lot of visual clutter including a lot of vertical and horizontal lines. For the view tabs section it might be better to add the line back. It helps separating the view tab section from the view content section. Have a look at the Intellij screenshot to see how this looks like. With the additional line the thin blue line feels a lot stronger.
> > What are your thoughts?
> 
> I personally have the feeling that proposal 2 is a good one. It not only makes the "blue" line feeling thicker but also helps in separating the view tabs from the view content.

Agree with that. I would also prefer a visual horizontal separation better, especially for inactive tabs.

PR #2119, opened 2024-07-22
Find/Replace Overlay: adapt to target theming

### LINKED ISSUE #2118: Find/Replace Overlay not adapted to theme of Editor
The Find/Replace Overlay should adapt to the theming of the editor, currently it looks like a foreign body

grafik

Wittmaxi:
grafik

The find/replace overlay now gets it's colors from the targetted view if the target is a text editor.

fixes #2118

Before:
grafik

Light theme: 
grafik

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -466,11 +466,17 @@ public class FindReplaceOverlay extends Dialog {
 	 * grab it's color and then immediately dispose of that bar.
 	 */
 	private void retrieveBackgroundColor() {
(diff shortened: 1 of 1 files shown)


Wittmaxi:
@thomasritter @BeckerWdf You are closer to the theme revamp than I am. What do you think of this change? Is this the best way to apply your theme to our overaly?

thomasritter:
Thanks for the heads-up. I will start using it to get a feel for it.

We usually look at other applications for inspirations first. So, I would recommend that you look at:

- VS Code
- Intellij
- Firefox, Chrome
- Notepad
- ...

See how they integrated it. Get some inspirations and find out what makes most sense in the Eclipse environment.

Wittmaxi:
@thomasritter Do you think it would make sense to merge this as a kind of "dogfooding"-PR and possibly change or revert it later on?

IntelliJ has the search themed to match with the Editor:
grafik

VScode does the same (except vscode has the editor lighter than the outside)
grafik

The windows editor has a "real overlay", I don't think it really applies to our case
grafik

Based on this, I believe that my change is an improvement for the theming. In any case, the new change in theming makes the Overlay stick out like a sore thumb - we should find *some* solution for it

vogella:
LGTM

PR #2124, opened 2024-07-23
fix issue with folder names like "a, b"

astrograph:
"a, b" (without the quotes) is a valid Windows folder name, but when using Open in -> System Explorer the folder is not shown in the System explorer but the "My Computer" View


mickaelistria:
Thank you!

PR #2125, opened 2024-07-23
Workbench restart is unavailable if --launcher.noRestart is used

dhendriks:
This new argument just got added. See https://github.com/eclipse-equinox/equinox/pull/595. Here we make sure the workbench understands it.

Should be rare that people actually use the option and start the IDE, as it is not meant for that. This PR detects it and informs the user restart is not available, as documented (see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2195).

I haven't tested it. Not sure how to build a new IDE with Equinox I-build and this PR both included.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/E4Workbench.java
+++ b/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/E4Workbench.java
@@ -20,6 +20,7 @@ import java.net.URI;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.UUID;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.e4.core.commands.ExpressionContext;
 import org.eclipse.e4.core.contexts.ContextInjectionFactory;
 import org.eclipse.e4.core.contexts.EclipseContextFactory;
@@ -201,6 +202,12 @@ public class E4Workbench implements IWorkbench {
 
 	@Override
 	public boolean restart() {
+		if (!canRestart()) {
+			String message = "The workbench cannot be restarted. This application was started with the '--launcher.noRestart' argument. Instead of restarting, shut down and start again manually."; //$NON-NLS-1$
+			ILog.get().error(message);
+			return false;
+		}
+
 		this.restart = true;
 		return close();
 	}
@@ -210,6 +217,22 @@ public class E4Workbench implements IWorkbench {
 	 */
 	public boolean isRestart() {
 		return restart;
+	}
+
+	/**
+	 * Returns whether the workbench can be {@link #restart() restarted}.
+	 *
+	 * <p>
+	 * It can not be restarted if the <code>--launcher.noRestart</code> argument is
+	 * given.
+	 * </p>
+	 *
+	 * @return <code>true</code> if the workbench can be restarted.
+	 */
+	public static boolean canRestart() {
+		String commands = System.getProperty("eclipse.commands", ""); //$NON-NLS-1$ //$NON-NLS-2$
+		String NO_RESTART_ARGUMENT = "--launcher.noRestart"; //$NON-NLS-1$
+		return commands.lines().map(String::strip).noneMatch(NO_RESTART_ARGUMENT::equals);
 	}
 
 	/**
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -1404,6 +1404,11 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 	 */
 	/* package */
 	boolean close(int returnCode, final boolean force) {
+		if (returnCode == PlatformUI.RETURN_RESTART && !E4Workbench.canRestart()) {
+			informNoRestart();
+			return false;
+		}
+
 		this.returnCode = returnCode;
 		final boolean[] ret = new boolean[1];
 		BusyIndicator.showWhile(null, () -> ret[0] = busyClose(force));
@@ -2555,11 +2560,21 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 	@Override
 	public boolean restart() {
+		if (!E4Workbench.canRestart()) {
+			informNoRestart();
+			return false;
+		}
+
 		return close(PlatformUI.RETURN_RESTART, false);
 	}
 
 	@Override
 	public boolean restart(boolean useCurrrentWorkspace) {
+		if (!E4Workbench.canRestart()) {
+			informNoRestart();
+			return false;
+		}
+
 		if (Platform.inDevelopmentMode()) {
 			// In development mode, command line parameters cannot be changed and restart
 			// will always be EXIT_RESTART. Also see setRestartArguments method
@@ -2581,6 +2596,18 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 			}
 		}
 		return close(PlatformUI.RETURN_RESTART, false);
+	}
+
+	/**
+	 * Inform the user that the workbench can't be restarted. If the workbench UI is
+	 * not available, this method does nothing.
+	 */
+	private void informNoRestart() {
+		final Display display = getDisplay();
+		if (display != null && !display.isDisposed()) {
+			MessageDialog.openError(null, WorkbenchMessages.Workbench_CantRestart_Title,
+					WorkbenchMessages.Workbench_CantRestart_Message);
+		}
 	}
 
 	/**
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
@@ -554,6 +554,12 @@ InstallationDialog_ShellTitle={0} Installation Details
 Workbench_NeedsClose_Title = Restart Needed
 Workbench_NeedsClose_Message = A required plug-in is no longer available and the Workbench needs to be restarted. You will be prompted to save if there is any unsaved work.
 
+Workbench_CantRestart_Title = Restart unavailable
+Workbench_CantRestart_Message = The workbench cannot be restarted.\n\
+This application was started with the '--launcher.noRestart' argument.\n\
+\n\
+Instead of restarting, shut down and start again manually.
+
 ErrorPreferencePage_errorMessage = An error has occurred when creating this preference page.
 
 ListSelection_title = Selection Needed
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchMessages.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchMessages.java
@@ -586,6 +586,9 @@ public class WorkbenchMessages extends NLS {
 	public static String Workbench_NeedsClose_Title;
 	public static String Workbench_NeedsClose_Message;
 
+	public static String Workbench_CantRestart_Title;
+	public static String Workbench_CantRestart_Message;
+
 	public static String ErrorPreferencePage_errorMessage;
 
 	public static String ListSelection_title;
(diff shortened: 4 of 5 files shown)


bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/workbench/IWorkbench.java
```diff
@@ -157,6 +157,12 @@ public interface IWorkbench {
 	/**
 	 * restart the workbench
 	 *
+	 * <p>
+	 * If the <code>--launcher.noRestart</code> argument is given, this method
+	 * should indicate to the user that the restart can't be completed, and it
+	 * should then return <code>false</code>.
+	 * </p>
+	 *
```
HannesWell: This `IWorkbench` interface, similar to the other one is marked with `@noimplement`, which means that only Eclipse itself is supposed to implement it and downstream consumers should not implement it (it is not enforced, but strongly discouraged not to do it).

Therefore I suggest to not mention the launcher option here and at the other `IWorkbench` and just leave these two interfaces unchanged.
In general I assume that this option is only used rarely in UI applications (it's actually not intended to be used then) and therefore we should only add reference to it where really necessary, to no complicate the usual cases.

bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/E4Workbench.java
```diff
@@ -212,6 +219,30 @@ public boolean isRestart() {
...
+			commands = ""; //$NON-NLS-1$
+		}
+		String[] args = commands.split("\n"); //$NON-NLS-1$
+		for (String arg : args) {
+			if ("--launcher.noRestart".equals(arg)) { //$NON-NLS-1$
+				return false;
+			}
+		}
+		return true;
+	}
```
HannesWell:
```suggestion
	public static boolean canRestart() {
		String commands = System.getProperty("eclipse.commands", ""); //$NON-NLS-1$ //$NON-NLS-2$
		String NO_RESTART_ARGUMENT = "--launcher.noRestart"; //$NON-NLS-1$
		return commands.lines().map(String::strip).noneMatch(NO_RESTART_ARGUMENT::equals);
	}
```

bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/E4Workbench.java
```diff
@@ -201,6 +202,12 @@ public boolean close() {
 
 	@Override
 	public boolean restart() {
+		if (!canRestart()) {
+			String message = "Can't restart the workbench, because the --launcher.noRestart argument is used."; //$NON-NLS-1$
```
HannesWell:
```suggestion
			String message = "Can't restart the workbench. This application was started with the '--launcher.noRestart' argument."; //$NON-NLS-1$
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
```diff
@@ -554,6 +554,9 @@ InstallationDialog_ShellTitle={0} Installation Details
 Workbench_NeedsClose_Title = Restart Needed
 Workbench_NeedsClose_Message = A required plug-in is no longer available and the Workbench needs to be restarted. You will be prompted to save if there is any unsaved work.
 
+Workbench_CantRestart_Title = Restart Unavailable
```
HannesWell:
```suggestion
Workbench_CantRestart_Title = Restart unavailable
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
```diff
@@ -554,6 +554,9 @@ InstallationDialog_ShellTitle={0} Installation Details
...
+Workbench_CantRestart_Message = Can't restart the workbench, since the '--launcher.noRestart' argument was used when starting the application.
```
HannesWell:
```suggestion
Workbench_CantRestart_Message = Can't restart the workbench. This application was started with the '--launcher.noRestart' argument.\n\
Shutdown and start again manually.
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -1404,6 +1404,11 @@ public boolean close() {
 	 */
 	/* package */
 	boolean close(int returnCode, final boolean force) {
+		if (returnCode == PlatformUI.RETURN_RESTART && !canRestart()) {
```
HannesWell:
```suggestion
		if (returnCode == PlatformUI.RETURN_RESTART && !E4Workbench.canRestart()) {
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2555,11 +2560,21 @@ public WorkbenchWindow openWorkbenchWindow(IAdaptable input, IPerspectiveDescrip
 
 	@Override
 	public boolean restart() {
+		if (!canRestart()) {
```
HannesWell:
```suggestion
		if (!E4Workbench.canRestart()) {
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2555,11 +2560,21 @@ public WorkbenchWindow openWorkbenchWindow(IAdaptable input, IPerspectiveDescrip
...
+			informNoRestart();
+			return false;
+		}
+
 		return close(PlatformUI.RETURN_RESTART, false);
 	}
 
 	@Override
 	public boolean restart(boolean useCurrrentWorkspace) {
+		if (!canRestart()) {
```
HannesWell:
```suggestion
		if (!E4Workbench.canRestart()) {
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2583,6 +2598,42 @@ public boolean restart(boolean useCurrrentWorkspace) {
...
+	 * not available, this method does nothing.
+	 */
+	public void informNoRestart() {
+		final Display display = getDisplay();
+		if (display != null && !display.isDisposed()) {
+			MessageDialog.openError(null, WorkbenchMessages.Workbench_CantRestart_Title,
+					WorkbenchMessages.Workbench_CantRestart_Message);
+		}
+	}
+
```
HannesWell: We can just re-use the method from `E4Workbench`
```suggestion
```

HannesWell (changes requested):
> I haven't tested it. Not sure how to build a new IDE with Equinox I-build and this PR both included.

To test this you can for example create a new `Eclipse Application` launch configuration based on Features and only add the `org.eclipse.sdk` feature. Then in the arguments tab you can simply add `--launcher.noRestart` as program argument.

I have made some comments below for a few enhancements.
In general instead of informing the user via a Error-Dialog or an error log entry just an `UnsupportedOperationException` could be thrown. But the current variant is probably more friendly.

In order to disable the File -> Restart menu entry you should also add the following constructor to `org.eclipse.ui.internal.handlers.RestartWorkbenchHandler`:
```
	public RestartWorkbenchHandler() {
		setBaseEnabled(E4Workbench.canRestart());
	}
```

dhendriks:
> > I haven't tested it. Not sure how to build a new IDE with Equinox I-build and this PR both included.
> 
> To test this you can for example create a new `Eclipse Application` launch configuration based on Features and only add the `org.eclipse.sdk` feature. Then in the arguments tab you can simply add `--launcher.noRestart` as program argument.

Is there a specific Oomph setup for Eclipse UI? Will that then pull in the latest I-build with the new launcher?

> I have made some comments below for a few enhancements. In general instead of informing the user via a Error-Dialog or an error log entry just an `UnsupportedOperationException` could be thrown. But the current variant is probably more friendly.

Indeed, that was the idea. I kept it.

> In order to disable the File -> Restart menu entry you should also add the following constructor to `org.eclipse.ui.internal.handlers.RestartWorkbenchHandler`:
> 
> ```
> 	public RestartWorkbenchHandler() {
> 		setBaseEnabled(E4Workbench.canRestart());
> 	}
> ```

The problem with that, is that a user doesn't know *why* it is disabled. Hence, I opted to keep it enabled, but instead give a message to the user that explains why it can't be completed.

dhendriks:
I made all the changes, as best I could. I squashed the commit. Ready for next review.

merks:
This setup is a bit overkill that it includes everything 

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment

but if you back up the wizard to the Projects page you can deselect projects you don’t need/want.  It installs the nightly Ibuild and populates the target platform with that as well.

dhendriks:
I broke some import statements, but fixed them now.

dhendriks:
File -> Restart works correctly:

image

laeubi:
> File -> Restart works correctly:

I must confess that from a Users Experience Point of View it does not work correctly.

If the action is not available it should either:

1. not shown at all  _- or -_
2. disabled with a proper tooltip text


Also the message is to much technical for a user to understand (too much information) and unlikely to be actionable.

dhendriks:
> > File -> Restart works correctly:
> 
> I must confess that from a Users Experience Point of View it does not work correctly.

It works as designed, but apparently you're looking for a different design.

> If the action is not available it should either:
> 
> 1. not shown at all  _- or -_
> 2. disabled with a proper tooltip text
> 
> Also the message is to much technical for a user to understand (too much information) and unlikely to be actionable.

Not showing it has the downside that a user then does not know why it is not shown, when normally it is shown. Similar with disabling it, as was [suggested before](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2125#pullrequestreview-2203516183). You suggest it is possible to disable, but have a tooltip text that explains why it is not shown? That would be nice. But, I've never heard of special disabled text tooltips. How does one set those?

laeubi:
> Not showing it has the downside that a user then does not know why it is not shown, when normally it is shown.

It depends, there are many things that are not shown in Eclipse, so this is the usual behavior. At least I have never used this Restart function anyways in > 10 years of Eclipse so some even won't notice that.
In combination with this option needing specifically be enabled and requiring to know what it does anyways it does not sounds like anything harmful to hide it completely in such case.

In contrast to that, I now get a hard to understand error popup that indicates I should remove the option I carefully have put into my product configuration before... but why have I put it there if I want to support restarts?!?

>  You suggest it is possible to disable, but have a tooltip text that explains why it is not shown? 

Yes, I have not checked if its possible.

dhendriks:
I've disabled the File -> Restart in case the option is given:

image

HannesWell (approved):
Thank you @dhendriks for the update and of course this PR in general.

> In contrast to that, I now get a hard to understand error popup that indicates I should remove the option I carefully have put into my product configuration before... but why have I put it there if I want to support restarts?!?

I also agree that first of all we should preventing restarts from being possible. A dedicated tool-tip for the disabled state would of course be nice in this case, but with the given use-case I think it's also fine to not have it.
As we already said in the very beginning the new `--launcher.noRestart` option is not intended for UI applications anyway. And if one stil adds them to a config, that person should remember that we noticing the disabled Restart entry. And if still someone adds that option to a product those devs should probably take care on their own to handle it better (e.g. add the tool-tip too).

Even though restart is now disabled if that option is present there can still be cases where the dialog can pop-up. I'm fine with that (again since it should actually not be used I'm also fine with a not 100% perfect solution) but tried to further enhance the readability of the text in the dialog.

With that I'm fine with this change. If there are no objections I'll submit this tomorrow evening.

dhendriks:
> [...] tried to further enhance the readability of the text in the dialog.

I changed the logged message in `E4Workbench` to again be the same as the one from the dialog.

HannesWell:
> > [...] tried to further enhance the readability of the text in the dialog.
> 
> I changed the logged message in `E4Workbench` to again be the same as the one from the dialog.

Thanks. Also for all your contributions on this topic. It was great work. If you are interested in contributing more to Eclipse, that would be welcome. :)

PR #2126, opened 2024-07-23
Only prompt for restart on zoom change when not rescaling at runtime

HeikoKlare:
When the zoom of the primary monitor changes, the workbench asks for a restart to adapt the workbench windows to the changed zoom. With the runtime rescaling functionality recently introduced to SWT, this behavior is only desired if that rescaling functionality is not activated. This change adapts the functionality for showing the restart prompt to only come up if the rescaling functionality is deactivated.

<s>⚠️ Requires recent SWT API changes, thus CI runs will only succeed after the next I-Build.</s>

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchWindow.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchWindow.java
@@ -934,25 +934,31 @@ public class WorkbenchWindow implements IWorkbenchWindow {
 
 			getShell().setData(this);
 			trackShellActivation();
-			/**
-			 * When SWT zoom changes for primary monitor, prompt user to restart Eclipse to
-			 * apply the changes.
-			 */
-			getShell().addListener(SWT.ZoomChanged, event -> {
-				if (getShell().getDisplay().getPrimaryMonitor().equals(getShell().getMonitor())) {
-					int dialogResponse = MessageDialog.open(MessageDialog.QUESTION, getShell(),
-							WorkbenchMessages.Workbench_zoomChangedTitle,
-							WorkbenchMessages.Workbench_zoomChangedMessage, SWT.NONE,
-							WorkbenchMessages.Workbench_RestartButton, WorkbenchMessages.Workbench_DontRestartButton);
-					if (event.doit && dialogResponse == 0) {
-						getWorkbenchImpl().restart(true);
-					}
-				}
-			});
-
+			addZoomChangeListenerToPromptForRestart();
 		} finally {
 			HandlerServiceImpl.pop();
 		}
+	}
+
+	private void addZoomChangeListenerToPromptForRestart() {
+		getShell().addListener(SWT.ZoomChanged, event -> {
+			/**
+			 * Prompt for restart when an SWT zoom change for the primary monitor occurs,
+			 * only when rescaling at runtime is not activated.
+			 */
+			if (getShell().getDisplay().isRescalingAtRuntime()) {
+				return;
+			}
+			if (getShell().getDisplay().getPrimaryMonitor().equals(getShell().getMonitor())) {
+				int dialogResponse = MessageDialog.open(MessageDialog.QUESTION, getShell(),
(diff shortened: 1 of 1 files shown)


akoch-yatta:
Looks good to me. I tested it on Windows and it works as expected for me, having the flag activated or not.

fedejeanne (commented):
I haven't tested yet but I have 2 points to consider:
1. As briefly discussed in our daily, if activating the "auto scale on runtime" fails then the method `Display::isRescalingAtRuntime` should return `false` otherwise the restart dialog will not be shown even though the application is still running in "the old mode" _i.e._ **not** rescaling at runtime. Different PR though, this code is fine here.
2. What would happen if one changes the mode programmatically via `Display::setRescalingAtRuntime`? Would the user be prompted again if one deactivates the rescaling at runtime (_i.e._ "would the old mode be completely restored, with prompt and everything")?

fedejeanne:
>     2. What would happen if one changes the mode programmatically via `Display::setRescalingAtRuntime`? Would the user be prompted again if one deactivates the rescaling at runtime (_i.e._ "would the old mode be completely restored, with prompt and everything")?

EDIT: 2. What would happen if one changes the mode programmatically via `Display::setRescalingAtRuntime` and sets it back to `false`? ...

HeikoKlare:
Thank you for the comments!

> * As briefly discussed in our daily, if activating the "auto scale on runtime" fails then the method `Display::isRescalingAtRuntime` should return `false` otherwise the restart dialog will not be shown even though the application is still running in "the old mode" _i.e._ **not** rescaling at runtime. Different PR though, this code is fine here.

Yes, makes sense. I will provide an according PR to SWT.

> * What would happen if one changes the mode programmatically via `Display::setRescalingAtRuntime`? Would the user be prompted again if one deactivates the rescaling at runtime (_i.e._ "would the old mode be completely restored, with prompt and everything")?

Then one has broken the contract of `setRescalingAtRuntime()`. We could also move the check inside the event handler, but then the check will be unnecessary unless some user broke the API.

HeikoKlare:
See eclipse-platform/eclipse.platform.swt#1363 for the first mentioned issue.

fedejeanne:
> > * What would happen if one changes the mode programmatically via `Display::setRescalingAtRuntime`? Would the user be prompted again if one deactivates the rescaling at runtime (_i.e._ "would the old mode be completely restored, with prompt and everything")?
> 
> Then one has broken the contract of `setRescalingAtRuntime()`. We could also move the check inside the event handler, but then the check will be unnecessary unless some user broke the API.

You mean that part about the undefined behavior? (excerpt from the Javadoc of `Display::setRescalingAtRuntime(boolean)`

```
 * ... This is only safe to call as long as
 * no shell has been created for this display. When changing the value after a
 * shell has been created for this display, the effect is undefined.
...
```

In that case, wouldn't it be better to make `Display::setRescalingAtRuntime` non-api (either reduce visibility or mark it as `@noreference`)?

fedejeanne:
> > > * What would happen if one changes the mode programmatically via `Display::setRescalingAtRuntime`? Would the user be prompted again if one deactivates the rescaling at runtime (_i.e._ "would the old mode be completely restored, with prompt and everything")?
> > 
> > 
> > Then one has broken the contract of `setRescalingAtRuntime()`. We could also move the check inside the event handler, but then the check will be unnecessary unless some user broke the API.
> 
> You mean that part about the undefined behavior? (excerpt from the Javadoc of `Display::setRescalingAtRuntime(boolean)`
> 
> ```
>  * ... This is only safe to call as long as
>  * no shell has been created for this display. When changing the value after a
>  * shell has been created for this display, the effect is undefined.
> ...
> ```
> 
> In that case, wouldn't it be better to make `Display::setRescalingAtRuntime` non-api (either reduce visibility or mark it as `@noreference`)?

Ignore it, we spoke about it already.

HeikoKlare:
Let's proceed with the discussion about the SWT API at SWT. I made the workbench behavior more "resilient" by always having the zoom change listener attached and make it check for the runtime rescaling behavior every time. In constrast to the efforts for actually rescaling the UI, this single check is irrelevant (performance-wise). This way the workbench implementation is able to adapt the restart behavior even if someone changes the rescaling mode at runtime (given that this will be possible at some point in time in the future).

PR #2132, opened 2024-07-26
Correct icon for closing replace bar of find/replace overlay

HeikoKlare:
The icon for closing the replace bar was accidentally added in a version that is vertically flipped compared to the comparable icon in other tools, namely an arrow pointing up instead of down. This change corrects the icon.

Before:
image

After: 
image

Comparison VS Code:
image

Here is the according images repo PR: https://github.com/eclipse-platform/eclipse.platform.images/pull/88


HeikoKlare:
@soethoff fyi, because, if I remember correctly, you also mentioned this issue to me.

vogella:
Looks way better in the new version

PR #2135, opened 2024-07-27
Use DS annotations instead of handcrafted XML

laeubi:
This also fixes a build warning due to usage of wildcard pattern for components, I don't see any reason to use a wildcard here.

Beside that, weight = 100 seems strange here as usually one would use a service ranking instead but I have leave this as-is for now.

FYI @robstryker


HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.ui/pull/2344 this is obsolete isn't it?

PR #2136, opened 2024-07-29
Find/replace overlay: invert keyboard navigation direction for history

HeikoKlare:
To navigate through the history of searches in the find/replace overlay via keyboard buttons, you currently have to press the "up" button for moving "downwards" in the history (i.e., from the most recent to the oldest search input) and "up" for the opposite direction. This is counter-intuitive, as it is the opposite direction of what you expect from the entry order in this drop-down menu of the history entries. In particular, one would usually expect to move to the most recent search input via the "down" button while you currently have to press the "up" button.

This change inverts the keyboard navigation direction for the search history.

As proposed by @HannesWell: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1990#issuecomment-2253140609

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
@@ -69,7 +69,7 @@ public class HistoryTextWrapper extends Composite {
 	private void listenForKeyboardHistoryNavigation() {
 		addKeyListener(KeyListener.keyPressedAdapter(e -> {
 			if (e.keyCode == SWT.ARROW_UP || e.keyCode == SWT.ARROW_DOWN) {
-				int stepDirection = e.keyCode == SWT.ARROW_UP ? 1 : -1;
+				int stepDirection = e.keyCode == SWT.ARROW_UP ? -1 : 1;
 				navigateInHistory(stepDirection);
 			}
 		}));


laeubi:
> In particular, one would usually expect to move to the most recent search input via the "down" button while you currently have to press the "up" button.

In Linux bash the up button gave the most recent entered command (not sure about windows).

HeikoKlare:
> In Linux bash the up button gave the most recent entered command (not sure about windows).

Indeed, thank you! There are argument for and against both directions and I do not have a strong opinion on which one to take here.

Pro _current direction_: consistent with history navigation in other tools like CLIs, VS code search

Pro _inverse direction_: consistent with behavior of drop-down menus and with existing behavior of Eclipse find/replace dialog

In my opinion, the crucial point is the drop-down menu. The reference tools (CLIs, VS code) don't have that. With an open drop-down menu the navigation direction is currently inverse of the one without the menu being open. And you usually navigate to the first entry of the drop-down menu by pressing "down" rather than "up". But I am unusure whether that should outweigh the consistency with history navigation in other tools not providing a drop-down menu.

laeubi:
So is this a feature of the drop down (then I would expect "operating system default" behavior), otherwise one might think about a somewhat "dynamic" behavior:

- If the textfield is empty and I press "up" (==bring **up** the last entry) then it iterates through the history without showing the popup like on CLI, if I press down I at some point in time will come back to the "empty state" and "reset"
- If the textfield is empty and I press "down" (==bring **down** the menu), the popup opens and I can navigate with up/down though the entries in the menu, if I press up I at some point in time will come back to the "empty state" and "reset"

HannesWell:
Thank you Heiko for this one.

> > In particular, one would usually expect to move to the most recent search input via the "down" button while you currently have to press the "up" button.
> 
> In Linux bash the up button gave the most recent entered command (not sure about windows).

It's the same on Windows cmd and powershell.

In general it is of course to look at how other IDEs do something and follow the 'industry standards' so that users don't have to learn everything new when using Eclipse. But at least I personally don't see it as the ultimate goal to behave like a specific other tool, because if one wants exactly the behavior of that tool that person will just use that tool.
Eventually I think the behavior should fit in the specific use story of a specific part.
And while I find the behavior of CLIs to navigate to the last command using the `UP` button intuitive (because the last command is above the current line), in this case I find it more intuitive to use the `DOWN` button to get the previous search because the history is represented by a drop-down list.
I expected the `DOWN` button to be the right one when using it the very first time and even now that I now that I have to press `UP` I do it wrong regularly.

> So is this a feature of the drop down (then I would expect "operating system default" behavior), otherwise one might think about a somewhat "dynamic" behavior:

I'm not sure if making it dynamic would just lead to full confusion.
Nevertheless I think it could indeed be be convenient if pressing down opens the drop-down menu because then one can see more of the history immediately and can target the desired entry better.

Btw. if I open the drop-down menu with an empty text field and then press down, the selection directly jumps to the second entry.

HeikoKlare:
Thank you for your input, @laeubi and @HannesWell! I understand all your arguments and, as mentioned before, I also do not have a strong opinion in one direction or the other. In general, I find the idea for a dynamic behavior pretty interesting, but I am also concerned that it might lead to more confusion than it will help. In particular, it will not be sufficient to apply the "special" navigation behavior when the text field is empty, as it will usually be initialized with the previous search string or the one selected in the editor. So we would need to have precise idea and specification about when that behavior should apply.

> Nevertheless I think it could indeed be be convenient if pressing down opens the drop-down menu because then one can see more of the history immediately and can target the desired entry better.

That would mean that navigating to the most recent entry requires two key presses (down + enter or anything else to close the drop down), while it currently only requires one. That is also consistent with existing behavior. We might have Alt+Down for opening the drop-down menu (like the old find/replace dialog has, even though I didn't know about that until I tried just now). During implementation of the history, I thought about having a view (just like the drop-down menu but without interaction options) showing like the one or two history entries below and above the current one while navigating through the history via up/down buttons. That would, in some way, also address Hannes' idea of seeing more of the history immediately.

I am not completely sure how to proceed. Here are my current thoughts:
* I would like to not overcomplicate the interaction, thus rather keep it simple like it is now with only very specific improvements/adaptations.
* I consider consistency with the existing behavior and the behavior of the drop-down menu more important than consistency with CLI tools. In particular, because in CLI tools the navigation direction is "natural" (as you can see the execution history when moving up in the view) whereas an input field does not have a natural navigation direction (except for the order in the drop-down menu).
* In conclusion, I would currently be in favor of making the change as proposed and maybe have some follow-up tasks for additional improvement ideas. What do you think?

> Btw. if I open the drop-down menu with an empty text field and then press down, the selection directly jumps to the second entry.

That is definitely a bug that should be easy to resolve. I have created #2139 for that. Thank you, Hannes!

laeubi:
I just tried it out with the "classic" dialog and at least under Linux when there are even bindings for Page Down/Up (last entry / first Entry) but for example the Mouse Wheel is not working. Also once I open the popup the focus changes from the text field to the drown down list.

So maybe it is more a matter of if the popup is open or not?

- Popup open -> use up/down as for "visual" navigation
- Popup closed -> use up/down like today

Given that there is already some preferences (that are a bit hard to find as I first looked under General > Search, then General Editors and finally found with General > Editors > Text Editors but for example Structured Text Editors do not offers that?) one *might* even make this configurable ...

But as I just never noticed (in the "classic" dialog) that I can use arrow keys at all .. I'm maybe ot the best one to give final advice so I would leave it up to you and Hannes to decide.

laeubi:
By the way bash even offers [search in history](https://superuser.com/questions/7414/how-can-i-search-the-bash-history-and-rerun-a-command) e.g. with prefix the command with `!` and other crazy stuff... so I wonder if it would be worth to have a "history" control that is simply shared by the "classic" and the "embedded" search so we simply have consistent behavior regardless of what is finally implemented.

BeckerWdf:
> * I consider consistency with the existing behavior and the behavior of the drop-down menu more important than consistency with CLI tools. In particular, because in CLI tools the navigation direction is "natural" (as you can see the execution history when moving up in the view) whereas an input field does not have a natural navigation direction (except for the order in the drop-down menu).

I agree with that.

Wittmaxi:
@BeckerWdf @laeubi the search history component is already implemented in such a way that it is be relatively easy to reuse in the Dialog.
I am in favor of having one particular direction and sticking with it. I don't yet see how a dynamic behavior could help the user.
When implementing the search direction, I copied what VSC did. However, I agree that the semantic of "scrolling through the dropdown with arrow keys" is very appealing and maybe even a bit more accurate.
I would prefer not to open the history box when pressing down (the least new things I can open during a workflow, the better). The compromise with pressing Alt+Down to open the history box sounds like a good idea instead

HannesWell (approved):
> * In conclusion, I would currently be in favor of making the change as proposed and maybe have some follow-up tasks for additional improvement ideas. What do you think?

I would be more than fine with that. :)
I also think it should be consistent within itself and that dynamic behavior can be confusing.

Opening the drop-down with Alt+Down also sound good to me. This way the default stays simple.

HeikoKlare:
I have created:
- #2145
- #2146

I hope they properly summarize the ideas and proposals in this discussion. Feel free to add to them or change them. I have added them with low priority to the find/replace overlay umbrella issue as well. Thank you for all the interesting input!

Let's have this merged to provide the existing behavior of the find/replace dialog also in the find/replace overlay for now and proceed with further discussions in those issues.

PR #2141, opened 2024-07-30
Refresh ImageBasedFrame on ZoomChange of control

akoch-yatta:
This commit adds a listener for the ZoomChanged event to the wrapped control of an ImageBasedFrame. If the listener is notified of this event this means, that the size and position of the wrapped control could be changed due to the zoom change. Therefor, the control is packed and repositioned. 

## How to test
The event is currently thrown for wrapped controls only in win32 with "swt.autoScale.updateOnRuntime"-flag set to true. For the other OS the event will not be thrown for this control in any case to my knowledge.

@HeikoKlare Can you have a look at my change and test it?

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/62 and https://github.com/eclipse-platform/eclipse.platform.swt/issues/131

### DIFF
--- a/bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
+++ b/bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
@@ -66,6 +66,10 @@ public class ImageBasedFrame extends Canvas {
 			ImageBasedFrame frame = (ImageBasedFrame) event.widget;
 			frame.setCursor(null);
 		});
+		toWrap.addListener(SWT.ZoomChanged, event -> {
+			toWrap.pack(true);
+			setFramedControlLocation();
+		});
 
 		addMouseMoveListener(e -> {
 			// Compute the display location for the handle
--- a/bundles/org.eclipse.e4.ui.widgets/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.widgets/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-SymbolicName: org.eclipse.e4.ui.widgets
-Bundle-Version: 1.4.100.qualifier
+Bundle-Version: 1.4.200.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Export-Package: org.eclipse.e4.ui.widgets;x-friends:="org.eclipse.e4.ui.workbench.swt,org.eclipse.e4.ui.workbench.addons.swt"
 Require-Bundle: org.eclipse.swt;bundle-version="[3.6.0,4.0.0)"


bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
```diff
@@ -66,6 +66,10 @@ public ImageBasedFrame(Composite parent, Control toWrap, boolean vertical,
 			ImageBasedFrame frame = (ImageBasedFrame) event.widget;
 			frame.setCursor(null);
 		});
+		toWrap.addListener(SWT.ZoomChanged, event -> {
```
akoch-yatta, line 69: I could guard it with `if (getDisplay().isRescalingAtRuntime())` so the listener is only registered in this case

fedejeanne: If you do that and then some one changes the behavior (calls `setRescalingAtRuntime`) then the listener will be missing. If you want to guard then then always add listener but put the `if` **inside** the logic of the listener. I don't see a reason to guard it in the first place though since `ZoomChanged` events mean that the zoom changed and this logic needs to be executed, am I right? I mean one can always be conservative and guard against unnecessary events, but that would mean that one hides the problem (the problem being the **unnecessary** event).

fedejeanne (dismissed):
Changes LGTM and the PR has the intended effect (no regressions found either).

But there is a build problem. Please look into that.

The request for changes is only because of the build problem (versions?)

fedejeanne:
I just rebased, let's see if that fixes the problem.

bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
akoch-yatta: Yes, I totally agree. Thats why I didn't include the guard yet in the PR. I would leave it as is.

HeikoKlare:
The version of the changed plugin needs to be incremented, as shown in the logs:

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.ui.widgets: Only qualifier changed for (org.eclipse.e4.ui.widgets/1.4.100.v20240731-0633). Expected to have bigger x.y.z than what is available in baseline (1.4.100.v20231201-1637)

akoch-yatta:
> The version of the changed plugin needs to be incremented, as shown in the logs:
> 
> org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.ui.widgets: Only qualifier changed for (org.eclipse.e4.ui.widgets/1.4.100.v20240731-0633). Expected to have bigger x.y.z than what is available in baseline (1.4.100.v20231201-1637)

Ah alright, I see. This plug-in only contains ImageBasedFrame, so it is not really incremented that often. I increased the version to 1.4.200. Is it okay to do that inside of the PR?

fedejeanne:
> > The version of the changed plugin needs to be incremented, as shown in the logs:
> > org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.ui.widgets: Only qualifier changed for (org.eclipse.e4.ui.widgets/1.4.100.v20240731-0633). Expected to have bigger x.y.z than what is available in baseline (1.4.100.v20231201-1637)
> 
> Ah alright, I see. This plug-in only contains ImageBasedFrame, so it is not really incremented that often. I increased the version to 1.4.200. Is it okay to do that inside of the PR?

I don't see why not. @merks, any thoughts?

HeikoKlare:
Yes, it's okay and even preferred to do that in the same PR. It ist just recommended to have a separate commit for it, so in case the change needs to be reverted, the version bump can remain.

akoch-yatta:
> Yes, it's okay and even preferred to do that in the same PR. It ist just recommended to have a separate commit for it, so in case the change needs to be reverted, the version bump can remain.

Alright. I split it up and made a separate commit for the version bump

HeikoKlare (approved):
The changes are sound and I have successfully tested them on Windows and Linux/GTK.

On Windows, it has the intended effect when using rescaling at runtime and it has no effect without that functionality (as no event is generated then).

On Linux/GTK, there is functionality for generating `ZoomChanged` events upon DPI changes, but I did not find any way to produce the OS-side events (neither by moving between monitors, changing the scale between 100% and 200%, using fractional scaling with different values or changing values during runtime).

On MacOS, no `ZoomChanged` events are used.

PR #2149, opened 2024-08-02
[Sticky Scrolling] Don't scroll if caret offset is not in the text range

### LINKED ISSUE #2127: IllegalArgumentException at StickyScrollingControl.java:519
Just found multiple occurrences of this error in the log. I assume the code in asyncExec should double check the constraints once again, because it is not guaranteed that it will be executed as the immediate next UI task.

```
java.lang.IllegalArgumentException: Index out of bounds
	at org.eclipse.swt.SWT.error(SWT.java:4903)
	at org.eclipse.swt.custom.StyledText.getLineAtOffset(StyledText.java:3585)
	at org.eclipse.ui.internal.texteditor.stickyscroll.StickyScrollingControl$StickyScollingCaretListener.lambda$0(StickyScrollingControl.java:519)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5040)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4520)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:639)
```

Code is from #1719. 
@Christopher-Hermann, could you please check?

Christopher-Hermann:
The scrolling based on caret movement is done in a asynchronous task to keep the impact on the UI thread as low as possible. If the source viewer document is changed between caret movement and execution of the scrolling, it can happen that the offset is not in the text range any longer. In this case, scrolling is not done.

Fixes #2127

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
@@ -508,15 +508,15 @@ public class StickyScrollingControl {
 
 		@Override
 		public void caretMoved(CaretEvent event) {
-			int offsetEndPosition= sourceViewer.getTextWidget().getText().length();
+			int offsetEndPosition= sourceViewer.getTextWidget().getCharCount();
 			if (event.caretOffset == 0 || event.caretOffset == offsetEndPosition) {
 				return;
 			}
 			Display.getDefault().asyncExec(() -> {
-				if (!enableCaretListener) {
+				StyledText textWidget= sourceViewer.getTextWidget();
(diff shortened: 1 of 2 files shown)


laeubi:
> The scrolling based on caret movement is done in a asynchronous task to save performance. 

This is wrong. It is executed asynchronously but this does not save any performance (actually it will decrease performance by a tiny little bit).

Christopher-Hermann:
> > The scrolling based on caret movement is done in a asynchronous task to save performance.
> 
> This is wrong. It is executed asynchronously but this does not save any performance (actually it will decrease performance by a tiny little bit).

Ah yes, you are right. It's saving performance on the UI thread by shifting it to another thread. I will change the commit message.

laeubi:
I would simply omit the performance aspect. In general Display#asyncExec *always* needs to recheck the state as in the meanwhile anything can happen (including disposal of some UI components).

Issue #2114, opened 2024-07-19
UI refresh: a modern light and dark theme


thomasritter:
### Known issues

First version has been submitted. Here the list of already known issues we will work on.

- [x] Headers do not react to hovering events anymore. Add hover indicator. ( Dark, Light -> Win, Mac) 
--> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2130 
- [x]  “Git Staging” view does not have a grey background. This makes the grey buttons standout. Background should be grey. (Light -> Win, Mac) 
--> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2168
- [x]  "Types" view does not have a grey background 
--> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2153. 
--> Revert in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2181
- [x]  Editor line number strip should have the same color as the editor background. (Dark -> Win, Mac) 
--> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2744
- [x]  Editor line number strip should have the same color as the editor background. (Light -> Win, Mac) 
--> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2314
- [x]  Provide same themes for Linux (Dark, Light) 
--> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2137
- [x] [Issue in breadcrumb](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114#issuecomment-2241561690) 
--> https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1577
- [x] Missing alternate row colors in tables on macOS https://github.com/eclipse-platform/eclipse.platform.ui/issues/2171)
- [x]  Missing alternate row colors in trees on macOS https://github.com/eclipse-platform/eclipse.platform.ui/pull/2404
- [x] [Editor Tabs looks greyed out](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2182) --> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2294
- [x] Multiple Tabs have blue highlight --> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2286
- [x] [Labels in Progress View have wrong background](https://github.com/eclipse- noplatform/eclipse.platform.ui/issues/2313) Not reproducible any more.
[...]

PR #2153, opened 2024-08-03
Fix for background color of some views like Types, Console etc

mvm-sap:
Implementing fix for background color change to some of the views like Console View, Types View etc.
Background color of these views was white, with this change it has been changed to grey(#f8f8f8) to harmonize with other views.
Refer  https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114#issue-2418424128 -  "Types" view does not have a grey background. Fix has been provided in Mac, Win and Linux.

image

image


BeckerWdf:
tested on macOS. Looks good.

PR #2160, opened 2024-08-05
Find/replace overlay: avoid disposed exception on closing

HeikoKlare:
When closing a find/replace overlay while a history drop-down menu is open, an exception can occur because the state of a disposed widget is updated. This change ensures that only when the widget is not disposed, its state is changed.


HeikoKlare:
Failing tests are unrelated and already documented: #294 and #926

PR #2163, opened 2024-08-05
Find/replace: allow changing whole word option when input is empty #2162

### LINKED ISSUE #2162: Find/replace overlay: enable "whole word" option when input field is empty

## Suggestion

* open a text editor
* open the find/replace overlay via CTRL+F / CMD+F
* enter anything into the find input field and clear it again

The "whole word" option is disabled. It can neither be activated nor activated. At least a single letter has to be entered to change this option.

It would be good to allow changing this option when nothing is entered in the input field to only move to whole matching words if desired. There does not seem to be a reason for disabling the option when the input field is empty. In particular, the option can be activated or deactivated after opening the overlay, even though the find input field is still empty. It will only be disabled when removing the contents of the find input field.

image

## Acknowledgements

This proposal is based on a finding by Celine Williams and Markus Wambach in a test session performed at SAP. Thank you for testing and reporting!

## Community

HeikoKlare:
The find/replace dialog and overlay do not allow to change the "whole word" search option when the input field is empty. There is, however, no reason for disabling the option when the input field is empty. With this change, the "whole word" option can also be activated and deactivated when the find input is empty. In addition, the enablement behavior of the option in the find/replace overlay is unified.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2162

### How it look

Without this change, the "whole word" option cannot be enabled/disabled when the find input field is empty. Here is how it looks for the dialog (overlay affected in the same way):
image

And here is how it looks with the change for the dialog:
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -128,7 +128,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	 * @return <code>true</code> if the given string is a word
 	 */
 	private static boolean isWord(String str) {
-		return str != null && !str.isEmpty() && str.chars().allMatch(Character::isJavaIdentifierPart);
+		return str != null && str.chars().allMatch(Character::isJavaIdentifierPart);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


HeikoKlare:
Does anyone see a reason why one should not be able to configure the "whole word" search option when the find input field is empty? If not, I would like to merge this to allow a user to configure their search without requiring them to first input some search string.

HeikoKlare:
If no one objects, I would like to merge this by the end of today.

Issue #1993, opened 2024-06-25
Find/Replace overlay:  "String xyz not found" message


jukzi:
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/661656a7-ec5a-4897-8d00-cbe6b906f464)

HeikoKlare:
@jukzi can you briefly explain what exactly is wrong here, so in particular what you would expect in your scenario?

jukzi:
minor: in the bottom left of the eclipse window the progress shortly says "String xyz not found"
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/d125a28c-04c3-4474-8141-d5648261ba6d)

1. The word "String" is something a user does not need to understand. Better: "'xyz' not found"
2. "xyz NOT FOUND" is a status not a progress, that should not disappear within a tens of a second


_Originally posted by @jukzi in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1990#issuecomment-2186273627_

HeikoKlare:
Thanks for the additional information!

> * The word "String" is something a user does not need to understand. Better: "'xyz' not found"

I agree. This has just been taken over from the existing dialog, so if we change that, it should be changed at both places.
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/755472/ee85cf6e-2d29-4e09-addf-5e4d81cd78c0)

Maybe something like "No search results for: xyz"?

> * "xyz NOT FOUND" is a status not a progress, that should not disappear within a tens of a second

I see, that is definitely a bug.

PR #2164, opened 2024-08-06
Find/replace: improve message when no search results were found #1993

HeikoKlare:
When a search in an editor does not yield any results, a status message with the content "String xyz is not found" with "xyz" being the search string. It is currently hard to understand which part of the message the actual search string is and maybe users do not even know the term "string".

With this change, the message for a search of "xyz" not yielding any results is changed to: "No search results for: xyz" This makes clear what the search string is without requiring to quote it.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1993

### Screenshots

Status message before the change:
image

Status message after the change:
image


HeikoKlare:
Opinions on the proposed change are appreciated.

The change effects all users of the `FindReplaceLogic`, i.e., both the existing find/replace dialog and the new find/replace overlay UI.

HeikoKlare:
If no one objects, I would like to merge this today, so that we have this improvement in the next release.

HeikoKlare:

PR #2167, opened 2024-08-06
EclipseCommandURIHandler: Fix decoding of query parameters.

### LINKED ISSUE #2166: eclipse+command:// URI handler incorrectly decodes query parameters
There is a bug in `org.eclipse.e4.ui.internal.workbench.swt.handlers.EclipseCommandURIHandler` in how URIs are parsed:

Consider the following example:

`eclipse+command://foo.bar?action=doSomethinWithUrl&url=https%3A%2F%2Feclipse.org%3Fparam1%3Dhi%26param2%3Dthere`

This should be deconstructed as follows:

```json
"command" : "foo.bar",
"parameters" : {
  "action" : "doSomethinWithUrl",
  "url" : "https://eclipse.org?param1=hi&param2=there"
}
```

What is actually deconstructed is:

```json
"command" : "foo.bar",
"parameters" : {
  "action" : "doSomethinWithUrl",
  "url" : "https://eclipse.org?param1=hi",
  "param2" : "there"
}
```

The single `url` parameter broke and parts of it were put into a separate parameter.

Reason:
`URI#getQuery()` is get called which URL-decodes the full thing and destroys semantics. Then, the result is split on `&` and `=`.

Correct would be to call
`URI#getRawQuery()`, then split on `&` and `=` and only then do the URL-decoding of the individual parts.

sratz:
`URI#getQuery()` URL-decodes the full query string, destroying semantics. The, the result is split on `&` and `=`.

Instead, we first need to use `URI#getRawQuery(),` then split on `&` and `=`, and only then do the URL-decoding of the individual parts.

Fixes: #2166

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.workbench.swt/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.workbench.swt;singleton:=true
-Bundle-Version: 0.17.400.qualifier
+Bundle-Version: 0.17.500.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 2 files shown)


mickaelistria:
Thank you!

PR #2170, opened 2024-08-07
Limit check for macOS major version to 14 for flipped splash image workaround

Phillipus:
- As background image flipping is fixed in macOS 15 limit the version check to version 14 only


Phillipus:
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/772#issuecomment-2272902189

cc @sratz

Phillipus:
If Apple's fix is in beta 5 it's likely it will be in the final macOS release. Merging this PR now would ensure it getting into Eclipse 4.33 ready for the macOS 15 release in late September/early October. If Apple's fix turns out not to make it into macOS 15 then it could be reverted for 4.34.

sratz:
> If Apple's fix is in beta 5 it's likely it will be in the final macOS release. Merging this PR now would ensure it getting into Eclipse 4.33 ready for the macOS 15 release in late September/early October.

Agreed, to avoid another round of _'The splash screen is flipped!'_, we should merge this for 2024-09.

> If Apple's fix turns out not to make it into macOS 15 then it could be reverted for 4.34.

Yep, fingers crossed that it will.

sratz (approved):

Phillipus:
> Agreed, to avoid another round of 'The splash screen is flipped!', we should merge this for 2024-09.

Could someone please merge.

PR #2174, opened 2024-08-09
Removing ADT specific changes related to breadcrumbs

mvm-sap:
While addressing an issue related to inconsistent background color for breadcrumbs in java editor, which is reported here - https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114#issuecomment-2241561690. Realized that piece of code was specific to ADT. Hence reverting the code.


BeckerWdf:
can you pls. show before and after screenshots?

BeckerWdf:
So thing changes in Platform and the Eclipse IDE?

mvm-sap:
> So thing changes in Platform and the Eclipse IDE?

Yes

PR #2179, opened 2024-08-11
Bug 242719 add SpinnerFieldEditor

sebthom:
This is a follow-up PR for https://bugs.eclipse.org/bugs/show_bug.cgi?id=242719

Similar to the approach in that issue I created the SpinnerFieldEditor based on the latest code of ScaleFieldEditor.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/SpinnerFieldEditor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/SpinnerFieldEditor.java
@@ -1,0 +1,370 @@
+/*******************************************************************************
+ * Copyright (c) 2003, 2015 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *     Wedia - Joel DRIGO (joel.drigo@wedia-group.com): Bug 470866
+ *     Sebastian Thomschke - Adapted code of ScaleFieldEditor for SpinnerFieldEditor
+ *******************************************************************************/
+package org.eclipse.jface.preference;
+
+import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Spinner;
+
+/**
+ * A field editor for an integer type preference. This class may be used as is,
+ * or subclassed as required.
+ *
+ * @since 3.35
+ */
+public class SpinnerFieldEditor extends FieldEditor {
+
+	/**
+	 * Value that will feed {@link Spinner#setDigits(int)}.
+	 */
+	private int digits = 0;
+
+	/**
+	 * Value that will feed {@link Spinner#setIncrement(int)}.
+	 */
+	private int incrementValue = 1;
+
+	/**
+	 * Value that will feed {@link Spinner#setMaximum(int)}.
+	 */
+	private int maxValue = 10;
+
+	/**
+	 * Value that will feed {@link Spinner#setMinimum(int)}.
+	 */
+	private int minValue = 0;
+
+	/**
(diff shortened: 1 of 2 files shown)


HannesWell:
Thank you for this enhancement.

Can anyone of the committers that is more involved into UI topics review this?
Then we can consider this for submission in the beginning of the next release cycle.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/SpinnerFieldEditor.java
```diff
@@ -0,0 +1,370 @@
...
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
```
Wittmaxi, line 12: Is this file copied from somewhere? The copyright header looks off somehow.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/SpinnerFieldEditor.java
```diff
@@ -0,0 +1,370 @@
...
+/**
+ * A field editor for an integer type preference. This class may be used as is,
+ * or subclassed as required.
+ *
+ * @since 3.35
+ */
+public class SpinnerFieldEditor extends FieldEditor {
+
+	/**
+	 * Value that will feed {@link Spinner#setDigits(int)}.
```
Wittmaxi, line 34: These comments don't help me understand the code. 
```suggestion
	 * amount of digits to display (?) by default (?)
```

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/SpinnerFieldEditor.java
```diff
@@ -0,0 +1,370 @@
...
+	/**
+	 * Creates a spinner field editor with particular spinner values.
+	 *
+	 * @param name      the name of the preference this field editor works on
+	 * @param labelText the label text of the field editor
+	 * @param parent    the parent of the field editor's control
+	 * @param min       the value used for {@link Spinner#setMinimum(int)}.
+	 * @param max       the value used for {@link Spinner#setMaximum(int)}.
+	 */
+	public SpinnerFieldEditor(final String name, final String labelText, final Composite parent, final int min,
```
Wittmaxi, line 94: This feels like an oddly specific constructor. Maybe consider providing a builder here?

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/SpinnerFieldEditor.java
```diff
@@ -0,0 +1,370 @@
...
+	 * Returns the value that will be used for {@link Spinner#setIncrement(int)}.
+	 *
+	 * @return the value (default 1).
+	 * @see Spinner#setIncrement(int)
+	 */
+	public int getIncrement() {
+		return incrementValue;
+	}
+
+	/**
```
Wittmaxi, line 161: A lot of these javadoc comments seem redundant. In particular, I am talking about comments referring to the setter of a getter.

```suggestion
	/** {@return the maximum value this spinner may take on (10 by default) }
```

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/SpinnerFieldEditor.java
```diff
@@ -0,0 +1,370 @@
...
+			spinner.addSelectionListener(widgetSelectedAdapter(event -> valueChanged()));
+			spinner.addDisposeListener(event -> spinner = null);
+		} else {
+			checkParent(spinner, parent);
+		}
+		return spinner;
+	}
+
+	/**
+	 * Set the value to be used for Spinner.setDigits(int) and update the spinner.
```
Wittmaxi, line 237:
```suggestion
	 * Sets the number of decimal places used by the receiver.
	 * @see Spinner#setDigits(int)
```

Wittmaxi (changes requested):
Thank you for your contribution!
Since adding this PR means increasing API, I think it is extra important to make sure the comments for all methods are speaking.

In particular, many of the javadoc-comments refer to another method instead of "just telling me what the method does".

The code looks sound, so do the test.

Is there any use for this componet? It would be great to see this component as a real component used in some follow-up-down-the-line-PR

Issue #2173, opened 2024-08-08
New light theme causes UI inconsistencies in Tabbed Properties View


Phillipus:
Recent changes to the light theme in the latest I-builds have set the background color for controls to a light grey (#f8f8f8). This [comment ](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114#issue-2418424128) suggests this is intended to create an overall darker background for a View. That's all very well if the View contains only one control, such as a Tree (Package Explorer), but if you have a View that contains many different controls you get an inconsistency in the controls' background colors. This is apparent in a [Tabbed Properties View](https://www.eclipse.org/articles/Article-Tabbed-Properties/tabbed_properties_view.html).

A Tabbed Properties View relies on [FormToolkit](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormToolkit.java) to create the controls. With the previous light theme the background color for most controls was white. With the latest changes some controls background colors are grey and some are white. This leads to an inconsistency in the UI and suggests that some controls are disabled. For example, a `Text` control is white (correct) but a `StyledText` control is grey and looks like it is disabled. Same for `Button` with style `SWT.TOGGLE | SWT.FLAT`.

The changes to the light theme have implemented the rule that most SWT controls that are in a View have their background color set to "#f8f8f8". This is not always desired.

Phillipus:
A `Table` in a View is the same grey color whether enabled or disabled so it's difficult to distinguish its enabled state (especially if the Table is empty). This is the case in any View, not just a Tabbed Properties View.

So a general point here is users will now not easily be able to distinguish enabled/disabled controls in Views.

iloveeclipse:
@Phillipus : could yuo please update description with screenshots before/after?

@mvm-sap, @BeckerWdf: could you please check? Probably related to https://github.com/eclipse-platform/eclipse.platform.ui/pull/2153 changes.

Phillipus:
Here's some before and after screenshots (taken on Windows 11)


### Text, StyledText and Button

1. The following screenshot is a Tabbed Property View containing a single-line `Text`, a multi-line `StyledText` control and a `Button` (with style `SWT.TOGGLE | SWT.FLAT`) _before_ the recent theme change. Backgrounds of all controls are white, controls are "enabled". The `Text` control is included for reference.

[...]

PR #2181, opened 2024-08-12
Reverting the bg color for StyledText, Button, Table controls

mvm-sap:
Reverting the changes done to bg colors of controls like StyledText, Button, Table in views.
Selected tab in TabbedPropertyList had grey background in view, even that has been changed to white with this change.
Issue  https://github.com/eclipse-platform/eclipse.platform.ui/issues/2173 has been addressed with this change.
Please again test and provide feedback if there are still some issues.

Below are the screenshots

Before:
image

After:
image

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
+++ b/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
@@ -154,8 +154,7 @@ CTabFolder Canvas {
 .View FormText,
 .View Hyperlink,
 .View Canvas,
-.View FigureCanvas,
-.View Table
+.View FigureCanvas
 {
 	background-color: #f8f8f8;
 }
@@ -168,13 +167,16 @@ CTabFolder Canvas {
 	background-color:#f8f8f8;
 }
 
-.View PageBook StyledText,
-.View Button{
+.View Button[style~='SWT.CHECK']{
 	background-color: inherit;
 }
 
-.View Section ToolItem{
-	background-color: #EAEAEA;
+.View Toolbar ToolItem{
+	background-color: #eaeaea;
+}
+
+.View TabbedPropertyList{
+	swt-tabBackground-color: #ffffff;
 }
 
 .View Composite PrependingAsteriskFilteredTree,
--- a/bundles/org.eclipse.ui.themes/css/e4_default_mac.css
+++ b/bundles/org.eclipse.ui.themes/css/e4_default_mac.css
@@ -124,8 +124,7 @@ CTabFolder Canvas {
 .View FormText,
 .View Hyperlink,
 .View Canvas,
-.View FigureCanvas,
-.View Table
+.View FigureCanvas
 {
 	background-color: #f8f8f8;
 }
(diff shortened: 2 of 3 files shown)


BeckerWdf:
@Phillipus: Can you pls. review and test?

Phillipus:
Thanks for the PR. I tested this with our RCP app and can confirm that the `StyledText`, `Button` controls in views and the selected tab in `TabbedPropertyList` are white in the light theme and the `Table` control has alternate row colors on Mac.

One slight issue on Mac is the background in a `Label` in a `Group` control:

Screenshot 2024-08-12 at 15 24 16

However, the same problem occurs in the current light theme on Mac.

BeckerWdf:
> However, the same problem occurs in the current light theme on Mac.

So this has to be handled separately..

Phillipus:
> So this has to be handled separately..

The following on Mac seems to fix it:

```
.View Group Label {
    background-color: unset;
}
```

tomaswolf:
Just noticed: the gray unstaged/staged changes trees in that abapGit view (probably also in the eGit staging view?) also look like they were disabled. (Perhaps they are, since there are no changes...) Perhaps trees inside forms also should be styled with a white background. How does e.g. the PDE manifest editor look with that new styling?

Phillipus:
> How does e.g. the PDE manifest editor look with that new styling?

If it's an EditPart it's not affected. This change affects controls in ViewParts. But if those same form-based controls are in a ViewPart they're grey.

Trees in ViewParts are grey so if you depend on that to show disabled state of a Tree you're out of luck.

mvm-sap:
> > So this has to be handled separately..
> 
> The following on Mac seems to fix it:
> 
> ```
> .View Group Label {
>     background-color: unset;
> }
> ```

If we unset, it will affect other views, find below the screenshot. This is in windows, might aswell affect Mac
image

I feel its better to set it this way:
```
.View Group Label{
background-color: inherit;
}
```

What do you think?

BeckerWdf:
> Just noticed: the gray unstaged/staged changes trees in that abapGit view (probably also in the eGit staging view?) also look like they were disabled. (Perhaps they are, since there are no changes...) Perhaps trees inside forms also should be styled with a white background. How does e.g. the PDE manifest editor look with that new styling?

With this change or before this change?

mvm-sap:
> Just noticed: the gray unstaged/staged changes trees in that abapGit view (probably also in the eGit staging view?) also look like they were disabled. (Perhaps they are, since there are no changes...) Perhaps trees inside forms also should be styled with a white background. How does e.g. the PDE manifest editor look with that new styling?

Yes, Staged and Unstaged will have grey backgrounds, as we have set the bg for a tree as grey in views.

BeckerWdf:
Can we merge this change?

BeckerWdf:
> and the `Table` control has alternate row colors on Mac.

I don't see that on my Mac. Where do you see that? Can you show a screenshot?

Phillipus:
> I feel its better to set it this way:

```.View Group Label{
background-color: inherit;
}
```

> What do you think?

That doesn't work.  As I said it's Mac only so only belongs in the Mac CSS file .

Phillipus:
> > and the `Table` control has alternate row colors on Mac.
> 
> I don't see that on my Mac. Where do you see that? Can you show a screenshot?

This PR reverts the grey table so the alternate row colors is expected and desired (as it used to be).

BeckerWdf:
If nobody objects I will merge this today.

BeckerWdf:
> Thanks for the PR. I tested this with our RCP app and can confirm that the `StyledText`, `Button` controls in views and the selected tab in `TabbedPropertyList` are white in the light theme and the `Table` control has alternate row colors on Mac.
> 
> One slight issue on Mac is the background in a `Label` in a `Group` control:
> 
> Screenshot 2024-08-12 at 15 24 16
> However, the same problem occurs in the current light theme on Mac.

So this existed already even before @mvm-sap's changes? If yes. Pls. open a separate issue. And if you already have a solution proposal pls. also send a PR for it.

PR #2185, opened 2024-08-12
fix: setting minimum > 100 via ScaleFieldEditor's constructor arg is ignored

sebthom:
While working on https://github.com/eclipse-platform/eclipse.platform.ui/pull/2179 I found an unexpected edge case in `ScaleFieldEditor` that is handled wrong.

When a value for the `minimum` argument of the ScaleFieldEditor constructor is set that is higher than the Scale`s default value for maximum (100 on Windows), then it is set to the maximum value instead.

Example:
```java
editor = new ScaleFieldEditor("prefKey", "someLabel", parent, 
              200 /**minimum*/, 400 /**maximum*/, 1, 1);

editor.getScaleControl().getMinimum(); // returns 0 instead of 200
```

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScaleFieldEditor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScaleFieldEditor.java
@@ -317,6 +317,10 @@ public class ScaleFieldEditor extends FieldEditor {
 		if (scale != null && !scale.isDisposed()) {
 			scale.setMinimum(getMinimum());
 			scale.setMaximum(getMaximum());
+			// Reapplying the minimum to ensure that the scale's value is correctly
+			// adjusted in scenarios where the new minimum exceeds the previous maximum.
+			scale.setMinimum(getMinimum());
+
 			scale.setIncrement(getIncrement());
 			scale.setPageIncrement(getPageIncrement());
 		}
--- a/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/preferences/AllPrefsTests.java
+++ b/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/preferences/AllPrefsTests.java
@@ -18,7 +18,12 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 
 @RunWith(Suite.class)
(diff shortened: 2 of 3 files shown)


sebthom:
@HeikoKlare @HannesWell could one of you please have a look at this PR. Would be nice if it could make it into 2024-09.

HannesWell (approved):
To me this change looks save and sound. I also thought about setting the max fist, but then the same problem could arise for different scenarios.
@HeikoKlare do you agree?

I have just updated this PR to add the new test class to the corresponding test suite so that it is picked up during test execution.

If yes, I think we can have this for RC1.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScaleFieldEditor.java
```diff
@@ -317,6 +317,10 @@ private void updateScale() {
 		if (scale != null && !scale.isDisposed()) {
 			scale.setMinimum(getMinimum());
 			scale.setMaximum(getMaximum());
+			// Reapplying the minimum to ensure that the scale's value is correctly
```
merks, line 320: It’s better to a this comment because otherwise is looks like a redundant mistake.

sebthom: I don't understand what you mean.

HannesWell: I think it is just a confirmation that this comment is important. :)

HannesWell (approved):
Tests are now executed and succeed.
Let's have this for RC1.

Thank you @sebthom for fixing this important corner case!

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScaleFieldEditor.java
merks: Yes, sorry for the typo.  Everything is so small on an iphone....

HeikoKlare:
Just for the record: looks also fine from my side. Thank you for the fix @sebthom and thank you, Hannes, for merging.

One nitpick: it might be a bit cleaner in such a situation to explicltly check whether you are in one of the corner cases (new min > old max _or_ new max < old min) and then execute the statements in the proper order. That would look less like a mistake that requires an additional comment but would me rather self explaining.

PR #2186, opened 2024-08-13
Moving TabFolderLayout into SWT.

### LINKED ISSUE #1317: Move TabFolderLayout into SWT and set it as a default in TabFolder
Today I debugged some issues with `TabFolder` and noticed that it at least behaves strange if tabs are added dynamically, after investigate a little bit more I tried to fix it with an own Layout as usual layout mangers (like `FillLayout`) get sometimes confused as `TabFolder` has more childs than are actually displayed.

Short after that I noticed that there are already some instances of `TabFolderLayout` in platform:
grafik

They all seem to be copies of each other (more or less) except the license header and when they where added, I therefore wonder if we should simply add that layout to SWT and even set it by default to `TabFolder`.

deepika-u:
Moving TabFolderLayout into SWT.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1317

This pr needs to be merged after https://github.com/eclipse-platform/eclipse.platform.swt/pull/1402


BeckerWdf:
> Fixes #1317

I don't see the relation to that issue. Can you pls. explain?

laeubi:
> > Fixes #1317
> 
> I don't see the relation to that issue. Can you pls. explain?

I fixed the link, it is meant to be 
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1317

merks:
We can’t just delete it can we?  It’s API.

laeubi:
Its internal ... but I suspect there might be (internal) references that needs updates?

merks:
Sorry I missed the internal in the package name.

deepika-u:
> Its internal ... but I suspect there might be (internal) references that needs updates?

Definitely there would be some references as per my expectation. But how can i know where all it is referring? and what all files needs the respective updates. Could someone point me to a file which lists those failures or something?

laeubi:
The best would be to first finish the SWT one so the new one is available.

Then you can setup a [platform-sdk-ide](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment) and use the usual "Find References" search, and replace all usages with the SWT one.

akurtakov:
Please make sure that the bundles require new enough swt that has that api.

deepika-u:
@akurtakov : Can you please check now? I have updated bundle versions as well now.

akurtakov:
Jenkins fails due to OOM. I'm merging this one despite it.

Issue #2182, opened 2024-08-12
Editor Tabs looks greyed out


jukzi:
in the past month the File names in the editor  changed from black to grey or the used font got thinner then a pixel. Was that intentional? looks less readable
now (I20240811-1800):
![image](https://github.com/user-attachments/assets/dd4f90c3-69a2-49f1-919a-0d9dac95fe32)
before (I20240710-1800):
![image](https://github.com/user-attachments/assets/1d1845fc-f14f-439a-87b5-57622167e77a)

Sadly the screenshots are not so good to see the difference due to subpixel rendering. But it was basically the first thing that fleshed my eyes after having not used eclipse for a month.

jukzi:
![image](https://github.com/user-attachments/assets/5f7f6659-9063-4a98-8aad-85d3ece15b43)

jukzi:
old black was 0,0,0. new color is 59,59,59
![image](https://github.com/user-attachments/assets/7fe69239-9e2e-4972-b831-483138fd1e20)

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114

Phillipus:
Agreed, the text color looks a bit pale now.

BeckerWdf:
@mvm-sap: Was this change by intention? If yes: Why. If not: Can we get the old color with the better contrast back?

mvm-sap:
> @mvm-sap: Was this change by intention? If yes: Why. If not: Can we get the old color with the better contrast back?

Yes, we followed VS Code tab font color codes. 
In editor
selected tab : #3b3b3b
unselected tab : #6a6a6a
In Views 
Both Selected and Unselected tabs : #3b3b3b

If this does not look good, may be I can try to change

BeckerWdf:
> > @mvm-sap: Was this change by intention? If yes: Why. If not: Can we get the old color with the better contrast back?
> 
> Yes, we followed VS Code tab font color codes. In editor selected tab : #3b3b3b unselected tab : #6a6a6a In Views Both Selected and Unselected tabs : #3b3b3b
> 
> If this does not look good, may be I can try to change

I really can agree that the contrast of text and background was better in the past. And good contrast is a good think.
Was the color different for selected and unselect tabs before your change?

mvm-sap:
> > > @mvm-sap: Was this change by intention? If yes: Why. If not: Can we get the old color with the better contrast back?
> > 
[...]

PR #2187, opened 2024-08-13
Change to font colors in selected/unselected tabs in views and editor

mvm-sap:
Change to the existing light theme https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114 also involved the change to the font colors of selected and unselected tabs in views and editors. This led to an issue from a user https://github.com/eclipse-platform/eclipse.platform.ui/issues/2182.

Selected tab in Views and Editor now has #00000 color code 
Unselected tabs in Views and Editor now have #3b3b3b color code 
This change has been done to improve the contrast between font color and background color of tabs.
Please let us know your views.

Screenshots can be found here.

Before:
image

After: 
image

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
+++ b/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
@@ -91,14 +91,6 @@ ColorDefinition#org-eclipse-ui-workbench-ACTIVE_NOFOCUS_TAB_TEXT_COLOR {
 	color: #3b3b3b;
 }
 
-ColorDefinition#org-eclipse-ui-workbench-ACTIVE_TAB_TEXT_COLOR {
-   color: #3b3b3b;
-}
-
-ColorDefinition#org-eclipse-ui-workbench-INACTIVE_TAB_TEXT_COLOR {
-   color: #3b3b3b;
-}
-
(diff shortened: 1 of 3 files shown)


BeckerWdf:
@jukzi: What's your opinion?

jukzi:
Looks better but still **feels** to pale at the inactive tabs:
image
Maybe one can get used to it. But by now i would still prefer the old contrast.  I guess the subjective feeling is also because eclipse IDE still shows high contrast icons in front of the name which makes the low contrast text have less contrast then the icons. But i don't need to focus on the icons. The text is the important thing.

"VS Code" and other IDEs not have those problem because they have general other layout without icons and a low contrast content in the problem view:
image

BeckerWdf:
> I guess the subjective feeling is also because eclipse IDE still shows high contrast icons in front of the name which makes the low contrast text have less contrast then the icons.

That's an interesting idea.
Btw: You can turn off the icons on view tabs in the preference. Then this issue no longer exists.

jukzi:
issue still exists because it does not disable the high contrast in the editor icons
image

BeckerWdf:
Replaced with https://github.com/eclipse-platform/eclipse.platform.ui/pull/2294

PR #2188, opened 2024-08-13
Find/replace overlay: improve first-time notification layout

HeikoKlare:
The notification pop-up on first usage of the find/replace overlay has a white bar at the bottom. This is because the contents widgets does not fill the whole composite of the pop-up, which, in turn, has a minimum size.

This change <s>adds a line break to the notification text to improve readability and</s> makes the contents use the complete vertical space of the pop-up.

The issue was reported here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1192#issuecomment-2285830201

## How it looks

Before:
image

After:
image

@jukzi thank you for reporting the issue. What do you think about the fix? I see some opportunities for improving the design of the notification pop-up in general (such as some padding for all the text within the popup), but that's independent from this use case.


jukzi:
looks better, thanks. implementation wise i think that the linebreak is a dirty hack and should not be needed at all.

HeikoKlare:
The line break is not necessary for fixing the layout. I can also remove it. Then it looks like this:
image

jukzi:
looks good.

HeikoKlare:
Alright, I've reverted the NLS change and only adapted the layout.

jukzi (approved):

HeikoKlare:
Failing test is unrelated and already documented: #926

PR #2205, opened 2024-08-18
Find/replace: fix search operation during replace in RegEx mode #2203

### LINKED ISSUE #2203: Find/replace overlay: repeated ENTER has no effect in regex mode
In NON-regex (standard) mode one can simply add a phrase in the search and replace fields and change one occurrence after the other by pressing ENTER repeatedly. But this does not work for me in regex mode. Here hitting ENTER has no effect and I must use the "arrow down" button to jump to the next occurrence each time. 

I think there is no need to describe the bug in more detail because @HeikoKlare could already reproduce it and confirmed this being a bug. 

## Steps to reproduce
Use find/replace overlay as described above in regex mode and hit ENTER key repeatedly which has no effect.

* Win11
* Eclipse SDK
Version: 2024-09 (4.33)
Build id: I20240731-1800

## Community

HeikoKlare:
When performing a search operation in the FindReplaceLogic, it currently always validates compatibility of the search options. In particular, having incremental search and regex search enabled at the same time will result in search operations not being performed. This results in problems when performing replace operations that also include search operations, as the search operation will not be executed then.

With this change, the search operations performed while doing a replace will not validate the search operations but always perform the requested search instead.

ℹ️  I consider this an intermediate solution for the upcoming 2024-09 release. It is a rather local change only affecting the replace behavior when incremental and regex search mode are enabled (as that's the only case in which the existing guard applied). The root cause is, in my opinion, a design flaw, as the FindReplaceLogic is (1) used to reflect all current search option, (2) at the same time disallows specific operations when options are conflicting while (3) consumers rely on applying these operations even though options are conflicting. This results in consumers switching options before performing some operations (such as disabling incremental mode when performing an explicit search). Improving this will probably be a bit more invasive than the current change and is only about design, not about functionality. Thus, from a risk assessment perspective, it should be done after 2024-09. I will do this in follow-up work.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2203

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -326,9 +326,13 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 
 	@Override
 	public boolean performSearch(String findString) {
+		return performSearch(findString, true);
+	}
+
+	private boolean performSearch(String findString, boolean validateSearchOptions) {
 		resetStatus();
 
-		if (isActive(SearchOptions.INCREMENTAL) && !isIncrementalSearchAvailable()) {
+		if (validateSearchOptions && (isActive(SearchOptions.INCREMENTAL) && !isIncrementalSearchAvailable())) {
 			return false; // Do nothing if search options are not compatible
 		}
 		boolean somethingFound = false;
@@ -543,7 +547,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	public boolean performReplaceAndFind(String findString, String replaceString) {
 		resetStatus();
 		if (performSelectAndReplace(findString, replaceString)) {
-			performSearch(findString);
+			performSearch(findString, false);
 			return true;
 		}
 		return false;
@@ -553,7 +557,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	public boolean performSelectAndReplace(String findString, String replaceString) {
 		resetStatus();
 		if (!isFindStringSelected(findString)) {
-			performSearch(findString);
+			performSearch(findString, false);
 		}
 		if (getStatus().wasSuccessful()) {
 			return replaceSelection(replaceString);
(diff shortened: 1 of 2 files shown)


Wittmaxi:
IMHO, A better solution (which respect architecture) is to temporarily disable Incremental search in the "performFindAndNext".
We recently moved to Incremental search ALWAYS performing incremental search even if performSearch was called (earlier, performSearch would perform a real search even if Incremental search was on). This left the bug that "performFindAndNext" didn't disable/enable the Incremental search option when searching next, even though "findAndNext" refers to a REAL find and not to an Incremental search.

A few months ago we had performIncrementalSearch vs performSearch, where performIncrementalSearch would do nothing if Incremental wasn't enabled and performSearch would never search incrementally, regardless of parametrization

I don't quite yet see how this is an architectural problem. The FindReplaceLogic has a state which makes it's operation behave the same way all the time in that same state. When doing "findAndNext" we expect a different behavior from when performing an incremental search (we want to actually jump to the next occurence, like in the FindReplaceDialog, not just "highlight it" without resetting a base location).

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -553,7 +557,7 @@ public boolean performReplaceAndFind(String findString, String replaceString) {
 	public boolean performSelectAndReplace(String findString, String replaceString) {
 		resetStatus();
 		if (!isFindStringSelected(findString)) {
-			performSearch(findString);
+			performSearch(findString, false);
```
Wittmaxi, line 560:
```suggestion   
disableOption(SearchOptions.INCREMENTAL);
			performSearch(findString, false);
			enableOption(SearchOptions.INCREMENTAL);
```

HeikoKlare: The design flaw is as follows: the logic know about search options (defined in an enum) and search operation (defined as methods). Now "incremental" is basically both: an option and an operation. You can activate the "incremental" option to say that an incremental base position for searching shall be used, and you can perform an incremental search, i.e., one that does not move to the next search result but update the current one. The latter are actually two operations: one performed by updating the search input string and one by performing an explicit search option (via pressing enter or a button) with different behaviors. These are two different things that are currently mixed up, as you can see at the fact that the incremental option needs to be changed just to make an operation behave differently and not because the user actually wanted to specify some different option.

I have already prepared a more holistic improvement, which completely removes all these necessities by hiding the responsibility for distinguishing these two cases in the logic, so no consumer will need to care about the different operation modes. But as mentioned, that's too invasive for the upcoming release as we are already after M3, so I will provide that improvement for 2024-12 M1.

I am not in favor of the proposed fix of temporarily disabling the incremental search option in the logic. It is more invasive, as it performs additional modification of the incremental base position. I would like to keep the impact as small as possible to be confident that we do not introduce any regressions at this point in time. Only reasons for me to integrate the proposed fix would be an additional benefit in terms of fixing some unintended behavior with the current fix that I have not found so far.

HeikoKlare:
Failing test is unrelated and already documented: #926

PR #2207, opened 2024-08-19
Load marker content generator details from marker content generator extensions (fix for #2193)

### LINKED ISSUE #2193: Adding new MarkerField via markerSupport extension point to problems view seems impossible
I tried adding a new `MarkerField` to the problems view using the _markerSupport_ extension point. That seems not to be possible with the latest [org.eclipse.ui.ide](https://github.com/eclipse-platform/eclipse.platform.ui/tree/master/bundles/org.eclipse.ui.ide) bundle. I was not able to select the new `MarkerField` in the problems (or markers) view's "Configure Columns..." dialog.

Using the _markerSupport_ extension point I added a new _markerContentGenerator_ and a _markerContentGeneratorExtension_ for the problem view's marker content generator (`id=org.eclipse.ui.ide.problemsGenerator`). While debugging the code, I found out that my marker content generator was initialized and used the new `MarkerField`, but the [`ContentGeneratorDescriptor`](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java) did not pick up the new `MarkerField`.

I guess, the method [`ContentGeneratorDescriptor#getAllFields()`](https://github.com/eclipse-platform/eclipse.platform.ui/blob/8ad21aa2384fb09ae4146c4905a2929d474bacff/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java#L107) is not correct. I think, the `ContentGeneratorDescriptor` has to collect all `MarkerField`s provided by the corresponding marker content generator as well as `MarkerField`s provided by the marker content generator extensions. Instead, `ContentGeneratorDescriptor#getAllFields()` ignores the registered extensions. It seems, the only place, where marker content generator extensions are considered, is the method [`ContentGeneratorDescriptor#getFilterReferences()`](https://github.com/eclipse-platform/eclipse.platform.ui/blob/8ad21aa2384fb09ae4146c4905a2929d474bacff/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java#L125).

Did I misunderstand something or do we have to handle marker content generator extensions in most of the `ContentGeneratorDescriptor`'s methods?

## Details

My marker support extension looked like that:

```java
   <extension
         point="org.eclipse.ui.ide.markerSupport">
      <markerField
            class="org.eclipse.lsp4e.ui.CustomMarkerField"
            id="org.eclipse.lsp4e.markerField.custom"
            name="Custom">
      </markerField>
      <markerContentGeneratorExtension
            generatorId="org.eclipse.ui.ide.problemsGenerator"
            id="org.eclipse.lsp4e.markerContentGenerator">
      </markerContentGeneratorExtension>
      <markerContentGenerator
            id="org.eclipse.lsp4e.markerContentGenerator"
            name="LSP4E MarkerContentGenerator">
         <markerFieldReference
               id="org.eclipse.lsp4e.markerField.custom">
         </markerFieldReference>
         <markerTypeReference
               id="diagnostic">
         </markerTypeReference>
      </markerContentGenerator>
   </extension>
```

## Community

travkin79:
This patch fixes issue #2193. The implementation and tests ensure loading of marker content generator extensions, including new marker fields,  marker types, and groups provided via marker content generator extensions.

This way, it is now possible to extends, e.g. the Problems View and the Markers View with new columns. For example, you could add a column with a URL to a vulnerability issue's detailed description (e.g. [CWE-200](https://cwe.mitre.org/data/definitions/200.html)) or in my case to an issue's description found by clang-tidy in a C++ program, like e.g. https://clang.llvm.org/extra/clang-tidy/checks/readability/magic-numbers.html.

gh2193_new_column

Fixes:  #2193

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
@@ -21,6 +21,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.TreeSet;
 import java.util.stream.Stream;
@@ -45,8 +46,10 @@ public class ContentGeneratorDescriptor {
 
 	private IConfigurationElement configurationElement;
 	private MarkerField[] allFields;
+	private MarkerField[] allFieldsWithExtensions;
 	private Collection<MarkerType> markerTypes;
 	private MarkerField[] initialVisible;
+	private MarkerField[] initialVisibleWithExtensions;
 	private Collection<MarkerGroup> groups;
 	private Collection<IConfigurationElement> generatorExtensions = new ArrayList<>();
 	private Map<String, MarkerType> allTypesTable;
@@ -79,6 +82,7 @@ public class ContentGeneratorDescriptor {
 	 */
 	public void addExtensions(Collection<IConfigurationElement> extensions) {
 		generatorExtensions = extensions;
+		clearCaches();
 	}
 
 	/**
@@ -97,7 +101,7 @@ public class ContentGeneratorDescriptor {
 	 * @return boolean
 	 */
 	public boolean allTypesSelected(Collection<MarkerType> selectedTypes) {
-		return selectedTypes.containsAll(markerTypes);
+		return selectedTypes.containsAll(getMarkerTypes());
 	}
 
 	/**
@@ -106,7 +110,14 @@ public class ContentGeneratorDescriptor {
 	 * @return {@link MarkerField}[]
 	 */
 	public MarkerField[] getAllFields() {
-		return allFields;
+		if (allFieldsWithExtensions == null) {
+			List<MarkerField> fields = new ArrayList<>();
+			fields.addAll(Arrays.asList(allFields));
+			getExtensionsDescriptorsStream().map(d -> Arrays.asList(d.getAllFields())).flatMap(Collection::stream)
+					.forEach(fields::add);
+			allFieldsWithExtensions = fields.toArray(MarkerField[]::new);
+		}
+		return allFieldsWithExtensions;
 	}
 
 	/**
@@ -157,7 +168,14 @@ public class ContentGeneratorDescriptor {
 	 * @return {@link MarkerField}[]
 	 */
 	public MarkerField[] getInitialVisible() {
-		return initialVisible;
+		if (initialVisibleWithExtensions == null) {
+			List<MarkerField> fields = new ArrayList<>();
+			fields.addAll(Arrays.asList(initialVisible));
+			getExtensionsDescriptorsStream().map(d -> Arrays.asList(d.getInitialVisible())).flatMap(Collection::stream)
+					.forEach(fields::add);
+			initialVisibleWithExtensions = fields.toArray(MarkerField[]::new);
+		}
+		return initialVisibleWithExtensions;
 	}
 
 	/**
@@ -215,6 +233,8 @@ public class ContentGeneratorDescriptor {
 				MarkerType[] types = MarkerTypesModel.getInstance().getType(IMarker.PROBLEM).getAllSubTypes();
 				markerTypes.addAll(Arrays.asList(types));
 			}
+			getExtensionsDescriptorsStream().map(ContentGeneratorDescriptor::getMarkerTypes).flatMap(Collection::stream)
+					.forEach(markerTypes::add);
 		}
 		return markerTypes;
 	}
@@ -247,7 +267,7 @@ public class ContentGeneratorDescriptor {
 		if (allTypesTable == null) {
 			allTypesTable = new HashMap<>();
 
-			Iterator<MarkerType> allIterator = markerTypes.iterator();
+			Iterator<MarkerType> allIterator = getMarkerTypes().iterator();
 			while (allIterator.hasNext()) {
 				MarkerType next = allIterator.next();
 				allTypesTable.put(next.getId(), next);
@@ -294,5 +314,28 @@ public class ContentGeneratorDescriptor {
 	 */
 	public void removeExtension(IConfigurationElement element) {
 		generatorExtensions.remove(element);
+		clearCaches();
 	}
+
+	private void clearCaches() {
+		allFieldsWithExtensions = null;
+		initialVisibleWithExtensions = null;
+		markerTypes = null;
+		groups = null;
+		allTypesTable = null;
+	}
+
+	private Stream<ContentGeneratorDescriptor> getExtensionsDescriptorsStream() {
+		if (generatorExtensions != null) {
+			MarkerSupportRegistry registry = MarkerSupportRegistry.getInstance();
+			return generatorExtensions.stream()
+					.map(extensionConfigElem -> extensionConfigElem
+							.getAttribute(MarkerSupportInternalUtilities.ATTRIBUTE_ID))
+					.filter(id -> id != null && !id.isBlank())
+					.map(contentGeneratorId -> registry.getContentGenDescriptor(contentGeneratorId))
+					.filter(generator -> generator != null);
+		}
+		return Stream.empty();
+	}
+
 }
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/markers/MarkerSupportViewTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/markers/MarkerSupportViewTest.java
@@ -14,7 +14,11 @@ package org.eclipse.ui.tests.markers;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import org.eclipse.jface.viewers.CheckboxTableViewer;
 import org.eclipse.swt.widgets.Button;
@@ -25,8 +29,13 @@ import org.eclipse.ui.internal.views.markers.ExtendedMarkersView;
 import org.eclipse.ui.internal.views.markers.FiltersConfigurationDialog;
 import org.eclipse.ui.internal.views.markers.MarkerContentGenerator;
 import org.eclipse.ui.tests.harness.util.UITestCase;
+import org.eclipse.ui.views.markers.MarkerField;
 import org.eclipse.ui.views.markers.MarkerSupportView;
+import org.eclipse.ui.views.markers.internal.ContentGeneratorDescriptor;
+import org.eclipse.ui.views.markers.internal.MarkerGroup;
 import org.eclipse.ui.views.markers.internal.MarkerMessages;
+import org.eclipse.ui.views.markers.internal.MarkerSupportRegistry;
+import org.eclipse.ui.views.markers.internal.MarkerType;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
@@ -123,6 +132,70 @@ public class MarkerSupportViewTest extends UITestCase {
 
 		boolean isLimitEnabled = view.isMarkerLimitsEnabled();
 		assertFalse(isLimitEnabled);
+	}
+
+	@Test
+	public void markerContentGeneratorExtensionLoaded() throws Exception {
+		MarkerSupportView view = (MarkerSupportView) PlatformUI.getWorkbench().getActiveWorkbenchWindow()
+				.getActivePage().showView(PROBLEM_VIEW_ID);
+
+		MarkerContentGenerator generator = getMarkerContentGenerator(view);
+		ContentGeneratorDescriptor descriptor = MarkerSupportRegistry.getInstance()
+				.getContentGenDescriptor(generator.getId());
+
+		assertNotNull(descriptor);
+
+		MarkerField[] allFields = descriptor.getAllFields();
+		List<String> allFieldNames = mapToNames(allFields);
+		String fieldName1 = "Problem Key";
+		String fieldName2 = "Problem Key V2";
+
+		assertTrue(
+				"Expected loading marker field '" + fieldName1
+						+ "' from marker content generator extensions, but got only " + allFieldNames,
+				allFieldNames.contains(fieldName1));
+		assertTrue(
+				"Expected recursively loading marker field '" + fieldName2
+						+ "' from marker content generator extensions, but got only " + allFieldNames,
+				allFieldNames.contains(fieldName2));
+
+		MarkerField[] initiallyVisibleFields = descriptor.getInitialVisible();
+		List<String> initiallyVisibleFieldNames = mapToNames(initiallyVisibleFields);
+
+		assertTrue("Expected marker field '" + fieldName1
+				+ "' from marker content generator extension being visible according to 'visible' attribute in the extension,"
+				+ " but only the following marker fields are visible " + initiallyVisibleFieldNames,
+				initiallyVisibleFieldNames.contains(fieldName1));
+		assertFalse("Expected marker field '" + fieldName2
+				+ "' from marker content generator extension being not visible according to 'visible' attribute in the extension,"
+				+ " but the following marker fields are visible " + initiallyVisibleFieldNames,
+				initiallyVisibleFieldNames.contains(fieldName2));
+
+		String markerTypeId = "org.eclipse.ui.tests.markers.artificial.problem";
+		MarkerType markerTypeFromExtension = descriptor.getType(markerTypeId);
+		List<String> markerTypeIds = descriptor.getMarkerTypes().stream().map(MarkerType::getId)
+				.collect(Collectors.toList());
+
+		assertNotNull("Marker type with id '" + markerTypeId + "' not loaded from marker content generator extension.",
+				markerTypeFromExtension);
+		assertTrue("Expected marker type id '" + markerTypeId + "' being in marker types list, but we have only "
+				+ markerTypeIds, markerTypeIds.contains(markerTypeId));
+
+		Collection<MarkerGroup> groups = descriptor.getMarkerGroups();
+		List<String> groupIds = groups.stream().map(MarkerGroup::getId).collect(Collectors.toList());
+		String groupId = "org.eclipse.ui.tests.test.extended";
+
+		assertTrue("Expected loading group id '" + groupId
+				+ "' from marker content generator extension, but got only the following group ids: " + groupIds,
+				groupIds.contains(groupId));
+	}
+
+	private List<String> mapToNames(MarkerField[] markerFields) {
+		if (markerFields == null || markerFields.length == 0) {
+			return Collections.emptyList();
+		}
+
+		return Arrays.stream(markerFields).map(mf -> mf.getName()).collect(Collectors.toList());
 	}
 
 	public static MarkerContentGenerator getMarkerContentGenerator(MarkerSupportView view) {
--- a/tests/org.eclipse.ui.tests/plugin.xml
+++ b/tests/org.eclipse.ui.tests/plugin.xml
@@ -3461,7 +3461,58 @@
        </markerContentGenerator>
         <markerContentGenerator
              id="org.eclipse.ui.tests.customScopeContentGenerator">
-       </markerContentGenerator>       
+       </markerContentGenerator>
+        <markerField
+              class="org.eclipse.ui.tests.markers.ProblemKeyMarkerField"
+              id="org.eclipse.ui.tests.markerField.problemKey"
+              name="Problem Key">
+        </markerField>
+        <markerField
+              class="org.eclipse.ui.tests.markers.ProblemKeyMarkerField"
+              id="org.eclipse.ui.tests.markerField.problemKeyV2"
+              name="Problem Key V2">
+        </markerField>
+        <markerContentGenerator
+              id="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator"
+              name="Additional Problem Marker Content Generator">
+           <markerFieldReference
+                 visible="true"
+                 id="org.eclipse.ui.tests.markerField.problemKey">
+           </markerFieldReference>
+           <markerTypeReference
+                 id="org.eclipse.ui.tests.markers.static.analysis.problem">
+           </markerTypeReference>
+           <markerTypeReference
+                 id="org.eclipse.ui.tests.markers.artificial.problem">
+           </markerTypeReference>
+        </markerContentGenerator>
+        <markerContentGeneratorExtension
+              generatorId="org.eclipse.ui.ide.problemsGenerator"
+              id="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator">
+           <markerGrouping
+                 id="org.eclipse.ui.tests.test.extended"
+                 label="Extended Problem Category">
+           </markerGrouping>
+        </markerContentGeneratorExtension>
+        <markerContentGeneratorExtension
+              generatorId="org.eclipse.ui.ide.allMarkersGenerator"
+              id="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator">
+        </markerContentGeneratorExtension>
+        <markerContentGenerator
+              id="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator2"
+              name="Additional Problem Marker Content Generator 2">
+           <markerFieldReference
+                 visible="false"
+                 id="org.eclipse.ui.tests.markerField.problemKeyV2">
+           </markerFieldReference>
+           <markerTypeReference
+                 id="org.eclipse.ui.tests.markers.static.analysis.problem">
+           </markerTypeReference>
+        </markerContentGenerator>
+        <markerContentGeneratorExtension
+              generatorId="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator"
+              id="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator2">
+        </markerContentGeneratorExtension>
    </extension>
    <extension
          id="categoryTestMarker"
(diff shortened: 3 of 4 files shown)


merks:
Note that we are in the RC1 phase of development where it's more difficult to push though changes that might cause regressions.  I think this will unfortunately need to wait until the 4.34 stream opens for development.

Also, have a read through this:

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-a-pull-request

E.g., 

> You will usually be asked to squash your commits to a single one before a PR will be merged.

travkin79:
> Note that we are in the RC1 phase of development where it's more difficult to push though changes that might cause regressions. I think this will unfortunately need to wait until the 4.34 stream opens for development.
> 
> Also, have a read through this:
> 
> https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-a-pull-request
> 
> E.g.,
> 
> > You will usually be asked to squash your commits to a single one before a PR will be merged.

Thank you for the hints, Ed. I guess, I have to wait at least until the latest commit in the master branch has a successful CI run. When will the 4.34 stream be open for development? Is there a timeline?

merks:
Here's the release train schedule:

https://github.com/eclipse-simrel/.github/blob/main/wiki/SimRel/2024-09.md

the calendar details:

https://calendar.google.com/calendar/u/0/embed?src=gchs7nm4nvpm837469ddj9tjlk@group.calendar.google.com&dates=20240901/20240930&hl=en&mode=AGENDA

The platform is due to contribute its final build on August 30:

image

So I expect the following week that new I-builds for 4.34 will be configured and at some point during that week master will be declared open.

rubenporras:
@merks , without putting much pressure, I think this is a very nice addition to the product, and if the new extension point is not used, an small change.

Would it be possible to have it merged before RC1?

akurtakov:
Not used but going in a release means this will become API without being even used with changes following 2 years deprecation period. IMO this is really bad idea. Practically tonight should be the last build with functional changes for 4.33 and getting new functionality in it means there is no room for maneuver. 
This one should wait couple of weeks and be submitter for 4.34 release. Releasing is already too cumbersome job to risk adding extra work at this point. 
Please see https://www.eclipse.org/lists/eclipse-dev/msg12304.html which describes what should happen during current week. 
With every release procedures are streamlined and freeze periods are reduced but this doesn't happen magically - it requires far bigger investments in stabilizing tests and etc. I haven't seen a clean build in long period (e.g. https://download.eclipse.org/eclipse/downloads/drops4/I20240819-1800/ ) which to me means we have reached the most open state of development for now. I understand this may not look strictly related (or even totally unrelated) to your request but for those of us that have to throw the whole release out all these things are one and the same as they make us spend time daily trying to get to a manageable state.

travkin79:
I'd appreciate it if someone would review my PR, maybe @trancexpress?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
```diff
@@ -185,23 +207,41 @@ public Collection<MarkerGroup> getMarkerGroups() {
 	 * @return Collection of {@link MarkerType}
 	 */
 	public Collection<MarkerType> getMarkerTypes() {
-		if (markerTypes == null) {
```
trancexpress: Any reason to drop the caching here? Do you expect dynamic additions to this list?

From the changes here I assume whatever new addition there is, its contributed via some plug-in. It should be OK to keep `markerTypes` and fill it with the new content, so that `getMarkerTypes()` does its computations only once and then keeps the result in `markerTypes`. E.g. if `allTypesSelected()` is being called a lot by whatever code.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
```diff
@@ -185,23 +207,41 @@ public Collection<MarkerGroup> getMarkerGroups() {
...
-				String elementName = configElement.getAttribute(MarkerSupportInternalUtilities.ATTRIBUTE_ID);
-				MarkerType[] types = MarkerTypesModel.getInstance().getType(elementName).getAllSubTypes();
-				markerTypes.addAll(Arrays.asList(types));
-				markerTypes.add(MarkerTypesModel.getInstance().getType(elementName));
-			}
-			if (markerTypes.isEmpty()) {
-				MarkerType[] types = MarkerTypesModel.getInstance().getType(IMarker.PROBLEM).getAllSubTypes();
-				markerTypes.addAll(Arrays.asList(types));
+		Set<MarkerType> markerTypes = new HashSet<>();
+		addMarkerTypesFrom(configurationElement, markerTypes);
```
trancexpress: Is it possible to keep the old code here and move the new code to a method? Would make reading the diff easier.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
```diff
@@ -227,14 +267,12 @@ public MarkerType getType(String typeId) {
 	 * @return Map of {@link String} to {@link MarkerType}
 	 */
 	public Map<String, MarkerType> getTypesTable() {
-		if (allTypesTable == null) {
```
trancexpress, line 247: This again removes a cache. Is that necessary?

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/markers/MarkerSupportViewTest.java
```diff
@@ -125,6 +130,59 @@ public void limitDisabled() throws Exception {
...
+		MarkerContentGenerator generator = getMarkerContentGenerator(view);
+		ContentGeneratorDescriptor descriptor = MarkerSupportRegistry.getInstance()
+				.getContentGenDescriptor(generator.getId());
+		assertNotNull(descriptor);
+
+		MarkerField[] allFields = descriptor.getAllFields();
+		boolean foundAdditionalField1 = false;
+		boolean foundRecursivelyAddedField2 = false;
+		for (MarkerField field : allFields) {
+			if (ProblemKeyMarkerField.class.equals(field.getClass())) {
```
trancexpress: Maybe better to define some helper method that filters `allFields` with this `if` statement, then pack `field.getName()` to a result `List<String>`.

Then you can just do an `assertTrue("..., but found: " + names, names.contains("Problem Key"));`

This has the benefit of actually seeing what is in `allFields` w.r.t. the test; it would be easier to understand failures (especially sporadic ones). It would also reduce the length of the test and make it easier to read.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/markers/MarkerSupportViewTest.java
```diff
@@ -125,6 +130,59 @@ public void limitDisabled() throws Exception {
...
+			if ("Problem Key".equals(field.getName())) {
+				foundAdditionalVisbileField = true;
+			}
+		}
+		assertTrue("Visible attribute from marker fields from marker content generator extensions not loaded.",
+				foundAdditionalVisbileField);
+
+		MarkerType markerTypeFromExtension = descriptor.getType("org.eclipse.ui.tests.markers.artificial.problem");
+		assertNotNull("Marker type from marker content generator extensions not loaded.", markerTypeFromExtension);
+		assertTrue(descriptor.getMarkerTypes().contains(markerTypeFromExtension));
```
trancexpress: Please add some fail message here that lists the contents of `getMarkerTypes()`.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
```diff
@@ -20,7 +20,9 @@
 import java.util.HashMap;
```
trancexpress, line 21: I'm guessing `eclipse.platform.ui/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF` needs a version bump too?

tests/org.eclipse.ui.tests/plugin.xml
```diff
@@ -3411,7 +3411,56 @@
        </markerContentGenerator>
         <markerContentGenerator
              id="org.eclipse.ui.tests.customScopeContentGenerator">
-       </markerContentGenerator>       
+       </markerContentGenerator>
+        <markerField
```
trancexpress, line 3465: I see the extensions here result in the new columns automatically being visible in the `Problems` view.

I'm not 100% sure we want that; assuming only 1 bundle makes such additions, there shouldn't be an issue.

But assuming multiple bundles would use the new extension, we would have lots of extra columns as soon as those bundles are added to the IDE. Generally contributions are rarely immediately visible (such as views, editors, perspectives and so on, even toolbar buttons often belong to activities).

I can't judge if this is something we want or not. @iloveeclipse should the user / maintainer need to enable/add new columns to the `Problems` view, if such are being contributed by plug-ins in the platform? Or is it fine to show them automatically (as soon as the plug-in is in the platform)? See e.g. Eclipse running with the `org.eclipse.ui.tests` plug-in (in particular, the new test columns "Problem Key" and "Problem Key V2", those would essentially be contributions of plug-ins using the new capabilities added in this PR):

image

My initial expectation when screening this PR was that the `Problems` view would make the new columns available for addition, when going in the view options.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
travkin79: Thank you for reviewing, @trancexpress.

I was not sure if another version bump is needed, since a [preceding commit](https://github.com/eclipse-platform/eclipse.platform.ui/commit/b6441169f712a036bbd2bf43d7857826020e21c1) had a version bump on that bundle already.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
travkin79: I dropped the caching here and in other methods because of the extensions that are being added via `addExtensions(Collection<IConfigurationElement>)` or removed via `removeExtension(IConfigurationElement)`. Most getters in this class require loading elements from these extensions. I also guess that the extensions are provided by other plug-ins. I'm just not sure when extensions are added or removed and assumed that this can happen dynamically.

I think, if we want to keep collections of marker types, groups, fields, etc. up to date (caching), we would have to add or remove those provided by extensions each time an extension is added or removed via `addExtensions(Collection<IConfigurationElement>)` or `removeExtension(IConfigurationElement)`. I think, that would complicate the code and would require additional tests. Recursive loading of extensions makes thing even more complicated.

What do you think, @trancexpress?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
iloveeclipse: No, it was already bumped for 4.34

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
travkin79: I extracted two methods here, the old code into one method and the new code into another method. I think, it's more readable now.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/markers/MarkerSupportViewTest.java
travkin79: Good point. I changed the checks accordingly. Thank you.

travkin79:
It seems, I fell into the [pitfall](https://github.com/orgs/community/discussions/10369) of not knowing, I have to explicitly publish my answers to the reviewers' comments.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
trancexpress: Sorry I didn't get a notification for this.

Do you have an actual case where your change doesn't work, given the cache? From my POV, better to keep what is already in the code; it likely speeds up some use case. I'm not aware of dynamically adding plug-ins to Eclipse at runtime, at least for us that is not a use case. If someone has this use case and has issues with the code here (prior to this change) they can open a ticket and the problem can be looked into.

Or am I misunderstanding something here? Does the cache interfere with your change somehow?

iloveeclipse:
> This way, it is now possible to extends, e.g. the Problems View and the Markers View with new columns.

Could you please add a screenshot with such example?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
iloveeclipse: The caching makes sense because 1) extensions are rarely added/removed **on the fly** 2) this code would need to read config area that could be on NFS every time 3) the Problems view is almost always used & updated on selection/build changes etc.
So please don't compute extension points on every call

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
iloveeclipse: Not, and shouldn't be removed.

tests/org.eclipse.ui.tests/plugin.xml
travkin79: > My initial expectation when screening this PR was that the Problems view would make the new columns available for addition, when going in the view options.

I admit, I had the same expectation (new columns would have to be added via the 'Configure Columns' dialog, but I did not test if that's really the case. I think, we should change that so that the users decide what the want to add to the problems view / markers view.

iloveeclipse: What surprises me looking on this PR is the absence of new API added => means, it looks you use existing extensions point data to show some potentially already existing extensions in the Problems view that were hidden before.

If this assumption is correct, then extension point should be updated with the extra argument "yes, that contribution should be shown in the Problems view".

Otherwise, existing code will cause unexpected data to be shown that wasn't ever supposed to be shown.

iloveeclipse: BTW, there is also "Markers" view too, and it might show more content as Problems view by default. 
So it should be clear for the extension point in which of the views the content is supposed to be shown.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
travkin79: I understand the reasons for using caching.

Nevertheless, unfortunately, in this case, keeping the list of provided extensions up to date becomes very complicated. My first experiments with filling the lists / arrays of marker types, marker fields, etc. for caching and updating them when loading / unloading extensions corrupt recursive loading of marker support extensions, I just run into an infinite chain of method calls and a stack overflow issue. The whole concept of loading marker support extensions was not built for recursively loading marker support extensions.

I'll check that once again. Maybe there is another way of implementing that with caching.

iloveeclipse: > Nevertheless, unfortunately, in this case, keeping the list of provided extensions up to date becomes very complicated.

Start with simple code, don't update extensions, and you will cover 99.9% of use cases already. If someone runs into troubles because of that more complex code can be added to cover remaining 0.1% of use cases.

tests/org.eclipse.ui.tests/plugin.xml
travkin79: I did not change the existing API, that's correct. In theory, someone could have implemented marker support extensions with additional marker fields or types that are not usable at the moment and would become usable after merging this PR. (I think, this is very unlikely.) Would that be a problem if each user would still have to explicitly add these new fields to the list of visible fields in the problems or marker view? (I think, I should adapt this PR to let the user explicitly add the new fields to the views. I hope that can be done easily.)

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
travkin79: The markerSupport extension point offers a visible field in its markerTypeReference entries. It's documentation says:

image

It seems, the default of making the fields visible in case there is no entry in the visible field is questionable. Changing the default would usually brake many existing extensions, making visible fields invisible, but since loading marker content generator extensions does not work at the moment, we might safely change the default. I'll check that.

tests/org.eclipse.ui.tests/plugin.xml
trancexpress: OK, having looked into the change. It seems quite odd that `ContentGeneratorDescriptor` lists the extensions `markerContentGeneratorExtension`. I would have expecte code that cares about marker content to check the extensions, on top of the "main" content. However:

1. There is no direct API to iterate over extensions of `ContentGeneratorDescriptor`.
2. `ContentGeneratorDescriptor.getFilterReferences()` goes over the extensions of the descriptor

So the PR does seem to simply add missing functionality.

Also odd to me is that `ContentGeneratorDescriptor.getFilterReferences()` does no caching, unlike the other methods. I assume this was the reason why @travkin79 removed the existing caching. From my POV, we should keep the caching. The main reason for the lack of caching is, I assume, `ContentGeneratorDescriptor.removeExtension()`. That one is likely called dynamically. `ContentGeneratorDescriptor.addExtensions()` is called (so far) only when constructing the descriptor.

So, if we want to actually be correct, I believe we must cache more dynamically. Possibly on the add and remove methods... I think for the `removeExtension` method this is probably necessary... The removes would be coming from: `ExtensionTracker.registryChanged()`

I'll look into this. In the worst case we cache as in the previous state in the PR (which I changed). I.e. cache the descriptor contributions, go over the extensions on-the-fly (no idea how costly that would be, its already done for the filters though... so nothing new).

Essentially, the `plugin.xml` glue is this:

```
        <markerContentGeneratorExtension
              generatorId="org.eclipse.ui.ide.problemsGenerator"
              id="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator">
```

With the extension being:

```
       <markerContentGenerator
              id="org.eclipse.ui.tests.additionalProblemMarkerContentGenerator"
              name="Additional Problem Marker Content Generator">
           <markerFieldReference
           	     visible="true"
                 id="org.eclipse.ui.tests.markerField.problemKey">
```

This tells Eclipse, which view will be extended (the `Problems` view in this case) and with what.

If an existing contribution results in a column that is now visible, but this is not expected, that contribution should be adjusted to use `visible="false"`.

Though I have doubts that there are such contributions, considering that so far they do nothing.

trancexpress: @travkin79  in the PR description / commit message you mention that groups are also touched by this change. Yet in the previous state of the PR I didn't see any change for group code?

trancexpress:
I did grep what platform/Eclipse/Xtext repositories I have, the extension in question is barely used. I only find mentions in the platform UI repository.

```
In directory: eclipse.platform.ui/
bundles/org.eclipse.ui.ide/schema/markerSupport.exsd:            <element ref="markerContentGeneratorExtension" minOccurs="0" maxOccurs="unbounded"/>
bundles/org.eclipse.ui.ide/schema/markerSupport.exsd:   <element name="markerContentGeneratorExtension">
bundles/org.eclipse.ui.ide/schema/markerSupport.exsd:            A markerContentGeneratorExtension is an extension to an existing markerContentGenerator.
bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/MarkerSupportRegistry.java:        private static final String MARKER_CONTENT_GENERATOR_EXTENSION = "markerContentGeneratorExtension"; //$NON-NLS-1$
tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/markers/MarkerSupportViewTest.java:    public void markerContentGeneratorExtensionLoaded() throws Exception {
tests/org.eclipse.ui.tests/plugin.xml:       <markerContentGeneratorExtension
tests/org.eclipse.ui.tests/plugin.xml:       </markerContentGeneratorExtension>
tests/org.eclipse.ui.tests/plugin.xml:        <markerContentGeneratorExtension
tests/org.eclipse.ui.tests/plugin.xml:        </markerContentGeneratorExtension>
tests/org.eclipse.ui.tests/plugin.xml:        <markerContentGeneratorExtension
tests/org.eclipse.ui.tests/plugin.xml:        </markerContentGeneratorExtension>
tests/org.eclipse.ui.tests/plugin.xml:        <markerContentGeneratorExtension
tests/org.eclipse.ui.tests/plugin.xml:        </markerContentGeneratorExtension>
```

trancexpress:
@travkin79 feel free to do code clean-ups after we've merged the PR, I've adjusted it to make the review as easy as possible.

trancexpress:
OK, last remaining item:

@travkin79 , you mention groups in the original commit message. I didn't notice a change in the group code though (aside from removing the caching). Can you clarify? `ContentGeneratorDescriptor.getMarkerGroups()` is supposed to be supported too?

travkin79:
Hi @trancexpress,
Thank you for looking deeper into my changes and for improving the code. I think, your solution of clearing caches as soon as extensions change works better than calculating all collection always on demand.

> @travkin79 feel free to do code clean-ups after we've merged the PR, I've adjusted it to make the review as easy as possible.

I guess, you mean, before merging the PR, right? I think, I can only commit changes to my fork. I think, I'll make a few minor improvements in the fork / PR today.

> @travkin79 , you mention groups in the original commit message. I didn't notice a change in the group code though (aside from removing the caching). Can you clarify? ContentGeneratorDescriptor.getMarkerGroups() is supposed to be supported too?

You're right. I wanted to support groups, fields (all and the initially visible ones), and types. There were no significant changes on the groups code.

trancexpress:
> > @travkin79 feel free to do code clean-ups after we've merged the PR, I've adjusted it to make the review as easy as possible.
> 
> I guess, you mean, before merging the PR, right? I think, I can only commit changes to my fork. I think, I'll make a few minor improvements in the fork / PR today.

No, I do mean after the PR is merged. Please keep clean-ups separate of this change, unless you mean to do clean-ups in the added code.

> > @travkin79 , you mention groups in the original commit message. I didn't notice a change in the group code though (aside from removing the caching). Can you clarify? ContentGeneratorDescriptor.getMarkerGroups() is supposed to be supported too?
> 
> You're right. I wanted to support groups, fields (all and the initially visible ones), and types. There were no significant changes on the groups code.

Why is there no change in the groups code? Shouldn't the groups also take groups of extensions? I.e. how its done for fields and marker types?

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/views/markers/internal/ContentGeneratorDescriptor.java
```diff
@@ -277,5 +297,26 @@ public void initializeFromConfigurationElement(
 	 */
 	public void removeExtension(IConfigurationElement element) {
 		generatorExtensions.remove(element);
+		clearMarkerFieldCaches();
+	}
+
+	private void clearMarkerFieldCaches() {
```
travkin79: @trancexpress, I think, we have to also clear the collections `groups` and `allTypesTable` here, since these are also filled from extensions. I would rename the method to `clearCaches()` since we basically would clear almost all collections here.
In addition, we have to ensure that at all places where we access these collections, we use getters that ensure, these collections are initialized / filled.

For better traceability, I'll commit these two changes in additional commits. If you agree to these changes, we can squash the commit to a single commit and eventually rebase on master before merging.

travkin79:
> Why is there no change in the groups code? Shouldn't the groups also take groups of extensions? I.e. how its done for fields and marker types?

Yes, you're right. But the code from master already does that. It's just not that obvious. The method `addDefinedGroups(Collection<MarkerGroup>)` does run through all extensions and adds the groups accordingly. The new test case ensures that groups from extensions are loaded, too.

> No, I do mean after the PR is merged. Please keep clean-ups separate of this change, unless you mean to do clean-ups in the added code.

Ok. I guess that's for keeping the PRs as simple as possible to simplify reviews.

travkin79:
There's another thing left open: We have to decide whether to change the default of making new marker fields invisible by default. I think, we can safely do that since the marker content generator extensions didn't work before. Thus, no fields would suddenly disappear because of our change.

I'm not sure if we should also change the extension point, as suggested by @iloveeclipse, in order to make clear, in which views the fields will be visible. I tend to leave it as it is, assuming that new fields will be invisible by default and a user would decide whether to add them to certain views (e.g. problems view, markers view).

trancexpress:
> There's another thing left open: We have to decide whether to change the default of making new marker fields invisible by default. I think, we can safely do that since the marker content generator extensions didn't work before. Thus, no fields would suddenly disappear because of our change.
> 
> I'm not sure if we should also change the extension point, as suggested by @iloveeclipse, in order to make clear, in which views the fields will be visible. I tend to leave it as it is, assuming that new fields will be invisible by default and a user would decide whether to add them to certain views (e.g. problems view, markers view).

From my POV this PR is a bug fix, adding missing functionality. If any client code unintentionally adds visible fields (either due to the default when not specifying visibility, or due to explicitly specifying a visible field), that client will have to be adjusted to work with latest the platform (i.e. set the visibility as desired). So from my side, leave as is - the change follows the API documentation.

@iloveeclipse ?

trancexpress:
Will merge the PR tomorrow. Let me know if there is any feedback that has not been addressed.

PR #2209, opened 2024-08-19
Find/replace: move input state and incremental search handling into FindReplaceLogic

HeikoKlare:
<s>⚠️ This PR consists of three subsequent changes. Due to their dependency and because they should not be merged before development for 2024-12 M1 starts, I chose to provide them in three separate commits but one PR instead of having three interdependent PRs. It would still be possible to factor out single commits into separate PRs, if necessary.</s>

The PR addresses these three concerns:
- The FindReplaceLogic is a stateful representation of a search context for a search target. It currently already has a state of the used search options. Still, the find and replace string used for performing operations on the target are passed to each operation instead of keeping it in the logic object itself. Keeping this information in the logic object simplifies its interface and allows to encapsulate more of the incremental logic in the logic class, as it then has the information about when the search input changes. 
- The FindReplaceLogic provides different methods for retrieving the active/available state of different search options.
- Currently, consumers of the FindReplaceLogic are responsible for properly handling incremental searches by executing a search when the input text changes and by disabling incremental mode when an ordinary search shall be performed.

It provides these three according improvements:
- Adds the find and replace input string state to the FindReplaceLogic class and simplifies its interface accordingly. In addition, it works towards a clear model (logic) and view (dialog/overlay) separation.
- Unifies the state retrieval by providing an isAvailable() and isAvailableAndActive() method accepting a search option instance, such that the information can be retrieved for any search option of interest in a unform way.
- Incremental "search-as-you-type" is automatically performed by the FindReplaceLogic when the incremental search option is
activated. Explicitly performing a search does require to deactivate the incremental mode anymore.

This also replaces the intermediate solution from #2205

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -61,7 +61,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	@Override
 	public void setFindString(String findString) {
 		this.findString = Objects.requireNonNull(findString);
-		if (isIncrementalSearchAvailable() && isActive(SearchOptions.INCREMENTAL)) {
+		if (isAvailableAndActive(SearchOptions.INCREMENTAL)) {
 			performSearch(true);
 		}
 	}
@@ -130,14 +130,28 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	}
 
 	@Override
-	public boolean isIncrementalSearchAvailable() {
-		return !isRegExSearchAvailableAndActive();
+	public boolean isAvailable(SearchOptions searchOption) {
+		switch (searchOption) {
+		case INCREMENTAL:
+			return !isAvailableAndActive(SearchOptions.REGEX);
+		case REGEX:
+			return isTargetSupportingRegEx;
+		case WHOLE_WORD:
+			return !isAvailableAndActive(SearchOptions.REGEX) && isWord(findString);
+		case CASE_SENSITIVE:
+		case FORWARD:
+		case GLOBAL:
+		case WRAP:
+		default:
+			return true;
+		}
 	}
 
 	@Override
-	public boolean isWholeWordSearchAvailable() {
-		return !isRegExSearchAvailableAndActive() && isWord(findString);
+	public boolean isAvailableAndActive(SearchOptions searchOption) {
+		return isAvailable(searchOption) && isActive(searchOption);
 	}
+
 	/**
 	 * Tests whether each character in the given string is a letter.
 	 *
@@ -146,11 +160,6 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	 */
 	private static boolean isWord(String str) {
 		return str != null && str.chars().allMatch(Character::isJavaIdentifierPart);
-	}
-
-	@Override
-	public boolean isRegExSearchAvailableAndActive() {
-		return isActive(SearchOptions.REGEX) && isTargetSupportingRegEx;
 	}
 
 	@Override
@@ -185,7 +194,8 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		Point lineSelection = extensionTarget.getLineSelection();
 		scope = new Region(lineSelection.x, lineSelection.y);
 
-		int offset = isActive(SearchOptions.FORWARD) ? scope.getOffset() : scope.getOffset() + scope.getLength();
+		int offset = isAvailableAndActive(SearchOptions.FORWARD) ? scope.getOffset()
+				: scope.getOffset() + scope.getLength();
 
 		extensionTarget.setSelection(offset, 0);
 		extensionTarget.setScope(scope);
@@ -318,7 +328,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	private boolean performSearch(boolean validateSearchOptions) {
 		resetStatus();
 
-		if (validateSearchOptions && (isActive(SearchOptions.INCREMENTAL) && !isIncrementalSearchAvailable())) {
+		if (validateSearchOptions && !isAvailable(SearchOptions.INCREMENTAL) && isActive(SearchOptions.INCREMENTAL)) {
 			return false; // Do nothing if search options are not compatible
 		}
 		boolean somethingFound = false;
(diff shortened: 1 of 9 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -312,7 +337,7 @@ private boolean replaceSelection(String replaceString) {
 
 		boolean replaced;
 		try {
-			replaceSelection(replaceString, isRegExSearchAvailableAndActive());
+			replaceSelection(isAvailableAndActive(SearchOptions.REGEX));
```
Wittmaxi: why should this local state be passed anyway?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
```diff
@@ -25,6 +25,20 @@
  */
 public interface IFindReplaceLogic {
 
+	/**
+	 * Sets the string to be used for searching in find and replace operations.
+	 *
+	 * @param findString the find string to use, must not be null
+	 */
+	public void setFindString(String findString);
```
Wittmaxi, line 38: Somehow this method does more than I would expect just from reading this header.
The method
1) performs a null-check
2) updates incremental search
3) thus invalidates the search status

I am not sure if the name "setFindString" is good or not, however the javadocs should definitely somehow reflect these side-effects

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -497,20 +514,19 @@ private Point replaceSelection(String replaceString, boolean regExReplace) {
 	 * Returns whether the specified search string can be found using the given
 	 * options.
 	 *
-	 * @param findString             the string to search for
 	 * @return <code>true</code> if the search string can be found using the given
 	 *         options
 	 *
 	 */
-	private boolean findNext(String findString) {
+	private boolean findNext(boolean updateFromExistingBaseLocation) {
```
Wittmaxi:
```suggestion
	private boolean findNext(boolean updateFromIncrementalBaseLocation) {
```

to be consistent with the variable called `incrementalBaseLocation`

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -527,49 +543,55 @@ private boolean findNext(String findString) {
 		return true;
 	}
 
-	private int calculateFindBeginningOffset() {
+	private int calculateFindBeginningOffset(boolean updateFromExistingBaseLocation) {
```
Wittmaxi, line 508:
```suggestion
	private int calculateFindBeginningOffset(boolean updateFromIncrementalBaseLocation) {
```

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
```diff
@@ -1128,6 +1139,22 @@ private void updateFindHistory() {
...
 
+	/**
+	 * Called after executed find action to update the history.
+	 */
+	private void updateReplaceHistory() {
+		if (okToUse(fReplaceField)) {
+			fReplaceField.removeModifyListener(fReplaceModifyListener);
+
+			// XXX: Workaround for Combo bug on Linux (see bug 404202 and bug 410603)
+			if (Util.isLinux())
```
Wittmaxi:
```suggestion
			if (Util.isLinux()) {
```

Now is a good opportunity to introduce braces into this code!

Wittmaxi (changes requested):
Some minor details to change. In particular, side effects are introduced to operations which look like they wouldn't have side-effects (updating search string performs an incremental search operation without being documented)

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
HeikoKlare: True, I have removed that parameter and simplified and inlined this method completely in 69af3d0.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
HeikoKlare: Yes, I have added them at this place and the other, similar code location in 69af3d0.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
HeikoKlare: Fixed in 9978946

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
HeikoKlare: Fixed in 9978946

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
HeikoKlare: That you for the catch! Indeed, the contract of `setFindString()` needs to be updated accordingly.

The null check is conforming to the method's documentation, requiring the parameter not to be null. The null check only ensures that it fails as early as possible.

The update of the incremental search and the change of the status was not documented yet. These are no "side effects" but intended effects when setting the find string while incremental search is active, so in my opinion the method's implementation and name are fine and only the contract needs to be updated accordingly. I have done that in 9978946.

Wittmaxi: The new comment looks good now, it clarifies well what happens under the hood.

Wittmaxi (approved):

HeikoKlare:
Failing test is unrelated and already documented:
- #1941

PR #2210, opened 2024-08-20
fix "Implicit narrowing conversion in compound assignment" warnings

jukzi:
All false positives, but avoidable with integer arithmetic instead of floating point.

https://github.com/eclipse-platform/eclipse.platform.ui/security/code-scanning/16
https://github.com/eclipse-platform/eclipse.platform.ui/security/code-scanning/15 
https://github.com/eclipse-platform/eclipse.platform.ui/security/code-scanning/2
https://github.com/eclipse-platform/eclipse.platform.ui/security/code-scanning/1

### DIFF
--- a/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/RefactoringDescriptor.java
+++ b/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/RefactoringDescriptor.java
@@ -357,7 +357,7 @@ public abstract class RefactoringDescriptor implements Comparable<RefactoringDes
 	public final int hashCode() {
 		int code= getDescription().hashCode();
 		if (fTimeStamp >= 0)
-			code+= (17 * fTimeStamp);
+			code+= 17 * Long.hashCode(fTimeStamp);
 		return code;
 	}
 
--- a/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/RefactoringDescriptorProxy.java
+++ b/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/RefactoringDescriptorProxy.java
@@ -108,7 +108,7 @@ public abstract class RefactoringDescriptorProxy extends PlatformObject implemen
 		int code= getDescription().hashCode();
 		final long stamp= getTimeStamp();
 		if (stamp >= 0)
-			code+= (17 * stamp);
+			code+= 17 * Long.hashCode(stamp);
 		return code;
 	}
 
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/ResourceFilterGroup.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/ResourceFilterGroup.java
@@ -2772,8 +2772,8 @@ class MultiMatcherCustomFilterArgumentUI implements ICustomFilterArgumentUI {
 		setupDescriptionText(null);
 	}
 
-	private String[] timeIntervalPrefixes = {"s", "m", "h", "d"};    //$NON-NLS-1$//$NON-NLS-2$//$NON-NLS-3$ //$NON-NLS-4$
-	private double[] timeIntervalScale = {60, 60, 24};
+	private static final String[] TIME_INTERVAL_PREFIXES = { "s", "m", "h", "d" }; //$NON-NLS-1$//$NON-NLS-2$//$NON-NLS-3$ //$NON-NLS-4$
+	private static final long[] TIME_INTERVAL_SCALE = { 60, 60, 24 };
 
 	private String convertToEditableTimeInterval(String string) {
 		if (string.isEmpty())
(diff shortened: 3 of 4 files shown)


bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/dialogs/ResourceFilterGroup.java
```diff
@@ -2772,8 +2772,8 @@ public void widgetSelected(SelectionEvent e) {
 		setupDescriptionText(null);
 	}
 
-	private String[] timeIntervalPrefixes = {"s", "m", "h", "d"};    //$NON-NLS-1$//$NON-NLS-2$//$NON-NLS-3$ //$NON-NLS-4$
-	private double[] timeIntervalScale = {60, 60, 24};
+	private static final String[] timeIntervalPrefixes = { "s", "m", "h", "d" }; //$NON-NLS-1$//$NON-NLS-2$//$NON-NLS-3$ //$NON-NLS-4$
```
merks: Making it static final is not related to the pr topic.  That’s okay, but generally static final fields use upper case names in the existing code base. I think we had that discussion…

jukzi:
https://github.com/eclipse-platform/eclipse.platform.ui/security/code-scanning?query=is%3Aopen+pr%3A2210+
image

bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/RefactoringDescriptorProxy.java
```diff
@@ -108,7 +108,7 @@ public final int hashCode() {
 		int code= getDescription().hashCode();
 		final long stamp= getTimeStamp();
 		if (stamp >= 0)
-			code+= (17 * stamp);
+			code+= 17 * Long.hashCode(stamp);
 		return code;
```
HannesWell, line 112: The entire body of this method could be replaced by
`Objects.hash(getDescription(), getTimeStamp())`

Only disadvantage would be the Long object creation, but I assume that's something that is not relevant.

bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/RefactoringDescriptor.java
```diff
@@ -357,7 +357,7 @@ public final long getTimeStamp() {
 	public final int hashCode() {
 		int code= getDescription().hashCode();
 		if (fTimeStamp >= 0)
-			code+= (17 * fTimeStamp);
+			code+= 17 * Long.hashCode(fTimeStamp);
```
HannesWell, line 360: The entire body of this method could be replaced by
`Objects.hash(getDescription(), fTimeStamp)`

Only disadvantage would be the Long object creation, but I assume that's something that is not relevant.

jukzi: unfortunately i don't know why https://github.com/eclipse-platform/eclipse.platform.ui/commit/c9a4ae2aa889a6cf9cdd36d1873b826334af5027 explicitly added the >=0 check (-1 means no timestamp) .

iloveeclipse: Looks like a bug, if hashcode depends on object state, it shouldn't change. But the whole class is written in a way that hashcode is broken by design, it depends on two fields that can be set after object is created, and also non atomic operation. So hashcode can return few different values inbetween.

merks: Maybe I misunderstand the comment. Even a list or a set have hashcode that depends on state that can change. It’s only broken if two instances that are equals have different hash codes.  Of course using them as keys in a map or set and changing their state is after is broken.  But is that the case here?

HannesWell: > unfortunately i don't know why [c9a4ae2](https://github.com/eclipse-platform/eclipse.platform.ui/commit/c9a4ae2aa889a6cf9cdd36d1873b826334af5027) explicitly added the >=0 check (-1 means no timestamp) .

Good question. I don't know it either. But as Ed said it, as long as equals and hashCode correspond to each other with respect to the hashCode/equals contract everything is fine.
And since `equals()` always tests if two `RefactoringDescriptor` objects have equal `timestamp` and `descriptor` actually `hashCode()` should be implemented accordingly, regardless of the value.
And on first sight I cannot see a specialty that requires something else.


> It’s only broken if two instances that are equals have different hash codes. Of course using them as keys in a map or set and changing their state is after is broken.

Oh yes that's really fun. As a beginner  you have really understood the fundamentals of hashCode/equals() after debugging errors because of such misuse. 😅

jukzi: As the author of that >= 0 check did multiple commits in that area to improve memory and performance for the refactoring history it feels better to not remove that check and not introduce ephemeral memory allocations.
BTW the only usecase of the hashcode i found was in tests like :

```
org.eclipse.ltk.internal.core.refactoring.history.DefaultRefactoringDescriptorProxy(org.eclipse.ltk.core.refactoring.RefactoringDescriptorProxy).hashCode() line: 108	
	java.util.HashMap<K,V>.hash(java.lang.Object) line: 338	
	java.util.HashMap<K,V>.put(K, V) line: 618	
	java.util.HashSet<E>.add(E) line: 229	
	java.util.HashSet<E>(java.util.AbstractCollection<E>).addAll(java.util.Collection<? extends E>) line: 338	
	org.eclipse.ltk.internal.core.refactoring.history.RefactoringHistoryManager.readRefactoringDescriptorProxies(org.eclipse.core.filesystem.IFileStore, java.lang.String, java.util.Collection<org.eclipse.ltk.core.refactoring.RefactoringDescriptorProxy>, long, long, org.eclipse.core.runtime.IProgressMonitor, java.lang.String) line: 215	
...
	org.eclipse.jdt.ui.tests.refactoring.RenameTests18.testMethodReference0() line: 292	
```

There the use of modifiable timestamp looks fine as it is not modified after it is set once

HannesWell: > As the author of that >= 0 check did multiple commits in that area to improve memory and performance for the refactoring history it feels better to not remove that check and not introduce ephemeral memory allocations.

That change was done almost 20 years ago. I think today (if not already back than) that this is probably a micro-optimization that no one will notice a benefit from. One can probably not even measure a difference, unless one benchmarks only the `hashCode()` method. Though it would be interesting to have a JMH case for that I will not spend time on it.

PR #213, opened 2022-07-15
@AboutToHide method of dynamic menu contribution not called

iloveeclipse:
Original patch from https://bugs.eclipse.org/bugs/show_bug.cgi?id=578964
most likely had a typo => fixed condition

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/207

PR #2211, opened 2024-08-21
Adapted the client to use new Browser API

amartya4256:
The contribution alloows the client to utilize the new API from SWT Browser to check if the url is the custom text url of the browser.

Contributes to #213

Must be merged only after https://github.com/eclipse-platform/eclipse.platform.swt/pull/1409 is merged

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
@@ -367,8 +367,9 @@ public class BrowserInformationControl extends AbstractInformationControl implem
 		 * Avoids flickering when replacing hovers, especially on Vista in ON_CLICK mode.
 		 * Causes flickering on GTK. Carbon does not care.
 		 */
-		if (Util.isWin32()) 
+		if (Util.isWin32()) {
 			shell.moveAbove(null);
+		}
 
 		super.setVisible(true);
 	}
@@ -583,6 +584,20 @@ public class BrowserInformationControl extends AbstractInformationControl implem
 	 */
 	public BrowserInformationControlInput getInput() {
 		return fInput;
+	}
+
+	/**
+	 * @param loc the location to be checked
+	 *
+	 * @return Returns whether the given location represents the location to which the this browser
+	 *         control has navigated via calling {@link #setContent(Object)}.
+	 *         <p>
+	 *         It may be used to validate in a location listener whether the changed location
+	 *         belongs to custom content or whether it belongs to some other location to which the
+	 *         user has navigated in the control).
+	 */
+	public boolean isLocationForCustomText(String loc) {
+		return fBrowser.isLocationForCustomText(loc);
 	}
 
 	@Override
--- a/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
+++ b/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
@@ -465,7 +465,7 @@ public class BrowserViewer extends Composite {
 					if (!event.top)
 						return;
 					if (combo != null) {
-						if (!"about:blank".equals(event.location)) { //$NON-NLS-1$
+						if (!browser.isLocationForCustomText(event.location)) {
 							combo.setText(event.location);
 							addToHistory(event.location);
 							updateHistory();
--- a/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
+++ b/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
@@ -474,7 +474,7 @@ public class BrowserViewer extends Composite {
 					}
 					if (showToolbar) {
 						// enable auto-refresh button if URL is a file
-						File temp = getFile(browser.getUrl());
+						File temp = getFile(browser.getUrlOptional().get());
 						if (temp != null && temp.exists()) {
 							autoRefresh.setEnabled(true);
 							if (autoRefresh.getSelection()) {
@@ -680,7 +680,7 @@ public class BrowserViewer extends Composite {
 	}
 
 	private void toggleAutoRefresh() {
-		File temp = getFile(browser.getUrl());
+		File temp = getFile(browser.getUrlOptional().orElse(null));
 		if (temp != null && temp.exists() && autoRefresh.getSelection()) {
 			refresh();
 			fileChangedWatchService(temp);
@@ -955,7 +955,7 @@ public class BrowserViewer extends Composite {
 		};
 		browser.addLocationListener(locationListener2);
 
-		File temp = getFile(browser.getUrl());
+		File temp = getFile(browser.getUrlOptional().get());
 		if (temp != null && temp.exists()) {
 			file = temp;
 			timestamp = file.lastModified();
@@ -1005,9 +1005,11 @@ public class BrowserViewer extends Composite {
 						for (WatchEvent<?> event : key.pollEvents()) {
 							final Path changedPath = (Path) event.context();
 							Display.getDefault().asyncExec(() -> {
-								if (browser.getUrl().endsWith(changedPath.toString())) {
-									browser.refresh();
-								}
+								browser.getUrlOptional().ifPresent(url -> {
+									if (url.endsWith(changedPath.toString())) {
+										browser.refresh();
+									}
+								});
 							});
 						}
 						key.reset();


bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
```diff
@@ -367,7 +367,7 @@ public void setVisible(boolean visible) {
 		 * Avoids flickering when replacing hovers, especially on Vista in ON_CLICK mode.
 		 * Causes flickering on GTK. Carbon does not care.
 		 */
-		if (Util.isWin32()) 
+		if (Util.isWin32())
```
HeikoKlare: Please revert formatting change or use the chance to put braces around the expression block.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
```diff
@@ -585,6 +585,13 @@ public BrowserInformationControlInput getInput() {
 		return fInput;
 	}
 
+	/**
+	 * @return true if the location is of the browser's custom text page, otherwise false
+	 */
+	public Boolean isLocationForBrowserCustomText(String loc) {
```
HeikoKlare: Is there a reason for using the boxed type instead of primite `boolean`? If not, please change to use boolean.
And please use a speaking parameter name, e.g., `location`.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
```diff
@@ -585,6 +585,13 @@ public BrowserInformationControlInput getInput() {
 		return fInput;
 	}
 
+	/**
+	 * @return true if the location is of the browser's custom text page, otherwise false
+	 */
```
HeikoKlare: Please add specification for the parameter and please improve precision of the specification of the methods behavior/return type. It should cover when this method is expected to return true, as consumer of this class does not know anything about the underlying browser implementation and API. It could be something like 
``` 
Returns whether the given location represents the location to which the this browser control has navigated via calling {@link #setContent(Object)}.
<p>
It may be used to validate in a location listener whether the changed location belongs to custom content or whether it belongs to some other location to which the user has navigated in the control).
```

This also leads to a follow up: probably this method should be called `isLocationForCustomContent`, as it's about the location to be set by calling `setContent()`.

HeikoKlare (commented):
Thank you for this adaptation. I have left some comments at the code/API level. Please also improve the commit message to explain the concrete intent, as both "the client" and "the new Browser API" are unspecified. The header could be something like "Add method for identifying location for custom content page in BrowserInformationControl". And add an explaining message of the change according to https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
amartya4256: I'll add braces

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
amartya4256: Done

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
amartya4256: Nope it was a mistake. Fixed it

laeubi:
```suggestion
	public boolean isCustomTextPage(String loc) {
```

The old name is utterly confusing without the javadoc and the discussions here I won't be able to even guess what this method is supposed for, maybe even rename for SWT as well?

Beside that, why can't one get acces to the fbrowser and ask it directly?

bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
```diff
@@ -465,7 +465,7 @@ public void changed(LocationEvent event) {
 					if (!event.top)
 						return;
 					if (combo != null) {
-						if (!"about:blank".equals(event.location)) { //$NON-NLS-1$
+						if (!browser.isLocationForCustomText(event.location)) {
```
laeubi, line 468: Maybe this should be part of the event?!

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
HeikoKlare: > The old name is utterly confusing without the javadoc and the discussions here I won't be able to even guess what this method is supposed for, maybe even rename for SWT as well?

I agree that the name needs to be changed (https://github.com/eclipse-platform/eclipse.platform.ui/pull/2211#discussion_r1746601006)
 
> Beside that, why can't one get acces to the fbrowser and ask it directly?

That would break encapsulation. The browser is, by intent, properly hidden as an implementation detail of the `BrowserInformationControl`. Allowing a user to access leads (1) to worse design (because of less encapsulation, more necessity to know about the internals to properly use it) and (2) introduces the risk of problems as the current implementation assumes that the browser is hidden and cannot change its state due to access by anyone else.

bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
HeikoKlare: You mean to have an additional flag in `LocationEvent` with this information? I like the idea from a design POV. And considering the existing PRs for adaptations to the new API, we might not even need a new method in the `BrowserInformationControl` for checking a location then, as this method is only used to validate a location in a location change event (since you cannot directly retrieve the URL from the browser, as you don't have access to that, anyway).

@amartya4256 what do you think?

amartya4256: I assume, this flag would be set at the creation of the LocationEvent. It means we will have to adapt this for all the browsers. Am I right?

sratz: All the browser implementations that create those `LocationEvent`s, yes.

I also like this idea.
We already have the `LocationEvent#top` flag, so putting a `isLocationForCustomText` flag there would be nice.
And discovering this flag in the `LocationEvent` directly is a lot more intuitive than discovering that there is a utility method in `Browser`

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
```diff
@@ -585,6 +586,20 @@ public BrowserInformationControlInput getInput() {
...
+	 * @param loc the location to be checked
+	 *
+	 * @return Returns whether the given location represents the location to which the this browser
+	 *         control has navigated via calling {@link #setContent(Object)}.
+	 *         <p>
+	 *         It may be used to validate in a location listener whether the changed location
+	 *         belongs to custom content or whether it belongs to some other location to which the
+	 *         user has navigated in the control).
+	 */
+	public boolean isLocationForCustomText(String loc) {
```
sratz, line 599: could be final

bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserViewer.java
laeubi: **But please** find a better name... :-)

I have no clue for what one needs to check for "custom text", but in this example it looks more the code wants to know if the browser is displaying a real URL (in contrast to something artificial), so maybe even `locationValid` or similar.

HeikoKlare: So let's have this PR on hold and try out the extension of the `LocationEvent`. That seems to reflect the intent best.

> **But please** find a better name... :-)

We are always open for good proposals :-) What about something like `showsCustomContent`? `locationValid` does not reflect the meaning, as it's not about validity, it is specifically about whether custom content (the Browser API uses the term "text", which is why it is currently called "custom text") is shown. An unresolvable URL or one not fitting a URL scheme may also be considered "invalid".

laeubi: > An unresolvable URL or one not fitting a URL scheme may also be considered "invalid".

But that might have the same effect as then the browser shows a custom browser error page maybe...

So in what cases one needs to check for "custom text"? For me it all seems related to if there is actually content shown that was fetched from an URI versus content shown that is provided by direct text (e.g in memory, but I can only guess this becomes important when I check the location URI somehow.

So to summarize, maybe it would have been better that `location` would have been an `Optional<String>` or at least `null` ... now we need a flag to tell the the user "the location is nothing useful as a real location" aka invalid.

So maybe one should better deprecate `location` and replace it with a real `Optional<URI> locationURI` making the whole checks and API method redundant?

HeikoKlare: > So maybe one should better deprecate `location` and replace it with a real `Optional<URI> locationURI` making the whole checks and API method redundant?

That could be a solution, yes. Still, it raises some questions for me:
1. Why is an Optional so much better than having an additional flag that it justifies the process of deprecation and removal? Of course, it is a means to avoid inconsistency, but at the same time you lose expressiveness (as no one knows what an empty option means without reading its documentation within its generation context. So instead of an optional, this should at least be a proper datatype exposing methods to retrieve the required information while hiding how the storage and retrieval is done.
2. If an empty optional means an invalid URL, this also means I will never be able to retrieve such invalid URLs, even if I was interested in them or even requires that information.
3. Most important: What would be your definition of invalid (i.e., an empty Optional)? I am not an expert on the client side, so I do not know about all the semantics of the clients when processing this information. The status quo is that clients usually check for `about:blank` to do specific things.  For example, JDT checks for `about:blank` (to identify custom content) and then has further matching for different URI schemes. So the definition would need to fit to all the existing consumers, which are currently quite restricted to matching `about:blank`.

amartya4256: I think changing the type of location would lead to other complications too. All the clients which are consuming it would have to adapt to it as well. Isn't it a bit risky?

laeubi: > Most important: What would be your definition of invalid (i.e., an empty Optional)?

An empty optional simply means that nor URI is displayed but "something else" ... I really think that "custom text" thing is confusing, no such thing exits in a browser.

A browser displays a _document_ it might be initially loaded from an _URL_ ... but if we enable JS this document can be completely replaced by something else.

That the SWT BrowserControl has a `setURL(...)` and `setText(...)` is an implementation detail but none of them is "custom" ... the difference is more that for the first the browser has (possibly e.g. see data urls!) fetch some (possibly remote) bytes and then decode them into a document. But this is also "custom" content, the browser can only use the former to resolve relative URIs while with `setText` it is not possible that easy.

That the location return `about:blank` seems more an implementation detail of the browser than a design decision of the browser control. So even if people do it to check _something_ it really checks if the page has `document.location.href` that points to some URL suitable to resolve relative URIs

HeikoKlare: I see most of the recent concerns related to the term "custom", and I understand that this is confusing, so it would be great if we can improve the according names. Indepedent from that, we have at least two use cases for locations and location change events that are or can currently be realized:
1. Retrieve the actual location from which the browser has loaded contents (for whatever reasons).
2. Identify for some location change event whether this is caused by a call to `setText()` for providing what we used to call "custom" text/content. To this end, clients compare "about:blank" against the location. That's what we have, e.g., seen in JDT: https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/1df6c00479ebd2cb7030ea92ae10a28112a8b367/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/viewsupport/JavaElementLinks.java#L601

If we provide some new API, it needs to allow clients to implement these use cases, and it should allow them to replace their "about:blank" comparisons. I do not want to challenge what these clients are doing (because there can be an arbitrary, unknown number of further clients that do the same and that we cannot change or investigate whether it would be possible to change them). An additional flag in the location event allows to implement these use cases. I still do not see how this would be possible by making the location optional and thus hiding the actual location in some cases (in addition to the other drawbacks I have mentioned).

> That the SWT BrowserControl has a `setURL(...)` and `setText(...)` is an implementation detail but none of them is "custom"

I think part of the problem is that this is not true. This is API of `Browser`, not any implementation detail. It is something that clients rely on. And they rely on the ability to investigate a location change event and identify if that was triggered by a call to `setText()`.

Don't get me wrong: It would be great to see a better than the proposed solution to address this, but I currently don't see it. In case there is one, I would like to see it along with the required adaptations of the clients.

> That the location return `about:blank` seems more an implementation detail of the browser than a design decision of the browser control.

Exatly. That's why we would like to hide this from clients and this implementation detail may change.

laeubi: > I think part of the problem is that this is not true. This is API of Browser, not any implementation detail. It is something that clients rely on. And they rely on the ability to investigate a location change event and identify if that was triggered by a call to setText().

You can browse `about:blank` it must not be `setText`, maybe **also** `setText` for some unknown reasons, result in the location value return `about:blank` but that has nothing to do with `setText`. Actually to have a locationchange event, the location must be **different** from the one before...

The more interesting question is, if I really need to want to know why not set a flag before calling setText... so I doubt ist really meant as an indication of setText is called... It might happen it works, but for me its undefined behavior unless the API explicitly state this  but it does not it only says:

> Renders a string containing HTML. The rendering of the content occurs asynchronously. 

Location listener says

> This listener interface may be implemented in order to receive a LocationEvent notification when a Browser navigates to a different URL.

so technically the URL could be empty maybe , null, or even about:blank if I call setText **once** but successive calls should *not* trigger a location change. 

Next we look at the event itself that says

> A LocationEvent is sent by a Browser to LocationListener's when the Browser navigates to a different URL. This notification typically occurs **when the application navigates to a new location with Browser.setUrl(String) or when the user activates a hyperlink**.

it says nothing about that I should exspect a locationchange on setText, nor that I can rely on any particular location set see:

```
       /**
	 * The URL of this event, escaped and encoded for consumption by
	 * {@link java.net.URI#URI(String)}.
	 */
	public String location;
```

so I agree that `about:blank` is an URI that can be consumed by `new URI()` but assuming that this tells me anything about if `setText("<html>/html>")` was called or `setUrl("about:blank")` is not the case even though it might happens as part of the implementation details of the browser.

HeikoKlare: Nothing in the previous post conflicts with the current proposal. No one has said that "about:blank" is a unique identifier for whether `setText()` was called. It is, of course, not. Still some clients (that I have shared) rely on the implication: call `setText()` => location "about:blank" (some then assuming a location change event, some only retrieving the URL). Whether or not the API currently guarantees this does not really matter, as we have clients relying on this behavior. With new API, we could even ensure equivalence between a specific location and `setText()` being called, by using a unique identifier instead of "about:blank", but that does not even matter.

I still don't see how this brings us forward. What is the concrete proposal? And how must this concrete proposal be adapted by clients?

laeubi: I have proposed that we need something to tell the user "this location is not a (regular/valid/usual/whatever) URI" you claimed that the user wants to know that `setText` is called. That is not the case. It might be a sideffect of `setText` call that a location change event is triggered with a location of about:blank, but this is nothing that is guaranteed so an API that states it returns `true` wen `setText` is called sounds wrong to me (and limit us in the implementation of modern browsers).

It still will work as before for the clients (and maybe must due to legacy reasons), but the API should simply state that the URI is probably a synthetic/invalid/unset/whatever, it should not state anything about "custom" or "text" at all... (my opinion of course) and give false expectation that it has anything to do with setText/setURL.

This is exactly what one calls "browser quirks" ...

amartya4256:
Had a look at how we could introduce a method for optional getUrl and all the other references which contain the URL. Overall, the idea was to get rid of about:blank text but while working on it I realized there are a lot of places where about:blank cannot be removed since it is referenced in jsp, bookmarks, where we don't have any reference of browser and also in many places. In my opinion, we use about:blank as a convention to display custom text all eclipse wide and since it is consistent for every platform, wecan keep it this way for now and keep using about:blank. In case of Edge, of course it uses a different URL, but it imitates that as about:blank since that is the convention. Hence, with this I am closing this PR.

PR #2213, opened 2024-08-22
Improve performance for large selection

Bananeweizen:
In a custom RCP application I noticed that selecting a huge amount of files (by holding Shift-Scroll down for some seconds) gets slower with the growing size of the selection.

We should short-circuit the evaluation the moment the overall result is known to be false.


jukzi (approved):
lgtm

Issue #2180, opened 2024-08-12
exclude NON-binary files from file search by file extension


jukzi:
i just saw a a slow file search where it tried to search in *.png files
I would say *.png files should be considered as binary. or not?
![image](https://github.com/user-attachments/assets/0a710d15-f868-4942-8f8f-a4f7c6f0e0d6)

Wittmaxi:
I don't think a user would actively search for a string inside of a png-file. The only case I could see myself doing that is, for example, when looking for some metadata in the file headers - which is treating the png as a binary file.

Wittmaxi:
Intuitively, I would say that a fix for this can be rather easy

Wittmaxi:
The relevant file seems to be TextSearchVisitor in method `hasBinaryMethod`.
However, this method only checks for this:
1) are we a text file? -> return false
2) does the file contain "\0" somewhere? -> return true

I see multiple problems with this approach
1) we are relying on nullbytes being present in the file to see if it is a Binary
2) we are scanning for a very long time (in fact, most binary files where we aren't 100% sure will be fully scanned through JUST to see if they should be excluded from file search !!!!!!!!!!!!). This certainly has a huge speed impact and is performed even if we wanted to search in binary files

Strikingly, when adding 0-bytes into a genuine text-file, it was automatically considered a binary and never searched through.

A better solution (which would be a much greater overhaul, I'm afraid) would be to store which file-extensions or file headers are to be considered binary and which ones are not to be consided binary.

I am wondering, do you mind sharing some PNG that is eronneously scanned even though it shouldn't be? I have just added my pictures folder into the workspace and searched for some common file signatures and some other string sequences which happened to be in the hexdump of my image files. I could (correctly) not find any of those in the png-files.

laeubi:
> we are relying on nullbytes being present in the file to see if it is a Binary

This is a very common approach to check if a file is a binary one if no other measure is available because the Zero Byte is something that very often occurs in binary numbers and is the end of string character in C strings.

As you mentioned this means the file still needs to be scanned **and might appear in the log output**, but the important part is that it is not searched for the text-string itself.

> A better solution (which would be a much greater overhaul, I'm afraid) would be to store which file-extensions

File extensions are a very worse thing to qualify file types, but if one want one can already exclude with `!.*png` in the search box

> headers are to be considered binary

[...]

PR #2215, opened 2024-08-23
TextSearchVisitor: don't check for binary files if not required

Wittmaxi:
If a workspace search is allowed to search in binary files, we should not perform a (possibly expensive) check for whether a file is binary - we are going to search through it regardless of the result of that query.

As discussed in
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2180

Note: the performance increase is not yet optimal - we still load the file to ram, the difference is that now, we don't scan through the file for zero-bytes. A "good" solution would recognize file extensions and not load those files to begin with.

### DIFF
--- a/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -206,7 +206,7 @@ public class TextSearchVisitor {
 				} else {
 					try {
 						charsequence = fileCharSequenceProvider.newCharSequence(file);
-						if (hasBinaryContent(charsequence, file) && !fCollector.reportBinaryFile(file)) {
+						if (!fCollector.reportBinaryFile(file) && hasBinaryContent(charsequence, file)) {
 							return Status.OK_STATUS;
 						}
 						occurences = locateMatches(file, charsequence, matcher, monitor);


jukzi:
> A "good" solution would recognize file extensions and not load those files to begin with.

performance optimizations should show the benefit by measurement, and please lets go for a "good" solution - opening a file is typically the costly thing while reading the content is almost not noticeable if its in the OS cache.

laeubi (approved):
Looks good to me, this is obviously an improvement as the check for binary file is the one that most likely more costly than that of asking the collector if it wants binary content.

Wittmaxi:
@jukzi my current focus is finishing up the tasks in the Umbrella issue for the Find/Replace overlay. I understand that this is more of a "low hanging fruit"-PR, I might be able to come up with a more substantial improvement in a few months - and yet, I can't commit to anything at this point.

If this is OK for you, I will merge this PR once master opens up again

jukzi (changes requested):
If you can provide a measurement that shows the benefit i am good with it. But please never ever merge a performance improvement without a measurement. We have way too many regressions by such well-intentioned commits having bad side effect.

My guess is `fCollector.reportBinaryFile(file)` will return false. i.e. the inverse of that is true i.e. you do not change anything with this commit.

laeubi:
@jukzi please read the PR description for what case this is meant:

> If a workspace search is allowed to search in binary files,

So this is for the case where searching in binary files is **enabled**, meaning even if we want to search in binary files (currently second check!) it is first check if it is a binary file just to the discard the result because we want to search in *Every file* regardless if binary or not...

If binary search is **disabled** it doesn't matter much as you pointed out.

Wittmaxi:
I cannot take ownership of this anymore. I believe I have seen another PR which fixed this already, so this PR is obsolete

Issue #2218, opened 2024-08-26
Disable UI Freeze Error log by default


jukzi:
When something takes long it is contra productive that the error log pops up and robs the focus, disturbing my work even more. It should do so on real errors, but some things simply take long. Therefore by default the error log should only pop up on errors and freezes should not be reported as errors.

![image](https://github.com/user-attachments/assets/171cc9f1-1bda-4dbd-bab0-31882c719b1c)

Downgrading freezes error/warning to warnings/info does not work so good with public API likeorg.eclipse.ui.monitoring.PreferenceConstants.LONG_EVENT_ERROR_THRESHOLD_MILLIS

WDYT?

vogella:
IMHO this change is fine. The period in which I hoped that all long operations will run out of the UI thread has passed, looks like this expectation was unrealistic.

I would not do this shortly before the release , better to wait for a new release cycle.

PR #2220, opened 2024-08-26
Error Log: do not pop up on info/warning/freeze by default #2218

jukzi:
As it is counterproductive to pop up error log when something took too long.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2218
image
image

### DIFF
--- a/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/preferences/MonitoringPreferenceInitializer.java
+++ b/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/preferences/MonitoringPreferenceInitializer.java
@@ -30,7 +30,7 @@ public class MonitoringPreferenceInitializer extends AbstractPreferenceInitializ
 
 		store.setDefault(PreferenceConstants.MONITORING_ENABLED, false);
 		store.setDefault(PreferenceConstants.LONG_EVENT_WARNING_THRESHOLD_MILLIS, 500); // 0.5 sec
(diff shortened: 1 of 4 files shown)


vogella:
See issue for discussion, tests should be updated/ fixed

bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/preferences/MonitoringPreferenceInitializer.java
```diff
@@ -30,7 +30,7 @@ public void initializeDefaultPreferences() {
 
 		store.setDefault(PreferenceConstants.MONITORING_ENABLED, false);
 		store.setDefault(PreferenceConstants.LONG_EVENT_WARNING_THRESHOLD_MILLIS, 500); // 0.5 sec
-		store.setDefault(PreferenceConstants.LONG_EVENT_ERROR_THRESHOLD_MILLIS, 1000); // 2 sec
+		store.setDefault(PreferenceConstants.LONG_EVENT_ERROR_THRESHOLD_MILLIS, 100_000); // 100 sec
```
Wittmaxi, line 33: So we still pop up a warning on freeze, except we allow freezes to be much longer before we pop up an info?

I think that's an okay option, especially since it is relatively easy to implement compared to it's revenue (i.e. you are not always annoyed by a popup on smaller freezes)

jukzi: We will still **log** a warning, but **activate** the error view only for 100s lags (i.e. almost never)

PR #2221, opened 2024-08-26
search: fail fast for binary file types without opening the file #2180

jukzi:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2180

assumes that files are binary if they have a IContentType that is not text


jukzi:
Measurement result:
Platform workspace, searching "utf" in "*" - but not in binary files:
Without this change:
Job "Searching '"utf8"'" took 3561ms
With this change:
Job "Searching '"utf8"'" took 2149ms

=> more the 1.5 times faster

jukzi:
The trick here is that this change avoids evaluating IContentDescription - which did access the file:
image

jukzi:
now `findContentTypesFor()` has not optimal performance but at least it does not open the file:
image
maybe one could further speed up `org.eclipse.core.internal.content.ContentTypeCatalog.internalFindContentTypesSorted ()`

Issue #2225, opened 2024-08-27
workspace path not writable on a smb home share


andy1805:
Since the release 2024-03 we cant use the eclipse IDE for our exams because the students have to save the workspace on their home share and that's not more possible. The last release that worked was 2023-12. 

> Quote Ed Merks from eclipse forum: The logic was changed during the 2024-03 release cycle and now calls this method to test:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/e3628a3a67c44771465d7d317b3424920c425090/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java#L491-L504

In our system the exam software opens a folder on the students home share, where they have to save their workspace, and only files under this folder will be transferred into the teachers home share.
I have no idea about java, I only have to install eclipse in a way that's usable for exams. I don't know why the logic have be changed and why it recognize the path as not writable because it is. I can only imagine that it checks if the local user has write rights and they log in via Samba User.
![EclipseFehler](https://github.com/user-attachments/assets/afb903bf-2fa6-4024-ab95-325844ab9642)

merks:
What probably works as a workaround is to specify the workspace on the command line:

eclipse.exe -data L:\eclipse_2024-06

Can you try if that works?

andy1805:
Hi Ed,

thanks for your help. 

Yes, you're absolutely right. I already tested this workaround and it works. But how i wrote before, I don't know before the exam, where exactly the workspace have to be saved. So for example the students have to change the shortcut of eclipse.exe with "-data L:\Klassenarbeiten\M122-LB2-INP23C\eclipse_2024-06" (M122 is the modul, lb2 means "Leistungsbeurteilung 2" and the class name "INP23B") and this every time again, because it's always another path.

That's not practible at all. I can use this way for normal work but not for exams.

jukzi:
`The path is not writable by the current user` was introduced with https://github.com/eclipse-platform/eclipse.platform.ui/pull/1522. @laeubi  can you please take a look?

laeubi:
> `The path is not writable by the current user` was introduced with #1522. @laeubi can you please take a look?

No as I don't have any SMB setup like described by the user, and as verified with @merks the location obviously *is* writable so this is either a Windows Bug or JDK Bug but nothing that can be fixed by platform without further information or access to this special setup.

merks:
[...]

PR #2226, opened 2024-08-28
fix workspace considered not writable on a smb home share #2225

jukzi:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2225

java.io.File.canWrite() and java.nio.file.Files.isWritable(Path) can not be trusted on windows. they may return wrong values. see for example JDK-8282720, JDK-8148211, JDK-8154915

=> allow the user to press "Launch" even if jdk states the directory is not writable

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
@@ -474,9 +474,6 @@ public class ChooseWorkspaceDialog extends TitleAreaDialog {
 			}
 			String normalisedPath = path.normalize().toString();
 			String normalisedPathWithSeperator = normalisedPath + File.separator;
-			if (!isWritable(path)) {
-				return NLS.bind(IDEWorkbenchMessages.ChooseWorkspaceDialog_NotWriteablePathWarning, normalisedPath);
-			}
 			if (normalisedPathWithSeperator.contains(TILDE)) {
 				return NLS.bind(IDEWorkbenchMessages.ChooseWorkspaceDialog_TildeNonExpandedWarning, normalisedPath);
 			}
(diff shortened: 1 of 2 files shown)


jukzi:
image

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/messages.properties
```diff
@@ -1049,7 +1049,7 @@ ChooseWorkspaceDialog_recentWorkspaces=&Recent Workspaces
 ChooseWorkspaceDialog_ResolvedAbsolutePath=Full path: {0} 
 ChooseWorkspaceDialog_TildeNonExpandedWarning=\u26A0\uFE0F '~' is not expanded, full path: {0}
 ChooseWorkspaceDialog_InvalidPathWarning=\u26A0\uFE0F The path is invalid on this system: {0}
-ChooseWorkspaceDialog_NotWriteablePathWarning=\u26A0\uFE0F The path is not writable by the current user: {0} 
+ChooseWorkspaceDialog_NotWriteablePathWarning=\u26A0\uFE0F The path may not writable by the current user: {0}
```
merks:
```suggestion
ChooseWorkspaceDialog_NotWriteablePathWarning=\u26A0\uFE0F The path may not be writable by the current user: {0} 
```

iloveeclipse: `The path may not **be** writable by the current user: {0}`

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
```diff
@@ -474,33 +474,37 @@ protected String getUnexpectedPathHint() {
...
-	 * @return
-	 */
-	private boolean isWritable(Path path) {
-		if (Files.exists(path)) {
-			return Files.isWritable(path);
-		}
-		Path parent = path.getParent();
-		if (parent != null) {
-			return isWritable(parent);
+	private boolean mayBeWritable(Path path) {
```
merks: I find the name a little awkward.  English you have may be and maybe and they have subtly different mean, allowed to be versus possibly.  In the end, if this method returns true, the file is definitely writable I think, right?  I think all the bugs are this returning false when it should return true. 

I would just leave the method as isWritable with the description of it being unreliable...

merks (changes requested):
This is similar to what I do in the installer where I warn the user the location isn't writable but I don't prevent them from trying to use it anyway.  So I think that's good approach.

merks:
@akurtakov 

I think when this finds closure today, we should get this into RC2.

laeubi (approved):
I'm fine with what ever is assumed suitable as I didn't introduce this check as a "hard check" I just improved once an existing implementation in terms of layout and cover of error cases.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceDialog.java
jukzi: depending on how the method is implemented (io/nio) the returned value did show wrong values in both directions. It is unreliable and should reflect that with the name. Changed to "maybe"

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/messages.properties
jukzi: thanks

merks (approved):

jukzi:
Please formally state that that you do the approval as PMC decision so that it can be merged in the current phase.

akurtakov:
Assuming multiple people test and verify I'm fine with this late fix.

merks:
I've tested this myself, blanking out the location so that I get a prompt:

image

It works well for me.

PMC +1

jukzi:
ignoring unrelated error `ExpressionContext illegally implements IEvaluationContext`

PR #2228, opened 2024-08-29
Resolve problem markers "Unlikely argument type for equals()"

jukzi:
trick the compiler to ignore false positives:
* IHyperlinkSegment is related to ParagraphSegment
* ITabbedPropertySheetPageContributor is related to IWorkbenchPart

as they have common implementations.

### DIFF
--- a/bundles/org.eclipse.ui.views.properties.tabbed/src/org/eclipse/ui/views/properties/tabbed/TabbedPropertySheetPage.java
+++ b/bundles/org.eclipse.ui.views.properties.tabbed/src/org/eclipse/ui/views/properties/tabbed/TabbedPropertySheetPage.java
@@ -315,8 +315,8 @@ public class TabbedPropertySheetPage
 		 * The properties view has been activated and the current page is this
 		 * instance of TabbedPropertySheetPage
 		 */
-		boolean thisActivated = part instanceof PropertySheet
-			&& ((PropertySheet) part).getCurrentPage() == this;
+		boolean thisActivated = part instanceof PropertySheet propertySheet
+			&& propertySheet.getCurrentPage() == this;
 
 		/*
 		 * When the active part changes and the part does not provide a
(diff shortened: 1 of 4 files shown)


laeubi:
> trick the compiler to ignore false positives
> as they have common implementations.

I actually think this is correct (and better way) and not "tricking" the compiler who can't know that :+1:

jukzi:
the compiler could/should show the same warning for an unlikely instanceof.
For example `new Object() instanceof MyInterface` is not only unlikely but even impossible.

laeubi:
> the compiler could/should show the same warning for an unlikely instanceof. For example `new Object() instanceof MyInterface` is not only unlikely but even impossible.

That's a completely different topic. In the case you presented here, the compiler has no doubt of the instance of the object (and not only its possible types).

In the case in the PR the argument is a parameter and its an interface, so if have an argument of type `Object` the compiler can never know if it *might* implement the interface.

So here we have an (abstract) class (ParagraphSegment) that do not implement the interface `IHyperlinkSegment` and due to uses **the default contract of equals/hashcode** and there is no guarantee that all `ParagraphSegment` extensions are implement `IHyperlinkSegment` and only those are then passed to the method, so this equals is indeed unlikely (but not impossible) to be what one wants. Instead for example an identity check or a comparison of properties might be a better choice.

So with your additional check you now ensure that at least the implementation is a `ParagraphSegment` and there is a higher chance for the objets to be actually compatible and therefore no warning is emitted.

bundles/org.eclipse.ui.views.properties.tabbed/src/org/eclipse/ui/views/properties/tabbed/TabbedPropertySheetPage.java
```diff
@@ -325,16 +325,15 @@ protected void handlePartActivated(IWorkbenchPart part) {
...
 		 */
-		if (!thisActivated && !part.equals(contributor)
+		if (!thisActivated && !(part instanceof ITabbedPropertySheetPageContributor p && p.equals(contributor))
 				&& !part.getSite().getId().equals(contributor.getContributorId())) {
 			/*
 			 * Is the part is a IContributedContentsView for the contributor,
 			 * for example, outline view.
 			 */
 			IContributedContentsView view = Adapters.adapt(part, IContributedContentsView.class);
-			if (view == null
```
HannesWell, line 335: It looks like removing the `view ==null` case is causing a lot of NPEs in certain situations:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2275

HannesWell: And I don't know if the case `view.getContributingPart() == null` is relevant, but before if that was true the if's condition was false, now it is true.

jukzi: it was != null condition.
cp instanceof Something=> cp!=null

HannesWell: > cp instanceof Something=> cp!=null

That's right, but the problem is the negation is now applied to the result of the `&&` and not to its RHS.
Instead of 
`!(view.getContributingPart() instanceof ITabbedPropertySheetPageContributor cp && cp.equals(contributor))`
it should probably be
`view.getContributingPart() instanceof ITabbedPropertySheetPageContributor cp && !cp.equals(contributor)`

PR #2229, opened 2024-08-29
AbstractTextEditor: resolve "Potential null pointer access"

jukzi:
"The variable nextAnnotationPosition may be null at this location"

The code assumed "distance=Integer.MAX_VALUE" will cause short circuit "currentDistance < distance" while theoretically equality could happen with currentDistance=Integer.MAX_VALUE.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -6998,7 +6998,8 @@ public abstract class AbstractTextEditor extends EditorPart implements ITextEdit
 					if (currentDistance < 0)
 						currentDistance= endOfDocument + currentDistance;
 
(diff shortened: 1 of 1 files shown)


jukzi:
will ignore unrelated "baseline 'workspace' is disposed" https://github.com/eclipse-pde/eclipse.pde/issues/145

HeikoKlare:
> will ignore unrelated "baseline 'workspace' is disposed" [eclipse-pde/eclipse.pde#145](https://github.com/eclipse-pde/eclipse.pde/issues/145)

In my opinion, this kind of issue should not be ignored, because it makes the build fail immediately when it occurs. A bunch of plug-ins may not be built/tested. So the results are incomplete for the affected OS. When this occurs, we should retrigger the build for that OS unless we are absolutely sure that the PR may not affect OSes differently, so that we can rely on the results of the builds for the other OSes.

jukzi:
i agreee and normally do rebuild. but the rebuild button was not available when i looked and this change is impossible to be OS depended

jukzi:
ignoring random failing testSingleSelectionAndTwoInitialSelectionsWithInitialPattern https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #2232, opened 2024-09-02
SearchHistoryMenu: improve scrolling through selection with arrow keys

### LINKED ISSUE #2139: Find/replace overlay: history drop-down navigates to second entry when pressing "down" key

## Steps to reproduce

* Open any text editor and press CTRL+F to open the find/replace overlay
* Perform at least two explicit search operations (input + ENTER) to have the two entries in the history
* Open the drop-down menu for the search input field
* Press the "down" arrow key

I expected: the first entry in the list to be selected

But got: the second entry in the list being selected

Reported by @HannesWell: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2136#issuecomment-2257128241

* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): I20240728-1800

## Community

Wittmaxi:
Scrolling through the selection of the Search History now correctly starts at the first item and cycles through the boundaries (ie. scrolling down from the last item returns correctly to the first item of the list)

fixes #2139

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
@@ -54,6 +54,7 @@ class SearchHistoryMenu extends Dialog {
 	private int width;
 	private Table table;
 	private TableColumn column;
+	private int selectedIndexInTable = -1;
 
 	public SearchHistoryMenu(Shell parent, HistoryStore history, Consumer<String> historyEntrySelectedCallback) {
 		super(parent);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
```diff
@@ -93,6 +94,36 @@ public Control createContents(Composite parent) {
 	}
 
 	private void attachTableListeners() {
+		table.addKeyListener(new KeyListener() {
```
HeikoKlare: You can use KeyAdapter to avoid empty method implementations

HeikoKlare (commented):
Wouldn't it also be an option to set the selection to the first (or even to the currently selected, if existing) entry in the table? Then you would always have a selection and do not require this additional functionality replicating the complete navigation functionality.

Note that in usual dropdowns, hovering over other elements also moves the selection to these entries, so that the keyboard navigation will proceed from the entry you have hovered over last.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
Wittmaxi: I knew this comment might come but chose to stick to a regular "KeyListener" The reason is twofold:

1) I have state which I don't want to manage in the outer class (i.e. indexInTable). In my view, that is responsibility of the keylistener which manages where to scoll to
2) I also want to encapsulate `moveSelectionInTable`

What do you think?

HeikoKlare: It makes sense to have both within the listener, so that's fine for me. But why wouldn't that be possible with a KeyAdapter? Only different should be that you can avoid empty method implementations, but you still have an anonymous class in which you can encapsulate the things you have mentioned.

Wittmaxi:
The hover is now sticky such that the navigation with the arrow keys resumes from the position where the last hover happened.
This also meant that some state (the scrol positoin) had to be moved into the entire Search History Menu instead of in the key listener (see comment above).

That way, I could fix the comment about using the adapter for the listener on the fly.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
```diff
@@ -92,7 +93,35 @@ public Control createContents(Composite parent) {
...
 	private void attachTableListeners() {
+		table.addListener(SWT.MouseMove, event -> {
+			Point point = new Point(event.x, event.y);
+			TableItem item = table.getItem(point);
+			if (item != null) {
+				table.setSelection(item);
+				selectedIndexInTable = table.getSelectionIndex();
+			}
+		});
+		table.addKeyListener(KeyListener.keyPressedAdapter(e -> {
```
HeikoKlare, line 116: nitpicky: The same method already registered another KeyListener at the table. Would it make sense to combine them? And maybe switch over the `e.keyCode` in case of having multiple if/else clauses.

HeikoKlare (commented):
The change looks okay and works as expected. I only have a nitpicky proposal for improvement.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
Wittmaxi: Yes, it makes sense to combine them, good catch!

https://github.com/eclipse-platform/eclipse.platform.ui/commit/b6186f57fbec28d23260fc3a2f3d1101b910c3c5

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/SearchHistoryMenu.java
Wittmaxi: Fixed in https://github.com/eclipse-platform/eclipse.platform.ui/commit/b6186f57fbec28d23260fc3a2f3d1101b910c3c5
see comment down below

Wittmaxi:
@HeikoKlare if this is fine with you, I will merge by end of day

deepika-u:
Verified on the below build applying the pr and i am able to select 1st entry when using down arrow key on Find/replace overlay.

Eclipse SDK
Version: 2024-12 (4.34)
Build id: I20240923-1800
OS: Windows 11, v.10.0, x86_64 / win32
Java vendor: Eclipse Adoptium
Java runtime version: 21.0.1+12-LTS
Java version: 21.0.1

Issue #2099, opened 2024-07-18
Find/Replace Overlay: can hide views which are in "detached mode"


Wittmaxi:
![grafik](https://github.com/user-attachments/assets/d816808a-9b62-4b8f-ba06-5b5f9acc8083)

Wittmaxi:
Regression since https://github.com/eclipse-platform/eclipse.platform.ui/commit/a5689de5d997f9fc8c40bd92af32a800f285b1ae

Issue #2093, opened 2024-07-16
Example handler for an alternative inline display


vogella:
@HeikoKlare and @Wittmaxi while cleanup up my inbox I found this example code for showing something inline of the current editor. I wanted to share it with you so that you are aware of this option. No error check are included in the example code, it is just a demo which might help you for future developments.

Result: 

![image](https://github.com/user-attachments/assets/afa47e56-d13c-46a1-985d-17dd067d9d2c)

````
package handler.handlers;

import java.util.List;

import org.eclipse.core.commands.AbstractHandler;
import org.eclipse.core.commands.ExecutionEvent;
import org.eclipse.core.commands.ExecutionException;
import org.eclipse.e4.ui.model.application.MApplication;
import org.eclipse.e4.ui.model.application.ui.advanced.MArea;
import org.eclipse.e4.ui.model.application.ui.basic.MPartStack;
import org.eclipse.e4.ui.model.application.ui.basic.MWindow;
import org.eclipse.e4.ui.workbench.modeling.EModelService;
import org.eclipse.e4.ui.workbench.modeling.EPartService;
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Rectangle;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Composite;
import org.eclipse.swt.widgets.Control;
import org.eclipse.ui.IWorkbenchWindow;
import org.eclipse.ui.handlers.HandlerUtil;

public class SampleHandler extends AbstractHandler {
	
	
	MPartStack pipStack = null;
	Composite pipComp = null;

	@Override
	public Object execute(ExecutionEvent event) throws ExecutionException {
		IWorkbenchWindow window = HandlerUtil
				.getActiveWorkbenchWindowChecked(event);
		MApplication theApp = (MApplication) window
				.getService(MApplication.class);
		MWindow win1 = theApp.getChildren().get(0);
		EPartService ps = win1.getContext().get(EPartService.class);
		EModelService ms = win1.getContext().get(EModelService.class);

		if (pipStack == null) {
			// Find the MArea
			List<MArea> areas = ms.findElements(win1, null, MArea.class, null);
			if (areas.size() > 0) {
				MArea area = areas.get(0);
				List<MPartStack> areaStacks = ms.findElements(area, null,
						MPartStack.class, null);
				if (areaStacks.size()==1) {
					pipStack =	ms.createModelElement(MPartStack.class);
[...]

PR #2234, opened 2024-09-02
EXPERIMENT: display the Find/Replace overlay as partstack

Wittmaxi:
Experiment with display the Find/Replace Overlay as a Part in a Partstack instead of an overlay shell.

Caveats:
- the part has the annoying tendency to stick to the side instead of allowing to be "positionned to the top all the time"
- keypresses are still interpreted as keypresses targeted towards the Editor behind (ie. pressing Ctrl+Backspace delets a full word in the editor not in the overlay)

First time opening the overlay (something is wrong...)
grafik

Second time pressing Cltr+F
grafik

Adapted from https://github.com/eclipse-platform/eclipse.platform.ui/issues/2093

I do not believe that this solution can bring improvement to #2099


HeikoKlare:
Closing as superceded by #2254.

PR #2235, opened 2024-09-02
FindReplaceOverlay: add support from mnemonics from the Dialog

### LINKED ISSUE #2080: Users of the Find/Replace dialog want to use Alt + A to perform "Search all"
Some users are used to pressing "Alt+A" to perform search all which was a mnemonic used in the FindReplaceDialog.
Those users should be able to press Alt+A in the Overlay as well so that they don't need to "learn" new key combinations

This issue was originally brought up by user u/n4te on [reddit](https://www.reddit.com/r/eclipse/comments/1dle5ts/new_find_dialog/)

Wittmaxi:
In order to preserve the workflow of users who use the Mnemonics of the Find/Replace Dialog a lot, add support to these key combinations as well

fixes #2080

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -82,18 +82,22 @@ public class FindReplaceOverlay {
 		private static final List<KeyStroke> SEARCH_BACKWARD = List.of( //
 				KeyStroke.getInstance(SWT.SHIFT, SWT.CR), KeyStroke.getInstance(SWT.SHIFT, SWT.KEYPAD_CR));
 		private static final List<KeyStroke> SEARCH_ALL = List.of( //
+				KeyStroke.getInstance(SWT.MOD1, SWT.CR), KeyStroke.getInstance(SWT.MOD1, SWT.KEYPAD_CR),
+				KeyStroke.getInstance(SWT.ALT, 's'));
+		private static final List<KeyStroke> REPLACE_ALL = List.of( //
 				KeyStroke.getInstance(SWT.MOD1, SWT.CR), KeyStroke.getInstance(SWT.MOD1, SWT.KEYPAD_CR));
 		private static final List<KeyStroke> OPTION_CASE_SENSITIVE = List.of( //
-				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'C'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'c'));
+				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'C'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'c'),
+				KeyStroke.getInstance(SWT.ALT, 'c'));
 		private static final List<KeyStroke> OPTION_WHOLE_WORD = List.of( //
 				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'D'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'd'));
 		private static final List<KeyStroke> OPTION_REGEX = List.of( //
 				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'P'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'p'));
 		private static final List<KeyStroke> OPTION_SEARCH_IN_SELECTION = List.of( //
-				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'I'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'i'));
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -80,19 +80,28 @@
 public class FindReplaceOverlay extends Dialog {
 	private final class KeyboardShortcuts {
 		private static final List<KeyStroke> SEARCH_FORWARD = List.of( //
-				KeyStroke.getInstance(SWT.CR), KeyStroke.getInstance(SWT.KEYPAD_CR));
+				KeyStroke.getInstance(SWT.CR), KeyStroke.getInstance(SWT.KEYPAD_CR),
+				KeyStroke.getInstance(SWT.ALT, 'n'));
```
HeikoKlare: In the dialog, all the added bindings also work when having the shift key pressed. Maybe an additional KeyStroke instance should be added for that combination as well?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -80,19 +80,28 @@
...
+				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'C'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'c'),
+				KeyStroke.getInstance(SWT.ALT, 'c'));
 		private static final List<KeyStroke> OPTION_WHOLE_WORD = List.of( //
-				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'W'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'w'));
+				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'W'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'w'),
+				KeyStroke.getInstance(SWT.ALT, 'w'));
 		private static final List<KeyStroke> OPTION_REGEX = List.of( //
-				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'P'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'p'));
+				KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'P'), KeyStroke.getInstance(SWT.MOD1 | SWT.SHIFT, 'p'),
+				KeyStroke.getInstance(SWT.ALT, 'X'));
```
HeikoKlare: This needs to be 'x' instead of 'X'.

HeikoKlare (commented):
In general, it's great to have the existing mnemonics of the dialog supported in the overlay as well. My only concern is that they derivate from existing behavior in the sense that for the search options/action they are only available when the search input field has focus, while the replace actions are only available when the replace input field has focus. So when the focus is anywhere else (any toolbar item or the other input field), the shortcuts are not available. This was already a limitation before, but it somehow becomes worse when readding the existing mnemonics, as they were dialog-unique and accessible from _everywhere_ in the dialog. Maybe we should extend the design to also support that case?

Wittmaxi:
After merging into the FindReplaceOverlay which is drawn as a composite inside of the page, some of the "alt+..." keybindings don't work anymore, and trigger a global action instead (e.g. alt+s opens "source" from the menu bar at the top).
Some mnemonics still work.

I believe this is a reason to drop this feature all together.
1) we don't communicate to any user which mnemonics exists, this is merely a help to old users who are used to the old mnemonics
2) choosing new mnemonics would not help in our case since we don't communicate to the user which mnemonics they can expect to work

This is the table of mnemonics which work:

| Action                  | Has Mnemonic? | Mnemonic Key(s)   |
|-------------------------|---------------|-------------------|
| SEARCH_FORWARD           | No            | -                 |
| SEARCH_BACKWARD          | No            | -                 |
| SEARCH_ALL               | Yes           | ALT + S           |
| REPLACE_ALL              | No            | -                 |
| OPTION_CASE_SENSITIVE    | Yes           | ALT + C           |
| OPTION_WHOLE_WORD        | No            | -                 |
| OPTION_REGEX             | No            | -                 |
| OPTION_SEARCH_IN_SELECTION | Yes         | ALT + L           |
| CLOSE                    | No            | -                 |
| TOGGLE_REPLACE           | No            | -                 |

I have updated my PR with a subset of all the mnemonics - every registered mnemonic still works.

HeikoKlare:
> I believe this is a reason to drop this feature all together.

I agree with this. In particular, now the overlay is not a separate dialog with its completely independent shortcut processing anymore. Shortcuts defined for the editor scope will apply, such that depending on the individual configuration different shortcuts may be conflicting. E.g., configuring some editor action for shortcut ALT+C would lead to a conflict with the support for the find/replace dialog mnemonic to toggle case-sensitive search. This also demands for making even the current shortcuts for the overlay configurable so that conflicting shortcuts are detected/avoided.
In summary, I propose to not provide any implicit shortcuts just for "backwards compatibliity" but to only support the explicitly defined shortcuts that are also mentioned in the tooltips (and make them configurable).

Wittmaxi:
@HeikoKlare in this case, I am now closing this PR

PR #2239, opened 2024-09-04
reduce some "Potential resource leak: '' may not be closed"

jukzi:


jukzi:
image

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CopyFilesAndFoldersOperation.java
```diff
@@ -528,7 +530,11 @@ private void copyExisting(IResource source, IResource existing, IProgressMonitor
 			IFile sourceFile = getFile(source);
 
 			if (sourceFile != null) {
-				existingFile.setContents(sourceFile.getContents(), IResource.KEEP_HISTORY, subMonitor.split(1));
+				try (InputStream contents = sourceFile.getContents()) {
+					existingFile.setContents(contents, IResource.KEEP_HISTORY, subMonitor.split(1));
+				} catch (IOException closeException) {
+					// never happens
```
HannesWell, line 536: Can't we use `IFile.copy()` in this case? 
Would be shorter and can be faster.

jukzi: does it have a flag to override existing resource?

merks: https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardCopyOption.html#REPLACE_EXISTING

jukzi: That is not a parameter of IFile.copy

merks: Sorry, I read it on my IPhone as Files.copy.

HannesWell: > does it have a flag to override existing resource?

Doesn't look like it has one yet. But it would definitely be worth adding it.

PR #2240, opened 2024-09-04
fix IFindReplaceLogic.java:184: error: unexpected end tag: </p>

jukzi:
reported in jenkins build log:
 [ERROR] MavenReportException: Error while generating Javadoc

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
@@ -181,7 +181,6 @@ public interface IFindReplaceLogic {
 	 * <li>Backward search operations will use the end of the selection as the
 	 * starting point.</li>
 	 * </ul>
-	 * </p>
 	 */
 	void resetIncrementalBaseLocation();
 
--- a/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor; singleton:=true
-Bundle-Version: 3.18.0.qualifier
+Bundle-Version: 3.18.100.qualifier
 Bundle-Activator: org.eclipse.ui.internal.texteditor.TextEditorPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
```diff
@@ -181,7 +181,6 @@ public interface IFindReplaceLogic {
 	 * <li>Backward search operations will use the end of the selection as the
 	 * starting point.</li>
 	 * </ul>
-	 * </p>
```
Wittmaxi, line 184: Didn't know that javadocs doesn't require closing tags. Looks sound to me :)

Wittmaxi (approved):

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/IFindReplaceLogic.java
laeubi: > Didn't know that javadocs doesn't require closing tags. Looks sound to me :)

That's wrong! The closing tag is just misplaced and implicitly closed by the `<ul>` tag already.

So the correct one (without relying on implicit close) would be:
```
* <p>
* The "current selection" refers to the range of text that is currently
* highlighted or selected within the FindReplaceTarget. This selection can be
* either a single position (if no range is selected) or a range of text.
*
* </p>
* <p>
* When handling range selections:
* </p>
* <ul> 
* .....
```

if you want to have two paragraphs...

jukzi: `Some HTML element types allow authors to omit end tags (e.g., the P`
https://stackoverflow.com/a/33787857/9549750

laeubi: As mentioned this does not mean that the paragraph is now closed where previously the `</p>` was placed it is already closed before (implicitly), this is even an example here https://html.spec.whatwg.org/multipage/grouping-content.html

> List elements (in particular, [ol](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element) and [ul](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element) elements) cannot be children of [p](https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element) elements.

And further down

> Authors wishing to conveniently style such "logical" paragraphs consisting of multiple "structural" paragraphs can use the [div](https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element) element instead of the [p](https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element) element.

So it is not true that it is _not required_ here (e.g. optional) is _simply wrong_ (and that's what javac complains about) because at that point _there is no open paragraph_ anymore that can be closed (due to the list element).

See also:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1858

merks: It seems best and never wrong to balance them, keeping in mind that ol and ul can’t nest in p.

PR #2247, opened 2024-09-06
Find/replace overlay: unify async executions, avoid NPEs #2244

### LINKED ISSUE #2244: Find/replace overlay: logs NPE
I repeatedly get NPEs in the log when using the find/replace overlay on Linux (X11), (XFCE / Xubuntu 22.04)

```
!ENTRY org.eclipse.ui 4 0 2024-09-05 10:39:45.524
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.Shell.getDisplay()" because the return value of "org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.getShell()" is null
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.updatePlacementAndVisibility(FindReplaceOverlay.java:848)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5040)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4520)
```

They are all the same, getShell() retuning null at this line:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/31a902749e53ae095fb98c40459ba8cfc70d6d0d/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java#L848

I don't have a way to reliably reproduce, but it seems to happen when focus is coming in to the IDE.

Note that the test on the line above is `true` if `getShell() == null`:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/31a902749e53ae095fb98c40459ba8cfc70d6d0d/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java#L847-L848

https://github.com/eclipse-platform/eclipse.platform.ui/blob/31a902749e53ae095fb98c40459ba8cfc70d6d0d/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java#L882-L887

HeikoKlare:
Several operation in the find/replace overlay need to be executed asynchronously as they may be called in OS callback which can lead to deadlocks in GTK. Currently, this processing is distributed and repeated across the class and did not check preconditions (in particular the dialog being opened and its shell being present) appropriate in every case. In some cases, when the overlay is already or temporarily closed but an asynchronous task is still to be executed, that task may fail with an NPE.

This change moves the functionality for the asynchronous processing to a central place, ensuring that the precondition requiring the shell to exist to always be evaluated. Factoring out this functionality also makes explicit that the called operation is executed only if the overlay respectively its shell are currently open. It also fixes the mentioned NPEs.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2244

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -16,7 +16,7 @@ package org.eclipse.ui.internal.findandreplace.overlay;
 import static org.eclipse.ui.internal.findandreplace.overlay.FindReplaceShortcutUtil.registerActionShortcutsAtControl;
 
 import java.util.List;
-import java.util.function.Supplier;
+import java.util.function.Consumer;
 
 import org.osgi.framework.FrameworkUtil;
 
@@ -156,7 +156,7 @@ public class FindReplaceOverlay extends Dialog {
 		setShellStyle(SWT.MODELESS);
 		setBlockOnOpen(false);
 		targetPart = part;
-		targetPartVisibilityHandler = new TargetPartVisibilityHandler(targetPart, this::getShell, this::close,
+		targetPartVisibilityHandler = new TargetPartVisibilityHandler(targetPart, this::asyncExecIfOpen, this::close,
 				this::updatePlacementAndVisibility);
 	}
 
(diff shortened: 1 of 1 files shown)


Wittmaxi:
Looks Good!

HeikoKlare:
I cannot reproduce the NPEs reported in #2244 with this fix. Just repeatedly opening the overlay via CTRL+F resulted in those logged exceptions (without any user-noticable behavior). Doing the same with the fix works fine.

Ubuntu 22.04.4 LTS, GNOME 42.9, GTK 3.24.33

PR #2248, opened 2024-09-06
[performance] QuickTextSearcher avoid streaming

jukzi:
For small files read and decode the whole file content at once.
* Avoid exceptions.

Quick Search (crtl alt shift L) the whole platform workspace (any String
that can not be found): 4s -> 2sec

### DIFF
--- a/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
+++ b/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
@@ -13,6 +13,9 @@
 package org.eclipse.text.quicksearch.internal.core;
 
 import java.io.InputStreamReader;
+import java.io.Reader;
+import java.io.StringReader;
+import java.io.UnsupportedEncodingException;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -23,6 +26,7 @@ import java.util.regex.Pattern;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
@@ -36,6 +40,7 @@ import org.eclipse.text.quicksearch.internal.util.LightSchedulingRule;
 import org.eclipse.text.quicksearch.internal.util.LineReader;
 
 public class QuickTextSearcher {
+	private static int MAX_BUFFER_LENGTH = 999_999; // read max 1MB bytes => max 2MB chars.
 	private final QuickTextSearchRequestor requestor;
 	private QuickTextQuery query;
 
@@ -155,7 +160,7 @@ public class QuickTextSearcher {
 			if (canceled.getAsBoolean()) {
 				return false;
 			}
-			try (LineReader lr = new LineReader(new InputStreamReader(f.getContents(true), f.getCharset()),
+			try (LineReader lr = new LineReader(getReader(f),
 					maxLineLength)) {
 				String line;
 				int lineIndex = 1;
@@ -179,6 +184,15 @@ public class QuickTextSearcher {
 			return true;
 		}
 
+		private static Reader getReader(IFile f) throws UnsupportedEncodingException, CoreException {
+			String shortString = toShortString(f);
+			if (shortString != null) {
+				return new StringReader(shortString);
+			} else {
+				return new InputStreamReader(f.getContents(true), f.getCharset());
+			}
+		}
+
 		@Override
 		public void resume() {
 			//Only resume if we don't already exceed the maxResult limit.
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
```diff
@@ -121,7 +123,27 @@ private SearchInFilesWalker createWalker(PriorityFunction priorities) {
...
-
+	/**
+	 * Try to get a content as String. Avoids Streaming.
+	 */
+	private static String toShortString(IFile file) {
+		/**
+		 * Just any number such that the most source files will fit in. And not too
+		 * big to avoid out of memory.
+		 **/
+		int MAX_BUFFER_LENGTH = 999_999; // max 2MB.
```
laeubi: 2 MB are `2_097_152` bytes so comment is confusing.... beside that constants should be defined in top of class file, computed values are often easier to read and change using a computation if they are computed e.g.

```suggestion
		int MAX_BUFFER_LENGTH = 2* 1024*1024; //Read 2 MB at maximum.
```

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
```diff
@@ -121,7 +123,27 @@ private SearchInFilesWalker createWalker(PriorityFunction priorities) {
...
+		try {
+			byte[] content = file.readNBytes(MAX_BUFFER_LENGTH);
+			int length = content.length;
+			if (length >= MAX_BUFFER_LENGTH) {
```
laeubi: This looks strange, why should the method read more than requested? In any case the data is there why not use it? And why then not better check for length <= 0 ?

the javadoc says:

```
	 * @param maxBytes The maximal length of the returned array. If maxBytes is
	 *                 greater or equal to the file length the whole content is
	 *                 returned. If maxBytes is smaller then the file length then
	 *                 only the first maxBytes bytes are returned.
```

so this will never return a larger array. looking furhter down, it feels like you first want to check the file-size.

laeubi (commented):
I'm not sure one can really say the improvement is from String#splitWithDelimtter it seem to change the logic considerably e.g. I suspect it to always read a short string

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
laeubi: By the way even though such method is missing in the API if used in a loop like here it is often better to allocate an array once and reuse that array, like in `java.io.InputStream.readNBytes(byte[], int, int)`

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
jukzi: The file is typically utf8 but the resulting String is 16bit/char in memory if there is a single non ascii char.
BTW that is all copy&paste from org.eclipse.search.internal.core.text.FileCharSequenceProvider.toShortString(IFile) and just works

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
```diff
@@ -132,22 +154,52 @@ protected void visit(IFile f, IProgressMonitor mon) {
...
+				if (string != null) {
+					// fast path:
+					Pattern lineSeparator = lineSeparators.get();
+					String[] lines = lineSeparator.splitWithDelimiters(string, -1);
+					int lineIndex = 1;
+					int offset = 0;
+					for (String line : lines) {
+						// line or separator?
+						if (line.length() > 2 || !lineSeparator.matcher(line).matches()) {
+							if (checkCanceled(mon)) {
```
laeubi: The cancel check is misplaced here it should at best be the first check in the loop.

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearcher.java
laeubi: > BTW that is all copy&paste from 

There it is a static constant on top of the file

>  and just works

Copy & Paste does not proof correctness... the length check will simply never be true (or the implementation is wrong and violates the API contract) as it can under no circumstances return an array with a length larger than requested.

~That means the method will never return `null` (unless IO or runtime errors) and therefore your code that checks for `null` here (**in a different context than where it was copied from**!) will never get null and therefore always only read 1 MB even if the file is 1 GB large... so something seems fishy here.~

Strike that, it checks if it is equal as well... that's very confusing check here. So maybe the speed improvement is more from this method if you have many small files than the usage of the new split method.

jukzi:
Took a solution that does not require higher JDK.

without this change:
image
with this change:
image

PR #2249, opened 2024-09-06
[performance] parallel Quick Search PoC

jukzi:
14s->2s with 6 cores

### DIFF
--- a/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
+++ b/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
@@ -48,7 +48,7 @@ public class QuickTextQuery {
 
 	private boolean caseInsensitive;
 	private String orgPattern; //Original pattern case preserved even if search is case insensitive.
-	private Pattern pattern;
+	final Pattern pattern;
 
 	/**
 	 * A query that matches anything.
@@ -60,15 +60,11 @@ public class QuickTextQuery {
 	public QuickTextQuery(String substring, boolean caseInsensitive) {
 		this.orgPattern = substring;
 		this.caseInsensitive = caseInsensitive;
-		createMatcher(substring, caseInsensitive);
+		String regex = createRegEx(substring);
+		pattern = Pattern.compile(regex, caseInsensitive ? Pattern.CASE_INSENSITIVE : 0);
 	}
 
-	/**
-	 * Compile a pattern string into an RegExp and create a Matcher for that
-	 * regexp. This is so we can 'compile' the pattern once and then keep reusing
-	 * the matcher or compiled pattern.
-	 */
-	private void createMatcher(String patString, boolean caseSensitive) {
+	private String createRegEx(String patString) {
 		StringBuilder segment = new StringBuilder(); //Accumulates text that needs to be 'quoted'
 		StringBuilder regexp = new StringBuilder(); //Accumulates 'compiled' pattern
 		int pos = 0, len = patString.length();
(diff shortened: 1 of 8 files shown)


bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
```diff
@@ -48,7 +48,7 @@ public int getOffset() {
 
 	private boolean caseInsensitive;
 	private String orgPattern; //Original pattern case preserved even if search is case insensitive.
-	private Pattern pattern;
+	private final ThreadLocal<Pattern> patterns;
```
laeubi: Javadoc says: 

> Instances of this class are immutable and are safe for use by multiple concurrent threads

so maybe better compile the pattern only once and pass it as an argument instead of a string?

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
```diff
@@ -48,7 +48,7 @@ public int getOffset() {
 
 	private boolean caseInsensitive;
 	private String orgPattern; //Original pattern case preserved even if search is case insensitive.
-	private Pattern pattern;
+	final Pattern pattern;
```
merks, line 51: Looking from my iPhone I can’t tell if the not being private is intentional.

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
jukzi: great. now its simpler

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
jukzi: it's intentional package private now to share it across threads

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
```diff
@@ -60,15 +60,11 @@ public QuickTextQuery() {
 	public QuickTextQuery(String substring, boolean caseInsensitive) {
```
laeubi, line 60:
```suggestion
	public QuickTextQuery(Pattern pattern) {
```

Maybe better make the caller pass the pattern here that makes it more flexible and could use static methods to construct Pattern from String without the need of constructor calling private methods.

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
laeubi: Maybe use a getter instead? Is the pattern reqully required or not more the matcher?

So one can thing of `public Matcher createMatcher(String input);`

jukzi: Pattern is thread safe, while Matcher is not.  This works. Unless you come up with a good reason to change it i would keep it as is. As the pattern is final as in a record there is no benefit of hiding it behind a getter.

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextQuery.java
jukzi: I don't see the benefit to change this.

Issue #2015, opened 2024-07-01
Find/Replace Overlay: Allow rebinding shortcuts


Wittmaxi:
As proposed by @BeckerWdf https://github.com/eclipse-platform/eclipse.platform.ui/pull/2008#issuecomment-2197046092

HeikoKlare:
The complexity with this issue seems to be that key bindings are attached to actions, for which I have not seen an easy way to be bound to simple SWT widgets, like used in the find/replace overlay. Usually, they are attached to workbench elements, like menu items or toolbar items.
Without defining key bindings for actions, I do not see a way to (1) make them configurable in a standard way and (2) to display the key bindings in a standard way (like using the icons for the modifier keys on Mac rather than writing "CMD", "CTRL" etc. by hand).

PR #2254, opened 2024-09-07
Find/replace overlay: replace shell with integrated composite #2099

### LINKED ISSUE #1447: New find/replace overlay broken under wayland
Trying to use new find/replace under Wayland is almost entirely impossible, It constantly moves and hides. Just watch the video

https://github.com/user-attachments/assets/9b8085fa-ea45-45ed-934b-e047c1501628

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x ] Linux

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Fedora 40, Gnome on Wayland

4. JRE/JDK version
 Java 23

### LINKED ISSUE #2099: Find/Replace Overlay: can hide views which are in "detached mode"

### LINKED ISSUE #2246: Find/replace overlay: does not take focus after detach
on Windows OS:

open an editor
detach it (drag to another screen or "detach" in the tabs context menu)
press strg+f -> opens search
enter "hallo"
-> "hallo" edits the document. it is supposed to be written in the search dialog

After detaaching mutliple times it is sometimes not possible to open Find/replace overlay at all.
Errors found in the error log:

```
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:500)
	at org.eclipse.swt.widgets.Widget.getDisplay(Widget.java:622)
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay$TargetPartVisibilityHandler.partHidden(FindReplaceOverlay.java:271)

```

```
Caused by: org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:500)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:419)
	at org.eclipse.swt.widgets.Control.setBackground(Control.java:3071)
	at org.eclipse.ui.internal.findandreplace.overlay.AccessibleToolBar.setBackground(AccessibleToolBar.java:63)
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.applyOverlayColors(FindReplaceOverlay.java:388)
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.open(FindReplaceOverlay.java:365)

```

```
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:500)
	at org.eclipse.swt.widgets.Widget.getDisplay(Widget.java:622)
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.asyncUpdatePlacementAndVisibility(FindReplaceOverlay.java:212)
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.lambda$0(FindReplaceOverlay.java:207)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:247)

```

```
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:500)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:419)
	at org.eclipse.swt.widgets.Control.getParent(Control.java:1449)
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.isInvalidTargetShell(FindReplaceOverlay.java:886)
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.lambda$22(FindReplaceOverlay.java:849)

```

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.Shell.setVisible(boolean)" because the return value of "org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.getShell()" is null
	at org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay.updatePlacementAndVisibility(FindReplaceOverlay.java:844)

```

HeikoKlare:
The FindReplaceOverlay is currently realized as a separate shell (more precisely, a JFace Dialog), which is placed at a proper position on top of the workbench shell. This has some drawback:
- It has to manually adapt to movements of the parent shell or the target part/widget
- It has to manually hide and show depending on visibility changes of the target part/widget
- It does not follow events of the target immediately, i.e., movements are always some milliseconds behind, minimize/maximize operations/animations are not synchronous etc.
- It does not locate properly when the platform uses Wayland, as manual shell positioning is not possible there.

This change replaces the dialog-based implementation of the FindReplaceOverlay with an in-place composite-based implementation. A composite is created in the target widget and placed relative to this composite. In consequence, the overlay automatically follows all move, resize, hide/show operations of the target widget. With this change, when the overlay has focus, general shortcuts for the editor (such as opening types/resources but also undo/redo) still work, while content specific shortcut (e.g., comment out current line) are disabled.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1447

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2099

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2246

### Known (Accepted) Issues / Drawbacks

- Proper handling of shortcuts requires the disablement of key bindings for the target text editor in order to (1) react to shortcuts at all and (2) in order to not execute unintended commands in the editor, such as commenting out a line while the overlay has focus. The implementation to achieve this is adapted from the breadcrumb implementation, which is kind of hacky but currently the only way I see to do this.
- Two of the shortcuts had to be changed due to conflicts with other shortcuts that can now be accessed from within the overlay. On the contrary, all shortcuts registered for editor, such as opening types or resources, are now accessible from within the overlay, which may be considered a benefit.
- For the other composite of the overlay, custom color handling had to be implemented, because it will usually be embedded into a StatusTextEditor containing a StyledText that uses a custom way of coloring as well, which is then inherited by the overlay. In order to have a consistent look, the colors of the outer composite of the overlay have to be retrieved and set in a custom way.

### <s>WiP</s>
<s>This is work in progress. At least the following is not working:</s>
- <s>Cursor: the cursor is shown according to the contents of the target widget, not according to the widgets in the overlay. I.e., on a button still the text cursor is shown.</s>
- <s>Shortcuts: shortcuts do currently not work as they are processed by the target widget and do not reach the overlay. Considering #2015, we may directly implement the shortcut as actions that can be reconfigured.</s>

Despite these problems that may require some additional effort, the overall design and integration with this approach makes more sense to be, as the overlay is really "integrated" into the target widget, as it is supposed to be. It gets rid of all the additional functionality to manually update position, size and visibility state according to changes of the target widget.

### How it looks
This is how it looks on Ubuntu using Wayland:
Screenshot from 2024-09-07 16-42-40

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -13,32 +13,29 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.findandreplace.overlay;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.List;
-import java.util.function.Consumer;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.osgi.framework.FrameworkUtil;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.BusyIndicator;
 import org.eclipse.swt.custom.StyledText;
-import org.eclipse.swt.events.ControlEvent;
 import org.eclipse.swt.events.ControlListener;
 import org.eclipse.swt.events.FocusEvent;
 import org.eclipse.swt.events.FocusListener;
 import org.eclipse.swt.events.KeyListener;
 import org.eclipse.swt.events.MouseListener;
-import org.eclipse.swt.events.ShellAdapter;
-import org.eclipse.swt.events.ShellEvent;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Point;
-import org.eclipse.swt.graphics.RGBA;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
-import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.ScrollBar;
 import org.eclipse.swt.widgets.Scrollable;
 import org.eclipse.swt.widgets.Shell;
@@ -46,25 +43,22 @@ import org.eclipse.swt.widgets.Text;
 import org.eclipse.swt.widgets.ToolItem;
 import org.eclipse.swt.widgets.Widget;
 
+import org.eclipse.core.runtime.Status;
+
 import org.eclipse.jface.bindings.keys.KeyStroke;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.IDialogSettings;
-import org.eclipse.jface.dialogs.IPageChangedListener;
-import org.eclipse.jface.dialogs.PageChangedEvent;
 import org.eclipse.jface.fieldassist.TextContentAdapter;
 import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.jface.resource.JFaceColors;
-import org.eclipse.jface.window.Window;
 
 import org.eclipse.jface.text.FindReplaceDocumentAdapter;
 import org.eclipse.jface.text.FindReplaceDocumentAdapterContentProposalProvider;
 import org.eclipse.jface.text.IFindReplaceTarget;
 import org.eclipse.jface.text.ITextViewer;
 
-import org.eclipse.ui.IPartListener2;
 import org.eclipse.ui.IWorkbenchPart;
-import org.eclipse.ui.IWorkbenchPartReference;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.fieldassist.ContentAssistCommandAdapter;
 import org.eclipse.ui.internal.findandreplace.FindReplaceLogic;
@@ -72,13 +66,14 @@ import org.eclipse.ui.internal.findandreplace.FindReplaceMessages;
 import org.eclipse.ui.internal.findandreplace.HistoryStore;
 import org.eclipse.ui.internal.findandreplace.SearchOptions;
 import org.eclipse.ui.internal.findandreplace.status.IFindReplaceStatus;
-import org.eclipse.ui.part.MultiPageEditorSite;
+import org.eclipse.ui.internal.texteditor.TextEditorPlugin;
 
+import org.eclipse.ui.texteditor.AbstractTextEditor;
 import org.eclipse.ui.texteditor.IAbstractTextEditorHelpContextIds;
 import org.eclipse.ui.texteditor.ITextEditorActionDefinitionIds;
 import org.eclipse.ui.texteditor.StatusTextEditor;
 
-public class FindReplaceOverlay extends Dialog {
+public class FindReplaceOverlay {
 	private final class KeyboardShortcuts {
 		private static final List<KeyStroke> SEARCH_FORWARD = List.of( //
 				KeyStroke.getInstance(SWT.CR), KeyStroke.getInstance(SWT.KEYPAD_CR));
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -143,26 +134,46 @@ private final class KeyboardShortcuts {
...
 	private ToolItem replaceAllButton;
 
-	private Color backgroundToUse;
+	private Color widgetBackgroundColor;
 	private Color normalTextForegroundColor;
 	private boolean positionAtTop = true;
-	private final TargetPartVisibilityHandler targetPartVisibilityHandler;
 	private ContentAssistCommandAdapter contentAssistSearchField, contentAssistReplaceField;
 
+	private class FixColorComposite extends Composite {
```
Wittmaxi: Why use this kind of composite? What do we fix here?

Wittmaxi:
There are a few extra pixels on the right side 
grafik

Wittmaxi:
looks like a very contructive commits which fixes multiple shortcomings of the current Find/Replace Overlay

Wittmaxi (approved):

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: It fixes that the CSS engine repeatedly overwrites the background color of composite with one defined by the theme. This is definitely not a perfect way to deal with the necessity to use some other component's background color here, but it's the only workaround I found so far. I am open for a better solution.

Note that this does not mean that the component is not properly themed: it just takes the themed background color of the shell instead of the one of the overlay's outer composite.

Wittmaxi: I am really not familiar (at all) with the CSS engine, but maybe we can add some CSS-tag to the Find/Replace overlay for custom styling? Drawback: all other stylesheets will need to style the overlay.

Anyway, this makes sense to me now and I believe that this is an okay workaround

HeikoKlare: I think we need to check again how the styling behaves in the overlay in general. Currently, the widgets inherit styling from the outer composite (usually the the composite of the editor's StyledText) except for the outermost composite within the overlay, which would otherwise simply be black in dark theme (i.e., have the same color as the editor, thus without any border). But I am not sure what happens if the composite was something else, i.e., if the colors will still fit then. Note that the proposed solution gets rid of all the existing custom color applications in the current implementation. It only needs to style two widgets instead of all the others.

We should definitely have another look at this topic.

akurtakov:
This looks very promising. What is holding it off ?

HeikoKlare:
> This looks very promising. What is holding it off ?

It's the key bindings. Most of them (e.g., those for all the search options) are currently not working in this proposal. The reason is that now that the overlay is not a separate dialog with its own key input handling, the active workbench part (i.e., the target text editor of the overlay) now processes key events. The two consequences are that (1) the find/replace key bindings do not work (at least most of them) as the input event won't even reach the overlay  and (2) the usual text editor key bindings are also evaluated when the overlay has focus (e.g., you can comment out the current text editor line while the overlay has focus with the according shortcut). Both behaviors are uninteded.
I am not yet sure how to properly solve this. I thought about adding some separate scope for the overlay, but I am not sure if it is even possible to have completely independent key binding scopes within the same workbench part. Maybe it is not sufficient to only integrate the overlay at the SWT layer (with just some composite at the right place) but also requires some integration in the workbench models (comparable to the proposal here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2093#issue-2411252154)? If anyone with more knowledge in the key binding concept knows a/the way to properly handle this I would very much appreciate input on this.

HeikoKlare:
This change of replacing the embedding of the overlay using dialog with an inplace composite seems to work almost fine now. I have documented some known drawbacks/issues (which I would consider acceptable for now) in the original post. In my opinion, the benefits (working on Linux/Wayland, resolution of several known issues) outweights them.

There are some test failures on Linux I will have a look into once I am at a Linux system again. After that, I would like to give this a try to finally have the overlay working on Linux, and apply further incremental improvements if necessary afterwards. Looking forward to your opinions on this, in particular @Wittmaxi.

HeikoKlare:
Test failures have been fixed and manual tests on all platforms (Windows, MacOS, Linux with/without Wayland) looked promising. Just found that undo/redo shortcuts now also work as discussed in some request for the overlay (i.e., they now apply to the editor instead of the search/replace input field). E.g., if you perform a replace and want to undo it, you can do that right away with the according shortcut without the need to switch focus to the editor before.

I plan to keep this open for feedback until next Wednesday, 9th October (and then merge), as I won't be able to respond to potential needs for follow-up actions before. In case someone wants to have this merged earlier, feel free to take over and do so.

Issue #888, opened 2022-06-07
Inbuilt support for content assist on typing like vscode


gayanper:
Today for JavaEditor we can achieve this my providing all the letters in the activation trigger. But the experience is not so good due to auto activation doesn't debounce the key events.  For example when you are trying continuously we don't want to content assist to popup and preselect wrong suggestions which the developer doesn't need.

The current activation logic can be extended to support debounce, We could introduce a configuration to enable this new activation feature for all valid keys with debounce. One simple implementation would be to change the current activation logic by resetting the last activation time when ever a key is pressed and reduce the delay to something like 100ms in this new mode.

gayanper:
@mickaelistria  @vogella  Do you think this is a good idea, if so I can try out and push a PR

mickaelistria:
Yes, it's a good idea that was tracked in https://bugs.eclipse.org/bugs/show_bug.cgi?id=101420 . I suggest you look into the comments and blockers in more details before going too deep. It's not as easy as it seems in the current state and we'll probably need preliminary fixes in JDT or other before this can be achieve smoothly.

mickaelistria:
See also eclipse-platform/eclipse.platform.text#66 which enables it by default in generic editor. It's very nice to use!

PR #2261, opened 2024-09-11
[WIP] Keep content assist popup open even if user input has no matches

### LINKED ISSUE #2260: Keep content assist popup open even if user input has no matches
Eclipse bugzilla ticket: https://bugs.eclipse.org/bugs/show_bug.cgi?id=545075

As the title goes, ideally keep the content assist pop-up as the user types (aside from a few special characters such as empty space). By doing this, the user doesn't have to invoke content assist again after some typo.

I'm guessing this is partially work that would be needed for: https://github.com/eclipse-platform/eclipse.platform.ui/issues/888

trancexpress:
WIP, DONT MERGE

Fixes: #2260

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java
@@ -371,9 +371,11 @@ class CompletionProposalPopup implements IContentAssistListener {
 			}
 			fFilterOffset= offset;
 
-			if (proposals != null && !proposals.isEmpty())
+			if (proposals != null && !proposals.isEmpty()) {
 				setProposals(proposals, fIsFilteredSubset);
-			else {
+			} else if (fContentAssistant.isShowEmptyList() && !shouldHideOnNoProposals(event, offset)) {
+				setProposals(Collections.emptyList(), false);
+			} else {
 				hide();
 				if (fContentAssistant.isAutoActivation() && offset > 0 && event != null) {
 					try {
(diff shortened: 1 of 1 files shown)


trancexpress:
There are artifacts with the current change, will have to look into those when I find time:


[gh2260_content_assist_artifacts.webm](https://github.com/user-attachments/assets/efac548a-60b8-45ec-9dd7-3e237e4cacdd)

Opened https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1643 for this.

mickaelistria:
The proposal seems like a clear enhancement to me. I would enjoy the IDE providing this. One thing I believe is important is that if user closes the popup (hits Escape), the popup doesn't show again except if it has actual proposals.

trancexpress:
> One thing I believe is important is that if user closes the popup (hits Escape), the popup doesn't show again except if it has actual proposals.

To be clear, https://github.com/eclipse-platform/eclipse.platform.ui/issues/2260 is for keeping the content assist popup once opened with the usual invocation. We don't aim for https://github.com/eclipse-platform/eclipse.platform.ui/issues/888, i.e. open the content assist popup automatically as soon as there is a proposal.

So we don't change how the popup is shown, we only keep it open once shown.

Though maybe I misunderstand your concern? Is there some preference that lets content assist show up automatically on typing, that I must test with?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/CompletionProposalPopup.java
```diff
@@ -2026,4 +2028,13 @@ public void setSorter(ICompletionProposalSorter sorter) {
 	void sortProposals(final List<ICompletionProposal> proposals) {
 		proposals.sort(fSorter::compare);
 	}
+
+	private static boolean shouldHideOnNoProposals(DocumentEvent event, int offset) {
+		try {
+			char c= event.getDocument().getChar(offset - 1);
+			return c == ' ';
```
jukzi, line 2035: better https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isWhitespace-char- ?

trancexpress:
Unfortunately no time for this.

PR #2262, opened 2024-09-12
Use LTK Refactoring when copying (duplicating) a Project.

N1k145:
The Idea is that currently it is not possible to modify a project when it is copied and renamed. This leads to usability issues for example in PDE, where the rename of a project modifies the MANIFEST.MF but a copy does not. This solves this by providing a copy project refactoring that can be extended using a copy participant.

This is related to #2155

### DIFF
--- a/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyProjectChange.java
+++ b/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyProjectChange.java
@@ -1,0 +1,153 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Vector Informatik GmbH - initial implementation
+ *******************************************************************************/
+package org.eclipse.ltk.core.refactoring.resource;
+
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.SubMonitor;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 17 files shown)


bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyProjectChange.java
```diff
@@ -0,0 +1,153 @@
...
+import org.eclipse.ltk.internal.core.refactoring.Messages;
+import org.eclipse.ltk.internal.core.refactoring.RefactoringCoreMessages;
+import org.eclipse.ltk.internal.core.refactoring.RefactoringCorePlugin;
+
+/**
+ * {@link Change} that copies a project
+ *
+ * @since 3.15
+ */
+public class CopyProjectChange extends ResourceChange {
```
mickaelistria, line 48: Does this need to be API ? It seems like none of the current use-cases we have for it make it necessary to be API and only the ID is used externally, or did I miss something. If not, then let's keep it internal.

N1k145: You are right, this does not need to be public.
I did it to aligne with the other ResourceChange Classes (e.g. DeleteResourceChange)
The main interaction should happen over the refactoring participants

mickaelistria: > The main interaction should happen over the refactoring participants

Would the refactoring participant typically need to access those classes?

N1k145: I had a look at our internal and PDE Implementations of RenameParticipants and noticed that they can create new changes, e.g. https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PluginManifestChange.java
So it could be helpfull for some usecases where you want to start a project rename from somewhere else than the eclipse internal rename dialog, that you can create this change object by hand

mickaelistria: Let's keep it as API then. Thanks for checking!

N1k145: No Problem, thank you for looking into it

N1k145:
@mickaelistria thank you very much for merging

mickaelistria:
> thank you very much for merging

Thank you for the patch!
Please keep actively looking at reported issues for a few weeks to detect potential related issues.

jukzi:
This PR introduced several new warnings in the workspace. can you please fix them?
image

mickaelistria:
I think the plugin.properties was missing in the commit.

N1k145:
@jukzi @mickaelistria thank you for noticing, I will look into it directly...

PR #2266, opened 2024-09-13
[StickyScrolling] Refactoring sticky lines provider and handler

Christopher-Hermann:
Refactors the sticky lines provider and extract the method to an interface. This is a preparation step for providing a extension point for sticky lines provider.

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/IStickyLinesProvider.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/IStickyLinesProvider.java
@@ -1,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2024 SAP SE.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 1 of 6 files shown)


Christopher-Hermann:
Failing test is unrelated.

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/IStickyLinesProvider.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.eclipse.swt.custom.StyledText;
+
+import org.eclipse.jface.text.source.ISourceViewer;
+
+/**
+ * A sticky lines provider calculates the sticky lines for a given source viewer. The sticky lines
+ * will be displayed in the top area of the editor.
+ * 
+ * TODO move to public package and add since 3.19
```
BeckerWdf, line 26: I think you will change this once the extension point is available and *released* as API

BeckerWdf:
I just read through very briefly..

BeckerWdf (approved):

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/IStickyLinesProvider.java
Christopher-Hermann: Yes, correct. But I want to keep the change with the extension point as small as possible so that it is easier to review

Issue #2245, opened 2024-09-06
IOConsoleTests: AutoRegisterSchemeHandlersJob still running on I20240905-0340 MacOS


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20240905-0340/testresults/html/org.eclipse.debug.tests_ep434I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

```
org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
java.lang.AssertionError: Test triggered errors in IOConsole
java.lang.AssertionError: Test triggered errors in IOConsole
Suppressed: java.lang.AssertionError: Test triggered errors in IOConsole
at org.eclipse.debug.tests.console.IOConsoleTests.lambda$1(IOConsoleTests.java:138)
at java.base/java.util.concurrent.ConcurrentLinkedQueue.forEachFrom(ConcurrentLinkedQueue.java:1037)
at java.base/java.util.concurrent.ConcurrentLinkedQueue.forEach(ConcurrentLinkedQueue.java:1054)
at org.eclipse.debug.tests.console.IOConsoleTests.assertNoError(IOConsoleTests.java:136)
at org.eclipse.debug.tests.console.IOConsoleTests.testBackspaceControlCharacter(IOConsoleTests.java:542)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
Caused by: java.lang.RuntimeException: Some job is still running or waiting to run:
'AutoRegisterSchemeHandlersJob(19)'/org.eclipse.urischeme.AutoRegisterSchemeHandlersJob
thread info: "Worker-2: AutoRegisterSchemeHandlersJob" prio=5 Id=47 RUNNABLE
at java.base@17/java.lang.String.substring(String.java:2710)
at java.base@17/java.lang.String.subSequence(String.java:2742)
at java.base@17/java.util.regex.Pattern.split(Pattern.java:1272)
at java.base@17/java.lang.String.split(String.java:3150)
at java.base@17/java.lang.String.split(String.java:3196)
at org.eclipse.urischeme.internal.registration.RegistrationMacOsX.determineHandlerLocation(RegistrationMacOsX.java:88)
at org.eclipse.urischeme.internal.registration.RegistrationMacOsX.getSchemesInformation(RegistrationMacOsX.java:67)
[...]

PR #2267, opened 2024-09-13
improve RegistrationMacOsX performance #2245

jukzi:
Pattern matching with String.split("-{80}\n") is quiet expensive especially if the string contains more then 80 dashes in a row.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2245

### DIFF
--- a/bundles/org.eclipse.urischeme/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.urischeme/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.urischeme;singleton:=true
-Bundle-Version: 1.3.400.qualifier
+Bundle-Version: 1.3.500.qualifier
 Automatic-Module-Name: org.eclipse.urischeme
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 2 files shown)


sratz (commented):
Change looks good in principle.

However, there are exactly 80 dashes in that output, not more than 80.

Are you sure this will actually boost performance?

jukzi:
> Are you sure this will actually boost performance?

I dont't have mac to test it locally.  It's just that this split was logged so often that it can be seen as hotspot and local generic tests show that the alternative regex is faster, also reusing the splitted result may help

sratz (approved):
Ok, then let's merge this and see if it helps 👍

BeckerWdf:
> > Are you sure this will actually boost performance?
> 
> I don't have mac to test it locally. It's just that this split was logged so often that it can be seen as hotspot and local generic tests show that the alternative regex is faster, also reusing the splitted result may help

There are JUnit tests that have "real-life" console outputs. See: [TestUnitRegistrationMacOsX](https://github.com/eclipse-platform/eclipse.platform.ui/blob/023f3dea8db60e5300c55821661ff2b39cde2241/tests/org.eclipse.tests.urischeme/src/org/eclipse/urischeme/internal/registration/TestUnitRegistrationMacOsX.java#L39). These should be runnable on all OSes. Can you provide numbers how much faster they run?

BeckerWdf:
Thanks for taking care of performance.

PR #2271, opened 2024-09-14
Find/replace overlay: do not escape selected text when not regex #2270

### LINKED ISSUE #2270: [Find/Replace] Pre-selected search string is always escaped even if RegEx search is disabled

## Steps to reproduce

If one opens the Find/Replace overlay with text in an editor being selected as initial search input, that search input is now always escaped even if RegEx search is disabled (see screenshot below).
This has the consequence that the non-RegEx search does not find the selected text because it treats the escape characters literally.

grafik

Furthermore the search input is partally hidden/truncated at left, as displayed in the screenshot.

@Wittmaxi or @HeikoKlare could you please have a look at this?
IIRC this was introduced round about in the past week.

* OS & version: `Windows 10`
* Eclipse IDE/Platform version (as shown in *Help > About*): `Version: 2024-12 (4.34), Build id: I20240913-1800`

## Community

HeikoKlare:
When opening the find/replace overlay with a text selection in the target editor, this text is set as the find input to the overlay. When in regex mode, the string is escaped accordingly. Currently, the string is also escaped when regex mode is not activated.

With this change, the find input string is not escaped when regex mode is not activated. An according test case is added.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2270

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -977,7 +977,7 @@ public class FindReplaceOverlay extends Dialog {
 			findReplaceLogic.deactivate(SearchOptions.GLOBAL);
 			searchInSelectionButton.setSelection(true);
 		} else if (!selectionText.isEmpty()) {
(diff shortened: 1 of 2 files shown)


deepika-u:
I verified this fix and it works fine on below build

Eclipse SDK
Version: 2024-12 (4.34)
Build id: I20240917-1800
OS: Windows 11, v.10.0, x86_64 / win32
Java vendor: Eclipse Adoptium
Java runtime version: 21.0.1+12-LTS
Java version: 21.0.1

Issue #1532, opened 2024-01-21
Replace "Default" text editor with "Generic" text editor


iloveeclipse:
Recently I've seen few cases where current "Generic Text Editor" (`org.eclipse.ui.genericeditor.GenericEditor`) was competing with "Text Editor" (`org.eclipse.ui.DefaultTextEditor`) and IDE decided to use later one, even if "Generic" is almost always a better choice, especially together with [tm4e](https://github.com/eclipse/tm4e). For example, because IDE doesn't allow to contribute "default" editor for content types, simply first found editor contribution is used to open "unknown" text files (the one from simple Text editor), and not the "Generic" one. Also having two "similarly named" editors in the IDE confuses users and leads to [lengthy discussions](https://github.com/eclipse-platform/eclipse.platform.ui/issues/1385) which name should one use for these editors.

That all seem to be a problem of a **parallel co-existence** of two "plain text editors" in same IDE - old "simple" text editor and the "generic" one. This co-existence is caused by the IDE evolution, where the "generic" text editor (with more functionality under the hood) was added later but the old & simple one was not removed because it was (and still is) widely used.

What I would propose is to evaluate if we can simply **replace** editor contribution of the "simple" text editor with "generic" one, **without removing implementation class / editor id**, to stay compatible as much as possible.

Note: I know there is a `GenericEditorWithIconAssociationOverride` that can be easily changed to open `org.eclipse.ui.genericeditor.GenericEditor` instead of `org.eclipse.ui.DefaultTextEditor`, but that one doesn't allow to **hide** all the default text editor contributions from the UI and keeps it in the file associations/content types preferences, so it is not quite what needed here.

What I'm thinking about is to extend the `org.eclipse.ui.editors` extension point with `alias` or `override` element that would internally redirect all uses of one editor id with another one and so "hide" all contributions of the old one, eliminating duplicated and confusing editor contributions appearing in the UI.

So the `org.eclipse.ui.genericeditor.GenericEditor`  would say "I'm an alias/override for  `org.eclipse.ui.DefaultTextEditor`, so use me if anyone refers to  `org.eclipse.ui.DefaultTextEditor`".

[...]

Issue #1385, opened 2023-12-06
Relabel "Generic Editor" (or "Advanced Text Editor")


mickaelistria:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/) doesn't provide the feature.

## Suggestion

Here and there, we see references to "Generic Editor" or "Advanced Text Editor". The first one is not really user-oriented as the name is more related to the internals of the Editor and it doesn't tell much about what it can do better; the second tries to emphasize that it should be preferred, but it doesn't tell much about what cases it's supposed to serve better.
So I would recommend that we consider a better label for this editor.
I currently have 2 suggestions:
* Generic Code Editor
* Generic Source Editor

I like keeping the "Generic" to emphasize that there can be some better specialized editors, but telling "Code" or "Source" also tells that some convenience features for developers (brackets matching, collapse range...) can already be available for this editor.

## Community

- [x] I understand suggesting an enhancement doesn't mandate anyone to implement it. Other contributors may consider this suggestion, or not, at their own convenience. The most efficient way to get it fixed is that I implement it myself and contribute it back as a good quality patch to the project.

vogella:
I think we should keep it simple. Proposal:

- Generic Editor / Advanced Text Editor -> Text Editor or Code Text Editor
-  Text Editor -> Simple Text Editor

This should make it clear that "Simple" should not be used

vogella:
@laeubi WDYT?

mickaelistria:
> I think we should keep it simple.

I'm not convinced your proposal is substantially simpler than other ones. I even fear it may introduce some confusion if "Text Editor" suddenly start pointing at something else. IMO we should keep the "Text Editor" as it since it's not a bad name and it's a long established one, and only focus on the "Generic Editor" which is less good and less established.

vogella:
What about "Code Text Editor"?

laeubi:
`Generic Code Editor` seems good for me form the point of an IDE, maybe one should relabel the other to `Plain Text Editor` then.

More important I think for any Eclipse EPP package the default should be the code editor (as one likes to edit code most of the time) and if tm4e support generic syntax highlight in such case tm4e should be included in all EPPs as well (/CC @jonahgraham )

laeubi:
> Here and there, we see references to "Generic Editor"

[...]

PR #2277, opened 2024-09-16
Rename Text Editor into Plain Text Editor

mickaelistria:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1532 and https://github.com/eclipse-platform/eclipse.platform.ui/issues/1385


mickaelistria:
@merks @vogella and what do you think about this (IMO simpler) one?

merks:
I personally like the "Plain Text" qualifier; it sets low expectations as is appropriate.  That doesn't mean everyone will like it.  😝

vogella:
I like this change.

mickaelistria:
This one seems consensual enough: it was proposed on another issue, and is supported by the people who've been most active on related discussions. So let's just merge it now.

PR #2278, opened 2024-09-16
Rename Generic Text Editor into Generic Code Editor

### LINKED ISSUE #1385: Relabel "Generic Editor" (or "Advanced Text Editor")

mickaelistria:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1532 and https://github.com/eclipse-platform/eclipse.platform.ui/issues/1385

### DIFF
--- a/bundles/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.ui.genericeditor;singleton:=true
-Bundle-Version: 1.3.400.qualifier
+Bundle-Version: 1.3.500.qualifier
 Bundle-Vendor: %Bundle-Vendor
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Require-Bundle: org.eclipse.ui.workbench.texteditor;bundle-version="3.10.0",
--- a/bundles/org.eclipse.ui.genericeditor/plugin.properties
+++ b/bundles/org.eclipse.ui.genericeditor/plugin.properties
@@ -14,7 +14,7 @@
 #                                 - Bug 521382 default highlighter
 #     Simon Scholz <simon.scholz@vogella.com> - Bug 527830
 ###############################################################################
-genericEditor_name=Generic Text Editor
+genericEditor_name=Generic Code Editor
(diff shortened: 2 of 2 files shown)


vogella:
I would drop the "Generic". "Code Editor" or "Text Editor" (if the old one is renamed) is short and IMHO less confusing for the user.

mickaelistria:
Let's go incrementally. Choosing a name never is a consensual exercise, going baby steps is more likely to create progress.
Moreover, the "Generic Code Editor" is usually a good fit for "generic code edition", but if there is a dedicated editor (eg "Java editor"), then the "Generic Code Editor" seems a bit too generic as it wouldn't provide as many language features. So Generic is not too harmful at the moment.

vogella:
I'm fine with the incremental change idea. Change is fine for me.

merks:
I wonder if the end-user considers *.html, *.xm, or *.target to be code?  (I dislike naming discussions; you can never please everyone.   But I really wonder what is code and what is not code?

mickaelistria:
> But I really wonder what is code and what is not code?

I don't know if this even can have an answer :shrug: My interpretation is that any textual language that's not prose is more or less code. ie if it's not meant to talk to directly to a human and if the language is tailored first for a machine to understand or manipulate or executre or render... it, it's code to me.

mickaelistria:
One more argument in favor of "Code" is that many people use Visual Studio Code (executable is `code`) to edit their HTML files and AFAIK there has been no argument that the name wasn't suitable.

HannesWell:
> I would drop the "Generic". "Code Editor" or "Text Editor" (if the old one is renamed) is short and IMHO less confusing for the user.

I would also be fine with just `Text Editor`and `Code Editor`. If I read the later I would expect that the editor can handle 'codish' content.
That the editor is a generic `Code Editor` seems to be implied for me because a generic name like `Code` is used.

And I would also expect that an editor that is named more specifically can handle that specific thing better.
I.e. when having a `Code Editor` and a `Java Editor` I would expect the later to handle Java files best. 

That's just my 2ct. If the majority prefers `Generic Code Editor` I would also be fine.

PR #2282, opened 2024-09-17
[test] delete ControlObservableValueTest.png after test

jukzi:
2MB artifact in i builds

### DIFF
--- a/tests/org.eclipse.jface.tests.databinding/src/org/eclipse/jface/tests/internal/databinding/swt/ControlObservableValueTest.java
+++ b/tests/org.eclipse.jface.tests.databinding/src/org/eclipse/jface/tests/internal/databinding/swt/ControlObservableValueTest.java
@@ -235,13 +235,17 @@ public class ControlObservableValueTest extends AbstractDefaultRealmTestCase {
 		System.out.println("active shell (5): " + shell.getDisplay().getActiveShell());
 
 		System.out.println("Value (should be true): " + value.getValue());
-		Screenshots.takeScreenshot(getClass(), getClass().getSimpleName(), System.out);
+		try {
+			assertTrue(value.getValue());
 
-		assertTrue(value.getValue());
-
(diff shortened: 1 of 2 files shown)


jukzi:
@HeikoKlare what is that png for? it is saved in ever build. for example
https://download.eclipse.org/eclipse/downloads/drops4/I20240916-1800/testresults/ep434I-unit-mac64-java17_macosx.cocoa.x86_64_17/org.eclipse.jface.tests.internal.databinding.swt.ControlObservableValueTest.ControlObservableValueTest.png

akurtakov:
Screenshots like this are generally a "debugging" tool when trying to fix UI test. If they are deleted like that they can not be inspected so it's probably better to not generate them at all.

jukzi:
If the test fails the screenshots remain as artifact. That could be useful. Maybe better they would only be generated on fail?

akurtakov:
> If the test fails the screenshots remain as artifact. That could be useful. Maybe better they would only be generated on fail?

If you manage to do that it would be awecome.

jukzi:
ignoring unrelated `Download of osgi.bundle,org.eclipse.jface.examples.databinding,1.4.400.v20240424-0956 failed` on linux

HeikoKlare:
> what is that png for? it is saved in ever build. for example

> Screenshots like this are generally a "debugging" tool when trying to fix UI test.

The places where screenshots are saved that I have seen so far are for documenting and debugging inproper focus handling, like here:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/be54f35f78bf182dad6b3d4beefc9d7f0d209068/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceUITest.java#L84-L87

So your fix to only save the screenshot in case the failure that needs to be analyzed via that screenshot occurs is the right thing to do.

PR #2286, opened 2024-09-17
Light (Preview) theme - change in the behavior of highlights for selected tabs

mvm-sap:
All selected tabs in Light (Preview) theme had blue highlight(underline) irrespective of stack being active or inactive. At any point in time only one tab could be active and it has to be blue. This change has been done by making blue highlight only for the active selected tab and rest of the selected tabs which are inactive will have grey highlight.
Please find the below screen shots:

Before:
image

image

After :
image

image

PR #2287, opened 2024-09-17
org.eclipse.ui.themes - Version bump for 4.34 stream

mvm-sap:


BeckerWdf (changes requested):
don't do a separate PR but a second commit in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2286

BeckerWdf:
included in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2286

Issue #2279, opened 2024-09-16
UI freeze when searching over large files. Search View Filter viewer limit is not applied for children in TreeViewer.


raghucssit:
This issue is already reported here. Please close this one on Bugzilla as duplicate.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=575271

UI freezes over around 12 seconds if have a large search result in the second level tree onwards. The reason is Search View Filter viewer limit is not applied for children.
Use below snippet to generate 100k files with searchable text in it. Initiate File Search with text `text here`
  ```
      public static void main(String[] args) throws IOException {
                Path folder = Paths.get("...");
                if (!Files.deleteIfExists(folder)) {
                	Files.createDirectory(folder);
                }
                int n = 100_000;
                String baseContent = "text here";
                for (int i = 0; i < n; ++i) {
                        System.out.println("Writing file " + i);
                        StringBuilder content = new StringBuilder();
                        int k = 1;
                        if (i == 1001) {
                        	k = 1000;
                        }
                        for (int j = 0; j < k; ++j) {
                        	content.append(baseContent);
                        	content.append(" n");
                        	content.append(i);
                        	content.append(System.lineSeparator());
                        }
                        String fileName = "file" + i + ".txt";
                        Path file = folder.resolve(fileName);
                        try {
                                Files.write(file, content.toString().getBytes());
                        } catch (IOException e) {
                                System.err.println("Failed to write file: " + fileName);
                                e.printStackTrace();
                        }
                }
        }
```

jukzi:
Can you please provide screenshot how you search and what the result looks like? Is this a linux issue only?

raghucssit:
This is on linux. I did not test on windows.
UI will not show not responding but nothing can be performed on UI for 12 seconds and more.

100k files in PE. PE don't freeze because there is global viewer limit.
![100k-pe](https://github.com/user-attachments/assets/aa8500c6-839f-4dce-8467-6f9d7ae53296)


Search results view tries to expand the `dir` node to select first match.
![100k-sv](https://github.com/user-attachments/assets/68052726-87c0-4fd9-8abb-0211667a17a9)

iloveeclipse:
After playing a bit more with search view after merged changes I saw two things: 

1) We still have a full hang (forever on the example with 100.000 files here) after doing "F5" (run the current search again) - this is because the view update code calls `AbstractTextSearchResult.getMatchCount()` on every element to update and that traverses every match...
[...]

PR #2288, opened 2024-09-17
Reduce freeze times on updating search results on resource changes

iloveeclipse:
If the search view shows ~100000 matches, deleting resources that are shown in the search view freezes Eclipse for many minutes. With this change the freeze is only about 30 seconds.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2279

To test, perform steps from https://github.com/eclipse-platform/eclipse.platform.ui/issues/2279 and after search results view shows matches, delete generated files in Package Explorer.

### DIFF
--- a/bundles/org.eclipse.search/newsearch/org/eclipse/search/ui/text/AbstractTextSearchResult.java
+++ b/bundles/org.eclipse.search/newsearch/org/eclipse/search/ui/text/AbstractTextSearchResult.java
@@ -21,6 +21,7 @@ import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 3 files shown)


iloveeclipse:
@raghucssit : I had not enough time to play with this patch today, but it seemed to work (**on top** of your https://github.com/eclipse-platform/eclipse.platform.ui/pull/2280).

Could you please double check that that works as expected?

bundles/org.eclipse.search/newsearch/org/eclipse/search/ui/text/AbstractTextSearchResult.java
```diff
@@ -308,6 +309,19 @@ public int getMatchCount() {
 		return count;
 	}
 
+	/**
+	 * @return {@code true} if the result is not empty
+	 * @since 3.17
+	 */
+	public boolean hasMatches() {
+		for (Entry<Object, Set<Match>> entry : fElementsToMatches.entrySet()) {
```
jukzi, line 317: when only values are needed better iterate over `Map.values()`

raghucssit:
@iloveeclipse I tested this PR. Changes LGTM.
I tried with different viewer limits like 5k, 10k, 20k, 50k, etc. Default limit is 1k.
With default limit there was a noticeable lag while clearing all matches. Now it works like magic. Very very fast.
Upto 20k viewer limit it clears the matches in no time.(Regular users may not go beyond this)
I did not observe any problem.

Issue #423, opened 2022-10-21
Remove dependencies on org.apache.commons.jxpath


mickaelistria:
org.apache.commons.jxpath is a very old library, full of CVE and unmaintained for 14 years.
On the other end, the JDK ships some build path implementation that is maintained in high quality just like everything in the JDK.

We should remove deps on jxpath and use JDK standard implementation.

mickaelistria:
@merks Is there something already existing in EMF land to handle XPath without jxpath and that Platform could reuse?

akurtakov:
Or is this bundle helpful/used in anyway? It never got to 1.0.0.

merks:
@mickaelistria Sorry, no.

mickaelistria:
> Or is this bundle helpful/used in anyway? It never got to 1.0.0.

Unfortunately, it seems used (or at least referenced) by org.eclipse.e4.model.workbench and by org.eclipse.e4.tools which is transtively used by the model.spy.

ptziegler:
What would be a good starting point for removing this dependency? I noticed that for the 2024-09, the platform switched to the official Maven jar. This now introduces some very nasty dependencies like jdom and commons-collections, which I'd really like to not have in our application:

![image](https://github.com/user-attachments/assets/c9710edb-2e94-443d-a45c-2a233af9875e)

Depending on how difficult this is, I might be able to get some time to work on this topic.

laeubi:
@ptziegler the easiest way is to remove the dependency from the manifest (either the package or the bundle) and then you will see the compile errors and can decide how to replace this with standard java XPath. If that's done one can remove it from the target entirely and see if there are some "hidden" references.

Even if we can not yet remove all usages it might be worth to start bundle-by-bundle.

ptziegler:
Hm... this will be tricky. At least for the e4 workbench, this library is used to look up elements relative to the application model, rather than the current fragment:

![image](https://github.com/user-attachments/assets/a33c2a86-2d2b-4806-ae81-099a4ebe7db0)

With it only used being here:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/a43889deed6bece2600069c81d5b2368d2d86137/bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/fragment/impl/StringModelFragmentImpl.java#L355-L393

In order to switch to the Java XPath API, one would need to keep track of the XML structure and then go:

[...]

PR #2289, opened 2024-09-17
Various minor improvements for E4 EMF Xpath

HannesWell:
This PR contains three commits with minor improvements of the E4 EMF Xpath bundle:

1. Unify and clean-up definitions of generic type arguments
2. Import org.apache.commons.jxpath packages
as suggested in https://github.com/eclipse-platform/eclipse.platform.ui/issues/423#issuecomment-2355284101
3. Add `XPathContext.stream()` and strengthen type of getValue(String, Class)

The new method is for example useful in:
https://github.com/eclipse-pde/eclipse.pde/blob/ba11267159a1f5eed242af8f18dadc299f763637/e4tools/bundles/org.eclipse.e4.tools.emf.ui/src/org/eclipse/e4/tools/emf/ui/internal/common/component/StringModelFragment.java#L587-L600

### DIFF
--- a/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
+++ b/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2015 BestSolution.at and others.
+ * Copyright (c) 2010, 2024 BestSolution.at and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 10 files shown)


bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
```diff
@@ -42,19 +46,35 @@ public interface XPathContext {
 	 *            required type
 	 * @return Object found
 	 */
-	Object getValue(String xpath, Class<?> requiredType);
+	<T> T getValue(String xpath, Class<T> requiredType);
```
HannesWell, line 51: Since the method signature is the same after generic erasure this is a binary and source compatible change.
But as far as I can tell, this method and its overload is not used within the SDK at all, besides tests.

HannesWell:
If there are no unresolved remarks I plan to submit this tomorrow (European) evening.

PR #2294, opened 2024-09-18
Light (Preview) Font color change for text in tabs

mvm-sap:
Font color of all tabs selected or unselected are set to black  #000000 with this change to have a better contrast.

Before :
image

After:
image

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/e4_preview_gtk.css
+++ b/bundles/org.eclipse.ui.themes/css/e4_preview_gtk.css
@@ -85,15 +85,15 @@ ColorDefinition#org-eclipse-ui-workbench-ACTIVE_NOFOCUS_TAB_BG_END {
 }
 
 ColorDefinition#org-eclipse-ui-workbench-ACTIVE_NOFOCUS_TAB_TEXT_COLOR {
-	color: #3b3b3b;
+	color: #000000;
 }
 
 ColorDefinition#org-eclipse-ui-workbench-ACTIVE_TAB_TEXT_COLOR {
-   color: #3b3b3b;
+   color: #000000;
 }
 
 ColorDefinition#org-eclipse-ui-workbench-INACTIVE_TAB_TEXT_COLOR {
-   color: #3b3b3b;
+   color: #000000;
 }
 
 .MTrimmedWindow {
@@ -197,13 +197,13 @@ CTabFolder Canvas {
 
 /* text color and background color of unselected tabs in editor*/
 #org-eclipse-ui-editorss CTabItem{
-	color: '#6A6A6A';
+	color: '#000000';
 	background-color: '#f8f8f8';
 }
 
 /*text color and background color of selected tab in editor */
 #org-eclipse-ui-editorss CTabItem:selected{
-	color: '#3b3b3b';
+	color: '#000000';
 	background-color: '#FFFFFF';
 }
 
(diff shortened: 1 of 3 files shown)


BeckerWdf:
how is this different to https://github.com/eclipse-platform/eclipse.platform.ui/pull/2187 ?

mvm-sap:
> how is this different to #2187 ?

In #2187  
Unselected Tabs in view and editor had  #3b3b3b
and 
Selected Tabs in view and editor had #000000

But with this PR 
All tabs, selected and unselected both in editor and view has #000000

We need to delete #2187

BeckerWdf (approved):

BeckerWdf:
looks good on macOS for me.
@jukzi: Can you pls. also test and say if this addresses your issues?

jukzi:
looks better, but it's hard to see which secondary tab is active:
image
used to look like this where it was obvious which tab is the active view by background color:
image

BeckerWdf:
> but it's hard to see which secondary tab is active

Do you mean the "Console" tab in your screenshot? It's now indicated by the grey underline.

jukzi:
> Do you mean the "Console" tab in your screenshot? It's now indicated by the grey underline.

yes. Unfortunately that grey underline is not as prominent as the changed background. isn't there even some option to diable that underline? Then it would be totally unclear. 
Maybe i can get used to the underline, but at a first glance the old style was more expressive.

BeckerWdf:
> > Do you mean the "Console" tab in your screenshot? It's now indicated by the grey underline.
> 
> yes. Unfortunately that grey underline is not as prominent as the changed background. isn't there even some option to diable that underline? Then it would be totally unclear. Maybe i can get used to the underline, but at a first glance the old style was more expressive.

Yes that's true but that's the intention of that change. As it's not the "primary active tab" but only the "secondary active tab" is should be not as prominent. We want to give the "main actor" more focus and the "secondary active tab" is not only a "supporting actor". But we could make the grey underline a bit darker grey to make it a bit more prominent but not _too_ prominent.
What about something like this:
Screenshot 2024-09-19 at 09 39 27
compared to the current state:
Screenshot 2024-09-19 at 09 40 53

jukzi:
darker looks better. maybe even one pixel thicker?

jukzi:
The best would be if the high contrast icons of the inactive  tabs gets reduced :-) just as idea

BeckerWdf:
> The best would be if the high contrast icons of the inactive tabs gets reduced

That's a nice idea but currently not possible to implement. We could save this for a further improvement via a separate PR.

BeckerWdf:
> maybe even one pixel thicker?

We just made the line thinner in the last release. This is something we can improve (and maybe make it bigger again) later on.

BeckerWdf:
> darker looks better. maybe even one pixel thicker?

I just pushed a second commit into this pr making the grey darker.
Now looks like the one on the right side. Left side is the current grey color:
image

BeckerWdf:
https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2294/4/console reports a lot of API breakages.
They are not caused by this PR and for it looks like false alarms.
Does anybody have a clue what's happening?

jukzi:
> Does anybody have a clue what's happening?

may be a random fail retry.

jukzi (approved):
LGTM

PR #2296, opened 2024-09-18
Find/replace overlay: store search in history when leaving overlay #2291

### LINKED ISSUE #2291: Find/Replace Overlay:  History missing
History is only added when i hit RETURN. but since the search already starts when i paste a String i do not need to hit RETURN. Instead i use the mouse to click into the editor where the result is found=> no history added. 
But later when i want todo the same search again i  am missing the last search i did.

Maybe add the current search to history when the overlay is left?

HeikoKlare:
The find/replace overlay only stores a search input in the history when an explicit search operation (forward/backward/all search) is performed. In case only an incremental search (via search-as-you-type) is performed, this input will not be added to the history. In some cases, no explicit search is executed but the final result of a search-as-you-type input is sufficient for the user. Since this final result, after they finished typing, also represented a relevant search input, it should be added to the history.

This change addresses the issue by adding a search input to the history also when the overlay is left, i.e., the focus is moved to somewhere else, such as the target editor or some other view.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2291


HeikoKlare:
@jukzi fyi

HeikoKlare:

Issue #2297, opened 2024-09-19
Improve "Editors" Preference Page


BeckerWdf:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

Currently the "Editor" preference page looks like this:
<img width="882" alt="Screenshot 2024-09-19 at 09 01 49" src="https://github.com/user-attachments/assets/618d8182-f1a1-4bd6-8b4c-97632fd1a24b">

I just was wondering why the "_Number of opened..._" setting is indented  but the "_Size of.._" setting is not.
After some thinking I realized that the "_Number of opened.._" setting depends on the checkbox above.
Due to the "_Size of.._" input field the distance between the input field and the checkbox is bigger then the distance between the other check boxes. This also makes it harder to recognize that these two are linked.

I would propose we put a "_Close mode_" frame around it. Something similar as we have it here:
<img width="884" alt="Screenshot 2024-09-19 at 09 02 27" src="https://github.com/user-attachments/assets/2d768fc9-fe33-4774-b120-cccd80a26ef0">

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

PR #2298, opened 2024-09-19
Improve UX for "Close mode" settings in editor preference page

praveen-skp:
Issue: #2297 

This pull request adds a "Close mode" frame in the editor preference to make the "Close editors automatically" and "Number of opened editors before closing" grouping clear.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
@@ -493,10 +493,10 @@ OpenPerspectiveDialogAction_tooltip=Open Perspective
 
 #---- General Preferences----
 PreferencePage_noDescription = (No description available)
-PreferencePageParameterValues_pageLabelSeparator = \ >\ 
+PreferencePageParameterValues_pageLabelSeparator = \ >\
 ThemingEnabled = E&nable theming
 ThemeChangeWarningText = Restart for the theme changes to take full effect
-ThemeChangeWarningTitle = Theme Changed 
+ThemeChangeWarningTitle = Theme Changed
 # --- Workbench -----
 WorkbenchPreference_openMode=Open mode
 WorkbenchPreference_doubleClick=D&ouble click
@@ -729,8 +729,8 @@ PageLayout_missingRefPart=Referenced part does not exist yet: {0}.
 # Keys used in the reuse editor which is released as experimental.
 # ==============================================================================
 PinEditorAction_toolTip=Pin Editor
-WorkbenchPreference_reuseEditors=&Close editors automatically
-WorkbenchPreference_reuseEditorsThreshold=Number of opened editors before closi&ng:
+WorkbenchPreference_reuseEditors=&Close editors if there are more than
+WorkbenchPreference_reuseEditors_closing=editors open
 WorkbenchPreference_reuseEditorsThresholdError=The number of opened editors should be more than 0.
 WorkbenchPreference_recentFiles=Size of &recently opened files list:
 WorkbenchPreference_recentFilesError=The size of the recently opened files list should be between 0 and {0}.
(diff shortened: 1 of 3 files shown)


praveen-skp:
Close mode

The new UI with "Close mode" grouping

laeubi:
`Close mode` as a group name reads a bit strange here, I would expect then having different modes to select but the configuration do not really offer that.

My suggestion would be to merge both settings into one line instead of grouping them e.g.

`[ ] Close oldest if there are more than _____ editors open`

BeckerWdf:
> `Close mode` as a group name reads a bit strange here

What about "Closing behaviour"?

BeckerWdf:
> My suggestion would be to merge both settings into one line instead of grouping them

This is also a good idea.

praveen-skp:
This is how the new UI looks.
Disabled
Enabled

BeckerWdf:
Can the input field made shorter so that the line doesn't get too long. I think place for around 4 chars would be sufficient.

What about changing the text to: "Close oldest editor if there more then ... editor open"

What does "oldest" mean? 
a) The one that was open the longest time, or 
b) the one where the time of last use is most far in the past?

praveen-skp:
I checked the current behaviour and noticed that it's not always the oldest editors that are being closed.

### Use case 1:
If I have the following editors opened in order:
1, 2, 3, 4, 5, 6 (1 being the oldest, focus is on 6)

- Set editor limit to 3
- Open editor 7
- The current available editors are 7, 2, 3

It deletes the latest 3 editors (6, 5, 4) and replaces the oldest editor.

---

### Use case 2:
If I have the following editors opened in order:
1, 2, 3, 4, 5, 6 (1 being the oldest, focus is on 5)

- Set editor limit to 3
- Open editor 7
- The current available editors are 7, 2, 3

It deletes the latest 3 editors (5, 6, 4) and replaces the oldest editor (1).

---

### Use case 3:
If I have the following editors opened in order:
1, 2, 3, 4, 5, 6 (1 being the oldest, focus is on 2)

- Set editor limit to 3
- Open editor 7
- The current available editors are 7, 3, 4

It deletes the latest 3 editors (2, 6, 5) and replaces the oldest editor (1).

---

### Use case 4:
If I have the following editors opened in order:
1, 2, 3 (1 being the oldest, focus is on 3)

- Set editor limit to 3
- Open editor 4
- The current available editors are 4, 2, 3

It replaces the oldest editor (1).

---

### Use case 5:
If I have the following editors opened in order:
1, 2, 3 (1 being the oldest, focus is on 1)

- Set editor limit to 3
- Open editor 4
- The current available editors are 1, 4, 3

It replaces the oldest editor (2).

praveen-skp:
It looks like its not always the oldest editors that are being closed. 
So shall we change the text to something generic like: 
Close editor(s) if there are more than ... editors open

BeckerWdf:
> Close editor(s) if there are more than ... editors open

yes.

BeckerWdf:
> > Close editor(s) if there are more than ... editors open
> 
> yes. Maybe  without the () simply "Close editors if..."

praveen-skp:
Reduced size

Reduced the size of the input field to show two characters, to match the text limit of 2 characters in the code.

BeckerWdf:
merge commit are not allowed in PR. Can you pls. squash and rebase on top of current master please?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
```diff
@@ -493,12 +493,13 @@ OpenPerspectiveDialogAction_tooltip=Open Perspective
...
 PreferencePage_noDescription = (No description available)
-PreferencePageParameterValues_pageLabelSeparator = \ >\ 
+PreferencePageParameterValues_pageLabelSeparator = \ >\
 ThemingEnabled = E&nable theming
 ThemeChangeWarningText = Restart for the theme changes to take full effect
-ThemeChangeWarningTitle = Theme Changed 
+ThemeChangeWarningTitle = Theme Changed
 # --- Workbench -----
 WorkbenchPreference_openMode=Open mode
+WorkbenchPreference_closeMode=Close mode
```
BeckerWdf: we don't need this text any more. Pls. remove it here and from the messages class

BeckerWdf (changes requested):

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
BeckerWdf: I did remove it

BeckerWdf (approved):
I plan to merge this once CI-Build is happy

PR #2299, opened 2024-09-19
Chevron: filter with '*' in front

jukzi:
image

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
@@ -403,6 +403,9 @@ public abstract class AbstractTableInformationControl {
 			fSearchPattern = null;
 		} else {
 			SearchPattern patternMatcher = new SearchPattern();
+			if (!pattern.startsWith("*")) { //$NON-NLS-1$
+				pattern = "*" + pattern; //$NON-NLS-1$
+			}
 			patternMatcher.setPattern(pattern);
 			fSearchPattern = patternMatcher;
 		}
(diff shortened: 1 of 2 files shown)


jukzi:
A second opinion about this change would be appreciated.

BeckerWdf:
> A second opinion about this change would be appreciated.

Sure. Can you please describe the problem you want to fix? "Chevron: filter with '*' in front" is not very speaking..

jukzi:
until now the filter finds only files starting with the String:
image
and would not find others:
image
That behaviour may be OK for Open-Type Dialog which filters from thousands of classes, but there are at max 99 items to filter in the chevron and i often find that i would have liked it to find the `*Test` too

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
```diff
@@ -403,6 +403,9 @@ private void setMatcherString(String pattern) {
 			fSearchPattern = null;
 		} else {
 			SearchPattern patternMatcher = new SearchPattern();
+			if (!pattern.startsWith("*")) { //$NON-NLS-1$
+				pattern = "*" + pattern; //$NON-NLS-1$
```
BeckerWdf, line 407: Change looks good to me.
But one question: Is this class only used for the Cevron-List? If not we may create unwanted side effects..

jukzi: As far as i have searched yes. But i wonder how many options those classes have even though they are internal and only used once:
image

BeckerWdf (approved):

jukzi:
ignoring random failing PartRenderingEngineTests https://github.com/eclipse-platform/eclipse.platform.ui/issues/751

PR #2309, opened 2024-09-20
Mw find next action use find replace logic

Wittmaxi:
Instead of re-implementing it's own logic, the FindNextAction now reuses
the FindReplaceLogic for it's backend.


Wittmaxi:
this PR is too broad to be merged untested. Adding unit tests is not as easy since there currently exists no test infrastructure for the FindNExtAction

Issue #2176, opened 2024-08-09
File->'Close Editor'  menu doesn't work with e4 editors.


eobrienPilz:
## Steps to reproduce

To reproduce you will need to define an editor part in an e4 fragment.
It should be marked as closable and given the 'Editor' tag.
Run the code and launch the editor
The Close menu option is available as a right click popup on the editor tab and it works.

![image](https://github.com/user-attachments/assets/e69cd8af-d8e7-4f12-8522-97e3f735098e)

However the File->Close Editor menu option is disabled.

![image](https://github.com/user-attachments/assets/27e0fd9d-b2d5-4302-95ba-6b37af687749)

From debuging I can see that the handler code (CloseEditorHandler) doesn't consider e4 editors.
It expects IEditorPart and checks that ISources.ACTIVE_EDITOR_NAME is set on the evaluation context.
E4 editors dont use IEditorPart or set ISources.ACTIVE_EDITOR_NAME.

From a quick prototype it seems to work for all editors (e4 and legacy) if the menu handlers are defined using an e4 fragment.

e.g. This seems to work for both E4 and legacy editors.

public class CloseActiveEditorHandlerE4 {
	@CanExecute
	boolean canExecute(@Named(IServiceConstants.ACTIVE_PART) MPart part) {
		boolean isEditor = part.getTags().stream().filter(t -> t.equals(Workbench.EDITOR_TAG)).count() > 0;
		return isEditor && part.isCloseable();	
	}

	@Execute
	void execute(EPartService partService, @Named(IServiceConstants.ACTIVE_PART) MPart part) {
		partService.hidePart(part, false);
	}
}

The same issue exists with File->'Close All Editors'.

feilimb:
Pull request created to address this: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2315

PR #2315, opened 2024-09-23
Modify the 'Close Editor' handler and enabled when evaluation to supp…

feilimb:
Modify the 'Close Editor' handler and '_enabledWhen_' evaluation to add support for both compatibility-layer editor parts, _and_ new Parts contributions which represent an Editor and are contributed via eg. a PartDescriptors in a Model Fragment addition.

Associated with Issue#2176 (https://github.com/eclipse-platform/eclipse.platform.ui/issues/2176).

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/CloseEditorHandler.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/CloseEditorHandler.java
@@ -14,16 +14,21 @@
 
 package org.eclipse.ui.internal;
 
+import java.util.List;
+import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
-import org.eclipse.core.expressions.EvaluationResult;
-import org.eclipse.core.expressions.Expression;
-import org.eclipse.core.expressions.ExpressionInfo;
-import org.eclipse.core.expressions.IEvaluationContext;
+import org.eclipse.e4.core.contexts.IEclipseContext;
+import org.eclipse.e4.ui.model.application.ui.basic.MPart;
+import org.eclipse.e4.ui.workbench.IWorkbench;
+import org.eclipse.e4.ui.workbench.modeling.EPartService;
 import org.eclipse.ui.IEditorPart;
-import org.eclipse.ui.ISources;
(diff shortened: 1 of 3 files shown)


feilimb:
Could anyone pick this up? @jukzi ? @iloveeclipse ?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/CloseEditorHandler.java
```diff
@@ -33,43 +33,42 @@
...
-		registerEnablement();
-	}
+public class CloseEditorHandler extends AbstractHandler {
 
 	@Override
-	public Object execute(ExecutionEvent event) throws ExecutionException {
-		IWorkbenchWindow window = HandlerUtil.getActiveWorkbenchWindowChecked(event);
-		IEditorPart part = HandlerUtil.getActiveEditorChecked(event);
-		window.getActivePage().closeEditor(part, true);
+	public Object execute(ExecutionEvent event) {
```
mickaelistria: IMO, we should always try to look in the event to ensure we get the right editor. Other strategies can have flaws and return another editor than the one that triggered the event.
Would it be possible to get the e4 element (MPart) from the event context instead?

If not, I think it would be worth keeping usage of `HandlerUtil.getActiveEditorChecked(event)` if it's not null, and if null, then failback to an alternative strategy.

feilimb: Thanks for the feedback - yes that makes sense, I have updated the PR now such that the event context is used to retrieve the active part.  If the active part is an IEditorPart I perform the previous logic, else if it is an E4Wrapper such as when we contribute a PartDescriptor via a model fragment it is dealt with via the EPartService.

mickaelistria:
Thanks for the change, this looks much better to me.
Let's try to merge it as soon as CI completes (particularly Jenkins build). I have rebased the commit to increase the chances of success.

feilimb:
> Thanks for the change, this looks much better to me. Let's try to merge it as soon as CI completes (particularly Jenkins build). I have rebased the commit to increase the chances of success.

Many thanks Mickael, lets see how the CI jobs go this time.

feilimb:
Also, just a heads up : I may look into a similar approach in a separate PR for handling the '_Close Active Editors_' (plural), if I ever get that far I might tag you in that PR if that's OK.

PR #2317, opened 2024-09-23
Replace o.e.test.performance dep with o.e.test

akurtakov:
The only reason to depend on o.e.test.performance bundle in these was the use of Screenshots class which should have not lived in performance test ever.


akurtakov:
Cleanup towards fixing https://github.com/eclipse-platform/eclipse.platform/issues/1564

akurtakov:
Build is marked unstable due to unrelated new warnings AFAICT.

Issue #2323, opened 2024-09-25
'Restore Defaults' button uses a different value when USE_MARKER_LIMITS preference is false.


eobrienPilz:
The FiltersConfigurationDialog has a 'Restore Defaults' button to restore the dialog to its default state.

![image](https://github.com/user-attachments/assets/f294d085-4284-471a-923c-4ddcbfa13e4f)

If USE_MARKER_LIMITS preference is false then 'items per group' is restored to 1000.

`int markerLimits = useMarkerLimits ? preferenceStore.getInt(IDEInternalPreferences.MARKER_LIMITS_VALUE) : 1000;`

This is in contrast to the default value for 'items per group' of 100 when Eclipse is first launched. 

`node.putInt(IDEInternalPreferences.MARKER_LIMITS_VALUE, 100);`

The restore default button should restore the 'items per group' to the same value as on first launch.

* note : this issue is not normally reproducible in Eclipse as USE_MARKER_LIMITS preference is true by default. 
It can be seen when USE_MARKER_LIMITS preference is being set to false via plugin_customization.ini for example.

deepika-u:
@eobrienPilz 
How do i get to FiltersConfigurationDialog? Can you please share the steps to recreate the problem as such.

eobrienPilz:
@deepika-u 
Steps to reproduce

- Add the following line to org.eclipse.platform\plugin_customization.ini	
	- org.eclipse.ui.ide/USE_MARKER_LIMITS=false
- Add the following line to the VM arguments section of the launch configuration
	-Declipse.pluginCustomization="path to the code/git/eclipse.platform/platform/org.eclipse.platform/plugin_customization.ini"
- Launch eclipse from the launch configuration
- Open the problem view
- Click the filters button

 ![image](https://github.com/user-attachments/assets/1855075d-922b-4391-b708-a91dbd16466b)
- Note that use limits is unchecked and the Items per group is disabled
- Note that Items per group default is set to 100
- Click 'Restore Defaults'
- Note that Items per group is set to 1000

eobrienPilz:
@deepika-u  Let me know if you need anything else to reproduce the issue.

deepika-u:
Hi @eobrienPilz,
I have my eclipse installation at C:\e\i\eclipse-SDK-I20240601-0610_4.32rc2a

[...]

PR #2325, opened 2024-09-25
#2323 'Restore Defaults' uses different value when USE_MARKER_LIMITS pref is false.

eobrienPilz:
Proposed fix for issue #2323 is to use 100 as the default value on first launch and also on pressing the 'Restore Defaults' button.


eobrienPilz:
Is anyone available to review this ? 
Its just a small fix and shouldn't have any side effects.

merks:
I wonder if this really the correct fix.  Isn't it possible for a product to override defaults?  

Should it maybe be like this?

```
int markerLimits = preferenceStore.getInt(IDEInternalPreferences.MARKER_LIMITS_VALUE);
```

Indeed, this looks like an example of such customization:

image

eobrienPilz:
Yes, I am not sure if there is a reason why pressing restore defaults uses a hard coded value. I'll see if I can change it so that it always uses the stored default like you suggest.

merks:
Yes, the 1000 is definitely fishy.  😕

eobrienPilz:
I made the update and tested it.  The restore defaults button now sets the the stored preference value, as expected.

merks:
🙏

PR #2326, opened 2024-09-25
Set up workflow for automated version increments in pull-requests 

HannesWell:
Set up for this repository the workflow for automated version increments in pull-requests added via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352

Before this can work complete, the EF infra team has to add the required PAT of the Platform-bot to this repo as described in the linked PR and requested in
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5020#note_2825941

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,5 @@ on:
     branches: [ master ]
 
 jobs:
-  check-freeze-period:
-    if: github.event_name == 'pull_request'
-    uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/verifyFreezePeriod.yml@master
-  check-merge-commits:
-    if: github.event_name == 'pull_request'
-    uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/checkMergeCommits.yml@master
   build:
     uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/mavenBuild.yml@master
-
--- a/.github/workflows/version-increments.yml
+++ b/.github/workflows/version-increments.yml
@@ -1,0 +1,14 @@
+name: Publish Version Check Results
+
+on:
+  workflow_run:
+    workflows: [ 'Pull-Request Checks' ]
+    types: [ completed ]
+
+jobs:
+  publish-version-check-results:
+    uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/publishVersionCheckResults.yml@master
+    with:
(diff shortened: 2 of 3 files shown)


.github/workflows/version-increments.yml
```diff
@@ -0,0 +1,14 @@
+name: Publish Version Check Results
+
+on:
+  workflow_run:
+    workflows: [ 'Pull-Request Checks' ]
```
jukzi, line 5: So it does  wait for all jobs and not only for "check-versions"`?

.github/workflows/ci.yml
```diff
@@ -10,12 +10,5 @@ on:
     branches: [ master ]
 
 jobs:
```
jukzi, line 12: is an empty "jobs:" section relevant anymore?

jukzi:
It would be nice if this could be tested with significant PR before rolling it out to every repo.
With significant PR i mean something non trivial like
* needs multiple versions bumps
* not all needed version bumps are reported in the first build
* as a result a feature needs to be updated.

.github/workflows/version-increments.yml
HannesWell: Exactly.
As far as I can tell only the completion of an entire workflow triggers an event on which another workflow can run:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

If you now a way to react on the completion of a single job in a new workflow that can run with access to secrets, please let me know.
But all other jobs should run much faster, so it's not really a problem atm.

.github/workflows/ci.yml
HannesWell: It's not empty, there is still the `build` job.

HannesWell:
https://github.com/eclipse-pde/eclipse.pde/pull/1413

> It would be nice if this could be tested with significant PR before rolling it out to every repo.
> With significant PR i mean something non trivial like

That has been done extensively and it worked well in:
- https://github.com/eclipse-pde/eclipse.pde/pull/1413
- https://github.com/eclipse-pde/eclipse.pde/pull/1414

and before that in my PDE fork:
https://github.com/HannesWell/eclipse.pde/pulls

Of course committers should still check the entire content of a PR as usual.

HannesWell:
The PAT has been added via https://github.com/eclipse-equinox/.eclipsefdn/pull/1, so this is ready for submission.

For `eclipse.platform` I have already verified it works.

PR #2327, opened 2024-09-25
Find/replace dialog: properly update search string #2322

### LINKED ISSUE #2322: [Find/Replace] changed search string is ignored by search dialog
I believe this is a regression in the recent code related to Find/Replace.

I'm using "old" Find/Replace dialog, latest SDK build I20240923-1800.

1) Open FindReplaceDialog.java in the IDE
2) Select `findReplaceLogic` at line 83 and trigger Ctrl+F to open find/replace dialog
3) Click once on Find => should jump to line 107 and highlight `findReplaceLogic` string (OK).
4) In the find/replace dialog, type "." after `findReplaceLogic` search text and click on "Find"
5) The dialog **continues to search with the old search string** and selects next `findReplaceLogic` occurrence on line 108.
6) Change search direction to "backward" and click "Find" two times => we are at line 83 again, where **no match** for `findReplaceLogic.` should be found!

image

HeikoKlare:
The find/replace dialog does currently not properly update its search string. When changing the search string and pressing return to execute a search, the previous search string is taken. The reason is an existing workaround for event handling of Combos in GTK.

This change ensures that the search and replace input strings are properly updated upon every text modification in the according widgets.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2322

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -139,12 +139,12 @@ class FindReplaceDialog extends Dialog {
 
 		@Override
 		public void modifyText(ModifyEvent e) {
+			modificationHandler.run();
 			// XXX: Workaround for Combo bug on Linux (see bug 404202 and bug 410603)
 			if (fIgnoreNextEvent) {
 				fIgnoreNextEvent = false;
 				return;
 			}
-			modificationHandler.run();
 			evaluateFindReplaceStatus();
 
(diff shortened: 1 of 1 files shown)


Wittmaxi:
Looks sound, however I wonder if it wouldn't be better to completely get rid of the combo boxes and use the input-element that the overlay uses.

1) we get reusable history features and do less inside of the Dialog itself
2) the guard-clause for the combo-box bug becomes irellevant - implicitly getting rid of the bug

HeikoKlare:
That might be a follow-up change. Still it has the risk of introducing regressions in the dialog, so we need to weigh that against the benefits of making that change. Aliging the two UIs makes sense, but we should ensure that the existing dialog remains usable in the same way as now for everyone sticking to it.

PR #2330, opened 2024-09-26
[FindNextAction] Introduce unit tests

Wittmaxi:
Introduces unit tests for the FindNextAction.
I don't achieve 100% test coverage, but I test for the most important parts. Appart from the "naive" functionality, in particular I test for:
1) search options are taken into account
2) the search history is honored
3) whole - word search is only performed if the selection is a word
4) RegExes are correctly escaped

### DIFF
--- a/tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/editors/tests/FindNextActionTest.java
+++ b/tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/editors/tests/FindNextActionTest.java
@@ -1,0 +1,263 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Vector Informatik GmbH - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.ui.editors.tests;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+
+import java.util.ResourceBundle;
+
+import org.hamcrest.Matchers;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.osgi.framework.FrameworkUtil;
+
+import org.eclipse.core.runtime.CoreException;
+
+import org.eclipse.core.resources.IFile;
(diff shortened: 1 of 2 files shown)


Wittmaxi:
The test is now in the editors.test package. Reason is that I need to start a whole editor in order to test the FindNextAction using IDE.openEditor. Depending on IDE in the workbench.editors package is problematic and such @HeikoKlare and I convened on moving the test to a more sensible place.

tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/editors/tests/FindNextActionTest.java
```diff
@@ -0,0 +1,253 @@
...
+	private static ResourceBundle fgBundleForConstructedKeys= ResourceBundle.getBundle(BUNDLE_FOR_CONSTRUCTED_KEYS);
+
+	public void openEditorAndFindNextAction(String content, boolean direction) {
+		try {
+			IFolder folder= ResourceHelper.createFolder("EncodingChangeTestProject/EncodingChangeTests/");
+			fFile= ResourceHelper.createFile(folder, "file" + fCount + ".txt", content);
+			fFile.setCharset(null, null);
+			fCount++;
+		} catch (CoreException e) {
+			fail();
```
akurtakov: Please no such empty fail() calls, it becomes a nightmare figuring out what happened when such a fail occurs. As a bare minimum pass e.getMessage() but it's even better to not catch the exception but let the method throws it and let JUnit fail the test itself with the message and full stack.

tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/editors/tests/FindNextActionTest.java
```diff
@@ -0,0 +1,253 @@
...
+		}
+		IWorkbench workbench= PlatformUI.getWorkbench();
+		IWorkbenchPage page= workbench.getActiveWorkbenchWindow().getActivePage();
+		try {
+			fEditor= IDE.openEditor(page, fFile);
+		} catch (PartInitException e) {
+			fail();
```
akurtakov: Likewise.

tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/editors/tests/FindNextActionTest.java
HeikoKlare: Yes, these fail statements are actually completely unnecessary. I am currently updating the PR accordingly (and improving it in general).

Wittmaxi:
@HeikoKlare thank you for updating the PR. I had copied over the method for opening an editor from another test - yet I agree that failing siletnyl is not a good idea.

HeikoKlare:
Thank you for adding the tests, @Wittmaxi! They are executed and succeed as expected.
Other failing tests are unrelated and documented: #751, #370

PR #2333, opened 2024-09-27
Remove reflective access from find/replace tests #2060

### LINKED ISSUE #2060: Find/Replace Overlay unused fields
In this PR (https://github.com/eclipse-platform/eclipse.platform.ui/pull/1990), some of the setBackground operations are useless and only there so that I don't get a warning for unused fields. I need those fields to stay in the class so I can access them in a test.
What is the best solution here?

HeikoKlare:
The tests for the FindReplaceDialog and FindReplaceOverlay currently use reflection to access specific UI elements. This ties the test implementations to implementation details of the production classes (i.e., specific hidden field to be present) and particularly requires the production code to contain (hidden) fields even if they would not be required just to provide according tests.

This change replaces the reflective access (at the code level) with widget extraction functionality based on unique information about the widget to be searched for. This may also be considered reflective (at the configuration level), but at least it gets rid of requiring the code to contain specific fields just in order to write tests that need to extract them.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2060

This is also a preparation for:
- #2254

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -104,6 +104,8 @@ public class FindReplaceOverlay extends Dialog {
 				KeyStroke.getInstance(SWT.MOD1, 'R'), KeyStroke.getInstance(SWT.MOD1, 'r'));
 	}
 
+	public static final String ID_DATA_KEY = "org.eclipse.ui.internal.findreplace.overlay.FindReplaceOverlay.id"; //$NON-NLS-1$
+
 	private static final String REPLACE_BAR_OPEN_DIALOG_SETTING = "replaceBarOpen"; //$NON-NLS-1$
 	private static final double WORST_CASE_RATIO_EDITOR_TO_OVERLAY = 0.95;
 	private static final double BIG_WIDTH_RATIO_EDITOR_TO_OVERLAY = 0.7;
@@ -130,9 +132,9 @@ public class FindReplaceOverlay extends Dialog {
 	private ToolItem wholeWordSearchButton;
 	private ToolItem caseSensitiveSearchButton;
 	private ToolItem regexSearchButton;
-	private ToolItem searchUpButton;
-	private ToolItem searchDownButton;
-	private ToolItem searchAllButton;
+	private ToolItem searchBackwardButton;
+	private ToolItem searchForwardButton;
+	private ToolItem selectAllButton;
 	private AccessibleToolBar closeTools;
 	private ToolItem closeButton;
 
@@ -370,6 +372,7 @@ public class FindReplaceOverlay extends Dialog {
 		}
 		overlayOpen = true;
 		applyOverlayColors(backgroundToUse, true);
+		assignIDs();
 		updateFromTargetSelection();
 		searchBar.forceFocus();
 
@@ -391,6 +394,25 @@ public class FindReplaceOverlay extends Dialog {
 		}
 	}
 
+	@SuppressWarnings("nls")
+	private void assignIDs() {
+		replaceToggle.setData(ID_DATA_KEY, "replaceToggle");
+		searchBar.setData(ID_DATA_KEY, "searchInput");
+		searchBackwardButton.setData(ID_DATA_KEY, "searchBackward");
+		searchForwardButton.setData(ID_DATA_KEY, "searchForward");
+		selectAllButton.setData(ID_DATA_KEY, "selectAll");
+		searchInSelectionButton.setData(ID_DATA_KEY, "searchInSelection");
+		wholeWordSearchButton.setData(ID_DATA_KEY, "wholeWordSearch");
+		regexSearchButton.setData(ID_DATA_KEY, "regExSearch");
+		caseSensitiveSearchButton.setData(ID_DATA_KEY, "caseSensitiveSearch");
+
+		if (replaceBarOpen) {
+			replaceBar.setData(ID_DATA_KEY, "replaceInput");
+			replaceButton.setData(ID_DATA_KEY, "replaceOne");
+			replaceAllButton.setData(ID_DATA_KEY, "replaceAll");
+		}
+	}
+
 	private void applyOverlayColors(Color color, boolean tryToColorReplaceBar) {
 		closeTools.setBackground(color);
 		closeButton.setBackground(color);
@@ -400,9 +422,9 @@ public class FindReplaceOverlay extends Dialog {
 		wholeWordSearchButton.setBackground(color);
 		regexSearchButton.setBackground(color);
 		caseSensitiveSearchButton.setBackground(color);
-		searchAllButton.setBackground(color);
-		searchUpButton.setBackground(color);
-		searchDownButton.setBackground(color);
+		selectAllButton.setBackground(color);
+		searchBackwardButton.setBackground(color);
+		searchForwardButton.setBackground(color);
 
 		searchBarContainer.setBackground(color);
 		searchBar.setBackground(color);
@@ -511,20 +533,20 @@ public class FindReplaceOverlay extends Dialog {
 
 		searchTools.createToolItem(SWT.SEPARATOR);
 
-		searchUpButton = new AccessibleToolItemBuilder(searchTools).withStyleBits(SWT.PUSH)
+		searchBackwardButton = new AccessibleToolItemBuilder(searchTools).withStyleBits(SWT.PUSH)
 				.withImage(FindReplaceOverlayImages.get(FindReplaceOverlayImages.KEY_FIND_PREV))
 				.withToolTipText(FindReplaceMessages.FindReplaceOverlay_upSearchButton_toolTip)
 				.withOperation(() -> performSearch(false))
 				.withShortcuts(KeyboardShortcuts.SEARCH_BACKWARD).build();
 
-		searchDownButton = new AccessibleToolItemBuilder(searchTools).withStyleBits(SWT.PUSH)
+		searchForwardButton = new AccessibleToolItemBuilder(searchTools).withStyleBits(SWT.PUSH)
 				.withImage(FindReplaceOverlayImages.get(FindReplaceOverlayImages.KEY_FIND_NEXT))
 				.withToolTipText(FindReplaceMessages.FindReplaceOverlay_downSearchButton_toolTip)
 				.withOperation(() -> performSearch(true))
 				.withShortcuts(KeyboardShortcuts.SEARCH_FORWARD).build();
-		searchDownButton.setSelection(true); // by default, search down
+		searchForwardButton.setSelection(true); // by default, search down
 
-		searchAllButton = new AccessibleToolItemBuilder(searchTools).withStyleBits(SWT.PUSH)
+		selectAllButton = new AccessibleToolItemBuilder(searchTools).withStyleBits(SWT.PUSH)
 				.withImage(FindReplaceOverlayImages.get(FindReplaceOverlayImages.KEY_SEARCH_ALL))
 				.withToolTipText(FindReplaceMessages.FindReplaceOverlay_searchAllButton_toolTip)
 				.withOperation(this::performSelectAll).withShortcuts(KeyboardShortcuts.SEARCH_ALL).build();
@@ -759,6 +781,7 @@ public class FindReplaceOverlay extends Dialog {
 
 		updatePlacementAndVisibility();
 		applyOverlayColors(backgroundToUse, true);
+		assignIDs();
 		replaceBar.forceFocus();
 	}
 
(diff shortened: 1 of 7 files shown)


laeubi:
@HeikoKlare A usual way is to specify an ID in the Widget#setData (see also https://github.com/eclipse-platform/eclipse.platform.swt/discussions/549 ) so one can identify them in test. SWT Bot on the other hand also contains some useful helper methods in this regards.

Wittmaxi:
I do not yet see the added benefit here - can you please elaborate a bit?

I understand that we want to get rid of reflection. The reason we don't like reflection is because it adds a new predicate for the code to run - i.e. changing the class structure of FindReplaceOverlay changes the behavior of FindReplaceOverlayTest as a side-effect.
This change makes the test rely on the text that is set to the FindReplaceOverlay, instead of the class structure - i.e. changing the text of the FindReplaceOverlay changes the behavior of FindReplaceOverlayTest as a side-effect.

This effectively means that we get rid of one dependency to introduce another dependency.
Since the class is coupled with the test, it a future developer changing the FindReplaceOverlay class would see a test failing and think "oh, that makes sense. The design is bad, but I understand why this test fails".
On the other hand, having a unit test fail because I change the text of a tooltip might not be very intuitive for me, if I am not familiar with the code. This gets even worse if we hire e.g. a copywriter to write texts (... far fetched, sure).

@laeubi's approach of setting IDs sounds more sound to me, but I think I am missing some information here: what is the (true) problem that we are trying to solve here?

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/WidgetExtractor.java
```diff
@@ -0,0 +1,90 @@
+package org.eclipse.ui.internal.findandreplace;
```
akurtakov, line 14: Missling licenseheader. Please make sure it's there before pushing in.

HeikoKlare:
> I do not yet see the added benefit here - can you please elaborate a bit?

The benefit is that production code does not need to be aligned with test code anymore. Currently, production code needs to have the fields that are reflectively accessed to make the test work. With the proposed change, that's different as the tests use production information (texts, tooltips) to identify some widget. So yes, if you change the production code you may need to adapt tests (but that's somehow expected). But you do not need to keep otherwise unnecessary production code (e.g., unused fields) just to make the tests work.

> usual way is to specify an ID in the Widget#setData (see also [eclipse-platform/eclipse.platform.swt#549](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/549) ) so one can identify them in test. SWT Bot on the other hand also contains some useful helper methods in this regards.

In general, that seems to be a better approach as you tie to IDs instead of NLS, which are more prone to change. I am just a bit concerned to introduce IDs only for these UI elements, because that would make it similar to the existing approach as the IDs would not serve any other purpose than implementing tests: production code is adapted to test code (which requires IDs) instead of the other way round. I think we would need to have a general pattern for such IDs to rely on by convention instead of introducing a specific convention for this use case. So all approaches have drawbacks, but maybe adding IDs is the least bad one as it should at least be stable.

With respect to SWTBot, I remember a discussion that we do not want to use it in the platform projects (I think it's not even available via the TP), which is why I avoided using it. In general, having a proper UI testing framework instead of multiple use-case-specific solutions would of course be preferable.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/WidgetExtractor.java
HeikoKlare: Thanks for the hint!

Wittmaxi:
@HeikoKlare 
> So all approaches have drawbacks, but maybe adding IDs is the least bad one as it should at least be stable.

The stability of the id-approach is what sells it for me.

I had tried SWTBot and RedDeer and I am confident that neither are helpful to us. I remember having documented the drawbacks of either in an internal chat (I believe it was related to 1) deprecation 2) problems with inclusion into eclipse 3) neither really "worked")

laeubi:
> In general, that seems to be a better approach as you tie to IDs instead of NLS, which are more prone to change. I am just a bit concerned to introduce IDs only for these UI elements

Currently your test uses IDs already (you can also call it a "name" if you feel more comfortable with that), the NLS could probably change as well, it was just a hint that assign data (it must not be an String!) is often a very powerful approach that decouples the provider from the consumer, e.g. JFace uses that to connect the wrapper classes to the widget.

HeikoKlare:
You are right, the current proposal also uses IDs. Only difference is that the used ID information is required by production code and thus present anyway (as the widgets need to have some name/tooltip/...) while an "explicit" ID will (currently) only be added for tests. But it can, of course, also be used for other purposes. I will adapt the PR accordingly.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayTest.java
```diff
@@ -45,8 +45,8 @@ public class FindReplaceOverlayTest extends FindReplaceUITest<OverlayAccess> {
 	public OverlayAccess openUIFromTextViewer(TextViewer viewer) {
 		Accessor actionAccessor= new Accessor(getFindReplaceAction(), FindReplaceAction.class);
 		actionAccessor.invoke("showOverlayInEditor", null);
-		Accessor overlayAccessor= new Accessor(actionAccessor.get("overlay"), "org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlay", getClass().getClassLoader());
-		return new OverlayAccess(getFindReplaceTarget(), overlayAccessor);
+		FindReplaceOverlay overlay= (FindReplaceOverlay) actionAccessor.get("overlay");
```
laeubi, line 48:
```suggestion
		FindReplaceOverlay overlay= (FindReplaceOverlay) viewer.getControl().getData(FindReplaceOverlay.class.getName()
```

Just as an example on how one can "transport" an object through the data of the underlying control, of course the `FindReplaceOverlay` needs to set the data accordingly once it is installed in a `TextViewer` and remove itself once it is removed. That way it is very easy to check if a given `TextViewer` has a `FindReplaceOverlay` or not (not only for tests).

If this is needed on multiple places one can even has

`static Optional<FindReplaceOverlay> FindReplaceOverlay.of(TextViewer viewer)`

that encapsulates the details on how a `FindReplaceOverlay` know its connection to a `TextViewer`.

HeikoKlare: That looks like a great idea for further reducing reflective access of the tests (in this case related to the `FindReplaceAction` instead of the dialog/overlay itself) and maybe can also improve the `FindReplaceAction` itself because it could relieve the action from tracking the overlay by just attaching it to the target text editor. I would like to follow-up on this in a subsequent PR.

HeikoKlare:
I have now replaced the widget extraction based on NLS information with explicit IDs. Those IDs are simple strings that need to match between production code and tests as well. An alternative would be to not store the IDs in the individual widgets and traverse the widgets to identify them but to register the individual widgets with those IDs as data at the "root widget" of the overlay/dialog and directly retrieve them from that widget via those IDs. That would make the additional widget traversal logic obsolete but would then require the dialog/overlay classes to expose their root widget. With #2254 the FindReplaceOverlay will not be a dialog and not expose the root widget anymore but hide it is an implementation detail, which would then need to be exposed only for the sake of testability.

@laeubi what do you think of the current solution?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -391,6 +394,25 @@ private void restoreOverlaySettings() {
...
+		searchBar.setData(ID_DATA_KEY, "searchInput");
+		searchBackwardButton.setData(ID_DATA_KEY, "searchBackward");
+		searchForwardButton.setData(ID_DATA_KEY, "searchForward");
+		selectAllButton.setData(ID_DATA_KEY, "selectAll");
+		searchInSelectionButton.setData(ID_DATA_KEY, "searchInSelection");
+		wholeWordSearchButton.setData(ID_DATA_KEY, "wholeWordSearch");
+		regexSearchButton.setData(ID_DATA_KEY, "regExSearch");
+		caseSensitiveSearchButton.setData(ID_DATA_KEY, "caseSensitiveSearch");
+
+		if (replaceBarOpen) {
```
Wittmaxi, line 409: Careful: since this method is only being called at startup and on open, there might be a case when not every overlay component has an ID assigned;

1) toggle replace off
2) close and re-open the overlay
3) toggle replace on --> replaceBar does not have an ID set 

This method should likely be called from applyOverlayColors or additionally from toggleReplace

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -400,9 +422,9 @@ private void applyOverlayColors(Color color, boolean tryToColorReplaceBar) {
 		wholeWordSearchButton.setBackground(color);
 		regexSearchButton.setBackground(color);
 		caseSensitiveSearchButton.setBackground(color);
-		searchAllButton.setBackground(color);
-		searchUpButton.setBackground(color);
-		searchDownButton.setBackground(color);
+		selectAllButton.setBackground(color);
```
Wittmaxi, line 425: Side-note: I have always been confused about whether we "search" or we "find" or we "select". I tried being consistent but I feel like there are more places where we have inconsistencies.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/WidgetExtractor.java
```diff
@@ -0,0 +1,84 @@
...
+		return findWidget(rootContainer, ToolItem.class, id);
+	}
+
+	private <T extends Widget> T findWidget(Composite container, Class<T> type, String id) {
+		List<T> widgets= findWidgets(container, type, id);
+		assertFalse("more than one matching widget found for id '" + id + "':" + widgets, widgets.size() > 1);
+		return widgets.isEmpty() ? null : widgets.get(0);
+	}
+
+	@SuppressWarnings("unchecked")
```
Wittmaxi: can we narrow down the unchecked warning into a smaller method?

```java
	private <T extends Widget> List<T> findWidgets(Composite container, Class<T> type, String id) {
		List<Widget> children= new ArrayList<>();
		children.addAll(List.of(container.getChildren()));
		if (container instanceof ToolBar toolbar) {
			children.addAll(List.of(toolbar.getItems()));
		}
		List<T> result= new ArrayList<>();
		for (Widget child : children) {
			if (type.isInstance(child)) {
				if (id.equals(child.getData(idDataKey))) {
					addChildIntoList(result, child);
				}
			}
			if (child instanceof Composite compositeChild) {
				result.addAll(findWidgets(compositeChild, type, id));
			}
		}
		return result;
	}

	@SuppressWarnings("unchecked")
	private <T extends Widget> void addChildIntoList(List<T> result, Widget child) {
		result.add((T) child);
	}
```

This might be overly verbose, but I believe it adds the benefit of being very explicit about how and which warning we violate - and it makes sure no other warning in findWidgets is being unintentionally ignored

Wittmaxi (approved):
Great improvement of the tests! I am not 100% sure that assignIDs in FindReplaceOverlay is correct.
From my static analysis this is the current status:
For the current tests it is certainly good enough, but a future tester might(should!) assume that it assigns IDs to every component - which it currently does not necessarily. The issue arises when the replaceBar is not open while opening the overlay. Since assignIDs is only called when opening the Overlay first, it will never react to toggling the replace bar open in a later stage of the overlay usage.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: It is called from `open()` and `openReplaceDialog()`, which covers all three cases.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: In particular for "search" and "find" there is probably no sharp distinction. I propose to simply resolve imprecise names whenever we find them.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/WidgetExtractor.java
HeikoKlare: Valid point. It would also be possible to simply attach the annotation to a single line instead of the whole method. But since we have the class at hand anyway, I have replaced the line producing a warning with a cast operation of that class.

HeikoKlare:
Failing tests are unrelated and already documented: #2346

Issue #2329, opened 2024-09-26
Add dropdown history to the "in" field of the Quick Search dialog window


mplukas:
It would be very helpful if the Quick Search dialog window (Ctrl+Alt+Shift+L) kept a [short] history of the values used in the **"in"** field and allowed to select from it (similar like "File name patterns" field in the regular Search window (Ctrl+H)).

jukzi:
Will you contribute such PR?

Wittmaxi:
Hey, you might want to look into the HistoryTextWrapper class. By re-using it, implementing this should be ~very easy~ doable.
The plan: replace the input box in the quick search with that HistoryTextWrapper class or something similar.

PValdovinos:
I'm working on this :)

HannesWell:
> I'm working on this :)

Great. You are one of the students from CodeDays, aren't you?

Much success on this task and don't hesitate to ask for help if you need any.

jannisCode:
Because I did not hear anything anymore and did not know if this is still worked on, I implemented the feature here:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2541
@HannesWell @Wittmaxi @jukzi Would you like to take a look at the new PR? 😅

jukzi:
> I'm working on this :)

@PValdovinos are you still working on this?

PR #2334, opened 2024-09-27
Quick Search - History for "search in"

Wittmaxi:
grafik

This PR can be a start for anybody willing to take on this issue - I think it is good enough "as is", but I would like somebody to double-check if adding the dependencies I added is okay

### DIFF
--- a/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
+++ b/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
@@ -237,9 +237,6 @@ public class QuickSearchDialog extends SelectionStatusDialog {
 	private Image getBlankImage() {
 		if (blankImage==null) {
 			blankImage = new Image(Display.getDefault(), 1, 1);
-//			GC gc = new GC(blankImage);
-//			gc.fillRectangle(0, 0, 16, 16);
-//			gc.dispose();
 		}
 		return blankImage;
 	}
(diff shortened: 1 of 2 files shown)


Wittmaxi:
#2329

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
```diff
@@ -811,6 +791,7 @@ public void getName(AccessibleEvent e) {
 
 		pattern.addModifyListener(e -> {
 			applyFilter(false);
+			searchIn.storeHistory();
```
jukzi, line 794: on every character typed?

Wittmaxi: Yes, because every new character typed starts a new search. Storing history is cheap, so I don't think that's a problem

jukzi: compare to https://github.com/eclipse-platform/eclipse.platform.ui/issues/2291

Wittmaxi: I believe you refer to
> Since the overlay has incremental search enabled by default (other than the existing dialog), adding every searched term to the history is inappropriate, as typing "test" would result in history entries "t", "te", "tes" and "test".

However, here we are doing something slightly differently: the search history is applied to the "search in" field - and we store the "search in" history every time some search is performed - while typing in the search bar, there is no change of the "search in" field. Thus, we save the same string to history for every keystroke. However, since we also remove all other occurrences of that exact string in the history, the operation does nothing.

Anyway, I don't want to invest too much into this PR. It is a good model for a new contributor wanting to play with this issue.

The main problems are the dependencies; I am not so sure about importing two internal packages

jukzi: please mark as draft

PR #2337, opened 2024-09-28
Find/replace overlay: improve replace toggle button appearance

HeikoKlare:
The button to toggle whether the replace bar in a find/replace overlay is shown currently appears to be a rather heavyweight button with a border. With this change, the button is replace with a lightweight toolbar item like used for all other buttons in the overlay. This improves the appearance as well as the implementation as the same implementation patterns can now be applied to all buttons in the overlay.

This is also a preparatory change for #2254 as with that change optmizing the appearance of the button would otherwise become difficult.

#### Windows
Before:
{97308575-0F78-4232-A151-3E93115F274C}
{D7A666D1-B597-46C2-A220-CD829ACBA500}
{FB7A0830-8AAE-4204-B2AC-BE2A9DFCCF3A}
{F2B46CA7-A9FD-4D6C-B2D5-A8C8B0710BB2}

After:
{4C0AFC57-EA34-4582-A506-50D820761F7C}
{5D4F867C-8ACA-44F8-A8E6-D317EE2D3E9D}
{4DC065F8-126A-476C-A732-1677DC81A31C}
{FBCCA0D9-14C9-47DA-8C86-0635C544B538}

#### MacOS
Before:
image
image
image
image

After:
image
image
image
image

#### Linux
Before:
image
image
image
image

After:
image
image
image
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItem.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolItem.java
@@ -22,6 +22,7 @@ import org.eclipse.swt.widgets.ToolBar;
 import org.eclipse.swt.widgets.ToolItem;
 
 import org.eclipse.jface.bindings.keys.KeyStroke;
+import org.eclipse.jface.layout.GridDataFactory;
 
 class AccessibleToolItem {
 	private final ToolItem toolItem;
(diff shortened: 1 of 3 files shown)


HeikoKlare:
@Wittmaxi what do you think? Is there anything speaking against this change, maybe any reason why this was not the solution in the beginning?

Wittmaxi:
I think this change looks good, improves code quality and makes maintenance easier in the future. Once the tests go green, this can be merged IMHO (is there a reason this is still a draft?)

grafik
Slight caveat: this outline doesn't really feel to me like it belongs there.

HeikoKlare:
> Slight caveat: this outline doesn't really feel to me like it belongs there.

Do you mean the missing margin to the right or the small size of the button in general? The former is fixed in the recent push. The latter is something I do not have a solution for since toolbars / toolbar items doe seems to be (easily) sizable.

HeikoKlare:
Let's give this a try.

PR #2339, opened 2024-09-30
Fix compilation error due unresolved org.eclipse.jdt.annotation

iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/issues/1424


iloveeclipse:
Couldn't reproduce with I20240930-1800 anymore

merks:
Very strange an annoying.  😢

PR #2344, opened 2024-09-30
Generated all OSGi Declarative Services component files by PDE-DS/Tycho

HannesWell:
Remove the now generated OSGi Declarative Services component-xml files from git.

Resolve wildcards in Service-Component header.

@laeubi do you want to have a look at this as well?

### DIFF
--- a/bundles/org.eclipse.e4.ui.progress/OSGI-INF/progress.xml
+++ b/bundles/org.eclipse.e4.ui.progress/OSGI-INF/progress.xml
@@ -1,8 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" name="org.eclipse.e4.ui.progress">
-   <implementation class="org.eclipse.e4.ui.progress.internal.ProgressServiceCreationFunction"/>
-   <service>
-      <provide interface="org.eclipse.e4.core.contexts.IContextFunction"/>
-   </service>
-   <property name="service.context.key" type="String" value="org.eclipse.e4.ui.progress.IProgressService"/>
-</scr:component>
--- a/bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/internal/ProgressServiceCreationFunction.java
+++ b/bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/internal/ProgressServiceCreationFunction.java
@@ -16,8 +16,13 @@ package org.eclipse.e4.ui.progress.internal;
 
 import org.eclipse.e4.core.contexts.ContextFunction;
 import org.eclipse.e4.core.contexts.ContextInjectionFactory;
+import org.eclipse.e4.core.contexts.IContextFunction;
 import org.eclipse.e4.core.contexts.IEclipseContext;
+import org.eclipse.e4.ui.progress.IProgressService;
+import org.osgi.service.component.annotations.Component;
 
+@Component(service = IContextFunction.class)
+@IContextFunction.ServiceContextKey(IProgressService.class)
 public class ProgressServiceCreationFunction extends ContextFunction {
 
 	@Override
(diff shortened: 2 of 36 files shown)


bundles/org.eclipse.e4.ui.progress/OSGI-INF/progress.xml
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" name="org.eclipse.e4.ui.progress">
-   <implementation class="org.eclipse.e4.ui.progress.internal.ProgressServiceCreationFunction"/>
-   <service>
-      <provide interface="org.eclipse.e4.core.contexts.IContextFunction"/>
-   </service>
-   <property name="service.context.key" type="String" value="org.eclipse.e4.ui.progress.IProgressService"/>
```
laeubi, line 7: We might want to define a component property type for this `@ServiceContectKey(Class))` this will avoid needing to manually set such property

bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/internal/ProgressServiceCreationFunction.java
```diff
@@ -16,8 +16,12 @@
 
 import org.eclipse.e4.core.contexts.ContextFunction;
 import org.eclipse.e4.core.contexts.ContextInjectionFactory;
+import org.eclipse.e4.core.contexts.IContextFunction;
 import org.eclipse.e4.core.contexts.IEclipseContext;
+import org.osgi.service.component.annotations.Component;
 
+@Component(service = IContextFunction.class, property = {
+		"service.context.key=org.eclipse.e4.ui.progress.IProgressService" })
```
laeubi: We might want to define a component property type for this @ServiceContectKey(Class)) this will avoid needing to manually set such property

bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/text/DirtyFileSearchParticipant.java
```diff
@@ -38,6 +40,7 @@
 
 import org.eclipse.search.internal.core.text.DirtyFileProvider;
 
+@Component(service = DirtyFileProvider.class, immediate = false, property = { "weight:Integer=100" })
```
laeubi, line 43: Maybe replace weight with ServiceRank?

HannesWell: Actually it would have been a good choice when this service was introduced. But now that it is defined as property to sort difference service implementations I would consider this an API and for now I don't want to implement support for both in `DirtyFileSearchParticipantServiceTracker`.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/04dc50b8b643a72416cb1779741827e36e9d2066/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/DirtyFileSearchParticipantServiceTracker.java#L36

bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/internal/ProgressServiceCreationFunction.java
HannesWell: Yes that would be handy here. Proposed it in
- https://github.com/eclipse-platform/eclipse.platform/pull/1577

HannesWell:
I have now found out why this leads to many more test-failures, by replying the build and capturing the generated DS component files and noticing that the properties of the component types are missing.

The reason is very likely that I defined `ServiceContextKey` with `RetentionPolicy.SOURCE` and thus the annotations are not recorded in the compiled class files. I did this because these annotations are not meant to be used at runtime, only by tools at compile time. But Tycho respectively BND operates on compiled classes so the annotations have to be present in the compiled classes...
I created a PR to fix that and will update all other pending ones:
- https://github.com/eclipse-platform/eclipse.platform/pull/1587

Looking at `org.osgi.service.event.propertytypes.EventTopics` CLASS should be sufficient (in the component XML files I captured during the builds that property worked):
https://github.com/osgi/osgi/blob/b8336992a313dc092b8332f0e3aae4cff216bb03/org.osgi.service.event/src/org/osgi/service/event/propertytypes/EventTopics.java#L42-L46

HannesWell:
Finally this works. The test-failures are unrelated.
Submitting.

PR #2352, opened 2024-10-02
IWorkbenchPageTest: delete project after shutdown

jukzi:
To avoid error messages about missing file after testOpenEditors3()


jukzi:
unrelated random fail testSingleSelectionAndTwoInitialSelectionsWithInitialPattern  https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

Issue #2367, opened 2024-10-08
NPE in ContributedPartRenderer$1.setFocus


jukzi:
logged in https://download.eclipse.org/eclipse/downloads/drops4/I20241007-1800/testresults/ep434I-unit-win32-java17_win32.win32.x86_64_17/org.eclipse.e4.ui.tests.UIAllTests.txt

```
!ENTRY org.eclipse.e4.ui.workbench 4 0 2024-10-07 23:41:23.155
!MESSAGE Error disposing widget for : org.eclipse.e4.ui.model.application.ui.basic.impl.WindowImpl null
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.e4.ui.workbench.IPresentationEngine.focusGui(org.eclipse.e4.ui.model.application.ui.MUIElement)" because "pe" is null
	at org.eclipse.e4.ui.workbench.renderers.swt.ContributedPartRenderer$1.setFocus(ContributedPartRenderer.java:106)
	at org.eclipse.swt.widgets.Decorations.restoreFocus(Decorations.java:763)
	at org.eclipse.swt.widgets.Decorations.WM_ACTIVATE(Decorations.java:1527)
	at org.eclipse.swt.widgets.Shell.WM_ACTIVATE(Shell.java:2374)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4722)
	at org.eclipse.swt.widgets.Canvas.windowProc(Canvas.java:342)
	at org.eclipse.swt.widgets.Decorations.windowProc(Decorations.java:1482)
	at org.eclipse.swt.widgets.Shell.windowProc(Shell.java:2347)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5155)
	at org.eclipse.swt.internal.win32.OS.DestroyWindow(Native Method)
	at org.eclipse.swt.widgets.Control.destroyWidget(Control.java:737)
	at org.eclipse.swt.widgets.Shell.destroyWidget(Shell.java:736)
	at org.eclipse.swt.widgets.Widget.release(Widget.java:906)
	at org.eclipse.swt.widgets.Widget.dispose(Widget.java:478)
	at org.eclipse.swt.widgets.Decorations.dispose(Decorations.java:396)
	at org.eclipse.e4.ui.workbench.renderers.swt.SWTPartRenderer.disposeWidget(SWTPartRenderer.java:186)
[...]

PR #2370, opened 2024-10-08
ContributedPartRenderer$1.setFocus: prevent NPE #2367

jukzi:
during PartRenderingEngineTests

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2367


jukzi:
@vogella can you please review, that code comes from you.

vogella:
I assume this is a test only scenario? IPresentationEngine cannot be null in a real application IIRC.

Change is fine for me.

jukzi:
> I assume this is a test only scenario?

yes, thanks

PR #2371, opened 2024-10-08
[tests] log a TestException instead of RuntimeException

jukzi:
nightly logfile is flooded with intentional RuntimeException that are hard to distinguish from unintentional exceptions.


akurtakov:
There is smth I don't understand here - What's the point in throwing this custom exception if nothing catches/checks it ? Wouldn't it be the same to not throw anything at all?

jukzi:
It will be logged as "TestException", and when i read the logfile i do not need to dig into if its a real failure. Is there any mechanisms that can prevent such errors from being logged? that would be even better.
As far as i understand those Tests tests that it works even if a customer throws an exception. The exceptions are catched and logged only.

PR #2374, opened 2024-10-09
[test] TestLabelProvider: reduce font not disposed warnings

jukzi:
to a single "SWT Resource was not properly disposed" during LabelProviderTest

(and fixed the font to be not bold)


jukzi:
> Therefore the following files need a version increment:

That bot is really helpful. thanks @HannesWell

HannesWell:
> > Therefore the following files need a version increment:
> 
> That bot is really helpful. thanks @HannesWell

You're welcome and I'm glad it's helpful. I always have a little joy if I see that bot's message in an email. :)

PR #2386, opened 2024-10-11
Increase InternalDialog width hint from 50 characters to 70 characters

merks:
At 50 characters, the dialog almost never is able to display a message without wrapping:

image

Even at 70 characters it's still small but at least there is a better chance that the text will fit:

image


laeubi:
@merks not sure if you want to even more improve the situation, I think it would be good to fetch the monitor for the shell then get the size and use a percentage based approach (maybe with a min + max value) instead of a fixed value maybe.

merks:
I'm not sure I want to go down that rabbit hole.  😜 

Search the regex `convertWidthInCharsToPixels\([0-9]` in the SDK workspace and find the 250+ matches:

image

The design intent in general is to display the specified/expected number of average characters which of course scales properly depending on fonts and resolutions.

If we want to make improvements, this seldom-seen dialog is probably not the most important place to focus such improvements, and we'd probably want provide support for such improvements in a central location, rather than distributed across the framework.

PR #2387, opened 2024-10-11
[WIP] [E4 Xpath] Replace apache.commons.jxpath by javax.xml.xpath

ptziegler:
This is a continuation of #2290 and attempts to perform an XPath evaluation of the EObjects via the native javax.xml.xpath API rather than relying on JXPath. The Java model is converted to an XML structure where each EObject is mapped to a DOM node. The XPath is applied on the root node and the resulting nodes are then converted back to EObjects.


ptziegler:
@HannesWell As the tests will hopefully show, the current state should support generic XPaths such as `//mainMenu/children` and also properly handle the parent context (albeit untested). The only thing that doesn't work the same way is `/`.

HannesWell:
Nevertheless, thank you for your help!

PR #2393, opened 2024-10-12
PreferencePageParameterValues_pageLabelSeparator add trailing space

merks:
- This property is supposed to have a training space which was specified by `\ ` but the space as removed which then continues the property value onto the next line which is not the desired effect and results in the ThemingEnabled property being missing.


merks:
There seems to be an avalanche of failures, but I'm doubtful any of  it is related to this trivial change.  Does anyone know what's going on?

laeubi:
Seems to be around

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2449
- 
maybe?

merks:
I'm going to assume that while there major problems, they are not related to adding a space back from where it was recently remove.

PR #2396, opened 2024-10-13
Add component property types for IModelProcessorContribution attributes

HannesWell:
These annotation simplify the specification of the 'beforefragment' and 'apply' service property for
IModelProcessorContribution implementations and makes it type-safe and more robust:
```
@Component(service = IModelProcessorContribution.class)
@IModelProcessorContribution.Beforefragment(true)
@IModelProcessorContribution.Apply(IModelProcessorContribution.APPLY_ALWAYS)
public class SampleContribution implements IModelProcessorContribution { 
```

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/workbench/modeling/IModelProcessorContribution.java
+++ b/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/workbench/modeling/IModelProcessorContribution.java
@@ -14,8 +14,11 @@
 
 package org.eclipse.e4.ui.workbench.modeling;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
 import java.util.Collections;
 import java.util.List;
+import org.osgi.service.component.annotations.ComponentPropertyType;
 
 /**
  * Service component interface to be able to register model processors via OSGi
@@ -32,7 +35,7 @@ public interface IModelProcessorContribution {
 	/**
 	 * Service property key for specifying the beforeFragment attribute, which
 	 * specifies if the processor has to be invoked before model fragments are
-	 * added. If not specified it defaults to <code>true</code>.
+	 * added. If not specified it defaults to {@code true}.
 	 */
 	String BEFORE_FRAGMENT_PROPERTY_KEY = "beforefragment"; //$NON-NLS-1$
 	/**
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/workbench/modeling/IModelProcessorContribution.java
```diff
@@ -44,11 +47,33 @@ public interface IModelProcessorContribution {
...
+	 * invoked before model fragments are added. If not specified it defaults to
+	 * {@code true}.
+	 *
+	 * @since 1.16
+	 * @see #BEFORE_FRAGMENT_PROPERTY_KEY
+	 */
+	@ComponentPropertyType
+	@Retention(RetentionPolicy.SOURCE)
+	@interface Beforefragment {
+		boolean value();
```
laeubi:
```suggestion
		boolean value() default true;
```

bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/workbench/modeling/IModelProcessorContribution.java
```diff
@@ -59,8 +84,30 @@ public interface IModelProcessorContribution {
...
+	 * @since 1.16
+	 * @see #APPLY_PROPERTY_KEY
+	 * @see #APPLY_ALWAYS
+	 * @see #APPLY_INITIAL
+	 *
+	 */
+	@ComponentPropertyType
+	@Retention(RetentionPolicy.SOURCE)
+	@interface Apply {
+		boolean value();
```
laeubi:
```suggestion
		String value() default "always";
```

HannesWell: Indeed the return type is wrong, copy-paste mistake. Thanks for pointing this out.
The return type also makes sense.
Applied both.

bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/workbench/modeling/IModelProcessorContribution.java
HannesWell: Makes sense. Thanks.
Applied it.

HannesWell:
If there are no more remarks I plan to submit this tomorrow (European) evening.

Ideally we would also have some places within the SDK were this would be used, but I could not find any place were this is immediately possible. If I understood it correctly, in order to use it, we would have to migrate some (or all) of the `org.eclipse.e4.workbench.model.processor` extensions e.g. in `org.eclipse.ui.workbench` or `org.eclipse.e4.ui.workbench.addons.swt` to be OSGi declarative services?
@fipro78 is that correct? Since you contributed the possibility to define processors as OSGi DS, would you be interested in migrating the existing ones or want to give advice?

fipro78:
@HannesWell 
Thanks for the pointer. I have created a PR for this: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2403

The option to use DS for contributing a model processor was intended to be an option for people that don't like the extension points in their code. There was no need to migrate the platform model processors.

Also note that for the existing ones there is no need for the component property types, as the `beforefragment` property defaults to `true` if not set.

And btw the `service` attribute on the `@Component` in your snippet is not necessary, as it is derived from the `implements IModelProcessorContribution`. So actually the information is redundant. Just in case you wonder why I have not used it in my PR.

HannesWell:
Thanks for that.

But this change is ready now.

Issue #2402, opened 2024-10-15
Migrate model processors from extension point to OSGi DS


fipro78:
Since 4.20 it is possible to contribute model processors via OSGi DS (see [New & Noteworthy](https://eclipse.dev/eclipse/news/4.20/platform_isv.php#register-model-processor))

This way it is possible to get rid of a plugin.xml to contribute the model processors and completely rely on OSGi DS. It also supports the OSGi dynamics.

The model processors in the platform should adapt the OSGi DS approach to get rid of the extension points.

laeubi:
Please note that currently there is a bug in the modeleditor that makes it sometimes not possible to migrate:
- https://github.com/eclipse-pde/eclipse.pde/issues/1307

beside that its a great feature I use quite often!

PR #2403, opened 2024-10-15
Impl #2402 - Migrate model processors from extension point to OSGi DS

fipro78:


HannesWell (changes requested):
Thanks for this one the change itself looks sound, but if I start an SDK from my workspace with it the start-up fails:

```
org.eclipse.e4.core.di.InjectionException: Unable to process "BindingService.manager": no actual value was found for the argument "BindingManager".
	at org.eclipse.e4.core.internal.di.InjectorImpl.reportUnresolvedArgument(InjectorImpl.java:462)
	at org.eclipse.e4.core.internal.di.InjectorImpl.resolveRequestorArgs(InjectorImpl.java:453)
	at org.eclipse.e4.core.internal.di.InjectorImpl.internalInject(InjectorImpl.java:128)
	at org.eclipse.e4.core.internal.di.InjectorImpl.internalMake(InjectorImpl.java:386)
	at org.eclipse.e4.core.internal.di.InjectorImpl.make(InjectorImpl.java:312)
	at org.eclipse.e4.core.contexts.ContextInjectionFactory.make(ContextInjectionFactory.java:203)
	at org.eclipse.ui.internal.Workbench$31.runWithException(Workbench.java:2337)
	at org.eclipse.ui.internal.StartupThreading$StartupRunnable.run(StartupThreading.java:35)
	at org.eclipse.swt.widgets.Synchronizer.syncExec(Synchronizer.java:183)
	at org.eclipse.ui.internal.UISynchronizer.syncExec(UISynchronizer.java:133)
	at org.eclipse.swt.widgets.Display.syncExec(Display.java:4883)
	at org.eclipse.ui.internal.StartupThreading.runWithoutExceptions(StartupThreading.java:93)
	at org.eclipse.ui.internal.Workbench.initializeDefaultServices(Workbench.java:2331)
	at org.eclipse.ui.internal.Workbench.init(Workbench.java:1714)
	at org.eclipse.ui.internal.Workbench.runUI(Workbench.java:2825)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:632)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:175)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:662)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:598)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1441)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1414)
```

My first guess is that `ModelAssembler` somehow registers contributions through OSGi services differently. I already tried to process services before extensions, but that didn't help. I can further look into this, but probably not before next week-end.

HannesWell:
And regarding your remark in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2396#issuecomment-2413177986
> And btw the service attribute on the @Component in your snippet is not necessary, as it is derived from the implements IModelProcessorContribution. So actually the information is redundant. Just in case you wonder why I have not used it in my PR.

That's right, but using it explicitly was a suggestion from @laeubi to ensure that services don't get implemented 'accidentally' if for example later another interface is implemented that's not intended as service. And I think that's reasonable.

laeubi:
> That's right, but using it explicitly was a suggestion from @laeubi to ensure that services don't get implemented 'accidentally' if for example later another interface is implemented that's not intended as service. And I think that's reasonable.

Its actually a bit different and why I think actually specify it always is good on the long term:

1. Assume you implement *no interface* then your component gets immediately activated
2. Later you let it implement an interface then it is no longer getting activated unless the interface is actually fetched as a service (+ it is registered as such service)

Other case:

1. Assume you implement *two interface* then your component gets registered as that service
2. Later you let it derive from a base class and remove that interface then it is no longer registered as such service because _only direct interfaces are considered_, or you choose to implement an extension of such interface then it becomes suddenly a different service.

So actually if one relies on the default, it means one must check each time something is refactored there, because of this I simply always add it explicitly so its always clear and less prone to accidental changes. e.g. PDE also gives an error if the component does not implement the declared service (anymore).

fipro78:
For the service attribute discussion, well from the maintenance point of view I understand this. Fine for me. And actually necessary to fix the issue.

About the issue, well we now have a startup order issue. The `BindingToModelProcessor` requires the `ContextToModelProcessor` and the `CommandToModelProcessor` the be executed before. Otherwise the necessary instances in the context are not yet available.

I see the following two errors in the logs:

```
!ENTRY org.eclipse.ui 4 4 2024-10-21 06:45:41.953
!MESSAGE Command manager was null in org.eclipse.ui.internal.BindingToModelProcessor

!ENTRY org.eclipse.ui 4 4 2024-10-21 06:45:41.987
!MESSAGE Context manager was null in org.eclipse.ui.internal.BindingToModelProcessor
```

I have now added a Condition service to describe the activation order dependency. With the extension points this was done by registering the extensions in the correct oder in the plugin.xml.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/BindingToModelProcessor.java
```diff
@@ -40,15 +40,20 @@
...
+@Component(service = IModelProcessorContribution.class)
 public class BindingToModelProcessor implements IModelProcessorContribution {
 
 	private Map<String, MBindingContext> contexts = new HashMap<>();
 	private Map<String, MCommand> commands = new HashMap<>();
 	private Map<String, MBindingTable> tables = new HashMap<>();
 	private Set<MKeyBinding> keys = new HashSet<>();
 
+	@Reference(target = "(osgi.condition.id=commandAndContext)")
+	Condition commandAndContextCondition;
```
laeubi: Why not simply use two references here instead? This looks a bot over complicated.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/CommandAndContextCondition.java
```diff
@@ -0,0 +1,19 @@
+package org.eclipse.ui.internal;
```
akurtakov: Missing license header.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/BindingToModelProcessor.java
fipro78: Well, we could replace the condition reference with the [Satisfying Condition](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-satisfying.condition) Component Annotation later on when DS 1.5 works with PDE DS Annotation Support. And there would be then no never used members in the class that might confuse others.

But yes, probably only syntactic sugar. I will update it for two directly and never used references.

laeubi: Maybe add a comment why these are fetched (but not used) otherwise someone might remove them later on.

fipro78: Already done with the latest change ;)

HannesWell (commented):
Thank you for the update and sorry for the late reply.
The change looks good and worked well in my testing.

But please squash your three commits into one with a commit message covering the resulting change, there is no need to effectively capture the review discussion in the git-history (you remove the version bump and let the bot add it again).
Then this is ready from my side.

HannesWell (approved):
Thank you for the updates and this change in general.
The code looks good, now only the build has to succeed.

HannesWell:
Test failure it's unrelated, submitting.

Thanks again!

Issue #2395, opened 2024-10-13
New Light theme - horizontal row lines are not visible in Tree control on Mac


Phillipus:
- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

If a `Tree` control has multiple columns and lines visible (`Tree#setLinesVisible(true)`) it is helpful to have horizontal row lines visible as can be seen in Eclipse's Problems view. On Windows these lines are displayed as horizontal grey lines and, on Mac, as alternating row colors.

However, on Mac, with the new light theme, these alternating row colors are not visible.

Mac Tree on standard light theme:

<img width="1046" alt="mac before" src="https://github.com/user-attachments/assets/4f571c47-7321-49ca-8065-c9b351e5233e">

Mac tree with new light theme:

<img width="1048" alt="mac after" src="https://github.com/user-attachments/assets/5f4d1eb6-f612-42b1-82f2-baafebc5fd62">

Compare the above with Windows using the new light theme:

![windows](https://github.com/user-attachments/assets/7ff1e56e-9374-4d25-9b13-8c70c851300b)

Whereas Windows has the horizontal row lines visible in the new theme, Mac no longer has the equivalent alternating row lines which can be helpful for accessibility.

Note - this is a follow on from https://github.com/eclipse-platform/eclipse.platform.ui/issues/2171 where that issue concerned the Table control and I mistakenly regarded the Problems view as containing a Table not a Tree.

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

BeckerWdf:
This should be fixed with https://github.com/eclipse-platform/eclipse.platform.ui/pull/2404
Pls. reopen if you still see the issue.

PR #2404, opened 2024-10-15
[MAC] - Light(Preview) theme - Fix for missing alternate row colors for Tree 

mvm-sap:
Tree Viewers had alternate row colors if they had coloumns, in Light Theme. This was missing in Light(Preview) theme. This has been fixed with this change by making swt-lines-visible (Tree Property) to true.
Please refer #2395

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/e4_preview_mac.css
+++ b/bundles/org.eclipse.ui.themes/css/e4_preview_mac.css
@@ -103,7 +103,6 @@ CTabFolder Canvas {
 }
 
 .View Composite,
-.View Composite Tree,
 .View Composite Label,
 .View ToolBar,
 .View Group,
@@ -143,6 +142,11 @@ CTabFolder Canvas {
 
 .View TabbedPropertyList{
 	swt-tabBackground-color: #ffffff;
+}
+
+
+.View Composite Tree[swt-lines-visible=false]{
+	background-color: #f8f8f8;
 }
 
 .View Composite PrependingAsteriskFilteredTree,


bundles/org.eclipse.ui.themes/css/e4_preview_mac.css
```diff
@@ -126,6 +125,11 @@ CTabFolder Canvas {
 	background-color: #f8f8f8;
 }
 
+.View Composite Tree{
+	background-color: #f8f8f8;
```
BeckerWdf: Tested on macOS.
Zebra-Lines are not visible because of this line. If this is removed. Zebar-Lines are visible.

BeckerWdf:
```suggestion
```

BeckerWdf (changes requested):

BeckerWdf (commented):
looks good

BeckerWdf:
with this PR:
Screenshot 2024-10-15 at 16 45 24

before:
Screenshot 2024-10-15 at 16 46 48

mvm-sap:
> with this PR: Screenshot 2024-10-15 at 16 45 24
> 
> before: Screenshot 2024-10-15 at 16 46 48

In after image, Project Explorer and Outline view have white background

BeckerWdf:
> In after image, Project Explorer and Outline view have white background

You are right. So they requirement would be: Show the zebra-lines in Trees *but only* when the tree has more then one column. Correct?

mvm-sap:
> > In after image, Project Explorer and Outline view have white background
> 
> You are right. So they requirement would be: Show the zebra-lines in Trees _but only_ when the tree has more then one column. Correct?

Yes

Phillipus:
> So they requirement would be: Show the zebra-lines in Trees _but only_ when the tree has more then one column. Correct?

Actually zebra lines are only visible when `Tree#setLinesVisible(true)`

bundles/org.eclipse.ui.themes/css/e4_preview_mac.css
```diff
@@ -195,6 +194,10 @@ CTabFolder Canvas {
 	background-color: #ffffff;
 }
 
+#org-eclipse-ui-navigator-ProjectExplorer Tree{
```
BeckerWdf: We should reference concrete views of the Eclipse SDK.

BeckerWdf:
Do it get it right: Mac creates these zebra lines for trees only if they have more then two "columns"? Is that a feature of the native UI toolkit or is this controlled by the SWT implementation?

Is it correct that setting background color does overwrite the "show visible lines" setting? Is that think that is controlled by the SWT implementation?

Do we have a possibility to tell SWT so set the background only if "show visible lines" is false / if the tree as more then two "columns"?

Phillipus:
> Mac creates these zebra lines for trees only if they have more then two "columns"? 

No. Zebra lines are only visible when `Tree#setLinesVisible(true)` with any amount of columns (or none).

mvm-sap:
> > Mac creates these zebra lines for trees only if they have more then two "columns"?
> 
> No. Zebra lines are only visible when `Tree#setLinesVisible(true)` with any amount of columns (or none).

Below is my understanding

 **Light theme** 
 Project explorer, Outline View - Tree Control - has only one column - bg color is white - setLinesVisible is false
 Problems View  - Tree Control -  has more than one column - bg has alternate row colors - setLinesVisible is true

**Light (Preview) theme**
 we just set the bg color of Tree control to grey (#F8F8F8) in CSS file
```
.View Tree {
swt-background-color: #f8f8f8 
}
```

 Project explorer, Outline View - Tree Control - has only one column - bg color is grey - setLinesVisible is false
 Problems View  - Tree Control -  has more than one column - bg color is grey - **setLinesVisible is set to false automatically**

BeckerWdf:
I just pushed a fix for this.
@Phillipus: Can you pls test this on your side?

Phillipus:
> I just pushed a fix for this.
@Phillipus: Can you pls test this on your side?

Tested, and I see zebras. 🦓 🦓 🦓 🦓 

Thanks!

BeckerWdf (approved):

mvm-sap:
> I just pushed a fix for this. @Phillipus: Can you pls test this on your side?

Looks like a good solution. Thanks, Matthias.

PR #2405, opened 2024-10-15
[FindNextAction] synchronize search history with FindReplaceOverlay

Wittmaxi:
synchronize search history with the FindReplaceOverlay and FindReplaceDialog.

This fix is not as extensive as a prior attempt (https://github.com/eclipse-platform/eclipse.platform.ui/pull/2308) and thus does not require as much testing. Since this is a high-priority issue, we want this to be fixed ASAP.

1) search histories and options are now stored at a centralized place - in the options for the FindReplaceAction class.
2) the name of the search history settings section is now unified
3) the search history component of the Find/Replace Overlay now updates it's history on the fly to accomodate.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindNextAction.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindNextAction.java
@@ -340,14 +340,13 @@ public class FindNextAction extends ResourceAction implements IUpdate {
 	 * @return the dialog settings to be used
 	 */
 	private IDialogSettings getDialogSettings() {
-		IDialogSettings settings = PlatformUI.getDialogSettingsProvider(FrameworkUtil.getBundle(FindNextAction.class))
-				.getDialogSettings();
-		fDialogSettings= settings.getSection(FindReplaceDialog.class.getName());
+		IDialogSettings settings = PlatformUI
+				.getDialogSettingsProvider(FrameworkUtil.getBundle(FindReplaceAction.class)).getDialogSettings();
+		fDialogSettings = settings.getSection(FindReplaceAction.class.getClass().getName());
 		if (fDialogSettings == null)
-			fDialogSettings= settings.addNewSection(FindReplaceDialog.class.getName());
+			fDialogSettings = settings.addNewSection(FindReplaceAction.class.getClass().getName());
 		return fDialogSettings;
 	}
-
 	/**
 	 * Initializes itself from the dialog settings with the same state
 	 * as at the previous invocation.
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindNextAction.java
```diff
@@ -340,12 +340,10 @@ private int findAndSelect(int offset, String findString, boolean forwardSearch,
 	 * @return the dialog settings to be used
 	 */
 	private IDialogSettings getDialogSettings() {
-		IDialogSettings settings = PlatformUI.getDialogSettingsProvider(FrameworkUtil.getBundle(FindNextAction.class))
+		IDialogSettings settings = PlatformUI
+				.getDialogSettingsProvider(FrameworkUtil.getBundle(FindReplaceAction.class))
 				.getDialogSettings();
-		fDialogSettings= settings.getSection(FindReplaceDialog.class.getName());
-		if (fDialogSettings == null)
-			fDialogSettings= settings.addNewSection(FindReplaceDialog.class.getName());
```
HeikoKlare: This will break the `FindReplaceDialog`, as it still uses the subsection according to its class name. Shouldn't we use a subsection in the dialog settings of the bundle, e.g., using the class name of the `FindReplaceAction` class, like done with the `FindReplaceDialog` before?

HeikoKlare (changes requested):
Sharing the settings between the dialog and the overlay is good. The current proposal will produce a regression in the dialog, which needs to be resolved.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindNextAction.java
Wittmaxi: I have made it so that FindReplaceDialog, FindReplaceOverlay and FindNextAction have the exact same method for getting the Dialog Settings in https://github.com/eclipse-platform/eclipse.platform.ui/commit/5847b3df0e414cd6b3716e3be916ee634fe902fe
I have contemplated adding this as a public method to the FindReplaceAction, however I chose not to in order not to pollute API

Wittmaxi:
random fails documented in 

https://github.com/eclipse-platform/eclipse.platform.ui/issues/195
https://github.com/eclipse-platform/eclipse.platform.ui/issues/294
https://github.com/eclipse-platform/eclipse.platform.ui/issues/751

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindNextAction.java
HeikoKlare: @Wittmaxi Now you extract different settings sections, as each class (the action, overlay and dialog) retrieve the one according their own class instead of the one for the same class (such as the action):
```java
IDialogSettings dialogSettings = settings.getSection(getClass().getName());
```

HeikoKlare (approved):

HeikoKlare:
Failing test is unrelated and documented: #926

PR #2406, opened 2024-10-15
Increase left margins in F&R overlay

fedejeanne:
Allow for the error icon to appear completely when necessary e.g. when searching for an invalid regular expression.

### Before this PR
The error icon is cut on the left:

image

### After this PR

image


Wittmaxi:
How does this look without the icons?

fedejeanne:
> How does this look without the icons?

Like this:

image

Why do you ask? Did I miss anything?

fedejeanne:
Failing test is unrelated #195

HeikoKlare:
> Like this:
> 
> image

Now we have quite a large whitespace to the left of the input field, just because of the (seldom) use case of having an invalid regex entered. Isn't there any other solution to the problem than this tweak?

fedejeanne:
LGTM

image

PR #2407, opened 2024-10-15
Multi line error message fixed

jannisCode:
## What it does
This PR solves the issue from here: (https://github.com/vi-eclipse/Eclipse-Platform/issues/130#issue-2585021204)

## Before
The error message did not handle the different widths of characters well. Especially the ^ was on the wrong charachter, basically making it useless.
image

## After
Now the ^ is on the right character, no matter what the user types in.
image

### DIFF
--- a/bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
+++ b/bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
@@ -19,7 +19,9 @@ import java.util.regex.PatternSyntaxException;
 
 import org.eclipse.jface.fieldassist.ControlDecoration;
 import org.eclipse.jface.fieldassist.FieldDecorationRegistry;
+import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.widgets.Control;
 
 /**
  * This class contains methods to validate and decorate search fields.
@@ -41,11 +43,10 @@ public class SearchDecoration {
 	 *                         the validation.
 	 */
 	public static boolean validateRegex(String regex, ControlDecoration targetDecoration) {
-		String errorMessage = getValidationError(regex);
+		String errorMessage = getValidationError(regex, targetDecoration.getControl());
 		if (errorMessage.isEmpty()) {
 			targetDecoration.hide();
 			return true;
-
 		}
 
 		Image decorationImage = FieldDecorationRegistry.getDefault()
@@ -62,21 +63,54 @@ public class SearchDecoration {
 	 * @return The appropriate error message if the regex is invalid or an empty
 	 *         string if the regex is valid.
 	 */
-	private static String getValidationError(String regex) {
+	private static String getValidationError(String regex, Control targetControl) {
 		try {
 			Pattern.compile(regex);
 			return ""; //$NON-NLS-1$
 		} catch (PatternSyntaxException e) {
-			String message = e.getLocalizedMessage();
-
-			// Only preserve the first line of the original error message.
-			int i = 0;
-			while (i < message.length() && "\n\r".indexOf(message.charAt(i)) == -1) { //$NON-NLS-1$
-				i++;
-			}
-
-			return message.substring(0, i);
+			return buildValidationErrorString(e, targetControl);
 		}
+	}
+
+	private static String buildValidationErrorString(PatternSyntaxException e, Control targetControl) {
+		String description = e.getDescription();
+		int errorIndex = e.getIndex();
+
+		if (errorIndex == -1) {
+			return description;
+		}
+
+		GC gc = new GC(targetControl);
+		String pattern = e.getPattern();
+
+		StringBuilder validationErrorMessage = new StringBuilder();
+
+		validationErrorMessage.append(description);
+		validationErrorMessage.append(" at index ").append(errorIndex).append(System.lineSeparator()); //$NON-NLS-1$
+		validationErrorMessage.append(pattern).append(System.lineSeparator());
+
+		String stringToIndexString = pattern.substring(0, errorIndex + 1);
+		String hairSpace = "\u200A"; //$NON-NLS-1$
+		int hairSpaceWidth = gc.stringExtent(hairSpace).x;
+
+		int stringToIndex = gc.stringExtent(stringToIndexString).x;
+		String lastCharacter = stringToIndexString.substring(stringToIndexString.length() - 1);
+
+		int widthLastChar = gc.stringExtent(lastCharacter).x;
+		int upWidth = gc.stringExtent("^").x; //$NON-NLS-1$
+
+		double howFar = stringToIndex - widthLastChar / 2 - upWidth / 2;
+		int currentWidth = 0;
+
+		while (currentWidth < howFar) {
+			currentWidth += hairSpaceWidth;
+			validationErrorMessage.append(hairSpace);
+		}
+
+		validationErrorMessage.append("^"); //$NON-NLS-1$
+		gc.dispose();
+
+		return validationErrorMessage.toString();
 	}
 
 }


bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -68,14 +73,39 @@ private static String getValidationError(String regex) {
...
 
-			// Only preserve the first line of the original error message.
 			int i = 0;
 			while (i < message.length() && "\n\r".indexOf(message.charAt(i)) == -1) { //$NON-NLS-1$
 				i++;
 			}
+			int j = i + 2;
+			while (j < message.length() && "\n\r".indexOf(message.charAt(j)) == -1) { //$NON-NLS-1$
+				j++;
+			}
```
fedejeanne: Extract into method and call twice. Give the method a meaningful name like `indexOfLineBreak`

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -68,14 +73,39 @@ private static String getValidationError(String regex) {
...
+
+			String firstLine = message.substring(0, i);
+
+			String indexString = firstLine.replaceAll("\\D+", ""); //$NON-NLS-1$ //$NON-NLS-2$
+			int errorIndex = Integer.parseInt(indexString);
+
+			sBuilder.append(firstLine);
+			sBuilder.append(System.lineSeparator());
+
+			String secondLine = message.substring(i + 2, j); // the +2 because of the \n\r
```
fedejeanne: Linux and Mac use only `\n` so the `+2` won't work. You should work with `String::lastIndexOf` and `System.lineSeparator()` instead.

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -68,14 +73,39 @@ private static String getValidationError(String regex) {
...
+			sBuilder.append(secondLine);
+			sBuilder.append(System.lineSeparator());
+
+			String subsString = secondLine.substring(0, errorIndex);
+			String string = ""; //$NON-NLS-1$
```
fedejeanne: Give these 2 meaningful names.

fedejeanne (commented):
I have some minor comments regarding the coding style.

I also tested it and sometimes the `^` is a bit off.

Regex: `[` --> ✔️ 
image

Regex: `[d` --> ✔️ 
image

Regex: `[df` --> ❌ (a bit to the right)
image

Maybe you can work with [Hair spaces](https://en.wikipedia.org/wiki/Whitespace_character#Hair_spaces_around_dashes) or [Thin spaces](https://en.wikipedia.org/wiki/Thin_space)?

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -14,17 +14,20 @@
 
 package org.eclipse.ui.internal;
 
+
```
jannisCode, line 17: .

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -41,6 +44,8 @@ private SearchDecoration() {
 	 *                         the validation.
 	 */
 	public static boolean validateRegex(String regex, ControlDecoration targetDecoration) {
+		gc = new GC(targetDecoration.getControl());
+
```
jannisCode, line 48:
```suggestion
```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -41,6 +44,8 @@ private SearchDecoration() {
 	 *                         the validation.
 	 */
 	public static boolean validateRegex(String regex, ControlDecoration targetDecoration) {
+		gc = new GC(targetDecoration.getControl());
```
jannisCode: todo: write unit test to highlight reliance on Java-specific behavior (RegEx Error message formats)

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
merks: It's odd to see \n\r when windows uses \r\n...

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -14,17 +14,20 @@
...
 import org.eclipse.jface.fieldassist.ControlDecoration;
 import org.eclipse.jface.fieldassist.FieldDecorationRegistry;
+import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
 
 /**
  * This class contains methods to validate and decorate search fields.
  */
 public class SearchDecoration {
+	private static GC gc;
```
merks: A static gc?

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
merks: A GC is created by not disposed.

Wittmaxi:
I think it is important to mention here that this PR relies very heavily on the RegEx error message format.
Any changes to that format or any errors on the compiler side would *fail silently*

https://docs.oracle.com/javase/8/docs/api/java/util/regex/PatternSyntaxException.html#getIndex--
https://docs.oracle.com/javase/8/docs/api/java/util/regex/PatternSyntaxException.html#getPattern--

Java *DOES* have a method that helps with building your own error message - the catch: the FindReplaceTargetStatus object does not provide them. It should be rather easy to add those into the object and to modify the message generator if required.

Wittmaxi:
Modify in org.eclipse.ui.internal.findandreplace.status.InvalidRegExStatus

```java

	public int getIndex() {
		return regExException.getIndex();
	}

	public String getPattern() {
		return regExException.getPattern();
	}

```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
fedejeanne: It works because it matches either `\n` or `\r`, but I agree: it's odd.

merks: I don't know how "foobar".indexOf.. can match anything except exactly "foobar".  In any case, whether it works or not  not evident because there is no test that it works on any OS...

Never mind, I see what you mean now.   Very strange way to express that.  😕

merks: Perhaps `x.replaceAll("(?s)[\r\n].*", "")` is a more concise way to trim off all but the first line.  One make compile the pattern with Pattern.DOT_ALL to make it more effecient.

jannisCode: Fixed that with @Wittmaxi ´s suggestion. That way we don´t even need the while loop!!

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: Thanks for the suggestion, with @Wittmaxi s help, I don´t need that anymore, so that is fixed

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: Changed that, so that the GC is created within the method and then given as an Argument to getValidationError(). I hope that makes it better !

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: added the gc.dispose()

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -14,11 +14,13 @@
 
 package org.eclipse.ui.internal;
 
+
```
fedejeanne:
```suggestion
```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -41,7 +43,9 @@ private SearchDecoration() {
 	 *                         the validation.
 	 */
 	public static boolean validateRegex(String regex, ControlDecoration targetDecoration) {
-		String errorMessage = getValidationError(regex);
+		GC gc = new GC(targetDecoration.getControl());
```
fedejeanne: Don't create the GC here. Create it in the method were it is actually needed (and dispose it there too).

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
fedejeanne: Don't create it outside the method where it is actually used just to pass it as a parameter, it pollutes the code (see my comment below)

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -62,20 +67,33 @@ public static boolean validateRegex(String regex, ControlDecoration targetDecora
...
 	 */
-	private static String getValidationError(String regex) {
+	private static String getValidationError(String regex, GC gc) {
 		try {
 			Pattern.compile(regex);
 			return ""; //$NON-NLS-1$
 		} catch (PatternSyntaxException e) {
-			String message = e.getLocalizedMessage();
+			String description = e.getDescription();
+			int errorIndex = e.getIndex();
```
fedejeanne: According to the JavaDoc, `getIndex()` may return -1. In that case you should only show the error message (no extra lines)

fedejeanne:
@jannisCode please adapt the commit message to suit the [recommendations](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations)

It still looks odd with `[df`
image

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: Done

jannisCode:
I now changed the while loop, so that it works better with the example [df. Unfortunately it doesn´t handle very wide characters like m or o as well as before then. 
image
image

fedejeanne (commented):
As discussed in today's call, here are some tips:
- Don't pass the `ControlDecoration` as a parameter, only the `Control`,
- Return earlier in case `e.getIndex() == -1` and avoid creating a `StringBuffer`,
- Extract methods from the whole `catch` block to improve readability,
- Use **hair spaces** (they are thinner than _thin spaces_)
- Try to compute the amount of _hair spaces_ that you need to add before the `^`. In the end, the ideal result is when the tip of the "`^`" is placed right in the middle of the offending character.

jannisCode:
Thank you for your comment! I used your suggestions and made adjustments accordingly.

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -62,21 +62,55 @@ public static boolean validateRegex(String regex, ControlDecoration targetDecora
 	 * @return The appropriate error message if the regex is invalid or an empty
 	 *         string if the regex is valid.
 	 */
-	private static String getValidationError(String regex) {
+	private static String getValidationError(String regex, org.eclipse.swt.widgets.Control targetControl) {
+
 		try {
 			Pattern.compile(regex);
 			return ""; //$NON-NLS-1$
+
```
fedejeanne:
```suggestion
```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -19,6 +19,7 @@
 
 import org.eclipse.jface.fieldassist.ControlDecoration;
 import org.eclipse.jface.fieldassist.FieldDecorationRegistry;
+import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
```
fedejeanne, line 23: Import `org.eclipse.swt.widgets.Control` to use it later

```suggestion
import org.eclipse.swt.graphics.GC;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.widgets.Control;
```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -62,21 +62,55 @@ public static boolean validateRegex(String regex, ControlDecoration targetDecora
...
 			Pattern.compile(regex);
 			return ""; //$NON-NLS-1$
+
 		} catch (PatternSyntaxException e) {
-			String message = e.getLocalizedMessage();
+			return buildValidationErrorString(e, targetControl);
+		}
+	}
+
+	private static String buildValidationErrorString(PatternSyntaxException e, org.eclipse.swt.widgets.Control targetControl) {
```
fedejeanne:
```suggestion
	private static String buildValidationErrorString(PatternSyntaxException e, Control targetControl) {
```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -62,21 +62,55 @@ public static boolean validateRegex(String regex, ControlDecoration targetDecora
...
+		GC gc = new GC(targetControl);
+
+		String description = e.getDescription();
+		int errorIndex = e.getIndex();
+		String pattern = e.getPattern();
+
+		if (errorIndex == -1) {
+			return description;
+		}
```
fedejeanne: The GC and the pattern are needed later, no need for them if you do an early return.

```suggestion
		String description = e.getDescription();
		int errorIndex = e.getIndex();

		if (errorIndex == -1) {
			return description;
		}
		
		GC gc = new GC(targetControl);
		String pattern = e.getPattern();
```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -62,21 +62,55 @@ public static boolean validateRegex(String regex, ControlDecoration targetDecora
...
+		StringBuilder sBuilder = new StringBuilder();
```
fedejeanne: Rename 
```suggestion
		StringBuilder validationErrorMessage = new StringBuilder();
```

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
```diff
@@ -62,21 +62,55 @@ public static boolean validateRegex(String regex, ControlDecoration targetDecora
...
+
+		sBuilder.append(description);
+		sBuilder.append(" at index ").append(errorIndex).append(System.lineSeparator()); //$NON-NLS-1$
+		sBuilder.append(pattern).append(System.lineSeparator());
+
+		String stringToIndexString = pattern.substring(0, errorIndex + 1);
+		String buildString = ""; //$NON-NLS-1$
```
fedejeanne: Get rid of this one, you don't need it. You need to `append` to the `StringBuilder` directly

fedejeanne (commented):
The results are much better now. There are only a few comments about the coding but after they are addressed, this PR will be ready to go 👍

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

bundles/org.eclipse.ui/src/org/eclipse/ui/internal/SearchDecoration.java
jannisCode: done

fedejeanne:
Failed check: see #2432

fedejeanne:
@jannisCode which changes did you introduce when you force-pushed 3c3d8cc57b7e68d96a665fd12ffc9728b8b494fa? I can't see any difference in the code when I compare it to my previous commit (9362717d0ce4ce9d67459e1123bd18adaef99293), only that you removed me as co-author.

jannisCode:
@fedejeanne I am pretty sure that it was just an update with rebase. I definitely didn´t mean to remove you as co-author

fedejeanne:
> @fedejeanne I am pretty sure that it was just an update with rebase. I definitely didn´t mean to remove you as co-author

👍 
I removed some unnecessary empty lines and pushed again. I also wanted to test if the failing tests were unrelated to this PR (I think they are).

After the checks are done, I will merge this. Please do not push any more changes.

fedejeanne (approved):

fedejeanne:
@jannisCode care to add a N&N entry? :-)

PR #2411, opened 2024-10-16
Changing the name from "Eclipse UI" to "eclipseui" to avoid space in path names

elsazac:
Refs: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1460/

Changing the path name from "Eclipse UI" to "eclipseui" and "Eclipse UI Editor Support" to  "eclipseuieditorsupport" ,
to avoid space in path names and to make it UNIX friendly.

Files changed belongs to :  `org.eclipse.ui.workbench`


elsazac:
This is the first bundle created as suggested in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1460#issuecomment-2371559558, and even though it's a draft PR, it's open for review. However, please don't land it until all the bundles are covered.

akurtakov:
@elsazac For each such change you have to make sure that the full agregator build still works - e.g. this will break javadoc build and https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b61229494d094a01678b663371379fce580df2d7/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml#L594 have to be changed. Please prep a corresponding PR.
Also I believe it's best to handle bundle by bundle fully and land the changes in so build issues like the one pointed can be handled easier instead of landing multiple bundles at once and thus making it harder to fix things after that.

elsazac:
> @elsazac For each such change you have to make sure that the full agregator build still works - e.g. this will break javadoc build and https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b61229494d094a01678b663371379fce580df2d7/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml#L594 have to be changed. Please prep a corresponding PR. 

@akurtakov - thanks, I thought about that, and hence I kept this PR in draft. I will bring up the releng-aggregator PR before undrafting this PR.

elsazac:
> Also I believe it's best to handle bundle by bundle fully and land the changes in so build issues like the one pointed can be handled easier instead of landing multiple bundles at once and thus making it harder to fix things after that.

my original plan was to make multiple PRs for as many bundles for `Eclipse UI` subfolders, and un-draft all at once when everything is ready. But reading your comment and thinking about it further - I think a better approach would be to :
 - prepare the releng-aggregator PR for this change (ie, add a new sourcepath ( ie, (`eclipseui`) ) for the new bundle  )
 - Also at the same time retain the original `Eclipse UI` version in the `sourcepath`
 - retain the original `classpathentry` and `source` in `bundles/org.eclipse.ui.workbench/.classpath` and `bundles/org.eclipse.ui.workbench/build.properties` while adding a new one (ie, `eclipseui`) for the new bundle
 - land this PR, and fix any potential issues thereon
 - repeat this for all the bundles
 - finally remove the original entries (ie `Eclipse UI`) from `pom.xml`, `build.properties` and `.classpath`, that points to the old folder which is now unused.

What do you think?

akurtakov:
This sounds overly complicated to me. Smth like:
* grep in aggregator to find where "Eclipse UI" is used
* do all the changes locally
* verify it builds locally
* land all PRs at once

HannesWell:
In order to very this change in an aggregator build in advance you could also create a temporary/draft PR against https://github.com/eclipse-platform/eclipse.platform.releng.aggregator, where you modify the .gitmodules file to include this PR's branch.
I.e. you have to change
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b61229494d094a01678b663371379fce580df2d7/.gitmodules#L26-L28
to 
```
[submodule "eclipse.platform.ui"]
  path = eclipse.platform.ui
  url = https://github.com/elsazac/eclipse.platform.ui.git
  branch = bundle2
```

elsazac:
> This sounds overly complicated to me. Smth like:
> 
> * grep in aggregator to find where "Eclipse UI" is used
> * do all the changes locally
> * verify it builds locally
> * land all PRs at once


Sure, I will proceed accordingly.

elsazac:
> In order to very this change in an aggregator build in advance you could also create a temporary/draft PR against https://github.com/eclipse-platform/eclipse.platform.releng.aggregator, where you modify the .gitmodules file to include this PR's branch.

@HannesWell Thank you for the suggestion. I’ll incorporate this.

laeubi:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/?tab=readme-ov-file#how-to-build-the-eclipse-sdk shows how one can build aggregator locally, it takes a while but might be more efficient than waiting for the Jenkins, you can for example simply remove the checked out submodule and replace it with a link to your forked repository.

sravanlakkimsetti:
Possibly you may use https://ci.eclipse.org/platform/job/eclipse-aggregator-master/ modify this job according to your needs

elsazac:
@akurtakov This is ready for review

This has been tested locally and the following changes are required.

Changes in a nutshell :

 - 11 files moved to new folder `eclipseui` 
 - Made entry in `.classpath` and `build.properties` for the new path.
 - Retain the old entries (for the sake of other files, will be removed after this exercise)
 - Adjusted releng aggregator to accommodate this change.

bundles/org.eclipse.ui.workbench/.classpath
```diff
@@ -2,7 +2,8 @@
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="Eclipse UI"/>
+	<classpathentry kind="src" path="eclipseui"/>
+        <classpathentry kind="src" path="Eclipse UI"/>
```
akurtakov: Why do you keep "Eclipse UI" if you renamed it to eclipseui?

elsazac: There are more files in `Eclipse UI` path yet to be moved. if I remove it now, the build will break due to missing files in the available classpath. the entry `Eclipse UI` is safe to be removed only after all the bundles are modified and landed.

akurtakov:
So in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1460#issuecomment-2376318086 you said that you will do one PR per bundle, what happened ?

bundles/org.eclipse.ui.workbench/.classpath
akurtakov: What is a "bundle" for you?

elsazac: This bundle alone contains over 1,000 files. The earlier feedback was that it is too much to review. It seems we may have a disconnect in terms of  bundle versus folder as the unit for movement. Would you be open to moving the entire folder into a single PR?

akurtakov: Your previous PR did the change for multiple bundles at once (other than o.e.ui.workbench too). To make it clear - please do all the changes in org.eclipse.ui.workbench bundle at once (even if they are thousand). 
Do not involve other bundles (like search in previous PR) that can and will complicate changes in other releng scripts.

elsazac: The `search` bundle was already created and merged a while ago,  https://github.com/eclipse-platform/eclipse.platform.ui/pull/1536. This PR ie, https://github.com/eclipse-platform/eclipse.platform.ui/pull/1460 was specifically done for `org.eclipse.ui.workbench` which contains two changes (`Eclipse UI` and `Eclipse UI Editor Support`) and this contained only the changes from `org.eclipse.ui.workbench` . This was what I pointed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1460#issuecomment-1903843312

elsazac: I skimmed through all the comments in the previous PR. I guess the disconnect mainly stemmed from the last few comments in #1460 which suggested close and rework, despite having a per-bundle change.

the only affected bundle remaining right now is `org.eclipse.ui.workbench` and that requires renaming of two folders with thousands of files inside. I am going to make that change in one single PR (this PR) and ofc, accommodate the same in `releng.agreegator`, but pls let me know if you are in sync with me.

akurtakov: Yes, that's it. Sorry for the long and mismatched communication.

elsazac:
Made changes in `org.eclipse.ui.workbench` bundle. Will make a corresponding change in releng aggregator as well.

elsazac:
@akurtakov This is ready now.

Could you review and let me know of any changes?

akurtakov:
Looks good to me. Please merge (together with aggregator PR) and start I-build immediately after that when you can monitor the build for breakage.

elsazac:
Merging this now and will initiate an I-Build to check for any potential failures.

PR #2399, opened 2024-10-14
[StickyScrolling] Move complex line adaption to central handler

Christopher-Hermann:
Move the adaption of the sticky lines to the growing sticky lines control into the central handler class. This change simplifies the sticky lines provider, making implementations for specific languages more straightforward and maintainable.

Preparation for #2398

PR #2412, opened 2024-10-17
[StickyScrolling] Move text and style calculation to StickyLine

Christopher-Hermann:
Move the text and style calculation to the StickyLine itself in order to enable the sticky line provider to overwrite the default behavior. This is needed to apply custom texts or custom styles for the sticky lines.

Preparation for #2398

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
@@ -14,8 +14,6 @@
 package org.eclipse.ui.internal.texteditor.stickyscroll;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.StringJoiner;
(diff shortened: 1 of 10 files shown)


Christopher-Hermann:
Depends on #2399

Christopher-Hermann:
@tobiasmelcher Maybe you can have a look. I implemented your idea from the POC via the new IStickyLine interface

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
```diff
@@ -244,14 +242,19 @@ private void styleStickyLines() {
...
-			stickyLinesStyleRanges.addAll(getStickyLineStyleRanges(stickyLine, stickyLineTextOffset));
-			stickyLineTextOffset+= stickyLine.text().length() + System.lineSeparator().length();
+		int stickyLineOffset= 0;
+		for (IStickyLine stickyLine : stickyLines) {
+			StyleRange[] styleRanges= stickyLine.getStyleRanges();
+
+			if (styleRanges != null) {
+				for (StyleRange styleRange : styleRanges) {
+					styleRange.start+= stickyLineOffset;
+					stickyLineText.setStyleRange(styleRange);
```
tobiasmelcher: Performance question: calling setStyleRanges() and passing all styleRanges in one call might be faster than calling setStyleRange separately for each range? I have not made any measurements, it is just an idea.

tobiasmelcher (approved):
thanks a lot Christopher for taking care of this improvement.

bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControl.java
Christopher-Hermann: Good point, done. Thanks for the review.

iloveeclipse:
This PR causes regression => https://github.com/eclipse-platform/eclipse.platform.ui/issues/2496

PR #2416, opened 2024-10-17
Create static variant of ImageDescriptor.imageDescriptorFromURI()

ptziegler:
This method is intended to be used as a wrapper for createFromURL() to avoid having to deal with the checked MalformedURLException. However, this method is effectively unusable, as it requires the user to already have an instance of the ImageDescriptor they want to created.

Instead, create a new, static createFromURI() method, mark the old method as deprecated and internally delegate to the new method.

The old method was created as a response to Bug 559656 [1], in order to match the signature of IResourceUtilities. But because the latter is accessed via an OSGi service, it doesn't need to be static. Which is something that doesn't really work for image descriptors.

[1] https://bugs.eclipse.org/bugs/show_bug.cgi?id=559656

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -205,6 +205,32 @@ public abstract class ImageDescriptor extends DeviceResourceDescriptor<Image> {
 	}
 
 	/**
+	 * Convenient method to create an ImageDescriptor from an URI.
+	 *
+	 * Delegates to {@link ImageDescriptor#createFromURL(URL)}. <em>Important</em>
+	 * This method should only be used when it's guaranteed that the given
+	 * {@link URI} is also a valid {@link URL}, in order to avoid the
+	 * {@link MalformedURLException} thrown by {@link URI#toURL()}.
+	 *
+	 * If the URI is {@code null} or not a valid {@link URL}, then an image from
+	 * {@link #getMissingImageDescriptor()} will be returned.
+	 *
+	 * @param uriIconPath The URI of the image file.
+	 * @return a new image descriptor
+	 *
+	 * @since 3.36
+	 */
+	public static ImageDescriptor createFromURI(URI uriIconPath) {
+		try {
+			return ImageDescriptor.createFromURL(uriIconPath != null ? uriIconPath.toURL() : null);
+		} catch (MalformedURLException e) {
+			// return the missing image placeholder to indicate
+			// the incorrect call without interfering with the user flow
+			return getMissingImageDescriptor();
+		}
+	}
+
+	/**
 	 * Convenient method to create an ImageDescriptor from an URI
 	 *
 	 * Delegates to ImageDescriptor createFromURL
@@ -213,15 +239,11 @@ public abstract class ImageDescriptor extends DeviceResourceDescriptor<Image> {
 	 * @return a new image descriptor
 	 *
 	 * @since 3.19
+	 * @deprecated Use {@link #createFromURI(URI)} instead.
 	 */
+	@Deprecated(since = "3.36", forRemoval = true)
 	public ImageDescriptor imageDescriptorFromURI(URI uriIconPath) {
-		try {
-			return ImageDescriptor.createFromURL(new URL(uriIconPath.toString()));
-		} catch (MalformedURLException | NullPointerException e) {
-			// return the missing image placeholder to indicate
-			// the incorrect call without interfering with the user flow
-			return getMissingImageDescriptor();
-		}
+		return createFromURI(uriIconPath);
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


ptziegler:
@vogella You contributed the original method. wdyt?

vogella:
LGTM, thanks @ptziegler

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -224,6 +224,22 @@ public ImageDescriptor imageDescriptorFromURI(URI uriIconPath) {
...
+	 *
+	 * Delegates to ImageDescriptor createFromURL
+	 *
+	 * @param uriIconPath The URI of the image file.
+	 * @return a new image descriptor
+	 *
+	 * @since 3.19
+	 * @deprecated Use {@link #createFromURI(URI)} instead.
+	 */
+	@Deprecated(since = "3.36")
```
laeubi: Mark for removal maybe?

ptziegler: Done. No point in keeping the old method when there is a drop-in replacement.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -212,9 +212,9 @@ public static ImageDescriptor createFromURLSupplier(boolean useMissingImage, Sup
...
 	 * @return a new image descriptor
 	 *
-	 * @since 3.19
+	 * @since 3.36
 	 */
-	public ImageDescriptor imageDescriptorFromURI(URI uriIconPath) {
+	public static ImageDescriptor createFromURI(URI uriIconPath) {
 		try {
 			return ImageDescriptor.createFromURL(new URL(uriIconPath.toString()));
 		} catch (MalformedURLException | NullPointerException e) {
```
laeubi: As we are currently changing the method, catching NPE is really really **BAD**!
We should check for `null` like in `createFromURL`
```suggestion
		} catch (MalformedURLException e) {
```

ptziegler: In e.g. createFromURL(), we have an explicit null check where getMissingImageDescriptor() is returned. I think it makes more sense to do the same here, rather than forwarding the exception.

ptziegler:
I've also noticed that `new URL(String)` will be deprecated in Java 20, so I switched to the equivalent `URI.toURL()`.

merks:
Be **very careful**.   It looks so innocent and obvious, bu there are cases where URI creation will thrown an exception where URL creation does not.  E.g., if there is a space in the string

vogella:
@laeubi do you still have open concerns here?

laeubi (approved):
Looks fine for me, using this method somewhere in the code (even in a testcase) might help to demonstrate its usefulness.

Maybe we should mention in the javadoc that this only works for URIs that can be converted to URLs?

ptziegler:
> Maybe we should mention in the javadoc that this only works for URIs that can be converted to URLs?

Adding a disclaimer sounds like a good idea.

HannesWell:
> Be **very careful**. It looks so innocent and obvious, bu there are cases where URI creation will thrown an exception where URL creation does not. E.g., if there is a space in the string

That's right.
But here an URI is converted into a URL and as far as I know URIs are strict regarding their content and therefore the conversion URI -> URL should be (relatively) save? But maybe I'm wrong. It's definitively not true for the opposite: URL -> URL can have many problems since URLs are not strictly validating their content.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -205,25 +205,50 @@ public static ImageDescriptor createFromURLSupplier(boolean useMissingImage, Sup
...
 	 */
-	public ImageDescriptor imageDescriptorFromURI(URI uriIconPath) {
+	public static ImageDescriptor createFromURI(URI uriIconPath) {
+		if (uriIconPath == null) {
+			return getMissingImageDescriptor();
+		}
 		try {
-			return ImageDescriptor.createFromURL(new URL(uriIconPath.toString()));
-		} catch (MalformedURLException | NullPointerException e) {
+			return ImageDescriptor.createFromURL(uriIconPath.toURL());
```
HannesWell: We could also delegate the handling of the null case to `createFromURL()` and avoid coding the handling here again. But it's just a suggestion, I don't have a strong preference on it.
```suggestion
			return ImageDescriptor.createFromURL(uriIconPath != null ? uriIconPath.toURL() : null);
```

HannesWell (approved):
Lets move this forward. The change looks good to me and unless somebody objects I plan to submit this within the next days.

Thank you @ptziegler.

> > Be **very careful**. It looks so innocent and obvious, bu there are cases where URI creation will thrown an exception where URL creation does not. E.g., if there is a space in the string
> 
> That's right. But here an URI is converted into a URL and as far as I know URIs are strict regarding their content and therefore the conversion URI -> URL should be (relatively) save? But maybe I'm wrong. It's definitively not true for the opposite: URL -> URL can have many problems since URLs are not strictly validating their content.

Since there was no opposition I assume that assessment is right, but with the current state of the doc even a more dangerous behavior would be fine.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
ptziegler: I think it's better to move the null-handling to `createFromURL()` to really make sure that both methods behave the same on a `null` argument.

HannesWell: Agree.

HannesWell (approved):
If there are no further remarks, I plan to submit this Monday (European) evening.

HannesWell:
Test failures are unrelated, submitting.

Thanks again.

PR #2419, opened 2024-10-18
Simplify ResourceInitialSelectionTest

akurtakov:
Make it plain JUnit test.
Simplify DisplayHelper usage to remove one anonymous class file and have lambda instead.


akurtakov:
Hopefully contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/294 by reducing unneeded operations thus making results more predictable.

PR #2434, opened 2024-10-21
Set Edge data directory to workspace on Windows

amartya4256:
This contributes to adding the workspace address to the display data on the initilization of the workbench. This contribution allows further SWT components to utilize this information.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1013

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -37,6 +37,7 @@ import java.io.StringWriter;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -74,6 +75,7 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.ListenerList;
 import org.eclipse.core.runtime.MultiStatus;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Platform.OS;
 import org.eclipse.core.runtime.SafeRunner;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
@@ -140,6 +142,7 @@ import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.window.IShellProvider;
 import org.eclipse.jface.window.Window;
 import org.eclipse.osgi.internal.location.LocationHelper;
+import org.eclipse.osgi.service.datalocation.Location;
 import org.eclipse.osgi.service.runnable.StartupMonitor;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
@@ -291,6 +294,8 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 	public static final String PROP_EXIT_CODE = "eclipse.exitcode"; //$NON-NLS-1$
 	private static final String CMD_DATA = "-data"; //$NON-NLS-1$
+
+	private static final String EDGE_USER_DATA_FOLDER = "org.eclipse.swt.internal.win32.Edge.userDataFolder"; //$NON-NLS-1$
 
 	private static final class StartupProgressBundleListener implements ServiceListener {
 
@@ -451,6 +456,10 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 	private Workbench(Display display, final WorkbenchAdvisor advisor, MApplication app, IEclipseContext appContext) {
 		this.advisor = Objects.requireNonNull(advisor);
 		this.display = Objects.requireNonNull(display);
+		if (OS.isWindows()) {
+			setEdgeDataDirectory(this.display);
+		}
+
 		application = app;
 		e4Context = appContext;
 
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
I've updated the data key with the proper value added to SWT via: 
- eclipse-platform/eclipse.platform.swt#1548

I've restricted the additional data to Windows and added proper handling for workspace-free applications (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1548#issuecomment-2426515530). I've also simplified the retrieval of the workspace directory path string using more "proper" conversion instead of toString-conversions.

@fedejeanne can you have a look please?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -510,6 +519,19 @@ public void eventLoopException(Throwable exception) {
 				new WorkbenchLocationService(IServiceScopes.WORKBENCH_SCOPE, this, null, null, null, null, 0));
 	}
 
+	private static void setEdgeDataDirectory(Display display) {
+		Location workspaceLocation = Platform.getInstanceLocation();
+		if (workspaceLocation != null) {
+			URI workspaceLocationURI;
```
fedejeanne: Nit: you can declare it and use it inside the `try`.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -510,6 +519,19 @@ public void eventLoopException(Throwable exception) {
 				new WorkbenchLocationService(IServiceScopes.WORKBENCH_SCOPE, this, null, null, null, null, 0));
 	}
 
+	private static void setEdgeDataDirectory(Display display) {
+		Location workspaceLocation = Platform.getInstanceLocation();
+		if (workspaceLocation != null) {
```
fedejeanne: Nit: early return instead. It reduces nesting/complexity.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -510,6 +519,19 @@ public void eventLoopException(Throwable exception) {
...
+			URI workspaceLocationURI;
+			try {
+				workspaceLocationURI = workspaceLocation.getURL().toURI();
+				display.setData(EDGE_USER_DATA_FOLDER, Paths.get(workspaceLocationURI).toString());
```
fedejeanne: I assume that `Platform.getInstanceLocation()` never returns a _relative_ Location (if that is even a thing). If that's the case then this code is fine, if that is **not** the case and therefore `workspaceLocationURI` may be _relative_ then adding a call to `Path::toAbsolutePath` would be necessary.

fedejeanne (commented):
@HeikoKlare the code looks fine. I only have 2 nit-picks and one question/suggestion regarding _relative_ locations (if that's even a thing).

If the question is irrelevant then please consider this PR **approved** on my account :-)

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
HeikoKlare: The location will always be absolute. When initializing the workspace, the path will always be made absolute. When initially passing some relative location, it will be resolve relative to the install directory upon workspace initialization.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
HeikoKlare: Indeed, fixed with latest commit.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
HeikoKlare: Good point. Resolved with latest commit.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
fedejeanne: Thanks for checking it!

fedejeanne (approved):

HeikoKlare:
Failing test is unrelated and documented: #195

sratz:
With this change we now have a directory named `EBWebView` very prominently lying around in the user's workspace next to all the projects:

```
eclipse_workspace\
    .metadata\
    asdf\
    EBWebView\
    foo_bar\
    my_demo_project\
    ...
```

Should we maybe put this into a separate `.` directory?

```
eclipse_workspace\
    .metadata\
    .swt_edge\
        EBWebView\
    asdf\
    foo_bar\
    my_demo_project\
    ...
```

Or move it into the plugin-specific instance-scoped directory?

```
eclipse_workspace\
    .metadata\
        .plugins\
            org.eclipse.swt\
                EBWebView\
    asdf\
    foo_bar\
    my_demo_project\
    ...
```

HeikoKlare:
> Or move it into the plugin-specific instance-scoped directory?

That seems to be a proper place for this folder. Makes sense to move it there instead of having it placed directly in the workspace.

Proposal: #2469

Issue #2432, opened 2024-10-21
OutOfMemoryError reported by jenkins validation build after org.eclipse.ui.tests.UiTestSuite


iloveeclipse:
See https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2431/1/consoleFull that didn't changed code, just updated bundle versions, so the problem must be coming from either environment or code changes before.

The question is: what is causing that, and which heap space is actually used on jenkins?

```
Exception in thread "Active Thread: Equinox Container: aa943991-8bd5-44f7-a55a-1a7bc7647820" java.lang.OutOfMemoryError: Java heap space
Exception in thread "WorkbenchTestable" org.eclipse.swt.SWTException: Failed to execute runnable (java.lang.OutOfMemoryError: Java heap space)
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.widgets.Synchronizer.syncExec(Synchronizer.java:209)
	at org.eclipse.ui.internal.UISynchronizer.syncExec(UISynchronizer.java:133)
	at org.eclipse.swt.widgets.Display.syncExec(Unknown Source)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Testable.runTest(E4Testable.java:118)
	at org.eclipse.tycho.surefire.osgibooter.AbstractUITestApplication.runTests(AbstractUITestApplication.java:38)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Testable.lambda$0(E4Testable.java:79)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.OutOfMemoryError: Java heap space

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Worker-11"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Worker-22"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Worker-23"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Worker-24"
```

iloveeclipse:
Also in https://ci.eclipse.org/platform/job/eclipse.platform.ui/view/change-requests/job/PR-2224/11/consoleFull

```
[...]

PR #2438, opened 2024-10-21
DO NOT MERGE! Dummy change to check OOMs in jenkins

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2432 and https://github.com/eclipse-platform/eclipse.platform/issues/1592

This change shouldn't affect any test at all, it is not a code change.


iloveeclipse:
So this is official, API tooling doesn't work, at least for `org.eclipse.ui.workbench` bundle 

image

akurtakov:
Should this one stay open?

iloveeclipse:
> Should this one stay open?

Surely not :-)

PR #2442, opened 2024-10-25
Remove long time not needed ICU4J references in tests

akurtakov:
These are not referenced at all thus simply deleted.


akurtakov:
MacOS failures are unrelated to the change.

PR #2446, opened 2024-10-26
CloseAllHandlerTest cleanup

akurtakov:
Use static imports and pattern matching to ease reading.

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/e4/CloseAllHandlerTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/e4/CloseAllHandlerTest.java
@@ -12,6 +12,12 @@
 
 package org.eclipse.ui.tests.e4;
 
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/e4/CloseAllHandlerTest.java
```diff
@@ -119,46 +124,46 @@ public void testCloseMixedEditorTypes() {
...
 		canExecute = handlerService.canExecute(parameterizedCommand);
-		Assert.assertFalse(canExecute);
+		assertFalse(canExecute);
 
 		// scenario 2: open a compatibility layer editor
 		IFileEditorInput input = new DummyFileEditorInput();
 		Object activeWindow = applicationContext.getActive(ISources.ACTIVE_WORKBENCH_WINDOW_NAME);
-		Assert.assertTrue("Active workbench window not found.", activeWindow instanceof IWorkbenchWindow);
+		assertTrue("Active workbench window not found.", activeWindow instanceof IWorkbenchWindow);
 		IWorkbenchWindow window = (IWorkbenchWindow) activeWindow;
```
HannesWell, line 145: Probably the following would be possible, but not everyone finds it more readable.
```suggestion
		assertTrue("Active workbench window not found.", activeWindow instanceof IWorkbenchWindow window);
```

akurtakov: Unfortunately this is not possible as that way window variable is local to the assertTrue call and thus not resolved outside of it.

PR #2448, opened 2024-10-26
Microsoft Defender status retrieval: Add 'not running' status

### LINKED ISSUE #2447: Failed to retrieve Microsoft Defender status: Process terminated with unexpected result: Not running

## Steps to reproduce

Disable Windows defender completely, e.g. using https://www.sordum.org/9480/defender-control-v2-1/

`powershell -Command (Get-MpComputerStatus).AMRunningMode` then returns an additional status not documented by Microsoft: `Not running`.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/54849021b21ba173e6008f44d294423017ea456a/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java#L325-L328

should be adjusted to include that.

I wonder: Maybe this method should be more graceful to unknown statuses, e.g. log them and return `false` instead of throwing an Exception.

## Here is some relevant log output

From `<workspace>/.metadata/.log`
```
org.eclipse.core.runtime.CoreException: Failed to retrieve Microsoft Defender status.
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.isWindowsDefenderActive(WindowsDefenderConfigurator.java:331)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.runExclusionCheck(WindowsDefenderConfigurator.java:159)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.lambda$0(WindowsDefenderConfigurator.java:106)
Caused by: java.io.IOException: Process terminated with unexpected result:
Not running
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.isWindowsDefenderActive(WindowsDefenderConfigurator.java:328)
	... 4 more
```

* OS & version: Windows 11 Build 27729

sratz:
Fixes #2447

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -323,7 +323,8 @@ public class WindowsDefenderConfigurator implements EventHandler {
 			return switch (onlyLine.toLowerCase(Locale.ENGLISH).strip()) {
 			// Known values as listed in
 			// https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-windows#use-powershell-to-check-the-status-of-microsoft-defender-antivirus
-			case "sxs passive mode", "passive mode", "" -> false; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+			// "not running" status appears to be undocumented (https://github.com/eclipse-platform/eclipse.platform.ui/issues/2447)
+			case "sxs passive mode", "passive mode", "not running", "" -> false; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
 			case "normal", "edr block mode" -> true; //$NON-NLS-1$//$NON-NLS-2$
 			default -> throw new IOException("Process terminated with unexpected result:\n" + String.join("\n", lines)); //$NON-NLS-1$//$NON-NLS-2$
 			};


bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -323,7 +323,8 @@ private static boolean isWindowsDefenderActive(IProgressMonitor monitor) throws
 			return switch (onlyLine.toLowerCase(Locale.ENGLISH).strip()) {
 			// Known values as listed in
 			// https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-windows#use-powershell-to-check-the-status-of-microsoft-defender-antivirus
-			case "sxs passive mode", "passive mode", "" -> false; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+			// "not running" status appears to be undocumented (#2447)
```
HannesWell: Using the full URL is future proof in case repos are merged/split/renamed again one day.
```suggestion
			// "not running" status appears to be undocumented (https://github.com/eclipse-platform/eclipse.platform.ui/issues/2447)
```

HannesWell (approved):
Looks good to me, thanks.

> I wonder: Maybe this method should be more graceful to unknown statuses, e.g. log them and return false instead of throwing an Exception.

Regarding that question from your issue: I considered that as well, but deliberately decided against it, because an unknown value could also mean the defender is active, e.g. if the defender is changed in a documented way in the future.
Then we would at least notice it even if none monitors the defender doc.
So while this is a bit more cumbersome in the beginning I think it's the saver way on the long run.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -323,7 +323,8 @@ private static boolean isWindowsDefenderActive(IProgressMonitor monitor) throws
...
+			case "sxs passive mode", "passive mode", "not running", "" -> false; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
```
HannesWell: Another NON-NLS tag is missing
```suggestion
			case "sxs passive mode", "passive mode", "not running", "" -> false; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
```

HannesWell (changes requested):

HannesWell (approved):
Looks good, thanks for the update. 

Ideally squash locally and force push as one commit, for context see the discussion at https://github.com/eclipse-pde/eclipse.pde/pull/1455#issuecomment-2437524603

PR #2455, opened 2024-10-28
Create Color Definition for Reuse

BeckerWdf:
Instead of writing the same RBG color codes over and over again in various places in the CSS files only do it once.
Define a new color and re-use that color definition at all other locations. This also adds a semantic name to the color code.

Unfortunately color definitions cannot refer to other color definitions. So at some places we still have to provides the color code. These places are marked with a comment.

~~The new defined color is now also visible in the "Colors and Fonts" preference page. I am unsure if this is something we want. If users change this this does not work 100% because of the limitation above.~~

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
+++ b/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
@@ -19,73 +19,73 @@
 @import url("platform:/plugin/org.eclipse.ui.themes/css/e4_basestyle.css");
 
 ColorDefinition#org-eclipse-ui-workbench-ACTIVE_UNSELECTED_TABS_COLOR_START {
-	color: #F8F8F8;
+	color: #f8f8f8; /* same as '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND' cannot use it directly*/
 }
 
 ColorDefinition#org-eclipse-ui-workbench-ACTIVE_UNSELECTED_TABS_COLOR_END {
-	color: #F8F8F8;
+	color: #F8F8F8; /* same as '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND' cannot use it directly*/
 }
 
 ColorDefinition#org-eclipse-ui-workbench-ACTIVE_TAB_OUTER_KEYLINE_COLOR {
-	color: #E5E5E5;
+	color: #e5e5e5;
 }
 ColorDefinition#org-eclipse-ui-workbench-ACTIVE_TAB_INNER_KEYLINE_COLOR {
-	color: #f8f8f8;
+	color: #f8f8f8; /* same as '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND' cannot use it directly*/
 }
 ColorDefinition#org-eclipse-ui-workbench-ACTIVE_TAB_OUTLINE_COLOR {
-	color: #E5E5E5;
+	color: #e5e5e5;
 }
 
 ColorDefinition#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START {
-	color: #F8F8F8;
+	color: #f8f8f8; /* same as '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND' cannot use it directly*/
 }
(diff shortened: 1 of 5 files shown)


BeckerWdf:
I only tested on macOS.

BeckerWdf:
@mvm-sap, @sratz what do you think?

BeckerWdf:
> The new defined color is now also visible in the "Colors and Fonts" preference page. I am unsure if this is something we want. If users change this this does not work 100% because of the limitation above.

With [845377a](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2455/commits/845377a245d979d94f16f3fa7c881cf524f66ecb) this issue is fixed.

bundles/org.eclipse.ui.themes/css/e4_preview_gtk.css
```diff
@@ -33,11 +33,11 @@ ColorDefinition#org-eclipse-ui-workbench-ACTIVE_TAB_OUTLINE_COLOR {
 }
 
 ColorDefinition#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START {
-	color: #F8F8F8;
+	color: #F8F8F8; /* same as '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND' cannot use '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND' directly*/
```
sratz: maybe shorten this a bit to not have to mention it twice:
```suggestion
	color: #F8F8F8; /* same as '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND' cannot reference it directly*/
```

Is this something we can improve technically?

sratz (commented):
Are we somehow committing to the id of the definition `org-eclipse-ui-workbench-SECONDARY_BACKGROUND` and have to keep it stable?

I guess with the latest `hidden` flag, we can consider this ID to be purely internal and should be able to rename this?

bundles/org.eclipse.ui.themes/css/e4_preview_gtk.css
BeckerWdf: > maybe shorten this a bit to not have to mention it twice:
> 
> Is this something we can improve technically?

I am not sure. I didn't yet have a look into it.

BeckerWdf:
> Are we somehow committing to the id of the definition `org-eclipse-ui-workbench-SECONDARY_BACKGROUND` and have to keep it stable?
> 
> I guess with the latest `hidden` flag, we can consider this ID to be purely internal and should be able to rename this?

I would say we can consider it internal as it only appears in the CSS files. Anyway: Do you have a better name?

sratz:
> I would say we can consider it internal as it only appears in the CSS files.

👍

> Anyway: Do you have a better name?

No, hence the question 😉

vogella:
I think it is great to allow the user to change the default colour. I think it should be relatively easy to implement that colors refer to others in CSS. What exactly is the requirement here @BeckerWdf ?

BeckerWdf:
> think it should be relatively easy to implement that colors refer to others in CSS. What exactly is the requirement here?

What I want to do is this:

```
ColorDefinition#org-eclipse-ui-workbench-ACTIVE_UNSELECTED_TABS_COLOR_START {
	color: '#org-eclipse-ui-workbench-SECONDARY_BACKGROUND';
}
```
instead of:
```
ColorDefinition#org-eclipse-ui-workbench-ACTIVE_UNSELECTED_TABS_COLOR_START {
	color: #F8F8F8;
}
```

@vogella: Would be great if you could help here. Feel free to push a separate PR / an additional commit to this PR

BeckerWdf:
This PR needs to be reworked after #2549 is merged.

BeckerWdf:
pr-head fails with:

`12:20:08.576 [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.10:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jface: Baseline and reactor have the same fully qualified version, but different content`

@HannesWell: What's the issue here?

PR #2456, opened 2024-10-28
Get rid of ArrayUtil.equals(Object[], Object[])

akurtakov:
Use JUnit's assertArrayEquals or JVM's Arrays.equals instead.


akurtakov:
Unrelated test failure.

Issue #2458, opened 2024-10-29
DynamicTest.testDynamicRegistry fails randomly


fedejeanne:
Test `org.eclipse.ui.tests.activities.DynamicTest.testDynamicRegistry` fails randomly.

Last time it failed on **Mac** in this PR: #2407 

```java
org.eclipse.ui.tests.activities.DynamicTest.testDynamicRegistry -- Time elapsed: 21.31 s <<< FAILURE!
java.lang.AssertionError: Activity Listener not called
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.ui.tests.activities.DynamicTest.testDynamicRegistry(DynamicTest.java:480)
...
```

PR #2459, opened 2024-10-29
Ignore missing versions on required bundles for Java editor examples

fedejeanne:
Change the PDE preferences from "Warning" to "Ignore".

image


fedejeanne:
Test failure is unrelated #2458

PR #2461, opened 2024-10-29
[win] Adding experimental preference to enable update on runtime behavior

### LINKED ISSUE #134: Add experimental preference to enable update on runtime behavior
Where: 
UI:
- Preferences -> General -> Appereance
- Best case: Preference is only visible on Windows

Code:
- perhaps org.eclipse.e4.ui.swt.win32

What is affected:
- swt.autoscale: if not active -> change nothing, if active -> change
- Browser: , swt.autoscale, swt.autoscale.updateOnRuntime

What:
- Evaluate preference at workbench initialization: 
- configure Display accordingly: Display::setRescalingAtRuntime
- Edge as default browser -> Set the system property "org.eclipse.swt.browser.DefaultType" to edge

This preference must be disabled by default

ShahzaibIbrahim:
In the UI Preferences -> General -> Appearance, there is a new HiDPI setting that could be checked to enable the monitor-specific scaling of the UI. This check will also enable Edge browser by default. This is an experimental feature to test the multi-monitor HiDPI support before it is actually released and set by default. For now, this preference is disabled by default.

**Note: This preference check is only available in Windows** 

image

When changing the preference, the dialog will appear to prompt the restart. 

image

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
@@ -705,4 +705,14 @@ public interface IWorkbenchPreferenceConstants {
 	 * @since 3.130
 	 */
 	String LARGE_VIEW_LIMIT = "largeViewLimit"; //$NON-NLS-1$
+
+	/**
+	 * <p>
+	 * <strong>EXPERIMENTAL</strong>. Whether the UI adapts to DPI changes at
+	 * runtime. It only effects Windows.
+	 * </p>
+	 *
+	 * @since 3.133
+	 */
+	String RESCALING_AT_RUNTIME = "RESCALING_AT_RUNTIME"; //$NON-NLS-1$
 }
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -584,7 +584,7 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 				int orientation = store.getInt(IPreferenceConstants.LAYOUT_DIRECTION);
 				Window.setDefaultOrientation(orientation);
 			}
-
+			setRescaleAtRuntimePropertyFromPreference(display);
 			if (obj instanceof E4Application) {
 				E4Application e4app = (E4Application) obj;
 				E4Workbench e4Workbench = e4app.createE4Workbench(getApplicationContext(), display);
@@ -676,6 +676,15 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 			}
 		});
 		return returnCode[0];
+	}
+
+	private static void setRescaleAtRuntimePropertyFromPreference(final Display display) {
+		boolean rescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
+				.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
+		if (rescaleAtRuntime) {
+			display.setRescalingAtRuntime(rescaleAtRuntime);
+			System.setProperty("org.eclipse.swt.browser.DefaultType", "edge"); //$NON-NLS-1$ //$NON-NLS-2$
+		}
 	}
 
 	private static void setSearchContribution(MApplication app, boolean enabled) {
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchMessages.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchMessages.java
@@ -34,6 +34,16 @@ public class WorkbenchMessages extends NLS {
 
 	public static String ThemingEnabled;
 
+	public static String HiDpiSettingsGroupTitle;
+
+	public static String RescaleAtRuntimeEnabled;
+
+	public static String RescaleAtRuntimeDisclaimer;
+
+	public static String RescaleAtRuntimeSettingChangeWarningTitle;
+
+	public static String RescaleAtRuntimeSettingChangeWarningText;
+
 	public static String ThemeChangeWarningText;
 
 	public static String ThemeChangeWarningTitle;
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
@@ -37,6 +37,7 @@ import org.eclipse.core.runtime.IExtension;
 import org.eclipse.core.runtime.IExtensionPoint;
 import org.eclipse.core.runtime.IExtensionRegistry;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Platform.OS;
 import org.eclipse.core.runtime.RegistryFactory;
 import org.eclipse.core.runtime.preferences.DefaultScope;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
@@ -53,6 +54,7 @@ import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.fieldassist.ControlDecoration;
 import org.eclipse.jface.fieldassist.FieldDecorationRegistry;
+import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.preference.PreferencePage;
 import org.eclipse.jface.util.Util;
@@ -73,6 +75,7 @@ import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Group;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPreferenceConstants;
@@ -114,6 +117,7 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 	private boolean highContrastMode;
 
 	private Button themingEnabled;
+	private Button rescaleAtRuntime;
 
 	private Button hideIconsForViewTabs;
 	private Button showFullTextForViewTabs;
@@ -135,6 +139,7 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 			layout.horizontalSpacing = 10;
 			comp.setLayout(layout);
 			createThemeIndependentComposits(comp);
+			createHiDPISettingsGroup(comp);
 			return comp;
 		}
 
@@ -180,6 +185,8 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 		createHideIconsForViewTabs(comp);
 		createDependency(showFullTextForViewTabs, hideIconsForViewTabs);
 
+		createHiDPISettingsGroup(comp);
+
 		if (currentTheme != null) {
 			String colorsAndFontsThemeId = getColorAndFontThemeIdByThemeId(currentTheme.getId());
 			if (colorsAndFontsThemeId != null && !currentColorsAndFontsTheme.getId().equals(colorsAndFontsThemeId)) {
@@ -190,6 +197,30 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 
 		Dialog.applyDialogFont(comp);
 		return comp;
+	}
+
+	private void createHiDPISettingsGroup(Composite parent) {
+		if (!OS.isWindows()) {
+			return;
+		}
+		createLabel(parent, ""); //$NON-NLS-1$
+		Group group = new Group(parent, SWT.LEFT);
+		group.setText(WorkbenchMessages.HiDpiSettingsGroupTitle);
+
+		GridData gridData = new GridData(SWT.FILL, SWT.CENTER, true, false);
+		gridData.horizontalSpan = ((GridLayout) parent.getLayout()).numColumns;
+		group.setLayoutData(gridData);
+		group.setFont(parent.getFont());
+		GridLayout layout = new GridLayout(1, false);
+		group.setLayout(layout);
+		Label infoLabel = new Label(group, SWT.WRAP);
+		infoLabel.setText(WorkbenchMessages.RescaleAtRuntimeDisclaimer);
+		infoLabel.setLayoutData(GridDataFactory.defaultsFor(infoLabel).create());
+		createLabel(group, ""); //$NON-NLS-1$
+
+		boolean initialStateRescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
+				.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
+		rescaleAtRuntime = createCheckButton(group, WorkbenchMessages.RescaleAtRuntimeEnabled, initialStateRescaleAtRuntime);
 	}
 
 	private void createThemeIndependentComposits(Composite comp) {
@@ -227,7 +258,6 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 		GridData gridData = new GridData();
 		gridData.horizontalIndent = 20;
 		dependent.setLayoutData(gridData);
-
 		boolean parentState = parent.getSelection();
 		dependent.setEnabled(parentState);
 
@@ -341,6 +371,14 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 				.getSelection();
 		prefs.putBoolean(PartRenderingEngine.ENABLED_THEME_KEY, themingEnabled.getSelection());
 
+		boolean isRescaleAtRuntimeChanged = false;
+		if (rescaleAtRuntime != null) {
+			boolean initialStateRescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
+					.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
+			isRescaleAtRuntimeChanged = initialStateRescaleAtRuntime != rescaleAtRuntime.getSelection();
+			apiStore.setValue(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, rescaleAtRuntime.getSelection());
+		}
+
 		prefs.putBoolean(CTabRendering.USE_ROUND_TABS, useRoundTabs.getSelection());
 		try {
 			prefs.flush();
@@ -367,19 +405,22 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 			colorFontsDecorator.hide();
 
 			if (themeChanged || colorsAndFontsThemeChanged) {
-				showRestartDialog();
+				showRestartDialog(WorkbenchMessages.ThemeChangeWarningTitle, WorkbenchMessages.ThemeChangeWarningText);
 			}
 		}
 		if (themingEnabledChanged) {
-			showRestartDialog();
+			showRestartDialog(WorkbenchMessages.ThemeChangeWarningTitle, WorkbenchMessages.ThemeChangeWarningText);
+		}
+		if (isRescaleAtRuntimeChanged) {
+			showRestartDialog(WorkbenchMessages.RescaleAtRuntimeSettingChangeWarningTitle,
+					WorkbenchMessages.RescaleAtRuntimeSettingChangeWarningText);
 		}
 
 		return super.performOk();
 	}
 
-	private void showRestartDialog() {
-		if (new MessageDialog(null, WorkbenchMessages.ThemeChangeWarningTitle, null,
-				WorkbenchMessages.ThemeChangeWarningText, MessageDialog.NONE, 2,
+	private void showRestartDialog(String title, String warningText) {
+		if (new MessageDialog(null, title, null, warningText, MessageDialog.NONE, 2,
 				WorkbenchMessages.Workbench_RestartButton, WorkbenchMessages.Workbench_DontRestartButton)
 						.open() == Window.OK) {
 			Display.getDefault().asyncExec(() -> PlatformUI.getWorkbench().restart());
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
@@ -497,6 +497,11 @@ PreferencePageParameterValues_pageLabelSeparator = \ >\
 ThemingEnabled = E&nable theming
 ThemeChangeWarningText = Restart for the theme changes to take full effect
 ThemeChangeWarningTitle = Theme Changed
+RescaleAtRuntimeSettingChangeWarningTitle = DPI Setting Changed
+RescaleAtRuntimeSettingChangeWarningText = Restart for the DPI setting changes to take effect
+HiDpiSettingsGroupTitle = HiDPI settings
+RescaleAtRuntimeEnabled = Monitor-specific UI &scaling
+RescaleAtRuntimeDisclaimer = EXPERIMENTAL! Activating this option will dynamically scale all windows according to the monitor they are currently in. It will also set the default browser to Edge in order to provide the appropriate scaling of content displayed in a browser. This feature is still in development and therefore considered experimental.
 # --- Workbench -----
 WorkbenchPreference_openMode=Open mode
 WorkbenchPreference_doubleClick=D&ouble click


laeubi:
As we have ways to know the platform, maybe only show it when running on windows?

Also either make PerMonitorV2 a link to some documentation or remove implementation details from the description.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -192,6 +200,32 @@ protected Control createContents(Composite parent) {
...
+
+		GridData gridData = new GridData(SWT.FILL, SWT.CENTER, true, false);
+		gridData.horizontalSpan = ((GridLayout) composite.getLayout()).numColumns;
+		buttonComposite.setLayoutData(gridData);
+		buttonComposite.setFont(composite.getFont());
+		GridLayout layout = new GridLayout(1, false);
+		buttonComposite.setLayout(layout);
+		Label infoLabel = new Label(buttonComposite, SWT.WRAP);
+		infoLabel.setText(
+				"This preference is Windows only. Activating it will change the DPI awareness of the IDE to PerMonitorV2 and dynamically scale all windows according to the monitor it is currently assigned to. An additional effect of this preference is to set the default browser to Edge to provide the appropriate scaling of content displayed in a browser. This feature is still in development and therefore marked as experimental."); //$NON-NLS-1$
```
BeckerWdf: Shouldn't we move this text to a messages.properties file?

I would write:
"This setting is specific to the Windows operating system..."

ShahzaibIbrahim:
> As we have ways to know the platform, maybe only show it when running on windows?

@laeubi Forgot to mention but yes, we are only showing it in Windows. I have updated the description.

laeubi:
If it is only shown on windows, then I don't think it needs to be mentioned as a "Windows only" in the text.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
ShahzaibIbrahim: The only reason I didn't bother to put the message in the file is that later it would be easier to remove this temporary option (just from one place) but yeah  I will move it.

BeckerWdf:
> Also either make PerMonitorV2 a link to some documentation or remove implementation details from the description.

One could also add such details into the documentation page for this preference page.

ShahzaibIbrahim:
@BeckerWdf @laeubi for now I have removed implementation details from the description. I have asked if we have documentation page for this right now. If yes, then I will link it.

BeckerWdf (approved):

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -656,6 +656,17 @@ public void update() {
...
 	}
 
+	private static void setRuntimePropertyFromPreference(final Display display) {
+		boolean updateOnRunTime = PrefUtil.getAPIPreferenceStore()
+				.getBoolean(IWorkbenchPreferenceConstants.UPDATE_ON_RUNTIME_KEY);
+		display.setRescalingAtRuntime(updateOnRunTime);
+		if (updateOnRunTime) {
+			System.setProperty("org.eclipse.swt.browser.DefaultType", "edge"); //$NON-NLS-1$ //$NON-NLS-2$
+		} else {
+			System.setProperty("org.eclipse.swt.browser.DefaultType", "ie"); //$NON-NLS-1$ //$NON-NLS-2$
```
fedejeanne: By explicitly setting this to `ie` you are also overriding the value of `-Dorg.eclipse.swt.browser.DefaultType` that might have been set from outside. I would remove the `else` and let the user select whichever browser s/he wants if "update on runtime" is not activated.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -114,6 +116,9 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 	private boolean highContrastMode;
 
 	private Button themingEnabled;
+	private Button updateOnRunTime;
+	boolean initialStateUpdateOnRuntime = PrefUtil.getAPIPreferenceStore()
```
fedejeanne: make it `private`

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -192,6 +200,31 @@ protected Control createContents(Composite parent) {
 		return comp;
 	}
 
+	private void createUpdateOnRuntimeButton(Composite composite) {
```
fedejeanne:
```suggestion
	private void createHiDPISettingsGroup(Composite parent) {
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -192,6 +200,31 @@ protected Control createContents(Composite parent) {
...
+		if (!OS.isWindows()) {
+			return;
+		}
+		createLabel(composite, ""); //$NON-NLS-1$
+		Group buttonComposite = new Group(composite, SWT.LEFT);
```
fedejeanne:
```suggestion
		Group group = new Group(composite, SWT.LEFT);
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -341,6 +373,11 @@ public boolean performOk() {
 				.getSelection();
 		prefs.putBoolean(PartRenderingEngine.ENABLED_THEME_KEY, themingEnabled.getSelection());
 
+		boolean isUpdateOnRuntTmeChanged = (updateOnRunTime != null
```
fedejeanne: fix 2 typos
```suggestion
		boolean isUpdateOnRuntimeChanged = (updateOnRuntime != null
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -373,13 +410,19 @@ public boolean performOk() {
 		if (themingEnabledChanged) {
 			showRestartDialog();
 		}
+		if (isUpdateOnRuntTmeChanged) {
+			showRestartDialog("DPI Setting Changed", "Restart for the effect to take place"); //$NON-NLS-1$ //$NON-NLS-2$
```
fedejeanne: Externalize these strings.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
```diff
@@ -497,6 +497,8 @@ PreferencePageParameterValues_pageLabelSeparator = \ >\
 ThemingEnabled = E&nable theming
 ThemeChangeWarningText = Restart for the theme changes to take full effect
 ThemeChangeWarningTitle = Theme Changed
+UpdateUIZoomTitle = Update UI scaling on runtime
+UpdateUIZoomDesc = Activating it will dynamically scale all windows according to the monitor it is currently assigned to. An additional effect of this preference is to set the default browser to Edge to provide the appropriate scaling of content displayed in a browser. This feature is still in development and therefore marked as experimental.
```
fedejeanne:
```suggestion
UpdateUIZoomDesc = EXPERIMENTAL! Activating this option will dynamically scale all windows according to the monitor they are currently in. It will also set the default browser to Edge in order to provide the appropriate scaling of content displayed in a browser. This feature is still in development and therefore considered experimental.
```

fedejeanne (dismissed):
It works pretty well, I only have some minor comments about naming/typos and one important detail about preserving the selected browser when the new preference is set to `false`.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
```diff
@@ -705,4 +705,13 @@ public interface IWorkbenchPreferenceConstants {
 	 * @since 3.130
 	 */
 	String LARGE_VIEW_LIMIT = "largeViewLimit"; //$NON-NLS-1$
+
+	/**
+	 * <p>
+	 * Whether the UI updates on runtime during dpi change
+	 * </p>
```
HeikoKlare: It's not only about updating the UI, it's actually about adapting to DPI changes
```suggestion
	 * <p>
	 * Whether the UI adapts to DPI changes at runtime
	 * </p>
```
Please also flag this preference as "experimental" and state that it only affects Windows.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
```diff
@@ -705,4 +705,13 @@ public interface IWorkbenchPreferenceConstants {
...
+	 *
+	 * @since 3.133
+	 */
+	String UPDATE_ON_RUNTIME_KEY = "UPDATE_ON_RUNTIME_KEY"; //$NON-NLS-1$
```
HeikoKlare: `UPDATE_ON_RUNTIME` is very generic and might conflict to other things. Could we use something more specific like at least `RESCALE_ON_RUNTIME` or `ADAPT_DPI_AT_RUNTIME` or the like?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -678,6 +678,15 @@ public void update() {
 		return returnCode[0];
 	}
 
+	private static void setRuntimePropertyFromPreference(final Display display) {
+		boolean updateOnRunTime = PrefUtil.getAPIPreferenceStore()
```
HeikoKlare: Please use consistent way of writing "runtime" --> `updateOnRuntime`.
And please use more descriptive name (see previous comment)

fedejeanne (approved):

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -678,6 +678,15 @@ public void update() {
 		return returnCode[0];
 	}
 
+	private static void setRuntimePropertyFromPreference(final Display display) {
```
HeikoKlare: This is a very generic method name for a very specific task. Shouldn't is be specific to processing the preference for rescaling?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
```diff
@@ -678,6 +678,15 @@ public void update() {
...
+		boolean updateOnRunTime = PrefUtil.getAPIPreferenceStore()
+				.getBoolean(IWorkbenchPreferenceConstants.UPDATE_ON_RUNTIME_KEY);
+		display.setRescalingAtRuntime(updateOnRunTime);
```
HeikoKlare: This changes existing behavior. Executing `display.setRescalingAtRuntime(false)` makes the application (more specifically the display thread) run in `System` DPI awareness mode while currently it just runs with whatever it was started with. So this must be put in the if statement as well.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/WorkbenchMessages.java
```diff
@@ -34,6 +34,14 @@ public class WorkbenchMessages extends NLS {
 
 	public static String ThemingEnabled;
 
+	public static String UpdateUIZoomTitle;
+
+	public static String UpdateUIZoomDesc;
+
+	public static String RuntimeSettingChangeWarningTitle;
+
+	public static String RuntimeSettingChangeWarningDesc;
```
HeikoKlare: Please use comprehensible names (see previous comments). In particular "RuntimeSettingChangeWarningTitle" is rather unspecific.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -192,6 +200,31 @@ protected Control createContents(Composite parent) {
...
+		GridData gridData = new GridData(SWT.FILL, SWT.CENTER, true, false);
+		gridData.horizontalSpan = ((GridLayout) parent.getLayout()).numColumns;
+		group.setLayoutData(gridData);
+		group.setFont(parent.getFont());
+		GridLayout layout = new GridLayout(1, false);
+		group.setLayout(layout);
+		Label infoLabel = new Label(group, SWT.WRAP);
+		infoLabel.setText(WorkbenchMessages.UpdateUIZoomDesc);
+		GridData labelGridData = new GridData(SWT.FILL, SWT.CENTER, true, false);
+		labelGridData.widthHint = 400; // Adjust this value to control initial wrapping width
```
HeikoKlare: Hardcoding this width value look inappropriate. Itsn't there an existing constant for this? Or any other way of making this horizontally fill the dialog, whatever its size may be?

Note that there is also `GridDataFactory` with some default to simplify the process of layout data generation.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -212,6 +245,9 @@ protected void createHideIconsForViewTabs(Composite composite) {
 				CTabRendering.HIDE_ICONS_FOR_VIEW_TABS_DEFAULT);
 		hideIconsForViewTabs = createCheckButton(composite, WorkbenchMessages.ViewsPreference_hideIconsForViewTabs,
 				actualValue);
+		GridData gridData = new GridData();
+		gridData.horizontalIndent = 20;
+		hideIconsForViewTabs.setLayoutData(gridData);
```
HeikoKlare: Why it is necessary to now apply this grid data to `hideIconsForViewTabs` instead of `dependent` in the `createDependency` method like before?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
```diff
@@ -497,6 +497,10 @@ PreferencePageParameterValues_pageLabelSeparator = \ >\
 ThemingEnabled = E&nable theming
 ThemeChangeWarningText = Restart for the theme changes to take full effect
 ThemeChangeWarningTitle = Theme Changed
+RuntimeSettingChangeWarningTitle = DPI Setting Changed
+RuntimeSettingChangeWarningDesc = Restart for the effect to take place
+UpdateUIZoomTitle = Update UI scaling on runtime
```
HeikoKlare: May it be more self-explaining when putting the `monitor-specific` property in the label? E.g.
```suggestion
UpdateUIZoomTitle = Monitor-specific UI scaling
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -341,6 +373,11 @@ public boolean performOk() {
 				.getSelection();
 		prefs.putBoolean(PartRenderingEngine.ENABLED_THEME_KEY, themingEnabled.getSelection());
 
+		boolean isUpdateOnRuntimeChanged = (updateOnRuntime != null
+				&& initialStateUpdateOnRuntime != updateOnRuntime.getSelection());
+		apiStore.setValue(IWorkbenchPreferenceConstants.UPDATE_ON_RUNTIME_KEY,
+				updateOnRuntime == null ? false : updateOnRuntime.getSelection());
```
HeikoKlare: Shouldn't this whole block be put into a `updateOnRuntime != null` condition? Why should the value be stored in the preference store if the button is not present (i.e., if we are not on Windows, if I understand correctly) at all?

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -373,13 +410,20 @@ public boolean performOk() {
...
 	}
 
 	private void showRestartDialog() {
-		if (new MessageDialog(null, WorkbenchMessages.ThemeChangeWarningTitle, null,
-				WorkbenchMessages.ThemeChangeWarningText, MessageDialog.NONE, 2,
+		showRestartDialog(WorkbenchMessages.ThemeChangeWarningTitle, WorkbenchMessages.ThemeChangeWarningText);
+	}
+
+	private void showRestartDialog(String title, String warningText) {
+		if (new MessageDialog(null, title, null, warningText, MessageDialog.NONE, 2,
```
HeikoKlare: Please properly refactor this to either always have the consumer decise which title and text to put in the dialog or to have one parameterless method per use case. Now we have one method for custom title/text and one with some unspecified default values (which by accident default to themeing). I would propose to just have the method with two parameters and make the method consumers pass the required information.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -114,6 +116,9 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 	private boolean highContrastMode;
 
 	private Button themingEnabled;
+	private Button updateOnRuntime;
+	private boolean initialStateUpdateOnRuntime = PrefUtil.getAPIPreferenceStore()
+			.getBoolean(IWorkbenchPreferenceConstants.UPDATE_ON_RUNTIME_KEY);
```
HeikoKlare: Why is this solved different then the themeing property does it? That one is simply retrieved from the preference store when identifying whether the value has changed instead of storing it in a field. In my opinion, that solution is better as this field just serves as an unnecessary and potentially error-prone cache.

HeikoKlare (changes requested):
Great to finally have a proper preference for enabling the enhanced HiDPI support on Windows, thanks!

Still, in the current state, this PR introduces a regression and needs some revision regarding design and naming. See my detailed comments.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
fedejeanne: That would be a 350

```java
infoLabel.setLayoutData(GridDataFactory.defaultsFor(infoLabel).create());
```

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
fedejeanne: I'd go with `RESCALING_AT_RUNTIME` since it matches the **public** (API) method `Display::setRescalingAtRuntime` and would adapt the names of other methods, parameters, constants, fields and information texts in this PR too in order to keep it consistent.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
ShahzaibIbrahim: It's a leftover. I was trying something and later forgot to move it back.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -127,6 +132,8 @@ protected Control createContents(Composite parent) {
 		Composite comp = new Composite(parent, SWT.NONE);
 
 		themingEnabled = createCheckButton(comp, WorkbenchMessages.ThemingEnabled, engine != null);
+		initialStateUpdateOnRuntime = PrefUtil.getAPIPreferenceStore()
+				.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
```
HeikoKlare: As mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2461#discussion_r1824038278, I do not yet understand why we need to store this information in a field instead of retrieving it when required, like done for the themeing.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -114,6 +117,8 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 	private boolean highContrastMode;
 
 	private Button themingEnabled;
+	private Button updateOnRuntime;
```
HeikoKlare: Can we call this `rescaleAtRuntime` to be consistent with the general naming? Same applies to other variables in this class.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/messages.properties
```diff
@@ -497,6 +497,11 @@ PreferencePageParameterValues_pageLabelSeparator = \ >\
 ThemingEnabled = E&nable theming
 ThemeChangeWarningText = Restart for the theme changes to take full effect
 ThemeChangeWarningTitle = Theme Changed
+RescaleAtRuntimeSettingChangeWarningTitle = DPI Setting Changed
+RescaleAtRuntimeSettingChangeWarningDesc = Restart for the effect to take place
+UpdateUIZoomGroupTitle = HiDPI settings
+UpdateUIZoomTitle = Monitor-specific UI scaling
+UpdateUIZoomDesc = EXPERIMENTAL! Activating this option will dynamically scale all windows according to the monitor they are currently in. It will also set the default browser to Edge in order to provide the appropriate scaling of content displayed in a browser. This feature is still in development and therefore considered experimental.
```
HeikoKlare: Can we please improve the property names here?
E.g.:
- `UpdateUIZoomGroupTitle` -> `HiDpiSettingsGroupTitle`
- `UpdateUIZoomTitle` -> `RescaleAtRuntimeEnabled`
- `UpdateUIZoomTitle` -> `RescaleAtRuntimeDisclaimer`

And maybe add a mnemonic for the new option? If I see correctly "s" is not used yet, so you might use `Monitor-specific UI &scaling`

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -192,6 +202,28 @@ protected Control createContents(Composite parent) {
 		return comp;
 	}
 
+	private void createHiDPISettingsGroup(Composite parent) {
+		if (!OS.isWindows() || true) {
```
HeikoKlare: The options do not appear anymore because this always evaluates to `true`, such that the method is never executed.

HeikoKlare (commented):
Regarding functionality, this looks good now (except for the case that the options do not show up anymore because of a faulty conditions).

Some minor code style comments are remaining.

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
ShahzaibIbrahim: So that it could be used across the class. I needed this field info while creating the HiDpi setting group and also when Ok is performed. For theming they checked the initial state by simply checking if theme != null. I hope it make sense

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
ShahzaibIbrahim: Sorry, leftover...

bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
ShahzaibIbrahim: Now, I am retrieving the value only in the methods where its needed, separately.

HeikoKlare (approved):
From my side, this also looks good now. Thank you!

HeikoKlare:
Failing test is unrelated/already documented: #370

PR #2464, opened 2024-10-29
Stabilize DynamicTest

akurtakov:
* Redo assertions to use assertEquals for better reporting.
* Remove useless catch blocks with fails
* Use DisplayHelper to spin the event loop


akurtakov:
Should help with https://github.com/eclipse-platform/eclipse.platform.ui/issues/2458

PR #2469, opened 2024-10-30
Place Edge browser userdata directory in metadata space of workspace

HeikoKlare:
The userdata directory for the Edge browser is currently placed as an immediate child of the workspace directory (if existing), thus mixing up with actual projects in the workspace. With this change, the directory is placed in the metadata folder of the SWT bundle within the workspace.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/2434#issuecomment-2447396324

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/Workbench.java
@@ -147,6 +147,7 @@ import org.eclipse.osgi.service.runnable.StartupMonitor;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.SWTException;
+import org.eclipse.swt.browser.Browser;
 import org.eclipse.swt.custom.BusyIndicator;
 import org.eclipse.swt.graphics.DeviceData;
 import org.eclipse.swt.graphics.FontData;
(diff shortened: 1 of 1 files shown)


sratz:
Looks good in principle.

One more thought: Is this EBWebView directory platform-independet or could we potentially run into inconsistencies / problems in case Windows-on-ARM users migrate from `win32.win32.x86_64` to `win32.win32.aarch64` with the same workspace?

HeikoKlare:
> One more thought: Is this EBWebView directory platform-independet or could we potentially run into inconsistencies / problems in case Windows-on-ARM users migrate from `win32.win32.x86_64` to `win32.win32.aarch64` with the same workspace?

Interesting question/scenario. The `EBWebView` folder is placed inside the `.metadata/.plugins/org.eclipse.swt` folder of the workspace. So this is independent of the platform (architecture) and I would expect this to also work in case of a migration. Do you agree?
{2CF1E943-F9C4-443C-83A2-D9BF528B8A9F}

I am not sure whether the data within that folder allows for migration between different systems/architectures, but that issue would at least be independent of where the folder is placed.

sratz:
Right, my thought was whether there is anything platform-specific inside that folder, and we should instead put it into the fragment directories
`.metadata/.plugins/org.eclipse.swt.win32.win32.x86_64`
`.metadata/.plugins/org.eclipse.swt.win32.win32.aarch64`

But that is probably overthinking the problem, since if we did not override this path and stick to the default path, it also would be a generic platform-independent directory.

So would the `%AppData%\Local\Microsoft\Edge\User Data\` directory of a regular Edge browser for that matter.


Remaining question: Who is the 'owner' of that directory
- Is it `org.eclipse.ui.workbench` because here we make the decision of where to put the Edge cache?
- Is it `org.eclipse.swt` because SWT is what uses it?

HeikoKlare:
> Remaining question: Who is the 'owner' of that directory
> 
> * Is it `org.eclipse.ui.workbench` because here we make the decision of where to put the Edge cache?
> * Is it `org.eclipse.swt` because SWT is what uses it?

I am not sure whether there is a "correct" answer in the sense that there is some agreement, convention or the like that indisputably determines where to place it. I would go with the latter open (in `org.eclipse.swt`), because SWT contains and operates the browser, so I consider the browser user data to be metadata belonging to SWT. Putting it in `org.eclipse.ui.workbench` would feel a bit more "random" to me, because you could probably actually find a different place (i.e., bundle) in Eclipse UI where the directory could be determined. But I have no strong opinion on this, so if someone has valid arguments for another option, I will glady adapt the PR.

sratz (approved):

HeikoKlare:
Failing test already documented: #370

PR #2472, opened 2024-10-31
Avoid "null" in "Show In" menu when there is no key binding

orionll:
When the "Show In" command has no key binding (we disable it for our reasons in our product), there is a `null` label in the menu:

image

The proposed change fixes this issue.

### DIFF
--- a/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/ShowInActionProvider.java
+++ b/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/ShowInActionProvider.java
@@ -45,6 +45,6 @@ public class ShowInActionProvider extends CommonActionProvider {
 				? bindingService
 						.getBestActiveBindingFormattedFor(IWorkbenchCommandConstants.NAVIGATE_SHOW_IN_QUICK_MENU)
 				: ""; //$NON-NLS-1$
-		return WorkbenchNavigatorMessages.ShowInActionProvider_showInAction_label + (keyBinding != null ? '\t' + keyBinding : ""); //$NON-NLS-2$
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/ShowInActionProvider.java
```diff
@@ -45,6 +45,6 @@ private String getShowInMenuLabel() {
 				? bindingService
 						.getBestActiveBindingFormattedFor(IWorkbenchCommandConstants.NAVIGATE_SHOW_IN_QUICK_MENU)
 				: ""; //$NON-NLS-1$
-		return WorkbenchNavigatorMessages.ShowInActionProvider_showInAction_label + '\t' + keyBinding;
+		return WorkbenchNavigatorMessages.ShowInActionProvider_showInAction_label + (keyBinding != null ? "\t" + keyBinding : ""); //$NON-NLS-1$ //$NON-NLS-2$
```
Wittmaxi, line 48: From a static analysis, this looks sound. I looked into other ActionProviders and couldn't see in a glance why they don't have the same problem and how they solve that problem. Is there any mechanism we might have missed here?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/ShowInActionProvider.java
```diff
@@ -45,6 +45,6 @@ private String getShowInMenuLabel() {
...
 	}
-}
\ No newline at end of file
+}
```
Wittmaxi, line 50: please keep the diff clean!

Issue #416, opened 2022-10-17
[Dark Theme] Checkbox created with toolkit on dialog has wrong background


marcushoepfner:
A button with SWT.CHECK gets a wrong background color (light grey instead of dark grey) under the following conditions:

- created on a dialog
- toolkit used (FormToolkit.createButton)
- darktheme

In an editor this does not happen.

<img width="153" alt="Screenshot 2022-10-17 at 13 30 50" src="https://user-images.githubusercontent.com/29301444/196166511-a59e5ed3-404c-47a6-b7fc-66ab3c514fc9.png">

Does happen on Mac. Not on Windows.

akurtakov:
Can you share the snippet to reproduce?

marcushoepfner:
sure.
[background.test_1.0.0.202210171431.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/9800462/background.test_1.0.0.202210171431.zip)

(It's actually a jar, but github does not support. You might need to rename back to .jar)

marcushoepfner:
I have started analyzing this:
- FormToolkit calls adapt() when creating the checkbox; a background color is set
- afterwards the css engine overwrites this background color via Control.setBackground(Color color) with null
- this leads to the light grey eventually

Compared to e.g. a label created with toolkit
- FormToolkit calls adapt() when creating the label; a background color is set
-  afterwards the css engine overwrites this background color via Control.setBackground(Color color) with a concrete color

Is just the color in css missing?

marcushoepfner:
Does someone has a hint how to fix this? If so, I can take over and try.

akurtakov:
Can you share the code for this jar so it can be used for debugging?

marcushoepfner:
Oh damn. I forgot. Thanks for bearing with me.
Here it is, same disclaimer: rename from zip to jar
[background.test.source_1.0.0.202210191250.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/9819767/background.test.source_1.0.0.202210191250.zip)

marcushoepfner:
Hi, any progress on this one?

akurtakov:
Unfortunately not many people look into MacOS specific issues (@Phillipus only lately) . In my opinion this could be SWT bug but it could be css styling issue too. In order to get this issue moving people with interest (and access) in MacOS have to step up but you can further help by trying to reduce the snippet to pure SWT snippet one .

Phillipus:
[...]

PR #2474, opened 2024-10-31
Better Checkbox Coloring on Dialog when using FormToolkit

marcushoepfner:
Issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/416

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/e4-dark_mac.css
+++ b/bundles/org.eclipse.ui.themes/css/e4-dark_mac.css
@@ -48,7 +48,7 @@ Tree[swt-lines-visible=true] {
 	background-color: unset;
 }
 
-Button  {
+Button[style~='SWT.PUSH']  {
 	background-color: unset;
 	color: unset;
 }


marcushoepfner:
Before:
Screenshot 2024-10-31 at 14 22 56

After:
Screenshot 2024-10-31 at 14 23 55

But unfortunately also influences:
Screenshot 2024-10-31 at 14 31 09

So there is a bit more to fix.
@BeckerWdf, can you help here?

vogella:
Could you use the shell style for selection?

/* Style for SWT.PUSH buttons in dialogs */
Shell[style*=SWT.APPLICATION_MODAL] Button[type=SWT.PUSH] {
    background-color: #007ACC; /* Change to your desired background color */
}

bundles/org.eclipse.ui.themes/css/e4-dark_mac.css
```diff
@@ -48,7 +48,7 @@ Tree[swt-lines-visible=true] {
 	background-color: unset;
 }
 
-Button  {
+Button[style~='SWT.PUSH']  {
 	background-color: unset;
```
BeckerWdf, line 52: we could also try to use "inherit" instead of "unset"

PR #2224, opened 2024-08-27
Added dynamic tab alignment support in MultiPageEditorPart

praveen-skp:
Issue: #2223 

This pull request introduces a new feature in Eclipse that allows users to configure the alignment of tabs in multi-page editors. The tabs, which are currently fixed at the bottom, can now be dynamically positioned on the top or bottom, based on user preference. 

Key Changes:
- Added a new preference for multi-page editor tab alignment.
- Added a preference change listener to MultiPageEditorPart.
- Updated tab style based on the user's preference.

New preference: 
New preference selected

Tabs on top of the screen:
Expected look

PR #2476, opened 2024-11-01
Correct bundle version after changes in MultiPageEditorPart

HeikoKlare:
Recent API changes to MultiPageEditorPart and IWorkbenchPreferenceConstants have been introduced after 3.133 has already been released. Thus, the micro version bump made since then is insufficient. This change corrects the version but and the according since tags to 3.134.

Fixes API errors introduced by https://github.com/eclipse-platform/eclipse.platform.ui/pull/2224 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/2461:
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/IWorkbenchPreferenceConstants.java
@@ -634,7 +634,7 @@ public interface IWorkbenchPreferenceConstants {
 	 * The default value for this preference is: {@link SWT#BOTTOM}
 	 * </p>
 	 *
-	 * @since 3.133
+	 * @since 3.134
 	 */
 	String ALIGN_MULTI_PAGE_EDITOR_TABS = "ALIGN_MULTI_PAGE_EDITOR_TABS"; //$NON-NLS-1$
 
@@ -726,7 +726,7 @@ public interface IWorkbenchPreferenceConstants {
 	 * runtime. It only effects Windows.
 	 * </p>
 	 *
-	 * @since 3.133
+	 * @since 3.134
 	 */
 	String RESCALING_AT_RUNTIME = "RESCALING_AT_RUNTIME"; //$NON-NLS-1$
 }
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/part/MultiPageEditorPart.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/part/MultiPageEditorPart.java
@@ -172,7 +172,7 @@ public abstract class MultiPageEditorPart extends EditorPart implements IPageCha
 	 *
 	 * @param event the {@link PropertyChangeEvent} triggered by a change in the
 	 *              preference store
-	 * @since 3.133
+	 * @since 3.134
 	 */
 	protected boolean isUpdateRequired(PropertyChangeEvent event) {
 		if (event.getProperty().equals(IWorkbenchPreferenceConstants.ALIGN_MULTI_PAGE_EDITOR_TABS)) {
@@ -327,14 +327,14 @@ public abstract class MultiPageEditorPart extends EditorPart implements IPageCha
 	 * @return {@code SWT.TOP} if the user prefers tabs to be aligned on top,
 	 *         {@code SWT.BOTTOM} if the user prefers tabs to be aligned on the
 	 *         bottom.
-	 * @since 3.133
+	 * @since 3.134
 	 */
 	protected int getTabStyle() {
 		return getAPIPreferenceStore().getInt(IWorkbenchPreferenceConstants.ALIGN_MULTI_PAGE_EDITOR_TABS);
 	}
 
 	/**
-	 * @since 3.133
+	 * @since 3.134
 	 */
 	protected IPreferenceStore getAPIPreferenceStore() {
 		return PrefUtil.getAPIPreferenceStore();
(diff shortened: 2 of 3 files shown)


HeikoKlare:
As another confirmation that this is the right thing to do: the missing version bump was event reported in the Jenkins builds of #2224, see, e.g., https://github.com/eclipse-platform/eclipse.platform.ui/runs/32320535119
```
[API ERROR] File MANIFEST.MF at line 5: The minor version should be incremented in version 3.133.100, since new APIs have been added since version 3.133.0 (location: /home/jenkins/agent/workspace/eclipse.platform.ui_PR-2224/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF)
```

HannesWell:
> As another confirmation that this is the right thing to do: the missing version bump was event reported in the Jenkins builds of #2224, see, e.g., https://github.com/eclipse-platform/eclipse.platform.ui/runs/32320535119

This is also visble in the log of the latest `master` build:
https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/master/747/consoleFull

```
[INFO] --- tycho-apitools:4.0.8:verify (verify) @ org.eclipse.ui.workbench ---
[INFO] Resolve API baseline for org.eclipse.platform:org.eclipse.ui.workbench:eclipse-plugin:3.133.100-SNAPSHOT with linux/gtk/x86_64
[INFO] API Analysis finished in 42 s.
[ERROR] [API ERROR] File MANIFEST.MF at line 5: The minor version should be incremented in version 3.133.100, since new APIs have been added since version 3.133.0 (location: /home/jenkins/agent/workspace/eclipse.platform.ui_master/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF)
[WARNING] 1 API problems can't be mapped to the compiler log!
```
I really wonder why this doesn't fail the build? @laeubi any idea? Since the error is detected, I would first assume the problem in the integration with Tycho, not in the API-tools itself.

laeubi:
Both the maven and compiler checks have shown additional issues here:

grafik

so at best reviewers would check before merge that we do not add new issues, the main build even shows one more

grafik

Other than that If I remember correctly we currently use an older version of API tools, that suffers from a bug where the `failOnError` is not accounted in all cases see [here](https://github.com/eclipse-tycho/tycho/commit/846eb0bda2c71de77bde1dfaf9fd5c19ffdcc4bc)

iloveeclipse:
> have shown additional issues here:

But why they are both "green" ? Shouldn't they have red crosses?

iloveeclipse:
> Other than that If I remember correctly we currently use an older version of API tools, that suffers from a bug where the `failOnError` is not accounted in all cases see [here](https://github.com/eclipse-tycho/tycho/commit/846eb0bda2c71de77bde1dfaf9fd5c19ffdcc4bc)

 @laeubi : Is there a new version available? If yes, could you please update aggregator to use it?

laeubi:
> But why they are both "green" ? Shouldn't they have red crosses?

Because platform.ui has chosen to make them green:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/c1c681ada78722481d2b01a0790bf87f966975e4/Jenkinsfile#L33

>  Is there a new version available? If yes, could you please update aggregator to use it?

There was some issues with deadlocks but no one cared enough to analyses the root cause so an older API-Tools version was used in the meanwhile. I have now created a PR to configure the changed behavior now but it is not yet released:
- https://github.com/eclipse-tycho/tycho/pull/4295

iloveeclipse:
> Because platform.ui has chosen to make them green:
> 
> [eclipse.platform.ui/Jenkinsfile](https://github.com/eclipse-platform/eclipse.platform.ui/blob/c1c681ada78722481d2b01a0790bf87f966975e4/Jenkinsfile#L33)

@jukzi: I'm not sure it was meant to "mute" API checks: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1739

We should revert this to make API errors "visible" again.

BeckerWdf:
> As another confirmation that this is the right thing to do: the missing version bump was event reported in the Jenkins builds of #2224, see, e.g., https://github.com/eclipse-platform/eclipse.platform.ui/runs/32320535119
> 
> ```
> [API ERROR] File MANIFEST.MF at line 5: The minor version should be incremented in version 3.133.100, since new APIs have been added since version 3.133.0 (location: /home/jenkins/agent/workspace/eclipse.platform.ui_PR-2224/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF)
> ```

Thanks at @HeikoKlare for taking care. I simply overlooked this.

iloveeclipse (approved):

jukzi:
> We should revert this to make API errors "visible" again.

https://github.com/eclipse-platform/eclipse.platform.ui/pull/1739/files did not mute warnings. see https://github.com/eclipse-platform/eclipse.platform.ui/pull/1739#issuecomment-1984484691, also proofen by the other PRs failing by the warnings.
In the particular case of https://github.com/eclipse-platform/eclipse.platform.ui/runs/32320534400 the "problem" is that same number of warnings have been identified fixed as solved:
image

iloveeclipse:
> > We should revert this to make API errors "visible" again.
> 
> https://github.com/eclipse-platform/eclipse.platform.ui/pull/1739/files did not mute warnings. see [#1739 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1739#issuecomment-1984484691), also proofen by the other PRs failing by the warnings. In the particular case of https://github.com/eclipse-platform/eclipse.platform.ui/runs/32320534400 the "problem" is that same number of warnings have been identified fixed as solved: image

Please comment on https://github.com/eclipse-platform/eclipse.platform.ui/issues/2486.

I honestly don't know what exactly causes github actions be green despite finding new issues, but @laeubi seems to know that.

Issue #2453, opened 2024-10-28
[test] NPE in BarContentAssistProcessor


jukzi:
logged when locally running org.eclipse.ui.genericeditor.tests.CompletionTest
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jface.text.IDocument.getLength()" because the return value of "org.eclipse.jface.text.ITextViewer.getDocument()" is null
	at org.eclipse.ui.genericeditor.tests.contributions.BarContentAssistProcessor.computeCompletionProposals(BarContentAssistProcessor.java:45)
	at org.eclipse.ui.genericeditor.tests.contributions.LongRunningBarContentAssistProcessor.computeCompletionProposals(LongRunningBarContentAssistProcessor.java:36)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$10(AsyncCompletionProposalPopup.java:378)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$9(AsyncCompletionProposalPopup.java:377)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)

```

nettozahler:
This one seems to be trivial to me and so is the bugfix. But maybe I have overlooked sth.

PR #2484, opened 2024-11-03
Bugfix for #2453: NPE in BarContentAssistProcessor.

nettozahler:
This bugfix simply adds a check for NULL which was missing. In case of NULL a default value is returned. Fxes #2453


mickaelistria:
While null-check are always welcome, I'm more curious about in which context can this issue be happening? The document isn't supposed to be null at that stage, so maybe the issue is earlier in the execution chain?

nettozahler:
Since the affected class belongs to test code and the NPE happens when locally running `org.eclipse.ui.genericeditor.tests.CompletionTest` I suggest not to spend more time for further investigations until this bug occurs in productive code. Deeper investigations would be out of my personal scope anyway.

Issue #2486, opened 2024-11-04
API errors not failing GH validation


iloveeclipse:
> But why they are both "green" ? Shouldn't they have red crosses?

Because platform.ui has chosen to make them green:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/c1c681ada78722481d2b01a0790bf87f966975e4/Jenkinsfile#L33

>  Is there a new version available? If yes, could you please update aggregator to use it?

There was some issues with deadlocks but no one cared enough to analyses the root cause so an older API-Tools version was used in the meanwhile. I have now created a PR to configure the changed behavior now but it is not yet released:
- https://github.com/eclipse-tycho/tycho/pull/4295

_Originally posted by @laeubi in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2476#issuecomment-2453878664_

iloveeclipse:
@laeubi : I'm not sure which part of the mentioned line is responsible for showing "green" vs "red" Github checks shown on GH UI.
Is this `publishAllIssues:false` or `ignoreQualityGate:true` or both or something else?

laeubi:
> Is this `publishAllIssues:false` or `ignoreQualityGate:true` or both or something else?

- `publishAllIssues` = should all (including existing) be shown in Github as code annotations
- `ignoreQualityGate` = If true, then the result of the quality gate is ignored when selecting a reference build. This option is disabled by default, so a failing quality gate will be passed from build to build until the original reason for the failure has been resolved.

iloveeclipse:
@laeubi : thanks, but sorry, I still don't understand which setting causes the failed API check to be shown as green instead as red? Could you please simply fix whatever you think is broken and push a PR that would "enable" github validation to fail on failed API check?

laeubi:
I don't know what is the desired outcome, when I started with this I let the check fail / become "red" whenever something new was detected, then several iterations / changes where performed e.g. now we use "DELTA" what mean if you solve one problem you are allowed to add one new without the checks to become red.

As long as warnings / reported problems are only seen as something that gets into peoples way to merge there PR as soon as possible instead of something we need to fix/improve I honestly don't see any "right" settings.

This is for example how PDE uses the checks what seem to work (for the PDE project) since a while:
https://github.com/eclipse-pde/eclipse.pde/blob/f3ba0239012344e49638f28431ff4340c1eeeb47/Jenkinsfile#L42

iloveeclipse:
[...]

PR #2488, opened 2024-11-04
added API Error example

jukzi:
DO NOT MERGE

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2486


jukzi:
@HannesWell the error message of "Pull-Request Checks" is not understandable in this case: `bundleLocation not found`

jukzi:
Error should have been: org.eclipse.ui.browser/META-INF `The minor version should be incremented in version 3.8.400, since new APIs have been added since version 3.8.300`

laeubi:
> error should have been: org.eclipse.ui.browser/META-INF `The minor version should be incremented in version 3.8.400, since new APIs have been added since version 3.8.300`

You have a missing since tag:
grafik

laeubi:
And error is shown here

grafik

PR #2490, opened 2024-11-04
added API Error example - mark maven unstable

jukzi:
DO NOT MERGE

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2486


jukzi:
i think we did not setup a qualitygate for mavenConsole since we had too many false positives due to flakyness

PR #2494, opened 2024-11-05
Notication rename according to numberschema

vogella:


vogella:
Mac failure is known https://github.com/eclipse-platform/eclipse.platform.ui/issues/370

vogella:
Now fails with the known and unrelated issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #2495, opened 2024-11-05
Notification example user interaction

vogella:


vogella:
Mac failure which will most likely occur is known https://github.com/eclipse-platform/eclipse.platform.ui/issues/370

PR #2504, opened 2024-11-08
Snippet to reproduce exception when calling getExpandedTreePaths()

ptziegler:
This snippet shows how to cause an AssertionFailedException when calling getExpandedTreePaths() on a virtual tree.

When clicking the "refresh" button for the first time, everything works fine.

On the second time, the following error is logged:

> !MESSAGE Invalid index value 2 >= 2

On the third time, an AssertionFailedException is thrown.

See:
https://github.com/eclipse-platform/eclipse.platform.ui/discussions/2491


ptziegler:
What seems to be important is that the parts of the tree must be outside the visible area. Meaning the exception isn't thrown when the shell is made larger.

I think what's happening is that the refresh() call only refreshes the visible tree items, leaving the viewer in an inconsistent state.

PR #2510, opened 2024-11-08
org.eclipse.search.core requires org.eclipse.core.resources 3.21.100

### LINKED ISSUE #1886: update m2e corrupts search
SCENARIO:
Eclipse 2024-06 for Java EE
Help -> About... ->  Installation details -> select M2E - Maven integration for Eclipse (2.6.1....), press Update...

Go through the update procedure (it ends up with   M2E - Maven Integration for Eclipse	2.6.2.20240828-1954).

After Eclipse restart  try to use Search in Eclipse

EXPECTED:
Search works

ACTUALLY:
```
Text Search
An internal error occurred during: "File Search Worker".
'byte[] org.eclipse.core.resources.IFile.readNBytes(int)'
```

The issue is similar to the one reported for another plug-in https://github.com/Genuitec/Copilot4Eclipse-community/discussions/23

Maybe it's related to Eclipse update itself and not to specific plugins?

howlger:
`org.eclipse.search.internal.core.text.FileCharSequenceProvider.toShortString(...)` requires the method `byte[] org.eclipse.core.resources.IFile.readNBytes(int)` which was introduced in org.eclipse.core.resources 3.21.100.

The increasing of the lower bound of that requirement was missed in https://github.com/eclipse-platform/eclipse.platform/commit/906371db294e0d090180248d673d0c84f7b2d2b9.

Fixes https://github.com/eclipse-m2e/m2e-core/issues/1886

### DIFF
--- a/bundles/org.eclipse.search.core/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.search.core/META-INF/MANIFEST.MF
@@ -12,7 +12,7 @@ Export-Package: org.eclipse.search.core.text,
  org.eclipse.search.internal.core.text;x-friends:="org.eclipse.search,org.eclipse.search.tests"
 Require-Bundle: 
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
- org.eclipse.core.resources;bundle-version="[3.14.0,4.0.0)",
+ org.eclipse.core.resources;bundle-version="[3.21.0,4.0.0)",
  org.eclipse.core.filesystem;bundle-version="[1.3.0,2.0.0)",
  org.eclipse.core.filebuffers;bundle-version="[3.5.0,4.0.0)",
  org.eclipse.ltk.core.refactoring;bundle-version="[3.5.0,4.0.0)",


bundles/org.eclipse.search.core/META-INF/MANIFEST.MF
```diff
@@ -12,7 +12,7 @@ Export-Package: org.eclipse.search.core.text,
  org.eclipse.search.internal.core.text;x-friends:="org.eclipse.search,org.eclipse.search.tests"
 Require-Bundle: 
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
- org.eclipse.core.resources;bundle-version="[3.14.0,4.0.0)",
+ org.eclipse.core.resources;bundle-version="[3.21.100,4.0.0)",
```
HannesWell: Specifying the service/micro version is usually not necessary because new API requires a new minor version.
```suggestion
 org.eclipse.core.resources;bundle-version="[3.21.0,4.0.0)",
```
Alternatively you could also simply require the current version `3.22`, as the bundles of the SDK should usually be of the same release.

HannesWell (changes requested):
Thank you for this PR. I just have a small request for change.

bundles/org.eclipse.search.core/META-INF/MANIFEST.MF
howlger: I see.

The [new method was added](https://github.com/eclipse-platform/eclipse.platform/commit/906371db294e0d090180248d673d0c84f7b2d2b9#diff-e43d5a9d9d26ee0b9f446aafe3c30449eaec98801cc41697d94aba0aaed8e164) two weeks after [the version of `org.eclipse.core.resources` has been increased to 3.21.0](https://github.com/eclipse-platform/eclipse.platform/commit/2d9238e189d52ad6083f6ac06aafd6275026648a#diff-2919459a3a8abd79ccc1c39389b1263eecd708e84b0e134bf41cde31d70be75c), but both changes have been between the [4.32 release](https://github.com/eclipse-platform/eclipse.platform/commits/R4_32) and [4.33M1](https://github.com/eclipse-platform/eclipse.platform/commits/S4_33_0_M1). Does this mean that `3.21` is sufficient, or should `3.22` be used, since there was no release in between, but a few 3.21 builds not containing the new required method?

What would have been the right thing to do? Bump the minor version in the commit that adds the new API? Or will the minor version be bumped before a new release to avoid multiple minor version bumps between releases?

HannesWell: > Does this mean that `3.21` is sufficient, or should `3.22` be used, since there was no release in between, but a few 3.21 builds not containing the new required method?

Yes, 3.21 is sufficient. The few builds are probably not available anymore since only the release repository is retained forever after the release. The Integration-build repositories built every night are usually deleted after the release.



> What would have been the right thing to do? Bump the minor version in the commit that adds the new API? Or will the minor version be bumped before a new release to avoid multiple minor version bumps between releases?

How it was done was done in `o.e.core.resources` is the right way. Since we don't want to skip versions, we only apply each version bump 'type' once in a release and only 'widen' the type.
I.e. if the first change in a release cycle is fully internal we bump only the service version, if then later in a cycle new API is added, we increment the minor version too. If an API addition is the first change, then the minor version is bumped immediately and subsequent changes, may be API additions or just internal changes, don't require a version bump anymore.

What was forgotten in this case was to increase the required version when the new method was used for the first time in this bundle.

howlger: Thanks for the explanation.

HannesWell (approved):
Thanks Holger for this fix.

In general please amend the existing commit of a PR and force-push it to your PR branch.
There is no need to capture the review process in the git-history and Github's squash-and-merge does not lead to satisfying results. I have done that for you know so that this can be submitted now.

HannesWell:
Test failure on Jenkins is unrelated.

Issue #2509, opened 2024-11-08
Paste into find overlay control does not work in multipage editor (manifest.mf, plugin.xml)


tobiasmelcher:
- [X] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/) - I downloaded eclipse-SDK-I20241107-1800-win32-x86_64

## Steps to reproduce

From a fresh installation and clean workspace:

* import an eclipse plugin project
* open a MANIFEST.MF file
* select the word "qualifier" in the text editor and press CTRL-C to copy the text into the clipboard
* select another word in the text editor, e.g. "plugin"
* press CTRL-F to open the new find overlay control shown inplace of the text editor. The word "plugin" is taken over in the find overlay which is correct. I expect that the find overlay control now has the focus.
* now press CTRL-V to paste "qualifier" into the find overly search control -> this is currently not the case, the word "qualifier" from the clipboard is pasted into the text editor and not into the find overlay

The same steps work fine in the Java source editor, but not in a multipage editor containing a text editor as page.
Screencast:

https://github.com/user-attachments/assets/3613e358-09ee-4002-b89e-1074be1a9ff0

## Tested under this environment:
* OS & version: Windows
* Eclipse IDE/Platform version (as shown in *Help > About*): ...
* eclipse-SDK-I20241107-1800-win32-x86_64

## Community

- [X] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

HeikoKlare:
Additional information: this also happens in other editors (like Java source code editor) when pasting into the replace input field.

HeikoKlare:
Fixed via:
- #2516

I used a wrong tag, so that this issue was not auto-closed by merging that PR.

tobiasmelcher:
Thanks a lot. Fix is also working fine in the Abap Development Tools source editors.

PR #2513, opened 2024-11-11
Find/replace overlay: allow pasting into replace input field #2509

HeikoKlare:
While actions of the target editor are properly deactivated when the find input field of the FindReplaceOverlay has focus, the same does not happen for the replace input field. In consequence, for example, you cannot paste clipboard content into the replace input field via the according keyboard shortcut (CTRL+V).

With this change, the functionality to deactivate target editor actions is also applied when the the replace input field has focus, in addition to the find input field.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/2509

### Demo
Behavior before fix:
findreplace_insert_replace_before

Behavior after fix:
findreplace_insert_replace_after

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -147,6 +147,37 @@ public class FindReplaceOverlay {
 	private ControlDecoration searchBarDecoration;
 	private ContentAssistCommandAdapter contentAssistSearchField, contentAssistReplaceField;
 
+	private FocusListener targetActionActivationHandling = new FocusListener() {
+		@Override
+		public void focusGained(FocusEvent e) {
+			setTextEditorActionsActivated(false);
+		}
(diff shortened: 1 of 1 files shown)


BeckerWdf:
> Contributes to #2509

This reported issue from #2509 still exists with this PR.

HeikoKlare:
> > Contributes to #2509
> 
> This reported issue from #2509 still exists with this PR.

This PR does not completely fix #2509. It just fixes pasting into the replace field for non-multipage-editors (for which I added an additional comment to #2509). Probably I should have opened a separate issue for that, but I first though they are highly related, as they are all about pasting into the overlay (and using other shortcuts).

HeikoKlare:
In my opinion, this should be fixed for the upcoming release, as it is quite annoying to not be able to use common shortcuts to cut, copy, paste etc. in the replace input field of the find/replace overlay.

@akurtakov @vogella may I ask for the PMC approval of one of you (or any other PMC member) to still merge this?

akurtakov:
I haven't reviewed the actual code change nor tested it but the issue looks important enough to worth PMC +1 from me for RC1.

akurtakov:
Code change looks good. Please push when verifications succeed.

HeikoKlare:
Thank you!

PR #2515, opened 2024-11-12
Find/replace overlay: improve focus/tab order #2161

### LINKED ISSUE #2161: Find/replace overlay: improve switching between find and replace input fields

## Suggestion

* open a text editor
* open the find/replace overlay via CTRL+F / CMD+F
* open the replace bar (if not yet open) via CTRL+R / CMD+R
* try to navigate from find input field to replace input field via pressing TAB

You have to tab through all the search options before reaching the replace input field.

I would like to be able to directly navigate between find and replace input fields via a single keyboard interaction (a single TAB press), as this is (in my opinion) the most important navigation scenario.

I see options how to improve this:
1. Change the tab order by having replace input field right after search input field before navigating through the option buttons.
2. Use a different way of keyboard interaction to navigate through the options and the input fields (such as tab for the input fields and Alt+Tab for the option or the like).

## Acknowledgements

This proposal is based on a finding by Andreia Schoknecht and Lukas Kolender in a test session performed at SAP. Thank you for testing and reporting!

## Community

HeikoKlare:
The FindReplaceOverlay currently uses the default focus order. When tabbing through the widgets, this means that when starting from the find input field you have to tab over all the search options before coming to the replace input field. Since the most often used navigation is between the two input fields for find and replace, they should come after each other in the focus order.

This change adapts the focus order of the involved elements:
- Tab between find and replace input field
- Tab between replace input field and the tools (starting from the search tools)
- Tab between search/close and replace search tools (instead of coming from search/close tools to the replace input field)

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2161

### Proposed Tab Order
image

### Behavior
Existing tab order:
findreplace_taborder_existing

Proposed tab order:
findreplace_taborder_new

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/AccessibleToolBar.java
@@ -74,4 +74,9 @@ class AccessibleToolBar extends Composite {
 		}
 	}
 
+	Control getFirstControl() {
+		Control[] children = getChildren();
+		return children.length == 0 ? null : getChildren()[0];
+	}
+
 }
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -38,6 +38,7 @@ import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.ScrollBar;
 import org.eclipse.swt.widgets.Scrollable;
 import org.eclipse.swt.widgets.Shell;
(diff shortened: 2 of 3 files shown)


HeikoKlare:
@thomasritter do you agree that this proposed order/behavior is as expected?
Unfortunately, I have been thinking about a "better" solution for too long, so I am (too) late for the upcoming release. But I consider it a great improvement with respect to usability of the find/replace UI, so maybe we could still bring it into RC1.

Wittmaxi:
This is the hack I was hoping we wouldn't need - yet, I fear it is the only logical solution to what we want (and I agree that we want this): a smooth transition from search to replace bar.

Is there a (good) way to wrap this functionality into "Accessible Toolbar"? Maybe we could add a method to the toolbars (or to the toolbar builder) allowing them to automatically do the dirty wiring so that we just need to tell toolbars A and B that "this button[A] ->  that button[b]"

HeikoKlare:
> This is the hack I was hoping we wouldn't need - yet, I fear it is the only logical solution to what we want (and I agree that we want this): a smooth transition from search to replace bar.

I agree that the solution is not that nice, but I would not call it a hack either. I do not see any option than a custom handling of tab events, as a custom definition leading to an automated handling of tab event is not allowed across composite border, at least on Windows (if I am not mistaken).

> Is there a (good) way to wrap this functionality into "Accessible Toolbar"? Maybe we could add a method to the toolbars (or to the toolbar builder) allowing them to automatically do the dirty wiring so that we just need to tell toolbars A and B that "this button[A] -> that button[b]"

We could wrap the establishment of each wiring into a method, but without quite some additional effort that would increase the number of listeners involved. Wrapping this into the accessible toolbar is not possible, because there are other elements than accessible toolbar involved (like the text field).

Wittmaxi:
> We could wrap the establishment of each wiring into a method, but without quite some additional effort that would increase the number of listeners involved. Wrapping this into the accessible toolbar is not possible, because there are other elements than accessible toolbar involved (like the text field).

I don't believe it would be worth it. One would expect that there is an inbuilt way to specify tab orders which do not respect the object-hierarchy (hence why I called this a hack when in reality you are right: this is a canonical solution)

HeikoKlare:
Test failure is unrelated and already documented:
- #370

thomasritter:
Sorry missed the mention. Thanks for fixing this @HeikoKlare!

Issue #2021, opened 2024-07-01
Umbrella Issue: find/replace overlay issues


Wittmaxi:
This Issue documents my views on priority for the issues of the find/replace overlay.
High issues are issues which are blocking a possible release

### high
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2033
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2011
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2007
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2006
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1994
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1992
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1989
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1983
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1991
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1951
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1945
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/1907
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2053
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2046
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2054
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2059
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2075
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2138
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2147
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2244
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2246
- [x] https://github.com/eclipse-platform/eclipse.platform.swt/issues/1447
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2285
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2322
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2161
[...]

PR #2516, opened 2024-11-12
Find/replace overlay: allow pasting in multi-page editors #2509

HeikoKlare:
While actions of the target editor are properly deactivated when the target is an ordinary, single-page editor, the same does not happen when the target is a multi-page editor. In consequence, for example, you cannot paste clipboard content into the overlay via the according keyboard shortcut (CTRL+V), but it will paste into the target editor instead.

With this change, the functionality to deactivate target editor actions is extended to also deactivate the relevant global actions (cut, copy, paste, etc.) that editors like multi-page editors fall back to.

Fixes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/2509

### Note
I hope there is a better solution to deactivate the according action handlers in a multi-page editor than this, but I did not find it so far. Still, I think this is better than not being able to use common shortcuts for cut, copy, paste etc. when using multi-page editors.

### Behavior
Without the fix:
findreplace_multipage_before

With the fix:
findreplace_multipage_after

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -13,9 +13,10 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.findandreplace.overlay;
 
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.osgi.framework.FrameworkUtil;
@@ -46,6 +47,7 @@ import org.eclipse.swt.widgets.Widget;
 
 import org.eclipse.core.runtime.Status;
 
+import org.eclipse.jface.action.IAction;
 import org.eclipse.jface.bindings.keys.KeyStroke;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.IDialogSettings;
@@ -60,6 +62,7 @@ import org.eclipse.jface.text.FindReplaceDocumentAdapterContentProposalProvider;
 import org.eclipse.jface.text.IFindReplaceTarget;
 import org.eclipse.jface.text.ITextViewer;
 
+import org.eclipse.ui.IActionBars;
 import org.eclipse.ui.IWorkbenchPart;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.fieldassist.ContentAssistCommandAdapter;
@@ -70,10 +73,12 @@ import org.eclipse.ui.internal.findandreplace.HistoryStore;
 import org.eclipse.ui.internal.findandreplace.SearchOptions;
 import org.eclipse.ui.internal.findandreplace.status.IFindReplaceStatus;
 import org.eclipse.ui.internal.texteditor.TextEditorPlugin;
+import org.eclipse.ui.part.MultiPageEditorSite;
 
 import org.eclipse.ui.texteditor.AbstractTextEditor;
 import org.eclipse.ui.texteditor.FindReplaceAction;
 import org.eclipse.ui.texteditor.IAbstractTextEditorHelpContextIds;
+import org.eclipse.ui.texteditor.ITextEditorActionConstants;
 import org.eclipse.ui.texteditor.ITextEditorActionDefinitionIds;
 import org.eclipse.ui.texteditor.StatusTextEditor;
 
@@ -148,6 +153,8 @@ public class FindReplaceOverlay {
 	private ContentAssistCommandAdapter contentAssistSearchField, contentAssistReplaceField;
 
 	private FocusListener targetActionActivationHandling = new FocusListener() {
+		private DeactivateGlobalActionHandlers globalActionHandlerDeaction;
+
 		@Override
 		public void focusGained(FocusEvent e) {
 			setTextEditorActionsActivated(false);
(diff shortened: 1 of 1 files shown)


HeikoKlare:
@HannesWell what do you think?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -167,15 +175,48 @@ private void setTextEditorActionsActivated(boolean state) {
...
+					globalActionHandlerDeaction = null;
+				}
+			}
 			try {
 				Method method = AbstractTextEditor.class.getDeclaredMethod("setActionActivation", boolean.class); //$NON-NLS-1$
 				method.setAccessible(true);
 				method.invoke(targetPart, Boolean.valueOf(state));
-			} catch (IllegalArgumentException | InvocationTargetException | IllegalAccessException | SecurityException | NoSuchMethodException ex) {
+			} catch (IllegalArgumentException | InvocationTargetException | IllegalAccessException | SecurityException
+					| NoSuchMethodException ex) {
```
HannesWell: If suitable, you could combine all reflective operations being catched.
```suggestion
			} catch (IllegalArgumentException | ReflectiveOperationException ex) {
```

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
```diff
@@ -167,15 +175,48 @@ private void setTextEditorActionsActivated(boolean state) {
...
 				TextEditorPlugin.getDefault().getLog()
 						.log(Status.error("cannot (de-)activate actions for text editor", ex)); //$NON-NLS-1$
 			}
 		}
+
+		static final class DeactivateGlobalActionHandlers {
+			private final static String[] ACTIONS = { ITextEditorActionConstants.CUT, ITextEditorActionConstants.COPY,
+					ITextEditorActionConstants.PASTE, ITextEditorActionConstants.DELETE,
+					ITextEditorActionConstants.SELECT_ALL, ITextEditorActionConstants.FIND };
```
HannesWell: Maybe make this more modern?
```suggestion
			private final static List<String> ACTIONS = List.of(ITextEditorActionConstants.CUT,
					ITextEditorActionConstants.COPY, ITextEditorActionConstants.PASTE,
					ITextEditorActionConstants.DELETE, ITextEditorActionConstants.SELECT_ALL,
					ITextEditorActionConstants.FIND);
```

HannesWell (approved):
Thank you for this fix! I really appreciate that.

I tested it and it works as desired, but I cannot tell if there is a generally better way to implement this.

Since I think this is an important fix, I approve this for submission for RC1 as PMC member:
PMC +1

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: Agreed. Replaced in 5c7a21c

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
HeikoKlare: That's a good idea. I had just copied this over and adapted it, but of couse makes sense to improve it as well. Done with 5c7a21c

HeikoKlare:
Thank you for the feedback and the approval, Hannes!

By accident, I had to search through several Manifests right after proposing this PR and experienced how annoying the issue is. Seems like I had no need to search in Manifests for quite a while. So now I definitely agree that this should have high priority and that it should go into the upcoming release.

HannesWell:
> So now I definitely agree that this should have high priority and that it should go into the upcoming release.

Yes, absolutely 😅 
Please submit this as soon as the build has completed.

HeikoKlare:
Failing test is unrelated: #370

chrisrueger:
@HeikoKlare I have this problem in Bndtools / BndEditor with **Eclipse 2024-12** which under the hood is a `FormEditor` which is a `MultiPageEditorPart`.

https://github.com/bndtools/bnd/blob/9ed91bd967559c0bf963753990e96719bf6ea4cb/bndtools.core/src/bndtools/editor/BndEditor.java#L104

It happens in the **Source tab** (which is a org.eclipse.ui.forms.editor.FormPage extends EditorPart implements IFormPage https://github.com/bndtools/bnd/blob/9ed91bd967559c0bf963753990e96719bf6ea4cb/bndtools.core/src/bndtools/editor/pages/BundleContentPage.java#L39)


`public class BndEditor extends ExtendedFormEditor implements IResourceChangeListener, IBndEditor {`

`public abstract class ExtendedFormEditor extends FormEditor {`

`public abstract class FormEditor extends MultiPageEditorPart  {`


Is this a different / new case which needs to be handled? (and maybe added to the umbrello issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/2021)

Or is there something Bndtools should do, to fix it?

merks:
@chrisrueger 

I fixed some related problems during the current release cycle because I noticed the PDE target editor was working poorly.  

I would suggest testing with 2025-03 / 4.35, though it's too late now to fix anything for that release.  Any fixes will have to happen for 2025-06 / 4.36...

chrisrueger:
> I would suggest testing with 2025-03 / 4.35,

I just tested with 2025-03 / 4.35 RC1 and I confirm the problem in Bndtools Editor is fixed 👍  
Great work  @merks

PR #2518, opened 2024-11-14
[Windows] Add warning about activating "Monitor-specific UI scaling" in preferences dialog

fedejeanne:
### [**Windows only!**]

Since the flickering issue (https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122) is still open and activating this option also sets **Edge** as the browser of choice, the user should be warned.

This is the new warning:

image

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
@@ -20,6 +20,7 @@
 package org.eclipse.ui.internal.dialogs;
 
 import static org.eclipse.jface.viewers.LabelProvider.createTextProvider;
+import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
 import static org.eclipse.ui.internal.registry.IWorkbenchRegistryConstants.ATT_COLOR_AND_FONT_ID;
 import static org.eclipse.ui.internal.registry.IWorkbenchRegistryConstants.ATT_OS_VERSION;
 import static org.eclipse.ui.internal.registry.IWorkbenchRegistryConstants.ATT_THEME_ASSOCIATION;
@@ -71,12 +72,14 @@ import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.program.Program;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Group;
 import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Link;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPreferenceConstants;
 import org.eclipse.ui.IWorkbenchPreferencePage;
@@ -213,9 +216,12 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 		group.setFont(parent.getFont());
 		GridLayout layout = new GridLayout(1, false);
 		group.setLayout(layout);
-		Label infoLabel = new Label(group, SWT.WRAP);
-		infoLabel.setText(WorkbenchMessages.RescaleAtRuntimeDisclaimer);
+		Link infoLabel = new Link(group, SWT.WRAP);
+		infoLabel.setText(WorkbenchMessages.RescaleAtRuntimeDisclaimer + System.lineSeparator() + System.lineSeparator()
+				+ WorkbenchMessages.EdgeBrowserDisclaimer);
 		infoLabel.setLayoutData(GridDataFactory.defaultsFor(infoLabel).create());
+		infoLabel.addSelectionListener(widgetSelectedAdapter(c -> Program.launch(c.text)));
+
 		createLabel(group, ""); //$NON-NLS-1$
 
 		boolean initialStateRescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchMessages.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchMessages.java
@@ -76,6 +76,8 @@ public class WorkbenchMessages extends NLS {
 
 	public static String StatusUtil_errorOccurred;
 
+	public static String EdgeBrowserDisclaimer;
+
 	// ==============================================================================
 	// Workbench Actions
 	// ==============================================================================
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
@@ -505,6 +505,8 @@ RescaleAtRuntimeSettingChangeWarningText = Restart for the DPI setting changes t
 HiDpiSettingsGroupTitle = HiDPI settings
 RescaleAtRuntimeEnabled = Monitor-specific UI &scaling
 RescaleAtRuntimeDisclaimer = EXPERIMENTAL! Activating this option will dynamically scale all windows according to the monitor they are currently in. It will also set the default browser to Edge in order to provide the appropriate scaling of content displayed in a browser. This feature is still in development and therefore considered experimental.
+EdgeBrowserDisclaimer = Note that the usage of Edge may currently lead to flickering when resizing a browser window (see <a href="https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122">this issue</a> for more details).
+
 # --- Workbench -----
 WorkbenchPreference_openMode=Open mode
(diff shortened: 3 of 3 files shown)


fedejeanne:
@HeikoKlare  since you made the N&N about the new setting (https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/243), can you please propose a PR to adapt the screenshot? I have a different windows theme and it wouldn't look good in the N&N

laeubi:
Would a link to the N&N not being better than clutter the UI with more an more text ... the more text the more unlikely it getting actually read.

fedejeanne:
> Would a link to the N&N not being better than clutter the UI with more an more text ... the more text the more unlikely it getting actually read.

Who's to say someone would click on a link and wait for the browser to render something if he/she wouldn't even read some text that's right in front of them and starts with a big **_WARNING!_**.

laeubi:
> Who's to say someone would click on a link and wait for the browser to render something if he/she wouldn't even read some text that's right in front of them and starts with a big **_WARNING!_**.

There is no guarantee, but a short 

> Read the following carefully ...some link ...

is likely getting more attention than a long text... also such text can be enhanced *after* the release.

And honestly if it is **that dangerous** that we now need **two** big WARNING message (actually it is not bold in the screenshot ...), we better should not ship this feature to the user at all. It is already mentions to be experimental, so either user *want* this then they don't care about some glitches, or the don't but adding multiple `warnings`, `danger`, `don't click this` does not feel right to let user make a good decision. 

We have had some of such experimental switches already ere and there, and none of them have more than a few words.

HeikoKlare:
I have to admit that I am not an expert in when flickering might be problematic for people sensitive to it. But note that the issue is not an unavoidable, automated flickering. It only happens when resizing a browser window and it only happens as long as you do a resizing operation, so once you release the mouse button or stop moving the mouse, the flickering stops as well. And you experience the same with other embedded applications like Word, Excel etc. in an Eclipse IDE as well, but probably not so many people use those capabilities.

That said, I appreciate the awareness for this and we should definitely think about whether and where to place that information. I agree with @laeubi that we now have quite too much text (and particular, too many paragraphs) and too much "experimental" and "warning" in the box that may prevent people from using the feature. In my opinion, we could still integrate the information but just as an additional sentence to the existing description of the "experimental" paragraph. Adding a sentence like this at the end might be sufficient: "Note that the usage of Edge may currently lead to flickering when resizing a browser window."

fedejeanne:
Then how about less "EXPERIMENTAL" and more focus on the WARNING?

image

HeikoKlare:
> Then how about less "EXPERIMENTAL" and more focus on the WARNING?

The purpose of the additional text is to notify that the feature is experimental, i.e., we know that it is had glitches and is not ready to be enabled by default, but we consider it useful for testing and maybe even productive use if remaining glitches are individually acceptable. Now just adding the new warning text about the Edge glitch and removing every information about the feature itself and in total being experimental is not what the text is supposed to say.

fedejeanne:
Then how about:

image

?

Less clutter in the "EXPERIMENTAL" part and a link in the "WARNING"

fedejeanne:
BTW requesting PMC approval to merge this once the text is proper @merks / @akurtakov

HeikoKlare:
Could we please just have a simple solution for this?

> Less clutter in the "EXPERIMENTAL" part and a link in the "WARNING"

My concern remains:

> that we now have quite too much text (and particular, too many paragraphs) and too much "experimental" and "warning" in the box that may prevent people from using the feature

I propose to keep the existing paragraph (if possible, even shorten it) without any additional "WARNING" or the like and just add one simple sentence without references to issues, potential seizures etc.. When you mention that things may flicker, people sensitive to that will consider not using that functionality anyway without requiring a verbose explanation in the preferences.

fedejeanne:
> I propose to keep the existing paragraph (if possible, even shorten it) without any additional "WARNING" or the like and just add one simple sentence without references to issues, potential seizures etc.. When you mention that things may flicker, people sensitive to that will consider not using that functionality anyway without requiring a verbose explanation in the preferences.

I disagree: the word "seizures" should be in it. We must be open about the possible implications, regardless of how low the chances are that someone may get seizures from it.

Would you like to propose a final text? Please explicitly warn about possible seizures, something like "may cause seizures". Skip the "WARNING" if you want.

HeikoKlare:
> I disagree: the word "seizures" should be in it. We must be open about the possible implications, regardless of how low the chances are that someone may get seizures from it.

With this argumentation, we would need to have a disclaimer when starting/downloading Eclipse, because, as said, when opening other embedded applications inside Eclipse (like MS Office products), you have the same flickering and no one warns you about that. Don't get me wrong: it's good to discuss this and ensure that we provide a proper information about this. But as said, I consider a warning that there might be flickering enough for people who may be sensitive to it to decide whether they want to take the risk or not. I cannot even say whether this kind of flickering may cause seizures, so I would just make that statement about flickering without judging about potential consequences.

My proposal remains that I would add something simple to the existing paragraph like:  "Note that the usage of Edge may currently lead to flickering when resizing a browser window." (https://github.com/eclipse-platform/eclipse.platform.ui/pull/2518#issuecomment-2476308361)

Since I have not seen any reasonable arguments to further clutter the dialog with more verbose statements so far, I do not agree with any more complicate statement. If someone else has a different opinion, feel free to overrule me.

fedejeanne:
> My proposal remains that I would add something simple to the existing paragraph like: "Note that the usage of Edge may currently lead to flickering when resizing a browser window." (https://github.com/eclipse-platform/eclipse.platform.ui/pull/2518#issuecomment-2476308361)

A full text from you would have been better but I assume you meant (original text + 1 sentence):

image

... not my first choice but I can live with a slightly modified version of it:

image

It has your sentence, the word "seizures" (my preference) and a link for more details (Christoph's request).

I hope I hit the middle-ground and we can all accept this proposal, @HeikoKlare and @laeubi ?

fedejeanne:
I assume the current text is fine since there have been no objections/proposals in the past days.

@merks / @akurtakov requesting PMC approval to merge for RC2

HeikoKlare:
> I assume the current text is fine since there have been no objections/proposals in the past days.

As said in [my previous comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2518#issuecomment-2478182620), I object to the current proposal.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -213,16 +219,27 @@ private void createHiDPISettingsGroup(Composite parent) {
...
 		createLabel(group, ""); //$NON-NLS-1$
 
 		boolean initialStateRescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
 				.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
 		rescaleAtRuntime = createCheckButton(group, WorkbenchMessages.RescaleAtRuntimeEnabled, initialStateRescaleAtRuntime);
 	}
 
+	private void openURL(String url) {
+		try {
+			Desktop.getDesktop().browse(new URI(url));
```
laeubi:
```suggestion
			Program.launch(url);
```

merks:
I tend agree with Heiko.  Personally, I think that people who are susceptible to seizures induced by flickering will be well aware of that fact and that we don't need to scream this out to every reader...

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
fedejeanne: Added, thank you

fedejeanne:
Majority has spoken. I changed the text.

fedejeanne:
Failing test is unrelated: https://github.com/eclipse-platform/eclipse.platform.ui/issues/195

@merks may I merge this?

akurtakov:
So the edge warning will be shown on mac/linux too? It sounds not useful and overwhelming.

fedejeanne:
> So the edge warning will be shown on mac/linux too? It sounds not useful and overwhelming.

It remains a Windows-only thing, nothing changed in that sense:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/69f6e45a6a3cc74f553fd827c7fb86dbfe7ed7e9/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java#L202-L205

fedejeanne:
Today is the last stabilization day for RC2 and tomorrow is the signoff ([calendar](https://calendar.google.com/calendar/u/0/embed?src=prfk26fdmpru1mptlb06p0jh4s@group.calendar.google.com)).

Do you think it would be possible to merge today or do you see any blockers, @merks and @akurtakov ? Anything I should do?

akurtakov:
I am not a fan of such kind of messages in preferences but it all depends on the "severity" of the issue . 
As I don't have access to windows at all I don't feel qualified to have the say here and defer to the rest of the @eclipse-platform/eclipse-platform-project-leads and/or PMC to have a say here.

fedejeanne:
> I am not a fan of such kind of messages in preferences but it all depends on the "severity" of the issue . As I don't have access to windows at all I don't feel qualified to have the say here and defer to the rest of the @eclipse-platform/eclipse-platform-project-leads and/or PMC to have a say here.

Understood. Thank you for the feedback.

FTR I added another way to reproduce the flickering to the original issue (https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122): the _JavaDoc_ view flickers.

fedejeanne:
Rendered moot by https://github.com/eclipse-platform/eclipse.platform.swt/pull/1588

Issue #2534, opened 2024-11-25
Eclipse Hover loses control in certain cases


mehmet-karaman:
## Steps to reproduce

From a fresh installation and clean workspace:
* Create a new Java Project and open a java class in the java editor
* Move the mouse to a place where the hover shows up. (eg. Class head and hover on the class name) 
* Double click on the Hover
* Minimize the Eclipse window and show it back.
* Move any other applications window to the foreground
* The hover is still visible above the foreign windows 

## I expected: 
The Sticky hover shouldn't be visible (closed when focus is lost)

## But got: 
Sticky hover is still there

## Tested with Eclipse Version:
Version: 2024-12 (4.34)
Build id: I20241022-1800

## Here is a video recorded in windows:
https://github.com/user-attachments/assets/2eb51c05-caee-4256-8b9b-72c78492a991

iloveeclipse:
Reproducible with latest 4.34 RC2 on RHEL 9.2 too, so this is not Windows specific.

mehmet-karaman:
The PR Was merged, so we can close this bug ticket now.

PR: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2540

PR #2540, opened 2024-11-25
Bugfix for Hover which is visible above other application windows. #2534

mehmet-karaman:
This patch is fixing the problem, where the HoverManagers listeners are deregistered but the hover is still visibile on top. This happens when the hover was sticky and after that eclipse was minimized and put back on foreground after it. In that case the listeners are already deregistered and there is no chance to react at all. So for this case the visibility has to be set explicitly when the close is going to stop.

### DIFF
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.text
-Bundle-Version: 3.26.0.qualifier
+Bundle-Version: 3.26.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
@@ -332,8 +332,6 @@ public class BrowserInformationControl extends AbstractInformationControl implem
 	@Override
 	public void setVisible(boolean visible) {
 		Shell shell= getShell();
-		if (shell.isVisible() == visible)
-			return;
 
 		if (!visible) {
 			super.setVisible(false);
@@ -342,6 +340,8 @@ public class BrowserInformationControl extends AbstractInformationControl implem
 			return;
 		}
 
+		if (shell.isVisible() == visible)
+			return;
 		/*
 		 * The Browser widget flickers when made visible while it is not completely loaded.
 		 * The fix is to delay the call to setVisible until either loading is completed
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControl.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControl.java
@@ -505,7 +505,7 @@ public abstract class AbstractInformationControl implements IInformationControl,
 
 	@Override
 	public void setVisible(boolean visible) {
-		if (fShell.isVisible() == visible)
+		if (visible && fShell.isVisible() == visible)
 			return;
 
 		fShell.setVisible(visible);


bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/StickyHoverManager.java
```diff
@@ -142,6 +144,14 @@ public void stop() {
 			if (fInformationControlToClose != null)
 				fInformationControlToClose.removeFocusListener(this);
 
+			// Was added to fix the bug #2534
```
iloveeclipse: Nice to have this, but it doesn't help the reader to understand the code (this info is available also by git blame). Better is to say **why** we manually touch the shell here.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/StickyHoverManager.java
```diff
@@ -142,6 +144,14 @@ public void stop() {
...
+			if (fInformationControlToClose instanceof AbstractInformationControl infoControl) {
+				Shell shell= infoControl.getShell();
+				if (!shell.isDisposed()) {
+					shell.setVisible(false);
```
iloveeclipse: The fix works, but why don't we have that in `hideInformationControl()` ?
1) The "Closer" can be set so it is not guaranteed this code will work for all implementations.
2) The `Closer.stop()` javadoc says `Tells this closer to stop monitoring the subject and the information control` which doesn't fit here.

What about this patch:
```
diff --git a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControlManager.java b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControlManager.java
index d7be920..2146db2 100644
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControlManager.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControlManager.java
@@ -26,2 +26,3 @@
 import org.eclipse.swt.widgets.Monitor;
 
@@ -1212,4 +1213,11 @@
 			fInformationControl.setVisible(false);
-			if (fInformationControlCloser != null)
+			if (fInformationControlCloser != null) {
 				fInformationControlCloser.stop();
+			}
+			if (fInformationControl instanceof AbstractInformationControl infoControl) {
+				Shell shell= infoControl.getShell();
+				if (!shell.isDisposed()) {
+					shell.setVisible(false);
+				}
+			}
 		}
```

mehmet-karaman: Thanks.. Going to use this patch, test this and undo my changes.

iloveeclipse: Please try. I have only quickly tested with Java editor, no idea if Xtext goes its own way, as I see many different implementations around.

mehmet-karaman: I've tested the solution in Java Editor and Xtext Editor.. Both are working.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControlManager.java
```diff
@@ -1210,8 +1211,15 @@ protected void hideInformationControl() {
...
 			storeInformationControlBounds();
 			fInformationControl.setVisible(false);
-			if (fInformationControlCloser != null)
+			if (fInformationControlCloser != null) {
 				fInformationControlCloser.stop();
+			}
+			if (fInformationControl instanceof AbstractInformationControl infoControl) {
+				Shell shell= infoControl.getShell();
+				if (!shell.isDisposed()) {
+					shell.setVisible(false);
```
iloveeclipse: Now seeing this line added below `fInformationControl.setVisible(false);` I wonder **why** is the shell not hidden by `fInformationControl.setVisible(false);` by default?

Could you please check that, and also in different re-incarnations of `fInformationControl`?

mehmet-karaman: The Java Doc hoveer is a BrowserInformationControl. The setVisible-Method call will lead to the following method implementation in org.eclipse.jface.internal.text.html.BrowserInformationControl.setVisible(boolean) ..

There it is checking shell.isVisible() .. which returns usually false, when eclipse is minimized and is not going to apply the visibility to the shell anymore.. But it will be visible if eclipse will come back (but the listeners are not registered)

mehmet-karaman: Same happens in the default implementation of org.eclipse.jface.text.AbstractInformationControl.setVisible(boolean) .. So we have to explicitly set the visibility to avoid showing the sticky hover when eclipse is back in foreground with deregistered listeners..

iloveeclipse: > The Java Doc hoveer is a BrowserInformationControl.

Note, we have multiple implementations around in SDK alone: 
image

Ideally we should check at least few of them how they work with this change

> There it is checking shell.isVisible() .. which returns usually false, when eclipse is minimized and is not going to apply the visibility to the shell anymore

Which brings a question: should we fix it **there** (`org.eclipse.jface.text.AbstractInformationControl.setVisible(boolean)` and others)?

If not, I would add a comment over the new block saying that for whatever reason known implementations of `IInformationControl.setVisible()` tend to omit `shell.setVisible()` call in some cases.

The change was added in 91e32933fa40aea0335d03bdebbbd304bfecebf5 without proper commit message.
I assume it was some optimization ( `shell.setVisible(true)` may do some more work as needed)?

The question if we should either remove this optimization completely or only have it for "true" case?

mehmet-karaman: Right.. Let me check that.. Seems more complicated than expected..

mehmet-karaman: There are 6 Implementations which override the setVisible() Method of the AbstractInformationControl class. 4 of them are using the super call so the adjustment of the AbstractInformationControl class is enough for those. The other 2 have a similar check so the same adjustment needs to be done there locally: BrowserInformationControl, XbaseInformationControl.

To be totally sure, I tried to avoid unnecessary behavior changes when visibility is going to be changed to false, it will be set to false (without checking isVisible() ) .. But for the true case there is no change.. (to avoid flickering or some other stuff which I couldn't foresee) ..

iloveeclipse (approved):
LGTM

iloveeclipse:
@mehmet-karaman : I've verified last patch on Linux, have you verified it on Windows?

iloveeclipse:
Just rebased. jenkins had an OOM error while running ui tests. Surely not related to this change.

@HannesWell , @eclipse-platform/eclipsefdn-releng  : are you aware about any environment changes for jenkins insances?

HannesWell:
> @HannesWell , @eclipse-platform/eclipsefdn-releng : are you aware about any environment changes for jenkins insances?

I'm not aware of any. Maybe this was just by chance. The latest build seem to have succeeded.

mehmet-karaman:
> @mehmet-karaman : I've verified last patch on Linux, have you verified it on Windows?

@iloveeclipse Yes, it works also on windows.

PR #2557, opened 2024-11-28
CSS: Allow Color And Font Definitions To Be Marked As Not Editable

BeckerWdf:
Color and Fonts can be defined as not editable if done via the extensions point. Provide the same possibility if they are defined via CSS.

Needed for #2455


BeckerWdf:
Failure are unrelated. Merging this.

PR #2558, opened 2024-11-28
Fix undo / redo icons according to standard

BeckerWdf:
Contributes to:
https://github.com/eclipse-platform/eclipse.platform.images/issues/113


vogella:
Our build infrastructure seems to be very unstable these days.... Maybe it is because master just opened up.

vogella:
Windows build failure looks unrelated as this only updates icons, I suggest to merge

PR #2549, opened 2024-11-27
Move "Light (Preview)" Theme to the "Light" Theme

BeckerWdf:
- Copies over all changes done in the "Light (Preview)" theme to the "Light" theme. So the changes done in the preview theme are now the default.

- Delete the "Light (Preview)" theme as it's no longer needed.

Also:
- Simplify the plugin.properties file to remove duplicated strings.

See: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114

PR #2559, opened 2024-11-28
Revert "Move "Light (Preview)" Theme to the "Light" Theme"

mickaelistria:
Reverts eclipse-platform/eclipse.platform.ui#2549
See whether the OOM happens


mickaelistria:
OOM too, looks like this change was not the cause.

PR #2561, opened 2024-11-29
Remove gradient from placeholder images in JFace

vogella:
The gradient placeholder is replace with just an empty image.


vogella:
@HannesWell sorry to ping you again, but why does the version bot not add a commit to update the jface plug-in version in this case?

BeckerWdf:
See also https://github.com/eclipse-platform/eclipse.platform.images/pull/116

BeckerWdf:
Can you please elaborate on this change?
Can you show before and after screenshots?

merks:
This seems so arbitrary, without justification.  What purpose can a bunch of completely transparent images serve?

vogella:
> Can you please elaborate on this change? Can you show before and after screenshots?

Sure, will do, just waiting for our "random" build failures to vanish.

vogella:
> This seems so arbitrary, without justification. What purpose can a bunch of completely transparent images serve?

We replaced a lot of (ugly) gradients in the past with solid colors. If you do not have a gradient in a component, you end up with a solid background color. The common path forward seems to be that gradients are not very modern looking and solid background are nicer. For reference you could compare the old forms styling from older releases (https://eclipse.dev/eclipse/news/4.5/M5/images/dark_theme_old_pde.png) with the latest styling.

BeckerWdf:
Any news on this?

vogella:
Build still failing, so no news here

vogella:
New (without the gradient in the right top corner) on the left side, old on the right side for Eclipse 4 application model

image

image

And the wizard for Git repo

image

image


Please note that the change only affect wizard which do not specify their own images. So JDT wizards, PDE wizards, platform wizards will not be affected.

BeckerWdf:
Ah now I get it.
The old version looked broken anyway to me.

PR #2564, opened 2024-12-02
Update tycho-build to 4.0.10

akurtakov:


akurtakov:
api tools failure prevents jenkins from succeeding but this is releng change that syncs tycho version thus merging.

PR #2565, opened 2024-12-02
PropertyPageNode: log full exception if any instead of popup only

jukzi:
This issue did not show the root exception:
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/570


jukzi:
@SougandhS this should help to get better exceptions in your case.

SougandhS:
> @SougandhS this should help to get better exceptions in your case.

Yess 👍

jukzi:
With this PR:
image

jukzi:
Random build failures are known:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/926
https://github.com/eclipse-platform/eclipse.platform.ui/issues/370

Issue #867, opened 2023-05-18
Random failing AnnotationManagerTest.testRemoveQuery


Shee43:
In the Build [I20230517-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20230517-1800/), the below tests in org.eclipse.search.tests.filesearch.AnnotationManagerTest failed  
> 
on [macM1](https://download.eclipse.org/eclipse/downloads/drops4/I20230517-1800/testresults/html/org.eclipse.search.tests_ep428I-unit-macM1-java17_macosx.cocoa.aarch64_17.html)
> 
> ```
>java.lang.AssertionError: expected:<0> but was:<6>
>at org.junit.Assert.fail(Assert.java:89)
>at org.junit.Assert.failNotEquals(Assert.java:835)
>at org.junit.Assert.assertEquals(Assert.java:647)
>at org.junit.Assert.assertEquals(Assert.java:633)
>at org.eclipse.search.tests.filesearch.AnnotationManagerTest.testRemoveQuery(AnnotationManagerTest.java:144)
>at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
>at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>at java.base/java.lang.reflect.Method.invoke(Method.java:568)
> ```

Shee43:
In the Build [I20230703-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/), the below tests in org.eclipse.search.tests.filesearch.AnnotationManagerTest failed

on [mac64](https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.search.tests_ep429I-unit-mac64-java17_macosx.cocoa.x86_64_17.html)

>```
>expected:<0> but was:<6>
>java.lang.AssertionError: expected:<0> but was:<6>
>at org.junit.Assert.fail(Assert.java:89)
>at org.junit.Assert.failNotEquals(Assert.java:835)
>at org.junit.Assert.assertEquals(Assert.java:647)
>at org.junit.Assert.assertEquals(Assert.java:633)
[...]

PR #2567, opened 2024-12-02
Implement filter-based TableViewer as a counterpart to the FilteredTree

ptziegler:
Projects such as PDE currently use the FilteredTree in combination with a child-less ITreeContentProvider to simulate a filtered table. This approach brings a big performance overhead, as the tree still has to handle functionality such as expansion and contraction of the (flat) tree structure which especially large trees/tables causes a significant overhead.


ptziegler:
See https://github.com/eclipse-pde/eclipse.pde/issues/1497 for a potential use-case.

Note: This also works with virtual tables:

[recording.webm](https://github.com/user-attachments/assets/97a8f563-5477-45a2-81dd-ca5cebbf8ff7)

vogella:
Did you also consider the FilteredTree from the E4 package? This way this class could also be used in pure e4 rcp apps.

ptziegler:
> Did you also consider the FilteredTree from the E4 package? This way this class could also be used in pure e4 rcp apps.

Not for the time being. And looking at it, it really is just a copy-paste of the E3 class. So ideally, I'd like to get some feedback on the current implementation before I repeat the same for the E4 one.

vogella:
IIRC we copied it and removed the E3 reference (IIRC the help system) and added the option to enable/ disable the filter at runtime.

If it's functionality is the same I think we should only add one class ( to the E4 package so that it can be used in ide and E4 RCP).

ptziegler:
Is there an E4 counterpart for the [TextMatcher](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/misc/TextMatcher.java)? Would it perhaps make sense to move it to e.g. `org.eclipse.text`?

vogella:
> Is there an E4 counterpart for the [TextMatcher](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/misc/TextMatcher.java)? Would it perhaps make sense to move it to e.g. `org.eclipse.text`?

I'm not aware of an e4 counterpart. Would be great to move this to an API package that everyone can use.

laeubi:
> > Is there an E4 counterpart for the [TextMatcher](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/misc/TextMatcher.java)? Would it perhaps make sense to move it to e.g. `org.eclipse.text`?
> 
> I'm not aware of an e4 counterpart. Would be great to move this to an API package that everyone can use.

By the way it is confusing to talk about E3/E4 "counterparts" here as these are completely independent from the used technique (e.g. it does not require any DI at all nor does it requires E3 extension points). So making the class somewhere more generic available would be beneficial and as it is a final class let the "old" extend the "new" should work flawlessly.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredTable.java
```diff
@@ -0,0 +1,240 @@
...
+import org.eclipse.ui.internal.misc.TextMatcher;
+import org.eclipse.ui.progress.WorkbenchJob;
+
+/**
+ * A simple control that provides a text widget and a table viewer. The contents
+ * of the text widget are used to drive a TextMatcher that is on the viewer.
+ *
+ * @since 3.135
+ */
+public non-sealed class FilteredTable extends AbstractFilteredViewer {
```
laeubi: Just an idea but maybe having a JFace `FilteredTableViewer` could be a better place for this so it can be used even outside of Eclipse Platform in SWT/JFace Applications.

Maybe even one can abstract the whole thing to use a `StructuredViewer` then it can be used with any table/tree/grid ...

vogella:
> By the way it is confusing to talk about E3/E4 "counterparts" 

It is not, new elements which can be used in E3 and pure E4 must of have dependencies to the old workbench implementation, as the e3 initialization code does not run in pure E4.

laeubi:
> It is not, new elements which can be used in E3 and pure E4 must of have dependencies to the old workbench implementation, as the e3 initialization code does not run in pure E4.

Them mentioned class do not have any dependencies to a (E3 or E4) workbench as far as I can see...

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredTable.java
ptziegler: > Just an idea but maybe having a JFace `FilteredTableViewer` could be a better place for this so it can be used even outside of Eclipse Platform in SWT/JFace Applications.

The `FilteredTable`/`FilteredTree` depend on the `WorkbenchJob`/`BasicUIJob`, so there is an inherent dependency to the E3/E4 workbench, which effectively prevents them to be moved to JFace.

Though it might be possible to specify a plain Job in the abstract base class and then have the implementations provide their respective instances... This would at the very least avoid the code duplication.

> Maybe even one can abstract the whole thing to use a `StructuredViewer` then it can be used with any table/tree/grid ...

The base class only specifies a `ColumnViewer` but one could probably go even lower...

ptziegler:
In principle, the TextMatcher class should work regardless of whether an E3 or E4 workbench is used. I've therefore moved the class from org.eclipse.ui.workbench to org.eclipse.jface.text. The same is true for the abstract base class I've created for the FilteredTable/Tree, so that I don't have to duplicate the implementation for the E4 classes.

Though I still have to think about how to best adapt the E4 classes...

laeubi:
> Is there an E4 counterpart for the [TextMatcher](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/misc/TextMatcher.java)? Would it perhaps make sense to move it to e.g. `org.eclipse.text`?

Interestingly enough I found we already have `org.eclipse.core.text.StringMatcher` so maybe that's another place to put a generic `TextMatcher` ...

ptziegler:
> Interestingly enough I found we already have org.eclipse.core.text.StringMatcher so maybe that's another place to put a generic TextMatcher ...

Given that this is an internal class in can be freely moved to wherever it's needed. But as far as I know, it's only used for UI stuff, so I don't think it makes sense to move it out of the UI repo.

ptziegler:
I've tought about this PR a little bit more and I'd like to split it into three parts:

1) Move TextMatcher to Platform
2) Move common components of the FilteredTree's into an abstract base class
3) Implement a FilteredTable using this base class.

HannesWell:
> 1. Move TextMatcher to Platform

I have probably oversaw something, but can you please explain why this is necessary?
`org.eclipse.text` doesn't have any UI dependency so using it in a non-UI app is perfectly fine. In fact there are already bundles, that reside in eclipse-platform that depend on it, like `org.eclipse.core.filebuffers`. The situation that `org.eclipse.text` is placed in `platform-ui` is indeed sub-optimal and misleading and that there are circular dependencies between platform and platform-ui is bad, but that's the situation and `o.e.text` is usable. IIRC there was even a discussion to move it.

vogella:
@ptziegler with the Equinox change in, can you update this PR?

bundles/org.eclipse.e4.ui.dialogs/.settings/.api_filters
```diff
@@ -0,0 +1,11 @@
...
+<component id="org.eclipse.e4.ui.dialogs" version="2">
+    <resource path="src/org/eclipse/e4/ui/dialogs/filteredtree/FilteredTable.java" type="org.eclipse.e4.ui.dialogs.filteredtree.FilteredTable$NotifyingTableViewer">
+        <filter id="571473929">
+            <message_arguments>
+                <message_argument value="TableViewer"/>
+                <message_argument value="NotifyingTableViewer"/>
+            </message_arguments>
+        </filter>
+    </resource>
+</component>
```
ptziegler, line 11: Not sure how to handle the API warning because the UI bundles are implementing the JFace viewers directly (just like the existing FilteredTree does).
Reason being that the filter cache is not properly cleared when e.g. a node is no longer filtered. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=187200

ptziegler:
That should hopefully be all. I've rebased everything and adapted the code to use the new API introduced in previous commits.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredTable.java
ptziegler: I thought about it a little bit more and in general, it still holds that the dependency to the Eclipse runtime prevents those filters to be moved to `org.eclipse.jface`. However, it would be possible to move them to e.g. `org.eclipse.jface.text`. I'm certainly not a huge fan of having a separate copy specifically for pure E4 applications. wdyt?

vogella:
Could the new FilteredTable also ONLY life in the e4 bundle?

vogella:
Thanks @ptziegler. One last question, would it be sufficient to have the FilteredTree in the e4 plug-in or do we also need the version in the org.eclipse.ui.dialogs?

ptziegler:
> Could the new FilteredTable also ONLY life in the e4 bundle?

> Thanks @ptziegler. One last question, would it be sufficient to have the FilteredTree in the e4 plug-in or do we also need the version in the org.eclipse.ui.dialogs?

Apologies. I suppose it would? It's perfectly fine to use the e4 dialogs in e.g. plain views and it's not like you can create "pure" e3 applications anymore. The only downside is that projects then get a new dependency to `org.eclipse.e4.ui.dialogs`, but I don't think it matters for such a small bundle.

This then however begs the question why the FilteredTree in `org.eclipse.ui.dialogs` wasn't deprecated, when a copy was created for the e4 component, given that they provide the same functionality.

vogella:
> > Could the new FilteredTable also ONLY life in the e4 bundle?
> 
> > Thanks @ptziegler. One last question, would it be sufficient to have the FilteredTree in the e4 plug-in or do we also need the version in the org.eclipse.ui.dialogs?
> 
> Apologies. I suppose it would? It's perfectly fine to use the e4 dialogs in e.g. plain views and it's not like you can create "pure" e3 applications anymore. The only downside is that projects then get a new dependency to `org.eclipse.e4.ui.dialogs`, but I don't think it matters for such a small bundle.

I think that should be fine. Please update the PR so that we have only one new implementation, the e4 one.
 
> This then however begs the question why the FilteredTree in `org.eclipse.ui.dialogs` wasn't deprecated, when a copy was created for the e4 component, given that they provide the same functionality.

IIRC the e4 FilteredTable did not support "quickSelectionMode" but I could be wrong. If it provides the same functionality, we should deprecate the old one.

vogella:
@ptziegler Is it OK for m e to push an update with the change in org.eclipse.ui.dialogs removed?

ptziegler:
> @ptziegler Is it OK for m e to push an update with the change in org.eclipse.ui.dialogs removed?

Yeah, sure. Sorry, things are a little busy again right now and I wanted to look at it again today.

> IIRC the e4 FilteredTable did not support "quickSelectionMode" but I could be wrong. If it provides the same functionality, we should deprecate the old one.

I remember again why the old class can't be deprecated: The filter control can be configured via the `org.eclipse.ui/SHOW_FILTERED_TEXTS` preference, which obviously doesn't exist in the e4 context.

vogella:
> > @ptziegler Is it OK for m e to push an update with the change in org.eclipse.ui.dialogs removed?
> 
> Yeah, sure. Sorry, things are a little busy again right now and I wanted to look at it again today.

Thanks for the feedback, in this case I wait for your update. No need to rush this.

ptziegler:
I got rid of the duplicate class in `org.eclipse.ui` and updated the example/test to use the e4 implementation instead.

ptziegler:
>  Error:  Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:4.0.11-SNAPSHOT:test (default-test) on project org.eclipse.search.tests: There are test failures.

Test on MacOS seems unrelated, as I only add new API.

vogella:
Thanks @ptziegler 

You may want to add this to the N&N.

vogella:
Test failure seems unrelated and seems to happen randomly. https://github.com/eclipse-platform/eclipse.platform.ui/issues/867

merks:
I'm trying to understand the reason for the deprecation of org.eclipse.ui.dialogs.FilteredTree.updateToolbar(boolean).

It's not unused:

image

And there are overrides that expect the method to be called to do things specific to the subclass:

image

So I wonder if these subclasses are broken now?

I noticed this in Oomph where there were no warnings but these new ones showed up with a new target platform:

image

ptziegler:
The `updateToolbar()` method used to be called within the update job of the `FilteredTree` up until 3726516c0ab381d3d29e24e412f73679e830aba2. When I initially saw this method I wondered how exactly it fit into this class. Because to me, it doesn't make much sense to define a method which is then never called, hence the deprecation.

> So I wonder if these subclasses are broken now?

If the subclasses override `doCreateRefreshJob()` and explicitly call `updateToolbar()` from there, everything is fine. But in case of e.g. `LaunchConfigurationFilteredTree` and `RegistryFilteredTree`, where this is not the case: Yes, they are broken. Not since now, but rather since over 6 years.

The motivation behind deprecating this method was to spot exactly those instances where subclasses interact with this method. Otherwise this probably would've gone unnoticed for another 6 years.

In hindsight, I probably should've noticed that something was off and looked closer at the history, as well as at some of the related projects. That is 100% on me.

So in the end, this means parts of 3726516c0ab381d3d29e24e412f73679e830aba2 need to be reverted so that `updateToolbar()` is called again, which then also needs to be undeprecated.

PR #2572, opened 2024-12-03
ToolItemUpdater: fix possible Integer overflow

jukzi:
Both DELAY and 1_000_000 are Integers. Multiplied they could exceed max integer. Currently the DELAY is small enough to not overflow, but changing the constant would silently overflow.


jukzi:
CI fails with unrelated errors that i do not understand
`MavenProject: org.eclipse.platform:org.eclipse.ui.tests:3.15.1900-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.platform.ui_PR-2572/tests/org.eclipse.ui.tests/.polyglot.META-INF: baseline and build artifacts have same version but different contents`
@akurtakov any idea?

jukzi:
ignoring unrelated random fail https://github.com/eclipse-platform/eclipse.platform.ui/issues/195

PR #2573, opened 2024-12-03
Use SubMonintor in TextChange

vogella:
done() call not necessary here and convert can also handle null.


vogella:
cc @ptziegler

PR #2574, opened 2024-12-03
Use SubMonintor in CheckConditionsContext

vogella:
done() call not necessary here and convert can also handle null.  Also split will also check for cancellation so the additional cancellation check is not necessary.


jukzi (approved):

vogella:
@ptziegler can you review?

vogella:
Planning to merge for next release once master opens up

PR #2575, opened 2024-12-03
Use SubMonintor in FileBufferOperationAction

vogella:
done() call not necessary here and convert can also show the label and set the number of work unit.


jukzi:
May i ask to add or modify an existing test to actually test the progress using FussyProgressMonitor? I have seen so much things broken by needless conversion to Subprogressmonitor without proper testing.

vogella:
> May i ask to add or modify an existing test to actually test the progress using FussyProgressMonitor? I have seen so much things broken by needless conversion to Subprogressmonitor without proper testing.

Its about removing the unnecessary done call, not about progress reporting.

jukzi:
The FussyProgressMonitor will check if its ok to remove the done call or not, i.e. if all progress is reported.

PR #2578, opened 2024-12-03
Remove done() call on SubMonitor in Refactoring

vogella:
done() call not necessary here


jukzi (dismissed):
The caller `Will not expect that the callee invokes done() on the monitor. The caller must either use an SubMonitor (or a similar implementation which does not require done() to be invoked), or it must take responsibility for calling done() on the monitor after the callee has finished.`
Reads like this done() is valid

vogella:
> The caller `Will not expect that the callee invokes done() on the monitor. The caller must either use an SubMonitor (or a similar implementation which does not require done() to be invoked), or it must take responsibility for calling done() on the monitor after the callee has finished.` Reads like this done() is valid

See Changes in Eclipse 4.7

PR #2581, opened 2024-12-04
Mark deprecated and unsupported constants for deletion in

vogella:
IWorkbenchPreferenceConstants

Also remove initialization of unsupported preference values


vogella:
Linux fail: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2603

PR #2585, opened 2024-12-04
Remove unused variable

vogella:


jukzi (changes requested):
only shifts the problem marker to "The allocated object is never used"

PR #2586, opened 2024-12-05
Revert "Use SubMonintor in TextChange"

jukzi:
Reverts eclipse-platform/eclipse.platform.ui#2573

see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1833


vogella:
Revert is fine for me for now, see the JDT report for analysis, I might get back to remove this call with the full consumption of units of work of the submonitor as the JDT test is testing for in a later commit.

vogella (approved):

PR #2587, opened 2024-12-05
Properly dispose font in FontRegistry upon display disposal

HeikoKlare:
When retrieving a font from a FontRegistry instance that is not cached in the registry yet, it will be created on the current display. In order to dispose the fonts of a display upon the display's disposal, the FontRegistry registers a dispose hook at the display instance. The current implementation only does this for the first display that is instantiated. For every further display, which may be created on Windows systems, no hook is added. Thus, if an additional display is disposed, its fonts remain in the FontRegistry, referring to a disposed display.

This change improved the FontRegistry implementation to properly register a dispose hook to every display for which it contains fonts. It also adds an according regression test.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.MissingResourceException;
 import java.util.ResourceBundle;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.jface.util.Policy;
@@ -207,7 +208,7 @@ public class FontRegistry extends ResourceRegistry {
 	 */
 	protected Runnable displayRunnable = this::clearCaches;
 
-	private boolean displayDisposeHooked;
+	private final Set<Display> displayDisposeHooked = ConcurrentHashMap.newKeySet();
 
 	private final boolean cleanOnDisplayDisposal;
 
@@ -492,7 +493,7 @@ public class FontRegistry extends ResourceRegistry {
 		if (display == null) {
 			return null;
 		}
-		if (cleanOnDisplayDisposal && !displayDisposeHooked) {
+		if (cleanOnDisplayDisposal && !displayDisposeHooked.contains(display)) {
 			hookDisplayDispose(display);
 		}
 
@@ -718,7 +719,7 @@ public class FontRegistry extends ResourceRegistry {
 		stringToFontRecord.clear();
 		staleFonts.clear();
 
-		displayDisposeHooked = false;
+		displayDisposeHooked.remove(Display.getCurrent());
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
```diff
@@ -207,7 +208,7 @@ void addAllocatedFontsToStale(Font defaultFont) {
 	 */
 	protected Runnable displayRunnable = this::clearCaches;
 
-	private boolean displayDisposeHooked;
+	private final Set<Display> displayDisposeHooked = new HashSet<>();
```
jukzi: may i ask if this can be used concurrently. Especially if one uses two Display-Threads on windows will they share this instance so that it be thread safe?

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/resources/FontRegistryTest.java
```diff
@@ -41,4 +52,47 @@ public void testBug544026() {
...
+		Font fontOnThisDisplayBeforeSecondDisplayDispose = fontRegistry.defaultFont();
+		Device displayOfFontOnSecondDisplay = fontOnSecondDisplay.getDevice();
+		// font registry returns same font for every display
+		assertEquals(secondDisplay, displayOfFontOnSecondDisplay);
+		assertEquals(fontOnThisDisplayBeforeSecondDisplayDispose, fontOnSecondDisplay);
+
+		// after disposing font's display, registry should reinitialize the font
+		secondDisplay.syncExec(secondDisplay::dispose);
+		Font fontOnThisDisplayAfterSecondDisplayDispose = fontRegistry.defaultFont();
+		assertNotEquals(fontOnThisDisplayAfterSecondDisplayDispose, fontOnSecondDisplay);
```
fedejeanne, line 74: I know this should always be the case but I would also check that the font has been disposed too, just in case:

```suggestion
		assertNotEquals(fontOnThisDisplayAfterSecondDisplayDispose, fontOnSecondDisplay);
		
		// When the display is disposed, its fonts are disposed too.
		assertTrue(fontOnSecondDisplay.isDisposed());
```

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
HeikoKlare: I made this thread safe in 11c2271.
If I am not mistaken, it may not be necessary, because even though threads may see different states of the data structure, only the data that a thread itself has added is relevant. I.e., the `contains` check only needs to see the display that may have been added by the exact same thread before. But that would be hard to understand and the thread-safe solution does also not introduce performance drawbacks.
An alternative might be to use a thread-local boolean indicating whether the display for the current thread was hooked, but I would also find that harder to understand (but that might be up to personal preference).

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/resources/FontRegistryTest.java
HeikoKlare: Added in 11c2271.

fedejeanne (approved):

vogella:
Nice to see that at least one PR builds green. These days this seems very rare even for things like Readme changes.....

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
fedejeanne: @jukzi has your concern been addressed? May I merge this PR?

jukzi: yes, thanks.

PR #2593, opened 2024-12-06
Feature Proposal: Rasterization of SVGs at Runtime for Eclipse Icons 

### LINKED ISSUE #1438: Improving Eclipse Icon Scaling by Supporting Vectorized Icons
Eclipse currently relies on PNG files for displaying icons, requiring two different sizes to handle various zoom levels. For zoom factors between 100% and 174%, the original-size icon is used, while for zoom levels between 175% and 200%, a double-sized icon is displayed. This approach has several drawbacks:

1. **Quality Loss with Autoscaling**: Enabling autoscaling for icons using the flags `-Dswt.autoScale=quarter` or `-Dswt.autoScale=exact` can cause quality degradation, particularly on high-DPI displays or at high zoom levels, since PNGs do not scale well.

2. **Maintenance Challenges**: Maintaining multiple PNG versions in different sizes is not a sustainable solution, especially as higher DPI displays may require even more icon variations.

Currently, new icons are manually created by converting SVGs to PNGs, a process that is both time-consuming and inefficient. By fully utilizing SVGs, which support resolution-independent scaling, these problems can be effectively addressed.

Previous discussions have highlighted these issues, such as those documented in this [Eclipse bug report](https://bugs.eclipse.org/bugs/show_bug.cgi?id=568215).

My bachelor’s thesis explores how SVGs can be better integrated into Eclipse to improve icon scalability and reduce maintenance overhead. If you have any insights or suggestions, please feel free to reach out or leave a comment!

Michael5601:
See the following [PR](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638) in Eclipse SWT for a detailed description of the changes.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1438.


HannesWell:
I just pushed the changes I made for/in our yesterdays 1:1 meeting, mainly adapt to the change that `o.e.swt.svg` is a fragment.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -59,6 +61,16 @@ public URLImageFileNameProvider(String url) {
 		public String getImagePath(int zoom) {
 			URL tempURL = getURL(url);
 			if (tempURL != null) {
+				SVGRasterizer rasterizer = SVGRasterizerRegistry.getRasterizer();
+				if (rasterizer != null) {
+					try (InputStream in = getStream(tempURL)) {
+						if (rasterizer.isSVGFile(in)) {
```
HannesWell: If we would just check the file-extension here we would save creating an `InputStream` for each image to render. The latter is probably a significant performance overhead.
Furthermore it would save us from making `SVGRasterizerRegistry` and `SVGRasterizer` public.

Michael5601: Please see my [comment](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638/commits/44e6f19d6eaa1cf09e59e1970b3e6c9c52a0285b#r1881066057).

We could nevertheless change the isSVGFile() call in Platform UI to a file-extension check. This way the method can't be deleted but `SVGRasterizerRegistry` and `SVGRasterizer` could be internal. 

Edit: How can we make `SVGRasterizer` internal? It is implemented by `JSVGRasterizer` so it must be visible.

Michael5601: I made `SVGRasterizerRegistry` internal and changed isSVGFile to a file-extension check.

Michael5601:
I removed the added SVGs from this PR. Please see the new [PR](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2643) for the SVGs if you want to test this feature.

In the meantime I locally changed all PNGs I could find in Platform UI, Platform, JDT UI and SWT for testing the Performance. I had no problems whatsoever when starting Eclipse and it was also quite easy and fast to add the SVGs and change the paths. I did not delete any PNGs in this process but changed ".png" to ".svg" in most places. 

I will report the performance data soon.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchImages.java
```diff
@@ -236,6 +236,7 @@ private static final void declareImages() {
 		declareImage(ISharedImages.IMG_DEF_VIEW, PATH_EVIEW + "defaultview_misc.png", true); //$NON-NLS-1$
 
 		declareImage(IWorkbenchGraphicConstants.IMG_LCL_CLOSE_VIEW, PATH_ELOCALTOOL + "close_view.png", true); //$NON-NLS-1$
+		declareImage(IWorkbenchGraphicConstants.IMG_LCL_PIN_VIEW, PATH_ELOCALTOOL + "pin_view.png", true); //$NON-NLS-1$
```
HannesWell: This change seems unrelated to the topic of this PR.
If this is an unintended left-over please remove it, if it's missing please add it in a separate PR.
```suggestion
		declareImage(IWorkbenchGraphicConstants.IMG_LCL_PIN_VIEW, PATH_ELOCALTOOL + "pin_view.png", true); //$NON-NLS-1$
```

features/org.eclipse.e4.rcp/feature.xml
```diff
@@ -294,6 +294,10 @@
          arch="aarch64"
          version="0.0.0"/>
 
+   <plugin
+         id="org.eclipse.swt.svg"
+         version="0.0.0"/>
+
```
HannesWell: In order to include the org.eclipse.swt.svg fragment in the provided Eclipse products this part is important.

As far as I can tell no code changes are necessary in Jface to make SVGs usable in Eclipse, aren't they?
Or to ask the other way, round: Would it make sense to just submit this part now?

HeikoKlare: The only change required is the adaptation of the `URLImageDataProvider`, which directly instantiates `ImageData` in a non-zoom-aware way. This needs to be enhanced in order to avoid that SVGs are rasterized at 100% and then auto-scaled.

laeubi: As mentioned [here](https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944#issuecomment-2753513977) I don't think dedicated inclusion in a feature is the right approach and instead a requirement should be added to bundles containing SVGs.

laeubi (dismissed):

features/org.eclipse.e4.rcp/feature.xml
HannesWell: Lets not have that discussion at multiple places. :) TLDR; I think this is fine:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944#issuecomment-2753574505

laeubi: I don't think its fine to introduce yet another implicit dependency. Just make it explicit and it will work everywhere.

Michael5601: @HannesWell can this be deleted now?

HeikoKlare: yes, please remove this for now

Michael5601: done.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
```diff
@@ -121,12 +125,15 @@ public boolean equals(Object o) {
...
 	@Override
 	public ImageData getImageData(int zoom) {
-		InputStream in = getStream(zoom);
-		if (in != null) {
-			try (BufferedInputStream stream = new BufferedInputStream(in)) {
-				return new ImageData(stream);
+		SourceAtZoom<InputStream> inputStreamAtZoom = getStream(zoom);
+		if (inputStreamAtZoom != null) {
+			try (BufferedInputStream stream = new BufferedInputStream(inputStreamAtZoom.source())) {
+				ElementAtZoom<ImageData> imageData = NativeImageLoader
```
HeikoKlare: This is one of the crucial points: `getStream(125)` will yield `null` for a file name referring to an SVG and thus no loading is performed.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
```diff
@@ -53,16 +57,16 @@ public String getImagePath(int zoom) {
...
+				String xResult = getFilePath(xName.source(), logIOException);
 				if (xResult != null) {
 					return xResult;
 				}
 			}
-			String xPath = getxPath(name, zoom);
+			SourceAtZoom<String> xPath = getxPath(name, zoom);
 			if (xPath != null) {
-				String xResult = getFilePath(xPath, logIOException);
+				String xResult = getFilePath(xPath.source(), logIOException);
```
HeikoKlare: This is another the crucial point: `getImagePath(125)` will yield `null` for a file name referring to an SVG.

HeikoKlare (commented):
If I am not mistaken, I am afraid that the proposed change is a no-op.

The purpose of the PR is to make the providers capable of processing files that can be used for retrieving images in a different zoom than what is currently provided by patterns for raster graphics files (like `@2x`, `32x32` and so on).
However, all used methods just return an element when the requested zoom exactly matches a file according to the above mentioned patterns. This means, the introduced `SourceAtZoom` will only contain an element if the zoom of that element fits to the zoom that was passed to the method returning it. So the zoom information in SourceAtZoom is not needed, as it's implicitly given by the contained element not being null anyway.

What really needs to be changed is that when, e.g., requesting image data for a path not following any of the patterns (i.e., currently being treated as a file of 100% zoom) and passing a zoom of, e.g., 175%, it should still be possible to retrieve data in that zoom from that file if the file access itself is capable of processing the zoom (as is available for SVGs). This does currently not happen.

Michael5601:
> If I am not mistaken, I am afraid that the proposed change is a no-op.
> 
> The purpose of the PR is to make the providers capable of processing files that can be used for retrieving images in a different zoom than what is currently provided by patterns for raster graphics files (like `@2x`, `32x32` and so on). However, all used methods just return an element when the requested zoom exactly matches a file according to the above mentioned patterns. This means, the introduced `SourceAtZoom` will only contain an element if the zoom of that element fits to the zoom that was passed to the method returning it. So the zoom information in SourceAtZoom is not needed, as it's implicitly given by the contained element not being null anyway.
> 
Is this only the case for the `FileImageDescriptor`? For the `URLImageDescriptor` the change in this PR should allow rasterization of SVGs in arbitrary sizes. I did not test it for the `FileImageDescriptor` yet and the `URLImageDescriptor` was also only called with the value 100 in my manual testing as the composite images were always generated with this value and then scaled afterwards.

HeikoKlare:
The `URLImageDescriptor` is (also) affected. When having a descriptor for an SVG, none of the xpath/xname matches will apply. And if none of them matches, the original source will only be used if zoom is 100. I.e., you can only retrieve the image data at zoom 100 for an SVG. Maybe I miss something, but that's how I understand the current code.

Consider for example the `CompositeImageDescriptor` implementation:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/b6af990cac142bdba33768e29ec0d62482f0ba54/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/CompositeImageDescriptor.java#L134-L144

The first call will yield `null` if zoom is not 100, so as a fallback the 100% image data is retrieved later on.

Michael5601:
> The `URLImageDescriptor` is (also) affected. When having a descriptor for an SVG, none of the xpath/xname matches will apply. And if none of them matches, the original source will only be used if zoom is 100. I.e., you can only retrieve the image data at zoom 100 for an SVG. Maybe I miss something, but that's how I understand the current code.
> 
> Consider for example the `CompositeImageDescriptor` implementation:
> 
> https://github.com/eclipse-platform/eclipse.platform.ui/blob/b6af990cac142bdba33768e29ec0d62482f0ba54/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/CompositeImageDescriptor.java#L134-L144
> 
> The first call will yield `null` if zoom is not 100, so as a fallback the 100% image data is retrieved later on.

The discussed problems should be fixed by commit 6a23f979b4494c1104a322b4001ba4b95c64ed41. I utilized the new `FileFormat.canLoadAtZoom()` API from this [change](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1987) for the `FileImageDescriptor` and `URLImageDescriptor`. It is now checked in advance whether loading an image for a specific zoom level is possible. If so, the corresponding path without "e.g. @2x" notation is returned or the relevant SVG data is directly loaded at that zoom level.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FileImageDescriptor.java
```diff
@@ -196,15 +202,16 @@ static String getxPath(String name, int zoom) {
 				int desiredHeight = Math.round((zoom / 100f) * currentHeight);
 				String lead = name.substring(0, matcher.start(1));
 				String tail = name.substring(matcher.end(2));
-				return lead + desiredWidth + "x" + desiredHeight + tail; //$NON-NLS-1$
+				String xPath = lead + desiredWidth + "x" + desiredHeight + tail; //$NON-NLS-1$
+				return new SourceAtZoom<>(xPath, desiredHeight);
```
HannesWell: I think this should have been the specified zoom. E.g. if I pass a path that contains a `/16x16/` folder and a zoom of `200`, the resulting `desiredHeight/Width` will be `32` and that's not the zoom of the result.
```suggestion
				return new SourceAtZoom<>(xPath, zoom);
```

HannesWell (commented):
After a reviewing session Michael and I agreed that the the `canLoad()` check should not be present for the `zoom == 100` case in the `ImageFileNameProviders` of both descriptor types as they always have to return the path for exactly the requested file-zoom.
The SWT Image class subsequently then can scale the image, depending on the type of Image (i.e. static vs. dynamically sizeable), while rasterizing it with a matching factor that reflects the relation of target-zoom to file-zoom. That doesn't work for SVGs if the filenameProvider returns the default (100%) file if the zoom is e.g. 150.

The `ImageDataProvider` is different as it returns an `ImageData` object that represents the image as it will be finally, without any (lossless) post-processing.
So while the `ImageDataProvider` returns an exact representation of an image for the requested zoom, the `ImageFilenameProvider` returns only a suitable source of an image (i.e. the file) at the requested zoom. But since the latter does not allow to supply any information about the 'native-zoom' of a file, the returned file must exactly match the expectation of the caller, i.e. the specified zoom or otherwise the subsequent rasterization of an SVG will be conducted with the wrong scaling factor. Therefore the `ImageFilenameProvider` implementations must not be smart and must just behave as before.


> However, all used methods just return an element when the requested zoom exactly matches a file according to the above mentioned patterns. This means, the introduced `SourceAtZoom` will only contain an element if the zoom of that element fits to the zoom that was passed to the method returning it. So the zoom information in SourceAtZoom is not needed, as it's implicitly given by the contained element not being null anyway.

That's actually right and I also checked this and step-by-step removed the `SourceAtZoom` and it showed that the passed zoom is actually not needed. At least except for the `zoom == 100` case in the `FileImgeDescriptor.getStream(int)` method, but that could have been resolved by refactoring the code a bit and adapting it more to the `URLImageDescriptor` class.
With that it was possible to avoid the `SourceAtZoom` entirely.
Then only changes to load ImageData directly where necessary, which also reflects the findings described above that the `ImageFilenameProviders` should be unchanged.

HannesWell (approved):
When the tests pass, I think this is ready for submission.

Michael5601:
Thank you for the session @HannesWell. It was really insightful for me. :)

Michael5601:
@HannesWell The tests passed. I also tested it manually again and it works perfectly. Would be great if we can submit. 

For everyone interested in trying this out:
This PR allows consumers outside of SWT to use SVGs. Essentially this can be seen by the composite icons in the eclipse workspace. 
To see changes the autoScaling-flag needs to be set to `quarter` or `exact` and one method needs to be adjusted in the class `CompositeImageDescriptor` as it works like a local implementation of the autoScaling-flag.
The method is `supportsZoomLevel(int)` and it must always return `true` to allow loading images for all zoom levels. 
We will fix this behaviour in a follow-up, which was already discussed [here](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1693).

HannesWell:
> The tests passed. I also tested it manually again and it works perfectly. Would be great if we can submit.

Great. Thanks for testing again!
I just fixed a minor flaw, where we didn't use the created `BufferedInputStream` but the originating stream directly. I also rearranged a few lines to have less changes in this PR.
Once the build succeeds I plan to submit this. Tomorrows I-build should then have this change available.

Thanks for all your awesome work on this topic!

PR #2602, opened 2024-12-09
Implement search-by-category for Import/Export wizard

### LINKED ISSUE #2576: New Wizard Dialog should also include Categories into Filter
If I search in the new project wizard for a category nothing is shown:

image

I would expect that all entries in the category are shown.

image

cc @ptziegler as he recently looked at the FilteredTree implementation.

ptziegler:
This adapts the WizardPatternFilter to also show the wizards, if the search string matches one of its containing categories.

Amends 532c0fb9b6851d77e3a7cc9f70a85c77a9ba3826
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2576

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WizardPatternFilter.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WizardPatternFilter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2015 IBM Corporation and others.
+ * Copyright (c) 2005, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -15,8 +15,11 @@
 package org.eclipse.ui.internal.dialogs;
 
 import java.util.ArrayList;
+import java.util.function.Function;
+import java.util.stream.Stream;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.ui.dialogs.PatternFilter;
+import org.eclipse.ui.wizards.IWizardCategory;
 
 /**
  * A class that handles filtering wizard node items based on a supplied matching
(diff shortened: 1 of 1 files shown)


vogella:
LGTM, tested it and works perfectly.

ptziegler:
Example:

image

Note that there is a special implementation for the content provider which flattens categories with only a single child:

image

This behavior is intentional, but whether it's a "good" behavior is a different question...

https://github.com/eclipse-platform/eclipse.platform.ui/blob/187669501065220bf4fdb793e4409ae8ca807905/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WizardContentProvider.java#L49-L52

vogella:
> This behavior is intentional, but whether it's a "good" behavior is a different question...

So you suggest to remove this flattening?

ptziegler:
> So you suggest to remove this flattening?

That would be my preference because the current behavior is not very intuitive. A `WizardCollectionElement` is flattened, if it doesn't have any other siblings. So the visibility of an element depends on your search string.

image

I found it confusion when `Test Import Add To Parent Category` was no longer a direct child of `General` after clearing the filter, even though that's exactly what the viewer told me.

However, this has been the behavior for over 20 years without any complaints. So I don't think the current behavior is wrong and if it is changed, then only as part of a separate commit.

vogella:
> That would be my preference because the current behavior is not very intuitive. A `WizardCollectionElement` is flattened, if it doesn't have any other siblings. So the visibility of an element depends on your search string.

I agree that we should remove this special treatment, I think consistent behavior is the UI is important. As you already into the code, please provide a separate PR for this adjustment.

vogella:
Linux builds seem to fail always unrelated to the actual change so I will assume that this is not related to this particular change. 

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.10:verify (verify) on project org.eclipse.ui.views.properties.tabbed: Execute ApiApplication failed

vogella:
Linux build issue reported here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2603

ptziegler:
> Linux builds seem to fail always unrelated to the actual change so I will assume that this is not related to this particular change.
> 
> org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.10:verify (verify) on project org.eclipse.ui.views.properties.tabbed: Execute ApiApplication failed

The builds really are flaky... 😦 

>  As you already into the code, please provide a separate PR for this adjustment.

https://github.com/eclipse-platform/eclipse.platform.ui/pull/2604

vogella:
The search test failure on Mac also seems unrelated, I checked them and I do not see the Wizard involved. Merging this one.

vogella:
@ptziegler thanks, works great in todays I-Build. I think this is a notable feature, can you add it to N&N 4.5? I think the repo for this is: https://github.com/eclipse-platform/www.eclipse.org-eclipse?tab=readme-ov-file

Issue #2603, opened 2024-12-09
Constant failing of the tycho-apitools-plugin for Linux in the GH actions


vogella:
Sorry if that has already been reported but I could not find a issue about it in this repo.

The Linux verification fails constantly for PR in this repo 

````
Error:  Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.10:verify (verify) on project org.eclipse.ui.views.properties.tabbed: Execute ApiApplication failed: InvocationTargetException: Component 'Tabbed Properties View' in the baseline 'workspace' is disposed [main]: Component was disposed here [Worker-2: Updating plug-in dependencies] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.10:verify (verify) on project org.eclipse.ui.views.properties.tabbed: Execute ApiApplication failed
````

@HannesWell @laeubi is this a known issue?

laeubi:
This also happen in Jenkins sometimes, it seems to be an timing issue, I now added the follwoing improvements / workarounds:

- https://github.com/eclipse-tycho/tycho/pull/4575
- https://github.com/eclipse-tycho/tycho/pull/4561

once they are available the failure should hopefully be gone, I tried to debug this a bit locally but due to it depends on the timing its quite hard to guess whats really going on.

So I added a breakpoint at the "disposed here" place and disposed the baseline manually as a test, then I see that after the second run the first breakpoint is never hit again and the retry actually works.

PR #2604, opened 2024-12-09
Remove special flattening-behavior for Wizard content providers

ptziegler:
This removes the feature from the "New" and "Import/Export" wizards, where container for wizard elements would be flattened, if it only contains a single, other container.

There are two reasons for this change:
- it is not consistent with the other viewers, that always show the complete tree path (e.g. the preference dialog).
- it doesn't work properly, as it is still possible to produce the undesirable behavior in the "New" wizard (for e.g. Java/JUnit/...)

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/2602 for context.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WizardContentProvider.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WizardContentProvider.java
@@ -46,11 +46,6 @@ public class WizardContentProvider implements ITreeContentProvider {
 				handleChild(childWizard, list);
 			}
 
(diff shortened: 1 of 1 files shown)


ptziegler:
This is an example of the flattening, where the intermediate `Test Import Parented Category` entry is skipped, making it look like `Test Import Add To Parent Category` is a direct child of `General`.
Screenshot_20241209_222353

This is an example using the _same_ filter, but in the `New` wizard. By the same logic, the `JUnit` entry should be skipped, even though it isn't...
Screenshot_20241209_222328

vogella:
Linux build failure seems unrelated, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/2603

PR #2606, opened 2024-12-10
Fixes to issue #2522 Enable Theming Checkbox

### LINKED ISSUE #2522: Enable theming preference check box is not working as expected and forces to restart twice

## Steps to reproduce

Install the setup of 'platform ui' by dragging the setup link in the banner of the installer : see [Setting your environment](https://github.com/eclipse-ide/.github/blob/main/CONTRIBUTING.md) and [projects guide](https://github.com/eclipse-ide/.github/blob/main/projects.md)

Open the preference dialog and filter on 'theme'. Disable theming if there is one selected by unchecking the 'Enable Theming' checkbox and restart. Eclipse restarts without any theme. 

Then reopen the preference dialog and filter on 'theme' then check the 'Enable theming' check box again and click Apply. 

Eclipse will ask to restart, while no Theme has been yet selected.. and no theme has been proposed.  The possible themes will appear after the restart and will be applicable, but we will have to restart again... this is not straightforward and totally counterintuitive.

Actually, when the 'Enable theming' is selected, it must display the themes and then propose to select one and restart when it is applied (even if the theme was not changed in the combo). 

This is the current situation when we switch from 'Enable theming' unchecked to checked:

| What we have  (unchecked to checked)          | What we expect       |
|-----------------------|----------------------|
| image |image |

On the other hand, if we unchecked the theme while it was checked, the list of themes should be hidden: 

| What we have (checked to unchecked)         | What we expect       |
|-----------------------|----------------------|
| image |image
 |

The Plugin Spy (Alt Shift F1), launched when the preference page is displayed says that this preference page is managed by the ViewsPreferencePage in the org.eclipse.ui.workbench plugin : 

image

A change event listener could be added on the preference checkbox editor so as to react on the change and display or not the bloc of theme list. 

* OS & version: macOS X 15.1 (sequoia).
* Eclipse IDE/Platform version (as shown in Help > About): Version: 2024-12 (4.34) Build id: I20241117-1800 (got from the platform.ui current setup)

## Community

mai-tran-03:
I added more functionality when the `enableTheming` button is selected. It would display the theme-dependent components if selected, otherwise hide them. We need to restart twice when theming is disabled.

Fixes [#2522 ](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2522)


IamLRBA (commented):
@mai-tran-03 Hi,
Thanks for the contribution, I am trying to understand your take on this issue. 
If possible, could you please share some screenshots showing if these changes tackle and rectify the issue when you reproduce it.
Thank You!

mai-tran-03:
@IamLRBA Hello,

The issue I'm trying to fix is when the user toggles the "Enable theming" button in Settings > General > Appearance, it should show/hide the selection of themes. It's counterintuitive to have to reset twice to apply the button changes.

image

My solution is to create separate functions to handle theme-dependent vs independent composites. In eclipse.ui.workbench/.../internal/dialogs/ViewsPreferencePage.java
> `createContents()`
1. `createEnableTheming()` - when the user first opens the preference settings page, it should display the above window with the current value of the enabled-theme key `themingEnabled` (on/off)

2. `createThemeDependentComposite()` - group together the elements, such as theme and color combo dropdown and tab icons, that depend on `themingEnabled` being selected; it will display/hide those elements when user toggles using `updateThemingEnablement()` (tacking on selectionListener to `themingEnabled` button)
- `createThemeIdCombo()` - move around this code (creating the theme dropdown menu) to encapsulate related functions

image

3. `createThemeIndependentComposite()` - group together the elements, such as buttons for round tabs, mixed font and color labels, and most recently used tabs, that don't depend on `themingEnabled` being selected

image

Below is a video demonstrating how my changes resolved part of the issue:
- `if engine == null` - the page only shows the basic buttons that do not rely on `themingEnabled`, that means disabling `themingEnabled` is the same as disabling `engine`
- The `engine` is needed to create the theme combo dropdown and other components that depend on the `engine` to be initialized
- Resetting twice: (1) initialize `engine` so it can `getThemes()`; and (2) apply any other new changes

[https://youtu.be/LdVrXZAM8zA?si=wCbBxKsdkJjCTSyg](url)

IamLRBA (commented):
Hey @mai-tran-03,
Well done, I'm interested in following what you are doing.
I see `themeDependentComp` was added but isn’t used yet. Should it affect UI behavior? If so, its integration might not be complete because i see you haven't initialized it yet.

mai-tran-03:
@IamLRBA Thank you for your feedback.
In my most recent commit [f241b1b](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2606/commits/f241b1be5e6a77f58193d15adc01f9cd70cc30a3), I removed the private global `themeDependentComp` variable, instead, I declared and initialized it in the `createThemeDependentComposite` method. So it is not referenced elsewhere in the class beyond that method.
- The variable is used inside `createThemeDependentComposite`, serving as a container for UI elements related to theme-dependent settings.
- It affects the UI, in which its visibility and layout depend on `updateThemingEnablement`, controlling whether the `themeDependentComp` section is visible based on the state of `themingEnabled`.

BeckerWdf:
@vogella: Can you provide some input why up to now these two restarts are necessary?

PR #2611, opened 2024-12-10
String intern org.eclipse.e4.ui.css.swt.dom.WidgetElement.swtStyles

jukzi:
To avoid needless duplicate strings in heap (only few styles are actually used).
image


akurtakov (approved):
This should help reducing the memory pressure.

jukzi:
ignoring unrelated flaky `NotifyingTreeViewer illegally extends TreeViewer`

vogella:
Always great to see Eclipse faster and leaner. Thanks @jukzi

PR #2618, opened 2024-12-11
ImportExistingArchiveProjectFilterTest: skip jdt #2432

jukzi:
When running locally and jdt plugins are available the test walked the whole contributed JDT items in the project explorer, which took much time and Heap memory

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2432

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/datatransfer/ImportExistingArchiveProjectFilterTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/datatransfer/ImportExistingArchiveProjectFilterTest.java
@@ -21,6 +21,7 @@ import java.util.Set;
 
 import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
@@ -154,7 +155,7 @@ public class ImportExistingArchiveProjectFilterTest extends UITestCase {
 		if (element instanceof IFolder) {
 			IFolder folder = (IFolder) element;
 			assertFalse(folder.getName().equalsIgnoreCase("res"));
-		} else {
+		} else if (element instanceof IResource) { // to expensive to walk other contributions like whole JRE from JDT
 			Object[] children = contentProvider.getChildren(element);
 			for (Object child : children) {
 				processElementAndChildren(child, contentProvider);


jukzi:
10s ->2s
image
image
1GB->500MB
image

image

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/datatransfer/ImportExistingArchiveProjectFilterTest.java
```diff
@@ -155,6 +155,9 @@ private void processElementAndChildren(Object element, ITreeContentProvider cont
 			IFolder folder = (IFolder) element;
 			assertFalse(folder.getName().equalsIgnoreCase("res"));
 		} else {
+			if (element.getClass().getName().startsWith("org.eclipse.jdt")) {
```
akurtakov: This is not optimal as e.g. org.eclipse.cdt can be installed on the machine and cause similar delay. Is it possible to "whitelist" rather than "blacklist" elements?

jukzi: i would prefer whitelist too. Just don't know what needs to be listed

akurtakov: A sane way to figure would be to create launch configuration with the minimum plugins that o.e.ui.tests requires and print the elements to get such a list.

merks: I'm a tiny bit confused.  This is in a test in the Platform's I-build.  In what case would CDT being installed be a problem?

merks: I see.  You mean when we launch from the IDE and we might have CDT in the target platform.  So this is not concern for the I-Build itself (and in the IDE we can always assign more memory, though being slow is not nice either there).

akurtakov: The change skips JDT contributions for the sake of speeding up tests, which is relevant for developers as the bundle itself doesn't require JDT thus tycho would not be sped up by this but only IDE runs of the test and developer can have any combination of plugins installed JDT, CDT, Mylyn, WildWeb,.... each of them slowing the test run. That's why I'm looking for whitelist (check only relevant aka Platform contributions or even smaller list) instead of blacklisting JDT.

jukzi: restricting IResource feels good and skips the same entries for me

PR #2619, opened 2024-12-11
Updating OpenCloseTest to JUnit5

vogella:
Adding jupiter API to MANIFEST.MF
Migrating OpenCloseTest to JUnit5

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/stress/OpenCloseTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/stress/OpenCloseTest.java
@@ -13,7 +13,8 @@
  *******************************************************************************/
 package org.eclipse.ui.tests.stress;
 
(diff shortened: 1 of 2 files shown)


vogella:
I assume that the build will fail as I now start to use JUnit5 for the first time in this plug-in @HeikoKlare please remind me what I have to do to make JUnit4 and JUnit5 co-exist.

akurtakov:
Having the test suite JUnit 3.x parts (extends TestCase) being converted to JUnit 4.x first would be nicer as I'm not looking for a mixture of JUnit 3/4/5 in the same test bundle.

HeikoKlare:
> I assume that the build will fail as I now start to use JUnit5 for the first time in this plug-in @HeikoKlare please remind me what I have to do to make JUnit4 and JUnit5 co-exist.

You would also need to migrate the test suites to JUnit 5 (at least those including the test migrated to JUnit 5). A JUnit 3/4 test suite will not be able to run a JUnit 5 test. Let me know if this is still relevant and if I should provide a PR for the test suite migration, @vogella.

Issue #2582, opened 2024-12-04
Align Text related icons according to UI guidelines


vogella:
After @BeckerWdf and @Michael5601 fixed the misaligned icons at the beginning of the main toolbar, the text icons starts to stand out as misaligned.

![image](https://github.com/user-attachments/assets/a8c198f9-9e09-4f21-82fe-8b73fd00ec5d)

![image](https://github.com/user-attachments/assets/e2f195d5-d83b-4b1c-93db-93f0d55c160d)
![image](https://github.com/user-attachments/assets/1d95da59-d452-492a-8c12-d065905a915c)
![image](https://github.com/user-attachments/assets/e791fd74-b098-4cca-b7d1-2e4f975a414a)

Would be great if these could also be aligned according to the current UI standard.

vogella:
@BeckerWdf can you help here?

BeckerWdf:
> @BeckerWdf can you help here?

Yes I can. I just need to find some time for it.

BeckerWdf:
@vogella: Is this issue now fixed?

vogella:
Thanks @BeckerWdf, it looks much better for me:

![image](https://github.com/user-attachments/assets/f0a8f046-7281-4f2d-bdd7-8b195194ddbc)

If my sense of order is not mistaken, the last icon looks still looks not centered to me (it is a bit to low AFAICS).

![image](https://github.com/user-attachments/assets/28c85d82-2379-425e-868d-4bff70f9782f)

Michael5601:
> Thanks @BeckerWdf, it looks much better for me:
> 
This is really a great improvement. I agree the last icon is still not right but the rest is fine.

BeckerWdf:
if you give me the name / location of the icon I can have a look at it.

Michael5601:
The icon is `pin_editor` in `org.eclipse.platform.doc.user\images` and `org.eclipse.ui\icons\full\etool16`. 

It also has duplicates in:

`org.eclipse.search\icons\full\elcl16\pin_view.svg`
`org.eclipse.jdt.ui\icons\full\elcl16\pin_view.svg`
`eclipse.platform.common\bundles\org.eclipse.jdt.doc.user\images\org.eclipse.jdt.ui\elcl16\pin_view.svg`

and maybe also `org.eclipse.ui\icons\full\elcl16\pin_view.svg` because it looks like a half version of the original icon. I don't know if this is wanted.

[...]

PR #2626, opened 2024-12-13
Fix Objects inside Icons

BeckerWdf:
Fix Icons according to Eclipse UI guideline.
Left and top pixel row should stay empty according to the guideline.

Also: make the "document" in the "wordwrap" and the "block_selection_mode" icon the same hight as they are shown next to each other in the toolbar.


BeckerWdf:
Contributes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2582

vogella:
LGTM

PR #2632, opened 2024-12-16
Run verification build with Java 21

akurtakov:
Prereq for
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2654


akurtakov:
This PR changes jenkins only so failures in gha could not be caused by this change.

Issue #2636, opened 2024-12-17
[performance] CloseUnrelatedProjectsAction.buildConnectedComponents(IProject[]) blocks UI startup


laeubi:
org.eclipse.ui.actions.CloseUnrelatedProjectsAction.buildConnectedComponents(IProject[]) calls the following code possibly in the UI thread:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/0344b2b062792a7336fe40edaefe768595e1d3f0/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java#L83

this leads to all classpath containers of all projects to be resolved what is a costly operation and blocks the whole IDE startup procedure see here:
- https://github.com/eclipse-pde/eclipse.pde/issues/1481

instead the computation should happen in the background and only the UI has to be updated once the operation succeeded.

vogella:
#3871 will be reverted with #3942 

This problem exists again. I do not plan to work on this again in the near future.

vogella:
Sampling the main thread of an Eclipse SDK every 500 ms during startup (115 samples,
Windows, Java 21) shows the UI thread BLOCKED for roughly 65 seconds, samples 26 to 102,
t=21.9s to t=86.5s. Full stack of the first blocked sample, trimmed at the bottom:

````
    "main" #1 BLOCKED (on object monitor)
      at org.eclipse.pde.internal.core.PluginModelManager.findModel(PluginModelManager.java:951)
      - waiting to lock <0x0000000605c90990> (a java.lang.Object)
      at org.eclipse.pde.internal.core.DynamicPluginProjectReferences.getDependentProjects(DynamicPluginProjectReferences.java:43)
      at org.eclipse.core.internal.resources.ProjectDescription.computeDynamicReferencesForProject(ProjectDescription.java:1037)
      at org.eclipse.core.internal.resources.ProjectDescription.getAllBuildConfigReferences(ProjectDescription.java:297)
      at org.eclipse.core.internal.resources.ProjectDescription.getAllReferences(ProjectDescription.java:252)
      at org.eclipse.core.internal.resources.Project.getReferencedProjects(Project.java:522)
      at org.eclipse.ui.actions.CloseUnrelatedProjectsAction.buildConnectedComponents(CloseUnrelatedProjectsAction.java:83)
[...]

PR #2638, opened 2024-12-19
Updated Code to Introduce Asynchronous Computation of Connected Compo…

IamLRBA:
# As per issue #2636 
The updated code introduces asynchronous computation of the connected components using a Job to avoid blocking the UI thread.

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java
@@ -23,23 +23,24 @@ import java.util.List;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
-import org.eclipse.core.resources.IResourceChangeEvent;
-import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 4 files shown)


mickaelistria:
Please remove the .idea/* files from the commit. The Eclipse Platform project usually doesn't maintain metadata for other IDEs

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java
```diff
@@ -66,37 +67,39 @@ public class CloseUnrelatedProjectsAction extends CloseResourceAction {
...
-				//each reference represents an edge in the project reference
-				//digraph from projects[i] -> references[j]
-				for (IProject reference : references) {
-					IProject setOne = set.findSet(project);
-					//note that referenced projects may not exist in the workspace
-					IProject setTwo = set.findSet(reference);
-					//these two projects are related, so join their sets
-					if (setOne != null && setTwo != null && setOne != setTwo) {
-						set.union(setOne, setTwo);
+	private static void buildConnectedComponentsInBackground(IProject[] projects, Runnable onComplete) {
```
laeubi, line 75:
```suggestion
	private static CompletableFuture<DisjointSet<IProject>> buildConnectedComponentsInBackground(IProject[] projects) {
```

laeubi (commented):
If you use a completable future to return the async computation and complete it with the result in the job then the caller can easily act on the result (e.g. execute something in UI or transform it further)

PR #2639, opened 2024-12-19
[Dark Theme] Fix for highlight issue in sub-tabs

mvm-sap:
With the dark theme changes, selected sub-tabs in the editor and views were missing blue highlights. This has been fixed with this PR. The screenshots below are from Windows. The same change has been incorporated in Mac and Linux.

Before:
Screenshot 2024-12-19 231102

After:
Screenshot 2024-12-19 225628

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/dark/e4-dark_tabstyle.css
+++ b/bundles/org.eclipse.ui.themes/css/dark/e4-dark_tabstyle.css
@@ -40,11 +40,10 @@ CTabFolder {
 CTabFolder[style~='SWT.DOWN'][style~='SWT.BOTTOM'] {
     /* Set the styles for the bottom inner tabs (Bug 430051): */
     swt-tab-renderer: url('bundleclass://org.eclipse.e4.ui.workbench.renderers.swt/org.eclipse.e4.ui.workbench.renderers.swt.CTabRendering');
-    swt-unselected-hot-tab-color-background: #161616; /* Bug 465711 */
+    swt-unselected-hot-tab-color-background: #161616;/*  Bug 465711 */
     swt-selected-tab-highlight: #316c9b;
     swt-selected-highlight-top: false;
 }
-
 
 CTabFolder.active {
     swt-selected-tab-highlight: #316c9b;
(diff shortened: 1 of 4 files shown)


BeckerWdf (changes requested):
I manually tested this. Works fine an fixes the bug.

I see that the changes are the same on all 3 platforms (CSS rules are identical).
Can't we then simply move this into the e4-dark_globalstyle.css only once instead of duplicating it three times?

mvm-sap:
> I manually tested this. Works fine an fixes the bug.
> 
> I see that the changes are the same on all 3 platforms (CSS rules are identical). Can't we then simply move this into the e4-dark_globalstyle.css only once instead of duplicating it three times?

Okay, I shall try to move and test again

BeckerWdf:
> I manually tested this. Works fine an fixes the bug.
> 
> I see that the changes are the same on all 3 platforms (CSS rules are identical). Can't we then simply move this into the e4-dark_globalstyle.css only once instead of duplicating it three times?

or in `e4-dark_tabstyle.css`

BeckerWdf:
The current state of the PR does not work correctly on macOS:
image

BeckerWdf (changes requested):
does not work on macOS

mvm-sap:
> does not work on macOS

I'm not sure why it isn't working. Shall I revert to a previous commit that made changes in all three platforms and worked fine? I don't have a MAC to test or debug, hence the problem.

BeckerWdf:
It's only the last change moving it to a common file that did not work. Your initial commit worked.

mvm-sap:
> It's only the last change moving it to a common file that did not work. Your initial commit worked.

Shall I revert it back to the changes done in initial commit ?

BeckerWdf:
> > It's only the last change moving it to a common file that did not work. Your initial commit worked.
> 
> Shall I revert it back to the changes done in initial commit ?

yes

mvm-sap:
> > > It's only the last change moving it to a common file that did not work. Your initial commit worked.
> > 
> > 
> > Shall I revert it back to the changes done in initial commit ?
> 
> yes

Could you please test again and confirm if its working?

BeckerWdf (approved):

BeckerWdf:
Works fine on macOS

PR #2643, opened 2024-12-20
Add all SVGs for Platform UI Icons 

Michael5601:
This PR adds SVGs for testing the new SVG feature. 

See this [PR](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593) for the feature.


HeikoKlare (commented):
I've just tested the SVG replacement with the recent addition of SVG support to SWT. To do so, I had to add the SVG fragment to the product. Note that this PR needs to be extended by adding the required capability for every bundle using SVGs in order to ensure that products using those bundles have a rasterized available.

I found that some icons are not rasterized in proper resolution but upscaled from 100%. When moving to a 175% monitor (from 100% primary monitor with monitor-specific scaling enabled), I got the following:
image

In particular, several icons from the `org.eclipse.search` bundle are not scaled properly in toolbars (probably it also affects others). May this require further changes such as the ones proposed in ...?
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593

Michael5601:
>Note that this PR needs to be extended by adding the required capability for every bundle using SVGs in order to ensure that products using those bundles have a rasterized available.
 
I updated the all manifests of bundles where I added SVGs with `Require-Capability`. 

> I found that some icons are not rasterized in proper resolution but upscaled from 100%. When moving to a 175% monitor (from 100% primary monitor with monitor-specific scaling enabled), I got the following:
> 
I tried loading the SVGs on 125% and 175% zoom respectively and it worked fine without blurry icons. Then I reproduced your setup with activated monitor-specific UI scaling and it also works fine. Can you please try again if the issue still persists and if so, can you maybe tell me how I can reproduce the problem?

> May this require further changes such as the ones proposed in ...?
> 
As as all the icons of `org.eclipse.search` are loaded with the URLImageFileNameProvider I am pretty sure the proposed changes in Platform UI have no impact here as they only adjust the URLImageDataProvider (and FileImageDescriptor).

HeikoKlare:
> I tried loading the SVGs on 125% and 175% zoom respectively and it worked fine without blurry icons. Then I reproduced your setup with activated monitor-specific UI scaling and it also works fine. Can you please try again if the issue still persists and if so, can you maybe tell me how I can reproduce the problem?

With the following setup, I can still reproduce it:
- Use monitor-specific scaling (which is now enabled by default anyway, so if you have an up-to-date IDE, that will be used automatically)
- Two monitors: primary monitor at 100% and secondary at 175%

When moving the windows to the secondary monitor, the marked icons are blurry.

HeikoKlare:
You can ignore my previous comment. I found the issue for the wrongly scaled SVGs: the Image implementation was erroneous. If the data for an image at the zoom level of the file (usually 100) was already loaded, this data is scaled instead of loading the file in proper zoom anew (as required for SVGs). That's easy to fix with a PR I provided today, so I added a fix that PR:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1987

Michael5601:
> You can ignore my previous comment. I found the issue for the wrongly scaled SVGs: the Image implementation was erroneous. If the data for an image at the zoom level of the file (usually 100) was already loaded, this data is scaled instead of loading the file in proper zoom anew (as required for SVGs). That's easy to fix with a PR I provided today, so I added a fix that PR:
> 
> * [Allow to test for loading image at zoom and fix faulty Win32 image load eclipse.platform.swt#1987](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1987)

Thats great, thank you! :)
I wanted to try it out on monday and I knew it must have to do with the 100% scaling as in my setup I tested moving from a 125% scaled monitor to 175% and thus not getting the blurry images. But after your comment I was pretty sure this was why I could not reproduce it.

PR #2647, opened 2024-12-24
[Light Theme]Fix for regression caused by defining colors for re-use

mvm-sap:
PR #2455 leads to regression in the behavior of highlight of selected tabs in inactive stack. This caused only in windows and Linux and has been fixed with this change.

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
+++ b/bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
@@ -190,7 +190,7 @@ CTabFolder Canvas {
 }
 
 .MPartStack{
-	swt-selected-tab-highlight: #a0a0a0;
+	swt-selected-tab-highlight: #8a8a8a;
 	swt-selected-highlight-top: false;
 }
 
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.ui.themes/css/e4_default_gtk.css
```diff
@@ -190,7 +190,7 @@ CTabFolder Canvas {
 }
 
 .MPartStack{
-	swt-selected-tab-highlight: #a0a0a0;
+	swt-selected-tab-highlight: #8a8a8a;
```
BeckerWdf, line 193: But in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2455/files#diff-c17f7415d0f31b4e2531cc20f515b87c405ce80099aa4730282ce6d27a59d0c8

I did not change this color: I did only change it from #A0A0A0 to #a0a0a0

Can you explain pls? What is bug already there before #2455 and not a regression of it?

mvm-sap: Yeah, looks like for linux this was not changed from #a0a0a0 to #8a8a8a before #2455. So, this is not a regression for linux.

BeckerWdf:
Failed test is unrelated.

PR #2650, opened 2024-12-30
Show "Arrow Symbols" instead up "UP" "DOWN" etc.

BeckerWdf:
On macOS we use arrow symbols (see
https://www.fileformat.info/info/unicode/char/2193/index.htm) when formatting key bindings (e.g. in tooltips or menu entries). This is much more compact then writing "UP", "DOWN", etc. and looks a lot nicer.

Let's also use this on Windows and Linux.


BeckerWdf:
On Windows key bindings that contain the cursor keys currently look like this:
Bild

On macOS they look like this:
image

So the use arrow symbols like "↑".

Native Apps on Windows use this:
image

So writing text instead of "↑" seems to be correct but if you see closely Eclipse writes "Left" and Visual Studio write "Arrow Left" which his even longer.

VSCode on Windows does this:
image
Yet another way "ArrowLeft".

Native GTK Apps do this:
image
Even thou I did not find a key binding with including curser keys I assume that they would also use "↑" instead of something like "Up" or "Arrow Up".

I would vote for changing this to always use the symbols like "↑" on all platforms. It looks nice and does not that that much screen space.

What do you think?

merks:
As long as they render nicely, that make good sense.  That's also what they look like on my keyboard.  😜

BeckerWdf:
> That's also what they look like on my keyboard. 😜

Haha. That's true.

BeckerWdf:
With this change windows looks like this:
Bild (1)

BeckerWdf:
can somebody on Linux test this PR and provide a screenshot?

tobiasmelcher (approved):
works as expected, thanks Matthias

tobiasmelcher:
screenshot from Ubuntu 22.04.5 LTS
ubuntu_keys

BeckerWdf:
> screenshot from Ubuntu 22.04.5 LTS 

Thanks for testing. This looks nice.

BeckerWdf:
If nobody objects I plan to merge this today EoB so that we can get feedback from users early.

BeckerWdf:
failed test is unrelated

PR #2660, opened 2025-01-03
support rendering of code minings at the very end of the document

### LINKED ISSUE #2157: Create a LineHeaderCodeMining at the end of the document
LineHeaderCodeMining allows only creation of code minings before a given line number, but not after.
This is needed if a code mining should be shown at the end of a document. 
I am giving a concrete example here:
screenshot_code_mining
This screenshot shows a code completion which is shown via two code minings. One AbstractCodeMining "ethods:" at the end  of the line where the caret is located and one LineHeaderCodeMining "add importing ..." with the rest of the completion at the end of the document. You can see that there is a whitespace character at line 6 at the very end of the document; this is the workaround I currently do so that the LineHeaderCodeMining can be rendered; before showing the code minings, the document is modified and a line break with an additional space is added at the end of the document so that rendering of the code minings is working. It would be great if this workaround would not be needed.

I would be happy to assist and provide a fix for this issue. But I would need some guidance how this could be achieved. Could you please give me some hints? There are currently two code mining classes CodeMiningLineHeaderAnnotation and CodeMiningLineContentAnnotation. Do we need a third class which represents a code mining at the end of the document (at offset >= document.getLength()) ? Or should CodeMiningLineHeaderAnnotation be extended so that it can also render a mining below a given line?

tobiasmelcher:
if the last line at the end of the document is empty

Fixes: #2157


mickaelistria:
Thanks!

PR #2664, opened 2025-01-06
Abort failing FindReplaceDocumentAdapter operations consistently #2657

### LINKED ISSUE #2657: FindReplaceOverlay/FindReplaceDialog: correcting regexp does not work
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2655#issuecomment-2568886267

having a java:
```
package a;

public class A {
	
}
public static void main2(String[] args) {
	
}
public static void main3(String[] args) {
	
}
}
```
1. open search overlay
2. enable regexp
3. enter "main"
4. press Search forward button
5. enter "main\n" in Replace Dialog
6. press Replace button -> does not work because of wrong line break (unituitive)
7. enter "main\r\n" in Replace Dialog
8.  press Replace button -> still does not work WHY?

HeikoKlare:
The implementation of FindReplaceDocumentAdapter#findReplace(), used by its API methods #find() and #replace(), currently leaves an inconsistent state when failing because of an invalid input or state. This inconsistent state affects the last executed operation type, which the adapter stores in order to identify whether a replace operation is preceded by an according find operation. In case the #findReplace() method is called with an invalid position or to execute a replace without a preceding find, the method aborts (throwing an exception) without storing the operation to be executed as the last executed operation, i.e., it leaves the adapter as if that method was never called. In case the method is called in regular expression mode and the expression used as find or replace input is invalid, the operation aborts (throwing an exception) but still stores the operation to be executed as the last executed operation, i.e., it leaves the adapter as if that method was called successfully.

This behavior is unexpected as is handles invalid inputs inconsistently. This also becomes visible in the existing consumers, such as FindReplaceTarget#replaceSelection() used by the FindReplaceDialog and FindReplaceOvleray. They assume that in case an exception occurs while trying to perform a replace operation, a subsequent replace should succeed without an additional find operation being required. This assumption does currently not hold.

This change corrects the behavior of the FindReplaceDocumentAdapter#findReplace() method to always leave the adapter in an unmodified state when the method fails because of being called with invalid input or in an inconsistent state. In addition, regular expressions with an unfinished character escape at the end now lead to a proper exception.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2657

### How it behaves

This is how it behaves after the fix
findreplace_replace_regex_fix

The inconsistent error indication (input colored red) is specific to the FindReplaceOverlay and will be fixed in a follow-up PR.

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/FindReplaceDocumentAdapter.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/FindReplaceDocumentAdapter.java
@@ -188,9 +188,6 @@ public class FindReplaceDocumentAdapter implements CharSequence {
 			}
 		}
 
-		// Set state
-		fFindReplaceState= operationCode;
-
 		if (operationCode == REPLACE || operationCode == REPLACE_FIND_NEXT) {
 			if (regExSearch) {
 				Pattern pattern= fFindReplaceMatcher.pattern();
@@ -199,7 +196,10 @@ public class FindReplaceDocumentAdapter implements CharSequence {
 					replaceText= interpretReplaceEscapes(replaceText, prevMatch);
 					Matcher replaceTextMatcher= pattern.matcher(prevMatch);
 					replaceText= replaceTextMatcher.replaceFirst(replaceText);
-				} catch (IndexOutOfBoundsException ex) {
+				} catch (IndexOutOfBoundsException | IllegalArgumentException ex) {
+					// These exceptions are thrown by Matcher#replaceFirst(), capturing information about
+					// invalid regular expression patterns, such as unfinished character escape sequences
+					// at the end of the pattern
 					throw new PatternSyntaxException(ex.getLocalizedMessage(), replaceText, -1);
 				}
 			}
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.text/src/org/eclipse/jface/text/FindReplaceDocumentAdapter.java
```diff
@@ -199,7 +196,7 @@ private IRegion findReplace(final FindReplaceOperationCode operationCode, int st
 					replaceText= interpretReplaceEscapes(replaceText, prevMatch);
 					Matcher replaceTextMatcher= pattern.matcher(prevMatch);
 					replaceText= replaceTextMatcher.replaceFirst(replaceText);
-				} catch (IndexOutOfBoundsException ex) {
+				} catch (IndexOutOfBoundsException | IllegalArgumentException ex) {
```
HeikoKlare, line 199: This is necessary to fulfill the intention of `interpretReplaceEscapes()` to leave an unfinished character escape at the end of the pattern as is (such as in the pattern `demo\`) in order to produce a proper error message (which is delivered through an `IllegalArgumentException` from `Matcher#replaceFirst()`).

jukzi: such comment should be put in the code

HeikoKlare: Indeed, I have added a comment to the code explaining why these exceptions are caught in f020fbe.

jukzi (approved):
lets try

PR #2670, opened 2025-01-07
Mark unused methods  for removal in WorkbenchWindowAdvisor

vogella:
These methods are not used since the 3.x migration, time to mark them for deletion.


vogella:
Took only approx. 2413 build tries to get a green build.....

merks:
> Took only approx. 2413 build tries to get a green build.....

Give or take 500.  😱  It's beyond annoying! 😭

PR #2672, opened 2025-01-07
Replace TextMatcher in favor of extended StringMatcher

ptziegler:
The functionality of the TextMatcher has been integrated in the StringMatcher class from the Equinox bundle, which allows this class to be used without a dependency to the Eclipse UI bundle.

Note: To get the same behavior as the TextMatcher, one needs to call matchWords() instead of match(). Furthermore, the pattern needs to be trimmed explicitly, where needed.

Contributes to
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2567


iloveeclipse:
> The functionality of the TextMatcher has been integrated in the StringMatcher class from the Equinox bundle

Please update bundle requirements to expect minimal version of equinox containing the new API

ptziegler:
> Please update bundle requirements to expect minimal version of equinox containing the new API

The Equinox bundle is re-exported by `org.eclipse.core.runtime`. So I assume you mean to adjust the version of the required bundle there and then bump the dependency to the runtime here?

iloveeclipse:
> The Equinox bundle is re-exported by `org.eclipse.core.runtime`. So I assume you mean to adjust the version of the required bundle there and then bump the dependency to the runtime here?

If core.runtime reexports new equinox API amd in this release didn't increased minor version segment, it has to bump it and after that the ui bundles that consume new API should update minimal range of core.runtime.

ptziegler:
I still haven't updated the lower bound for the dependency to org.eclipse.core.runtime, though...

vogella:
> I still haven't updated the lower bound for the dependency to org.eclipse.core.runtime, though...

Please push another PR for this.

iloveeclipse:
@vogella : please next time before merge read comments on PR.
@ptziegler : please add requested changes, thanks.

ptziegler:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2673

vogella:
https://github.com/eclipse-equinox/p2/pull/604

PR #2674, opened 2025-01-08
[StickyScrolling] Remove needless job to style sticky lines

Christopher-Hermann:
The styling is done in another Thread by asyncExe anyways, there is no need to do this in an Job. This overhead can be avoided.


tobiasmelcher (approved):

vogella:
I guess the old implementation followed the good old German "doppelt hält besser" approach.  ;-)

Thanks for this change, LGTM

PR #2676, opened 2025-01-08
AbstractDocument fix concurrency issues (BadPositionCategoryException)

jukzi:
By synchronized access to members used from different thread

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1067

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
@@ -345,12 +345,12 @@ public abstract class AbstractDocument implements IDocument, IDocumentExtension,
 
 		List<Position> list= fPositions.get(category);
 		if (list == null)
-			throw new BadPositionCategoryException();
+			throw new BadPositionCategoryException(category);
 		list.add(computeIndexInPositionList(list, position.offset), position);
 
 		List<Position> endPositions= fEndPositions.get(category);
 		if (endPositions == null)
-			throw new BadPositionCategoryException();
+			throw new BadPositionCategoryException(category);
 		endPositions.add(computeIndexInPositionList(endPositions, position.offset + position.length - 1, false), position);
 	}
 
@@ -514,7 +514,7 @@ public abstract class AbstractDocument implements IDocument, IDocumentExtension,
 
 		List<Position> c= fPositions.get(category);
 		if (c == null)
-			throw new BadPositionCategoryException();
+			throw new BadPositionCategoryException(category);
 
 		return computeIndexInPositionList(c, offset);
 	}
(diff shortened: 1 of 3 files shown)


jukzi:
i wonder if adding "synchronized" to the method is an API change as there is an error if overwriting methods do not add "synchronized" too

iloveeclipse:
> i wonder if adding "synchronized" to the method is an API change as there is an error if overwriting methods do not add "synchronized" too

There is SynchronizableDocument. I believe AbstractDocument was supposed not to be synchronized

laeubi (changes requested):
It looks wrong to synchronize on the document that then potentially uses a lock object. 
If synchronization is required, the lock object is needed to be set.

Also see comment here I think it is better solved on the consumer side:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1067#issuecomment-2577693789

bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
```diff
@@ -242,15 +243,6 @@ protected List<IDocumentPartitioningListener> getDocumentPartitioningListeners()
 		return asList(fDocumentPartitioningListeners);
 	}
 
-	/**
-	 * Returns all positions managed by the document grouped by category.
-	 *
-	 * @return the document's positions
-	 */
-	protected Map<String, List<Position>> getDocumentManagedPositions() {
```
iloveeclipse: I don't understand why API tooling doesn't complain here, but removing protected method from public class seems like API breakage for me.

jukzi:
What do you suggest?

Checking all usages seem to be impossible by the shear amount of usages.
One problem might be that SynchronizableDocument does not even offer/override `addPositionCategory` which will be the root cause to unsynchronized update the object. Another problem is that using the SynchronizableDocument will never reveal the stacktrace/thread which illegally called `addPositionCategory ` in another thread.

iloveeclipse:
> i wonder if adding "synchronized" to the method is an API change as there is an error if overwriting methods do not add "synchronized" too

See https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Evolving-Java-based-APIs-2.md

> Adding or removing the synchronized modifier also has a bearing on the method's behavior in a multi-threaded world, and may therefore raise a question of contract compatibility.

@jukzi : I'm not sure it is a good idea to change behavior of the basic class for documents with multiple inheritance, I see at least two clients that had extra synchronization enabled - SynchronizableDocument and XtextDocument. 

@szarnekow : FYI. XtextDocument uses own locking too.

jukzi:
I changed the commit to only synchronize SynchronizableDocument where needed/missing

laeubi (approved):
Looks fine to me

bundles/org.eclipse.text/src/org/eclipse/jface/text/AbstractDocument.java
iloveeclipse: Actually API tooling complained after explicit project build.

iloveeclipse:
Thanks, looks not so scary anymore.

jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform.ui/issues/1517

szarnekow:
> @szarnekow : FYI. XtextDocument uses own locking too.

Thanks for the ping. I was already reading the MR and it looks quite good. Nice catch @jukzi

PR #2681, opened 2025-01-10
Update org.eclipse.e4.ui.workbench.renderers.swt to Java 21

vogella:
For https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2739


vogella:
This might not be necessary, , it is only in which the pattern matching when the type is already of the tested one - that was not allowed in Java 17. Not sure how to find this with pattern matching.

PR #2684, opened 2025-01-10
AbstractTreeViewer do not auto-expand same nodes multiple times

jukzi:
fixes slow (Quick-)Type hierarchy for redundant superinterfaces

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1830

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -25,9 +25,11 @@ package org.eclipse.jface.viewers;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 import java.util.function.Function;
 
 import org.eclipse.core.runtime.Assert;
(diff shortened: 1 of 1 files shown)


jukzi:
@fedejeanne for the example from https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1922
image

fedejeanne:
I tested this in our big WS and it works like a charm: both the _Type Hierarchy_ and the _Quick Type Hierarchy_ of the problematic class (the "original `Broke1`") open in under **8s**! 🚀 👏

fedejeanne:
It just occurred to me: how does this PR affect other use cases where the user wants to expand the tree completely e.g. when he/she presses the button "Expand All"?
I'm away from the computer until Monday but I'd like to test this PR for more use cases before it gets merged. Can you please hold it, @jukzi ?

jukzi:
While i guess, all our views would benefit of it there may be views downstream that may be confused. While it does not make sense to have such "loops" in a tree, there may be for example both a quicklink (favorite) and the subtree itself in a view (like in windows file explorer). I think it doesn't make sense to expand both but still adopters may not expect it. So i suggest you take this PR as a proof of concept, enhance it to offer the behavior as API and let usages like jdt opt-in. Could you please provide such PRs?

jukzi:
to be replaced by https://github.com/eclipse-platform/eclipse.platform.ui/pull/2692

PR #2686, opened 2025-01-11
Add an example for providing an icon pack

laeubi:
There are some ongoing demands for providing custom icons to the Eclipse IDE or RCP applications.

This now adds an example of how such an "icon-pack" can be created and used inside Eclipse IDE.


vogella:
That is very cool. Thanks.

What would happen if we have multiple icon transformers defined? Last one to run wins?

laeubi:
> What would happen if we have multiple icon transformers defined? Last one to run wins?

If they try to transform the same resource one will win (but I haven't checked wich one) maybe the one with highest bundle id.

laeubi:
The build passes but we have heavy outage, merging to get some progress into this and it is only an example:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5511

PR #2688, opened 2025-01-13
Remove dispose call for Colors in StickyScrollingHandler

vogella:
According to the Javadoc on Colors:

 * Colors do not need to be disposed, however to maintain compatibility
 * with older code, disposing a Color is not an error.


vogella:
@Christopher-Hermann can you please have a look?

Christopher-Hermann (approved):

PR #2689, opened 2025-01-13
Correct activation of monitor-specific rescaling on Windows

HeikoKlare:
The activation of monitor-specific rescaling on Windows currently relies on the method Display#setRescalingAtRuntime(), which was found to not take the full intended effect.

This change thus replaces the call to that method based on the value of the according experimental preference by setting the according system property to activate that behavior in a proper way. It also removes the obsolete activation of Edge when the experimental preference is set, as Edge has been enabled by default.

Adapts to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1720

Requires https://github.com/eclipse-platform/eclipse.platform.swt/pull/1719

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
@@ -298,6 +298,8 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 	private static final String EDGE_USER_DATA_FOLDER = "org.eclipse.swt.internal.win32.Edge.userDataFolder"; //$NON-NLS-1$
 
+	private static final String SWT_RESCALE_AT_RUNTIME_PROPERTY = "swt.autoScale.updateOnRuntime"; //$NON-NLS-1$
+
 	private static final class StartupProgressBundleListener implements ServiceListener {
 
 		private final SubMonitor subMonitor;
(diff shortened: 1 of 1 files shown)


HeikoKlare:
@akoch-yatta

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -680,12 +680,16 @@ public void update() {
 		return returnCode[0];
 	}
 
-	private static void setRescaleAtRuntimePropertyFromPreference(final Display display) {
+	private static void setRescaleAtRuntimePropertyFromPreference() {
+		final String rescaleAtRuntimeProperty = "swt.autoScale.updateOnRuntime"; //$NON-NLS-1$
```
fedejeanne: I'd make this one a private constant and also use it in the `else`

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -680,12 +680,16 @@ public void update() {
...
 		boolean rescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
 				.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
```
fedejeanne: NIT: you could move this inside the `else`

fedejeanne (commented):
Looks good and works as expected, I only have 2 minor comments regarding coding practices.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
HeikoKlare: Done in 19e3364

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
HeikoKlare: Oh, yes, that was supposed to be used in the else block as well. Fixed with 19e3364.

akoch-yatta:
Looks good to me as well.

HeikoKlare:
Required SWT PR has been merged. Failing test is unrelated and documented: #867

PR #2690, opened 2025-01-14
Only evaluate expand condition once in AbstractTreeViewer

fedejeanne:
No need to evaluate for every child since the condition applies to the parent item.


fedejeanne:
Test failure is unrelated: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1517

PR #2692, opened 2025-01-14
Overload method AbstractTreeViewer::internalConditionalExpandToLevel

fedejeanne:
See @jukzi's comment: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2684#issuecomment-2586454515

## Is required by
- [ ] https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1932

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -28,6 +28,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.function.Predicate;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.IStatus;
@@ -139,6 +140,8 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	 */
 	private boolean isTreePathContentProvider = false;
 
+	private Predicate<Item> shouldItemExpand;
+
 	/**
 	 * Safe runnable used to update an item.
 	 */
@@ -166,6 +169,7 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	 */
 	protected AbstractTreeViewer() {
 		// do nothing
+		this.shouldItemExpand = createShouldItemExpand();
 	}
 
 	/**
@@ -1162,6 +1166,7 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 				control.setRedraw(false);
 			}
 			Widget w = internalExpand(elementOrTreePath, true);
+			shouldItemExpand = createShouldItemExpand();
 			if (w != null) {
 				internalExpandToLevel(w, level);
 			}
@@ -1170,6 +1175,17 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 				control.setRedraw(true);
 			}
 		}
+	}
+
+	/**
+	 * @return the predicate. It should return {@code true} if the received
+	 *         {@code Item} should be expanded when expanding the whole tree and
+	 *         {@code false} otherwise.
+	 *
+	 * @since 3.36
+	 */
+	protected Predicate<Item> createShouldItemExpand() {
+		return w -> true;
 	}
 
 	/**
@@ -1856,6 +1872,9 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 			createChildren(widget, false);
 			// XXX for performance widget should be expanded after expanding children:
 			if (widget instanceof Item it) {
+				if (!shouldItemExpand.test(it)) {
+					return;
+				}
 				setExpanded(it, true);
 			}
 			if (level == ALL_LEVELS || level > 1) {
--- a/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
+++ b/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.jface.tests.viewers;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
@@ -21,6 +22,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Predicate;
 
 import org.eclipse.jface.viewers.AbstractTreeViewer;
 import org.eclipse.jface.viewers.StructuredViewer;
@@ -28,6 +30,7 @@ import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Item;
 import org.eclipse.swt.widgets.Tree;
 import org.eclipse.swt.widgets.TreeItem;
 import org.eclipse.swt.widgets.Widget;
@@ -114,4 +117,47 @@ public class TreeViewerTest extends AbstractTreeViewerTest {
 		}
 	}
 
+	@Test
+	public void testExpandOnlyArbitraryChild() {
+
+		TestElement modelRoot = TestElement.createModel(3, 5);
+
+		// select an arbitrary child to be expanded
+		TestElement childToBeExpanded = modelRoot.getChildAt(modelRoot.getChildCount() - 1);
+
+		TreeViewer viewer = new TreeViewer(fShell) {
+			@Override
+			protected Predicate<Item> createShouldItemExpand() {
+				// Expand only "childToBeExpanded"
+				return it -> {
+					if (!(it.getData() instanceof TestElement)) {
+						throw new RuntimeException("Data: " + it.getData());
+					}
+
+					return it.getData() instanceof TestElement d && d.equals(childToBeExpanded);
+				};
+			}
+		};
+
+		viewer.setContentProvider(new TestModelContentProvider());
+		viewer.setInput(modelRoot);
+
+		// Even when expanding all elements, the provided predicate will take precedence
+		viewer.expandToLevel(AbstractTreeViewer.ALL_LEVELS);
+
+		Queue<TestElement> elements = new ConcurrentLinkedQueue<>(Arrays.asList(modelRoot.getChildren()));
+		while (!elements.isEmpty()) {
+			TestElement currentElement = elements.poll();
+
+			boolean shouldExpand = currentElement.equals(childToBeExpanded);
+
+			assertEquals("The expanded state of the element " + currentElement + " is wrong", shouldExpand,
+					viewer.getExpandedState(currentElement));
+			elements.addAll(Arrays.asList(currentElement.getChildren()));
+		}
+
+		// Double check that the desired element is there and that it was expanded
+		assertTrue("The element " + childToBeExpanded + " should have been expanded",
+				viewer.getExpandedState(childToBeExpanded));
+	}
 }


jukzi:
I think "internalConditionalExpandToLevel" is the wrong level of abstraction to make an API. Let alone the name "internal" ;-) 
May better offer some 
`Function<Widget, Boolean> getShouldChildrenExpand()`
or `boolean shouldExpandDuplicate()`
or `setExpandDuplicate(boolean)`
I don't know. but certainly not "Internal"Something.

fedejeanne:
Added `getShouldWigdetExpand()`

jukzi (dismissed):

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -1850,12 +1851,40 @@ protected Widget internalGetWidgetToSelect(Object elementOrTreePath) {
...
 			Function<Widget, Boolean> shouldChildrenExpand) {
+		internalConditionalExpandToLevel(widget, level, getShouldWigdetExpand(), shouldChildrenExpand);
+	}
+
+	/**
+	 * @return the predicate that determines if a widget should be expanded when
+	 *         expanding the tree.
+	 * @since 3.36
+	 */
+	protected Predicate<Widget> getShouldWigdetExpand() {
```
HeikoKlare:
```suggestion
	protected Predicate<Widget> getShouldWidgetExpand() {
```

HeikoKlare (commented):
Would it be possible to hold back this change until https://github.com/eclipse-platform/eclipse.platform.ui/pull/1072#pullrequestreview-2552406645 is resolved?
The method to be modified has been introduced last year, but we may need to change it as there was something like a breaking API change caused by it which we may need to revise.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
fedejeanne: Corrected, thank you!

fedejeanne:
> Would it be possible to hold back this change until [#1072 (review)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1072#pullrequestreview-2552406645) is resolved? The method to be modified has been introduced last year, but we may need to change it as there was something like a breaking API change caused by it which we may need to revise.

Hm, I'd like to have this one and https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1932 for M2.

Why not solve the other issue separately? The fix for [#1072 (review)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1072#pullrequestreview-2552406645) will have to expand widgets and its children regardless of _how_ to do it _i.e._ it can make (and should) make use of `getShouldWidgetExpand()`.

HeikoKlare:
> Hm, I'd like to have this one and [eclipse-jdt/eclipse.jdt.ui#1932](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1932) for M2.

I want to have a solution for the other issue soon (before M2), so it should not prevent this from being in before M2. It may just make the resolution more complicated as it further a method whose extract may have been invalid and needs to be reverted.

> Why not solve the other issue separately? The fix for [#1072 (review)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1072#pullrequestreview-2552406645) will have to expand widgets and its children regardless of _how_ to do it _i.e._ it can make (and should) make use of `getShouldWidgetExpand()`.

If you have a solution, it would be great if you can share. I did not think about the solution so far as I just got aware of the issue. To restore the original recursive behavior we may need to revert the method extraction, which will also affect this change.

fedejeanne:
I still don't see why the need of postponing this change. If you need to revert the other change then you still can integrate this PR into the other one.

Why pass the buck and make this PR adapt to other one? 

Also keep in mind that I don't only want to merge **this** PR, I also want the one in JDT so @jukzi would have to rush that one too which can lead to more errors.

HeikoKlare:
As I said: 
> To restore the original recursive behavior we may need to revert the method extraction, which will also affect this change.

We potentially have to replace the extracted method with an inlined (recursively called one) again, so you cannot simply add new parameters like you are doing in tihs PR.

> Why pass the buck and make this PR adapt to other one?

Because it might be possible that the other PR was _invalid_ and this one would be as well (for the same reason, an API breakage).

I don't understand the problem with just waiting for some investigation. I will try to do that asap (hopefully today, maybe tomorrow).

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -1850,12 +1851,40 @@ protected Widget internalGetWidgetToSelect(Object elementOrTreePath) {
 	 */
 	private void internalConditionalExpandToLevel(Widget widget, int level,
 			Function<Widget, Boolean> shouldChildrenExpand) {
+		internalConditionalExpandToLevel(widget, level, getShouldWidgetExpand(), shouldChildrenExpand);
```
HeikoKlare: Why do you need to pass the result of that new method as an argument at all? Wouldn't it be sufficient to just call that method from within the existing method?

jukzi: the map needs to be initialized only once

jukzi:
maybe you could agree on a setExpandDuplicate(boolean) which should be independent on how its implemented?

HeikoKlare:
> maybe you could agree on a setExpandDuplicate(boolean) which should be independent on how its implemented?

I currently have no strong opinion for or against any possible solution. This solution is also fine for me if will still be possible to apply it once the regression is fixed.

I have just created a PR with a potential fix of the regression: 
- #2707

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
HeikoKlare: You mean the map in the consumer (in JDT)? It is dangerous rely on this specific behavior of the `AbstractTreeViewer` implementation as that is an implementation detail that may easily be changed. There is no direct consumer (e.g., in the same bundle) relying on only executing the getter once per tree-expand operation and there is no specification or test ensuring it. So it will eventually be simplified in the future, having an impact on the consumer which relied on a specific property that is not part of the contract.

fedejeanne: The contract is similar to the one established by `shouldChildrenExpand`, which is already accepted and respected. There is no need to over complicate this.

jukzi: > y executing the getter once per tree-expand operation

I agree that a getter name is not intuitive for that. Such method should better be called "create"Something and the implementation should add a test that it is only called once.

fedejeanne: @HeikoKlare I'll try to rethink this PR to see if it can be adapted to yours. Any hints would be appreciated.

fedejeanne:
@HeikoKlare I just reviewed your PR and I think I can adapt this one to it. Let's process yours with a higher priority and then, once it's merged, I can adapt this one.

fedejeanne:
> maybe you could agree on a setExpandDuplicate(boolean) which should be independent on how its implemented?

The reasons I'm trying to avoid the usage of "duplicate" are :
1. The meaning of "duplication" is very specific. In your initial implementation (#2684, which used a `Set<Object>`) duplication was detected by relying on what was returned by `widget.getData()`, but this is also an implementation detail. 
2. The use case itself is also very specific. Maybe some consumers have no problem with expanding duplicated items but would like to keep the ones whose content starts with the letter "X" collapsed.

A solution that gives the chance to interrupt the  unfolding and leaves the details to the consumer looks more flexible to me, that's why I tried to implement something more general by providing the `Widget` itself and letting the consumer make the choice.

laeubi:
> The meaning of "duplication" is very specific. 

Because of this JFace has [ViewerComparator](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ViewerComparator.java) to compare two items. Also trees are allowed to contain "duplicates" of course. One major drawback and performance problem of trees is that it does not maintain a parent<->child relation ([TreeNodeContentProvider](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TreeNodeContentProvider.java) try to solve this but requires special consideration of the elements shown in a tree).

fedejeanne:
I've adapted this PR according to the changes in #2707. 

This PR is ready to be reviewed again.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
```diff
@@ -139,6 +140,8 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	 */
 	private boolean isTreePathContentProvider = false;
 
+	private Predicate<Item> shouldItemExpand;
```
HeikoKlare, line 143: If this is to be a field that has a meaning across method calls, it should, e.g., be filled during intialization. If this value is only relevant during a call of `expandToLevel` (or other) method (and thus reinitialized in every call of that method), it should not be a field, as it is specific to the execution time of that method.
You see the issues with this in the currently failing tests: there are other callers of `internalCustomizedExpandToLevel`, which accesses this field without it being initialized. 

So I see two options:
- This predicate needs to be created within the method it is valid in and passed through the methods called from there, if necessary
- This precidate may be initialized at object initialization with a proper invariant holding for it during the object lifecycle

fedejeanne: I initialized it in the constructor too to avoid NPEs.

The tricky part is passing it all the way down through the method calls until it reaches `internalCustomizedExpandToLevel` without adding an extra parameter (or overloading) `internalExpandToLevel(Widget widget, int level)`. Which is why I was only able to support the use cases where the call comes through `expandToLevel(...)`.

Do you see any better way to do it? I couldn't come up with any.

HeikoKlare: This still means you are abusing a field of class to actually pass a method parameter. Fields of classes need to have a proper (or at least a reasonable) invariant they fulfill, but this one actually has not reasonable invariant (apart from "state x regarding last execution of y") as it is actually a method parameter.

I understand that changing the method signature is not easy/possible, but in my opinion solving this by abusing a field as a method parameter is an error prone design flaw. If you need to reinitialize the `shouldItemExpand` in some of the methods calling `internalCustomizedExpandToLevel`, why not in the others? Is there potentially some state in the predicate delivered by `createShouldItemExpand`? Then that one will usually be invalid while executing `internalCustomizedExpandToLevel` without having it properly initialized.

Wouldn't it be possible to achieve the same without any extension to AbstractTreeViewer by simply implementing the according subtype with an overwrittten `expandToLevel`, adding the predicate creation at the beginning of the method, and then consuming it in an overwritten `setExpanded`?

fedejeanne: Consuming it in `setExpanded` won't work because the execution needs to be stopped in its consumer (`internalCustomizedExpandToLevel`) which can only be achieved by changing its return value to `boolean` (API change, since it's a protected method) and doing an early return.

I've been looking at other subclasses of `AbstractTreeViewer` and I noticed that they end up hacking its way into the expansion mechanism by overriding methods. I think this may be a consequence of the little flexibility that the class offers. How about adding hook 2 methods that will be called only **once** right before and after the expansion runs? This would give the consumer/subclasses the chance to cleanly preserve some state _outside_ `AbstractTreeViewer` and hopefully avoid future hacks.

fedejeanne:
Moot

PR #2693, opened 2025-01-14
Mark long deprecated methods in AbstractUIPlugin for removal

vogella:
These methods are deprecated since a long time and should be marked for removal. This will enable us to delete them after the API deprecation period.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/plugin/AbstractUIPlugin.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/plugin/AbstractUIPlugin.java
@@ -269,7 +269,7 @@ public abstract class AbstractUIPlugin extends Plugin {
 	 *
 	 * @return IWorkbench the workbench for this plug-in
 	 */
-	@Deprecated
+	@Deprecated(forRemoval = true, since = "2025-03")
 	public IWorkbench getWorkbench() {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingHandler.java
```diff
@@ -122,9 +122,7 @@ private StickyScrollingControlSettings loadControlSettings(IPreferenceStore stor
 		int stickyScrollingMaxCount= store.getInt(EDITOR_STICKY_SCROLLING_MAXIMUM_COUNT);
 
 		Color lineNumberColor= new Color(PreferenceConverter.getColor(store, EDITOR_LINE_NUMBER_RULER_COLOR));
-		sourceViewer.getTextWidget().addDisposeListener(e -> lineNumberColor.dispose());
```
merks: It's a little confusing to see this change mixed into a PR about AbstractUIPlugin.  Maybe this is accidental and meant to be in a different PR about not disposing colors?

vogella: Sorry, forgot to rebase onto latest master. The already merged change is not removed from this PR.

PR #2694, opened 2025-01-14
remove unused FastProgressReporter, BackgroundContentProvider.sortMon

jukzi:
FastProgressReporter was not public and deprecated and the only used implementation acted like a NullProgressMonitor


jukzi:
unrelated fail in PartRenderingEngineTests  https://github.com/eclipse-platform/eclipse.platform.ui/issues/751

PR #2695, opened 2025-01-14
deprecate DefaultUndoManager for removal

jukzi:
only referenced in org.eclipse.jdt.ui.tests.leaks.UndoManagerLeakTest
see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1936


vogella:
LGTM

Issue #2521, opened 2024-11-18
DataFormatException with badColor3


opcoach:
Let's make sure issue is not already fixed in latest builds first.

- [X] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

Install the setup of 'platform ui' by dragging the setup link in the banner of the installer : see   [Setting your environment](https://github.com/eclipse-ide/.github/blob/main/CONTRIBUTING.md)   and [projects guide](https://github.com/eclipse-ide/.github/blob/main/projects.md)

Set a breakpoint in StringConverter line 568 (on the throw DataFormatException of asRGB(String) ) : 

![image](https://github.com/user-attachments/assets/54bb79b4-5c77-462e-af05-9bbd152f9c7f)

Launch with all the plugins and hit the breakpoint. 

The value received is 0,0, 1x which is not compliant with the format expected by StringConverter. Should be 0,0,1 or something with 3 Integers. 

If you go up in the stack trace, you will see that it concerns the color definition 'badColor3'. 

## Tested under this environment:
* OS & version: macOS X 15.1 (sequoia). 
* Eclipse IDE/Platform version (as shown in *Help > About*): Version: 2024-12 (4.34) Build id: I20241117-1800 (got from the platform.ui current setup).

## Community

- [X] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

opcoach:
Here is the stack trace : 

!ENTRY org.eclipse.ui 2 2 2024-11-18 15:04:13.529
!MESSAGE Could not parse value for theme color badColor3
!STACK 0
org.eclipse.jface.resource.DataFormatException: For input string: "1x"
	at org.eclipse.jface.resource.StringConverter.asRGB(StringConverter.java:568)
	at org.eclipse.ui.themes.ColorUtil.getColorValue(ColorUtil.java:159)
	at org.eclipse.ui.internal.themes.ColorDefinition.getValue(ColorDefinition.java:111)
	at org.eclipse.ui.internal.themes.ThemeElementHelper.installColor(ThemeElementHelper.java:272)
[...]

PR #2696, opened 2025-01-14
Remove stacktrace from DataFormatException but add raw value #2521

jukzi:
looks less scary
image

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2521

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorDefinition.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorDefinition.java
@@ -14,11 +14,11 @@
 package org.eclipse.ui.internal.themes;
 
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.e4.ui.internal.css.swt.definition.IColorDefinitionOverridable;
 import org.eclipse.jface.resource.DataFormatException;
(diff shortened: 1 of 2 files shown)


laeubi (commented):
This seem to loose any way to find our *what* is wrong here...

jukzi:
> This seem to loose any way to find our _what_ is wrong here...

there is nothing wrong. if you wanted to debug the thing you still can as the error message is unique:
image

laeubi:
> there is nothing wrong. 

Why do we log a message if nothing is wrong? 

In your screenshot it says can't parse value but _why_? So I would expect some kind of number format exception, wrong values, missing comma, ....

jukzi:
> In your screenshot it says can't parse value but _why_? So I would expect some kind of number format exception, wrong values, missing comma, ....

??? that was not given before either. now you at least see the bad value in the logging.

laeubi:
It would of course help to see the full exception message/trace before.... from the mentioned codepath it could handle different cases and some might contain a useful message.

jukzi:
There are literally 0 bugs in bugzilla and github about any DataFormatException. so no one ever needed such stacktrace

laeubi:
> There are literally 0 bugs in bugzilla and github about any DataFormatException. so no one ever needed such stacktrace

Or everyone was able to fix the issue just from the stack-trace before ;-)

akurtakov:
ECA is failing for submitter now and there was disagreement how to handle this one. Closing.

PR #2699, opened 2025-01-14
Remove o.e.jdt.ui dependency from navigator tests

akurtakov:
Reduces cyclic dependencies between git repositories.


akurtakov:
[Jst]PipelinedTest is an area which I don't have knowledge about so I am giving up here.

PR #2700, opened 2025-01-14
Extract shared code of FilteredTrees to abstract base class

ptziegler:
This moves the duplicate code of the E3.x and E4.x based FilteredTree class and moves it into an AbstractFilteredViewerComposite class located in the JFace component.

The base class is currently only implementing a tree-based filter, but is structured in such a way that it can easily be extended to also support e.g. table-based filters.

Due to being unused, the following fields and methods have been marked as deprecated:
- filterToolBar
- clearButtonControl
- updateToolbar(boolean)

Contributes to
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2567


ptziegler:
This diff is already pretty big, so I think it's better to separate the refactoring from the FilteredTable implementation.

vogella:
LGTM, also passes all the tests, thanks @ptziegler

PR #2703, opened 2025-01-15
Do not use Java editor in back/forward test

akurtakov:
Test was using Generic and Java editors but is now switched to use the DefaultText and Generic editors in an effort to reduce cross repo dependencies.


akurtakov:
It's a pre-req for https://github.com/eclipse-platform/eclipse.platform.ui/pull/2699 . Semantically there should be no change for the test as it's about navigation between 2 different editors (no matter which ones).

PR #2704, opened 2025-01-15
Activate monitor-specific scaling before Display instantiation

HeikoKlare:
Monitor-specific scaling needs to be activated before the display that is supposed to run in that mode is created. Currently, the scaling is activated in the Workbench based on an according preference after the Display has already been created.

This change moves the evaluation of the experimental preference and the according initialization of monitor-specific scaling to before the Display is created inside the Workbench.
In addition, the disclaimer is cleaned up from the statement about Edge, which is now already been enabled by default.


akoch-yatta (approved):
I had a look at the change and from it looks solid and sound to me. Moving the initialization before the Display creation is definitely necessary.

HeikoKlare:
Failing test first occurred in this PR, but is completely unrelated to the change. It was documented here:
- #2708

PR #2705, opened 2025-01-15
Use LTK Refactoring when copying (duplicating) a Project.

N1k145:
Added missing plugin.properties

### DIFF
--- a/bundles/org.eclipse.ltk.ui.refactoring/plugin.properties
+++ b/bundles/org.eclipse.ltk.ui.refactoring/plugin.properties
@@ -29,3 +29,7 @@ renameResource.description=Rename the selected resource and notify LTK participa
 renameResource.commandParameter.newName=Selected resource's new name.
 
 contentMergeViewers.refactoring.label=Refactoring Compare
+
+copyProject.name=Copy Project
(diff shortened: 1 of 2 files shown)


N1k145:
@jukzi @mickaelistria here is the fix

bundles/org.eclipse.ltk.ui.refactoring/plugin.properties
```diff
@@ -29,3 +29,7 @@ renameResource.description=Rename the selected resource and notify LTK participa
 renameResource.commandParameter.newName=Selected resource's new name.
 
 contentMergeViewers.refactoring.label=Refactoring Compare
+
+copyProject.name = Copy Project
+copyProject.commandParameter.newName = Selected project's new name.
+copyProject.commandParameter.newLocation = Location of the new project.
```
mickaelistria: Would it be better to try to have both sentence follow the same pattern, eg "The (name|location) for the new project" ?

N1k145: yes, i adapted it

mickaelistria (approved):

jukzi:
thanks

PR #2707, opened 2025-01-15
Restore recursive behavior of AbstractTreeViwer#internalExpandToLevel()

HeikoKlare:
The method `AbstractTreeViewer#internalExpandToLevel()` used be a recursively called method (by implementation and specification). As a protected method, it is part of the API.
With a recent change (see #1072), the method was changed to delegate to a different method, removing it's recursive property. Consumers that have created subtypes of the AbstractTreeViewer and rely the recursive behavior of that method by overwriting it face a regression by that change.

This change restores the existing, recursive behavior of that method while still keeping the reuse of the implementation across other use cases. A test case ensuring the contractual recursive execution is added.

### DIFF
--- a/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
+++ b/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
@@ -16,6 +16,13 @@ package org.eclipse.jface.tests.viewers;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import org.eclipse.jface.viewers.AbstractTreeViewer;
 import org.eclipse.jface.viewers.StructuredViewer;
 import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.swt.SWT;
@@ -23,6 +30,7 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Tree;
 import org.eclipse.swt.widgets.TreeItem;
+import org.eclipse.swt.widgets.Widget;
 import org.junit.Test;
 
 public class TreeViewerTest extends AbstractTreeViewerTest {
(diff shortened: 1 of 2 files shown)


tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
```diff
@@ -77,4 +85,33 @@ public void testAutoExpandOnSingleChildThroughEvent() {
 				fTreeViewer.getExpandedState(trivialPathRoot.getFirstChild().getFirstChild()));
 	}
 
+	@Test
+	public void testInternalExpandToLevelRecursive() {
```
fedejeanne, line 89: I'd move this one to `AbstractTreeViewerTest`

fedejeanne (commented):
Looks pretty good but it reverts also the optimization made in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2690/. I would be nice if this optimization could be preserved.

I'll try to rethink my PR (https://github.com/eclipse-platform/eclipse.platform.ui/pull/2692) to see if it can be used together with this one.

BTW the test should be moved.

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
HeikoKlare: I understand the idea and agree that this should better be in a test class for `AbstractTreeViewer`. But the `AbstractTreeViewerTest` is actually not a test class for that class but an _abstract_ test class for applying the defined test cases to all _realizations_ of that class. Since the added test does not make sense to be applied for every subclass of `AbstractTreeViewer` (as it always does exactly the same), I placed it in `TreeViewerTest`. It would probably be better placed in a separate class performing tests on a mock subtype of `AbstractTreeViewer`, but that would need to be part of a larger refactoring of the tests.

fedejeanne: Uff, bad naming then. Ok, agreed 👍

HeikoKlare:
> Looks pretty good but it reverts also the optimization made in #2690. I would be nice if this optimization could be preserved.

I have adapted the PR to include the performance improvement (but putting more of the custom expansion functionality in the lambda). Can you have another look whether that properly reflects your improvement, @fedejeanne?

fedejeanne (approved):

fedejeanne:
Test failure is unrelated: https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #2710, opened 2025-01-16
Add ViewerComparator methods in INavigatorSorterService

akurtakov:
As every ViewerSorter is ViewerComparator the new default methods can be trivially added to implementors.
There are expected to be none as the interface is marked as noimplement and noextend.


akurtakov:
image

PR #2714, opened 2025-01-16
[Win] Activate monitor-specific scaling before Display instantiation

### LINKED ISSUE #2712: IDE can't start anymore without preselected workspace
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/2704.
Starting IDE without selected workspace (with prompt) breaks immediately with this error:
```
!SESSION 2025-01-16 08:18:01.840 -----------------------------------------------
eclipse.buildId=4.35.0.I20250115-1800
java.version=21.0.2-13
java.vendor=N/A
Command-line arguments:  -os linux -ws gtk -arch x86_64

!ENTRY org.eclipse.osgi 4 0 2025-01-16 08:18:03.638
!MESSAGE Application error
!STACK 1
java.lang.IllegalStateException: The instance data location has not been specified yet.
	at org.eclipse.core.internal.runtime.DataArea.assertLocationInitialized(DataArea.java:61)
	at org.eclipse.core.internal.runtime.DataArea.getStateLocation(DataArea.java:146)
	at org.eclipse.core.internal.preferences.InstancePreferences.getBaseLocation(InstancePreferences.java:43)
	at org.eclipse.core.internal.preferences.SingletonEclipsePreferences.initializeChildren(SingletonEclipsePreferences.java:97)
	at org.eclipse.core.internal.preferences.SingletonEclipsePreferences.<init>(SingletonEclipsePreferences.java:37)
	at org.eclipse.core.internal.preferences.InstancePreferences.<init>(InstancePreferences.java:57)
	at org.eclipse.core.internal.preferences.InstancePreferences.internalCreate(InstancePreferences.java:64)
	at org.eclipse.core.internal.preferences.EclipsePreferences.create(EclipsePreferences.java:364)
	at org.eclipse.core.internal.preferences.EclipsePreferences.create(EclipsePreferences.java:352)
	at org.eclipse.core.internal.preferences.PreferencesService.createNode(PreferencesService.java:354)
	at org.eclipse.core.internal.preferences.EclipsePreferences.getOrCreate(EclipsePreferences.java:1161)
	at org.eclipse.core.internal.preferences.RootPreferences.getNode(RootPreferences.java:65)
	at org.eclipse.core.internal.preferences.RootPreferences.node(RootPreferences.java:53)
	at org.eclipse.core.internal.preferences.AbstractScope.getNode(AbstractScope.java:38)
	at org.eclipse.ui.preferences.ScopedPreferenceStore.getStorePreferences(ScopedPreferenceStore.java:213)
	at org.eclipse.ui.preferences.ScopedPreferenceStore.<init>(ScopedPreferenceStore.java:127)
	at org.eclipse.ui.plugin.AbstractUIPlugin.getPreferenceStore(AbstractUIPlugin.java:254)
	at org.eclipse.ui.internal.UIPlugin$1.getPreferenceStore(UIPlugin.java:93)
	at org.eclipse.ui.internal.util.PrefUtil.getAPIPreferenceStore(PrefUtil.java:74)
	at org.eclipse.ui.internal.Workbench.setRescaleAtRuntimePropertyFromPreference(Workbench.java:690)
	at org.eclipse.ui.internal.Workbench.createDisplay(Workbench.java:760)
	at org.eclipse.ui.PlatformUI.createDisplay(PlatformUI.java:185)
	at org.eclipse.ui.internal.ide.application.IDEApplication.createDisplay(IDEApplication.java:208)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:151)

```

@HeikoKlare : please take a look, that must be fixed ASAP.

HeikoKlare:
Monitor-specific scaling needs to be activated before the display that is supposed to run in that mode is created. Currently, the scaling is activated in the Workbench based on an according preference after the Display has already been created.

This change moves the evaluation of the experimental preference and the according initialization of monitor-specific scaling to before the Display is created inside the Workbench. Since the preference is now accessed at a point in time where no workspace may already be available, the preference was moved to the ConfigurationScope so that it is stored independent from the actual workspace.

The drawback with this approach is that for an Eclipse application started from a workspace, where the configuration area will usually be cleared before every application start, the preference will be reset on every restart. Thus, for that scenario, the system property would currently have to be set explicitly via the VM arguments.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2712

Closes https://github.com/eclipse-platform/eclipse.platform.ui/pull/2713

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
@@ -81,6 +81,7 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.dynamichelpers.IExtensionTracker;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.preferences.ConfigurationScope;
 import org.eclipse.e4.core.contexts.ContextFunction;
 import org.eclipse.e4.core.contexts.ContextInjectionFactory;
 import org.eclipse.e4.core.contexts.IEclipseContext;
@@ -687,8 +688,8 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 					+ " is configured (e.g., via the INI), but the according preference should be preferred instead." //$NON-NLS-1$
 			));
 		} else {
-			boolean rescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
-					.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
+			boolean rescaleAtRuntime = ConfigurationScope.INSTANCE.getNode(WorkbenchPlugin.PI_WORKBENCH)
+					.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, false);
 			System.setProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY, Boolean.toString(rescaleAtRuntime));
 		}
 	}
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
@@ -39,6 +39,7 @@ import org.eclipse.core.runtime.IExtensionRegistry;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Platform.OS;
 import org.eclipse.core.runtime.RegistryFactory;
+import org.eclipse.core.runtime.preferences.ConfigurationScope;
 import org.eclipse.core.runtime.preferences.DefaultScope;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.core.runtime.preferences.InstanceScope;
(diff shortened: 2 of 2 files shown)


HeikoKlare:
@laeubi may I ask you to have a look at this? I had not so much to do with preferences yet, so there is probably a better / more "correct" way to manage preferences in the configuration scope.
This may supercede or be a follow-up to https://github.com/eclipse-platform/eclipse.platform.ui/pull/2713

laeubi:
> The drawback with this approach is that for an Eclipse application started from a workspace, where the configuration area will usually be cleared before every application start, the preference will be reset on every restart

I already mentioned this here:

- https://github.com/eclipse-equinox/equinox/issues/203

I think Equniox is sadly doing "too much" maybe here...

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -687,8 +688,8 @@ private static void setRescaleAtRuntimePropertyFromPreference() {
 					+ " is configured (e.g., via the INI), but the according preference should be preferred instead." //$NON-NLS-1$
 			));
 		} else {
-			boolean rescaleAtRuntime = PrefUtil.getAPIPreferenceStore()
-					.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME);
+			boolean rescaleAtRuntime = ConfigurationScope.INSTANCE.getNode(WorkbenchPlugin.PI_WORKBENCH)
+					.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, false);
```
laeubi, line 692: should this possibly take a system property into account for the default value?

HeikoKlare: Question would be which one to take into account. What the preference currently does is to set a system property which configures the HiDPI behavior of SWT. If you specify that system property manually (via INI/VM argument), this will be used anyway (in the if-block above) and the preference will be ignored.
So if we want to take a preference into account, we should probably just remove the if/else and integrate the system property value into this preference evaluation. The reason for not doing it that way was that we wanted to log a warning, as the interaction between system property and preference may be hard to understand, so setting a VM argument or INI flag to define it should be avoided in favor of simply using the preference.

HeikoKlare:
> I already mentioned this here:
> 
> * [clean whipe out all bundle data eclipse-equinox/equinox#203](https://github.com/eclipse-equinox/equinox/issues/203)
> 
> I think Equniox is sadly doing "too much" maybe here...

Interesting. Thank you for the pointer! I did not even question the behavior, I just wanted to mention that it will have some impact on configuring (and in particular testing) the monitor-specific scaling feature.

HeikoKlare:
Failing test is unrelated and documented: https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
HeikoKlare: So I would keep as is for now (it conforms to the existing behavior also defaulting to "false"). Additionally, this is currently only an experimental preference (which is reasonably set to "false" by default) to be replaced by a "proper" one once the feature is ready to be used productively and potentially activated by default.

PR #2722, opened 2025-01-17
Adapt double buffering use cases for multi-zoom

### LINKED ISSUE #138: UI Glitches in multi zoom environments for line numbers
In the LineNumberRuler the initialization of Image with the DPIUtil::nativeZoom seems to lead to issue when scrolling

Scenario to reproduce seems to be similar to the following:

- Multi zoom setup (eg. 100%, 200%)
- Move eveything to 100%
- Open a test editor with line numbers with enough text to contain vertical scrollbars
- Detach another view and move it to 200%
- Use the mouse and move the scrollbar up and down- 
https://github.com/user-attachments/assets/b5c4b56c-42ab-4ce4-a65c-c9e4c4e175e5

akoch-yatta:
This PR replaces three use cases where GC together with Images where used to draw double buffered images. With new multi-zoom-setting in Windows this leads to multiple issues like:
- destructive scaling, e.g. up- or downscaling of the pixel data
- fragments, when buffer image and second buffer image use different zoom 

The changes utilize the newly added ImageGcDrawer to provide a dynamic callback to draw on a correctly initialized GC on demand.

Most important changes are:
- cleanup of the buffer image on zoom change
- Always do a full draw when a new variant is requested

### How to test

1. You need to use the branch from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1734 for SWT. 
2. Either make sure you have the the experimental monitor specific flag enabled or use `
-Dswt.autoScale=quarter` as additional VM arguments

1. Have at least two monitors with different zoom (e.g. one with 100, one with 150), detach a view and move it to the second monitor. Then you should be able to create fragments or other glitches in the text editors e.g. by just resizing it -> resizing always retriggers a full re-creation.

#### Example for OverviewRuler and LineNumberRulerColumn
If you have a monitor setup with 200% and 100%, have both (main view and detached view) on 200%, move one to the 100% monitor and resize the view remaining on 200%. You will see:
Before it blurry scales the 100% variant of the image and blurry scales the line number
Screenshot 2025-01-20 110042
After it correctly fetches the 200% variant of the image and the line numbers are redrawn sharply
Screenshot 2025-01-20 105208

#### Example for not refreshed line numbers
Use the same setup as above: 200% and 100% monitor, have both (main view and detached view) on 200%, move one to the 100% monitor and resize the view remaining on 200%. You should see the line number not correctly refreshed on scrolling though a file anymore

**Important**: As it is quite visible I want to mention, that the DefaultRangeIndicator will still looks glitchy like 
image
in some scenarios. This is an additional issue related to ImageData and the destructive scaling of ImageData and will be fixed in a separate PR. This is a pre-existing issue, but the monitor aware scaling makes it way more obvious.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/AnnotationRulerColumn.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/AnnotationRulerColumn.java
@@ -266,6 +266,13 @@ public class AnnotationRulerColumn implements IVerticalRulerColumn, IVerticalRul
 			fCachedTextWidget= null;
 		});
 
+		fCanvas.addListener(SWT.ZoomChanged, e -> {
+			if (fBuffer != null) {
+				fBuffer.dispose();
+				fBuffer= null;
+			}
+		});
+
 		fMouseListener= new MouseListener() {
 			@Override
 			public void mouseUp(MouseEvent event) {
@@ -517,24 +524,28 @@ public class AnnotationRulerColumn implements IVerticalRulerColumn, IVerticalRul
 				fBuffer= null;
 			}
 		}
-		if (fBuffer == null)
-			fBuffer= new Image(fCanvas.getDisplay(), size.x, size.y);
-
-		GC gc= new GC(fBuffer);
-		gc.setFont(fCachedTextWidget.getFont());
-		try {
-			gc.setBackground(fCanvas.getBackground());
-			gc.fillRectangle(0, 0, size.x, size.y);
-
-			if (fCachedTextViewer instanceof ITextViewerExtension5)
-				doPaint1(gc);
-			else
-				doPaint(gc);
-		} finally {
-			gc.dispose();
+		if (fBuffer == null) {
+			fBuffer= new Image(fCanvas.getDisplay(), this::doPaint, size.x, size.y);
+		} else {
+			GC gc= new GC(fBuffer);
+			try {
+				doPaint(gc, size.x, size.y);
+			} finally {
+				gc.dispose();
+			}
 		}
-
 		dest.drawImage(fBuffer, 0, 0);
+	}
+
+	private void doPaint(GC gc, int width, int height) {
+		gc.setFont(fCachedTextWidget.getFont());
+		gc.setBackground(fCanvas.getBackground());
+		gc.fillRectangle(0, 0, width, height);
+
+		if (fCachedTextViewer instanceof ITextViewerExtension5)
+			doPaint1(gc);
+		else
+			doPaint(gc);
 	}
 
 	/**
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
@@ -34,6 +34,7 @@ import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.FontMetrics;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.ImageGcDrawer;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.widgets.Canvas;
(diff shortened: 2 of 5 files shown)


HeikoKlare:
Some minor additions on the PR:
- Testing the changes without the experimental preference for monitor-specific scaling on Windows, the behavior should be the exact same as before the PR. The mentioned remaining glitches will only occur when having that experimental feature enabled.
- Only derivation from the point above is the appearance on GTK when changing fractional scaling settings of the used monitors while running an Eclipse application. In this case, one should experience an _improvement_ of the apperance of the line number ruler, see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1734#pullrequestreview-2560425476

This is how the improvement looks on GTK when switching from 100% to 150% scaling:
#### Before the changes
image

#### After the changes
image

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/AnnotationRulerColumn.java
```diff
@@ -517,10 +525,18 @@ private void doubleBufferPaint(GC dest) {
 				fBuffer= null;
 			}
 		}
-		if (fBuffer == null)
-			fBuffer= new Image(fCanvas.getDisplay(), size.x, size.y);
+		if (fBuffer == null) {
+			ImageGcDrawer imageGcDrawer= (gc, imageWidth, imageHeight) -> {
+				doPaint(gc, new Point(imageWidth, imageHeight));
+			};
+			fBuffer= new Image(fCanvas.getDisplay(), imageGcDrawer, size.x, size.y);
```
HeikoKlare: When changing the signature of `doPaint` to accept width/height instead of size, this could be simplified to:
```suggestion
			fBuffer= new Image(fCanvas.getDisplay(), this::doPaint, size.x, size.y);
```

HeikoKlare (commented):
The change looks good to me. I have tested it on all platforms together with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1734 and found no issues, but I will also test again today with the final state of the merged SWT changes.

I only have a minor comment on a potential code simplification.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/AnnotationRulerColumn.java
akoch-yatta: Nice improvement, I adapted the affected places

fedejeanne (approved):
Thank you for adding the use cases to test @akoch-yatta .

I tested (in Windows) and found no regressions and also that the 2 mentioned use cases were fixed by this PR. I couldn't find any more glitches,

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/VerticalRuler.java
```diff
@@ -223,27 +230,30 @@ private void doubleBufferPaint(GC dest) {
 				fBuffer= null;
 			}
 		}
-		if (fBuffer == null)
-			fBuffer= new Image(fCanvas.getDisplay(), size.x, size.y);
-
-		GC gc= new GC(fBuffer);
-		gc.setFont(fTextViewer.getTextWidget().getFont());
```
HeikoKlare, line 230: If I don't miss anything, this line of code got lost in the refactoring.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
```diff
@@ -681,8 +688,17 @@ private void doubleBufferPaint(GC dest) {
...
+				ILineRange lines= JFaceTextUtil.getVisibleModelLines(fCachedTextViewer);
+				if (lines == null) {
+					return;
+				}
+				// We redraw everything; paint directly into the buffer
+				initializeGC(gc, 0, 0, imageWidth, imageHeight);
+				doPaint(gc, lines);
+			};
+			fBuffer= new Image(fCanvas.getDisplay(), imageGcDrawer, size.x, size.y);
+		} else {
```
HeikoKlare, line 691: The indentation of the following code is broken by now enclosing it into an else block. Could we maybe separate the if and else case into distinct methods to be called here? That would, in my opinion, make the code cleaner and, as a side effect, preserve the indentation, so that we do not have unnecessary indentation changes.

akoch-yatta: I extracted it into two sub method

HeikoKlare (approved):

Issue #2651, opened 2025-01-02
Find/Replace Overlay: "Content Assist Available" ctrl+space does not work in .java files


jukzi:
When enabling regular expression i get a hint that ctrl+space would open content assist but it only adds a space
![image](https://github.com/user-attachments/assets/6714eeb1-57ab-4cc6-9a9d-ccc560d58c90)

jukzi:
windows, I20241230-1800

HeikoKlare:
That's interesting. The issue is not reproducible for me, it works as expected:
![image](https://github.com/user-attachments/assets/4f6707a8-7eab-4332-b058-5f6ce007faa7)

jukzi:
for .xml files it works for me too but not with .java source code

HeikoKlare:
Thank you, I missed the specialization to .java files. I can indeed reproduce the issue there. Seems to be related to how key bindings work / are disabled in (Java) editors. It's a rather complicated / hacky solution to support such an overlay for an editor.

jukzi:
Please provide what ever API is needed for editors like jdt and disable the hint when it is disabled by the editor

HeikoKlare:
Added this to the umbrella issue:
- #2021

PR #2724, opened 2025-01-17
Avoid FindReplaceOverlay setTextEditorActionsActivated resulting in NPE

merks:
- The FindReplaceOverlay's setTextEditorActionsActivated method reflectively invokes AbstractTextEditor.setActionActivation(boolean) and that method assumes that getEditorSite().getActionBarContributor() returns non-null but MultiPageEditorSite.getActionBarContributor() returns null so in PDE's target editor, this leads to an NPE.
- AbstractTextEditor.setActionActivation(boolean) must guard against a null action contributor to avoid NPE
- The logic in setTextEditorActionsActivated already has special case handling for MultiPageEditorSite which should be augmented to call IEditorActionBarContributor.setActiveEditor(IEditorPart) on the overall multi-page editor.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -188,6 +188,7 @@ import org.eclipse.jface.text.source.SourceViewerConfiguration;
 import org.eclipse.jface.text.source.VerticalRuler;
 
 import org.eclipse.ui.IActionBars;
+import org.eclipse.ui.IEditorActionBarContributor;
 import org.eclipse.ui.IEditorDescriptor;
 import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IEditorPart;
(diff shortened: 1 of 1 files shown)


merks:
To reproduce the original problem open a *.target file with the PDE's Target Editor, turn to the Source tab and type Ctrl-F.  I didn't even notice the logged errors initially, but I did notice that Ctrl-C didn't paste into the overlay but rather into the editor (and then I noticed that Revert didn't work for that editor, which is now fixed).  

Anyway, I don't know if this is the right approach for fixing this.  Also, given the careful order of actions in AbstractTextEditor.setActionActivation(boolean), the current change does not preserve that order/inverse-order for the setActiveEditor call and the processing of the actions.  I think more refactoring would be needed if that order is important.

Please feel free to reuse this PR in any way, shape, or form,  or to rewrite it.

merks:
This solution is simpler and I think more correct.

HeikoKlare (approved):
Thank you, @merks!

I have tested your change and it works for me as expected. Given that we accept the hack of reflectively calling `setActionActivation(boolean)`, I agree that this is probably the right thing to do. I do not see a risk of regression, as the only other "user" (in terms of reflective access) of the changed method in addition to the FindReplaceOverlay is JDT's breadcrumb for Java editors. Since they are usually not placed inside MultiPageEditors, they will not be affected at all. And for the FindReplaceOverlay we already have some special handling for other kinds of MultiPageEditors.

Interestingly, PDE's MultiPageEditor for bundle information (MANIFEST, build.properties etc.) and this one for target files seem to behave differently, as they require different mechanism to temporarily deactivate the editor's actions.

This kind of hacking into the action activation is rather unsatisfying, as it is difficult to make and keep it work properly (see also https://github.com/eclipse-platform/eclipse.platform.ui/issues/2651 for a related issue). From my understanding, this is all rooted in the fact that such addition/overlay implementations need their own action scope but as they do not participate in the ordinary workbench/part model handling the action activation, it required custom mechanisms.

fedejeanne (commented):
@merks I'm not really familiar with this code, Heiko supervised this feature and knows the code.

merks:
The failing test:

image

and the new warnings:

image

are not related to this change.

PR #2730, opened 2025-01-18
Stop playing with ENABLE_ANIMATIONS

akurtakov:
It had no effect for years so this change simply removes leftovers.


akurtakov:
Test failures unrelated and Jenkins is green.

PR #2732, opened 2025-01-19
Mark AnimatorFactory(and friends) for removal

akurtakov:
These haven't been used at all since Eclipse 3.3.


akurtakov:
Test failure could not be caused by deprecations.

PR #2733, opened 2025-01-19
Un-deprecate ListSelectionDialog constructor

akurtakov:
Deprecation has been done when Builder was added which seems to be a trend in the Java world.
Undeprecation is done now as :
* it's one of the very few (if there are any others) APIs that "mandate" going through a builder, which makes it inconsistent with rest
* documentation hasn't been adjusted accordingly (even the class level javadoc has example using the deprecated constructor)
* usages in the SDK hasn't been migrated to the builder approach (for years) thus
generating a number of deprecation warnings

Deprecation could be added again whenever the previous issues are resolved.


akurtakov:
Failing test unrelated to un-deprecation of API.

PR #2736, opened 2025-01-19
Add static method to the interface of ISharedImages to get the instance

laeubi:
Currently one needs to first obtain the PaltformUI Workbench and then call getSharedImages on this instance. As this actually maps down to WorkBenchPlugin that do not require a running workbench this limits the usage of such code only wanting to share images.

This adds a new ISharedImages.get() that directly calls the code the IWorkbench implementation would call and replace static calls to the workbench with this one.

See for example:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1952


akurtakov (approved):
Looks good to me to have more direct access.

Issue #2740, opened 2025-01-21
Regression: Line Numbers are broken in different ways on Linux/MacOS


BeckerWdf:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

I tried
* Open a Java Class and scroll up and down slowly

I expected: line number to be sharp

But got: partly blurry line numbers at the top or the bottom. See

<img width="234" alt="Image" src="https://github.com/user-attachments/assets/cf709f20-9566-46e2-88cc-c2fa1c2c884d" />

I think this is a regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/2722

## Tested under this environment:
* OS & version: macOS
* Eclipse IDE/Platform version (as shown in *Help > About*): newest i-build

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

BeckerWdf:
@akoch-yatta and @HeikoKlare can you please have a look?

akoch-yatta:
We have a look

iloveeclipse:
OMG. I see on Linux line numbers completely broken after scrolling up/down few times:

![Image](https://github.com/user-attachments/assets/ccd6a720-dea2-4acd-a49d-ac7cd7735506)

This must be fixed ASAP.

HeikoKlare:
Looks to me as if it might got worse with the recent changes, but it seems like this is long broken on GTK. I just tested with a 1,5-year-old SDK and got this:
Version: 2023-06 (4.28)
Build id: I20230605-0440

![Image](https://github.com/user-attachments/assets/55f29fd2-00be-43dd-a1a2-76269d5c919b)

iloveeclipse:
Whatever issues your I20230605-0440 version has is not related to recent regression. It is not the first time the line number ruler got a regression.

It works fine with reverted https://github.com/eclipse-platform/eclipse.platform.ui/pull/2722/commits/57666c4508c2fd48a9ceb29c0d4745cf5654eec6 and is broken with current master.

[...]

PR #2741, opened 2025-01-21
Possible fix for #2740

akoch-yatta:
Improves the behavior but doesn't seem to fully solve the problem.


iloveeclipse:
> Improves the behavior but doesn't seem to fully solve the problem.

Not sure what it improves, still see the problem of line numbers in blocks seemingly randomly repeated or same line number repeated, so it shows same one line for every line.

akoch-yatta:
This PR doesn't work, so I'll close it

PR #2742, opened 2025-01-21
Temporary workaround for LineNumberRulerColumn regression on Linux

akoch-yatta:
Having a debug session together with @HeikoKlare we got some insights about the underlying issue.

This PR is currently proposing having a platform-depending behavior for creating the main buffer image. Since we could identify the underlying issue in the Image implementation of GTK, we partially revert to the old implementation of the rule on Linux/GTK while using the updated one for Windows and MacOS. What this PR does in more detail:

#### Windows
It still uses the same logic as the initial PR. Reason: the line number ruler is not useable in the windows implementation with the monitor specific scaling enabled. As the previous implementation relied internally on the DPIUtil::getDeviceZoom we need a dynamic image creation for Windows. For that reason we really would like to have this logic in M2 for further evaluation

#### Mac OS
We have no issues with the changes from the PR to the ruler after https://github.com/eclipse-platform/eclipse.platform.swt/pull/1751, so we stick to the state after the PR.

#### GTK
We are running into the issue https://bugs.eclipse.org/bugs/show_bug.cgi?id=571166 again, because we are now using another init method as before, which uses `Cairo.cairo_image_surface_create` that is known to produce problems in the usage context of the line number ruler. We didn't have a proper solution until now to adapt the logic (inside the Image implementation for GTK) to adress this issue, but are working on it for M3.

Question from us is. Would this workaround acceptable as temporary "fix" for M2 or would we need to revert the whole LineNumberRuler changes, accepting the still broken ruler in windows with the monitor specific scaling active

<s>@HeikoKlare please extend on my explanation if you think something is missing</s>

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
@@ -42,6 +42,8 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 
+import org.eclipse.core.runtime.Platform.OS;
+
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.IRegion;
@@ -687,24 +689,25 @@ public class LineNumberRulerColumn implements IVerticalRulerColumn {
 		}
 
 		if (fBuffer == null) {
-			fBuffer= newFullBufferImage(size);
+			newFullBufferImage(visibleLines, size);
 		} else {
 			doPaint(visibleLines, size);
 		}
 		dest.drawImage(fBuffer, 0, 0);
 	}
 
-	private Image newFullBufferImage(Point size) {
-		ImageGcDrawer imageGcDrawer= (gc, imageWidth, imageHeight) -> {
-			ILineRange lines= JFaceTextUtil.getVisibleModelLines(fCachedTextViewer);
-			if (lines == null) {
-				return;
-			}
-			// We redraw everything; paint directly into the buffer
-			initializeGC(gc, 0, 0, imageWidth, imageHeight);
-			doPaint(gc, lines);
-		};
-		return new Image(fCanvas.getDisplay(), imageGcDrawer, size.x, size.y);
+	private void newFullBufferImage(ILineRange visibleLines, Point size) {
+		if (OS.isLinux()) {
+			fBuffer= new Image(fCanvas.getDisplay(), size.x, size.y);
+			doPaint(visibleLines, size);
+		} else {
+			ImageGcDrawer imageGcDrawer= (gc, imageWidth, imageHeight) -> {
+				// We redraw everything; paint directly into the buffer
+				initializeGC(gc, 0, 0, imageWidth, imageHeight);
+				doPaint(gc, visibleLines);
+			};
+			fBuffer= new Image(fCanvas.getDisplay(), imageGcDrawer, size.x, size.y);
+		}
 	}
 
 	private void doPaint(ILineRange visibleLines, Point size) {


iloveeclipse:
> Question from us is. Would this workaround acceptable as temporary "fix" for M2

Yes.

iloveeclipse:
And the patch works for me on RHEL 9.2

iloveeclipse:
Is this ready for review now?

HeikoKlare:
Yes, this should be ready for review (correct, @akoch-yatta?). I have updated the workaround to only be applied on Linux, as the reported MacOS bug had an independent cause that was properly fixed with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1751.

I have tested this workaround on Ubuntu 22.04 with X11 and found no issues with the line numbers so far. But I plan to also test with more complex monitor setups with different monitor scalings to ensure that are no regressions occur in such setups with the workaround as well.

akoch-yatta:
Yes, we are not able to find a proper fix for the GTK issue in SWT **today for M2**, so this workaround would be ready to review from my side.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
```diff
@@ -687,24 +689,27 @@ private void doubleBufferPaint(GC dest) {
...
 		} else {
 			doPaint(visibleLines, size);
 		}
 		dest.drawImage(fBuffer, 0, 0);
 	}
 
-	private Image newFullBufferImage(Point size) {
-		ImageGcDrawer imageGcDrawer= (gc, imageWidth, imageHeight) -> {
-			ILineRange lines= JFaceTextUtil.getVisibleModelLines(fCachedTextViewer);
-			if (lines == null) {
```
iloveeclipse, line 700: Why is this removed? both `doPaint()` variants will run into NPE if lines == null

iloveeclipse (commented):
Seem to work on GTK, I've tried various use cases including git blame & compare editors (known to uncover issues in line ruler). However see my comment on removed null check.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
HeikoKlare: That check should be obsolete as this method will only be called in case the visible lines are not "null". This is already evaluated in the caller (line 687).

iloveeclipse: The problem with this indirection is that if you open call hierarchy to understand who passes what you will see the call here that passes the data without checks, so you either assume it is not checked or you have to go one level deeper etc. Also next maintainer who will look into will assume it is safe not to check the result because it wasn't checked here.

What about to avoid second (expensive) call to `getVisibleModelLines()` and pass the visible lines from the caller to `newFullBufferImage()` instead of calling it yet another time there?

HeikoKlare: Sure, makes to only retrieve the value once anyway (both from a conceptual/comprehensibility as well as from a performance perspective). I adapted the PR to just retrieve the value once and pass it to the method in 4aefad1.

iloveeclipse (approved):

HeikoKlare:
I tested with different scaling settings now and also found no issues there. The only drawback is that the improvement of the line number ruler regarding scaling on GTK (reported in the original PR https://github.com/eclipse-platform/eclipse.platform.ui/pull/2722#issuecomment-2601696198) is also reverted by this.

iloveeclipse:
> The only drawback is that the improvement of the line number ruler regarding scaling on GTK (reported in the original PR [#2722 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2722#issuecomment-2601696198)) is also reverted by this.

Well, that still could be contributed by someone :-)

My point is that at least basic things should **always** work, so that one could use every nightly build in production (surely I do not propose it for everyone but I personally use every nightly), there will be always some corner cases that need extra time but that is OK.

HeikoKlare:
> My point is that at least basic things should **always** work, so that one could use every nightly build in production (surely I do not propose it for everyone but I personally use every nightly), there will be always some corner cases that need extra time but that is OK.

Absolutely. No change should introduce regressions regarding existing, in particular essential functionality. So it was out of question that we either have to fully revert the change to the line number ruler or find a fix for it soon. I am still not sure why we did not experience or perceive the issue when originally testing the PR, as we did testing for all three OS. Anyway, since we only have have positive feedback on this workaround, let's merge this for M2.

The improvement of HiDPI behavior for the line number ruler on GTK with the original PR was not even intended, but rather a side effect of the improvements for Windows :-)
Since we need to find a solution for the issue arising from the usage of `Cairo.cairo_image_surface_create` in Image to get rid of this workaround, the HiDPI improvement will then implicitly be restored.

iloveeclipse:
> I am still not sure why we did not experience or perceive the issue when originally testing the PR, as we did testing for all three OS

Because you didn't **used it for work** where you **need** correct line numbers. That's the reason I use every nightly for daily work.

HeikoKlare:
That's true. And that's always the difficulty / risk with changes relying on consistent cross-platform behavior, as there are places where the behavior (usually of SWT) is different across the platforms (like here). I usually work on Windows (also on the nightly version for daily work) and only use MacOS and Linux for cross-platform implementation / testing.

PR #2745, opened 2025-01-23
Add dependency check workflow

laeubi:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/.github/workflows#checkdependenciesyml


laeubi:
Verification do not test this

PR #2752, opened 2025-01-23
Properly sort marker resolution by relevance

laeubi:
Currently the marker resolution is sorted, but markers with the *lowest* relevance are show on top, and therefore these are selected by default. This is confusing and gives a bad user experience, also this is the reverse order than we use when offering the quickfix in the editor itself.

This now sorts the list that the marker with the highest relevance is on top and selected by default.


jukzi:
OK, regression from f13fb4997f5e7849da26285938a3a0aea1769747
however adding so many lines just to force your favorite style is questionable

laeubi:
The "new style" was hardly readable, understandable and debugable and introduced bugs ... So yes I always prefer "more lines" if it makes the code more clear because code is written once and read often.

Issue #2313, opened 2024-09-23
Light (Preview) Theme: Issues in Progress View


BeckerWdf:
- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

* Choose "Light (Preview)"

I tried

* Open the Progress View
* Build your complete workspace

But got: <img width="1378" alt="Screenshot 2024-09-23 at 08 34 57" src="https://github.com/user-attachments/assets/c5505b96-b88f-40ae-b727-37692d5273c8">

With the "Light" Theme this looks like this:
<img width="1045" alt="Screenshot 2024-09-23 at 08 35 41" src="https://github.com/user-attachments/assets/c00ad822-419c-4b1a-b5de-7164b396cb26">

I expected that the texts like "Building" don't have a white background but simply use the same background as the parent (the light grey color)

## Tested under this environment:
* OS & version: macOS

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

@mvm-sap: Can you pls. have a look?

mai-tran-03:
I'm working on this.
- Team Mai Tran/Andy Espinoza/Mahdi Alsalami (students), and Akshatha Anantharamu (mentor)

HannesWell:
Great. You are some of the students from CodeDays, aren't you?

Much success on this task and don't hesitate to ask for help if you need any.

mai-tran-03:
To get started, I replicated the issue and observed how the background color of this progress bar is rendered in different themes. The color is consistent for all other themes, except the light preview. Based on the keywords, using the global search tool, I searched for `progressBar` and `setBackground`. Then, I looked at any relevant packages and files where the progress bar could be created and the background could be called. Then, I followed this path `/bundles/org.eclipse.ui.workbench/Eclipse UI/org.eclipse.ui.internal.progress/` to the key locations of `DetailedProgressView.java` and `ProgressInfoItem.java`. 

[...]

PR #2753, opened 2025-01-23
[Light Theme]Fix for bg colors of entries in Progress View

mvm-sap:
Fix for issue [#2313](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2313)
Issue has been fixed in all three platforms Win, Mac and Linux.

### DIFF
--- a/bundles/org.eclipse.ui.themes/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.themes/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.themes;singleton:=true
-Bundle-Version: 1.2.2800.qualifier
+Bundle-Version: 1.2.2900.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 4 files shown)


mvm-sap:
@BeckerWdf Could you test once and confirm? Also, please update the screenshots as I have some issues uploading them.

BeckerWdf:
Tested on macOS:
It now looks like this:
image

But without your change it looks also good:
image

Can you pls. verify on your end?
Did the issue maybe only occur while the theme as the "Light (Preview)" theme?

BeckerWdf:
Pls. have a look at https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressInfoItem.java#L835

There the code checks for the current themeID. Depending on this some colors are set. 
We should try to get rid of this color-setting logic and do this only in via the CSS files. But that would be a different issue / PR.

BeckerWdf:
@mvm-sap: I just tested this PR once again. But even without your changes in this PR I cannot reproduce the issue. 
The progress view looks good to me:
image

Do you still see this issue reported in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2313 with Eclipse 4.36?

BeckerWdf:
Closing as the issue can no longer be reproduced. Re-open if this is still an issue.

PR #2755, opened 2025-01-24
Bug 567104 - Export Error if a File should be overriden

ptziegler:
The change done in 9c6c5519dfb245b2b1c9bf5b3686cf9c324bd67e introduces a regression due to which it is no longer possible to export a file to the file system, if the file already exists.

When copying a file using the Files API, one needs to explicitly set the REPLACE_EXISTING flag, in order to overwrite the target file, otherwise a FileAreadyExistsException is thrown.

Note: The "writeFile" method is only called when the user explicitly chose to overwrite the existing file. It is therefore safe to always set this flag.


ptziegler:
https://github.com/user-attachments/assets/5a40afc4-452a-4894-b8f5-a6dbcc36b549

iloveeclipse (approved):

Issue #2338, opened 2024-09-13
Sticky scrolling resets on comment lines


dgibson66:
When sticky scrolling encounters a line where the comment starts at column 1, it resets.  Unfortunately, this is exactly how Toggle Comment creates comments.  Please make sticky scrolling ignore comment lines.

HannesWell:
Sticky scrolling is a UI feature, therefore moved this to the eclipse.platform.ui repository.

@Christopher-Hermann can you please have a look at this?
But since Stick-Scrolling is not language sensitive and only considers indentation I fear that's how it is working.

Christopher-Hermann:
Yes, the issue is already know. 
Good news: I'm already working on an extension point so that every language can provide specific Sticky-Scrolling calculation.
Bad news: I assume it will take some time until it is adapted for java

Also see #1950 and #2128

Christopher-Hermann:
Fixed with #2756 and https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2149

Issue #2128, opened 2024-07-24
[Sticky scrolling] Commented-out code is being incorrectly recognized


jakub-suliga:
When code is commented out, sticky scrolling incorrectly detects the line.

To reproduce this issue:
Under org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider.CommentPosition.findFirstContent(CharSequence, int), some code is commented out.

![image](https://github.com/user-attachments/assets/919cd9ba-3a0f-4df4-8a49-1fb7a7ad1161)

In the method org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider.CommentPosition.computeProjectionRegions(IDocument), there is some commented-out code.
Before the the commented out line:

![image](https://github.com/user-attachments/assets/a0f32398-0371-41bf-96e8-d37d34e3c472)

And after:

![image](https://github.com/user-attachments/assets/72707c26-b4b9-45e6-9e6a-aec6c6b1930f)

tomaswolf:
Related to https://github.com/eclipse-platform/eclipse.platform.ui/issues/2115 and https://github.com/eclipse-platform/eclipse.platform.ui/issues/1950 . Sticky scrolling would need to be able to know what a comment is, which is different in different languages. AFAIK the current implementation is based purely on indentation? 

Perhaps it could consider document partitions, if any. Though I'm not sure there is any platform-wide "standard" to identify comment partitions, and not every editor might use partitions for comments...

Christopher-Hermann:
Correct, sticky scrolling is based purely on indentation. It is planned to provide a extension point so that every language can define a own implementation for the sticky scrolling calculation. For java we can then provide a implementation that is able to handle this kind of comments, annotations which are currently a problem or unusual code styles. 
If no specific implementation is provided, the default will still be based on indentation.

Unfortunately I had no time to look into this yet, but definitely want to pick this one up as soon as I have some time.

Christopher-Hermann:
Fixed with #2756 and https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2149

Issue #1950, opened 2024-06-11
Sticky scrolling should allow to be fed by language-specific info/annotations


mickaelistria:
Taking for example such Java code, in the middle of some file

```java
if (condition) {
  // some code
} else if (another condition) {
  // some other code
} else {
  // focus here
}
```
If editor top line is around "focus here", then the StickyLine would be `} else {` which is fine. But in this particular context, one would also like to get the conditions to make the Sticky more useful.
Exprected sticky lines for best result would be
```java
if (condition) {
} else if (another condition) {
} else {
```

On the contrary, some structures of the form
```
Stream.of()
	.map(...)
	.map(...)
	.filter(...)
	.map(x -> {
		// focus here
	})
```
Tend to render confusing or not so helpful information as sticky as it would only show
```
Stream.of()
	.map(x -> {
```

Although indentation already works very well (I wasn't expecting so good results), we'll most likely need to tune what's shown according to the file or language. The language support could create internal annotations of which lines are relevant and "stick-worthy", and the Sticky area shows lines tagged as such.

Christopher-Hermann:
Done with #2756

Christopher-Hermann:
Java already implemented the enhancement point: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2149

jld01:
CDT has implemented a sticky lines provider: https://github.com/eclipse-cdt/cdt/issues/1455

PR #2756, opened 2025-01-24
[StickyScrolling] Introduce enhancement point

Christopher-Hermann:
In order to implement editor/language specific sticky lines provider, a new extension point is introduced.

See issues:
- #2338
- #2128
- #1950

The idea is that the package of the extension point is internal in the first place.
1. We from SAP will provide a extension for the SAP language ABAP.
2. In https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1851 we provide a implementation for JDT. 

If both implementation works as expected, the API is most probably stable and we can change to package of the extension point to public usage.

### DIFF
--- a/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
@@ -15,7 +15,8 @@ Export-Package:
  org.eclipse.ui.internal.editors.text.codemining.annotation;x-internal:=true,
  org.eclipse.ui.internal.texteditor;x-internal:=true,
  org.eclipse.ui.internal.texteditor.stickyscroll;x-internal:=true,
- org.eclipse.ui.texteditor
+ org.eclipse.ui.texteditor,
+ org.eclipse.ui.texteditor.stickyscroll
 Require-Bundle: 
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.core.expressions;bundle-version="[3.9.0,4.0.0)",
--- a/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.editors; singleton:=true
-Bundle-Version: 3.19.200.qualifier
+Bundle-Version: 3.20.0.qualifier
 Bundle-Activator: org.eclipse.ui.internal.editors.text.EditorsPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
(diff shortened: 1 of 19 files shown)


bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
```diff
@@ -15,7 +15,8 @@ Export-Package:
  org.eclipse.ui.internal.editors.text.codemining.annotation;x-internal:=true,
  org.eclipse.ui.internal.texteditor;x-internal:=true,
  org.eclipse.ui.internal.texteditor.stickyscroll;x-internal:=true,
- org.eclipse.ui.texteditor
+ org.eclipse.ui.texteditor,
+ org.eclipse.ui.texteditor.stickyscroll;x-internal:=true
```
BeckerWdf: so we would remove the "x-internal:=true" part to make this "real" API once we are confident enough?

Christopher-Hermann: Yes, that would be the idea. But I'm not sure if this is the right approach or if this is a good idea.

mickaelistria:
Can this share a lot of code with the folding range?
Instead of a new dedicated provider API, could it be implemented through reconciler and annotations (just like Folding) ?

Christopher-Hermann:
> Can this share a lot of code with the folding range? Instead of a new dedicated provider API, could it be implemented through reconciler and annotations (just like Folding) ?

investigated the possibility, and it appears that annotations can likely be reused for sticky scrolling functionality. We could attach a depth attribute to each line that is a potential sticky line candidate. The sticky line handler could then determine which line should be attached based on the scroll index and the depth specified in the annotation.

However, utilizing annotations may reduce flexibility, which can be beneficial in certain scenarios. For example, if you want to apply different styles to the sticky lines or even add text that is not originally part of the editor, relying solely on annotations could be limiting.
See this example where I added a custom text with custom styling:
Bildschirmfoto 2025-01-29 um 14 54 29

danthe1st:
> However, utilizing annotations may reduce flexibility

I guess it would be possible to use folding annotations by default (if present, otherwise use indentation as done currently) and still allow overriding the sticky scrolling functionality?

BeckerWdf:
@mickaelistria: We would like to finish this early when master opens up again.
Can you pls. again have a look and thought about @Christopher-Hermann's comments?

Christopher-Hermann:
No updates or comments for a long time. So we will merge this change if there are no concerns in the next days/weeks as there is now also a JDT implementation: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1851

Need to check if a  isEnabled() method needs to be added

BeckerWdf:
testCreateAndRunWorkbench (RcpTestSuite PlatformUITest) is unrelated and documented in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1517

Regarding: testCancellingWhenRunning (org.eclipse.jface.text.tests.reconciler.FastAbstractReconcilerTest)
I could not find an issue for that. Assertion message is: "reconciler never ran in 5 seconds".
@Christopher-Hermann: Could this be related to your change?

Christopher-Hermann:
> @Christopher-Hermann: Could this be related to your change?

I guess not, but not 100% sure.
The test was never failing before when I introduced the sticky scrolling. In this change, only the extension point was introduced. The sticky scrolling behavior it self was not changed.

I will try to start the build again...

BeckerWdf (approved):

PR #2763, opened 2025-01-29
Fix: Statusline not cleared after job completion if sub-tasks are used

### LINKED ISSUE #2762: Statusline not cleared after job completion if sub-tasks are used

## Steps to reproduce

- Start a workspace with the `org.eclipse.ui.examples.job` plugin installed
- Open the "Job Factory" view
- Click on "Runnable in Widnow"

Image

I expected: Upon completion, the task name should be cleared again

* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): I20250129-0030

## Community

ptziegler:
Similar to `setTaskName()`, the string created when calling `subTask()` must also be stored in the `fTaskName`, rather than in `fMessageText` field. Otherwise the label of the status line is not cleared after calling done() on the progress monitor.

Because the text of the sub-task is calculated using the value of `fTaskName`, the original name of the task needs to be stored in a separate field to avoid the name of the previous sub-task to be contained in the next sub-task.

Closes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2762

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IWorkbenchWindowTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IWorkbenchWindowTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2013 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


ptziegler:
Without this fix, the latter test case should fail with `Status line was not updated. expected:<[]> but was:<[Task: SubTask]>`

Also pinging @jukzi because this continues adc4b6bab05b88af27cb5020996f8bd312f9ff35

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IWorkbenchWindowTest.java
```diff
@@ -197,4 +201,28 @@ public void testIsApplicationMenu() {
...
+			monitor.beginTask("Task", 1);
+		});
+		assertStatusText("");
+	}
+
+	@Test
+	public void testRunJobInStatusLineWithSubtasks() throws Throwable{
+		fWin.run(false, false, monitor -> {
+			monitor.beginTask("Task", 1);
+			monitor.subTask("SubTask");
```
jukzi, line 219: good that you add a test. Can we add an assertion that the given text was set?

ptziegler: Sure. I've added additional checks when executing the runnable.

jukzi: thanks

vogella:
Thanks @ptziegler

Issue #2764, opened 2025-01-29
'Switch to Editor' dialog does not include parts contributed via e4 model fragment which represent an Editor


feilimb:
### Steps to Reproduce

To reproduce you will need to define an editor part which is contributed via an e4 model fragment (eg. via a _PartDescriptor_), and can be opened eg. via a Menu addition.

It should be marked as closeable and given the magic '_Editor_' tag (see screenshot snippet below):

![Image](https://github.com/user-attachments/assets/f1b8b533-7c79-4d32-85e2-d9e58e6f92ba)

Launch the IDE application and open the newly contributed Part which represents an Editor and was contributed via a model fragment.

Go to the _Switch to Editor_ dialog, via _Window/Navigation/Switch to Editor_

The dialog does _not include_ the Part contributed via model fragment which represents an Editor, it will contain any other compatibility layer editors which have been opened, eg. the simple Text Editor.

### Notes

Looking at platform source code, there appears to multiple places where the '_Editor_' tag is checked to perform some handling based on its presence.

Would it be possible to update the Switch to Editor dialog, to also cater for e4 part(s) contributed which represent Editor(s).

iloveeclipse:
Could you please attach a simple bundle project that contributes such an editor, so whoever is interesting in the bugfix could quickly debug the problem without wasting time on creating the reproducible example? Thanks.

feilimb:
> Could you please attach a simple bundle project that contributes such an editor, so whoever is interesting in the bugfix could quickly debug the problem without wasting time on creating the reproducible example? Thanks.

Yes of course, apologies for not including in the original post. I will put that together and attach here today.

feilimb:
Sample plugin is now available on the master branch in the _org.eclipse.e4.ui.examples_ collection.

### Instructions
- Open the Run Configurations in your Platform SDK, select the '_Runtime Workspace_' run configuration, and choose the '_Select plug-ins.._' - with the following dialog enter the newly added plugin id, ie. '_org.eclipse.e4.ui.examples.e4editor_' - and choose _OK_, it should now be listed in the Plugins list/table, at the bottom, under '_Additional Plugins_', ensure the checkbox is checked.
- Run this _Runtime Workspace_ run configuration, to start up the IDE, close the _Welcome to Eclipse_ screen/view
- to demonstrate the issue, open eg. a dummy text file, via File / Open and just open any text file
[...]

PR #2766, opened 2025-01-30
Simple example of a dummy e4 editor addition, which is contributed via a Model Fragment

feilimb:
This pull request was created in response to the suggestion from @laeubi in the Issue #2764


examples/org.eclipse.e4.ui.examples.dummyeditor/src/org/eclipse/e4/ui/examples/dummyeditor/DummyEditor.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.eclipse.e4.ui.examples.dummyeditor;
+
+import jakarta.annotation.PostConstruct;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Text;
+
+public class DummyEditor {
```
laeubi:
```suggestion
public class ExampleE4Editor {

```

examples/org.eclipse.e4.ui.examples.dummyeditor/src/org/eclipse/e4/ui/examples/dummyeditor/DummyEditor.java
```diff
@@ -0,0 +1,29 @@
...
+
+	@PostConstruct
+	public void postConstruct(Composite parent) {
+		Text label = new Text(parent, SWT.ITALIC);
+		label.setText("Dummy Editor Content...");
```
laeubi:
```suggestion
		label.setText("E4 Editor Content...");

```

laeubi (commented):
Thanks for the example I think in gernal it is good but ther is no need to name it "dummy" :-)

feilimb:
> Thanks for the example I think in gernal it is good but ther is no need to name it "dummy" :-)

:) OK makes sense, I'll remove the 'dummy' references throughout.. and update the PR

examples/org.eclipse.e4.ui.examples.dummyeditor/src/org/eclipse/e4/ui/examples/dummyeditor/DummyEditor.java
feilimb: Modified as suggested in latest version.

examples/org.eclipse.e4.ui.examples.dummyeditor/src/org/eclipse/e4/ui/examples/dummyeditor/DummyEditor.java
feilimb: Modified as suggested in latest version.

laeubi (approved):
Looks good

PR #2767, opened 2025-01-31
Undeprecate updateToolbar() in FilteredTree  and call again on refresh

ptziegler:
This partially reverts 3726516c0ab381d3d29e24e412f73679e830aba2 due to which the updateToolbar() method is no longer called by the default refresh job.

Subclasses may override updateToolbar() without also overriding doCreateRefreshJob() which in return means that their toolbar is not updated anymore.

Because this method is called again, the deprecation tag introduced with 176f5030b1ed724274f88552b8bec14e43111832 is also removed


ptziegler:
See https://github.com/eclipse-platform/eclipse.platform.ui/pull/2567#issuecomment-2625726250 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/2567#issuecomment-2626359811

@vogella @merks

merks (approved):
Looks good.  Thanks for the super speedy response! 🚀

ptziegler:
The failure in ProgressViewTests seems sprodaric and unrelated to this change. Given that the test succeeds locally, I'll merge this now.

merks:
Thanks!🙏

PR #2770, opened 2025-01-31
introduce getMouseHover, getMouseOut and getMouseMove in ICodeMining

tobiasmelcher:
This change allows code mining implementors to react on mouse move events and call setCursor on the text widget for example.

We would like to implement a similar feature like shown in the following screenshot based on the eclipse code minings.
vscode_links
The code mining text color can already be set by overriding the draw method of LineHeaderCodeMining. The mouse cursor handle currently cannot be currently specified by the implementor; it is hard coded in AbstractInlinedAnnotation. This pull request introduces the methods getMouseHover, getMouseOut and getMouseMove in the ICodeMining interface so that the implementor can influence what mouse cursor is shown when hovering and moving over the mining. This API is very low level and technical, but it gives the implementor all possibilities to provide its own link-like controls in the code minings.

### DIFF
--- a/bundles/org.eclipse.jface.text/.settings/.api_filters
+++ b/bundles/org.eclipse.jface.text/.settings/.api_filters
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<component id="org.eclipse.jface.text" version="2">
+    <resource path="src/org/eclipse/jface/text/codemining/ICodeMining.java" type="org.eclipse.jface.text.codemining.ICodeMining">
+        <filter id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.jface.text.codemining.ICodeMining"/>
+                <message_argument value="getMouseHover()"/>
+            </message_arguments>
+        </filter>
+        <filter id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.jface.text.codemining.ICodeMining"/>
+                <message_argument value="getMouseMove()"/>
+            </message_arguments>
+        </filter>
+        <filter id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.jface.text.codemining.ICodeMining"/>
+                <message_argument value="getMouseOut()"/>
+            </message_arguments>
+        </filter>
+    </resource>
+</component>
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningLineContentAnnotation.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningLineContentAnnotation.java
@@ -70,11 +70,7 @@ public class CodeMiningLineContentAnnotation extends LineContentAnnotation imple
 	 * @param viewer   the viewer
 	 */
 	public CodeMiningLineContentAnnotation(Position position, ISourceViewer viewer) {
-		super(position, viewer);
-		fResolvedMinings= null;
-		fMinings= new ArrayList<>();
-		fBounds= new ArrayList<>();
-		afterPosition= false;
+		this(position, viewer, false);
 	}
 
 	/**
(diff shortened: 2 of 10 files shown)


BeckerWdf:
we have API Errors here.

As far as I see the version was incremented to 3.27 with https://github.com/eclipse-platform/eclipse.platform.ui/commit/816a0569186bf893c32fced6f663e7cea5164172

So it should be sufficient to add since tags with 3.27 to the new methods.
This was done with https://github.com/eclipse-platform/eclipse.platform.ui/compare/1213c88d1ce65113923fe104a9a9206c54023f96..85c70a26b6762e451edf75878ab0a8d42b02f640
but the build was complaining.

Incrementing to 3.28 now complains that the since tag on SurroundWithBracketsStrategy is wrong.

What's correct here?
Can somebody pls. help?

BeckerWdf:
Do API tools have issues with "Default" methods?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/codemining/ICodeMiningExtension.java
BeckerWdf, line 1: copyright header is missing

BeckerWdf (changes requested):

BeckerWdf (approved):

BeckerWdf:
@mickaelistria: What do you think about this change? Is it fine from your point of view?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/codemining/ICodeMiningExtension.java
```diff
@@ -0,0 +1,44 @@
...
+package org.eclipse.jface.text.codemining;
+
+import java.util.function.Consumer;
+
+import org.eclipse.swt.events.MouseEvent;
+
+/**
+ * @since 3.27
+ */
+public interface ICodeMiningExtension {
```
mickaelistria: Instead of a new interface, you can add the new methods as `default` returning `null` to the existing interface.

mickaelistria (commented):
I have no strong objection, just a suggestion inline.
I'm also wondering whether the API is actually on the best grain: here it looks like from the example that a link SelectionEvent could be a better fit. Thinking much further, it looks like the trend with code mining is to allow any rendering in here. Currently, it's just Text (drawn with native bits), but maybe it should become a free SWT Canvas to allow full widget power.

But again, no objection if you want to merge that now.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/codemining/ICodeMiningExtension.java
tobiasmelcher: I tried this with [1213c88] but then got errors from the API checks. Looks like that adding a default method implementation to an existing interface makes troubles with existing interface implementations, see https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Evolving-Java-based-APIs-2.md (Adding a default method will break an existing client type if it already implements another interface that declares a default method with a matching signature)

mickaelistria: Indeed, this would require API exception. But if we evaluate that the risk of method with colliding signature is about 0, then it's fine to add default interface methods and to add API tools exceptions

BeckerWdf: how can an "API tools exception" be added?

vogella: +1 for default implementation. Exception can be added via quick fix

BeckerWdf: so does the current version of the PR look good to you @mickaelistria and @vogella regarding the "default" methods and the API tools exception?

BeckerWdf:
> Currently, it's just Text (drawn with native bits)

One can also draw icons with it. We do that in our product "ABAP Development Tools" e.g. as a means to execute unit tests:
image

I find the current API quite powerful. Adding further convenience API on top (e.g. like rendering something like "Styled Text" or "Text with links" should be handled separately.

mickaelistria (approved):

PR #2775, opened 2025-02-05
Remove dependencies on org.apache.commons.jxpath

merks:


jukzi:
Removing dependencies is a good thing, thanks! :-)

merks:
> Removing dependencies is a good thing, thanks! :-)

Moreover this dependency is not well-maintained, i.e., not maintained at all, it has a rejected CVE associated with it that never really wants to die, and it drags in other fossilized content:

https://repo1.maven.org/maven2/org/jdom/jdom/1.1.3/

It really needs to go away!!

HannesWell:
Thank you Ed for your suggestions, I'm currently incorporating them into https://github.com/eclipse-platform/eclipse.platform.ui/pull/2290 as you suggested in our (private ) conversation.

HannesWell:

PR #2779, opened 2025-02-06
Find/Replace overlay can't be disabled by product customization

### LINKED ISSUE #2778: Find/Replace overlay can't be disabled by product customization
While moving to 4.35 we've found issues with Find/replace overlay and custom multi page editors. 
We are investigating whether is that a problem on our side or in the platform, and in the meantime we wanted to disable Find/Replace overlay by default via product customization. To my surprise that failed :-(

The preferences **dialog** shows expected value, but the code that decides whether to use Find/Replace or not (and at which position) is broken, as it only considers the instance (workspace) settings and not product/configuration settings. This is unfortunately a common mistake...

I will push a patch in a moment.

iloveeclipse:
To read values from *all* scopes one has to use IPreferencesService, not the InstanceScope!

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2778

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceAction.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceAction.java
@@ -21,7 +21,9 @@ import org.eclipse.swt.events.DisposeListener;
 import org.eclipse.swt.widgets.Shell;
 
 import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.IPreferencesService;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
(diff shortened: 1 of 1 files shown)


jukzi (commented):
hookDialogPreferenceListener(), removeDialogPreferenceListener() 
use the same pattern (line 360 ff), do they need adjustment too?

jukzi:
also update `org.eclipse.ui.internal.findandreplace.overlay.FindReplaceOverlayTest.testDisableOverlayViaPreference()`?

iloveeclipse:
> hookDialogPreferenceListener(), removeDialogPreferenceListener() use the same pattern (line 360 ff), do they need adjustment too?

No. The product preferences do not change **during the session**, only instance preferences. Therefore listeners and writers are unaffected. Only readers.

Regarding the test: I don't think anyone runs tests with product customization?

jukzi:
just thinking that even if it no actual problem at that places it might avoid future copy&paste errors

iloveeclipse:
> just thinking that even if it no actual problem at that places it might avoid future copy&paste errors

IPreferencesService doesn't have listeners / change capabilities.

iloveeclipse:
Jörg, if there is no other objections, I plan to merge today.

jukzi (approved):

Issue #2786, opened 2025-02-10
[Code mining] Editor content jumps around on every save in 4.35


Philipp0205:
Let's make sure issue is not already fixed in latest builds first.

- [X] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

The issue is reproducible with Eclipse 4.35.0 Integration Build: I20250209-1800 [Unstable!](https://wiki.eclipse.org/Platform-releng/Unstable_build)

## Steps to reproduce

From a fresh installation and clean workspace:

* Create a class and a method that is used in that class
* Enable code mining and show references 
* Enable "build automatically"

![Image](https://github.com/user-attachments/assets/f3d8c133-acfa-415b-9d2b-f9574ce8a817)

You will see the following issue: 

https://github.com/user-attachments/assets/59e17691-1451-42f9-87ac-bcdbe573f732

I expected: Code mining should not overlap like before.

But got: The overlapping issue with code mining, already present in earlier versions, has worsened in 4.34. Upon closer inspection, it seems that code mining elements are reloaded/repainted twice instead of once, as in versions before 4.33.

This behavior also causes lines to shift, with new lines constantly being added and removed, making code mining unusable in larger projects!

## Tested under this environment:
* RHEL 9 
*  Eclipse 4.35.0
## Community

- [X] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

iloveeclipse:
@tobiasmelcher : could you please check, I believe you've contributed few changes in this area? The one in question could be 9b78e0f62ca3849d3d746b184b447bb7feb3f98d.

The editor content jumps now visible every time a new line is added and file is saved in java editor, before (4.30) it was not **that** visible.

Without looking into the code/debugger for me it looks like the code minings are painted / re-layouted more times as before, so the effect of jumping editor lines is much more visible **or** the paint/re-layout operation is longer.

tobiasmelcher:
sure, I will take a look. Please give me some time to understand the scenario.

iloveeclipse:
[...]

PR #2788, opened 2025-02-10
Add basic code mining tracing option to trace rendering

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2786


iloveeclipse:
@mickaelistria, @tobiasmelcher : WDYT? At least some basic data?

mickaelistria:
I don't think it's so relevant to add this in the code; adding those as conditional breakpoints when needed is easy enough IMO.

iloveeclipse:
I was thinking about people who don't have SDK sources checked out - these options are tracing options everyone can set at runtime.

PR #2789, opened 2025-02-10
[E4 XPath] Add new API in EModelService to find elements via Xpath and deprecate E4-XPath bundle for removal

HannesWell:
As found in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2290, these methods seem not to provide any real value and just complicate the implementation.

I'm contemplating to even remove XPathContext.getValue() as it also complicates the implementation and has no real usage in the SimRel:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/c9f129d39bcd6e77a70d80421a18de7c7c768bf2/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java#L39-L51

### DIFF
--- a/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContextFactory.java
+++ b/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContextFactory.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2024 BestSolution.at and others.
+ * Copyright (c) 2010, 2025 BestSolution.at and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -19,7 +19,11 @@ import org.eclipse.e4.emf.xpath.internal.java.JavaXPathContextFactoryImpl;
  * Factory responsible to create an XPath-Context
  *
  * @param <T> the object type the XPath is created for
+ * @deprecated To query an E4-model use
+ *             {@code org.eclipse.e4.ui.workbench.modeling.EModelService#findMatchingElements(org.eclipse.e4.ui.model.application.MApplicationElement, String, Class)}
+ *             instead.
  */
+@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
 public abstract class XPathContextFactory<T> {
 
 	/**
@@ -38,10 +42,15 @@ public abstract class XPathContextFactory<T> {
 	 * @param parentContext parent context
 	 * @param contextBean   Object
 	 * @return XPathContext
+	 * @deprecated The parent-context does not provide any real value. Just use
+	 *             {@link #newContext(Object)}
 	 */
 	public abstract XPathContext newContext(XPathContext parentContext, T contextBean);
 
 	/**
+	 * Creates a new {@code XPathContextFactory<EObject>} that's suitable to query
+	 * the E4-model.
+	 *
 	 * @param <T> the object type the xpath is created for
 	 * @return Create a new XPath-Factory
 	 */
--- a/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathNotFoundException.java
+++ b/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathNotFoundException.java
@@ -20,6 +20,7 @@ package org.eclipse.e4.emf.xpath;
  *
  * @since 0.6
  */
+@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
 public class XPathNotFoundException extends IllegalArgumentException {
 	private static final long serialVersionUID = -4174244860692153739L;
 
--- a/bundles/org.eclipse.e4.emf.xpath/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.emf.xpath/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
-Bundle-SymbolicName: org.eclipse.e4.emf.xpath
+Bundle-SymbolicName: org.eclipse.e4.emf.xpath;deprecated:="This bundle is deprecated since 2025-03 (removal in 2027-03 or later)"
 Bundle-Version: 0.6.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Require-Bundle: org.eclipse.emf.ecore;bundle-version="2.35.0",
--- a/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/EcoreXPathContextFactory.java
+++ b/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/EcoreXPathContextFactory.java
@@ -21,6 +21,7 @@ import org.eclipse.emf.ecore.EObject;
  *
  * @since 1.0
  */
+@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
 public class EcoreXPathContextFactory {
 
 	/**
--- a/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
+++ b/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2024 BestSolution.at and others.
+ * Copyright (c) 2010, 2025 BestSolution.at and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -23,7 +23,11 @@ import java.util.stream.StreamSupport;
  * Context in which the xpath is executed
  *
  * @since 1.0
+ * @deprecated To query an E4-model use
+ *             {@code org.eclipse.e4.ui.workbench.modeling.EModelService#findMatchingElements(org.eclipse.e4.ui.model.application.MApplicationElement, String, Class)}
+ *             instead.
  */
+@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
 public interface XPathContext {
 
 	/**
--- a/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/internal/java/JavaXPathContextFactoryImpl.java
+++ b/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/internal/java/JavaXPathContextFactoryImpl.java
@@ -57,6 +57,7 @@ import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 
+@SuppressWarnings({ "deprecation", "removal" })
 public class JavaXPathContextFactoryImpl<T> extends XPathContextFactory<T> {
 
 	private static final XPathFactory XPATH_FACTORY = XPathFactory.newInstance();
--- a/bundles/org.eclipse.e4.ui.model.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.model.workbench/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.e4.ui.model.workbench;singleton:=true
-Bundle-Version: 2.4.400.qualifier
+Bundle-Version: 2.4.500.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-RequiredExecutionEnvironment: JavaSE-17
--- a/bundles/org.eclipse.e4.ui.model.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.model.workbench/META-INF/MANIFEST.MF
@@ -6,7 +6,11 @@ Bundle-Version: 2.4.400.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-RequiredExecutionEnvironment: JavaSE-17
-Export-Package: org.eclipse.e4.ui.model;x-friends:="org.eclipse.e4.ui.workbench,org.eclipse.e4.ui.workbench.swt,org.eclipse.ui.workbench",
+Export-Package: org.eclipse.e4.ui.model;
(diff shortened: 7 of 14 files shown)


bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContextFactory.java
```diff
@@ -38,13 +39,20 @@ public abstract class XPathContextFactory<T> {
...
 	public abstract XPathContext newContext(XPathContext parentContext, T contextBean);
 
 	/**
 	 * @param <T> the object type the xpath is created for
 	 * @return Create a new XPath-Factory
+	 * @deprecated The implementation returned by this method just supports
+	 *             {@link EObject} elements. Use a specializd factory instead like
+	 *             {@link EcoreXPathContextFactory}.
 	 */
+	@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
```
laeubi: For me it looks more that the javadoc is wrong, can we not simply fixup the javadoc:

- remove wrong `@param`
- mention it creates a default instance suitable for query the E4 model (that these are effectively `EObjects` is an implementation detail)

that would best match the intend of this method.

HannesWell: That's also an option and it allows to deprecate the specialized `EcoreXPathContextFactory` entirely and to make the generic-type of the `JavaXPathContextFactoryImpl` more correct.

HannesWell:
@ptziegler as you contributed test-cases and worked on the parent-context handling, I wonder if you are actually using it and therefore have objections against deprecating `XPathContextFactory.newContext(XPathContext parentContext, T contextBean)`.

The same question applies to `XPathContext.getValue()` to you ore anybody else?

laeubi:
I know it is an "API break" but should we not mark `XPathContextFactory` with `@NoExtend`?

HannesWell:
> I know it is an "API break" but should we not mark `XPathContextFactory` with `@NoExtend`?

I ask this question to myself as well. On one hand the Eclipse SDK only provide one specific implementation of that class that can only handle a specific type of model, while the interface is designed (maybe over-engineered)  generic. So there could be a demand to have other implementations that can handle other type of models.
On the other hand, there is no benefit from doing that as there is nothing that would work with such other specialization in the SDK and third-parties that would want to do that, could just use their own classes and abstraction.

The same question also applies to `XPathContext`. In the end I'm not sure about it as this entire API seems to be over-engineered to me for the use-cases we have in the SDK. So a lot could be reduced just from POV of the Eclipse SDK.

laeubi:
I think the only interesting use-case would be to query the Model programmatically with an xpath expression, maybe it was though of a way to be used with other EMF model, but tin this case it would have ben better be implemented/provided by emf itself.

For this purpose, I would rather want to have a method in `org.eclipse.e4.ui.workbench.modeling.EModelService` similar to the `findElements` or an `XPathSelector` that implements `org.eclipse.e4.ui.workbench.Selector`.

As all of this basically boils down to `org.eclipse.e4.ui.model.workbench` one could even thing about deprecate the whole bundle (PDE now supports a warning for users), and with the next release copy the now much simpler implementation into `org.eclipse.e4.ui.model.workbench`... then we won't need any "public" API anymore and if there is really users they can still use the "old" impl and platform can just not include the emf.xpath anymore.

HannesWell:
> I think the only interesting use-case would be to query the Model programmatically with an xpath expression, maybe it was though of a way to be used with other EMF model, but tin this case it would have ben better be implemented/provided by emf itself.

There is also one other user in PDE, see e.g.
- https://github.com/eclipse-pde/eclipse.pde/pull/1612

> 
> For this purpose, I would rather want to have a method in `org.eclipse.e4.ui.workbench.modeling.EModelService` similar to the `findElements` or an `XPathSelector` that implements `org.eclipse.e4.ui.workbench.Selector`.

Adding something like `EModelService.findMatchingElements(MApplicationElement contextBean, String xpath, Class<T> clazz)` sounds like a good idea. It would boil down the actually used API to the minimum and would make it more accessible.
I added this in a very first prototype.
However using the new method in the `StringModelFragmentImpl` class to replace the reference to the existing `XPathContext`, is not possible as `o.e.e4.ui.model.workbench` doesn't depend on `o.e.e4.ui.workbench` but the other way round.

> 
> As all of this basically boils down to `org.eclipse.e4.ui.model.workbench` one could even thing about deprecate the whole bundle (PDE now supports a warning for users), and with the next release copy the now much simpler implementation into `org.eclipse.e4.ui.model.workbench`... then we won't need any "public" API anymore and if there is really users they can still use the "old" impl and platform can just not include the emf.xpath anymore.

If we make the approach from above work, then we could deprecate the entire bundle and inline it's content to wherever the new API lands when the bundle is finally removed.

HannesWell:
> However using the new method in the `StringModelFragmentImpl` class to replace the reference to the existing `XPathContext`, is not possible as `o.e.e4.ui.model.workbench` doesn't depend on `o.e.e4.ui.workbench` but the other way round.

We could also add it as something internal in `o.e.e4.ui.model.workbench` use it there and make it public only through EModelService.

HannesWell:
> Adding something like `EModelService.findMatchingElements(MApplicationElement contextBean, String xpath, Class<T> clazz)` sounds like a good idea. It would boil down the actually used API to the minimum and would make it more accessible.

This also allows us to avoid the clean-up of the parameteriziation problem of `XPathContextFactory.newInstance()`.

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/XPathHelper.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.stream.Stream;
+import org.eclipse.core.runtime.ILog;
+import org.eclipse.e4.emf.xpath.XPathContextFactory;
+import org.eclipse.emf.ecore.EObject;
+
+public class XPathHelper {
+
+	@SuppressWarnings("removal")
+	public static <T> Stream<T> findMatchingElements(EObject contextBean, String xpath, Class<T> clazz) {
```
laeubi: Does this needs to be public API or can we move it to an internal package?

HannesWell: The package is already internal and just a friend to `org.eclipse.e4.ui.workbench`:
`org.eclipse.e4.ui.model;x-friends:="org.eclipse.e4.ui.workbench,org.eclipse.e4.ui.workbench.swt,org.eclipse.ui.workbench"`

HannesWell:
I have looked at the usage in PDE of the new API and in general it looks good, the only thing that makes using this cumbersome is the fact that the new API is an instance method and not a class/static method and one needs an `EModelService` instance. This makes it cumbersome to use in static methods (which is the case in PDE).

laeubi:
I'm not sure I found the right place but the only one I can found in PDE is

`org.eclipse.e4.tools.emf.ui.internal.common.component.StringModelFragment.getTargetClassFromXPath(MApplication, String)`

As it has the `MApplication` you can do:

`EModelService modelService = application.getContext().get(EModelService.class);`

if you want avoid an additional parameter.

HannesWell:
I have now added the missing documentation and think this is ready for submission.
I would like to land this for the upcoming 2025-03 release and think it's save to do this as this just deprecate methods and a bundle and adds one alternative API.

@laeubi are you fine with this now? @merks do you also want to have a look?

> As it has the `MApplication` you can do:
> 
> `EModelService modelService = application.getContext().get(EModelService.class);`
> 
> if you want avoid an additional parameter.

I tried this but it looks like it doesn't work in all cases, because (at least in some cases) that actually an MApplication loaded from a resource/file of an Editor and not the 'active' application model of the running application. And in that case the context is not set to these model-elements and `getContext()` therefore returns null.

So for PDE I either need to get the context purely static (not sure if that's possible) or pass an `EModelService` instance as parameter.

laeubi:
> So for PDE I either need to get the context purely static 

Du to the nature of E4 this is not reliable possible but I think the case of an editor is rather special.

But I think we have some cases:

1. PDE can simply "fake" it by setting its own EclipseContext to the loaded model
2. PDE can call an (internal) API on e4 for this purpose and become a friend of the package.

bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContextFactory.java
```diff
@@ -19,7 +19,11 @@
  * Factory responsible to create an XPath-Context
  *
  * @param <T> the object type the XPath is created for
+ * @deprecated To query an E4-model use
+ *             {@link org.eclipse.e4.ui.workbench.modeling.EModelService#findMatchingElements(org.eclipse.e4.ui.model.application.MApplicationElement, String, Class)}
+ *             instead.
  */
+@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
```
laeubi, line 26: I don't think we need any deprecation here if we deprecate the bundle

bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContextFactory.java
```diff
@@ -38,12 +42,18 @@ public abstract class XPathContextFactory<T> {
 	 * @param parentContext parent context
 	 * @param contextBean   Object
 	 * @return XPathContext
+	 * @deprecated The parent-context does not provide any real value. Just use
+	 *             {@link #newContext(Object)}
 	 */
+	@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
```
laeubi: same here

bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathNotFoundException.java
```diff
@@ -20,6 +20,7 @@
  *
  * @since 0.6
  */
+@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
```
laeubi, line 23: same here

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/XPathHelper.java
```diff
@@ -0,0 +1,29 @@
...
+ * Contributors:
+ *     Hannes Wellmann - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.e4.ui.model;
+
+import java.util.stream.Stream;
+import org.eclipse.e4.emf.xpath.XPathContextFactory;
+import org.eclipse.e4.ui.model.application.MApplicationElement;
+
+public class XPathHelper {
```
laeubi: I think it would be fair if PDE will just call this method then

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/XPathHelper.java
```diff
@@ -0,0 +1,29 @@
+/*******************************************************************************
+ * Copyright (c) 2025, 2025 Hannes Wellmann and others.
```
laeubi:
```suggestion
 * Copyright (c) 2025 Hannes Wellmann and others.
```

laeubi (commented):
Can the method that is currently at `XPathHelper` maybe betterl ife in `ModelServiceImpl` or `StringModelFragmentImpl`?

bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContextFactory.java
HannesWell: I removed the additional deprecation of some methods, but the explicit deprecation of the classes should be kept as one otherwise doesn't get a corresponding warning in the code. And from my experience warnings in Manifests are overlooked more easily than in code.
Furthermore the deprecation javadoc contains additional information.

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/XPathHelper.java
HannesWell: I thought about this as well. Then lets keep it simple. I added the e4.tools bundle in PDE as a friend of the containing package.

HannesWell:
> Can the method that is currently at `XPathHelper` maybe betterl ife in `ModelServiceImpl` or `StringModelFragmentImpl`?

Moving it to `ModelServiceImpl` is not feasible due to dependency structure and `StringModelFragmentImpl` would be a class were I would expect such functionality.
And as the comment in `XPathHelper` suggests, this class is intended to contain all the (then used) code that's currently in `JavaXPathContextFactoryImpl` when the `org.eclipse.e4.emf.xpath` bundle is finally removed. So it won't be that empty as it's now on the long run. With that in mind, having a separate class for it seems suitable to me (and as it's all internal it can be changed easily in the future if necessary).

HannesWell:
@laeubi I would like to submit this PR this evening so that I can apply the new API/methods in PDE tomorrow after the next I-build.
I think the most critical part is the new API method in `EModelService`:
`<T> Stream<T> findMatchingElements(MApplicationElement searchRoot, String xPath, Class<T> clazz);`

But I think that's fine. Everything else is internal and can be changed at any time later, if necessary. Do you have any critical points that should definitively be addressed?

bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContextFactory.java
laeubi: > but the explicit deprecation of the classes should be kept as one otherwise doesn't get a corresponding warning in the code. 

You should already get a warning in the Manifest about the deprecation.

laeubi:
Then I won't call it `XPathHelper` .. there are way to much "helper", "util" and a like and they tend to grow over time with no real scope. What about name it `ModelXPathSupport` or similar then? Beside that I think it is fine and goes into the right direction.

e.g. we already have three `P2Utils` (and two of them in PDE):
grafik

HannesWell:
> Then I won't call it `XPathHelper` .. there are way to much "helper", "util" and a like and they tend to grow over time with no real scope. What about name it `ModelXPathSupport` or similar then? Beside that I think it is fine and goes into the right direction.

Ok, as discussed in a private conversion I have renamed it to `ModelXPathEvaluator`.

laeubi (approved):
Thanks for working on this!
New `EModelService` method should maybe mention in N&N

HannesWell:
You are welcome. I'm happy if that whole XPath topic is finally reduced to the actually used minimum.

> New `EModelService` method should maybe mention in N&N

Yes. Together with the replacement of JXPath. That might be interesting for others as well.

HannesWell:
> > New `EModelService` method should maybe mention in N&N
> 
> Yes. Together with the replacement of JXPath. That might be interesting for others as well.

Done via https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/276.

PR #2793, opened 2025-02-12
Draw completion proposals always as focused

### LINKED ISSUE #1688: Dark theme: code completion selection does not have enough contrast -> align with other popular dark themes

Christopher-Hermann:
### Problem description
When opening the completion proposals via the keyboard, the focus will stay in the editor to be able to accept further user input. This is causing the completion proposal to be drawn in non focus colors. This colors can lead to UX problems, especially in dark theme. With this fix, the completion proposals are always drawn in focused colors.

Before the fix in dark theme, it was hard to sport the selected proposal:
image

With the fix, the proposal is colored in the focus color:
image

### How to retest

1. Open a text editor and run the code completion (ctrl+space)
2. Check that the selected completion proposal is colored in the focused color

Fixes #1688

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/TableOwnerDrawSupport.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/TableOwnerDrawSupport.java
@@ -70,7 +70,7 @@ public class TableOwnerDrawSupport implements Listener {
 		return (StyleRange[])item.getData(STYLED_RANGES_KEY + column);
 	}
 
-	private TableOwnerDrawSupport(Table table) {
+	public TableOwnerDrawSupport(Table table) {
 		int orientation= table.getStyle() & (SWT.LEFT_TO_RIGHT | SWT.RIGHT_TO_LEFT);
 		fSharedLayout= new TextLayout(table.getDisplay());
 		fSharedLayout.setOrientation(orientation);
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/contentassist/CompletionProposalDrawSupport.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/contentassist/CompletionProposalDrawSupport.java
@@ -1,0 +1,63 @@
+/*******************************************************************************
+ * Copyright (c) 2025 SAP SE.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     SAP SE - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jface.internal.text.contentassist;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Listener;
+import org.eclipse.swt.widgets.Table;
+
+import org.eclipse.jface.internal.text.TableOwnerDrawSupport;
+
+/**
+ * Provides custom drawing support for completion proposals. This class ensures that completion
+ * proposals are always rendered with a focused appearance.
+ *
+ * <p>
+ * This drawing behavior addresses the particular situation where the code completion is triggered
+ * via keyboard shortcut, leaving the editor focused. In such cases, without this custom drawing
+ * support, the completion proposal would appear unfocused, leading to a suboptimal coloring.
+ * </p>
+ */
+public class CompletionProposalDrawSupport implements Listener {
+
+	private final TableOwnerDrawSupport fTableOwnerDrawSupport;
+
+	private CompletionProposalDrawSupport(Table table) {
+		fTableOwnerDrawSupport= new TableOwnerDrawSupport(table);
+	}
+
+	public static void install(Table table) {
+		CompletionProposalDrawSupport listener= new CompletionProposalDrawSupport(table);
+		table.addListener(SWT.Dispose, listener);
+		table.addListener(SWT.MeasureItem, listener);
+		table.addListener(SWT.EraseItem, listener);
+		table.addListener(SWT.PaintItem, listener);
+	}
+
+	@Override
+	public void handleEvent(Event event) {
+		if (event.widget instanceof Control control && !control.isFocusControl() && (event.type == SWT.EraseItem || event.type == SWT.PaintItem) && event.gc != null) {
+			Color background= event.widget.getDisplay().getSystemColor(SWT.COLOR_TITLE_BACKGROUND);
+			Color foreground= event.widget.getDisplay().getSystemColor(SWT.COLOR_WHITE);
+			event.gc.setBackground(background);
+			event.gc.setForeground(foreground);
+		}
+
+		fTableOwnerDrawSupport.handleEvent(event);
+	}
+
+}
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/link/contentassist/CompletionProposalPopup2.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/link/contentassist/CompletionProposalPopup2.java
@@ -40,6 +40,7 @@ import org.eclipse.swt.widgets.Table;
 import org.eclipse.swt.widgets.TableItem;
 
 import org.eclipse.jface.internal.text.TableOwnerDrawSupport;
+import org.eclipse.jface.internal.text.contentassist.CompletionProposalDrawSupport;
 import org.eclipse.jface.preference.JFacePreferences;
 import org.eclipse.jface.resource.ColorRegistry;
 import org.eclipse.jface.resource.JFaceColors;
@@ -274,7 +275,7 @@ class CompletionProposalPopup2 implements IContentAssistListener2 {
 
 		fIsColoredLabelsSupportEnabled= fContentAssistant.isColoredLabelsSupportEnabled();
 		if (fIsColoredLabelsSupportEnabled) {
-			TableOwnerDrawSupport.install(fProposalTable);
+			CompletionProposalDrawSupport.install(fProposalTable);
 		}
 
 		fProposalTable.setLocation(0, 0);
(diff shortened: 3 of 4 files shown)


Christopher-Hermann:
@iloveeclipse I had fixed this issue in the past, but we had to revert the changes since you found some issues on Linux: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1690
Could you please check if this more local fix is working on Linux? 

On Mac and Windows it looks good.

iloveeclipse:
I will test later, but in general I would assume this is not for this release, M3 is planned for tomorrow and the change affects everyone using content assist.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/contentassist/CompletionProposalDrawSupport.java
```diff
@@ -0,0 +1,59 @@
...
+		table.addListener(SWT.Dispose, listener);
+		table.addListener(SWT.MeasureItem, listener);
+		table.addListener(SWT.EraseItem, listener);
+		table.addListener(SWT.PaintItem, listener);
+	}
+
+	@Override
+	public void handleEvent(Event event) {
+		Color background= event.item.getDisplay().getSystemColor(SWT.COLOR_TITLE_BACKGROUND);
+		Color forground= event.item.getDisplay().getSystemColor(SWT.COLOR_WHITE);
```
BeckerWdf: is setting "white" correct for all themes?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/contentassist/CompletionProposalDrawSupport.java
```diff
@@ -0,0 +1,59 @@
...
+ * Provides custom drawing support for completion proposals. This class ensures that completion
+ * proposals are always rendered with a focused appearance.
+ *
+ * <p>
+ * This drawing behavior addresses the particular situation where the code completion is triggered
+ * via keyboard shortcut, leaving the editor focused. In such cases, without this custom drawing
+ * support, the completion proposal would appear unfocused, leading to a suboptimal coloring.
+ * </p>
+ */
+public class CompletionProposalDrawSupport implements Listener {
```
iloveeclipse, line 35: **If** you really want rhis class be public API, it should have "since" tag and all public methods should be properly documented.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/contentassist/CompletionProposalDrawSupport.java
Christopher-Hermann: Not sure, at least on MacOS I'm not able to change the selection foreground color. It is always white.
I think there is no system color for selection foreground, or any ideas which one I could use?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/contentassist/CompletionProposalDrawSupport.java
Christopher-Hermann: Ah damn, you are right. I wanted to move it into a internal package, I will dot it 👍

iloveeclipse:
> Could you please check if this more local fix is working on Linux?
> 
> On Mac and Windows it looks good.

I see errors on every popup close:
```
!ENTRY org.eclipse.ui 4 0 2025-02-12 16:51:52.599
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.Widget.getDisplay()" because "event.item" is null
	at org.eclipse.jface.contentassist.CompletionProposalDrawSupport.handleEvent(CompletionProposalDrawSupport.java:54)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5862)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1652)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1678)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1657)
	at org.eclipse.swt.widgets.Widget.release(Widget.java:1396)
	at org.eclipse.swt.widgets.Control.release(Control.java:4753)
	at org.eclipse.swt.widgets.Composite.releaseChildren(Composite.java:1576)
	at org.eclipse.swt.widgets.Canvas.releaseChildren(Canvas.java:279)
	at org.eclipse.swt.widgets.Decorations.releaseChildren(Decorations.java:503)
	at org.eclipse.swt.widgets.Shell.releaseChildren(Shell.java:3437)
	at org.eclipse.swt.widgets.Widget.release(Widget.java:1403)
	at org.eclipse.swt.widgets.Control.release(Control.java:4753)
	at org.eclipse.swt.widgets.Widget.dispose(Widget.java:576)
	at org.eclipse.swt.widgets.Shell.dispose(Shell.java:3354)
	at org.eclipse.jface.text.contentassist.CompletionProposalPopup.hide(CompletionProposalPopup.java:1094)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.hide(AsyncCompletionProposalPopup.java:359)
	at org.eclipse.jface.text.contentassist.ContentAssistant.hide(ContentAssistant.java:2268)
	at org.eclipse.jface.text.contentassist.ContentAssistant$Closer.mouseDown(ContentAssistant.java:221)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:230)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5862)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1652)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5072)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4524)
```

bundles/org.eclipse.jface.text/src/org/eclipse/jface/contentassist/CompletionProposalDrawSupport.java
Christopher-Hermann: On my Mac, SWT.COLOR_WIDGET_HIGHLIGHT_SHADOW and SWT.COLOR_LIST_BACKGROUND does result in a white text color. But both don't seem to be right here.

Christopher-Hermann:
> I see errors on every popup close:

Yes I need to check this, there are also some "SWT Resource was not properly disposed" errors

iloveeclipse:
> there are also some "SWT Resource was not properly disposed" errors

These are most likely follow ups of original error

bundles/org.eclipse.jface.text/src/org/eclipse/jface/contentassist/CompletionProposalDrawSupport.java
Christopher-Hermann: Done

Christopher-Hermann:
> I see errors on every popup close:

Should be fixed.

Is the coloring working on Linux?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/contentassist/CompletionProposalDrawSupport.java
Christopher-Hermann: I removed the foreground color completely, by this the color defined in the operating system is used.
For background, SWT.COLOR_TITLE_BACKGROUND is giving the highlight color. This should work in every case. Thanks for the hint 👍

Christopher-Hermann:
@iloveeclipse can you confirm that the coloring is working on linux like expected?

MacOS and Windows is working. Then I would merge this change.

iloveeclipse:
> Then I would merge this change.

You can't merge the change, we are in the RC phase.

iloveeclipse:
> can you confirm that the coloring is working on linux like expected?

I can't observe the original problem on Linux. The proposal table without this change looks exactly as with the change (blue selection on dark background).

image

Christopher-Hermann:
If there are no objections I would merge the PR

szarnekow:
> If there are no objections I would merge the PR

Personally I find the focus color to distracting - it grabs too much of my attention while typing.

Christopher-Hermann:
> Personally I find the focus color to distracting - it grabs too much of my attention while typing.

Yes, good point. I will check if we can just change the non focus color to something more visible. The problem is, as far as I understood, that on Linux the code completion is always rendered with focus back ground color. So when changing the color, it will most probably change the behavior on Linux.

Christopher-Hermann:
I tried using SWT.COLOR_TITLE_INACTIVE_BACKGROUND if the control is not focused, but this is even more worse (at least on MacOS). 

image

So I see two ways to solve the issue:

1. Go with the approach as it is suggested in this PR: Just show the code completion as focus in any case
2. Introduce color definitions for table selections, as it was done in #1690. Later on, when work on #1690 is continued, we can reuse the already existing color definitions.

fedejeanne:
@Christopher-Hermann I rebased on the current `master` and also added a `final` modifier.

I also set this one to a draft since I am trying to integrate it into #3659 via cherry-pick (hence the rebase). If I succeed then this PR could be dropped. I hope that's fine by you?

Christopher-Hermann:
> @Christopher-Hermann I rebased on the current `master` and also added a `final` modifier.
> 
> I also set this one to a draft since I am trying to integrate it into #3659 via cherry-pick (hence the rebase). If I succeed then this PR could be dropped. I hope that's fine by you?

Yes, this could be done with the new PR. I also tried this out and it works fine.

I guess the only thing is the decision/discussion if we want to color the completion always as selected, or if we should keep the non focus color. I guess for the second we don't have to do anything, this should work out of the box with #3659

fedejeanne:
> I guess the only thing is the decision/discussion if we want to color the completion always as selected, or if we should keep the non focus color.

For me it makes sense to use the same color that lists/trees have when their items are selected. I'll try that and see how it looks.

Issue #2799, opened 2025-02-17
[e4] Adding tool items with before/after adds items in the wrong order in the toolbar


laeubi:
I have used the preference spy here as an example:

![Image](https://github.com/user-attachments/assets/7985c04b-3568-4761-889e-f58e95094fee)

If I choose nothing the "Position at list" it looks like this:

![Image](https://github.com/user-attachments/assets/b3e6b404-bd1b-4f44-8d1c-5a0ad2b123c9)

But if one uses `before:` or `after:` 

![Image](https://github.com/user-attachments/assets/99939259-3a93-42e3-9ed6-13b7074eb1f2)

it looks like this:

![Image](https://github.com/user-attachments/assets/3ccb1613-c812-45a3-90b3-9b3ffdfe6bc8)

PR #2800, opened 2025-02-17
Increment the index after the element was added to the list

laeubi:
Currently elements are merged in reverse order if an insert index is give because the add is always performed at the original index.

This now do the following:
- on each insert operation the index is incremented if given
- if the index is larger than the list size the element is added instead

Fixes 
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2799

### DIFF
--- a/bundles/org.eclipse.e4.ui.model.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.model.workbench/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.e4.ui.model.workbench;singleton:=true
-Bundle-Version: 2.4.500.qualifier
+Bundle-Version: 2.4.600.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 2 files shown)


laeubi:
It would be good to have this in RC1

iloveeclipse:
> It would be good to have this in RC1

This code/behavior exists "as is" for 13 years (since 4d10c1a37a911bb20f12b49088e20d72ed1eb9ea), so changing this behavior may affect others relying on that => I would prefer to fix this in 4.36.

laeubi:
> > It would be good to have this in RC1
> 
> This code/behavior exists "as is" for 13 years (since [4d10c1a](https://github.com/eclipse-platform/eclipse.platform.ui/commit/4d10c1a37a911bb20f12b49088e20d72ed1eb9ea)), so changing this behavior may affect others relying on that => I would prefer to fix this in 4.36.

What does not make it any better... The problem is that most of the time if this is sued one adds usually one element and you never notice it of course. As soon as one uses more than one element it is simply wrong.

And "relying" on the feature is quite hard, maybe people used it as a _workaround_ to add elements in reverse order then they can easily fix it to the correct behavior now.

iloveeclipse:
Once again, I'm not against the fix in general, I'm against **including it in RC**.
The point of having RC builds is to stabilize the system before release.

Reported issue isn't blocking or critical or anything that would  need an urgent fix, given the fact it was unnoticed for 13 years, so it doesn't need to be fixed in the RC, because every change may introduce regressions, especially if touching code not changed for so long time.

So assuming the fix is OK, let plan it for 4.36.

merks:
The fact that's it's been broken for 13 years and that people may be relying on the broken behavior is definitely a valid point.

That being said, I doubt anything in SimRel will notice this change and I doubt the people who might be affected will notice until after a release because they likely never test against pre-released versions.  I.e., I am somewhat in doubt that the timing of the fix, M1 versus RC1, will make it less painful for anyone...

iloveeclipse:
> I am somewhat in doubt that the timing of the fix, M1 versus RC1, will make it less painful for anyone...

The probability that a regression would be noticed is surely higher for period of 2 months after M1 compared to two weeks of RC's. 

As said, there is no reason to rush with this change, so being paranoid and to keep RC from unexpected regressions I would prefer this change to go into 4.36 M1.

akurtakov:
Will it be feasible to have a test so this doesn't regress again?

PR #2801, opened 2025-02-17
Add activity support for keybindings

### LINKED ISSUE #2859: Restore activity support for keybindings
We are facing an issue with bundle that contributes shortcuts (key bindings in Eclipse speak) to the workbench but by default is not enabled via "activities" and so not visible anywhere in Eclipse UI.

So **in theory**, no UI contributions from that plugin should be shown and active, but as of today the contributed bundle shortcuts are still active and can appear and the bound commands can be executed (and cause confusion).

This happens because activity support for key bindings was broken during e4 transition in 2010 and Eclipse 4.x happily processes key bindings from disabled activities.

The PR that reimplements activities support for key bindings: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2801.

iloveeclipse:
Currently Eclipse processes key bindings from "disabled" activities, activity support for key bindings was broken during e4 transition in 2010.

This change re-implements activity support for keybindings contributions.

- Don't show filtered key bindings in Keys preference page
- Hide bindings for filtered keys to BindingTableManager
- Fail command execution for filtered activity
- Update bindings state if enabled activities change

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2859

### DIFF
--- a/bundles/org.eclipse.e4.ui.bindings/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.bindings/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.bindings;singleton:=true
-Bundle-Version: 0.14.500.qualifier
+Bundle-Version: 0.14.600.qualifier
 Bundle-Name: %pluginName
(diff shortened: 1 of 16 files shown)


laeubi:
> check whether one should cache BindingTable.isActive(Binding) results (note: this would probably need an activity listener in e4 layer).

E4 uses `IEventBroker` instead of listeners or observable values in the Model.

PR #2802, opened 2025-02-18
fix concurrent use of AbstractTextSearchResult.fMatchEvent #992

### LINKED ISSUE #992: sporadic issue : Search results :  Incomplete search results are shown
Bug type : Sporadic
Eclipse version : 2023-09 (4.29) / 2025-03 too
Build id: I20230802-1800 / I20250216-1800
OS : Linux

### In 2023-09 (I20230802-1800)
1. Import the attached project. [p60.zip](https://github.com/eclipse-platform/eclipse.platform.ui/files/12250168/p60.zip)
2. Open File Search then search for text `some`
3. Check the search results in `Search View`. Mostly for the first time results are complete.
4. `Clear All` using "Search View" toolbar and hit `Refresh` next to it. Try this 4-5 times quickly.
5. It can be reproduced only by clicking `Refresh` in the `Search View` tool bar couple of time instantly.
6. Now observe the results in step 4/5 sometimes some results are missing. Please refer the attached images.

In development environment when I lauch a 'run-time' in debug mode this is very frequent. Even the first time search results are incomplete.
search-issue-missing-items
search-issue-missing-items-4-6

Below is the screenshot of the first serach result from run-time debug.
search-issue-missing-items-run-time-debug

### In 2025-03 M3 (I20250216-1800)
_(taken from my duplicate issue: https://github.com/eclipse-platform/eclipse.platform/issues/1741)_

Search for something (_Search > Search (`Ctrl + h`)_) with several hits, like `@Override`:

(first run)

Image

(second run)

Image

This is acceptable if I have thousands of hits since I probably won't miss the ones I don't see, but it's not ok when I look for something like `@Parameter`, which has only 72 hits:

(first run)
Image

(fifth run)
Image

jukzi:
Some search results where missing when the same object was concurrently passed to listeners. There is no point to reuse the instance.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/992

### DIFF
--- a/bundles/org.eclipse.search/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.search/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.search; singleton:=true
-Bundle-Version: 3.17.100.qualifier
+Bundle-Version: 3.17.200.qualifier
 Bundle-Activator: org.eclipse.search.internal.ui.SearchPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/bundles/org.eclipse.search/newsearch/org/eclipse/search/ui/text/AbstractTextSearchResult.java
+++ b/bundles/org.eclipse.search/newsearch/org/eclipse/search/ui/text/AbstractTextSearchResult.java
@@ -44,7 +44,6 @@ public abstract class AbstractTextSearchResult implements ISearchResult {
 
 	private final ConcurrentMap<Object, Set<Match>> fElementsToMatches;
 	private final List<ISearchResultListener> fListeners;
-	private final MatchEvent fMatchEvent;
 	private final AtomicInteger matchCount;
 
 	private MatchFilter[] fMatchFilters;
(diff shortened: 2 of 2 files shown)


merks:
I'm really happy to see changes for this!

iloveeclipse:
From my POV this PR is OK to merge.
@akurtakov : I saw I've merged some PR's from Jörg, how did you avoided the error with eca?
Just ignore it?

akurtakov:
Some PRs don't fail ECA https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2510 so I guess it's smth related with Joerg change of email in eclipse account. As the PR had valid ECA but now doesn't I really have no clue what should be done. I'm leaving it to the "collective" knowledge to decide.

iloveeclipse:
@waynebeaton : I guess you've had such case: the PR was set at the time the PR author (@jukzi) worked for a company X, using company X mail account. Now PR author is not working anymore for the X company. The email is still mentioned in commit metadata. 

Considering the fact that at the moment of PR creation ECA check was green, I assume we are OK to merge from the legal point of view. 

However, I fear the tooling might complain **after** the merge?

Should we simply rewrite commit metadata & push it with different (e.g. mine) author reference? Or can we merge?

waynebeaton:
> Considering the fact that at the moment of PR creation ECA check was green, I assume we are OK to merge from the legal point of view.

Correct.

Whether or not the tool will complain... I don't know. But if you can merge, then you should be good-to-go. AFAIK, there's no automatic checks after merge.

When we do one of our periodic checks (that we usually just do quietly), we will detect that the author was covered by an ECA (probably an MCCA based on the scenario) at the time

iloveeclipse (approved):

iloveeclipse:
Thanks @waynebeaton for a quick response!

PR #2803, opened 2025-02-18
Simplify concurrent AbstractTextSearchResult.fListeners

jukzi:
Avoids synchronization on every use.

Slight difference: the pathetic case of adding a listener multiple times would now also result in calling the listener as many times.


akurtakov:
@iloveeclipse Do you want to take a look at this one too?

iloveeclipse:
> @iloveeclipse Do you want to take a look at this one too?

It was all good before I've tried to resolve it with github web ui. That thing did a merge instead of rebase & merge.

iloveeclipse:
> > @iloveeclipse Do you want to take a look at this one too?
> 
> It was all good before I've tried to resolve it with github web ui. That thing did a merge instead of rebase & merge.

I've rebased & fixed merge conflict in the IDE.

iloveeclipse (approved):

PR #2808, opened 2025-02-20
Provide Explicit "isDarkTheme" Attribute

BeckerWdf:
Up to now we had some heuristics to check if a theme is a dark theme. This now makes this explicit.

- Add attribute to extension point
- Fill this attribute for the platforms dark theme
- Remove heuristics and now check this explicit flag

See also https://github.com/eclipse-platform/eclipse.platform.ui/pull/2797#issue-2850971179

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/Theme.java
+++ b/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/Theme.java
@@ -19,6 +19,7 @@ public class Theme implements ITheme {
 	private String id;
 	private String label;
 	private String osVersion;
+	private String isDark;
 
 	public Theme(String id, String label) {
 		this.id = id;
@@ -44,9 +45,24 @@ public class Theme implements ITheme {
 	}
 
 	@Override
+	public boolean isDark() {
+		if (isDark == null) {
+			// fallback for themes that don't yet set the "isDark" attribute
+			// will be removed in a later release
+			return id.contains("dark");
+		} else {
+			return Boolean.parseBoolean(isDark);
+		}
+	}
+
+	public void setIsDark(String isDark) {
+		this.isDark = isDark;
+	}
+
+	@Override
 	public String toString() {
 		return "Theme [id=" + id + ", label='" + label + "', osVersion="
-				+ osVersion + "]";
+				+ osVersion + " isDark=" + isDark + "]";
 	}
 
 
--- a/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
+++ b/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
@@ -81,6 +81,7 @@ public class ThemeEngine implements IThemeEngine {
 	private HashMap<String, List<IResourceLocator>> sourceLocators = new HashMap<>();
 
 	private static final String THEMEID_KEY = "themeid";
+	private static final String THEME_IS_DARK_KEY = "themeIsDark";
 
 	public static final String THEME_PLUGIN_ID = "org.eclipse.e4.ui.css.swt.theme";
 
@@ -121,6 +122,7 @@ public class ThemeEngine implements IThemeEngine {
 						String id = ce.getAttribute("id");
 						String os = ce.getAttribute("os");
 						String version = ce.getAttribute("os_version");
+						String isDarkTheme = ce.getAttribute("isDarkTheme");
 
 						/*
 						 * Code to support e4 dark theme on Mac 10.13 and older. For e4 dark theme on
@@ -154,7 +156,8 @@ public class ThemeEngine implements IThemeEngine {
 							basestylesheeturi = "platform:/plugin/" + ce.getContributor().getName() + "/"
 									+ basestylesheeturi;
 						}
-						registerTheme(themeId, label, basestylesheeturi, version);
+
+						registerTheme(themeId, label, basestylesheeturi, version, isDarkTheme);
 
 						//check for modified files
 						if (modifiedFiles != null) {
@@ -250,11 +253,11 @@ public class ThemeEngine implements IThemeEngine {
 	@Override
 	public synchronized ITheme registerTheme(String id, String label, String basestylesheetURI)
 			throws IllegalArgumentException {
-		return  registerTheme(id, label, basestylesheetURI, "");
+		return registerTheme(id, label, basestylesheetURI, "", "");
 	}
 
 	public synchronized ITheme registerTheme(String id, String label,
-			String basestylesheetURI, String osVersion) throws IllegalArgumentException {
+			String basestylesheetURI, String osVersion, String isDark) throws IllegalArgumentException {
 		for (Theme t : themes) {
 			if (t.getId().equals(id)) {
 				throw new IllegalArgumentException("A theme with the id '" + id
@@ -264,6 +267,9 @@ public class ThemeEngine implements IThemeEngine {
 		Theme theme = new Theme(id, label);
 		if (osVersion != "") {
 			theme.setOsVersion(osVersion);
+		}
+		if (isDark != "") {
+			theme.setIsDark(isDark);
 		}
 		themes.add(theme);
 		registerStyle(id, basestylesheetURI, false);
@@ -496,6 +502,7 @@ public class ThemeEngine implements IThemeEngine {
 			EclipsePreferencesHelper.setCurrentThemeId(theme.getId());
 
 			pref.put(THEMEID_KEY, theme.getId());
+			pref.putBoolean(THEME_IS_DARK_KEY, theme.isDark());
 			try {
 				pref.flush();
 			} catch (BackingStoreException e) {
@@ -560,6 +567,10 @@ public class ThemeEngine implements IThemeEngine {
 		return getPreferences().get(THEMEID_KEY, null);
 	}
 
+	private boolean getPreferenceThemeIsDark(boolean fallback) {
+		return getPreferences().getBoolean(THEME_IS_DARK_KEY, fallback);
+	}
+
 	private IEclipsePreferences getPreferences() {
 		return InstanceScope.INSTANCE.getNode(FrameworkUtil.getBundle(ThemeEngine.class).getSymbolicName());
 	}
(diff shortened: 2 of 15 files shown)


akurtakov:
My main concern with such "api" is what happens with e.g. gray #808080  are they dark or light? What about some dark(-ish) "blue"? It becomes a very thin line to decide with such boolean.

sratz:
I think it's pretty common for themes to be either dark or light. At least all major operating systems also have this kind of primary boolean distinction.

akurtakov:
@sratz I agree that OSes(GTK I know has it for sure) have such boolean distinction and I am not against this addition, just raising some concern about theming in general (not only in Eclipse) for which no one found better solution yet.

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/theme/ITheme.java
```diff
@@ -26,4 +26,10 @@ public interface ITheme {
 	 * @return the label
 	 */
 	String getLabel();
+
+	/**
+	 *
+	 * @return does the theme use a dark background color
+	 */
+	boolean getIsDarkTheme();
```
mickaelistria: What about just `isDarkTheme` ?

merks: Yes. Maybe even `isDark` is sufficient...

BeckerWdf: agree. See [b93d493](https://github.com/eclipse-platform/eclipse.platform.ui/commit/b93d49347e3ab9f934d2205ec658e4409e042e9c)

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/Theme.java
```diff
@@ -43,10 +44,19 @@ public String getOsVersion() {
 		return this.osVersion;
 	}
 
+	@Override
+	public boolean IsDark() {
```
merks: It should be `isDark`.

BeckerWdf: ah sure...

BeckerWdf: see [e4d78b0](https://github.com/eclipse-platform/eclipse.platform.ui/commit/e4d78b04722d351282c43576242bf81758ecb7f6)

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
```diff
@@ -607,7 +618,7 @@ public void restore(String alternateTheme) {
 				 * this case want to fall back to respect whether that previous choice was dark
 				 * or not. https://github.com/eclipse-platform/eclipse.platform.ui/issues/2776
 				 */
-				overrideWithDarkTheme = prefThemeId.contains("dark");
+				overrideWithDarkTheme = getPreferenceThemeIsDark();
```
sratz: here we lose the fallback in case the explicit new isDark preference is not (yet) set in an older installation.
I think we should let `getPreferenceThemeIsDark()` return a `Boolean` and in case of `null`, fall back to the heuristics.

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/Theme.java
```diff
@@ -43,10 +44,19 @@ public String getOsVersion() {
 		return this.osVersion;
 	}
 
+	@Override
+	public boolean isDark() {
+		return Boolean.parseBoolean(isDark);
```
sratz: for older theme contributions that have not yet updated to the explicit isDark attribute, we should probably also fall back to the heuristic here to maximize compatibility, otherwise, all legacy themes would suddenly become light.

BeckerWdf: > for older theme contributions that have not yet updated to the explicit isDark attribute, we should probably also fall back to the heuristic here to maximize compatibility, otherwise, all legacy themes would suddenly become light.

And when will we get rid of that fallback?

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
BeckerWdf: But how do we knot if the preference is not (yet) set?
JavaDoc of the get-methods say:

> Returns the value associated with the specified key in this node. Returns the specified default if there is no value associated with the key, or the backing store is inaccessible.

sratz: Ah, then something like:

```java
	private boolean getPreferenceThemeIsDark(boolean fallback) {
		return getPreferences().getBoolean(THEME_IS_DARK_KEY, fallback);
	}
```

```java
    ...
    overrideWithDarkTheme = getPreferenceThemeIsDark(prefThemeId.contains("dark"));
    ....
```

BeckerWdf: ah sure. That's a good idea.

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/Theme.java
sratz: Hmm, for a few releases?

We should announce this in the new-and-noteworthy, that themes may now be explicitly set to dark/light and therefore there is no need to have `...dark...` in the ID anymore.

Additionally, consumers are required to explicitly set this flag now as we want to remove the fallback with release X.

X being derived from the same rules as when we mark API as `forRemoval`.

BeckerWdf: see 8fe3319

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
BeckerWdf: see 8fe3319

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/Theme.java
BeckerWdf: > Hmm, for a few releases?
> 
> We should announce this in the new-and-noteworthy, that themes may now be explicitly set to dark/light and therefore there is no need to have `...dark...` in the ID anymore.
> 
> Additionally, consumers are required to explicitly set this flag now as we want to remove the fallback with release X.
> 
> X being derived from the same rules as when we mark API as `forRemoval`.

We can try this.

mickaelistria:
I'm still torn about whether "isDark" should be further declarative. Could we instead try to get the default background color from the them, and resolve whether the background color is dark or not; using eg the usual formula to convert RGB to greyscale (0.3 * R + 0.59 * B + 0.11 * G) and evaluating whether this grey is dark (eg < 122) ?

BeckerWdf:
> I'm still torn about whether "isDark" should be further declarative. Could we instead try to get the default background color from the them, and resolve whether the background color is dark or not; using eg the usual formula to convert RGB to greyscale (0.3 * R + 0.59 * B + 0.11 * G) and evaluating whether this grey is dark (eg < 122) ?

But that would replace one heuristic with another one. I think it's quite good to make this explicit. And keep in mind we don't have tons of themes in platform (and out there in the community). Adapting them should be easy.

BeckerWdf:
> get the default background color from the theme

What should that be? Pls. explain based on the existing themes which color definition would be the "default background color".

mickaelistria:
> But that would replace one heuristic with another one. 

Is this really just an heuristic or is the greyscale value the actual definition of whether a theme is dark or not ?
Note that I'm not against allowing people to explicit/override whether their theme is to be treated as dark or not if they want to; but I find very strange that the default decision made by the Platform wouldn't be sufficient in most cases. We can't expect every theme to adopt the extension (we do not control all the themes from the community, many of theme are probably built-in inside some low-maintained RCP apps...); and we cannot just force burden to theme developers about forcing them to set an extra flag if the value can be reliably deduced from the payload in the vast majority of cases.

mickaelistria:
> What should that be? Pls. explain based on the existing themes which color definition would be the "default background color".

Probably something like `Composite > background-color`; or even better way to infer it would be to be able to load the theme, create a dummy invisible shell, apply the theme on it and query the background color.

BeckerWdf:
> > What should that be? Pls. explain based on the existing themes which color definition would be the "default background color".
> 
> Probably something like `Composite > background-color`; or even better way to infer it would be to be able to load the theme, create a dummy invisible shell, apply the theme on it and query the background color.

To be honest I find this overcomplicated. For example we have code like this:

`boolean hasDarkTheme = getThemes().stream().anyMatch(t -> t.isDark());`

Do you really like to create a dummy invisible shell for each of the themes when the alternative is setting a little flag in the plugin.xml? And we don't have that many themes anyway. So I think the the cost / benefit relation is not very good for that.

sratz:
I agree with @BeckerWdf. Why implement a guessing algorithm when we can make it explicit?

Other IDEs also have an explicit flag, e.g. VS Code:
https://github.com/microsoft/vscode/blob/544674f984e97b8aa9a444cb5487cfc4c27f2b7a/src/vs/platform/theme/common/theme.ts#L6-L14

akurtakov:
@BeckerWdf Would you please rebase this one to fix the conflict?

Mailaender:
There is precedent for a boolean setting in Eclipse already at the text editor themes:

grafik

mickaelistria:
I would not take this part of TM4E as an example to rely on (it's not bad as it does the job well, but it didn't got through brainstorming to find the best sustainable solution).

akurtakov:
Closing as duplicate of https://github.com/eclipse-platform/eclipse.platform.ui/pull/4354 .

PR #2811, opened 2025-02-20
Don't remove parts from part stack on DnD drop

### LINKED ISSUE #2771: Part with "allow multiple" created dynamically disappears after moving back to PartStack
Hi,

with a few changes to a generated "Eclipse RCP application" project I can show the disappearing of parts.
The changes include the creation of a PartDescriptor for the generated view. The PartDescriptor allows 
multiple instances. The part is also removed from the PartStack in e4xmi.
In the generated Open-handler the part is dynamically created like this:
- find a PartStack
- if found
-- make it visible (in case the PartStack was minimized or is not visible after the last part closed)
-- create the part and add it as a child to the PartStack's children.
-- show the created part

This works as expected.

How a part disappears:
- moving a part from the PartStack to be next to the other parts of the PartStack
- moving the part back to the PartStack leads to one part (in the PartStack) being replaced with the moving part
- one part disappears

After digging around for a while I found that in StackDropAgent.dock(...) the part is removed
if the part's elementId is equal to a part's elementId in the PartStack (see variable "viewWithSameId").
Reading the comments in the code the intention was to remove Placeholder parts - if present.
As far as I see multiple instance of one part are not considered when searching for present Placeholder parts.

A fix could be to additionally test whether the part in the PartStack (variable "stackElement") is of type MPlaceholder and
only if it is a Placeholder instance _and_ the elementId's are equal it should be removed.

For this very small sample project I could test the described fix successfully. The parts can be moved around and back to the PartStack without disappearing.

What do you think?

Thanks.

OS: linux, Debian bookworm
Eclipse: 4.35

p-O-q:
Currently when dropping elements onto a part stack MStackElements are removed if their element ids are equal. This leads to disappearing parts in case of
- at least one part created using part descriptors allowing multiple instances is present in the part stack
- another such part is dropped onto the part stack

Following the comments/notes the intention is to remove place holder parts with the same element id as the element being dropped.

This fix ensures that only instance of type MPlaceholder are removed from the part stack.

Fixes #2771

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.addons.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.workbench.addons.swt/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.workbench.addons.swt;singleton:=true
-Bundle-Version: 1.5.600.qualifier
+Bundle-Version: 1.5.700.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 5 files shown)


p-O-q:
Closed b/c I missed to implement a test and left unnecessary code in test class.

PR #2816, opened 2025-02-25
Modernize to Java 17 switch expression.

deepika-u:
Update legacy switch block with Java 17 switch expression in org.eclipse.ui.workbench.texteditor package.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -78,18 +78,16 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 		}
 
 		switch (searchOption) {
-		case GLOBAL:
-			unsetSearchScope();
-			break;
-		case FORWARD:
-		case INCREMENTAL:
-			if (shouldInitIncrementalBaseLocation()) {
-				resetIncrementalBaseLocation();
+			case GLOBAL -> unsetSearchScope();
+			case FORWARD, INCREMENTAL -> {
+				if (shouldInitIncrementalBaseLocation()) {
+					resetIncrementalBaseLocation();
+				}
 			}
-			break;
-		// $CASES-OMITTED$
-		default:
-			break;
+			// $CASES-OMITTED$
+			default -> {
+				// No action needed; keeping an explicit default for clarity
+			}
 		}
 	}
 
@@ -131,19 +129,12 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 
 	@Override
 	public boolean isAvailable(SearchOptions searchOption) {
-		switch (searchOption) {
-		case REGEX:
-			return isTargetSupportingRegEx;
-		case WHOLE_WORD:
-			return !isAvailableAndActive(SearchOptions.REGEX) && isWord(findString);
-		case INCREMENTAL:
-		case CASE_SENSITIVE:
-		case FORWARD:
-		case GLOBAL:
-		case WRAP:
-		default:
-			return true;
-		}
+		return switch (searchOption) {
+			case REGEX -> isTargetSupportingRegEx;
+			case WHOLE_WORD -> !isAvailableAndActive(SearchOptions.REGEX) && isWord(findString);
+			case INCREMENTAL, CASE_SENSITIVE, FORWARD, GLOBAL, WRAP -> true;
+			default -> true;
+		};
 	}
 
 	@Override
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicMessageGenerator.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicMessageGenerator.java
@@ -41,22 +41,12 @@ public class FindReplaceLogicMessageGenerator implements IFindReplaceStatusVisit
 	@Override
 	public String visit(FindStatus status) {
 		FindStatus.StatusCode messageCode = status.getMessageCode();
-		String message;
-		switch (messageCode) {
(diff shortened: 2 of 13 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/texteditor/rulers/RulerColumnPlacement.java
```diff
@@ -67,18 +67,14 @@ private Set<RulerColumnPlacementConstraint> readIds(IConfigurationElement[] chil
...
-			default:
-				childHelper.fail(RulerColumnMessages.RulerColumnPlacement_illegal_child_msg);
-				continue;
-			}
+			boolean before = switch (name) {
+				case AFTER -> false;
+				case BEFORE -> true;
+				default -> {
+					childHelper.fail(RulerColumnMessages.RulerColumnPlacement_illegal_child_msg);
+					yield false; // Required because default must return a value.
```
iloveeclipse: This looks wrong. Is this autogenerated or manually changed?
Original code would continue for loop without adding anything to constraints.

iloveeclipse:
I'm not sure the change improves readability.
Is this PR manually crafted of is the result of automated quick fix?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/texteditor/rulers/RulerColumnPlacement.java
deepika-u: I had a doubt only in this context. So here it cannot be written in newer syntax does it mean?

deepika-u:
> I'm not sure the change improves readability. Is this PR manually crafted of is the result of automated quick fix?

I have did it manually.

iloveeclipse:
> I have did it manually

Why not with a cleanup? Or there is nothing available?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/texteditor/rulers/RulerColumnPlacement.java
iloveeclipse: No idea, have no time for this.

deepika-u: can this loop be like this to achieve the loop?
```
boolean before = switch (name) {
    case AFTER -> false;
    case BEFORE -> true;
    default -> {
        childHelper.fail(RulerColumnMessages.RulerColumnPlacement_illegal_child_msg);
        yield false;
    }
};
continue;
```

deepika-u:
> > I have did it manually
> 
> Why not with a cleanup? Or there is nothing available?

Any specific context of "cleanup" in this code are you referring to?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/texteditor/rulers/RulerColumnPlacement.java
iloveeclipse: No!

iloveeclipse (changes requested):
You can perform changes via Source->Cleanup... and the cleanup of your choice.
image

The proposed change breaks code.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/texteditor/rulers/RulerColumnPlacement.java
deepika-u: Let me check on it further.

deepika-u: Reverting this to original switch, thanks for the suggestion.

deepika-u:
> You can perform changes via Source->Cleanup... and the cleanup of your choice. 
> 

Thanks for the suggestion, even via this that file is not getting picked and updated. So i have reverted that file change(switch to original code). 
Can you review now please?

iloveeclipse:
> even via this that file is not getting picked and updated

That is the sign that the quick fix doesn't have an automated solution for conversion (yet).

> Can you review now please?

The new change shows some new warning, shows files with whitespace only changes. That shouldn't happen.

I only briefly checked **original PR** whether the proposed change make sense, found immediately a problem and beside the problem nothing interesting in the change itself. Approving a PR means I agree with the change, which is not the case here. 

I would propose to close this PR, and if you are eager to apply this particular refactoring on this particular bundle, do this with the automated cleanup **and only with that**.

For the new PR please find someone who is interested in such kind of reviews and let review it. I'm definitely not interested in this particular change.

akurtakov:
What is the status of this one?

deepika-u:
@akurtakov 
Did already addressed all the review comments regarding the code changes.
But the above code changes are from automated clean up and additional manual changes(which were missed by automated clean up tool) as well.

Updating the automated clean up - i agree needs to be improved but not sure how and who does it exactly(may be via JDT or JDT ui) which can be taken up separately.

So when you get some free time, can you review these changes please?

akurtakov:
@iloveeclipse Your change request is pending here. Is the PR good according to you?

iloveeclipse:
> @iloveeclipse Your change request is pending here. Is the PR good according to you?

It is not mine and I've answered already above: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2816#issuecomment-2684899239

akurtakov:
> > @iloveeclipse Your change request is pending here. Is the PR good according to you?
> 
> It is not mine and I've answered already above: [#2816 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2816#issuecomment-2684899239)

image

iloveeclipse:
I prefer to close this PR, I don't see any value in this change and the time ro properly review is not worth it.

PR #2821, opened 2025-03-03
Automatic cleanup with - Use StringBuilder instead of StringBuffer

deepika-u:
Used StringBuilder for better performance instead of StringBuffer.

### DIFF
--- a/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserManager.java
+++ b/bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserManager.java
@@ -158,7 +158,7 @@ public class BrowserManager extends Observable {
 
 			StringWriter writer = new StringWriter();
 			memento.save(writer);
-			String xmlString = writer.getBuffer().toString();
+			String xmlString = new StringBuilder(writer.toString()).toString();
 			IScopeContext instanceScope = InstanceScope.INSTANCE;
 			IEclipsePreferences prefs = instanceScope.getNode(WebBrowserUIPlugin.PLUGIN_ID);
 			prefs.put("browsers", xmlString); //$NON-NLS-1$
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserManager.java
```diff
@@ -158,7 +158,7 @@ protected void saveBrowsers() {
 
 			StringWriter writer = new StringWriter();
 			memento.save(writer);
-			String xmlString = writer.getBuffer().toString();
+			String xmlString = new StringBuilder(writer.getBuffer().toString()).toString();
```
merks: This seems bizarre.

merks (changes requested):
This just looks complicated and misguided.  Changing it to remove the `getBuffer()` call would be more sensitive:

image

laeubi:
> Used StringBuilder for better performance instead of StringBuffer.

I can't see any `StringBuffer` > `StringBuilder` conversion here ... @deepika-u please *always* review automatic cleanups and if you find any issues report them to JDT so then can be fixed.

bundles/org.eclipse.ui.browser/src/org/eclipse/ui/internal/browser/BrowserManager.java
deepika-u: Will be reverting this change.

merks: Changing `writer.getBuffer().toString()` to `writer.toString()` to hide that a buffer is involved would seem okay/better.

deepika-u: sure.

iloveeclipse: > This seems bizarre.

@deepika-u : was this automatically done? If so, could you please create jdt.ui isssue for that?

iloveeclipse:
The PR contains **only** really weird changes. 
*If* that is what proposed by jdt.ui tooling, this should be fixed in JDT UI and a ticket should be crated.

PR itself without these changes makes no sense => closing

deepika-u:
Thanks for your opinion and inputs on this.  I too got the intent of it. Sorry to cause confusion here.

PR #2825, opened 2025-03-04
[E4] Migrate o.e.e4.ui.w.addons.swt.test to JUnit5

p-O-q:
Migration of test bundle org.eclipse.e4.ui.workbench.addons.swt.test from JUnit4 to JUnit5.

### DIFF
--- a/tests/org.eclipse.e4.ui.workbench.addons.swt.test/src/org/eclipse/e4/ui/workbench/addons/cleanupaddon/CleanupAddonTest.java
+++ b/tests/org.eclipse.e4.ui.workbench.addons.swt.test/src/org/eclipse/e4/ui/workbench/addons/cleanupaddon/CleanupAddonTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 Vector Informatik GmbH and others.
+ * Copyright (c) 2024, 2025 Vector Informatik GmbH and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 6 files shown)


tests/org.eclipse.e4.ui.workbench.addons.swt.test/src/org/eclipse/e4/ui/workbench/addons/cleanupaddon/CleanupAddonTest.java
```diff
@@ -191,8 +192,8 @@ public Image adornImage(Image toAdorn, Image adornment) {
 	@Test
 	public void testRemovingPrimaryDataStackTransfersToRemainingStack() {
 		assumeFalse(
-				"not working reliably on Mac, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/1784",
-				OS.isMac());
+				OS.isMac(),
```
akurtakov: Have you considered using "native" Junit 5 way like https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/condition/DisabledOnOs.html ? Current code is fine, making code less Eclipse specific and thus more familiar to people is supposed to make the codebase easier for people to join.

p-O-q: @akurtakov Thank you for your suggestion! :)
No, actually I wasn't aware of it. I'd like to take a look at this and adapt the code accordingly. I think I will push the changes tomorrow.

akurtakov: There is a chance that JUnit integration in Eclipse or build reporting tools don't handle it well but even in that case it would help us improve junit integration (if there is and issue) so it could be only a bonus point.

p-O-q:
@akurtakov  As far as I can see, the warnings are not being introduced by my PR. I don't think there is anything I can do at the moment.
Please let me know if I am wrong or if I can help.

akurtakov:
Warnings are due to new stream opening and should be fixed with https://github.com/eclipse-platform/eclipse.platform.ui/pull/2826 . I'm rebasing this one.

akurtakov:
Thanks for this one! Much appreciated.

p-O-q:
@akurtakov Thank you for the effort you put into this PR :)

PR #2827, opened 2025-03-05
Change the NewWizard_title back to New

merks:
- Avoid using the description also as a title.

https://github.com/eclipse-platform/eclipse.platform/issues/1749


merks:
The failures appear unrelated.

PR #2828, opened 2025-03-05
Adjusting the Size of Splash Shell a/c to image

### LINKED ISSUE #219: The startup dialog in Eclipse is cutoff
Every time I start the application, the startup dialog it cutoff (with the quarter scaling check enabled). It start occurring to me ever since I updated my eclipse to Version: 2025-03 (4.35).

~The problem has been introduced sometime between these two I-BUILDS (https://github.com/vi-eclipse/Eclipse-Platform/issues/219#issuecomment-2647307845):~
- ~**20250106-1800** 🟢~
- ~**20250120-0740** 🔴~

Image

## To reproduce
- ~Primary Monitor: 150%, Secondary: 100%~ UPDATE: 1 monitor at 150% is enough to reproduce it
- Add `-Dswt.autoScale.updateOnRuntime=true` to `eclipse.ini`

## To debug remotely
- Add this to the `eclipse.ini` of the instance you want to debug
```
-Xdebug
-Xrunjdwp:server=y,transport=dt_socket,address=8833,suspend=y
-Dcom.sun.management.jmxremote
-Dcom.sun.management.jmxremote.port=9966
-Dcom.sun.management.jmxremote.authenticate=false
-Dcom.sun.management.jmxremote.ssl=false
```
(make sure the ports `8833` and `9966` are not in use and use `suspend=y` so you have enough time to attach a remote debugger)
- Create a **remote debug configuration** in the IDE that you use to debug and add the corresponding port (`8833` in my case)

Image

- Start the instance you want to debug
- Start the remote debug configuration

## To reproduce from the IDE itself
- Use the `platform.product` that is in the plugin `org.eclipse.platform.releng.tychoeclipsebuilder` (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product). That one has a splash screen too.

ShahzaibIbrahim:
The splash is initialized by the Equinox native launcher while image is set by SWT. SWT uses "correctly scaled images". This inconsistency leads to image cut off during start up.

image

### Proposed Solution

We are adjusting the size of the shell with the size of the image in case of inconsistency.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
@@ -820,8 +820,14 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 					if (splashShell == null)
 						return;
-					if (background != null)
+					if (background != null) {
 						splashShell.setBackgroundImage(background);
+						Rectangle imageBounds = background.getBounds();
+						Rectangle shellBounds = splashShell.getBounds();
+						if (imageBounds.width > shellBounds.width || imageBounds.height > shellBounds.height) {
+							splashShell.setSize(imageBounds.width, imageBounds.height);
+						}
+					}
 				}
 
 				Dictionary<String, Object> properties = new Hashtable<>();


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -820,8 +820,14 @@ public void run() throws Exception {
 
 					if (splashShell == null)
 						return;
-					if (background != null)
+					if (background != null) {
 						splashShell.setBackgroundImage(background);
+						Rectangle imageBounds = background.getBounds();
+						Rectangle shellBounds = splashShell.getBounds();
+						if (imageBounds.width > shellBounds.width || imageBounds.height > shellBounds.height) {
+							splashShell.setSize(imageBounds.width, imageBounds.height);
```
HeikoKlare, line 828: Isn't the issue that the image is too large/small for the shell (because Platform UI applies quarter scaling while Equinox native launcher uses integer200) and not that the shell does not fit to the image? Changing the shell size here will result in a rescaling of the splash during application startup, won't it?

HeikoKlare (commented):
To precise about the issue: if I am not mistaken, this _only_ happens with monitor-specific scaling enabled (or even when only `quarter` scaling is activated).

fedejeanne:
This PR currently resizes the shell to fit the image (see GIF below, made in a 125% monitor):

splash_resized

I'm in favor of keeping the initial size of the shell.

PR #2832, opened 2025-03-11
Make ContentTypes->default text editor as Generic Code Editor.

### LINKED ISSUE #2527: Make Generic Code Editor default for text files
Currently the (old) text editor is still the default for text files.

image

My understanding of https://github.com/eclipse-platform/eclipse.platform.ui/pull/1381 and https://github.com/eclipse-platform/eclipse.platform.ui/issues/1385 is that we agree that the Generic Code editor is the better choice. 

Should we also give the Generic Code Editor priority here? cc @mickaelistria and @iloveeclipse

deepika-u:
Make ContentTypes->default text editor as Generic Code Editor. 

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2527

**Before the pr**
When i come to Content Types page ->
1st screen
image

2nd screen when i select Text
image

3rd screen when i open a text file
image

**After the pr changes**
When i come to Content Types page ->
1st screen
image

2nd screen when i select Text
image

3rd screen when i open a text file
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -1358,6 +1358,13 @@ public class EditorRegistry extends EventManager implements IEditorRegistry, IEx
 		String key = IPreferenceConstants.DEFAULT_EDITOR_FOR_CONTENT_TYPE + contentType.getId();
 		String defaultEditorId = store.getString(key);
 		IEditorDescriptor descriptor = null;
+		if (defaultEditorId.isEmpty() && contentType.getId().contentEquals("org.eclipse.core.runtime.text")) { //$NON-NLS-1$
+			IEditorDescriptor genericEditor = descriptors.stream()
+					.filter(e -> "org.eclipse.ui.genericeditor.GenericEditor".equals(e.getId())) //$NON-NLS-1$
+					.findFirst().orElse(null);
+			store.setValue(key, genericEditor.getId());
+			return genericEditor;
+		}
 		if (defaultEditorId != null && !defaultEditorId.isBlank()) {
 			descriptor = descriptors.stream().filter(d -> defaultEditorId.equals(d.getId())).findFirst().orElse(null);
 		}


BeckerWdf (commented):
can you explain why reversing the list fixes the issue?

deepika-u:
> can you explain why reversing the list fixes the issue?

Could you hold on the review please, i found another bug with my fix. So once that is fixed may be you can review it please.

iloveeclipse:
Change on the preference page alone doesn't change **default** associations. Note, there will be most likely lot of tests failing which expect default test editor be opened if you switch defaults (depending on where and how you will do that).

deepika-u:
> Change on the preference page alone doesn't change **default** associations. Note, there will be most likely lot of tests failing which expect default test editor be opened if you switch defaults (depending on where and how you will do that).

Thanks for the feedback and headsup, even i was expecting some failures because default behavior is changing with the fix. That was the primary reason i have made it a draft than a normal pr.

deepika-u:
@BeckerWdf 
org.eclipse.ui.tests.dialogs.ResourceInitialSelectionTest.testSingleSelectionAndTwoInitialSelectionsWithInitialPattern is locally passing with the pr on mac
image (3)

I have tested this issue completely and also tried with xml, html files. They are not impacted.
Can you review now please?

deepika-u:
@vogella @BeckerWdf 
Can you review this pr please when you get some time?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
```diff
@@ -1358,6 +1358,13 @@ private IEditorDescriptor selectDefaultEditor(List<IEditorDescriptor> descriptor
 		String key = IPreferenceConstants.DEFAULT_EDITOR_FOR_CONTENT_TYPE + contentType.getId();
 		String defaultEditorId = store.getString(key);
 		IEditorDescriptor descriptor = null;
+		if (defaultEditorId.isEmpty() && contentType.getId().contentEquals("org.eclipse.core.runtime.text")) { //$NON-NLS-1$
+			IEditorDescriptor genericEditor = descriptors.stream()
+					.filter(e -> "org.eclipse.ui.genericeditor.GenericEditor".equals(e.getId())) //$NON-NLS-1$
```
mickaelistria, line 1363: IIRC, org.eclipse.ui.workbench does not depend on the genericeditor plugin, so this would be possibly incorrect.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
```diff
@@ -1358,6 +1358,13 @@ private IEditorDescriptor selectDefaultEditor(List<IEditorDescriptor> descriptor
 		String key = IPreferenceConstants.DEFAULT_EDITOR_FOR_CONTENT_TYPE + contentType.getId();
 		String defaultEditorId = store.getString(key);
```
mickaelistria, line 1359: If you want to change the defaultEditorId for this key, this is probably something to define the preferences; through a preference initializing or some plugin_customization.ini in the product.

mickaelistria (changes requested):
I think this is addressed at the wrong layer. The default editor is not something to be hardcoded but to be configured.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
deepika-u: I am back from my vacation. Let me check on this. Thanks for your inputs.

iloveeclipse: > through a preference initializing or some plugin_customization.ini in the product.

Correct. It would be
```
org.eclipse.ui.workbench/defaultEditorForContentType_org.eclipse.core.runtime.text=org.eclipse.ui.genericeditor.GenericEditor
```
and it would belong to the products

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
```diff
@@ -1358,6 +1358,13 @@ private IEditorDescriptor selectDefaultEditor(List<IEditorDescriptor> descriptor
...
 		IEditorDescriptor descriptor = null;
+		if (defaultEditorId.isEmpty() && contentType.getId().contentEquals("org.eclipse.core.runtime.text")) { //$NON-NLS-1$
```
iloveeclipse, line 1361: As author of this lines I've played with the idea to pick up the generic editor and make it default, but that could potentially break behavior of existing tests or products who depend on plain text editor usage.

Same effect can be achieved by a simple change of the generic editor contribution
```
diff --git a/bundles/org.eclipse.ui.genericeditor/plugin.xml b/bundles/org.eclipse.ui.genericeditor/plugin.xml
index 107f075..1711b4d 100644
--- a/bundles/org.eclipse.ui.genericeditor/plugin.xml
+++ b/bundles/org.eclipse.ui.genericeditor/plugin.xml
@@ -34,3 +34,3 @@
             contributorClass="org.eclipse.ui.editors.text.TextEditorActionContributor"
-            default="false"
+            default="true"
             icon="icons/full/obj16/generic_editor.png"
```

But guess why it wasn't done by default...

iloveeclipse (changes requested):
I don't think this is the right way to unconditionally change user preferences in a getter method.

There are two different other, less hacky solutions proposed in comments, please consider them as the path forward.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
deepika-u: Yes, org.eclipse.ui.workbench plugin does have neither org.eclipse.workbench.texteditor or org.eclipse.ui.genericeditor in its plugin dependencies.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/registry/EditorRegistry.java
deepika-u: > But guess why it wasn't done by default...

@iloveeclipse 
I think this is the simplest ever possible solution. It works so awesome. But why it should not be used this way?

mickaelistria: What happens if another editor also decides to declares itself as "default" ? Overall, this "default" flag is a mistake from the beginning: in an extensible environment (may it be the IDE, the OS...), nothing can pretend to be *the* default for task X as other competitive, and maybe better, solutions can be provided at any point of time. Setting default=true here isn't much better than harcoding it in the code, as it makes extensibility more complex.
The choice of the "best" or "recommended" default becomes more a choice of packaging (set in .product) or preference (set in preference page).

That was my dogmatic answer (which I still think is the only correct one). That said, I don't care that much and if most people want to change the flag here, I wouldn't see it as a big harm; as long as people who make the change are ready to face the criticism from the people who have set `default=true` in their RCP and might now see a different behavior (but maybe pushing them to use the Generic Editor is likely to help them on the longer run...).

iloveeclipse: > It works so awesome. But why it should not be used this way?

See above: that could potentially break behavior of existing tests or products who depend on plain text editor usage.

However, thinking once more about it while answering here, now with https://github.com/eclipse-platform/eclipse.platform.ui/issues/1659 fixed via https://github.com/eclipse-platform/eclipse.platform.ui/pull/1665 one could argue that customers expecting "old classic" text editor to be opened by default could simply customize their products by adding this line in product customization file:

`org.eclipse.ui.workbench/defaultEditorForContentType_org.eclipse.core.runtime.text=org.eclipse.ui.DefaultTextEditor`

Given that, the change would be not breaking because we would have a viable workaround. I've pushed https://github.com/eclipse-platform/eclipse.platform.ui/pull/2881, let see what tests say.

iloveeclipse:
Replaced by https://github.com/eclipse-platform/eclipse.platform.ui/pull/2881

PR #2838, opened 2025-03-14
Revert "spare control characters from content assist trigger"

### LINKED ISSUE #2723: ContentAssistProcessor no longer activated when CTRL key is pressed

## Steps to reproduce

- Create a custom source viewer with a default ContentAssist.
- Register a custom ContentAssistProcessor that is automatically activated when pressing space.
- Activate the processor by pressing CTRL+Space

I expected: The processor to be activated

But got: The processor remains inactive

* Windows 11
* Eclipse 2024-12

## Workaround

```java
ContentAssistant assistant = new ContentAssistant() {
    @Override
    protected AutoAssistListener createAutoAssistListener() {
        return new AutoAssistListener() {
            @Override
            public void keyPressed(KeyEvent e) {
                if (e.stateMask == SWT.CONTROL) {
                    e.stateMask = 0;
                }
                super.keyPressed(e);
            }
        };
    }
};
```

## Community

This is a result of https://github.com/eclipse-platform/eclipse.platform.ui/pull/1556 in an attempt to fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/891. As part of this change, any events where the CTRL key is pressed will no longer activate the content processor.

In hindsight, I wonder if it would've been a better approach to create a subclass for the AutoAssistListeners and use it wherever state masks are undesirable.

ptziegler:
This reverts commit aa42db11608ae15948c45836be3bb4a0a6d3827d, as the change has to be done on the JDT side [1].

[1] https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2086

Closes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2723

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/ContentAssistant.java
@@ -340,7 +340,7 @@ public class ContentAssistant implements IContentAssistant, IContentAssistantExt
 			if (e.character == 0 && (e.keyCode & SWT.KEYCODE_BIT) == 0)
 				return;
 
-			if (e.character != 0 && (e.stateMask == SWT.ALT || e.stateMask == SWT.CONTROL || e.stateMask == SWT.COMMAND))
+			if (e.character != 0 && (e.stateMask == SWT.ALT))
 				return;
 
 			TriggerType triggerType= getAutoActivationTriggerType(e.character);


ptziegler:
This side is looking good. But until I get a response from the JDT team, this is effectively blocked unless I want to reintroduce the original bug.

ptziegler:
Occasionally a warning may still pop up when deleting files. But this seems to be a false positive caused by rounding errors.  Perhaps it makes sense to make to only show this warning if the diff is significant (i.e. larger than 1e-15)?

image

PR #2846, opened 2025-03-19
Restore original Problems view sort order for severity and priority

### LINKED ISSUE #2845: Wrong markers sort order for severity and priority values
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1733.

Since 4.32 users can't (immediately) see errors if the most of reported problems are warnings. 

A picture says more then words:

Image

This is especially bad if the number of warnings exceeds 100 (default view limit) so the errors not shown at all :-(

The reason is that the default sort order changed in  https://github.com/eclipse-platform/eclipse.platform.ui/pull/1733, sorting higher values last (which is "usual" for generic values but unusual for problems and priorities where the highest values should be sorted first).

I'm going to provide a fix.

iloveeclipse:
Regression from f13fb4997f5e7849da26285938a3a0aea1769747.

One part of the regression was already fixed via
935b1dba41dd1e8f52a840c5057ae1eb925db159 (in `QuickFixPage`).

This change fixes remaining regressions that affect sort order of
markers where severity and priority attributes are involved.

Markers with higher values of severity/priority should be always sorted
first by default (the opposite to the "natural" sort order), because we
want that the *most important* markers are shown to the user by default
(without scrolling and filtering).

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2845

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerEntry.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerEntry.java
@@ -410,4 +410,31 @@ class MarkerEntry extends MarkerSupportItem implements IAdaptable {
 		MarkerEntry other = (MarkerEntry) obj;
 		return Objects.equals(marker, other.marker);
 	}
+
+	@Override
+	public String toString() {
+		StringBuilder builder = new StringBuilder();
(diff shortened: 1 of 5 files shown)


iloveeclipse:
Mac test fail in ResourceInitialSelectionTest is unrelated, merging.

PR #2856, opened 2025-03-25
Suppress warnings for Key Class

elsazac:
This PR adds @SuppressWarnings("deprecation") to avoid warnings caused by the usage of the deprecated Key class, which is marked for removal since 2024-03.


elsazac:
I’d like to proceed with merging this if there are no objections

PR #2858, opened 2025-03-25
Proper rescaling for win32 splash screen image

### LINKED ISSUE #219: The startup dialog in Eclipse is cutoff

akoch-yatta:
This commit adds an ImageDataProvider to properly scale the splash image, if it is shown in Windows. If the monitor specific scaling is active, the autoScale mode used when the native launcher calculated the initial size was calculated, might be different than when the splash image is refreshed via the Workbench. The ImageDataProvider ensures, that either the explicitly defined mode or the default mode is used.

### How to test

As I am not aware of an easy way to test the splash screen logic on runtime, I created a small Snippet to test the added ImageDataProvider.
You need to copy the splash.png from org.eclipse.platform next to this Snippet.

```java
public class SnippetSplash {

public static void main (String[] args) {
	Display display = new Display( );
	Shell shell = new Shell (display, SWT.NONE);
	shell.setText("Snippet Splash");
	shell.setLayout (new GridLayout ());
	final Image image = new Image (display, SnippetSplash.class.getResourceAsStream ("splash.png"));
	Image background = new Image(display, new ImageDataProvider() {
		@Override
			int convertedZoom = DPIUtil.getZoomForAutoscaleProperty(zoom,
					null);
			return image.getImageData(convertedZoom);
		}
	});
	shell.setBackgroundImage(image);
	int multiplier = 2; // 2 behaves like autoscale mode integer200 on 175% monitor or bigger, set it to 1 if want to "simulate" the initial splash on smaller zoom
	shell.setBoundsInPixels(100, 100, 452 * multiplier, 302 * multiplier);
	shell.open ();
	while (!shell.isDisposed ()) {
		if (!display.readAndDispatch ()) display.sleep ();
	}
	image.dispose ();
	display.dispose ();
}
}
```

@HeikoKlare feel free to have a look

This PR is an alternative for and thus closes https://github.com/eclipse-platform/eclipse.platform.ui/pull/2828

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
@@ -51,6 +51,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
 import org.eclipse.core.commands.Command;
 import org.eclipse.core.commands.CommandManager;
 import org.eclipse.core.commands.ExecutionException;
@@ -157,6 +158,7 @@ import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.graphics.Transform;
+import org.eclipse.swt.internal.DPIUtil;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Monitor;
@@ -315,7 +317,8 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 		@Override
 		public void serviceChanged(ServiceEvent event) {
 			subMonitor.setWorkRemaining(5).worked(1);
-			spinEventQueueToUpdateSplash(displayForStartupListener);
+			AUTOSCALE_ADAPTATION
+					.runWithInitialAutoScaleValue(() -> spinEventQueueToUpdateSplash(displayForStartupListener));
 		}
 	}
 
@@ -350,6 +353,11 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 	 * The splash handler.
 	 */
 	private static AbstractSplashHandler splash;
+
+	/**
+	 * Helper to adapt auto-scaling for the splash screen
+	 */
+	private static final AutoscaleAdaptation AUTOSCALE_ADAPTATION = new AutoscaleAdaptation();
 
 	/**
 	 * The display used for all UI interactions with this workbench.
@@ -627,9 +635,9 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 						properties);
 
 				// listener for updating the splash screen
-				ServiceListener serviceListener = null;
+				AtomicReference<ServiceListener> serviceListener = new AtomicReference<>();
 				createSplash = WorkbenchPlugin.isSplashHandleSpecified();
-				if (createSplash) {
+				Runnable splashCreation = () -> {
 
 					// prime the splash nice and early
 					workbench.createSplashWrapper();
@@ -649,10 +657,13 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 					if (handler != null && showProgress) {
 						IProgressMonitor progressMonitor = SubMonitor.convert(handler.getBundleProgressMonitor());
-						serviceListener = new Workbench.StartupProgressBundleListener(progressMonitor, display);
-						WorkbenchPlugin.getDefault().getBundleContext().addServiceListener(serviceListener);
+						serviceListener.set(new Workbench.StartupProgressBundleListener(progressMonitor, display));
+						WorkbenchPlugin.getDefault().getBundleContext().addServiceListener(serviceListener.get());
 					}
+				};
 
+				if (createSplash) {
+					AUTOSCALE_ADAPTATION.runWithInitialAutoScaleValue(splashCreation);
 				}
 
 				setSearchContribution(appModel, true);
@@ -661,8 +672,8 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 				if (returnCode[0] == PlatformUI.RETURN_OK) {
 					// run the e4 event loop and instantiate ... well, stuff
-					if (serviceListener != null) {
-						WorkbenchPlugin.getDefault().getBundleContext().removeServiceListener(serviceListener);
+					if (serviceListener.get() != null) {
+						WorkbenchPlugin.getDefault().getBundleContext().removeServiceListener(serviceListener.get());
 					}
 					e4Workbench.createAndRunUI(e4Workbench.getApplication());
 				}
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -912,6 +916,21 @@ private static Image getImage(Display display, InputStream input) {
...
+			// difference between the autoScale mode that is used by the native launcher and
+			// the one that is configured now. The ImageDataProvider ensures either the mode
+			// that is configured by the system property or the default mode (if the system
+			// property is not set) is used for the splash screen image.
+			return new Image(display, new ImageDataProvider() {
+				@Override
+				public ImageData getImageData(int zoom) {
+					@SuppressWarnings("restriction")
+					int convertedZoom = DPIUtil.getZoomForAutoscaleProperty(zoom,
+							System.getProperty(SWT_RESCALE_PROPERTY));
```
HeikoKlare: Instead of exposing the `getZoomForAutoscaleProperty` method and accessing further system properties, couldn't we simply store the original device zoom before modifying the autoscale settings? More precisely: before calling `setRescaleAtRuntimePropertyFromPreference()`, store the value of `DPIUtil.getDeviceZoom()` and use that for scaling the image.

akoch-yatta: Hmm, you mean because the image must always be scaled to that zoom anyway. Yes, you probably have a point. I will try that out

akoch-yatta: I adapted it like mentioned. I am not really happy with storing the zoom in a static field like that, but I don't have a better solution as of now.

HeikoKlare: I agree that the solution is not that nice, but we know that we cannot expect a really great solution given that we implicitly have to align Equinox and SWT assumptions regarding auto-scaling here. To me, the current solution looks acceptable as (1) it is local and internal to the `Workbench` class where also the modification of auto-scaling behavior happens and (2) it is exactly tied to what needs to be done (reflecting that device zoom changes during initialization).

I know this is rather nitpicky but given that we introduce more functionality regarding the autoscaling adaptation (when using monitor-specific scaling) in the `Workbench` class, what about centralizing that in a single class so that you can understand all the related functionality at once. Something like this:
```java
private static class AutoscaleAdaptation {
		private final int initialDeviceZoom;

		AutoscaleAdaptation() {
			this.initialDeviceZoom = getCurrentDeviceZoom();
		}

		public void setRescaleAtRuntimePropertyFromPreference() {
			if (System.getProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY) != null) {
				WorkbenchPlugin.log(Status.warning(SWT_RESCALE_AT_RUNTIME_PROPERTY
						+ " is configured (e.g., via the INI), but the according preference should be preferred instead." //$NON-NLS-1$
				));
			} else {
				boolean rescaleAtRuntime = ConfigurationScope.INSTANCE.getNode(WorkbenchPlugin.PI_WORKBENCH)
						.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, false);
				System.setProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY, Boolean.toString(rescaleAtRuntime));
			}
		}

		public Image adaptSplashImage(Image image) {
			if (getCurrentDeviceZoom() != initialDeviceZoom) {
				return new Image(image.getDevice(), new ImageDataProvider() {
					@Override
						return image.getImageData(initialDeviceZoom);
					}
				});
			}
			return image;
		}

		private static int getCurrentDeviceZoom() {
			return DPIUtil.getDeviceZoom();
		}
	}
```

This would even be rather "generic", not requiring to check for Windows and monitor-specific scaling etc. but can just always be applied in `getImage()` as it will only have an effect when the device zoom changed during initialization (as currently only happens for monitor-specific scaling on Windows).

akoch-yatta: I like the idea of encapsulating the logic into a specific class. I will try that one out

akoch-yatta: Question for me is where to instantiate this wrapper. In my adaption I am doing it now at the same place, where `setRescaleAtRuntimePropertyFromPreference `was called before

HeikoKlare: I also did it there when quickly testing it. And it also seems fine to me. You could argue that the method could me called multiple times, instantiating/overwriting the static field multiple times, but that would be strange anyway and is not part of any current initialization flow. To be safe, one could check whether it has already been created and if yes, not do it again (to ensure that the original stored values are preserved).

akoch-yatta: Sound good to me. I added a null check

HeikoKlare (commented):
When testing this, I found that in `EclipseSplashHandler` the progress bar, task information, version information and others, which are placed on top of the splash screen, also using auto-scaling and thus scale wrongly. So just adapting the image scaling is not sufficient.

We could of course adapt all those places faking other auto-scale calculation, but that seems like the wrong approach to me. I would be in favor of ensuring that everything related to splash updates is executed within the "original auto-scale setup". I.e., whenever splash screen operations are execute, the original auto-scale value should be used automatically (rather than explicitly overwriting something).

I have updated the PR with an according proposal. It requires an adaptation in SWT: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1964

@akoch-yatta can you please have a look?

## Comparison

### Original
splash_original

### Only Image Adaptation
splash_onlyimage

### Current Proposal
splash_updated

akoch-yatta:
I tested your adaptions with the latest I-Build RCP and they behave as expected 👍

HeikoKlare (approved):
Alright, let's give this a try.

PR #2862, opened 2025-03-27
Add SWT SVG fragment to E4 RCP feature

### LINKED ISSUE #1944: SWT tests not executed in I20250325-1800
Most likely regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638

See https://download.eclipse.org/eclipse/downloads/drops4/I20250325-1800/testResults.php

The tests page says:

Image

The p2 director logs say

* https://download.eclipse.org/eclipse/downloads/drops4/I20250325-1800/testresults/ep436I-unit-linux-x86_64-java23_linux.gtk.x86_64_23/directorLogs/director-org.eclipse.swt.tests.log
* https://download.eclipse.org/eclipse/downloads/drops4/I20250325-1800/testresults/ep436I-unit-linux-x86_64-java21_linux.gtk.x86_64_21/directorLogs/director-org.eclipse.swt.tests.log

```
Installing org.eclipse.swt.tests 3.107.800.v20250325-0759.
Installing org.eclipse.test 3.6.200.v20241031-0735.
Installing org.eclipse.ant.optional.junit 3.3.400.v20230714-1309.
Installing org.eclipse.test.performance 3.21.100.v20250311-0829.
Installing org.eclipse.test.performance.win32 3.1.600.v20230714-1309.
Cannot complete the install because one or more required items could not be found.Installation failed.

!SESSION 2025-03-25 22:48:11.695 -----------------------------------------------
	Software being installed: Eclipse SWT Tests 3.107.800.v20250325-0759 (org.eclipse.swt.tests 3.107.800.v20250325-0759)eclipse.buildId=4.36.0.I20250325-1800
java.version=21.0.6
java.vendor=Eclipse Adoptium

	Missing requirement: Eclipse SWT Tests 3.107.800.v20250325-0759 (org.eclipse.swt.tests 3.107.800.v20250325-0759) requires 'org.eclipse.equinox.p2.iu; org.eclipse.swt.svg 0.0.0' but it could not be found
There were errors. See log file: /home/jenkins/agent/workspace/AutomatedTests/ep436I-unit-linux-x86_64-java21/workarea/I20250325-1800/eclipse-testing/test-eclipse/eclipse/configuration/1742942891803.log
Framework arguments:  -application org.eclipse.equinox.p2.director -flavor tooling -installIUs org.eclipse.swt.tests,org.eclipse.test,org.eclipse.ant.optional.junit,org.eclipse.test.performance,org.eclipse.test.performance.win32 -repository file:/home/jenkins/agent/workspace/AutomatedTests/ep436I-unit-linux-x86_64-java21/workarea/I20250325-1800/eclipse-testing/testRepo
Command-line arguments:  -application org.eclipse.equinox.p2.director -consoleLog -debug -flavor tooling -installIUs org.eclipse.swt.tests,org.eclipse.test,org.eclipse.ant.optional.junit,org.eclipse.test.performance,org.eclipse.test.performance.win32 -repository file:/home/jenkins/agent/workspace/AutomatedTests/ep436I-unit-linux-x86_64-java21/workarea/I20250325-1800/eclipse-testing/testRepo

!ENTRY org.eclipse.equinox.p2.director 4 10053 2025-03-25 22:48:16.133
!MESSAGE Cannot complete the install because one or more required items could not be found.
!SUBENTRY 1 org.eclipse.equinox.p2.director 4 0 2025-03-25 22:48:16.133
!MESSAGE Software being installed: Eclipse SWT Tests 3.107.800.v20250325-0759 (org.eclipse.swt.tests 3.107.800.v20250325-0759)
!SUBENTRY 1 org.eclipse.equinox.p2.director 4 0 2025-03-25 22:48:16.133
!MESSAGE Missing requirement: Eclipse SWT Tests 3.107.800.v20250325-0759 (org.eclipse.swt.tests 3.107.800.v20250325-0759) requires 'org.eclipse.equinox.p2.iu; org.eclipse.swt.svg 0.0.0' but it could not be found

```

HeikoKlare:
The SVG fragment needs to be available in runtime and SDK products for the platform. This adds it to the feature that already includes the SWT bundle and native fragments.

This is an extract of https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593 only containing the feature changes (potential other changes to be discussed in that other PR).

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944

### DIFF
--- a/features/org.eclipse.e4.rcp/feature.xml
+++ b/features/org.eclipse.e4.rcp/feature.xml
@@ -295,6 +295,10 @@
          version="0.0.0"/>
 
    <plugin
+         id="org.eclipse.swt.svg"
+         version="0.0.0"/>
+
+   <plugin
          id="org.eclipse.jface"
          version="0.0.0"/>
 


merks:
I lost touch with the long thread about this fragment...  If we only need to ensure the bundle is in the update site, we could just add it here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/6fb609f562ddb855df4342f50902abafa6defe8f/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml#L40-L42

HeikoKlare:
From my understanding, the fragment is actually available in the update site but not included in the products we deploy (runtime binaries, SDK and SDK with test bundles).

Basis for this PR are the comments starting from this (for those who understandably do not want to read the whole thread 🙂 ): https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944#issuecomment-2756981490

laeubi:
> SDK and SDK with test bundles

Should it then not added to the test feature if it is only required by tests (at the moment).

HeikoKlare:
> Should it then not added to the test feature if it is only required by tests (at the moment).

And then add it to the productive feature next week? Or leave it out there because it's implicitly taken through required capabilities there, so releng for productive and test deployment becomes different and people will need to know about that when eventually something does not work anymore?

I have no strong opinion but I would really like to close the discussion and come to a conclusion here. For Eclipse products, the SVG fragment will be required once we use the first SVGs anyway, so I would not overcomplicate this.

I would be in favor of just adding the SVG fragment to the RCP feature. 
@akurtakov @HannesWell @laeubi objections, confirmations?

akurtakov:
Go for the RCP feature for now.

laeubi (changes requested):
> And then add it to the productive feature next week?

No, the whole problem currently is that platform do not require this at all... it is only required by a test-fragment at the moment. We do not need (or should) add it to **any** feature, like we do not add any 3rdparty dependencies to features for quite some time now.

Ad feature versions are hard requirements, they often cause numerous issues and require versionbumps if they are not part of the same build, and for me SWT is a 3rd party dependency ...

HeikoKlare:
When blocking further progress, please propose exactly what solution you expect (in the best case via a PR) and please explain why it will be better to have test feature explicitly including a fragment and productive products implicitly including it via capabilities, makes everything more confusing. The "correct" solution would be to improve the overall build process (so that the test product includes the fragment also via capability), but that's obviously out of scope.

laeubi:
>  The "correct" solution would be to improve the overall build process (so that the test product includes the fragment also via capability), but that's obviously out of scope.

If you are hunting for the "correct" way than we should not include a feature that is nowhere used. So basically we now have a feature in SWT that is nowhere used in platform, so the platform test fail .. that's really bad but somehow expected.

So whatever collects all test-bundles into the product should include whats only needed for the tests, and we should not include it somewhere "just in case". e4.rcp feature is *that* feature every RCP application depends on (not only platform / IDE).

A while back I tried to replace the "[Eclipse Testing Framework](https://wiki.eclipse.org/Platform-releng/Eclipse_Test_Framework)" but it was said that is works great and we should not touch it, so it seems fair that from time to time we need to invest extra time in keeping it updated. 

Given that limiations, I would add the fragment to the `org.eclipse.sdk.tests` feature that seem to contain already some stuff (e.g. mockito) that is missing in a regular environment.

HeikoKlare:
> If you are hunting for the "correct" way than we should not include a feature that is nowhere used. So basically we now have a feature in SWT that is nowhere used in platform, so the platform test fail .. that's really bad but somehow expected.
> 
> So whatever collects all test-bundles into the product should include whats only needed for the tests, and we should not include it somewhere "just in case". e4.rcp feature is _that_ feature every RCP application depends on (not only platform / IDE).

My comment was supposed to express that we cannot hunt for the "correct" way anyway, so we need to find the best "pragmatic" one. I don't think it is expected that SWT tests with a defined required capability fail because that capability is not provided as no productive consumer is there. I would actually expect the mechanism creating the test product to also consider the dependencies of the test bundles added to the SDK.

Everyone depending on `e4.rcp` will depend on SVG support anyway, so there is no harm in adding it there. We could also just add the required capability to any of the UI bundles that we will migrate to SVGs in the next weeks anyway, having the same effect.

Since I don't want to invest more into this rather pointless discussion, here we go: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2948

laeubi:
> I don't think it is expected that SWT tests with a defined required capability fail because that capability is not provided as no productive consumer is there. 

The [SWT test are running fine](https://github.com/eclipse-platform/eclipse.platform.swt/runs/39510284783) (apart from some known flapping one) as far as I can see.

> I would actually expect the mechanism creating the test product to also consider the dependencies of the test bundles added to the SDK.

Obviously it does, otherwise it would not fail :-)
The problem is that the source from that this product is installed does not include the bundle (for whatever reason we use a different one than we use for the ibuilds itself)... this is something I have no clue what was/is the rationale or if it is just "because it was always the case"....

Because of this I always liked to see that [SWT becomes a standalone project](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1621) with its own build, its on deployment / tests and its own releases... this would make thing much more decoupled, but it seems [I'm the only one seeing benefits](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1621#discussioncomment-11523748) in less coupling.

> Everyone depending on e4.rcp will depend on SVG support anyway, so there is no harm in adding it there. 

There is a very subtile difference between a (transitive) dependency and a bundle mentioned in a feature, a transitive dependency can be updated independently from a feature (if the version range matches), a feature does lock down an **exact version** what together with the excessive use of singleton bundles makes it impossible to upgrade the SVG bundle if we mention it in any features.

HeikoKlare:
> > I would actually expect the mechanism creating the test product to also consider the dependencies of the test bundles added to the SDK.
> 
> Obviously it does, otherwise it would not fail :-)

No, it fails because when starting the product it cannot resolve the requirement. The mechanism creating the product does not consider it, which is why it is kind of "incomplete", e.g.: https://download.eclipse.org/eclipse/downloads/drops4/I20250326-1800/download.php?dropFile=eclipse-Automated-Tests-I20250326-1800.zip
That one contains `org.eclipse.swt.tests` without `org.eclipse.swt.svg`, even though there is a dependency.

> Because of this I always liked to see that https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1621 with its own build, its on deployment / tests and its own releases... this would make thing much more decoupled, but it seems https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1621#discussioncomment-11523748 in less coupling.

I don't think it makes sense broaden this up further and have general discussions here. Feel free to proceed in that referenced discussion. Would just be great to not spread generic, misleading statements, like me not seeing benefits in less coupling. It's just that only seldom decisions are one-dimensional, which is why I weigh arguments against each other to find conclusions.

laeubi:
> No, it fails because when starting the product it cannot resolve the requirement. 

Are we talking about the same issue or has the error changed?

- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944

says:

> Cannot complete the install because one or more required items could not be found.Installation failed.
> Missing requirement: Eclipse SWT Tests 3.107.800.v20250325-0759 (org.eclipse.swt.tests 3.107.800.v20250325-0759) requires 'org.eclipse.equinox.p2.iu; org.eclipse.swt.svg 0.0.0' but it could not be found

That is an error from the P2 director that **installing** the test bundle is not possible, if it would be an error when **starting** the product the message would be different and the fragment would simply not be attached (or the bundle not resolve).

HeikoKlare:
If still relevant, please read again the thread in https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944.

The SDK test product used for running the tests does not contain the mentioned fragment. So the mechanism generating that product does not consider the dependency of `org.eclipse.swt.tests` to `org.eclipse.swt.svg` (or anything else providing the require capability).

HannesWell (approved):
Now that we have the first bundles containing and using SVGs I think we should finally add the swt.svg fragment to the feature to be consistent.

laeubi (dismissed):
I can't see any "consistency" reasons here to be honest, as mentioned earlier, please do not add anything that is included automatically to a feature (if it is not part of the project).

This will just lock down the version of the dependency and make it harder to upgrade.

merks:
I think that not adding it will actually make it harder to upgrade it because p2 tends to be conservative and does not do "unnecessary" deep updates:

https://github.com/eclipse-equinox/p2/issues/173

So it might well be that there are updates to the fragment available that would not be upgraded unless something specifies a lower bound that would cause an update.  Or?

---

I think if we generalized "please don't add anything included automatically" to also apply for bundles and not just for fragments we'd be removing a lot of bundles that are transitively required wouldn't we?  And would that be a good thing?

laeubi:
> I think that not adding it will actually make it harder to upgrade it because p2 tends to be conservative and does not do "unnecessary" deep updates

"deep updates" is actually something completely different, and only mean that p2 will not offer an (UI) update for this, this does not mean it won't update anything.

In any case if someone is concerned, it is not the responsibility of **platform.ui** feature to ensure this but swt shoudl offer a feature then.

> So it might well be that there are updates to the fragment available that would not be upgraded unless something specifies a lower bound that would cause an update

From my experience P2 does update transitive things if I update the "parent" IU. And even if not, if nothing requires a higher version there is no immediate need to update. Still I can explicitly install an update on my own.

> I think if we generalized "please don't add anything included automatically" to also apply for bundles and not just for fragments we'd be removing a lot of bundles that are transitively required wouldn't we?

Yes this would actually be the case, e.g we include a lot of third party stuff here that should likely be _removed_ instead of making the feature more eager to _require_ more, if we absolutely want this it would be better to _import_ things like we do for emf... so interestingly no one is concerned about emf not being updated even though not explicitly required here :-)

> And would that be a good thing?

1. less version bumps required
2. less chance for including something not actually required
3. less risk for including duplicate (problematic) versions of the same thing
4. ...

so yes I think it would be a good thing and we already did it in the past without any negative impact reported so far.

merks:
You appear to argue now that the feature should not include org.eclipse.swt nor the fragments:

image

---

Let's do a little though experiment.  Suppose the SVG fragment is updated and only that fragment's version is updated.  So there is a new update site and the only difference is a new version of the SVG fragment.  Running Check for Updates would not update it because that process looks for a feature with a higher version number so that it can make a request to replace the older feature with the newer feature.  Ergo, not including the fragment in the feature makes it hard to update the fragment.  This does not seem like a good thing.

laeubi:
> Let's do a little though experiment. Suppose the SVG fragment is updated and only that fragment's version is updated.
> So there is a new update site and the only difference is a new version of the SVG fragment.

That's the point, currently we have no SWT features (except the `org.eclipse.swt.tools.feature`), so actually there should be a `org.eclipse.swt.feature` that also changes because SWT was changed. At best there would even be a `org.eclipse.swt.svg.feature` so only this one changes.

> Ergo, not including the fragment in the feature makes it hard to update the fragment. This does not seem like a good thing.

Taking your example the other way round, suppose only SVG fragment is updated, that means we need to bump `org.eclipse.e4.rcp`, and maybe others just to update SWT in Eclipse and need a new release of everything because features are hard wired in products and so on.

merks (approved):
Given every other SWT fragment and bundle is included here, adding svg seems a reasonable approach.

Alternative ways of factoring the feature structure can be a different exercise in the future, if there is some significant benefit to be gained from it.

PR #2868, opened 2025-03-28
[Win32] Enable monitor-specific scaling by default

### LINKED ISSUE #92: Provide Productive Preference for Activating Multi-Monitor HiDPI Support
As a user of the Eclipse IDE (or any other Eclipse-RCP-based product) on Windows, I want to be able to (de-)activate multi-monitor HiDPI support within the tool instead of defining a startup flag.
To this end, we should add an Eclipse preference for enabling the multi-monitor HiDPI support on Windows. This allows to easily test and adopt the new functionality. It also gives us the chance to quite easily enable this preference by default at some point in time, when the new support is mature enough to be used by default (or to enable it for I-/M-builds so that a larger user base tests the new functionality).

Preference description could be something like `Enable HiDPI support for multiple monitors (Windows only)" or the like.

This should also remove the existing experimental preference.

HeikoKlare:
The monitor-specific UI scaling on Windows is disabled by default and can be activated via a preference.

This change inverts the default enablement: it enables the feature by default and allows to disable it via preference. The existing experimental preference is renamed to reset the configuration for every consumer. All information about the feature being experimental is removed.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/IWorkbenchPreferenceConstants.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/IWorkbenchPreferenceConstants.java
@@ -722,11 +722,10 @@ public interface IWorkbenchPreferenceConstants {
 
 	/**
 	 * <p>
-	 * <strong>EXPERIMENTAL</strong>. Whether the UI adapts to DPI changes at
-	 * runtime. It only effects Windows.
+	 * Whether the UI adapts to DPI changes at runtime. It only affects Windows.
 	 * </p>
 	 *
 	 * @since 3.134
 	 */
-	String RESCALING_AT_RUNTIME = "RESCALING_AT_RUNTIME"; //$NON-NLS-1$
+	String RESCALING_AT_RUNTIME = "monitorSpecificScaling"; //$NON-NLS-1$
 }
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchMessages.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchMessages.java
@@ -36,9 +36,9 @@ public class WorkbenchMessages extends NLS {
 
 	public static String HiDpiSettingsGroupTitle;
 
-	public static String RescaleAtRuntimeEnabled;
+	public static String RescaleAtRuntimeDescription;
 
-	public static String RescaleAtRuntimeDisclaimer;
+	public static String RescaleAtRuntimeEnabled;
 
 	public static String RescaleAtRuntimeSettingChangeWarningTitle;
 
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
@@ -55,7 +55,6 @@ import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.fieldassist.ControlDecoration;
 import org.eclipse.jface.fieldassist.FieldDecorationRegistry;
-import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.preference.PreferencePage;
 import org.eclipse.jface.util.Util;
@@ -76,7 +75,6 @@ import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
-import org.eclipse.swt.widgets.Group;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPreferenceConstants;
@@ -140,7 +138,7 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 			layout.horizontalSpacing = 10;
 			comp.setLayout(layout);
 			createThemeIndependentComposits(comp);
-			createHiDPISettingsGroup(comp);
+			createRescaleAtRuntimeCheckButton(comp);
 			return comp;
 		}
 
@@ -186,7 +184,7 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 		createHideIconsForViewTabs(comp);
 		createDependency(showFullTextForViewTabs, hideIconsForViewTabs);
 
-		createHiDPISettingsGroup(comp);
+		createRescaleAtRuntimeCheckButton(comp);
 
 		if (currentTheme != null) {
 			String colorsAndFontsThemeId = getColorAndFontThemeIdByThemeId(currentTheme.getId());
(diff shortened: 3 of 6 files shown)


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchMessages.java
```diff
@@ -1,1042 +1,1040 @@
...
-	public static String Util_List;
-
-	// Zoom change messages
-	public static String Workbench_zoomChangedTitle;
-	public static String Workbench_zoomChangedMessage;
-	public static String Workbench_RestartButton;
-	public static String Workbench_DontRestartButton;
-
-}
+/*******************************************************************************
```
HannesWell: Could it be that this PR changes some line-endings unintentionally?

HeikoKlare: Yes, thanks. I had a temporary misconfiguration in my Git CLI. I will update the PR.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/IWorkbenchPreferenceConstants.java
```diff
@@ -722,11 +722,10 @@ public interface IWorkbenchPreferenceConstants {
 
 	/**
 	 * <p>
-	 * <strong>EXPERIMENTAL</strong>. Whether the UI adapts to DPI changes at
-	 * runtime. It only effects Windows.
+	 * Whether the UI adapts to DPI changes at runtime. It only effects Windows.
```
fedejeanne: NIT:
```suggestion
	 * Whether the UI adapts to DPI changes at runtime. It only affects Windows.
```

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
```diff
@@ -503,8 +503,7 @@ ThemeChangeWarningTitle = Theme Changed
 RescaleAtRuntimeSettingChangeWarningTitle = DPI Setting Changed
 RescaleAtRuntimeSettingChangeWarningText = Restart for the DPI setting changes to take effect
 HiDpiSettingsGroupTitle = HiDPI settings
-RescaleAtRuntimeEnabled = Monitor-specific UI &scaling
-RescaleAtRuntimeDisclaimer = EXPERIMENTAL! Activating this option will dynamically scale all windows according to the monitor they are currently in. It is still in development and therefore considered experimental.
```
fedejeanne, line 507: I prefer adapting the text instead of removing it, so there is an explanation about _what the option does_ (since it is not self-explanatory for those who haven't been involved in the development/testing of the HiDPI functionality). 

image

fedejeanne (commented):
I tested this and I'd like to propose some minor changes.

Also, shouldn't the option be then enabled when starting on a new workspace? I started a brand new _Eclipse Application_ with a clean WS and it wasn't

image

HeikoKlare:
> Also, shouldn't the option be then enabled when starting on a new workspace? I started a brand new _Eclipse Application_ with a clean WS and it wasn't

@fedejeanne The preference is not stored in the workspace but in the configuration. Did you clean that or create a new one as well? At least I cannot reproduce this issue.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
HeikoKlare: I thought about that as well but dropped the idea for consistency reasons. I don't see other places where such additional information for a preference is given (actually the group around it is also uncommon). But I don't have a strong opinion on this. I would also be fine with readding the explanation.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/IWorkbenchPreferenceConstants.java
HeikoKlare: Adapted with latest commit.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
fedejeanne: Good point, the group is also "non-standard". How about no group and only a tooltip then?

image

Looks more like the rest of the preferences, right?

HeikoKlare: Looks great 👍 If you have already made the adaptation, can you just push it onto this PR? Thanks!

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -214,13 +213,9 @@ private void createHiDPISettingsGroup(Composite parent) {
...
 		GridLayout layout = new GridLayout(1, false);
 		group.setLayout(layout);
-		Label infoLabel = new Label(group, SWT.WRAP);
-		infoLabel.setText(WorkbenchMessages.RescaleAtRuntimeDisclaimer);
-		infoLabel.setLayoutData(GridDataFactory.defaultsFor(infoLabel).create());
-		createLabel(group, ""); //$NON-NLS-1$
 
 		boolean initialStateRescaleAtRuntime = ConfigurationScope.INSTANCE.getNode(WorkbenchPlugin.PI_WORKBENCH)
-				.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, false);
+				.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, true);
```
fedejeanne, line 208: This should also be done at line 375 (the default value is still `false` there).

fedejeanne (commented):
> > Also, shouldn't the option be then enabled when starting on a new workspace? I started a brand new _Eclipse Application_ with a clean WS and it wasn't
> 
> @fedejeanne The preference is not stored in the workspace but in the configuration. Did you clean that or create a new one as well? At least I cannot reproduce this issue.

Thank you for the pointer. I must have screwed up somewhere else when checking out your PR because I tried it again and I cannot reproduce it either. One pointer (for my future self too) is that when creating a new _Run Configuration_, this option is enabled by default:

image

This makes testing this kind of PRs impossible: I was getting always the default value (`true`) for the property and it was therefore always enabled after a restart. I disabled that option and was able to test this PR just fine.

I have one final remark: the initial value of `true` should also be applied in line 375 of the `ViewsPreferencePage`, otherwise changing the preference from `true` to `false` for the very 1st time doesn't not ask for a workbench restart (subsequent changes will ask for a restart though)

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
fedejeanne: Done in 056cc2f520820da2d509ac13dbc21f2616e1c1b2

merks:
There is a checkbox in each launch configuration whether to clear the configuration before launching.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
fedejeanne: I took the liberty of doing it myself in 1ee97cc5ca6a1565a238d041799064b10a639fec. If it's incorrect, please let me know (or revert)

fedejeanne:
> There is a checkbox in each launch configuration whether to clear the configuration before launching.

Do you mean this one? https://github.com/eclipse-platform/eclipse.platform.ui/pull/2868#pullrequestreview-2729084588

fedejeanne (approved):
From my point of view, this PR is now good to go ✔️ 

@HeikoKlare fyi I made 2 changes (2 commits), they are already squashed.

merks:
> > There is a checkbox in each launch configuration whether to clear the configuration before launching.
> 
> Do you mean this one? [#2868 (review)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2868#pullrequestreview-2729084588)

Yes. Reading on phone so harder to see all details. Sorry.

fedejeanne:
> > > There is a checkbox in each launch configuration whether to clear the configuration before launching.
> > 
> > 
> > Do you mean this one? [#2868 (review)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2868#pullrequestreview-2729084588)
> 
> Yes. Reading on phone so harder to see all details. Sorry.

No worries, I was just making sure I got it right :-) . Thank you, Ed!

fedejeanne:
Test failures are unrelated: 
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/294
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1952 fixes the test failure in Windows. **Need to wait until tonight's I-Build so the test stops failing in the PRs**

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
HeikoKlare: Indeed, good catch!

HeikoKlare:
Thank you! Changes look good to me.

fwiw: when introducing the preference, we had to move it to the configuration area because it has to be read before the workspace is available.

iloveeclipse:
We have a regression: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003

PR #2871, opened 2025-03-28
Add required capability for SVG support to org.eclipse.ui

HeikoKlare:
In order to use SVGs as icons, an according rasterizer has to be present. This is achieved by adding an according required OSGi capability.

In order to prevent issues that we had for SWT tests (https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944) also in production code, I propose to add the required capability to one of the UI bundles that will contain SVG before actually adding them. This is supposed to avoid that it breaks the build or the generated products because the according rasterizer fragment might be missing in the deployed products (runtime, SDK).


HeikoKlare:
Closing this as obsolete. Has been tested via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2957

According required capabilities need to be added via SVG-adding PRs to the affected bundles, such as in:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2643

PR #2876, opened 2025-03-31
Don't create listener when changing text in find-replace dialog #2865

### LINKED ISSUE #2865: FindReplaceDialog: indefinite listener generation and registration

## Steps to reproduce

Whenever changing the input of the find or replace input fields of a FindReplaceDialog, a new listener is registered. The more is typed in those fields, the more listeners will be registered:

Image

This is caused by registered a listener inside the text modification listener:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/b6af990cac142bdba33768e29ec0d62482f0ba54/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java#L146-L159

Code has been introduced and thus this is a regression caused by:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2373

@fedejeanne please have a look

fedejeanne:
Fixes #2865

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -148,12 +148,7 @@ class FindReplaceDialog extends Dialog {
 			// XXX: Workaround for Combo bug on Linux (see bug 404202 and bug 410603)
 			if (fIgnoreNextEvent) {
 				fIgnoreNextEvent = false;
-				return;
 			}
-			modificationHandler.run();
-			fFindField.addModifyListener(event -> {
-				decorate();
-			});
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
```diff
@@ -148,12 +148,7 @@ public void modifyText(ModifyEvent e) {
 			// XXX: Workaround for Combo bug on Linux (see bug 404202 and bug 410603)
 			if (fIgnoreNextEvent) {
 				fIgnoreNextEvent = false;
-				return;
 			}
-			modificationHandler.run();
```
HeikoKlare, line 153: @fedejeanne this change seems unrelated and incorrect: after this change, the `fIgnoreNextEvent` is obsolete, but it is required to fix a bug on Linux. Maybe the execution of the handler at the beginning of the method was wrong instead, but the code as it is now is broken.

fedejeanne: I reverted it in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2879 but the problem (https://bugs.eclipse.org/bugs/show_bug.cgi?id=404202) was there even before my PR so it requires further investigation: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2880

Issue #2875, opened 2025-03-31
Upstream fixed ScopedPreferenceStore from XText


pazi146:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/) doesn't provide the feature.

## Suggestion

Upstream the changes in `FixedScopedPreferenceStore` (Xtext) to `ScopedPreferenceStore` to fix the memory leak

old bugs:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=239033
https://bugs.eclipse.org/bugs/show_bug.cgi?id=362199

Relevant change: The handling of the `INodeChangeListener` is changed to be the same as the `IPreferenceChangeListener` (which fixes the memory leak)

What would be the steps to upstream this change with the proper attribution?

## Community

- [x] I understand suggesting an enhancement doesn't mandate anyone to implement it. Other contributors may consider this suggestion, or not, at their own convenience. The most efficient way to get it fixed is that I implement it myself and contribute it back as a good quality patch to the project.

iloveeclipse:
For the reference: https://github.com/eclipse-xtext/xtext/commit/e6ebfc52bc16f24d978a90f3c3f64c0a7b135c92

@pazi146 : do you plan to work on it?

pazi146:
@iloveeclipse the work is already done - I can make a pull request but I'm not sure how to properly attribute the work (or if I'm even allowed to do that with someone else's work)

iloveeclipse:
> I'm not sure how to properly attribute the work (or if I'm even allowed to do that with someone else's work)

If I read it right, the work was done by @szarnekow  via https://github.com/eclipse-xtext/xtext/commit/e6ebfc52bc16f24d978a90f3c3f64c0a7b135c92.

1) The code is under EPL, so sure you can contribute it to platform.
2) You can add @szarnekow mail via `Also-By: ` line in the git commit message (see https://www.eclipse.org/projects/handbook/#resources-commit)

pazi146:
Thank you for the info

PR #2878, opened 2025-04-01
Upstream memory leak fix for ScopedPreferenceStore

pazi146:
introduce fix created by Sebastian Zarnekow:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=239033 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=362199

Also-By: Sebastian Zarnekow <Sebastian.Zarnekow@gmail.com>

Context: #2875

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/preferences/ScopedPreferenceStore.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/preferences/ScopedPreferenceStore.java
@@ -12,6 +12,9 @@
  *     IBM Corporation - initial API and implementation
  *     Yves YANG <yves.yang@soyatec.com> -
  *     		Initial Fix for Bug 138078 [Preferences] Preferences Store for i18n support
+ *     Patrick Aigner -
+ *     		upstream fix created by Sebastian Zarnekow <Sebastian.Zarnekow@gmail.com>
+ *     		Fix for Bug 239033 [Preferences] ScopedPreferenceStore causes memory leak
  *******************************************************************************/
 package org.eclipse.ui.preferences;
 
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/preferences/ScopedPreferenceStore.java
```diff
@@ -224,8 +247,9 @@ private IEclipsePreferences getDefaultPreferences() {
 
 	@Override
 	public void addPropertyChangeListener(IPropertyChangeListener listener) {
-		initializePreferencesListener();// Create the preferences listener if it
-		// does not exist
+		// Create the preference listeners if they do not exist
```
laeubi, line 250: It looks like the life-cycle of the listener is solely bound to the `IPropertyChangeListener`, I therfore think it would be better to instead of handling it "globally" to instead have a wrapped listener object that is forwarding the events directly instead of holding an own instance in the store.

pazi146: What do you mean with a wrapped listener object?

It isn't dependent on a single `IPropertyChangeListener` but on the fact if any listeners are attached at all - no use listening if there is nothing to forward it to.

This copies over the relevant changes from: https://github.com/eclipse-xtext/xtext/blob/main/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/FixedScopedPreferenceStore.java

laeubi: > It isn't dependent on a single IPropertyChangeListener but on the fact if any listeners are attached at all 

Instead of one "global" listener used for all `IPropertyChangeListener` it should wrap this one into the required type and use this for the life-cycle management.

laeubi:
@pazi146 I have now created a PR can you check if it solves your observed problem?

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2886

PR #2881, opened 2025-04-01
Make GenericEditor default editor for org.eclipse.core.runtime.text

### LINKED ISSUE #2527: Make Generic Code Editor default for text files

iloveeclipse:
Make generic editor default for the base `org.eclipse.core.runtime.text` content type.

This shouldn't be a breaking change, given that with b3137d146f60760c214f2f96152c526251a5531c customers expecting "old classic" text editor to be opened by default could customize their products by adding this line in product customization file:

`org.eclipse.ui.workbench/defaultEditorForContentType_org.eclipse.core.runtime.text=org.eclipse.ui.DefaultTextEditor`

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2527

### DIFF
--- a/bundles/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.ui.genericeditor;singleton:=true
-Bundle-Version: 1.3.500.qualifier
+Bundle-Version: 1.3.600.qualifier
 Bundle-Vendor: %Bundle-Vendor
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Require-Bundle: org.eclipse.ui.workbench.texteditor;bundle-version="3.10.0",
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Note: the files with "unknown" content type will be still opened in the default text editor (just try with empty "blup" file). @deepika-u : maybe you could check that as well? It is not mandatory for the #2527 so it can be investigated independently.

Also note: this change changes not **only** `org.eclipse.core.runtime.text` base content type association but also changes all content types that are based on the `org.eclipse.core.runtime.text` and have no default editors configured.

So assuming a product relies on `content type_1`, `content type_2` and `content type_3` being opened in a default text editor, all three content types need a preference change to be set in product customization file. I think this is acceptable. 

If  product owners want to still use default text editor for **all** not configured child content types of `org.eclipse.core.runtime.text`, they could either remove generic text editor from target platform or update the product to work with generic text editor or contribute a better fix for the platform.

@HeikoKlare : any concerns? I would prefer to merge that in M1 to see how much ripples we will get (if any).

vogella (approved):

vogella:
Tested for .txt files and works well

deepika-u:
Hi,
Please find my observations as below.
- tested text files i see it works fine.
image

- tested .html files i see it works fine.
image

- tested .xml files, i see it works fine but the image against the file type is not same as in the case of .txt file image.
image

but when checked on .xml file => open with, it is showing "generic code editor" itself
image

image against file type txt with "open with" for reference 
image

- tested xlsx files, i see default editor is "generic code editor" showing fine, so icon is showing up fine.
image

- tested Content Types from preferences but here "generic code editor" is not made default.
image

If these are expected, please go ahead with the merge of this fix.

iloveeclipse:
I've merged the change now for M1 to see early regressions downstream.

PR #2882, opened 2025-04-01
[Find/Replace dialog][Linux] "Replace" buttons remain disabled when using "Enter"

### LINKED ISSUE #2880: [Find/Replace dialog][Linux] "Replace" buttons remain disabled when using "Enter"

- [x] I verified I can reproduce this issue **on Linux** against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

Same as https://bugs.eclipse.org/bugs/show_bug.cgi?id=410603

**On Linux**
- Open a file
- Make sure no text is selected
- Open the "old" search dialog (`Ctrl + f`) 
- Search for something that is in the file
- Hit `Enter` --> **The "replace" buttons remain disabled.** 

When using the mouse to _click_ the "**Find**" button, everything works as expected.

## Reproducible on
- ~Windows~
- [x] Linux
- [ ] Mac (??? not tested)

## Considerations
I was made aware of [the bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=410603) when working on another PR (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/2876#discussion_r2021326068) but it turns out that my PR did not cause it so reverting the mentioned changes (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/2879) doesn't fix it (maybe that PR is part of the fix but it's not the complete fix).

## Community

## Disclaimer
I can't work on this right now so I'm adding the tag "help wanted". Feel free to jump in.

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2880

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/FindReplaceDialog.java
@@ -129,13 +129,15 @@ class FindReplaceDialog extends Dialog {
 	 */
 	private class InputModifyListener implements ModifyListener {
 
-		private Runnable modificationHandler;
+		private final Runnable logicUpdateHandler;
+		private final Runnable uiUpdateHandler;
 
 		// XXX: Workaround for Combo bug on Linux (see bug 404202 and bug 410603)
 		private boolean fIgnoreNextEvent;
 
-		private InputModifyListener(Runnable modificationHandler) {
-			this.modificationHandler = modificationHandler;
+		private InputModifyListener(Runnable logicUpdateHandler, Runnable uiUpdateHandler) {
+			this.logicUpdateHandler = logicUpdateHandler;
+			this.uiUpdateHandler = uiUpdateHandler;
 		}
 
 		private void ignoreNextEvent() {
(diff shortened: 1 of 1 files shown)


HeikoKlare:
It might be that something has been mixed up in this handler by all the recent changes (including mine extracting the update of the button state from the handler). This change actually looks reasonable, but it will probably result in the same state of the code we had when this issue was reported:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2322

And that issue was fixed by moving the modification handler around (in the opposite direction as done by this change):
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2327

I can have a look at that once I am back from travel (on Thursday). If someone wants to reevaluate this change with the above mentioned information, feel free to do so.

iloveeclipse:
Thanks Heiko, if I will have time tomorrow, I will check which workaround for which regression caused which problem.

HeikoKlare:
@iloveeclipse I have checked the implementation and the existing workaround again and this is my understanding how it has to behave:
- the data passed to the underlying `FindReplaceLogic` (find and replace string) always have to updated when the modification listener is called
- the button state must only be updated on the second call of the listener on Linux (that's the existing workaround)
That's
Since the listener is parameterized with both these kinds of operations, which have to be called under different conditions, they cannot be combined in a single handler as they are right now. One solution is to extend the listener parameterization to take two callbacks, one for each of the above mentioned situations.

I have just pushed according changes to the PR branch for demonstration purposes. In my opinion that should solve all mentioned issues. Feel free to adapt/revert/merge them as desired and required. I will leave for vacation now, so I will probably not be able to contribute further to this for the next week, but I hope the above mentioned helps to finalize the fix.

iloveeclipse:
@HeikoKlare : thanks, it works, I've verified that.

PR #2885, opened 2025-04-02
misc: fix spelling

jbampton:
General spelling clean up all similar words


bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/CellEditor.java
```diff
@@ -692,7 +692,7 @@ public boolean isValueValid() {
 	 * @param keyEvent
 	 *            the key event
 	 */
-	protected void keyReleaseOccured(KeyEvent keyEvent) {
+	protected void keyReleaseOccurred(KeyEvent keyEvent) {
```
akurtakov, line 695: Unfortunately this is API so it can not be changed as every downstream project will break . Please revert this part.

akurtakov: If you don't plan to act on this one I'll close the PR as it can't be accepted in this form.

akurtakov:
Closing as there doesn't seem to be interest in fixing issue.

PR #2899, opened 2025-04-09
Merge FileImageDescriptor into URLImageDescriptor and clean it up

HannesWell:
In it's core the FileImageDescriptor queries the given context class for a resource at the given path and then operates on the returned URL or directly fetches the resource's stream. By optaining the resource's URL immediately and using an URLImageDescriptor with it a lot of similar code and logic from the FileImageDescriptor can be saved.

Additionally apply a few minor code clean-ups and remove a unused internal method.


laeubi:
Honestly this now feels it become completely unmaintainable and hardly understandable, can we not simply use an abstract base-class to share code instead of all this inlined lamda and hype-generic method calls?

HannesWell:
> Honestly this now feels it become completely unmaintainable and hardly understandable, can we not simply use an abstract base-class to share code instead of all this inlined lamda and hype-generic method calls?

I'm not so sure if that becomes much simpler, but I can have a look.
Actually this was intended as up-front work to reduce the duplicated changes for https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593. But maybe it's simpler to complete that first and make this a follow-up. Then we also know better was is needed for that.

HannesWell:
While looking at both classes again I got the impression that actually the `FileImageDescriptor` class could be fully replaced by `URLImageDescriptor`. There are probably some corner cases that still have to be handled (e.g. resource not found), but in general I didn't saw reason to keep the `FileImageDescriptor`.
Do you or anybody else know one?

HannesWell:
> While looking at both classes again I got the impression that actually the `FileImageDescriptor` class could be fully replaced by `URLImageDescriptor`.

I have now taken this path further and didn't found a reason speaking against it. The tests should now pass as well.

HannesWell:
This contains now much less, i.e. just one additional lambda in the zoom-image computation and is, from my side, ready for submission.
Does anybody want to review this?
If there are no objection respectively to request for more time to review, I plan to submit this tomorrow.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -84,7 +87,11 @@ protected ImageDescriptor() {
 	 * @return a new image descriptor
 	 */
 	public static ImageDescriptor createFromFile(Class<?> location, String filename) {
-		return new FileImageDescriptor(location, filename);
+		URL url = location != null ? location.getResource(filename) : null;
```
HeikoKlare: There was some more complex logic for retrieving the URL, in particular for the case that `location` was null. Is it intended to remove that?

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -138,35 +107,34 @@ public ImageData getImageData(int zoom) {
...
-				ImageData xdata = getImageData(xUrl, zoom, zoom);
-				if (xdata != null) {
-					return xdata;
-				}
+			return getZoomedImage(tempURL, url, zoom, u -> getImageData(u, zoom, zoom));
+		}
+		return null;
+	}
+
+	private static <R> R getZoomedImage(URL url, String urlString, int zoom, Function<URL, R> getImage) {
```
HeikoKlare: nit: this does actually not return an image (as indicated by the method name), but something depending on the type parameter, such as some image data or even just a path to the image. But I have to admit that I have no better generic name to propose.

HeikoKlare (commented):
I really like this simplification. I did not check in detail whether everything the FileImageDescriptor did is properly captured by the URLImageDescriptor as well. So far, I just found one place where I am not sure if behavior is properly preserved.

If you want me to do a more in-depth review, let me know, but I am not sure how soon I will find the time for it. So from my side, feel free to further process this PR if you are confident that everything works as expected.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
HannesWell: That's right. Initially I named that method `getZoomedImageSource()` but wanted to be a bit more compact here. However, that is probably a name that's more clear. Changed it to `getZoomedImageSource()`.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
HannesWell: You are right. I oversaw that part.
In incorporated the treatment as plain file-path when the location class is null. The remaining part of `FileImageDescriptor.getFilePath()` doesn't look relevant as we just want to determine the 'base' path here, i.e. without any `@2` or similar suffixes.

The only behavior difference I currently see is that the `FileImageDescriptor` looked up the resource in the Class(Loader), when the image was created and now the look-up is already performed when the `ImageDescriptor` is created.
Since Jface may run in an OSGi runtime, _theoretically_ new fragments for the bundle of the context class could be installed or uninstalled in the meantime, which means that Image resources could become available or unavailable later. The latter case would have the same result since the Image would not be loaded in either case, but the former would be different because the image would not be loaded although it became available in the meantime.

I would say that this case is not relevant, at least in the way how Eclipse (with P2) behaves as it usually requires a restart after installing new bundles.
But if somebody thinks this is a relevant we could still cover this by reworking the URLImageConstructor to get/construct the URL lazy using a `Supplier` or `Function`. Or, but I would have to check some details, we could create the URL for the bundle/resource root and construct the URL manually. But this would require that the returned URL is identical for host-bundle and fragment and I'm in doubt this is the case.

HannesWell: > But if somebody thinks this is a relevant we could still cover this by reworking the URLImageConstructor to get/construct the URL lazy using a `Supplier` or `Function`. Or, but I would have to check some details, we could create the URL for the bundle/resource root and construct the URL manually. But this would require that the returned URL is identical for host-bundle and fragment and I'm in doubt this is the case.

Alternatively, if JFace runs in an OSGi environment, we could obtain the bundle containing the given class and construct a `platform:/plugin/<bundle-symbolic-name>/<path>` URL from it. This would at least work as long as the bundle is (effectively) a singleton.

HeikoKlare: > I would say that this case is not relevant, at least in the way how Eclipse (with P2) behaves as it usually requires a restart after installing new bundles.

My conclusion here is also that the described scenario is rather irrelevant. So I would be in favor of not adding any additional complexity just to cover such a case that will not be relevant for anyone. Instead, let's keep/make the code as simple as possible and stick with the solution you currently propose.

laeubi: > I would say that this case is not relevant, at least in the way how Eclipse (with P2) behaves as it usually requires a restart after installing new bundles.

Keep in mind that the filesystem (!) could perfectly change while running and without a restart (e.g. in dev mode) but I don't think that this is of much relevance here and we not necessarily need "hot resource replacement".

HannesWell: Thank you both. So I think we have consensus here that, while it's possible that the new logic is insufficient in different scenarios, we all consider them as not relevant.

HannesWell:
Thank you for your review. Then I think this is ready for submission.
Nevertheless I have extracted the simplification part out into a separate PR
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2938

With that only a few a few changes remain in `URLImageDescriptor` for this change.

Once we have at least one working build I plan to submit this.

HeikoKlare:
@HannesWell Are you aware that this change introduced a test failure in `FileImageDescriptorTest`? That test fails in every PR since this one was merged and it's also reproducible locally that this change introduced the failure. It was also reported by the according checks for this for 😉  https://github.com/eclipse-platform/eclipse.platform.ui/runs/41211453350

HannesWell:
> Are you aware that this change introduced a test failure in `FileImageDescriptorTest`? That test fails in every PR since this one was merged and it's also reproducible locally that this change introduced the failure.

My bad, I oversaw that and confused it with the other intermittent test-failure. Sorry and thanks for the heads-up.
Just created a fix to handle the kind of illegal files passed in in that test:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2946

basilevs:
>     * [Merge FileImageDescriptor into URLImageDescriptor and clean it up #2899 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2899#issuecomment-2833461917)

The link is broken.

PS. RCPTT tests also broke due to changed `ImageDescriptor` representations. Users have to recapture Table/Tree verifications. Nothing can be done to avoid or fix this situation, because RCPTT effectively [relies ](https://github.com/eclipse-rcptt/org.eclipse.rcptt/blob/8e8e170ff43f8e13aa470281138fab007aff92e5/runtime/tesla/org.eclipse.rcptt.tesla.jface.aspects/src/org/eclipse/rcptt/tesla/jface/DescriptorInfo.java#L65)on `ImageDescriptor.toString()` which is an implementation detail.

HannesWell:
> The link is broken.

It was indeed the wrong link, just fixed it.

> PS. RCPTT tests also broke due to changed `ImageDescriptor` representations. Users have to recapture Table/Tree verifications. Nothing can be done to avoid or fix this situation, because RCPTT effectively [relies ](https://github.com/eclipse-rcptt/org.eclipse.rcptt/blob/8e8e170ff43f8e13aa470281138fab007aff92e5/runtime/tesla/org.eclipse.rcptt.tesla.jface.aspects/src/org/eclipse/rcptt/tesla/jface/DescriptorInfo.java#L65)on `ImageDescriptor.toString()` which is an implementation detail.

That's indeed unfortunate. But as you said, the exact result of `toString()` is not specified and one should not rely on it to have a specific value. If you think it's worth it and you can propose a simple fix to somehow restore the the result, we could 'fix' that.

basilevs:
> If you think it's worth it

It is not. An application can change icons in its UI at any release and that's effectively what happened. Testing tools comparing icons should not pretend, than nothing has changed, so test failures are expected.

I only mention the problem for context, so that bigger picture is more visible.

ptziegler:
> > If you think it's worth it
> 
> It is not. An application can change icons in its UI at any release and that's effectively what happened. Testing tools comparing icons should not pretend, than nothing has changed, so test failures are expected.
> 
> I only mention the problem for context, so that bigger picture is more visible.

Clients will likely have to update their tests anyway, due to the migration from PNGs to SVGs. So I think having both changes in a single release is actually good thing.

For example, this is an error now reported by our test suite:
> Line 79: verify-true: Assertion of cells[4].image.path failed: expected:<...full/obj16/warn_tsk.png> but was:<...full/obj16/warn_tsk.svg>.

laeubi:
By the way previously we have used the [Adapter Pattern](https://www.eclipse.org/articles/Article-Adapters/) to gather additional information.

 e.g. `Adapters.of(imageDescriptor, URL.class)`  would be much bettern than rely on a artificial toString ..

basilevs:
> By the way previously we have used the [Adapter Pattern](https://www.eclipse.org/articles/Article-Adapters/) to gather additional information.
> 
> e.g. `Adapters.of(imageDescriptor, URL.class)` would be much bettern than rely on a artificial toString ..

Unless the adapter is implemented by JFace, it is no better than direct downcast. And downcast requires internal API access. toString() was used to avoid use of internal APIs.

I see that adapter was added in 2022, but RCPTT still supports Eclipse Platform 2019-03, which makes the approach non-universal.

laeubi:
> I see that adapter was added in 2022, but RCPTT still supports Eclipse Platform 2019-03, which makes the approach non-universal.

This does not mean one could try it _first_ and then _fallback_ to the old approach...

PR #2901, opened 2025-04-10
handle a CancellationException without making it visible to the user

rubenporras:


iloveeclipse:
Could you please explain the original problem?

rubenporras:
The problem is that when `org.eclipse.lsp4e.operations.completion.LSContentAssistProcessor.LSContentAssistProcessor(boolean, boolean)` ends with a `CancellationException` (which happens anytime the user types before the actual completion has been computed), that exception is propagated and logged by the `SafeRunner`.

An alternative solution would be to adapt LSP4E to catch the `CancellationException` and throw an `OperationCanceledException` or to return an empty list. 

Do you think that is preferable?

iloveeclipse:
> Do you think that is preferable?

No idea without debugging the code, but I believe @mickaelistria did the work in this area, so he should be able to provide qualified answer.

mickaelistria:
Yes, I think updating LSP4E to report the usual OperationCanceledException (with the CancellationException being the cause) would be preferred.

rubenporras:
I will do that instead. Thanks for the feedback.

PR #2906, opened 2025-04-11
Deprecate "fastViewWidthRatio" extension point attribute & friends

ptziegler:
Fast views are unsupported and the attribute is not even read from the extension point. The corresponding method in the IViewDescriptor has been deprecated as well.


ptziegler:
image

vogella:
LGTM

PR #2907, opened 2025-04-11
Avoid implementations of ImageDescriptor.getImageData()

HannesWell:
`ImageDescriptor.getImageData()` is deprecated and sub-classes are advised to stop re-implementing it and instead implement `getImageData(int)`. The still existing implementations just replicate the default implementation and therefore can just be removed.

Additionally implementation of
`ImageDescriptor.getMissingImageDescriptor()` is simplified and `ImageDescriptor` implementations are cleaned-up.


bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -130,12 +130,6 @@ public boolean equals(Object o) {
 		return ((URLImageDescriptor) o).url.equals(this.url);
 	}
 
-	@Deprecated
-	@Override
-	public ImageData getImageData() {
-		return getImageData(getURL(url), 100, 100);
-	}
```
HannesWell, line 137: @Michael5601 FYI as this was a relevant point in our discussion at https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593. As before that change this implementation now just behaves as if `getImageData(100)` is called and therefore as the default implementation in the super-class. It can therefore just be removed.

bundles/org.eclipse.jface/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface;singleton:=true
-Bundle-Version: 3.36.100.qualifier
+Bundle-Version: 3.37.0.qualifier
```
HannesWell, line 5: Somehow API-tools considers the removal of the (deprecated) overriden method from `CompositeImageDescriptor` as API addition and demands to minor version bump.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/misc/ExternalProgramImageDescriptor.java
```diff
@@ -26,7 +25,7 @@
  */
 public class ExternalProgramImageDescriptor extends ImageDescriptor {
 
-	public Program program;
+	private Program program;
```
basilevs, line 28: Can RCPTT get a better `toString()` or an adapter to be able to introspect this?
A dirty hack is applied to work around the reduced access:
https://github.com/eclipse-rcptt/org.eclipse.rcptt/issues/174
https://github.com/eclipse-rcptt/org.eclipse.rcptt/commit/ba5e639067db35692bd4f10bf31a03ff63e2e681

laeubi: Yes implemention `IAdaptable` would be the right approach here to access internal data structure.

HannesWell: Done in
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2991

PR #2934, opened 2025-04-23
Redesign Message Dialog

### LINKED ISSUE #2929: Center text/save space on message dialogs?

## Suggestion

I reach state *Get any message dialog* with, for example an edit on a file and trying to close the editor without save.

I get 

Image

instead, I think we should consider removing the minimal size from the dialog and let it be as compact as possible, with the text centered on it. 

Looking at UI recommendations and habits from various OS, such a dialog would seem more consistent with the current state of art.

## Community

SougandhS:
Removes unwanted space and aligns message content & buttons to center

Before 
image <br>
After
Screenshot 2025-04-22 at 1 02 49 PM

Fixes : https://github.com/eclipse-platform/eclipse.platform.ui/issues/2929

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/IconAndMessageDialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/IconAndMessageDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -100,13 +100,15 @@ public abstract class IconAndMessageDialog extends Dialog {
 		if (message != null) {
 			messageLabel = new Label(composite, getMessageLabelStyle());
 			messageLabel.setText(message);
+			int xHint = message.length() > 40
+					? convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH)
+					: convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH - 50);
 			GridDataFactory
 					.fillDefaults()
 					.align(SWT.FILL, SWT.BEGINNING)
 					.grab(true, false)
-					.hint(
-							convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH),
-							SWT.DEFAULT).applyTo(messageLabel);
+					.hint(xHint, SWT.DEFAULT)
+					.applyTo(messageLabel);
 		}
 		return composite;
 	}


mickaelistria:
As the discussion on the bug reveals, there is no obvious consensus to find between aligned left vs centered; however, all OSs recommend to get the width fit the content (ie no extra space), so I think for a first iteration, we should focus on just saving the extra space (changing the min width to 0 or DEFAULT), and consider whether to center or not later; maybe if all OSs align on the same recommendation some day.

SougandhS:
Updated one

image

mickaelistria:
I'm seeing an issue with the given patch: the MessageDialog now tends to become too large:

Before
Screenshot From 2025-04-25 13-18-19

After
Screenshot From 2025-04-25 13-16-31

SougandhS:
> I'm seeing an issue with the given patch: the MessageDialog now tends to become too large:

Made minor adjustment for these cases

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/MessageDialog.java
```diff
@@ -328,7 +328,6 @@ protected Control createDialogArea(Composite parent) {
 		layout.marginWidth = 0;
 		composite.setLayout(layout);
 		GridData data = new GridData(GridData.FILL_BOTH);
-		data.horizontalSpan = 2;
```
merks: I thought there was previously already feedback about this change affecting all subclasses.  There are a great many so this seems not a good idea at all.

image

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/IconAndMessageDialog.java
```diff
@@ -100,13 +100,14 @@ protected Control createMessageArea(Composite composite) {
 		if (message != null) {
 			messageLabel = new Label(composite, getMessageLabelStyle());
 			messageLabel.setText(message);
+			int minWidthHint = convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH - 50);
```
merks: It's odd to compute this in advance and then only use it conditionally later.

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/IconAndMessageDialog.java
```diff
@@ -100,13 +100,14 @@ protected Control createMessageArea(Composite composite) {
...
 			GridDataFactory
 					.fillDefaults()
 					.align(SWT.FILL, SWT.BEGINNING)
 					.grab(true, false)
-					.hint(
-							convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH),
-							SWT.DEFAULT).applyTo(messageLabel);
+					.indent(5, 0)
+					.hint(message.length() > 40 ? minWidthHint : SWT.DEFAULT, SWT.DEFAULT)
```
merks: The number 40 is quite magic.  I wonder how things like font size and scaling might affect what this number should be?

merks (changes requested):
I think changing the horizontalSpan is a bad idea.  I'm not even sure why it's needed.

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/MessageDialog.java
SougandhS: Sorry, I had only configured the platform UI code in my workspace, which caused me to overlook the impact on the subclasses. I've now reverted the change and adjusted the value to 1 to remove the extra space.

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/IconAndMessageDialog.java
SougandhS: Fixed now

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/IconAndMessageDialog.java
SougandhS: I tested the implementation with few different font sizes and found the results to be satisfactory, which is why I proceeded with this approach.

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/MessageDialog.java
mickaelistria: See 
Screenshot From 2025-04-28 11-03-54
about a regression caused by changing the span.

BeckerWdf:
I asked our inhouse UX-Experts. This is what they say:

We would **not recommend** the centered alignment of dialog content in Eclipse – even for short texts like "Save ‘NAME.filetype‘?" – for several reasons:

- This dialog is displayed for only a few seconds and does not obscure any relevant content. Removing the empty line between the text and the buttons, as well as centering the text, worsens readability – we do not see any advantages in this approach.
- Left-aligned dialogs with right-aligned buttons are **common desktop practices** (e.g., Windows, macOS). Centered dialogs are more typical in mobile UIs where space is limited. On the desktop, space shouldn't be an issue.
- Prompts would wrap to two lines in the centered dialog quite easily. The current "airy issue" with short names **could potentially turn into a "squeezed issue"** with long names.

SougandhS:
> I asked our inhouse UX-Experts. This is what they say:

Thanks for checking  ✨

akurtakov:
@SougandhS So are you going to continue with some other ideas here or are you giving up on the idea (and the PR should be closed)?

SougandhS:
> @SougandhS So are you going to continue with some other ideas here or are you giving up on the idea (and the PR should be closed)?

I have a small idea.. [ Not sure everyone agrees :) ]
For small texts can we reduce the width a bit atleast ? 
image

SougandhS:
> I have a small idea.. [ Not sure everyone agrees :) ] For small texts can we reduce the width a bit atleast ? 

image

image

BeckerWdf:
> > I have a small idea.. [ Not sure everyone agrees :) ] For small texts can we reduce the width a bit atleast ?
> 
> image image

Where's the difference?

SougandhS:
> Where's the difference?

Before <br>
image

after <br>

image

BeckerWdf:
and how does this change look for other dialogs?

SougandhS:
> and how does this change look for other dialogs?

Yes did checked one
Before 
image
<br>
After<br>
image

mickaelistria:
I don't see a clear benefit with this change. IMO, the desired format is https://github.com/eclipse-platform/eclipse.platform.ui/pull/2934#issuecomment-2826495220 and anything else is not worth being pursued.
I don't know the details about why https://github.com/eclipse-platform/eclipse.platform.ui/pull/2934#issuecomment-2826495220 cannot be implemented reliably. Is it because changing the horizontal span is required?  Aren't there alternative that can be found? eg is the 2 columns layout used because of image? If so, then couldn't we dynamically decide of the span or layout according to whether image is null or not?

SougandhS:
>eg is the 2 columns layout used because of image? If so, then couldn't we dynamically decide of the span or layout according to whether image is null or not?

I tried modifying `IconAndMessageDialog.getColumnCount()` to return 1, but the extra space remained regardless of Image. It only disappeared after I also updated `data.horizontalSpan = 1` in `MessageDialog.createDialogArea(Composite).`

PR #2937, opened 2025-04-24
Replace deprecated File.toURL()

elsazac:
`java.io.File.toURL()` is deprecated and marked for removal. Replace it with `file.toURI().toURL()`, where `toURL()` is called on a `java.net.URI` instance, which is not deprecated and handles encoding correctly.

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/util/impl/resources/HttpResourcesLocatorImpl.java
+++ b/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/util/impl/resources/HttpResourcesLocatorImpl.java
@@ -13,7 +13,6 @@
  *******************************************************************************/
 package org.eclipse.e4.ui.css.core.util.impl.resources;
 
-import java.io.File;
 import java.io.InputStream;
 import java.net.URL;
 import org.eclipse.e4.ui.css.core.util.resources.IResourceLocator;
@@ -33,7 +32,7 @@ public class HttpResourcesLocatorImpl implements IResourceLocator {
 
 	@Override
 	public InputStream getInputStream(String uri) throws Exception {
-		URL url = new java.net.URL((new File("./")).toURL(), uri);
+		URL url = new URL(uri);
 		return url.openStream();
 	}
 
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceData.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceData.java
@@ -21,8 +21,11 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.StringTokenizer;
 
+import org.eclipse.core.filesystem.URIUtil;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.ConfigurationScope;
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/util/impl/resources/HttpResourcesLocatorImpl.java
```diff
@@ -33,7 +33,7 @@ public String resolve(String uri) {
 
 	@Override
 	public InputStream getInputStream(String uri) throws Exception {
-		URL url = new java.net.URL((new File("./")).toURL(), uri);
+		URL url = new java.net.URL((new File("./")).toURI().toURL(), uri);
```
HannesWell: This could have a few superfluous elements removed:
```suggestion
		URL url = new URL(new File("./").toURI().toURL(), uri);
```
Additionally I wonder if the context URL is even necessary at all?
The JavaDoc of the called `URL(URL context, String spec)` constructor says:
```
If the scheme component is defined in the given spec and does not match the scheme of the context,
then the new URL is created as an absolute URL based on the spec alone.
```
And since the passed string is definitively an `http`-URL and therefore has a scheme that is different from the `file`-scheme, I would say that the context is always discarded anyways.
So it could probably be just:
```suggestion
		URL url = new URL(uri);
```

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceData.java
```diff
@@ -540,13 +540,13 @@ private static URL getPersistenceUrl(URL baseUrl, boolean create) {
 			}
 
 			// make sure the file exists
-			url = new URL(dir.toURL(), PERS_FILENAME);
+			url = new URL(dir.toURI().toURL(), PERS_FILENAME);
```
HannesWell: Looking at the callers, the method `getPersistenceUrl()` should directly return a File or even better a Path.
In general the URL to File/Path conversion done here and in the caller are all actually not robust and only work by chance, i.e. in case the URL does not contain encoded characters.
Instead you should extract the `Path` (`File` would also work, but Path is more powerful) using `Path.of(URIUtil.toURI(baseUrl))` and then perform all resolution and existance checks directly on the Path/File API and never convert back to URL or URI.

HannesWell (changes requested):
Thank you for working on this.
This contributes to the large effort of https://github.com/eclipse-platform/eclipse.platform/issues/35.

bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/util/impl/resources/HttpResourcesLocatorImpl.java
elsazac: Makes sense, thanks for the suggestion.

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/ChooseWorkspaceData.java
elsazac: @HannesWell 

Is this the implementation you were referring to?

```
try {
			// make sure the directory exists
			URL url = new URL(baseUrl, PERS_FOLDER);
			File dir = new File(url.getFile());
			if (!dir.exists() && (!create || !dir.mkdir())) {
				return null;
			}

			// make sure the file exists
			Path p = Path.of(URIUtil.toURI(baseUrl.getPath() + File.pathSeparator + PERS_FILENAME));
			if (!Files.exists(p) && (!create || !p.toFile().createNewFile())) {
				return null;
			}

			return p.toUri().toURL();
		} catch (IOException e) {
			// cannot log because instance area has not been set
			return null;
```

HannesWell: Yes, that's the caller I was referring to.

HannesWell:
Thank you @elsazac for the update.
Unfortunately I misunderstood your previous comment. I meant that the method `getPersistenceUrl()` should just return a Path.
To do this the `Path ` representation of the `baseUrl` should be first extracted by using the method I'm introducing with
- https://github.com/eclipse-equinox/equinox/pull/691

but sadly that will probably not land anymore in this release cycle. But I plan to have this ready early next cycle and I suggest to put this on hold for that time. I'll keep you posted once that is ready, but also invite you to monitor that PR.
That being said, once you got the `Path` of the `baseUrl`, then all checks and concatenations can just be done on the Path object and the same applies for the caller.

elsazac:
> but sadly that will probably not land anymore in this release cycle. But I plan to have this ready early next cycle and I suggest to put this on hold for that time. I'll keep you posted once that is ready, but also invite you to monitor that PR. That being said, once you got the `Path` of the `baseUrl`, then all checks and concatenations can just be done on the Path object and the same applies for the caller.

Sure, I’ll keep an eye on the PR and wait for your update. Once it’s in, I can revisit and refactor accordingly.

PR #2939, opened 2025-04-24
Optimize Windows Defender exclusion script with native PowerShell #2930

### LINKED ISSUE #2930: Script to add an exclusion to Windows Defender may be simplified
[This script](https://github.com/eclipse-platform/eclipse.platform.ui/blob/9a586ab10ad4dd6397a123346303b992cf584a83/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java#L348-L352) to add an exclusion to Windows Defender seems overly complex

Already discussed [here](https://github.com/dbeaver/dbeaver/issues/37722) (DBeaver is built on Eclipse Platform).

## Suggestion

Replace the current

```powershell
$exclusions=@( `
 "C:\Program Files\DBeaver\dbeaver.exe"); `
 $existingExclusions=[Collections.Generic.HashSet[String]](Get-MpPreference).ExclusionProcess; `
 if($existingExclusions -eq $null) { $existingExclusions = New-Object Collections.Generic.HashSet[String] }; `
 $exclusionsToAdd=[Linq.Enumerable]::ToArray([Linq.Enumerable]::Where($exclusions,[Func[object,bool]]{param($ex)!$existingExclusions.Contains($ex)})); `
 if($exclusionsToAdd.Length -gt 0){ Add-MpPreference -ExclusionProcess $exclusionsToAdd }
```

With the simpler

```powershell
Add-MpPreference -ExclusionProcess 'C:\Program Files\DBeaver\dbeaver.exe'
```

## Rationale

I tested the simpler version and it seems that `Add-MpPreference` already does exactly the same work the current script does.

- It will APPEND the item to the list, not to REPLACE it.
- If the list doesn't exists, it will CREATE it.
- Tested on Windows 11 / Windows Powershell 5.1.

## Community

- [X] I understand suggesting an enhancement doesn't mandate anyone to implement it. Other contributors may consider this suggestion, or not, at their own convenience. The most efficient way to get it fixed is that I implement it myself and contribute it back as a good quality patch to the project.

IamLRBA:
cc
@HannesWell
Hello, 
Since the script as per @fdcastel, to add an exclusion to Windows Defender seems overly complex. Based on your concerns about the proposed `Add-MpPreference -ExclusionProcess 'C:\Program Files\DBeaver\dbeaver.exe'` script, this script replaces the original complex LINQ-based exclusion check with a more idiomatic and simpler version PowerShell implementation.

How It Addresses the your Concerns:
- **Duplicate Prevention**: HashSet.Contains() check is just as effective as LINQ
- **Single Admin Context**: All operations happen in one PowerShell invocation
- **No Growing Exclusion Lists**: Only genuinely new paths get added

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2930

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -335,21 +335,11 @@ public class WindowsDefenderConfigurator implements EventHandler {
 
 	public static String createAddExclusionsPowershellCommand(String extraSeparator) {
 		List<Path> paths = getExecutablePath();
-		// For detailed explanations about how to read existing exclusions and how to
-		// add new ones see:
+		// For detailed explanations about how to add new exclusions see:
 		// https://learn.microsoft.com/en-us/powershell/module/defender/add-mppreference?view=windowsserver2019-ps
-		// https://learn.microsoft.com/en-us/powershell/module/defender/get-mppreference?view=windowsserver2019-ps
-		//
-		// For .NET's stream API called LINQ see:
-		// https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable
 		String excludedPaths = paths.stream().map(Path::toString).map(p -> '"' + p + '"')
 				.collect(Collectors.joining(',' + extraSeparator));
-		final String exclusionType = "ExclusionProcess"; //$NON-NLS-1$
-		return String.join(';' + extraSeparator, "$exclusions=@(" + extraSeparator + excludedPaths + ')', //$NON-NLS-1$
-				"$existingExclusions=[Collections.Generic.HashSet[String]](Get-MpPreference)." + exclusionType, //$NON-NLS-1$
-				"if($existingExclusions -eq $null) { $existingExclusions = New-Object Collections.Generic.HashSet[String] }", //$NON-NLS-1$
-				"$exclusionsToAdd=[Linq.Enumerable]::ToArray([Linq.Enumerable]::Where($exclusions,[Func[object,bool]]{param($ex)!$existingExclusions.Contains($ex)}))", //$NON-NLS-1$
-				"if($exclusionsToAdd.Length -gt 0){ Add-MpPreference -" + exclusionType + " $exclusionsToAdd }"); //$NON-NLS-1$ //$NON-NLS-2$
+		return "Add-MpPreference -ExclusionProcess " + extraSeparator + excludedPaths; //$NON-NLS-1$
 	}
 
 	private static void excludeDirectoryFromScanning(IProgressMonitor monitor) throws IOException {


HannesWell:
Thank you for this contribution. The changes look good too me and the script indeed looks cleaner. I'm not very familiar with Powershell and that was what I came up with after searching the internet a bit.

However, I could confirm that @fdcastel's  [objection ](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2930#issuecomment-2825942539) is correct and this script could be even simpler.
If he doesn't want to contribute the simplification by himself, you could update this PR accordingly, but I would await his reaction first.

IamLRBA:
> Thank you for this contribution. The changes look good too me and the script indeed looks cleaner. I'm not very familiar with Powershell and that was what I came up with after searching the internet a bit.
> 
> However, I could confirm that @fdcastel's [objection ](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2930#issuecomment-2825942539) is correct and this script could be even simpler. If he doesn't want to contribute the simplification by himself, you could update this PR accordingly, but I would await his reaction first.

Thanks for your feedback @HannesWell, If @fdcastel approves I can proceed to update the PR...as per your guidance ofcourse!
Cheers!

fdcastel:
Apologies for the delay (I was abroad).

I agree with @HannesWell that the script can be simplified even further.

@IamLRBA Based on my tests (and I believe @HannesWell can confirm as well), just using the call 

```powershell
Add-MpPreference -ExclusionProcess $pathsToExclude
```
should be sufficient. There is no need for any if statements or for instantiating a new `Collections.Generic.HashSet[String]`.

IamLRBA:
Thank you for the feedback @fdcastel . You're right upon re-testing with just: `Add-MpPreference -ExclusionProcess $pathsToExclude`, the command handles duplicates cleanly without throwing errors. I’ll simplify the script accordingly and update the PR shortly.

IamLRBA:
@fdcastel, I've updated the script to simplify it as suggested. It now directly uses `Add-MpPreference -ExclusionProcess $pathsToExclude` without the extra conditionals or set creation. Let me know if anything else needs tweaking.
Thanks!
cc @HannesWell

fdcastel:
Looks good to me! I don’t have a Java setup here to test it out, but the changes seem fine. 🚀 👍🏻

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -342,14 +342,12 @@ public static String createAddExclusionsPowershellCommand(String extraSeparator)
...
 		// https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable
-		String excludedPaths = paths.stream().map(Path::toString).map(p -> '"' + p + '"')
-				.collect(Collectors.joining(',' + extraSeparator));
-		final String exclusionType = "ExclusionProcess"; //$NON-NLS-1$
-		return String.join(';' + extraSeparator, "$exclusions=@(" + extraSeparator + excludedPaths + ')', //$NON-NLS-1$
-				"$existingExclusions=[Collections.Generic.HashSet[String]](Get-MpPreference)." + exclusionType, //$NON-NLS-1$
-				"if($existingExclusions -eq $null) { $existingExclusions = New-Object Collections.Generic.HashSet[String] }", //$NON-NLS-1$
-				"$exclusionsToAdd=[Linq.Enumerable]::ToArray([Linq.Enumerable]::Where($exclusions,[Func[object,bool]]{param($ex)!$existingExclusions.Contains($ex)}))", //$NON-NLS-1$
-				"if($exclusionsToAdd.Length -gt 0){ Add-MpPreference -" + exclusionType + " $exclusionsToAdd }"); //$NON-NLS-1$ //$NON-NLS-2$
+		String excludedPaths = paths.stream().map(Path::toString).map(p -> "'" + p.replace("'", "''") + "'")
```
HannesWell: Can you please explain the advantage of using single-quotes (and escaping them) over double-quotes in a powershell script? As far as I can tell double-quotes are not allowed in Windows paths and therefore don't have to be escaped.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -342,14 +342,12 @@ public static String createAddExclusionsPowershellCommand(String extraSeparator)
 		//
 		// For .NET's stream API called LINQ see:
 		// https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable
```
HannesWell: This entire comment block (starting at its first line) can be replaced by the following:
```
		// For detailed explanations about how to add new exclusions see:
		// https://learn.microsoft.com/en-us/powershell/module/defender/add-mppreference?view=windowsserver2019-ps

```

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -342,14 +342,12 @@ public static String createAddExclusionsPowershellCommand(String extraSeparator)
...
-		String excludedPaths = paths.stream().map(Path::toString).map(p -> '"' + p + '"')
-				.collect(Collectors.joining(',' + extraSeparator));
-		final String exclusionType = "ExclusionProcess"; //$NON-NLS-1$
-		return String.join(';' + extraSeparator, "$exclusions=@(" + extraSeparator + excludedPaths + ')', //$NON-NLS-1$
-				"$existingExclusions=[Collections.Generic.HashSet[String]](Get-MpPreference)." + exclusionType, //$NON-NLS-1$
-				"if($existingExclusions -eq $null) { $existingExclusions = New-Object Collections.Generic.HashSet[String] }", //$NON-NLS-1$
-				"$exclusionsToAdd=[Linq.Enumerable]::ToArray([Linq.Enumerable]::Where($exclusions,[Func[object,bool]]{param($ex)!$existingExclusions.Contains($ex)}))", //$NON-NLS-1$
-				"if($exclusionsToAdd.Length -gt 0){ Add-MpPreference -" + exclusionType + " $exclusionsToAdd }"); //$NON-NLS-1$ //$NON-NLS-2$
+		String excludedPaths = paths.stream().map(Path::toString).map(p -> "'" + p.replace("'", "''") + "'")
+				.collect(Collectors.joining(','));
```
HannesWell: Please keep the addition of the `extraSeparator` as otherwise the formatting in the preferences view doesn't look good.
```suggestion
				.collect(Collectors.joining(',' + extraSeparator));
```

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -342,14 +342,12 @@ public static String createAddExclusionsPowershellCommand(String extraSeparator)
...
+		return String.join(";",
+				"$pathsToExclude = @(" + excludedPaths + ")",
+				"Add-MpPreference -ExclusionProcess $pathsToExclude"
```
HannesWell: Please keep the `extraSeparator`s here as well and keep the name of the name of the variable:
```suggestion
		return String.join(';' + extraSeparator, "$exclusions=@(" + extraSeparator + excludedPaths + ')', //$NON-NLS-1$
				"Add-MpPreference -ExclusionProcess $pathsToExclude"
```
Alternatively, as the script seems now easy enough, we could skip the creation of a variable and just pass the string-array directly to make it even more compact:
```suggestion
		return "Add-MpPreference -ExclusionProcess " + extraSeparator + excludedPaths; //$NON-NLS-1$
```

HannesWell (changes requested):
Thanks for the update.
Unfortunately this currently does not compile and a few points should be adjusted, but I made corresponding remarks below.

Furthermore, please squash your changes into one commit and force push that to your existing branch of this PR. Eventually we want to have this as one commit.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
IamLRBA: I had initially done this to handle the rare case of paths containing single quotes (which are technically valid in Windows paths)
However, as you noted, that might have just added unnecessary complexity.

IamLRBA:
@HannesWell, Thanks so much for the  corrections and changes suggested.
Please look at the current state of the script and we see what more we can change or if it is looking good.
Then we can proceed.
Cheers!

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -340,16 +340,12 @@ public static String createAddExclusionsPowershellCommand(String extraSeparator)
 		// https://learn.microsoft.com/en-us/powershell/module/defender/add-mppreference?view=windowsserver2019-ps
 		// https://learn.microsoft.com/en-us/powershell/module/defender/get-mppreference?view=windowsserver2019-ps
 		//
-		// For .NET's stream API called LINQ see:
-		// https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable
-		String excludedPaths = paths.stream().map(Path::toString).map(p -> '"' + p + '"')
+		// For detailed explanations about how to add new exclusions see:
+		// https://learn.microsoft.com/en-us/powershell/module/defender/add-mppreference?view=windowsserver2019-ps
+
+		String excludedPaths = paths.stream().map(Path::toString).map(p -> "\"" + p + "\"")
```
HannesWell: Please just keep the two double-quote char-literals instead of converting them to string-literals as the latter would require NON-NLS comments (i.e. `//$NON-NLS-1$//$NON-NLS-2$`) and this just adds unnecessary complexity:
```suggestion
		String excludedPaths = paths.stream().map(Path::toString).map(p -> '"' + p + '"')
```

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
```diff
@@ -340,16 +340,12 @@ public static String createAddExclusionsPowershellCommand(String extraSeparator)
 		// https://learn.microsoft.com/en-us/powershell/module/defender/add-mppreference?view=windowsserver2019-ps
 		// https://learn.microsoft.com/en-us/powershell/module/defender/get-mppreference?view=windowsserver2019-ps
 		//
-		// For .NET's stream API called LINQ see:
-		// https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable
-		String excludedPaths = paths.stream().map(Path::toString).map(p -> '"' + p + '"')
+		// For detailed explanations about how to add new exclusions see:
+		// https://learn.microsoft.com/en-us/powershell/module/defender/add-mppreference?view=windowsserver2019-ps
```
HannesWell: Please remove this comment as it duplicates the comments from above.
```suggestion
```

HannesWell (changes requested):
Thank you for the update and sorry for the delayed response.
The code change is mostly fine, just two remarks below.

After you have applied them, please locally squash all changes you made into your first commit and update the commit message if necessary to cover this change as a whole. Then fetch the upstream master and rebase on top of it. After that please force push the updated commit to this PR's branch so that eventually it just contains one commit.
Thank you :)

IamLRBA:
> Thank you for the update and sorry for the delayed response. The code change is mostly fine, just two remarks below.
> 
> After you have applied them, please locally squash all changes you made into your first commit and update the commit message if necessary to cover this change as a whole. Then fetch the upstream master and rebase on top of it. After that please force push the updated commit to this PR's branch so that eventually it just contains one commit. Thank you :)

@HannesWell, I think I have successfully squashed the commits into one

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
HannesWell: Applied this now.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
HannesWell: Applied this now.

HannesWell (approved):
> I think I have successfully squashed the commits into one

Yes you did, thanks for that. But the commit message needed a little tweaking.
Besides that I also applied my two pending requests and now this is ready for submission.

Thank you for this contribution to simplify the code. That's more than welcome.

IamLRBA:
> > I think I have successfully squashed the commits into one
> 
> Yes you did, thanks for that. But the commit message needed a little tweaking. Besides that I also applied my two pending requests and now this is ready for submission.
> 
> Thank you for this contribution to simplify the code. That's more than welcome.

Thank you for the opportunity @HannesWell , You were a very helpful mentor!

PR #2941, opened 2025-04-26
Create disabled shared image in org.eclipse.jface on-the-fly

HeikoKlare:
JFace uses a shared popup menu icon for which the disabled version is loaded as a pre-generated version from a specific file. This change replaces that with a programmatic on-the-fly disablement.


HeikoKlare:
Closing as disabled image is already provided as SVG, so let's keep it like that.

PR #2945, opened 2025-04-26
Use on-the-fly disabled icons in several bundles

HeikoKlare:
This replaces the usage of pre-generated disabled icons in different bundles. Most of the images / image descriptors were not reused anyway, so setting them at different actions is simply removes. Some shared image descriptors are adapted to generate the disabled version of the shared icon programmatically.


HeikoKlare:
Failing test is unrelated and what introduced by: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2899

PR #2948, opened 2025-04-28
Set size of InformationControl before setLocation

### LINKED ISSUE #241: JavaElement hover menu contains unnecessary scrollbars
In 225% (primary) I had the issue that the hover menu for methods or attributes contains scrollbars. It could be a rounding issue (size of the container is 1px too small)

Image

amartya4256:
This PR changes the order of how size and location of a newly created InformationControl is set. The order must be setSize followed by setLocation as setting location of the control first could cause DPI_CHANGE events in win32 since the OS is responsible for providing the size to the control o creation and tthis size can be very big spanning across multiple monitors.

contributes to
https://github.com/eclipse-platform/eclipse.platform.swt/issues/62 and https://github.com/eclipse-platform/eclipse.platform.swt/issues/127

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControlManager.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControlManager.java
@@ -1185,8 +1185,8 @@ abstract public class AbstractInformationControlManager {
 			cropToClosestMonitor(controlBounds);
 			location= Geometry.getLocation(controlBounds);
 			size= Geometry.getSize(controlBounds);
-			informationControl.setLocation(location);
 			informationControl.setSize(size.x, size.y);
+			informationControl.setLocation(location);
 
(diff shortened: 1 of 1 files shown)


fedejeanne (commented):
Since this PR doesn't change any observable behavior, I would change the text of the commit to something a bit simpler like:

```
Avoid unnecessary DPI_CHANGE event in Windows

Set the size of order of the information control before setting its location 
in order to avoid an unnecessary DPI_CHANGE event in Windows.

Setting the location of the control before setting its size could cause a 
DPI_CHANGE event in win32 since the OS is responsible for providing
the size to the control upon creation and this size can be very big, causing 
the control to span across multiple monitors.

contributes to
https://github.com/eclipse-platform/eclipse.platform.swt/issues/62 and
https://github.com/eclipse-platform/eclipse.platform.swt/issues/127
```

Other than that, LGTM ✔️

fedejeanne:
I reworded the commit and merged it. The checks had already passed.

PR #2949, opened 2025-04-29
Newer Workspace Version popup dialog ui alignment changes.

deepika-u:
Before the fix the dialog used to show up as 
image

After the fix the dialog now is showed up as 
image

This popup dialog basically comes when a workspace for example already used/edited/saved in 4.36 and now attempted to open with an older eclipse say 4.35 or lower, this pop up shows up. 

Change 1 : check box is aligned.
Change 2 : Button bar is center aligned.

Here we are changing the behavior of the current popup dialog leaving the original dialog alignments as-is.

### DIFF
--- a/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
+++ b/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2020 IBM Corporation and others.
+ * Copyright (c) 2003, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -56,6 +56,8 @@ import org.eclipse.osgi.service.datalocation.Location;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.layout.FillLayout;
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
@@ -726,6 +728,17 @@ public class IDEApplication implements IApplication, IExecutableExtension {
 			buttonLabelToId.put(IDEWorkbenchMessages.IDEApplication_version_exit, IDialogConstants.CLOSE_ID);
 			MessageDialogWithToggle dialog = new MessageDialogWithToggle(shell, title, null, message, severity,
 					buttonLabelToId, 0, IDEWorkbenchMessages.IDEApplication_version_doNotWarnAgain, false) {
+				@Override
+				protected Control createDialogArea(Composite parent) {
+					Composite composite = (Composite) super.createDialogArea(parent);
+					Button toggle = getToggleButton();
+					if (toggle != null && !toggle.isDisposed()) {
+						GridData toggleData = new GridData(SWT.BEGINNING, SWT.TOP, false, false);
+						toggleData.horizontalIndent = 50;
+						toggle.setLayoutData(toggleData);
+					}
+					return composite;
+				}
 				@Override
 				protected Shell getParentShell() {
 					// Bug 429308: Make workspace selection dialog visible


BeckerWdf:
I don't think that center-aligning the button brings an advantage.

deepika-u:
> I don't think that center-aligning the button brings an advantage.

Thanks for your opinion, will be reverting the 2nd change. Updated as per review comments.

tomaswolf:
Yes, don't center the buttons. If at all that would be done in #2929. (And this dialog here might make a good simple test case to check that whatever is done in #2929 doesn't mess up things.)

deepika-u:
> Yes, don't center the buttons. If at all that would be done in #2929. (And this dialog here might make a good simple test case to check that whatever is done in #2929 doesn't mess up things.)

Now after the fix, only the check box line is aligned and the buttons will continue to right align as earlier. 
From understanding with #2929 and its respective fix only the warning section will be updated but nothing to do with such kind of check box settings given for user to opt. Please correct me if i am not mistaken. 

Now after this fix only the **2nd line is aligned** changing the behavior of the current popup dialog **leaving the original dialog alignments as-is**.
image

deepika-u:
@BeckerWdf : can you check this when you have some time please?

deepika-u:
@tomaswolf @BeckerWdf 
Do you have any further suggestions regarding this fix? 

Here as per the suggestions buttons are right aligned. 
Warning is still left aligned. 
Only the check box part is aligned. Are we good?

BeckerWdf:
Aligning the text and the checkbox is fine for me. I know this dialog a lot. I never felt it was wrong. But I think having the checkbox left aligned (with the icon) indeed is not correct.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -727,6 +729,17 @@ protected ReturnCode checkValidWorkspace(Shell shell, URL url) {
 			MessageDialogWithToggle dialog = new MessageDialogWithToggle(shell, title, null, message, severity,
 					buttonLabelToId, 0, IDEWorkbenchMessages.IDEApplication_version_doNotWarnAgain, false) {
 				@Override
+				protected Control createDialogArea(Composite parent) {
```
BeckerWdf, line 732: is this the right place to fix this. Shouldn't we fix this in MessageDialogWithToggle?
So that for all the dialogs inheriting from this class the checkbox is aligned with the text and not with the icon?

deepika-u: > is this the right place to fix this. Shouldn't we fix this in MessageDialogWithToggle?

Checking on this, will update you soon.

deepika-u: I tried updating the method

deepika-u: Hi @BeckerWdf 
I tried updating the method createToggleButton() as below in MessageDialogWithToggle.java(leaving createDialogArea() untouched) but it is not being honoured to correct the indentation of check box as required.

		GridData data = new GridData(SWT.BEGINNING, SWT.TOP, false, false);
		data.horizontalIndent = 50;
		
Do you have any other thoughts?

deepika-u: @BeckerWdf : Can we merge this fix like this? if you dont have any further suggestions.

deepika-u: @BeckerWdf : When you get some free time, can you check on this and if no further suggestions - can we merge this please?

BeckerWdf: Sorry for missing this.
MessageDialogWithToggle#createToggleButton

looks like this:

```
	protected Button createToggleButton(Composite parent) {
		final Button button = new Button(parent, SWT.CHECK | SWT.LEFT);

		GridData data = new GridData(SWT.NONE);
		data.horizontalSpan = 2;
		button.setLayoutData(data);
		button.setFont(parent.getFont());

		button.addSelectionListener(widgetSelectedAdapter(e -> toggleState = button.getSelection()));

		return button;
	}
```

Wouldn't be more consistent from a UI point of view to add your changes there?

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -727,6 +729,17 @@ protected ReturnCode checkValidWorkspace(Shell shell, URL url) {
...
+					Composite composite = (Composite) super.createDialogArea(parent);
+					Button toggle = getToggleButton();
+					if (toggle != null && !toggle.isDisposed()) {
+						GridData toggleData = new GridData(SWT.BEGINNING, SWT.TOP, false, false);
+						toggleData.horizontalIndent = 50;
```
BeckerWdf, line 737: can you explain this "magic number"?

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
deepika-u: Tried at this place as per my previous comment but didnt help. -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2949#discussion_r2084307894

To be specific i have tried as below::
```
	protected Button createToggleButton(Composite parent) {
		final Button button = new Button(parent, SWT.CHECK | SWT.LEFT);

		GridData data = new GridData(SWT.BEGINNING, SWT.TOP, false, false);
		data.horizontalSpan = 2;
		data.horizontalIndent = 50;
		button.setLayoutData(data);
		button.setFont(parent.getFont());
		button.addSelectionListener(widgetSelectedAdapter(e -> toggleState = button.getSelection()));
		return button;
	}
```

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
deepika-u: It was by trial and error, by only which i could arrive to this value to align it with the text(1st warning line text where it starts) in the above line.

For reference -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2949#issuecomment-2838555786
image

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
BeckerWdf: Can you explain why this did not help?

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
BeckerWdf: May this 50 be wrong on other OSes?

deepika-u: I have only windows to check on. Not sure though.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
deepika-u: As stated above, the same change i have put in the 4.35 maintenance branch and built a complete eclipse local build. Installed and launched that eclipse to open a workspace used by 4.36 till then but the alignment is not corrected as per the code. The same is my observation in -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/2949#discussion_r2084307894

deepika-u:
@BeckerWdf 
Do you have any other way / suggestion to implement this?

deepika-u:
@BeckerWdf : Any other thoughts please?

deepika-u:
@BeckerWdf
Sorry to bother you again, do you have any other suggestions on this please?

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
BeckerWdf: I checked on macOS. Looks like this:
image

Anyway I am still not happy with this magic number. Can this be improved?

BeckerWdf (changes requested):

PR #2963, opened 2025-05-08
LogView: Date column should be more fine grained

chrisrueger:
- and show `yyyy-MM-dd HH:mm:ss.SSS` instead of just minute precision
- this is useful where you want to see the exact time of log events in cases where seconds or even milliseconds matter

Shows this:

image

instead of 

image

### DIFF
--- a/bundles/org.eclipse.ui.views.log/src/org/eclipse/ui/internal/views/log/LogSession.java
+++ b/bundles/org.eclipse.ui.views.log/src/org/eclipse/ui/internal/views/log/LogSession.java
@@ -15,9 +15,11 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.views.log;
 
+import static org.eclipse.ui.internal.views.log.LogEntry.GREGORIAN_SDF;
+import static org.eclipse.ui.internal.views.log.LogEntry.LOCAL_SDF;
+
 import java.io.PrintWriter;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
+import java.time.Instant;
 import java.util.Date;
 
 /**
@@ -31,8 +33,10 @@ public class LogSession extends Group {
 	 * @since 3.5
 	 */
 	public static final String SESSION = "!SESSION"; //$NON-NLS-1$
+
 	private String sessionData;
 	private Date date;
+	private String fDateString;
 
 	public LogSession() {
 		super(Messages.LogViewLabelProvider_Session);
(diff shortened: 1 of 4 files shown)


BeckerWdf:
This sounds like a reasonable thing to do.

BeckerWdf:
Does anyone from the other committers have any objections to merging this change?

sratz:
No objections, I really like this change.

Just some technical part:
In line 109, we could now do

```diff
-return dateFormat.format(entry.getDate());
+return entry.getFormattedDate();
```
That does internal caching of the log date string and might optimize things a bit.

Then, we could also give `org.eclipse.ui.internal.views.log.LogSession` a `getFormattedDate()` method that returns this.

Then get rid off the additional `org.eclipse.ui.internal.views.log.LogViewLabelProvider.dateFormat` in the label provider.

chrisrueger:
Thanks @sratz . Like this https://github.com/eclipse-platform/eclipse.platform.ui/pull/2963/commits/7e88f9f58176dee39a20b39a5f2320c3315dc06b ?

Once approved, I can squash the commits at the end.

Update: Here is a screenshot grouped by Session, to see the new `LogSession.getFormattedDate()`

image

bundles/org.eclipse.ui.views.log/src/org/eclipse/ui/internal/views/log/LogSession.java
```diff
@@ -42,11 +48,24 @@ public Date getDate() {
...
+		}
+		return fDateString;
+	}
+
 	public void setDate(String dateString) {
-		SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS"); //$NON-NLS-1$
 		try {
-			date = formatter.parse(dateString);
-		} catch (ParseException e) { // do nothing
+			LocalDateTime ldt = LocalDateTime.parse(dateString, LOCAL_SDF);
```
sratz: `LogEntry` does

```
Date date = Date.from(Instant.from(GREGORIAN_SDF.parse(stringToParse)));
if (date != null) {
	fDate = date;
	fDateString = LOCAL_SDF.format(fDate.toInstant());
}
```
Not sure what the difference is exactly and why we need a gregorian_sdf also, but since LogEntry is the probably more mature implementation, mabye
- make `org.eclipse.ui.internal.views.log.LogEntry.GREGORIAN_SDF` and `org.eclipse.ui.internal.views.log.LogEntry.LOCAL_SDF` package-protected
- use the constant from over there and just use the same parsing logic here as well.

chrisrueger: A bit more refactoring than I intended to do, but here we go :) https://github.com/eclipse-platform/eclipse.platform.ui/pull/2963/commits/f9166469325cb931ef5a4fc86f4d77cc639d7508

sratz: Thanks @chrisrueger 👍 :)

sratz:
@chrisrueger: Can you sign the ECA with Eclipse? Then we can get this in 2025-09 M1. See https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#setting-up-your-eclipse-and-github-account

chrisrueger:
> @chrisrueger: Can you sign the ECA with Eclipse? Then we can get this in 2025-09 M1. See https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#setting-up-your-eclipse-and-github-account

@sratz Hmm I actually have. This is not my first contribution. Is Eclipse.org having problems? 
I also cannot log in to my account at the moment. I remember there were issues in the past too, when I contributed something together with @laeubi

sratz:
> > @chrisrueger: Can you sign the ECA with Eclipse? Then we can get this in 2025-09 M1. See https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#setting-up-your-eclipse-and-github-account
> 
> @sratz Hmm I actually have. This is not my first contribution. Is Eclipse.org having problems? I also cannot log in to my account at the moment. I remember there were issues in the past too, when I contributed something together with @laeubi

Hmm, yes. I can't even log in to eclipse.org currently.

akurtakov:
Please add all failing ECA cases to https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6246 . As it's sporadic, let's hope that with some instance they will manage to figure what actually happens.

PR #2975, opened 2025-05-13
Improve empty workspace helper layout

### LINKED ISSUE #281: Links in empty Package Explorer are cut off
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003

akoch-yatta:
This PR adjusts the grid layout in EmptyWorkspaceHelper to address an issue in Windows with monitor specific scaling active. This commit will not fully solve the issue, but significantly lower the scenarios when it is triggered.

It reduces the effect the mentioned issue by letting the parent Composite fill its that. That way the issue will only occur when the View has a specific width. The proper fix will need additional effort in SWT which could cause other regressions, therefor I'd recommend this PR this late in the release cycle.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003

The Border are of course only added temporarily for the screenshots

#### Layout Before
ewh_before

#### Layout After
ewh_after


BeckerWdf (approved):
sounds reasonable

HeikoKlare (approved):
Looks good, thank you! Works as expected and (almost) resolves https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003

PR #2977, opened 2025-05-13
Add MouseEnter & MouseMove Event in PopupCloser

### LINKED ISSUE #280: [Edge] JavaDoc does not react to hovering/clicking
~See https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2061~
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2072

## Summary

- Hovering/clicking a JavaDoc should show the scrollbars and navigation controls.
- Using `Tab` or clicking in the lower part of the JavaDoc popup works as expected.

I assume this has something to do with an event not being properly triggered/handled by Edge.

## PRs
- [x] https://github.com/eclipse-platform/eclipse.platform.ui/pull/2977
- [ ] https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2289

### LINKED ISSUE #2072: [Regression] Can't click on JavaDoc popup, resize or scroll
From: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2187

Version: 2025-06 (4.36)
Build id: I20250423-0800
Windows 10

Image

Opening a popup JavaDoc as shown above does not allow to scroll it, making the continuation unreadable. It's also not resizable. In fact, clicking on it does nothing, but if you click on it and then switch to another non-eclipse window, the popups stay as an overlay.

This is a regression. In
Version: 2024-12 (4.34)
Build id: I20241120-1800

It is both resizable and scrollable, and the controls on the bottom appear. The windowing also behaves correctly after clicking on it.

Image

## Workaround
Use the `Tab` key to move from the forst popup to the JavaDoc popup

Image

amartya4256:
This commit adds MouseEnter and MouseMove event in the PopupCloser, as Edge browser can only react to mouse movements and not mouse clicks. This also makes the beahviour of PopupCloser consistent with AbstractHoverInformationControl:Closer.

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1551

fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2072

**Explanation:** The reason why edge browser doesn't react to a mouse hover over the jvadoc tooltip is because the controller responsible for managing the javadoc tooltip in this case is AdditionalInfoController and the closer needed to replace the tooltip on the mouse event is a PopUpCloser. The PopUpCloser doesn't have any mouse movement events (MouseMove & MouseEnter) registered like AbstractHoverInformationControl:Closer (Which is responsible for javadoc tooltip on code hover), hence the javadoc tooltip is not replaced.

**Fix:** Simply register PopUpCloser with the MouseMove and MouseEnter events and handle these events to trigger the replacement of the control similar to AbstractHoverInformationControl:Closer.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/PopupCloser.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/PopupCloser.java
@@ -109,6 +109,9 @@ class PopupCloser extends ShellAdapter implements FocusListener, SelectionListen
 
 			fDisplay.addFilter(SWT.Deactivate, this);
 			fDisplay.addFilter(SWT.MouseUp, this);
+
+			fDisplay.addFilter(SWT.MouseMove, this);
+			fDisplay.addFilter(SWT.MouseEnter, this);
 		}
 	}
 
@@ -130,6 +133,9 @@ class PopupCloser extends ShellAdapter implements FocusListener, SelectionListen
 
 			fDisplay.removeFilter(SWT.Deactivate, this);
 			fDisplay.removeFilter(SWT.MouseUp, this);
+
+			fDisplay.removeFilter(SWT.MouseMove, this);
+			fDisplay.removeFilter(SWT.MouseEnter, this);
 		}
 	}
 
@@ -206,6 +212,8 @@ class PopupCloser extends ShellAdapter implements FocusListener, SelectionListen
 				}
 				break;
 
+			case SWT.MouseEnter:
+			case SWT.MouseMove:
 			case SWT.MouseUp:
 				if (fAdditionalInfoController == null || fAdditionalInfoController.getInternalAccessor().isReplaceInProgress())
 					break;
@@ -224,7 +232,7 @@ class PopupCloser extends ShellAdapter implements FocusListener, SelectionListen
 							}
 
 							// XXX: workaround for https://bugs.eclipse.org/bugs/show_bug.cgi?id=212392 :
-							control.getShell().getDisplay().asyncExec(() -> fAdditionalInfoController.getInternalAccessor().replaceInformationControl(true));
+							fAdditionalInfoController.getInternalAccessor().replaceInformationControl(event.type == SWT.MouseUp);
 						}
 					}
 				}
--- a/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.editors; singleton:=true
-Bundle-Version: 3.20.0.qualifier
+Bundle-Version: 3.20.100.qualifier
 Bundle-Activator: org.eclipse.ui.internal.editors.text.EditorsPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/editors/text/TextEditorMessages.properties
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/internal/editors/text/TextEditorMessages.properties
@@ -12,7 +12,7 @@
 #     IBM Corporation - initial API and implementation
 ###############################################################################
 
-EditorsPlugin_additionalInfo_affordance=Press 'Tab' from proposal table or click for focus
+EditorsPlugin_additionalInfo_affordance=Press 'Tab' from proposal table or click/hover for focus
 EditorsPlugin_internal_error=Internal Error
 
 TextEditorPreferencePage_displayedTabWidth=Displayed &tab width:


fedejeanne:
The failing check seems unrelated and reminds me of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/2406. The error message this time is slightly different though:

```
13:48:57.785 [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.13-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.ui.workbench3: Baseline and reactor have the same fully qualified version, but different content
13:48:57.786 [ERROR] different
```

@merks / @iloveeclipse do you have any hints? Is the error safe to be ignored or should I try some fix? Thank you in advance

fedejeanne (commented):
The changes look fine and I tested **on Windows**, everything is fine. I haven't tested on Linux or Mac yet.

I would still like to **postpone merging this PR to the next M1**. The reason for that is that the affected class is quite central and the logic involved is not trivial. Applying this PR means that now 2 additional mouse events are handled/propagated and since there are so many subclasses of `IInformationControlExtension5` (which is official API) that could be receiving these 2 new events, this rather small change could have big repercussions in consumer code.

Also, the original issue (https://github.com/eclipse-platform/eclipse.platform.swt/issues/2072) does have a functioning **workaround** and it's very easy: simply press the `Tab` key. 

Last but not least, the (first) candidate build for M3 is going to be built today so there is no time to test it properly in the community. I'd hate to introduce some unforeseen bug just because we rushed this PR.

@vogella, @akurtakov I see that you were the last 2 people to do working changes to this class so I'd like to know what do you think.

HeikoKlare (commented):
In general, I think this change is fine both from technical side (to support proper usage of Edge inside PopupCloser) as well as from UX perspective (to replace the information control when hovering over it).

Still, some thoughts/concerns on this:
- The PopupCloser shows the information "Press `Tab`from proposal table or click to focus", which becomes invalid/incomplete with this change: it also receives focus when just hovering over it.
- Regarding focus: would it be possible to only replace the control when hovering over it without giving it focus and would that make sense? I find it a bit strange that I change the focus control just by hovering as that's quite unusual. When pressing tab or clicking that's kind of expected but not when hovering.
- When repeatedly hovering over the control, it seems to be replaced repeatedly (at least it flickers every time I click in the completion proposal popup and then hover over the additional information control). While that's fine when navigating via tab or when activating it by clicking into it, it's a bit strange that this repeatedly happens just on hovering over the already replaced additional information control, as you can see in this video:
popup_closer

amartya4256:
> In general, I think this change is fine both from technical side (to support proper usage of Edge inside PopupCloser) as well as from UX perspective (to replace the information control when hovering over it).
> 
> Still, some thoughts/concerns on this:
> 
> * The PopupCloser shows the information "Press `Tab`from proposal table or click to focus", which becomes invalid/incomplete with this change: it also receives focus when just hovering over it.

Shall we change the description?

> * Regarding focus: would it be possible to only replace the control when hovering over it without giving it focus and would that make sense? I find it a bit strange that I change the focus control just by hovering as that's quite unusual. When pressing tab or clicking that's kind of expected but not when hovering.

Implemented


> * When repeatedly hovering over the control, it seems to be replaced repeatedly (at least it flickers every time I click in the completion proposal popup and then hover over the additional information control). While that's fine when navigating via tab or when activating it by clicking into it, it's a bit strange that this repeatedly happens just on hovering over the already replaced additional information control, as you can see in this video:

- This happens because when you click back on the proposal, it replaces the already replaced control with the stage 1 control again (The one with no scroll bar) and then when you go back to the browser window it gets replaced again - leading to an extra replacement. It is identical behaviour to on IE with replace only on mouse Up. I tested it out. Although the replacement is super quick now the moment the cursor goes inside the browser window, that's why it feels snappy. I could potentially try to prevent it if it is possible. But I would propose we do that in a separate issue as an enhancement over the the mouse hover because we need to test this bheaviour for PopUpCloser with all the other consumer first and once we are sure, its not causing any problem, we can move ahead with enhancing it further.

HeikoKlare:
> > The PopupCloser shows the information "Press Tabfrom proposal table or click to focus", which becomes invalid/incomplete with this change: it also receives focus when just hovering over it.

> Shall we change the description?

Yes, we need to do that.

> > When repeatedly hovering over the control, it seems to be replaced repeatedly (at least it flickers every time I click in the completion proposal popup and then hover over the additional information control). While that's fine when navigating via tab or when activating it by clicking into it, it's a bit strange that this repeatedly happens just on hovering over the already replaced additional information control, as you can see in this video:

> This happens because when you click back on the proposal, it replaces the already replaced control with the stage 1 control again (The one with no scroll bar) and then when you go back to the browser window it gets replaced again - leading to an extra replacement. It is identical behaviour to on IE with replace only on mouse Up. I tested it out. Although the replacement is super quick now the moment the cursor goes inside the browser window, that's why it feels snappy. I could potentially try to prevent it if it is possible. But I would propose we do that in a separate issue as an enhancement over the the mouse hover because we need to test this bheaviour for PopUpCloser with all the other consumer first and once we are sure, its not causing any problem, we can move ahead with enhancing it further.

You are right. I was moving the mouse too fast so it looked a bit weird. But the behavior as you describe it totally makes sense and in my opinion is nothing we should/need to fix.

fedejeanne (approved):

PR #2983, opened 2025-05-15
Add missing SVG wizard banner icon in org.eclipse.ltk.ui.refactoring

BeckerWdf:
This adds SVG versions of the banners for the "Show Refactoring History Wizard".

See also https://github.com/eclipse-platform/eclipse.platform.images/pull/145


BeckerWdf:
Shouldn't we also delete the now unused PNGs?

HeikoKlare:
Thank you for creating the icon!

> Shouldn't we also delete the now unused PNGs?

Currently, we consider the PNG icons as deprecated for removal (i.e., to be removed in 2 years) and mention that in the according news: https://eclipse.dev/eclipse/markdown/?f=news/4.36/platform.md#support-for-svg-images
If I am not mistaken, this was what we agreed on in a community call to ensure that consumers bundles having direct references to those resources inside bundle have some time to adapt (if they even get aware of the upcoming removal as we cannot provide the information via code). Should we reconsider the decision?

HeikoKlare (approved):

PR #2985, opened 2025-05-15
Always set disabled icons in AbstractContributionItem

HeikoKlare:
AbstractionContributionItems only set a disabled icon on the ToolItem in case such a disabled icon is specified via an explicit URI to that icon. If no such URI is specified, the ToolItem creates a disabled icon on the fly. Since the ToolItem creates a new image every time the enablement state of the ToolItem changes, this leads to unnecessary repeated creation of images.
In contrast, ActionContributionItems always set the disabled icon for their ToolItem, even if no disabled image descriptor is specified for them, to avoid that the ToolItem repeatedly created that icon.

This change align the AbstractContributionItem handling for disabled icons with the one of ActionContributionItems: in case no URI to a disabled icon is specified, the SWT disablement mechanism is used to create a disabled version and to explicitly set it at the ToolItem. This avoids unnecessary repeated creation of the disabled version of an icon for those items.

@vogella maybe you know this code best and have an opinion on it? It would be nice to bring this in for M3/RC1 as otherwise the replacement of the pre-generated disabled icons (PNGs) with on-the-fly generated disabled version (based on SVGs) will lead to unnecessary, repeated image creations in the upcoming release.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/AbstractContributionItem.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/AbstractContributionItem.java
@@ -149,10 +149,17 @@ public abstract class AbstractContributionItem extends ContributionItem {
 	}
 
 	protected Image getImage(String iconURI, LocalResourceManager resourceManager) {
+		return createImage(iconURI, resourceManager, false);
+	}
+
+	private Image createImage(String iconURI, LocalResourceManager resourceManager, boolean disabled) {
 		Image image = null;
 
 		if (iconURI != null && iconURI.length() > 0) {
 			ImageDescriptor iconDescriptor = resUtils.imageDescriptorFromURI(URI.createURI(iconURI));
+			if (disabled) {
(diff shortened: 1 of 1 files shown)


laeubi (approved):
In general I think this sounds reasonable to do so.

At best the renderer would even do it lazy:

in the `AbstractContributionItem.getUpdateRunner()` the visible state of the item is updated, so one could there check if `enabled = false` and then generate and set the disabled image at that point in time with `modelItem.getWidget()`

HeikoKlare:
Thank you, @laeubi! 

> in the AbstractContributionItem.getUpdateRunner() the visible state of the item is updated, so one could there check if enabled = false and then generate and set the disabled image at that point in time with modelItem.getWidget()

I just found that `updateIcons()` is indeed executed on every state change (including enablement), but if the URI does not change, the icons are only generated once. I thus adapted the PR to add the on-the-fly-generated disabled image when `updateIcons()` is called for a disabled model item. This is not supposed to be the final solution. I would propose to completely revise the initialization to only create the enabled image and then create the disabled image (no matter if explicitly provided or generated on-the-lay) lazyily. But since that change would be more invasive and may have unexpected side effects, I would postpone it as a follow-up change for the next development cycle.

laeubi (approved):
Looks good to me

HeikoKlare:
Thank you!

PR #2925, opened 2025-04-15
Add SVGs to org.eclipse.e4.ui bundles

Michael5601:
This PR adds SVGs for all icons in the bundles `org.eclipse.e4.ui.progress`, `org.eclipse.e4.ui.workbench.swt`, `org.eclipse.e4.ui.workbench.renderers.swt`, `org.eclipse.e4.ui.workbench.addons.swt` and `org.eclipse.e4.ui.tests`, `org.eclipse.e4.ui.dialogs`.

See also this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1797) for more information.

PR #2994, opened 2025-05-19
Fix view_menu.svg

sratz:
view_menu.svg was the wrong svg file, not matching the .pngs.

Follow-up on #2925.

Supercedes and thus closes #2995.


BeckerWdf (approved):

sratz:
Was:
image

Should be:
image

HeikoKlare:
Just saw that we did the same thing at almost the same time :-)
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2995

However, note that the icon you propose is not correct. The existing SVG is not properly centered/sized. This is how it looks for me with your PR:
image

sratz:
Oh wow, we have like a million different versions of the same svg lying around :D

HeikoKlare:
> Oh wow, we have like a million different versions of the same svg lying around :D

Yeah, there were many SVGs for this, some were completely "wrong" (like the one you replaced), some just wrongly size. But actually I found none that completely fits to the existing PNGs. So the SVG I propose in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2995 is even a new one created out of the existing ones. Maybe I just missed the "correct" one lying around somewhere.

BeckerWdf:
> Oh wow, we have like a million different versions of the same svg lying around :D

Ups. Sorry. Maybe I messed this up years ago without noticing.

sratz:
I updated all the icons with https://github.com/eclipse-platform/eclipse.platform.images/pull/147, which is based on @HeikoKlare 's icon proposed in #2995.

HeikoKlare (approved):
Looks really great. Thank you, @sratz!

The icon now fits even better to the minimize/maximize icons next to them in the part stacks, as the stroke widths are better aligned. In addition, I tested on all fractional scales between 100% and 200% monitor zoom and it really looks good and much better than the previous PNG icon on all zoom levels.


Btw: the minimize/maximize icons are rendered programmatically. I quickly tried to do the same for the view_menu icon, but unfortunately GC-rendered circles do not look that good (even with anti-aliasing), which is why I cannot recommend trying out.

sratz:
Thanks, @HeikoKlare for the tests and the initial version of the updated icon!

I suppose we are good to merge this for RC1?

HeikoKlare:
> I suppose we are good to merge this for RC1?

Yes, +1 for doing so.

BeckerWdf (approved):

PR #2995, opened 2025-05-19
Correct SVG icon for showing menu in part stacks and shared images

HeikoKlare:
The SVG version of the icon for showing the menu in part stacks erroneously contained the "thin" version of that icon, showing an error instead of three dots. In addition, the shared image version of that icon contained the three dots shifted to the bottom right and smaller than intended.

This change replaces both SVG version with the same icon containing the correct three-dot-version of the view_menu icon.

### Before
image

### After
image

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/icons/full/elcl16/view_menu.svg
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/icons/full/elcl16/view_menu.svg
@@ -2,61 +2,82 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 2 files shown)


sratz:
I don't see this particular version of `view_menu.svg` (3656 bytes) anywhere else.
Was this corrected for this PR?

HeikoKlare:
> Was this corrected for this PR?

Yes, I took the original `view_menu.svg` from `org.eclipse.ui` and adapted it. I did not find an SVG version properly representing the existing PNGs, as all versions of that SVG had the three circles shifted to the bottom right and wrongly sized. Maybe I missed the correct one...

HeikoKlare:
Note: I am not an expert in creating SVGs, so feel free to check/update the proposed SVG. I tried to uniformly align the circles with the 16x16 grid to ensure that rasterization produces as sharp as possible results.

sratz:
> Note: I am not an expert in creating SVGs, so feel free to check/update the proposed SVG. I tried to uniformly align the circles with the 16x16 grid to ensure that rasterization produces as sharp as possible results.

Hmm, it does look different than the old .png. It's almost like it's squares, not circles.

sratz:
That icon was quite the mess:

Let's start at the source: I tried to re-create the PNGs that have been used over the last years as closely as possible in https://github.com/eclipse-platform/eclipse.platform.images/pull/147 and re-rendered everything.

BeckerWdf:
Can I help here or did you sort this out in the meantime?

sratz:
Closing in favor of #2994.

PR #2998, opened 2025-05-20
Workaround for unexpected scrollbars in MultipleHyperlinkPresenter

akoch-yatta:
This PR adds two additional points to the width and height when calculating the size of the composite containing the Hyperlinks in the MultipleHyperlinkPresenter. This serves as a workaround for a current limitation in the SWT implementation on Windows. With certain zoom settings (e.g., 125%, 175% or 225%), the calculated size may be too small, causing the Composite to show Scrollbars, although it calculated to not need scrollbars. This additional width/height is only added for windows and when no scrollbars should be thrown. This is intended as a temporary workaround.

**This is intended as a temporary workaround.** If we merge this, we should either revert it after the release to reintroduce the issue or have hopefully a proper fix inside of SWT. Still, as it affects a quite central element, I would propose to merge this PR.

#### How to test

Use this Snippet and open it on a 225% monitor.
```java
package test;

public class TestClass {
		// TODO Auto-generated method stub
	}
}
```
Without the PR you will see:
image

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/hyperlink/MultipleHyperlinkPresenter.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/hyperlink/MultipleHyperlinkPresenter.java
@@ -152,6 +152,14 @@ public class MultipleHyperlinkPresenter extends DefaultHyperlinkPresenter implem
 			fForegroundColor= foregroundColor;
 			fBackgroundColor= backgroundColor;
 			create();
+			getShell().addListener(SWT.ZoomChanged, event -> {
+				final Shell shell= getShell();
+				shell.getDisplay().asyncExec(() -> {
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
I've tested the workaround in a current SDK. It properly solves the mentioned scrollbar issue at all zooms that I was able to test (100, 125, 150, 175, 200, 225).

Since having those scrollbars would be annoying and significantly impact usability, this should go into RC1. I considers this safe to do because:
- The popup shell is self-contained, so the compensation values for autoscale rounding issues will not affect any enclosing widgets, layouts or the like. It will just slighly increase the shell size. And they are only applied when running on Windows, on which the changes have been tested.
- The zoom change listener will just reset the shell size. The shell is not manually resizable, so nothing can be overwritten here. It only happens on zoom change, thus if switching back to non-monitor-specific scaling, it would not even be executed.

PR #3001, opened 2025-05-22
Docu: Adding the org.eclipse.ui.selection to core expression document

vogella:


docs/Command_Core_Expressions.md
```diff
@@ -38,6 +38,7 @@ Some of the variables may not be set, depending on the current application conte
 | activePart | `org.eclipse.ui.IWorkbenchPart` |   The active part, which can be the same as the active editor.   | 3.2 |
 | activePartId | `java.lang.String` |   The ID of the currently active part.   | 3.2 |
 | activeSite | `org.eclipse.ui.IWorkbenchPartSite` |   The site of the currently active part.   | 3.2 |
+| org.eclipse.ui.selection | Object |   The current global selection if the `ESelectionService` is used. It is often used with <test/> elements with `org.eclipse.core.expressions.PropertyTester`, in programmatic core expressions, potentially with <iterate/> and <count/> elements.   | 4.0 |
```
merks: I wonder if you've previewed this?  I ask because one has to be careful with markdown that looks like XML, 
i.e., `<test/>` versus `>`<test/>`<`.

image

merks (changes requested):

docs/Command_Core_Expressions.md
vogella: The description was copied from one line below https://github.com/eclipse-platform/eclipse.platform.ui/pull/3001/files#diff-579451fec93073440823c9b658e9877518efbd6a40ae1ce5cfebea9a4aca105dR42

vogella: But true, the original line also does not render well: https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/docs/Command_Core_Expressions.md

merks: I'm not really sure what you are trying to tell me.  It doesn't really matter where it came from if it was already wrong where it come from, but probably I miss the point about.  I really am not sure what is at the targets of the two link that should draw my attention.

---
Maybe it's more clear if I make a picture of the following two lines:

See this line is blank: "<test/> <iterate/> <count/>"
This line is not: "`<test/>`  `<iterate/>`  `<count/>`"

Here is what I wrote:

image

Here is how it renders/previews:

image

I'm just trying to tell you that you need to use back quotes on markup that looks like XML element tag otherwise it will disappear and we don't want that.

vogella: @merks your feedback was clear, is valuable and will be acted upon. (as soon as I find the time).  I just tried to explain how this happened; I copied an existing line (which also has the described issue).

merks:
```suggestion
| org.eclipse.ui.selection | Object |   The current global selection if the `ESelectionService` is used. It is often used with `<test/>` elements with `org.eclipse.core.expressions.PropertyTester`, in programmatic core expressions, potentially with `<iterate/>` and `<count/>` elements.   | 4.0 |
```

merks: I see.  😁 

I just proposed the change without having see your comment first.  It's all fine other than this detail.

Also, as a general suggestion, when changing markdown, there is really no need to wait for a build.  I never do.  And anything that might need to be fixed can always be fixed after the fact...

merks:
This looks inconsistent with respect to the other table entries:

image

I guess it was never really previewed in the past:

image

It would be nicer if section titles were title case:

image

Do the changes need to be multiple commits or could we do them with a single commit?

vogella:
I will push an update of the PR for the errors I find.

vogella:
@merks please re-check

merks (approved):
I've not done a detailed review but it looks far better.  Thank you!

vogella:
Just docu, ignoring the "all the time and therefore not so random anymore infrastructure issues" build failure.

@merks thanks for the help

vogella:
Clicked the wrong button, will force push to remove the merge commit.

PR #3002, opened 2025-05-22
Add a negative requirement on darkest dark 2024.4.0.202412101111

merks:
https://github.com/eclipse-packaging/packages/issues/310

### DIFF
--- a/bundles/org.eclipse.jface/META-INF/p2.inf
+++ b/bundles/org.eclipse.jface/META-INF/p2.inf
@@ -1,0 +1,7 @@
+# Ensure that 4.36 is not installed with the 0.0.1,2024.4.0.202412101111 version of darkest dark devstyle.
+# https://github.com/eclipse-packaging/packages/issues/307
+requires.0.namespace=org.eclipse.equinox.p2.iu
+requires.0.name=com.genuitec.eclipse.theming.core.feature.feature.group
+requires.0.range = [0.0.1,2024.4.0.202412101111]
+requires.0.min = 0
+requires.0.max = 0


merks:
Probably this is the best place because this is where the class was deleted.

bundles/org.eclipse.jface/META-INF/p2.inf
```diff
@@ -0,0 +1,7 @@
+# Ensure that 4.36 is not installed with the 0.0.1,2024.4.0.202412101111 version of darkest dark devstyle.
+# https://github.com/eclipse-packaging/packages/issues/307
+requires.0.namespace=org.eclipse.equinox.p2.iu
+requires.0.name=com.genuitec.eclipse.theming.core.feature.feature.group
+requires.0.range = [0.0.1,2024.4.0.202412101111]
```
laeubi, line 5:
```suggestion
requires.0.range = [0.0.1,9999)
```

As we don't know in what version it will be fixed or if it ever will be fixed.

iloveeclipse: Compromise: `[0.0.1,2025)` ?

laeubi: > Compromise: `[0.0.1,2025)` ?

Sounds reasonable to me, lets hope we don't need to increase this every year :-P

merks: I'm sorry guys, but I've spent the entire day on this mess, and I'm **really** at the end of my patience.  Really, really.  I've had two wrong suggested changes so far and now one irrelevant suggest change.   Why can we not simply include the known broken version in the upper bound (that I've already tested) and call it done?  Maybe the number is not pretty enough? Not round enough?  My patience for this is definitely not enough. 😱 😭

iloveeclipse: > Maybe the number is not pretty enough? Not round enough? My patience for this is definitely not enough.

Exact :-)

Ed, we don't know what is the version policy there and whether a "breaking" change like in 2025.06 release would be fixed only in the next major bundle version. So having at least "not minor" version bump is something I would like to have to make sure we don't need to bump this version again here...

jonahgraham (approved):
Well I have asked genuitec to engage in the open on this. If they do they should be able to tell us an upper bound. If they don't we should take our best guess.

I think best guess is blocking currently published version. While genuitec aren't engaging very openly, they are at least saying they are working on this.

So +1 to how Ed has it
+0 to 2025)
And -1 to 9999)

bundles/org.eclipse.jface/META-INF/p2.inf
HeikoKlare: I have to admit that I do not seem to understand the version constraints properly, as otherwise I would have also been in favor of using `[0.0.1,2024.4.1)` or `[0.0.1,2025.0.0)` (to force an "active version bump", not qualifier update or the like).
But, if I understand correct, since you are confident they will not work or they are not reasonable, I give +1 to the existing proposal. And of course, thank you very much for taking care of the change and being patient with all proposals made on it!

akurtakov: Let's go the current way for this time . If the issue repeat we will have to go far stronger.

akurtakov (approved):

PR #3005, opened 2025-05-22
Make DnD consumer utilize internal state of Rectangle with APIs of and clone

### LINKED ISSUE #246: Drag and Drop Region moves with a different scale as compared to the cursor between 2 monitors
**Configuration for reproducing:** 

1. Right (Primary) - 150%
2. Left - 100%

**Steps to reproduce:**

- Keep your IDE on the primary monitor.
- Drag an editor across the monitors.
- The drag and drop rectangle moves away from the cursor.

Expected behaviour: The Drag and Drop region should always have its top left corner around the cursor.

### LINKED ISSUE #338: Adapt DnD consumer to use Rectangle.clone method to not lose internal references for a better Drag and Drop experience
As an eclipse user, I want to have a flawless CTab Drag and Drop experience between multiple monitors with different zooms.

**Configuration for reproducing:**

Right (Primary) - 150%
Left - 100%
Steps to reproduce:

- Keep your IDE on the primary monitor.
- Drag an editor across the monitors.
- The drag and drop rectangle moves away from the cursor.
- Expected behaviour: The Drag and Drop region should always have its top left corner around the cursor.

This happens because when a rectangle is copied into a new object using the default constructor, it loses its monitor information. However, the cursor location always has the right monitor zoom in the DnDManager and the rectangle created there should use that monitor reference while creating a rectangle from the cursor location and this rectangle should further only be copied using clone method to copy the right internal state of the rectangle so that we still have the monitor information - leading to right scaling.

amartya4256:
**Depends on:** https://github.com/eclipse-platform/eclipse.platform.swt/pull/2186

This PR enforces the usage of Rectangle.copyWith and Rectangle.of
while copying a rectangle from an existing rectangle and creating a
rectangle from a Point making sure the internal state of the source
resources are copied to the new resource.

**Configuration for reproducing:**

Right (Primary) - 150%
Left - 100%
Steps to reproduce:

- Keep your IDE on the primary monitor.
- Drag an editor across the monitors.
- The drag and drop rectangle moves away from the cursor.
- Expected behaviour: The Drag and Drop region should always have its top left corner around the cursor.

This happens because when a rectangle is copied into a new object using the default constructor, it loses its monitor information. However, the cursor location always has the right monitor zoom in the DnDManager and the rectangle created there should use that monitor reference while creating a rectangle from the cursor location and this rectangle should further only be copied using clone method to copy the right internal state of the rectangle so that we still have the monitor information - leading to right scaling.

contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/62 and https://github.com/eclipse-platform/eclipse.platform.swt/issues/128

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.addons.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.workbench.addons.swt/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.workbench.addons.swt;singleton:=true
-Bundle-Version: 1.5.700.qualifier
+Bundle-Version: 1.5.800.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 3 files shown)


fedejeanne (approved):
Tested it, looks good. Thank you @amartya4256 !

PR #3009, opened 2025-05-24
Add explicit type check to UpdateStrategy.convert()

### LINKED ISSUE #3008: ClassCastException in UpdateValueStrategy

## Steps to reproduce

Execute the following test case in `UpdateValueStrategyTest`:

```java
@Test
public void testUpdateValueWithStrategy() throws Exception {
	WritableValue<Integer> model = WritableValue.withValueType(Integer.class);
	WritableValue<Boolean> target = WritableValue.withValueType(Boolean.class);

	/*
	UpdateValueStrategy strategy1 = new UpdateValueStrategy() {
		@Override
		protected IStatus doSet(IObservableValue observableValue, Object value) {
			Integer modelValue = (Integer) value;
			if (modelValue != null && modelValue != 0) {
				return super.doSet(observableValue, true);
			}
			return super.doSet(observableValue, false);
		}
	};
	*/

	UpdateValueStrategy<Integer, Boolean> strategy2 = new UpdateValueStrategy<>() {
		@Override
		protected IStatus doSet(IObservableValue<? super Boolean> observableValue, Boolean value) {
			// ????
			return super.doSet(observableValue, value);
		}
	};

	DataBindingContext context = new DataBindingContext();
	context.bindValue(target, model, null, strategy2);

	model.setValue(1);
	assertEquals(target.getValue(), true);
}
```

When using `strategy1`, the test succeeds as expected. But when using `strategy2`, the following exception is thrown:

```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Boolean (java.lang.Integer and java.lang.Boolean are in module java.base of loader 'bootstrap')
	at org.eclipse.core.tests.databinding.UpdateValueStrategyTest$1.doSet(UpdateValueStrategyTest.java:1)
	at org.eclipse.core.databinding.ValueBinding.lambda$1(ValueBinding.java:206)
	at org.eclipse.core.databinding.Binding.lambda$3(Binding.java:191)
	at org.eclipse.core.databinding.observable.Realm$1.run(Realm.java:155)
	at org.eclipse.core.databinding.observable.Realm.safeRun(Realm.java:159)
	at org.eclipse.core.databinding.observable.Realm.exec(Realm.java:177)
	at org.eclipse.core.databinding.Binding.execAfterDisposalCheck(Binding.java:189)
	at org.eclipse.core.databinding.ValueBinding.lambda$0(ValueBinding.java:199)
	at org.eclipse.core.databinding.Binding.lambda$3(Binding.java:191)
	at org.eclipse.core.databinding.observable.Realm$1.run(Realm.java:155)
	at org.eclipse.core.databinding.observable.Realm.safeRun(Realm.java:159)
	at org.eclipse.core.databinding.observable.Realm.exec(Realm.java:177)
	at org.eclipse.core.databinding.Binding.execAfterDisposalCheck(Binding.java:189)
	at org.eclipse.core.databinding.ValueBinding.doUpdate(ValueBinding.java:163)
	at org.eclipse.core.databinding.ValueBinding$2.handleValueChange(ValueBinding.java:55)
	at org.eclipse.core.databinding.observable.value.ValueChangeEvent.dispatch(ValueChangeEvent.java:69)
	at org.eclipse.core.databinding.observable.ChangeManager.fireEvent(ChangeManager.java:118)
	at org.eclipse.core.databinding.observable.value.AbstractObservableValue.fireValueChange(AbstractObservableValue.java:86)
	at org.eclipse.core.databinding.observable.value.WritableValue.doSetValue(WritableValue.java:101)
	at org.eclipse.core.databinding.observable.value.AbstractObservableValue.setValue(AbstractObservableValue.java:70)
	at org.eclipse.core.tests.databinding.UpdateValueStrategyTest.testUpdateValueWithStrategy(UpdateValueStrategyTest.java:216)
```

## Community

ptziegler:
When using the DefaultConverter, type erasure may occur, where the "convert" method returns an object of type S, but treats it as type D. 

The conversion must fail early in such a case, in order to prevent obscure error messages when passing the value to successive methods.

Closes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3008

### DIFF
--- a/bundles/org.eclipse.core.databinding/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.core.databinding/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.databinding
-Bundle-Version: 1.13.500.qualifier
+Bundle-Version: 1.13.600.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.core.databinding;version="1.0.0",
--- a/bundles/org.eclipse.core.databinding/src/org/eclipse/core/databinding/UpdateStrategy.java
+++ b/bundles/org.eclipse.core.databinding/src/org/eclipse/core/databinding/UpdateStrategy.java
@@ -677,6 +677,10 @@ import org.eclipse.core.runtime.Status;
 
 		@Override
 		public Object convert(Object fromObject) {
+			// Explicit cast necessary due to potential type erasure
+			if (toType instanceof Class<?> clazz) {
+				return clazz.cast(fromObject);
+			}
 			return fromObject;
 		}
 
--- a/tests/org.eclipse.jface.tests.databinding/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.jface.tests.databinding/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.tests.databinding
-Bundle-Version: 1.12.600.qualifier
+Bundle-Version: 1.12.700.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.databinding;bundle-version="[1.3.0,2.0.0)",
--- a/tests/org.eclipse.jface.tests.databinding/src/org/eclipse/core/tests/databinding/UpdateStrategyTest.java
+++ b/tests/org.eclipse.jface.tests.databinding/src/org/eclipse/core/tests/databinding/UpdateStrategyTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007 IBM Corporation and others.
+ * Copyright (c) 2007, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -16,11 +16,15 @@ package org.eclipse.core.tests.databinding;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 import org.eclipse.core.databinding.UpdateValueStrategy;
 import org.eclipse.core.databinding.conversion.IConverter;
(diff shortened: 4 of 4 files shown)


ptziegler:
Note that this might break code that still uses raw types, because no explicit cast happens and the type erasure might not be noticed. But that's only because the code is relying on undefined behavior.

vogella:
LGTM

ptziegler:
Test failure is sporadic, unrelated and documented via https://github.com/eclipse-platform/eclipse.platform.ui/issues/3062

sratz:
This caused lots of our tests to fail with

```
java.lang.ClassCastException: Cannot cast java.lang.Boolean to boolean
 at java.base/java.lang.Class.cast(Class.java:4067)
 at org.eclipse.core.databinding.UpdateStrategy$DefaultConverter.convert(UpdateStrategy.java:682)
```

I guess auto-(un)boxing is a special case that is not correctly considered here?

laeubi:
@sratz you might want to add another testcase to show the problem? Are you using `boolean.class` as a parameter somewhere?!?

ptziegler:
I think the ambiguity comes from the fact that `boolean.class` and `Boolean.class` are both of type `Class<Boolean>`. I can have a look later today, but hopefully the check should be as simple as to check for a `Class.isPrimitive()`.

sratz:
> @sratz you might want to add another testcase to show the problem

I'll try.

> Are you using `boolean.class` as a parameter somewhere?!?

Yes, but not me specifically, but

`WidgetProperties.buttonSelection().observe(someCheckbox)` returns a `SWTObservableValueDecorator`, decorating a `org.eclipse.jface.internal.databinding.swt.ButtonSelectionProperty`, extending, `org.eclipse.jface.internal.databinding.swt.WidgetBooleanValueProperty<S>`, returning `Boolean.TYPE` as its value type, which is the primitive type.

sratz:
There is a shortcut:
When one of the `to/fromType`s is not a `Class`:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/69d52f3dee5b292baf2e3f1f9678047efcd2266c/bundles/org.eclipse.core.databinding/src/org/eclipse/core/databinding/UpdateStrategy.java#L127-L128

We take that path whenever EMF is involved, because EMFValueProperty does not return a `Class` in `getValueType()`:
https://github.com/eclipse-emf/org.eclipse.emf/blob/be7a5f1e8782e2801caa0b1c761548e5c8794436/plugins/org.eclipse.emf.databinding/src/org/eclipse/emf/databinding/internal/EMFValueProperty.java#L41-L44
instead, it returns e.g. an instance of `org.eclipse.emf.ecore.impl.EAttributeImpl`

merks:
> it returns e.g. an instance of `org.eclipse.emf.ecore.impl.EAttributeImpl`

Yes the multiplicity affects the type of value involved, e.g., list of the attribute’s type.

ptziegler:
I was wondering why the [IdentityConverter](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.core.databinding/src/org/eclipse/core/internal/databinding/conversion/IdentityConverter.java) wasn't used because the conversion between `Boolean.class`, `boolean.class` and `Boolean.TYPE` still works. But if the `fromType` is a plain object, then the `DefaultConverter` is used as a fallback and the issue appears.

sratz:
> I was wondering why the [IdentityConverter](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.core.databinding/src/org/eclipse/core/internal/databinding/conversion/IdentityConverter.java) wasn't used because the conversion between `Boolean.class`, `boolean.class` and `Boolean.TYPE` still works. But if the `fromType` is a plain object, then the `DefaultConverter` is used as a fallback and the issue appears.

Yes, the alternative to my change would have been to check `fromType` / `toType` independently for being `instanceof Class`, and replace with a boxed alternative, before handing it into `DefaultConverter`, but I think being robust inside `DefaultConverter` is probably the safest.

PR #3018, opened 2025-06-02
Use a Job as the Reconciler Thread

laeubi:
Currently AbstractReconciler uses a raw thread for reconciling requests this has some pitfalls:

1) There is no way for the user to see this job even if it hangs for a long time
2) If the job framework is suspended (e.g during init of the IDE) some code is still executed because the reconciler is called from the creation of the text viewer

This now replaces the bare thread with a job but retain current behaviour / API

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
@@ -15,7 +15,10 @@ package org.eclipse.jface.text.reconciler;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
 
 import org.eclipse.jface.text.DocumentEvent;
 import org.eclipse.jface.text.IDocument;
@@ -53,7 +56,7 @@ abstract public class AbstractReconciler implements IReconciler {
 	/**
 	 * Background thread for the reconciling activity.
 	 */
-	class BackgroundThread extends Thread {
+	class BackgroundThread extends Job {
 
 		/** Has the reconciler been canceled. */
 		private boolean fCanceled= false;
@@ -64,6 +67,10 @@ abstract public class AbstractReconciler implements IReconciler {
 		/** Is a reconciling strategy active. */
 		private boolean fIsActive= false;
 
+		private volatile boolean fIsAlive;
+
+		private boolean started;
+
 		/**
 		 * Creates a new background thread. The thread
 		 * runs with minimal priority.
@@ -72,8 +79,8 @@ abstract public class AbstractReconciler implements IReconciler {
 		 */
 		public BackgroundThread(String name) {
 			super(name);
-			setPriority(Thread.MIN_PRIORITY);
-			setDaemon(true);
+			setPriority(Job.DECORATE);
+			setSystem(true);
 		}
 
 		/**
@@ -98,7 +105,8 @@ abstract public class AbstractReconciler implements IReconciler {
 		/**
 		 * Cancels the background thread.
 		 */
-		public void cancel() {
+		@Override
+		protected void canceling() {
 			fCanceled= true;
 			IProgressMonitor pm= fProgressMonitor;
 			if (pm != null)
@@ -153,7 +161,9 @@ abstract public class AbstractReconciler implements IReconciler {
 					fDirtyRegionQueue.notifyAll();
 				}
 			}
-
+			synchronized (this) {
+				started= false;
+			}
 			informNotFinished();
 			reconcilerReset();
 		}
@@ -167,12 +177,12 @@ abstract public class AbstractReconciler implements IReconciler {
 		 * </p>
 		 */
 		@Override
-		public void run() {
-
+		public IStatus run(IProgressMonitor monitor) {
+			fIsAlive= true;
 			delay();
 
 			if (fCanceled)
-				return;
+				return Status.CANCEL_STATUS;
 
 			initialProcess();
 
@@ -217,6 +227,19 @@ abstract public class AbstractReconciler implements IReconciler {
 
 				fIsActive= false;
 			}
+			fIsAlive= false;
+			return Status.OK_STATUS;
+		}
+
+		public boolean isAlive() {
+			return fIsAlive;
+		}
+
+		public synchronized void start() {
+			if (!started) {
+				started= true;
+				schedule();
+			}
 		}
 	}
 
@@ -233,7 +256,7 @@ abstract public class AbstractReconciler implements IReconciler {
 		public void documentChanged(DocumentEvent e) {
 
 			if (fThread.isActive() || !fThread.isDirty() && fThread.isAlive()) {
-				if (!fIsAllowedToModifyDocument && Thread.currentThread() == fThread)
+				if (!fIsAllowedToModifyDocument && isRunningInReconcilerThread())
 					throw new UnsupportedOperationException("The reconciler thread is not allowed to modify the document"); //$NON-NLS-1$
 				aboutToBeReconciledInternal();
 			}
@@ -449,7 +472,7 @@ abstract public class AbstractReconciler implements IReconciler {
 		synchronized (this) {
 			if (fThread != null)
 				return;
-			fThread= new BackgroundThread(getClass().getName());
+			fThread= new BackgroundThread(getClass().getName() + " reconciler thread"); //$NON-NLS-1$
 		}
 
 		fDirtyRegionQueue= new DirtyRegionQueue();
@@ -614,14 +637,7 @@ abstract public class AbstractReconciler implements IReconciler {
 			return;
 
 		if (!fThread.isAlive()) {
-			try {
-				fThread.start();
-			} catch (IllegalThreadStateException e) {
-				// see https://bugs.eclipse.org/bugs/show_bug.cgi?id=40549
-				// This is the only instance where the thread is started; since
-				// we checked that it is not alive, it must be dead already due
-				// to a run-time exception or error. Exit.
-			}
+			fThread.start();
 		} else {
 			fThread.reset();
 		}
@@ -640,7 +656,10 @@ abstract public class AbstractReconciler implements IReconciler {
 	 * @return <code>true</code> if running in this reconciler's background thread
 	 * @since 3.4
 	 */
-	protected boolean isRunningInReconcilerThread() {
-		return Thread.currentThread() == fThread;
+	protected synchronized boolean isRunningInReconcilerThread() {
+		if (fThread == null) {
+			return false;
+		}
+		return Job.getJobManager().currentJob() == fThread;
 	}
 }


bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -64,6 +67,10 @@ class BackgroundThread extends Thread {
 		/** Is a reconciling strategy active. */
 		private boolean fIsActive= false;
 
+		private volatile Thread myThread;
+
+		private boolean started;
```
iloveeclipse, line 72: Github doesn't allow to add comments on unchanged code, but I believe if you are touching this area, it would make sense to make both `fIsActive` and `fCanceled` as volatile, as they are not accessed from synchronized blocks unlike other fields here.

laeubi:
As this changes a rather central part i'd like to merge this as as soon as master opens so we have enough time for testing/improving.

Currently it is also not optimal that the job/thread is sitting idle all the time. Instead it could just reschedule the job once there is some work. Also some check (e.g. for alive) seem questionable.

Another point is that the code seem to allow a "restart" but a new Background thread is never created.

mickaelistria:
That's a good step forward!

> Currently it is also not optimal that the job/thread is sitting idle all the time. 

Indeed, and going towards job will make it much easier to improve. So this step s definitely a good iterations towards a "cheaper" approach that wouldn't retain threads uselessly.

merks:
Yes this looks promising.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -72,8 +79,8 @@ class BackgroundThread extends Thread {
 		 */
 		public BackgroundThread(String name) {
 			super(name);
-			setPriority(Thread.MIN_PRIORITY);
-			setDaemon(true);
+			setPriority(Job.DECORATE);
+			setSystem(true);
```
iloveeclipse, line 83: The job should provide a family so it can be queries from outside whether the reconciler of specific family is still running or not. Possible would be to have one family for a dedicated viewer for which this job is started, this would require the viewer to be passed as argument here.

See `org.eclipse.core.runtime.jobs.Job.belongsTo(Object)`

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -167,12 +175,12 @@ public void reset() {
 		 * </p>
 		 */
 		@Override
-		public void run() {
-
+		public IStatus run(IProgressMonitor monitor) {
+			myThread= Thread.currentThread();
```
iloveeclipse: This is not safe, since job threads may change.

The only purpose of the `myThread` if I see it right is to check whether the change is from the same thread or not and to check whether the thread is running or not (`AbstractReconciler.BackgroundThread.isAlive()`).  Note, the last one seem to have a bug because the  `myThread` is not reset to null on `cancelling()`.

So if we move to Job API, we should remove `Thread` from this code completely.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -640,7 +654,10 @@ protected void reconcilerReset() {
 	 * @return <code>true</code> if running in this reconciler's background thread
 	 * @since 3.4
 	 */
-	protected boolean isRunningInReconcilerThread() {
-		return Thread.currentThread() == fThread;
+	protected synchronized boolean isRunningInReconcilerThread() {
+		if (fThread == null) {
+			return false;
+		}
+		return Thread.currentThread() == fThread.myThread;
```
iloveeclipse: Better to use `Job.getJobManager().currentJob() == fThread`

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -217,6 +225,19 @@ public void run() {
 
 				fIsActive= false;
 			}
+			myThread= null;
+			return Status.OK_STATUS;
+		}
+
+		public boolean isAlive() {
+			return myThread != null;
```
iloveeclipse: Better to replace `myThread` with a boolean flag that would be set in `run()` at begin and re-set on exiting `run()` (either by cancel or by termination). Note, if I see it right, current code has a bug that doesn't unset the  `myThread` on cancellation.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -449,7 +470,7 @@ public void install(ITextViewer textViewer) {
 		synchronized (this) {
 			if (fThread != null)
 				return;
-			fThread= new BackgroundThread(getClass().getName());
+			fThread= new BackgroundThread(getClass().getName() + " reconciler thread"); //$NON-NLS-1$
```
iloveeclipse, line 475: This shows same value for every editor of same type, ideally it should point to the file being parsed but I don't see right now any way to do so.

iloveeclipse (changes requested):
The change is interesting, but should be better tested & properly reviewed.

Please don't commit just because "we could do it in M1", also nightly the SDK builds should remain in at least some usable state.

Steps to break IDE:
- Open Java editor
- Cancel the new reconciler job
- Do some changes in code
- Watch for two jobs hanging forever: "Requesting Java AST from Selection" and "Decoration Calculations"

Stacks from debugger:

```
java.lang.Exception
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:176)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.SharedASTProviderCore.getAST(SharedASTProviderCore.java:138)
	at org.eclipse.jdt.internal.ui.viewsupport.SelectionListenerWithASTManager$PartListenerGroup.calculateASTandInform(SelectionListenerWithASTManager.java:167)
	at org.eclipse.jdt.internal.ui.viewsupport.SelectionListenerWithASTManager$PartListenerGroup$1.lambda$0(SelectionListenerWithASTManager.java:149)
	at org.eclipse.jdt.internal.core.JavaModelManager.cacheZipFiles(JavaModelManager.java:5708)
	at org.eclipse.jdt.internal.core.JavaModelManager.callReadOnly(JavaModelManager.java:5697)
	at org.eclipse.jdt.core.JavaCore.callReadOnly(JavaCore.java:6156)
	at org.eclipse.jdt.internal.ui.viewsupport.SelectionListenerWithASTManager$PartListenerGroup$1.run(SelectionListenerWithASTManager.java:149)
```

and 

```
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:176)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.SharedASTProviderCore.getAST(SharedASTProviderCore.java:138)
	at org.eclipse.jdt.ui.OverrideIndicatorLabelDecorator.getOverrideIndicators(OverrideIndicatorLabelDecorator.java:164)
	at org.eclipse.jdt.ui.OverrideIndicatorLabelDecorator.lambda$0(OverrideIndicatorLabelDecorator.java:138)
	at org.eclipse.jdt.internal.core.JavaModelManager.cacheZipFiles(JavaModelManager.java:5708)
	at org.eclipse.jdt.internal.core.JavaModelManager.callReadOnly(JavaModelManager.java:5697)
	at org.eclipse.jdt.core.JavaCore.callReadOnly(JavaCore.java:6156)
	at org.eclipse.jdt.ui.OverrideIndicatorLabelDecorator.computeAdornmentFlags(OverrideIndicatorLabelDecorator.java:129)
	at org.eclipse.jdt.ui.OverrideIndicatorLabelDecorator.decorate(OverrideIndicatorLabelDecorator.java:266)
	at org.eclipse.ui.internal.decorators.LightweightDecoratorDefinition.decorate(LightweightDecoratorDefinition.java:254)
	at org.eclipse.ui.internal.decorators.LightweightDecoratorManager$LightweightRunnable.run(LightweightDecoratorManager.java:105)
	at org.eclipse.ui.internal.decorators.LightweightDecoratorManager.decorate(LightweightDecoratorManager.java:359)
	at org.eclipse.ui.internal.decorators.LightweightDecoratorManager.getDecorations(LightweightDecoratorManager.java:345)
	at org.eclipse.ui.internal.decorators.DecorationScheduler$1.queue(DecorationScheduler.java:410)
	at org.eclipse.ui.internal.decorators.DecorationScheduler$1.run(DecorationScheduler.java:388)
```

laeubi:
> Cancel the new reconciler job

As mentioned above, this seems to be the case already (I just was not able to find out how this cancel procedure was meant to be implemented) so I didn't changed anything here.

If you look at the code whenever there is a cancel involved (or Thread ended for whatever reasons), then reset() is called what as far as I can see do simply nothing to respin a new thread or similar.

iloveeclipse:
> so I didn't changed anything here.

By switching to Jobs API you expose the possibility to cancel reconciler jobs now which can be both used by users and by code/tests. So a proper support for cancellation is part of the task.

mickaelistria:
@iloveeclipse Although it's not nice, not honoring the cancellation still seems valid according to the IJob contract: "Asks this job to cancel its execution. The cancellation can take an undertermined amount of time.". There is basically no guarantee of completion/implementation.

iloveeclipse:
> @iloveeclipse Although it's not nice, not honoring the cancellation still seems valid according to the IJob contract: "Asks this job to cancel its execution. The cancellation can take an undertermined amount of time.". There is basically no guarantee of completion/implementation.

Sure, but converting something that works to something that breaks UX/client code is surely not the right direction.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
laeubi: Done

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
laeubi: Done

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
laeubi: Given that before it was only the classname it is fine for now, but yes a more descriptive name would be better.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
laeubi: Done

laeubi (commented):
I applied some suggestions and also reset the started flag on reset now so the job possibly gets restarted.

iloveeclipse:
> and also reset the started flag on reset now so the job possibly gets restarted.

Steps above still produce same result as before.

merks:
Closed in favor of https://github.com/eclipse-platform/eclipse.platform.ui/pull/3025 which is a slight modification that the job continues to run even when there is an attempt to manually cancel it.

PR #3019, opened 2025-06-02
Adjust Icon References for using SVGs

Michael5601:
The bundle `org.eclipse.ui.workbench` contains references to icons in the bundle `org.eclipse.ui`. As these icons are now provided as SVG in `org.eclipse.ui` the references can be adjusted so SVGs are used for loading icons in `org.eclipse.ui.workbench`.


HeikoKlare (approved):
All referenced icons exist.

PR #3021, opened 2025-06-04
Add version bumps for 4.37

HannesWell:
Provoke version increment for 4.37.

Part of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3093


HannesWell:
Actually there are no versions to bump in this repo as we only have a two `pom.xml` files of aggregation folders left 🚀

PR #3025, opened 2025-06-05
Use a Job as the Reconciler Thread for AbstractReconciler

merks:
Currently AbstractReconciler uses a raw thread for reconciling requests. This has the problem that if the job framework is suspended (e.g, during startup of the IDE) some code is still executed because the reconciler is called from the creation of the text viewer and starts immediately.

This replaces the bare thread with a job but retains current behaviour. Note that the job cannot be manually terminated, just as the thread continued to run indefinitely until explicitly terminated.

https://github.com/eclipse-platform/eclipse.platform.ui/pull/3018

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
@@ -15,7 +15,10 @@ package org.eclipse.jface.text.reconciler;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
 
 import org.eclipse.jface.text.DocumentEvent;
 import org.eclipse.jface.text.IDocument;
@@ -53,7 +56,7 @@ abstract public class AbstractReconciler implements IReconciler {
 	/**
 	 * Background thread for the reconciling activity.
 	 */
-	class BackgroundThread extends Thread {
+	private class BackgroundThread extends Job {
 
 		/** Has the reconciler been canceled. */
 		private boolean fCanceled= false;
@@ -64,6 +67,10 @@ abstract public class AbstractReconciler implements IReconciler {
 		/** Is a reconciling strategy active. */
 		private boolean fIsActive= false;
 
+		private volatile boolean fIsAlive;
+
+		private boolean started;
+
 		/**
 		 * Creates a new background thread. The thread
 		 * runs with minimal priority.
@@ -72,8 +79,8 @@ abstract public class AbstractReconciler implements IReconciler {
 		 */
 		public BackgroundThread(String name) {
 			super(name);
-			setPriority(Thread.MIN_PRIORITY);
-			setDaemon(true);
+			setPriority(Job.DECORATE);
+			setSystem(true);
 		}
 
 		/**
@@ -98,7 +105,7 @@ abstract public class AbstractReconciler implements IReconciler {
 		/**
 		 * Cancels the background thread.
 		 */
-		public void cancel() {
+		public void doCancel() {
 			fCanceled= true;
 			IProgressMonitor pm= fProgressMonitor;
 			if (pm != null)
@@ -153,7 +160,9 @@ abstract public class AbstractReconciler implements IReconciler {
 					fDirtyRegionQueue.notifyAll();
 				}
 			}
-
+			synchronized (this) {
+				started= false;
+			}
 			informNotFinished();
 			reconcilerReset();
 		}
@@ -167,12 +176,12 @@ abstract public class AbstractReconciler implements IReconciler {
 		 * </p>
 		 */
 		@Override
-		public void run() {
-
+		public IStatus run(IProgressMonitor monitor) {
+			fIsAlive= true;
 			delay();
 
 			if (fCanceled)
-				return;
+				return Status.CANCEL_STATUS;
 
 			initialProcess();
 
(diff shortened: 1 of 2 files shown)


merks:
@iloveeclipse 

I was able to reproduce the problem you described and with this change the job continues to run which avoids the described problem.  In addition, I see that the editor starts with error markers that are subsequently cleaned up...

laeubi (approved):
Thant for catching up with this, ignoring the cancel request seems like a reasonable approach here, maybe we even want to override `cancel()` of the job and make it a noop with a comment why it is done here that way.

merks:
> Thant for catching up with this, ignoring the cancel request seems like a reasonable approach here, maybe we even want to override `cancel()` of the job and make it a noop with a comment why it is done here that way.

For our safety and comfort, that's not possible.  😁 

https://github.com/eclipse-platform/eclipse.platform/blob/c3dfa7bc07048f1db682284b32a4011290b81c55/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java#L289-L291

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -72,8 +79,8 @@ class BackgroundThread extends Thread {
 		 */
 		public BackgroundThread(String name) {
 			super(name);
-			setPriority(Thread.MIN_PRIORITY);
-			setDaemon(true);
+			setPriority(Job.DECORATE);
+			setSystem(true);
```
iloveeclipse, line 83: The job should provide a family so it can be queried from outside whether the reconciler of specific family is still running or not. Possible would be to have one family for a dedicated viewer for which this job is started, this would require the viewer to be passed as argument here. If that is too much, the minimal requirement would be that they should answer `true` to `job.belongsTo(AbstractReconciler.class)` (without introducing new API here).

See `org.eclipse.core.runtime.jobs.Job.belongsTo(Object)`

Background: in our tests (also in JDT debugger and I guess other places) we often check whether some jobs are still running after some operation, waiting that everything is "calm" before the next test step or on teardown.

Having here a family would allow to ignore jobs of that family without hard coding something like job.getClass().getName().equals ...

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -449,7 +471,7 @@ public void install(ITextViewer textViewer) {
 		synchronized (this) {
 			if (fThread != null)
 				return;
-			fThread= new BackgroundThread(getClass().getName());
+			fThread= new BackgroundThread(getClass().getName() + " reconciler thread"); //$NON-NLS-1$
```
iloveeclipse: I believe we should better undo this change, just keep the class name as before, additional text it doesn't bring much and in JDT case makes the job name definitely not better (beside the fact that this is a "job" now):

image

iloveeclipse (changes requested):
This looks better now, except two places.

I guess we might see some test hangs or slowdown because of that change for tests that open editors and do `waitForJobs()`, we will see...

Ideally if we submit that it would be **after** we have at least one "green" build in SDK.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
merks: Like this I believe, right?

image

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
merks: Sounds good.  Like this perhaps?
```
		@Override
		public boolean belongsTo(Object family) {
			return family == fViewer || AbstractReconciler.class.equals(family);
		}
```

iloveeclipse: Yes, and you can use == instead of equals for class object.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
iloveeclipse: Yes, just like it was before.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
merks: Done.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
merks: Done.

merks:
@iloveeclipse 

All the failures seem related to an unrelated API filter problem:

image

iloveeclipse:
> @iloveeclipse
> 
> All the failures seem related to an unrelated API filter problem:
> 
> image

@raghucssit : I guess it's same issue you mentioned?

merks:
@iloveeclipse 

I'll look at fixing that API issue too, but it's hard to juggle back and forth.  Also Ant current has errors in my workspace and Maven Publishing stopped working.  Too many fires to fight.  😱 

So if you're okay with the current content of this PR, maybe we could clear it out of the way?

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/junk
```diff
@@ -0,0 +1,665 @@
+/*******************************************************************************
```
iloveeclipse: @merks : looks like you've added "junk" file?

merks: Oh dear.  What an idiot!

I will rebase and force push with the API warning thing fixed a little later.

Sorry for the stupid noise.

merks:
Hopefully it builds this time with no failures!

merks:
@iloveeclipse 

We even have a green build now.   😁

iloveeclipse (approved):
Ed, please check SDK test results on the next SDK build. I fear there could be some hangs.

merks:
I will look at the builds tomorrow.

merks:
FYI, I started this build:

https://ci.eclipse.org/releng/job/Builds/job/I-build-4.37/14/

iloveeclipse:
Probably related regression: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2323

PR #3028, opened 2025-06-06
Fix regular expression replacement with anchors causing infinite recursion

### LINKED ISSUE #2820: Regular expression replacement with an anchor improperly recurses on each line
To remove a single leading space from each line of a file I did a regular expression replace-all: "^ " to "". This instead removed all leading spaces. I thought an anchor like "^" wouldn't be recursed unless some special recurse option was selected.

Eclipse IDE for Java Developers
Version: 2024-12 (4.34.0)
Build id: 20241128-0757

braedonbillingsley:
Fixes an issue where using anchors (like ^) in "Replace All" operations would cause infinite loops, removing more text than intended.

When performing a regex "Replace All" with patterns containing anchors like ^ (start of line), the current implementation would recurse infinitely on the same line. For example:

Pattern: ^  (start of line + space)
Replace with: `` (empty string)
Expected: Remove one leading space from each line
Actual: Remove ALL leading spaces from each line

Root Cause: The replaceAll() method uses a loop that continues until no matches are found.

Solution:
Added position tracking to detect when the same position is matched repeatedly:

Track the last found position (lastFoundPosition)
If the same position is found consecutively, advance past it manually
This breaks the infinite loop while preserving normal replacement behavior

Code Changes:
bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
Method: replaceAll()

Changes:
Added lastFoundPosition tracking variable
Added position comparison logic to detect repeated matches
Force advance past problematic positions when detected

Fixes [#2820](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2820)


merks:
Please rebase and squash to a single commit. We cannot incorporate merge commits.

PR #3029, opened 2025-06-07
Use TestUtil.cleanUp across the tests

merks:
- This ensures that the tests don't leak jobs that are only detected by a subsequent test.
- Use editor.close(false) rather than editor.dispose() to clean up the thread/job properly.

https://github.com/eclipse-platform/eclipse.platform.ui/pull/3025

### DIFF
--- a/tests/org.eclipse.ui.editors.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.ui.editors.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.editors.tests;singleton:=true
-Bundle-Version: 3.13.800.qualifier
+Bundle-Version: 3.13.900.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.jface.text.tests.codemining,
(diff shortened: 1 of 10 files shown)


HannesWell:
In some cases it might be more convenient to create for JUnit-4 tests a corresponding TestRule or for JUnit-5 a corresponding [Extension](https://www.baeldung.com/junit-5-extensions#3-lifecycle-callbacks).
The former is for example done in PDE at
https://github.com/eclipse-pde/eclipse.pde/blob/77b9fa8f8d2e8321976465ada201a69b01c5bcfb/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/util/ProjectUtils.java#L308-L312

which is then for example used at:
https://github.com/eclipse-pde/eclipse.pde/blob/77b9fa8f8d2e8321976465ada201a69b01c5bcfb/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/AbstractLaunchTest.java#L56

We also already have custom JUnit-5 extension here, but I don't remember from the top of my head where.

merks:
Yes, this seemed ugly and repetitive.  But I'm not at JUnit expert though even to my eye it seemed there must be a better way.

vogella:
Rule created https://github.com/eclipse-platform/eclipse.platform.ui/pull/3031 here, let me know if you like that, if something thinks that is a good approach, I can modify the tests to use this rule.

merks:
> Rule created #3031 here, let me know if you like that, if something thinks that is a good approach, I can modify the tests to use this rule.

I like it in principle a lot. I would expect that it should simplify something by virtue of deleting something. 

Thanks 🙏 in advance for improving this.

PR #3032, opened 2025-06-10
Provide file contents when determining content type for search #3014

### LINKED ISSUE #3014: File Search considers text files without file extension as binary (regression from #2180 when CDT is installed)

- [X] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/) (tested with I20250527-1800)

## Steps to reproduce

* Do **Help** > **Install new software**
* Paste "https://download.eclipse.org/tools/cdt/releases/12.0/" into the **Work with** field and type **Enter**
* Select **CDT Main Features** > **C/C++ Development Tools** and click **Next**, **Next**, **I accept...**, **Finish**, **Select All**, **Trust Selected**, and **Restart Now**

I tried

* Close the **Welcome** view
* Click **Create a project** under **Package Explorer**
* Select **General** > **Project** and click **Next**
* Type "dummyproject" as **Project name** and click **Finish**
* Do **File** > **New** > **File**
* Type "dummyfile" as **File name** and click **Finish**
* Type "foo" into the editor of "dummyfile" and do **File** > **Save**
* Close the "dummyfile" editor window
* Do **Search** > **File**, type "foo" into **Containing text** and click **Search**

I expected: "dummyfile" as single search result

But got: 0 matches in workspace

* OS & version: Windows 10 19045.5737
* Eclipse IDE/Platform version (as shown in *Help > About*): Version: 2025-06 (4.36), Build id: I20250527-1800

## Community

## Potential cause

I thin this is a regression caused by a combination of #2180 determining binary files by filename only, and eclipse-cdt/cdt#409 introducing a binary ELF File content type matching `[^.]+` file names.

I believe that the culprit is that the fix for #2180 does not consider the describer of binary file types, which in the ELF File case requires that the file starts with an ELF magic number:

https://github.com/eclipse-cdt/cdt/blob/2c82545ba0425ace706af241ee05fab6fb0080bc/core/org.eclipse.cdt.core/plugin.xml#L574-L593

I think `TextSearchVisitor.hasBinaryContentType` has to be extended to require that binary content types also match when considering the file content, maybe by also passing the content to `IContentTypeMatcher.findContentTypesFor` instead of just the file name.

miscose:
Ensures that describers are taken into account when deciding if a content type is binary. Adds a regression test that registers a binary content type with a describer and verifies that it only causes files with content matching the describer to be considered binary.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3014

### DIFF
--- a/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -442,7 +442,12 @@ public class TextSearchVisitor {
 	private final IContentType TEXT_TYPE = Platform.getContentTypeManager().getContentType(IContentTypeManager.CT_TEXT);
 
 	private boolean hasBinaryContentType(IFile file) {
-		IContentType[] contentTypes = Platform.getContentTypeManager().findContentTypesFor(file.getName());
+		IContentType[] contentTypes;
+		try (java.io.InputStream contents = file.getContents()) {
+			contentTypes = Platform.getContentTypeManager().findContentTypesFor(contents, file.getName());
+		} catch (IOException | CoreException e) {
+			return false; // unknown due to failure
+		}
 		for (IContentType contentType : contentTypes) {
 			if (contentType.isKindOf(TEXT_TYPE)) {
 				return false; // is text
--- a/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -446,7 +446,8 @@ public class TextSearchVisitor {
 		try (java.io.InputStream contents = file.getContents()) {
 			contentTypes = Platform.getContentTypeManager().findContentTypesFor(contents, file.getName());
 		} catch (IOException | CoreException e) {
-			return false; // unknown due to failure
+			SearchCorePlugin.log(e);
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -442,7 +442,12 @@ public IStatus search(TextSearchScope scope, IProgressMonitor monitor) {
 	private final IContentType TEXT_TYPE = Platform.getContentTypeManager().getContentType(IContentTypeManager.CT_TEXT);
 
 	private boolean hasBinaryContentType(IFile file) {
-		IContentType[] contentTypes = Platform.getContentTypeManager().findContentTypesFor(file.getName());
+		IContentType[] contentTypes;
+		try (java.io.InputStream contents = file.getContents()) {
+			contentTypes = Platform.getContentTypeManager().findContentTypesFor(contents, file.getName());
+		} catch (IOException | CoreException e) {
+			return false; // unknown due to failure
```
mickaelistria: In such case, wouldn't it be better to log it as a warning buut still resolve the contentTypes just based on the file name? That would usually lead to good enough results, wouldn't it?

miscose: Yes, that makes sense.

mickaelistria:
Thanks, I've created https://github.com/eclipse-platform/eclipse.platform.ui/pull/3034 which is a reorg of these commits. I plan to merge #3034 ASAP (when build succeeds).

mickaelistria:
Fixed with #3024

PR #3038, opened 2025-06-12
Mark ObservablesManager#runAndCollect for removal

vogella:
Deprecated since a very long time, time to mark for removal.

### DIFF
--- a/bundles/org.eclipse.core.databinding/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.core.databinding/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.databinding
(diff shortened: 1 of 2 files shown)


laeubi:
@vogella have you checked this is not used in platform anymore? I think last time I checked it was still used as there are some cases where you can not use with other methods.

vogella:
AFAICS only two calls in ModelEditor. Definitely something we need to update within the next two years before deleting the API.

laeubi:
I think we should not mark API for removal that we actively use anywhere (or mark that for removal as well).

vogella:
> I think we should not mark API for removal that we actively use anywhere (or mark that for removal as well).

As we have years to react I think marking for deletion is fine, even if we take longer we still do not have to delete.

laeubi:
As you initially wrote

> Deprecated since a very long time

and no one takes action just because we mark it for removal, I doubt it will be done now (especially as it is inside platform itself).

vogella:
Ok, abandon this change.

PR #3045, opened 2025-06-16
Replace deprecated key classes to use jface bindings API

elsazac:
The classes `KeySequence` and `KeyStroke` from `org.eclipse.ui.keys` are deprecated. They have been replaced to use the corresponding classes from `org.eclipse.jface.bindings.keys` .


mickaelistria:
Thank you!

PR #3049, opened 2025-06-18
Updated bundles with comparator errors

trancexpress:
Changes to ecj are to remove redundant bytecode, see: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3465

See: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3137


trancexpress:
The fail here is: https://github.com/eclipse-platform/eclipse.platform.ui/issues/195

PR #3056, opened 2025-06-20
[#2909] fix wrong hover size in composite hovers

### LINKED ISSUE #2909: Maven "pom.xml": popup for validation warning has wrong size
I created a maven project from the archetype "org.wildfly.archetype:wildfly-jakartaee-webapp-archetype:35.0.1.Final", then I added three dependencies to pom.xml:

```xml
<dependency>
	<groupId>jakarta.servlet.jsp</groupId>
	<artifactId>jakarta.servlet.jsp-api</artifactId>
	<scope>provided</scope>
</dependency>

<dependency>
	<groupId>org.glassfish.web</groupId>
	<artifactId>jakarta.servlet.jsp.jstl</artifactId>
	<version>3.0.1</version>
	<scope>provided</scope>
</dependency>

<dependency>
	<groupId>jakarta.servlet.jsp.jstl</groupId>
	<artifactId>jakarta.servlet.jsp.jstl-api</artifactId>
	<version>3.0.1</version>
	<scope>provided</scope>
</dependency>
```

One dependency has a unnecessary "version" element. So a warning "Overriding managed version 3.0.2 for jakarta.servlet.jsp.jstl-api" is reported.
The popuop when hovering over the warning text has a wrong size.

When hovering over "version", the message is completely visible, but when resizing the popup, additional info about the "version" element becomes visible. Is this intended?
Image

When hovering over the text content of "version", the message is clipped. A second warning is shown when resizing it:

Image

Hope this the right place to report it. Some time ago I reported a similar "popup is too small" problem for the xml editor (https://www.github.com/eclipse-lsp4e/lsp4e/issues/214) and was told to report it here. But as I could not reproduce the issue, I did not do this ;-). Anyway, in the other issue it was suggested that it might be related to multiple warnings in one place, and the second warning could be such a case, as there are two messages.

ghentschke:
These two changes require each other:

* Do not add sizes in computeCompositeSize as only the max. should be used in x-direction

* Prevent shrinking of the hover shell when the size calculation fails

fixes #2909

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultInformationControl.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultInformationControl.java
@@ -368,6 +368,11 @@ public class DefaultInformationControl extends AbstractInformationControl implem
 
 	@Override
 	public Point computeSizeHint() {
+		if (getShell().getChildren().length == 0) {
+			// no content yet, return default size
+			// computeSize would return 2,2 here
+			return getShell().getSize();
+		}
 		// see https://bugs.eclipse.org/bugs/show_bug.cgi?id=117602
 		int widthHint= SWT.DEFAULT;
 		Point constraints= getSizeConstraints();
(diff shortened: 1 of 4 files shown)


ghentschke:
This change will lead to this composite hover to be displayed (combination of: `MarkerInformationControl`, `FocusableBrowserInformationControl` and `DefaultInformationControl`):

image

mickaelistria (approved):

mickaelistria:
Thanks a lot!

PR #3063, opened 2025-06-24
Try to stabilize flacky PartRenderingEngineTest

### LINKED ISSUE #751: PartRenderingEngineTests test fails randomly

ptziegler:
Due to an oversight, the REMOVE_ON_HIDE tag is set on "partB" twice, as opposed to "partB" and "partC". This may occasionally keep the part stack rendered, because "partC" is not removed from the model when hidden.

Closes https://github.com/eclipse-platform/eclipse.platform.ui/issues/751

### DIFF
--- a/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
+++ b/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2023 IBM Corporation and others.
+ * Copyright (c) 2009, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -2445,7 +2445,7 @@ public class PartRenderingEngineTests {
 		partStackForPartBPartC.setSelectedElement(partB);
 
 		MPart partC = ems.createModelElement(MPart.class);
-		partB.getTags().add(EPartService.REMOVE_ON_HIDE_TAG);
+		partC.getTags().add(EPartService.REMOVE_ON_HIDE_TAG);
 		partStackForPartBPartC.getChildren().add(partC);
 		partStackForPartBPartC.setSelectedElement(partC);
 
--- a/tests/org.eclipse.e4.ui.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.e4.ui.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.e4.ui.tests;singleton:=true
-Bundle-Version: 0.15.800.qualifier
+Bundle-Version: 0.15.900.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.emf.ecore.xmi;bundle-version="2.4.0",


ptziegler:
This issue can hardly be reproduced locally, so I'll keep it as a draft for now and will periodically rebase it, to see whether the failure still occurs. The general problem is clear, I think, and the question is only how much one has to wait between the test steps.

tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
```diff
@@ -2468,6 +2468,7 @@ public void ensureCleanUpAddonCleansUp() {
 
 		assertTrue(" PartStack with children should be rendered", partStackForPartBPartC.isToBeRendered());
 		partService.hidePart(partB);
+		contextRule.spinEventLoop();
 		partService.hidePart(partC);
 		contextRule.spinEventLoop();
 		assertFalse(
```
laeubi: Instead one can consider to have a loop here that for example checks for a maximum of 10 seconds if the condition is meet and if not spins the event loop. I have did this on other places in a similar way.

ptziegler: I first wanted to see if this is sufficient, but I was also already playing with the idea of creating a new `spinEventLoop(int method)`, if this is still too unreliable:

```java
public void spinEventLoop(int time) {
    long start = System.currentTimeMillis();
    do {
        while (Display.getCurrent().readAndDispatch()) {
            Thread.yield();
        }
    } while (System.currentTimeMillis() - start < time);
}
```

ptziegler: False alarm: It's not a timing issue, the test is simply bugged. "partC" is missing the `REMOVE_ON_HIDE_TAG`, so it's never removed from the part stack.

ptziegler:
For the sake of documentation:

- The failure can be reproduced by simply running the test in a loop. After ~80-100 cycles, the problem should occur. Cause of the failure is the CleanupAddon not being executed.
- The test was added with a64d6c9d352aece17c808ea37385e395bbbda129 by @vogella and I'm fairly certain that this line was simply a copy-paste error.

vogella:
> For the sake of documentation:
> 
> * The failure can be reproduced by simply running the test in a loop. After ~80-100 cycles, the problem should occur. Cause of the failure is the CleanupAddon not being executed.
> * The test was added with [a64d6c9](https://github.com/eclipse-platform/eclipse.platform.ui/commit/a64d6c9d352aece17c808ea37385e395bbbda129) by @vogella and I'm fairly certain that this line was simply a copy-paste error.

Using partC looks right.

ptziegler:
Test failure is unrelated: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1517

PR #3071, opened 2025-06-27
Consider primitive types in explicit type check in DefaultConverter

sratz:
Follow-up on #3009.


ptziegler (approved):

ptziegler:
Thanks for already looking into the edge case this fast! :rocket:

PR #3072, opened 2025-06-27
Fix suppressWarnings

elsazac:
Remove 'deprecation' from @SuppressWarnings as it triggers an extra warning.

### DIFF
--- a/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
+++ b/bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
@@ -27,7 +27,7 @@ import java.util.stream.StreamSupport;
  *             {@code org.eclipse.e4.ui.workbench.modeling.EModelService#findMatchingElements(org.eclipse.e4.ui.model.application.MApplicationElement, String, Class)}
  *             instead.
  */
-@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
+//@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
 public interface XPathContext {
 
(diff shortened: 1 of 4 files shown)


akurtakov:
The PR adds 6 new warnings. It doesn't look like this is the correct change.

elsazac:
> The PR adds 6 new warnings. It doesn't look like this is the correct change.

Yeah, I am checking this.

bundles/org.eclipse.e4.emf.xpath/src/org/eclipse/e4/emf/xpath/XPathContext.java
```diff
@@ -27,7 +27,7 @@
  *             {@code org.eclipse.e4.ui.workbench.modeling.EModelService#findMatchingElements(org.eclipse.e4.ui.model.application.MApplicationElement, String, Class)}
  *             instead.
  */
-@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
+//@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
```
akurtakov, line 30: This is totally wrong thing to do as it has been deprecated for a very good reason.

elsazac: @akurtakov Sorry for pushing this,I was only testing the CI errors, and to see why the compiler throws a new warning during the build.;  it was just for investigation.

akurtakov: If you don't plan to finish this one I would rather see it closed as there are way too many open PRs that make it hard for anyone that tries to do any triaging.

akurtakov:
Closing as there is no progress for months.

PR #3088, opened 2025-07-05
Extract Reconciler Job Into an own class

laeubi:
Currently the `BackgroundThread` is an inner class of the `AbstractReconciler` and is tightly coupled with it what makes it quite hard to understand the interaction between both and to properly update a reconciler on document changes or refactor the code.

This is a first attempt to decouple the things by extracting the inner class in an own (package private) type to allow further refactoring operations and proper encapsulation.

Whats next:

Currently the change of the viewer initiates an exchange of the reconcile queue, this might or might not overlap with the background thread and the Job itself is not changed making it also not possible to identify the actual document. Instead the viewer and the queue should be passed in the constructor, the old job should be canceled and a new one must be scheduled to make this properly works. With that it would even be possible to find out the document providing input and set the name accordingly making the whole access to the shared data more threadsafe and easier to understand.

Then at the moment the job uses delays to wait for more things to do and sleeping otherwise unless notified. This notification for new things could be replaced by a scheduling of the job instead making much less threads running when many documents are open. Currently for each open Editor one thread is running forever unless the user closes the document.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
@@ -15,9 +15,7 @@ package org.eclipse.jface.text.reconciler;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
-import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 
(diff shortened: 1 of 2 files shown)


merks (approved):
It all seems fine in principle, though I imagine the same could be accomplished by leaving it where it was and simply making the class static.

laeubi:
Merging as no further concerns are raised.

iloveeclipse:
I had no time to review because I've spent time on original problem.
Please check https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2323#issuecomment-3045396964.

iloveeclipse:
@laeubi : also if you request reviews, ideally you would give more then one working day time. I'm trying to have some private life on weekends, so you can't expect review to happen on same working day.

PR #3089, opened 2025-07-07
FormText.paint(): don't pass (0,0) to Image on Linux

### LINKED ISSUE #3087: Cheat Sheet view reports multiple errors on use
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/3078 / e266ec17f85789de6e2245b6b628317acf5f5f64.

Open Chat Sheet view and select any topic from the "..." menu.
There will be multiple errors reported in the log
```
eclipse.buildId=4.37.0.I20250703-1800
java.version=21.0.6-7
java.vendor=N/A
Command-line arguments:  -os linux -ws gtk -arch x86_64 -data /data/4x_platform_workspace

org.eclipse.ui
Error
Fri Jul 04 14:52:24 CEST 2025
Unhandled event loop exception

java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4927)
	at org.eclipse.swt.SWT.error(SWT.java:4832)
	at org.eclipse.swt.graphics.Image.init(Image.java:1266)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:219)
	at org.eclipse.swt.graphics.Image.drawWithImageGcDrawer(Image.java:1179)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:705)
	at org.eclipse.ui.forms.widgets.FormText.repaint(FormText.java:1581)
	at org.eclipse.ui.forms.widgets.FormText.paint(FormText.java:1558)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:268)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5862)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1656)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1682)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1665)
	at org.eclipse.swt.widgets.Control.gtk_draw(Control.java:3895)
	at org.eclipse.swt.widgets.Scrollable.gtk_draw(Scrollable.java:394)
	at org.eclipse.swt.widgets.Composite.gtk_draw(Composite.java:506)
	at org.eclipse.swt.widgets.Canvas.gtk_draw(Canvas.java:173)
	at org.eclipse.swt.widgets.Widget.windowProc(Widget.java:2614)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:6861)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:6169)
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_do_event(Native Method)
	at org.eclipse.swt.widgets.Display.eventProc(Display.java:1605)
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_iteration_do(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4519)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:678)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:185)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:219)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:627)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:575)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1431)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1403)
```

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3087

### DIFF
--- a/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
+++ b/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
@@ -1555,7 +1555,14 @@ public class FormText extends Canvas {
 		gc.setForeground(getForeground());
 		gc.setBackground(getBackground());
 		Point size = getSize();
-		repaint(gc, 0, 0, size.x, size.y);
+		Rectangle paintBounds;
+		if (size.x == 0 && size.y == 0) {
+			// avoids crash on image creation with (0,0) image size
+			paintBounds = new Rectangle(e.x, e.y, e.width, e.height);
+		} else {
+			paintBounds = new Rectangle(0, 0, size.x, size.y);
+		}
+		repaint(gc, paintBounds.x, paintBounds.y, paintBounds.width, paintBounds.height);
 	}
 
 	private void repaint(GC gc, int x, int y, int width, int height) {


bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
```diff
@@ -1555,6 +1556,10 @@ private void paint(PaintEvent e) {
 		gc.setForeground(getForeground());
 		gc.setBackground(getBackground());
 		Point size = getSize();
+		if (size.x == 0 && size.y == 0 && Util.isLinux()) {
+			// On Linux, the size may be 0 and cause errors on Image creation.
+			size = new Point(1, 1);
+		}
```
fedejeanne:
```suggestion
		if (size.x == 0 && size.y == 0) {
			return;
		}
```

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
```diff
@@ -1555,6 +1556,10 @@ private void paint(PaintEvent e) {
 		gc.setForeground(getForeground());
 		gc.setBackground(getBackground());
 		Point size = getSize();
+		if (size.x == 0 && size.y == 0 && Util.isLinux()) {
+			// On Linux, the size may be 0 and cause errors on Image creation.
+			size = new Point(1, 1);
```
amartya4256: I think here we could just say `new Point (e.width, e.height)`.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
fedejeanne: I mean it makes more sense to not repaint at all (in any OS).

WDYT?

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
amartya4256: I think we don't need a Util.isLinix check. Since we never want to pass a zero as the dimension since that breaks the Image implementation.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
```diff
@@ -1555,6 +1556,10 @@ private void paint(PaintEvent e) {
 		gc.setForeground(getForeground());
 		gc.setBackground(getBackground());
 		Point size = getSize();
+		if (size.x == 0 && size.y == 0 && Util.isLinux()) {
```
amartya4256: I think we don't need a Util.isLinux check. Since we never want to pass a zero as the dimension on any platform since that breaks the Image implementation.

fedejeanne: Agree. See [my comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3089/files#r2189397914)

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
fedejeanne: I think that would be (another) arbitrary value which could (and probably would) end up being `{0, 0}` again? Just a guess, my WSL went nuts so I can't test it right now.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
iloveeclipse: Can you test it on Windows? On Linux it seem to work.

iloveeclipse:
So should we
1) return
2) use `new Point (e.width, e.height)`
3) Anything else?

I'm fine with return, at least on Linux it seem to work. However I don't have Mac and I would not like to test on Windows (need more time to setup environment).

So if @fedejeanne or @amartya4256 could test on other OS we can do what fits best.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
amartya4256: It won't. As it was the previous implementation. In the event, you always get a row of pixel to paint and in Linux in this case you get a (1,1). And since my commit broke it. The previous state was to repaint the bounds received from the event.

Then in that case, my bad, following could be even better:
`repaint(gc, e.x, e.y, e.width, e.height);` and then we do t have to hardcode it.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
fedejeanne: I see no problems on Windows, but to be honest: this issue (#3087) is not reproducible on Windows because the size (point) is **not** 0x0 so this change has no effect.

image

I see no other effects on Windows with my change (_i.e._ skipping the call to repaint)

fedejeanne: Sorry, I tested my version.

Yours ...

image

works too. Again, it produces no changes in Windows.

bundles/org.eclipse.ui.forms/src/org/eclipse/ui/forms/widgets/FormText.java
fedejeanne: I tested on Linux (at home) and using `repaint(gc, e.x, e.y, e.width, e.height);` repaints a square of 1x1, so it ends up being the same that Andrey proposed. 

Do you think it's really necessary to repaint at all if the control has no area to paint on? I think not, which is why I maintain that we should simply skip the call to paint instead of working around it and looking for a way to paint something.

Did I miss anything?

fedejeanne: FTR the 1x1 sounds to me like some other part of the code is doing a workaround and painting a pixel instead of doing nothing.

fedejeanne:
I tested the current PR (calling `return` if the size is `0x0`) and I saw no side effects on any OS (I tested for Mac and Linux at home). I propose we move on with this PR.

We can continue [this discussion](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3089#discussion_r2189431943) too if you want, but I'd like to have this fixed fast since I'm out of office the next 3 days and this PR fixes the regression.

Would you agree, @iloveeclipse and @amartya4256 ?

amartya4256:
I tested it out on both windows and linux and I do see no side effect of returning on if the size is 0. However, to avoid any unprecedented issue, I tested out with the following code and it seems the safest to me:

```
Point size = getSize();
Rectangle paintBounds = (size.x == 0 && size.y == 0) ? new Rectangle(e.x, e.y, e.width, e.height)
      : new Rectangle(0, 0, size.x, size.y);
repaint(gc, paintBounds.x, paintBounds.y, paintBounds.width, paintBounds.height);
```

iloveeclipse:
Interestingly, I've noticed while debugging the "bad" size of the FormText is only returned on fullscreen shells. If I minimize the window, no "zero" size is returned. Bug in GTK's implementation of `FormText.getSize()` or GTK3 bug?

sratz:
> Interestingly, I've noticed while debugging the "bad" size of the FormText is only returned on fullscreen shells. If I minimize the window, no "zero" size is returned. Bug in GTK's implementation of `FormText.getSize()` or GTK3 bug?

Hmm, we have about ~50 tests failing in our product due to this issue, but as far as I can tell none of them are using any fullscreen shells. Not really sure where exactly the (0,0) is coming from.

fedejeanne:
> Not really sure where exactly the (0,0) is coming from.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/3087#issuecomment-3043753197

fedejeanne:
I still don't understand why is it better to try and paint something when the size is `0x0` but since the rest seems confident that this is the best approach, I see no point in opposing it.

I tested the current PR in Mac and I didn't see any side-effects. Feel free to merge as I'll be offline for most of the next 3 days.

Thank you all for the input and thank you @iloveeclipse for the prompt fix!

iloveeclipse:
> I still don't understand why is it better to try and paint something when the size is `0x0`

it is just defensive style - use whatever was working before to stay compatible.

iloveeclipse:
OK, I think we have a consensus. Let's merge.

PR #3093, opened 2025-07-08
Perform initial work in the job

laeubi:
Currently the first thing a reconciler after startup does is to sleep for a while then call initialProcess then start the processing loop (that sleeps again...).

This now moves this initial work into the start job so we have a plain processing loop in the thread to make further improvements easier.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
@@ -55,29 +55,27 @@ abstract public class AbstractReconciler implements IReconciler {
 	/**
 	 * Background thread for the reconciling activity.
 	 */
-	class BackgroundThread extends Thread {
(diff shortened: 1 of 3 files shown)


iloveeclipse:
I'm not sure if that change would not break the implementation of existing clients, the code is very old to there is ahigh probability someone extended it.

BackgroundThread.run() javadoc says that it will call initialProcess() and that was **always** done in the BackgroundThread. Now with this change the code will run in a different thread.

laeubi:
> BackgroundThread.run() javadoc says that it will call initialProcess() and that was always done in the BackgroundThread. Now with this change the code will run in a different thread.

I think that is more it run _in_ a background thread than in _the_ `BackgroundThread` ... to prevent confusion I renamed that class now and don't let it implement Thread but instead creating one. I also adjuste the code so anyone that needs to know that it is inside the _reconcile background thread_ can still query it using `AbstractReconciler.isRunningInReconcilerThread()` that is the only place an implementation can actually hook in / extend anything.

PR #3100, opened 2025-07-10
Test.css.swt java21

vogella:


laeubi:
Should text-blocks not being a Java 17 feature?

PR #3101, opened 2025-07-10
Update test.css.core tests to Java 21

vogella:


laeubi:
@vogella please don't simply bump Java versions without any code changes, the BREE should only be incremented if it actually uses any new features as it often results in a ripple effect for consumers (not the case for tests but I think we should not do it there either "just in case".

Or is this connected to some particular issue to solve? Then you should mention it in the commit message.

vogella:
Yes I plan to use Java 21 features in the tests.

laeubi:
> Yes I plan to use Java 21 features in the tests.

The please add these changes to the PR to make it a more consistent change and one can see the benefits here.

PR #3115, opened 2025-07-17
Restore correct URL/URI handling in URLImageDescriptor

ptziegler:
The changes done with cee46311e84ba41e8a421b47099ad709e23761f1 cause an exception when URL is passed as an argument which contains improperly escaped characters (most noticeably whitespaces) and therefore causes an URISyntaxException when calling URL.toURI().

This does not occur when using IPath.fromOSString(), which is the approach that was done previously. But instead of converting this path back to an OS-dependent string, it is instead converted to a file, to check whether it exists and to only then convert it back to its String representation.

A test case with an ill-formed URL has been added to avoid a similar problem in the future.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
@@ -19,19 +19,18 @@ package org.eclipse.jface.resource;
 
 
 import java.io.BufferedInputStream;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
-import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.IAdaptable;
+import org.eclipse.core.runtime.IPath;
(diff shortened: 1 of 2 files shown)


ptziegler:
Test failure is unrelated again: https://github.com/eclipse-platform/eclipse.platform.ui/issues/926

Unless there are any objections, I plan on merging this later today when I'm back home.

HannesWell:
In general URL/URI handling and conversion to Path can be quite tricky and has a lot of corner-cases. I started to create a Utility that covers all these cases, but didn't have the time yet to finish it:
- https://github.com/eclipse-equinox/equinox/pull/691

Once it is finished, it would probably be good to use it here as well.

ptziegler:
> In general URL/URI handling and conversion to Path can be quite tricky and has a lot of corner-cases. I started to create a Utility that covers all these cases, but didn't have the time yet to finish it:
> 
>     * [Add manual trigger for 'updateRelease'-workflow and use project-bot eclipse-equinox/equinox#702](https://github.com/eclipse-equinox/equinox/pull/702)
> 
> 
> Once it is finished, it would probably be good to use it here as well.

I assume you got the numbers mixed up and meant https://github.com/eclipse-equinox/equinox/pull/691?

HannesWell:
> I assume you got the numbers mixed up and meant [eclipse-equinox/equinox#691](https://github.com/eclipse-equinox/equinox/pull/691)?

Yes, indeed. Thanks for the correction.
Please, note there is also https://github.com/eclipse-equinox/equinox/pull/920. but I intend to combine both into one which simplifies development.

HeikoKlare:
Thank you for the quick fix, @ptziegler!

PR #3117, opened 2025-07-18
Remove WS_CARBON check

elsazac:
This PR removes the only remaining WS_CARBON check from the codebase.The Carbon backend is no longer used or supported in SWT, as all modern macOS environments use the Cocoa backend (WS_COCOA). This check is now obsolete and can be removed. (https://github.com/eclipse-platform/eclipse.platform.ui/discussions/2964)

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/keys/BindingPersistence.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/keys/BindingPersistence.java
@@ -519,7 +519,7 @@ public class BindingPersistence extends PreferencePersistence {
 	 * @param commandService            The command service for the workbench; must
 	 *                                  not be <code>null</code>.
 	 */
-	@SuppressWarnings("removal")
+
 	private static final void readBindingsFromRegistry(final IConfigurationElement[] configurationElements,
 			final int configurationElementCount, final BindingManager bindingManager,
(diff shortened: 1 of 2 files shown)


vogella:
Thanks @elsazac

mickaelistria:
This is causing #3141 and should probably be reverted (one may add a line to log warning when using 'carbon' instead of 'cocoa') as it seems clear that the ecosystem, and even Platform, is not ready for dropping this legacy definition.

basilevs:
> even Platform

https://github.com/eclipse-platform/eclipse.platform.ui/blob/5922d0c3170d93c251cc48681a926eea03df8fb3/bundles/org.eclipse.ui/plugin.xml#L236-L240

I was hit by this problem when working in PDE and confirmed it also breaks navigation across simple text editors, so Platform was very much broken.

At the very least, Platform's plugin.xml files should be updated when this change is applied.

mickaelistria:
@basilevs if you have the opportunity to submit PRs replacing those `carbon` by `cocoa` in platform, that would be very welcome!

elsazac:
>if you have the opportunity to submit PRs replacing those `carbon` by `cocoa` in platform, that would be very welcome!

@mickaelistria I would like to work on this again, but I wanted to understand , when we replace carbon with cocoa and remove the code containing WS_CARBON, that change will now honour the bindings used in all the plugins across Eclipse. But, would this change affect any third-party plugins that might still rely on carbon?

PR #3127, opened 2025-07-26
Stop building source features

akurtakov:
They are not deployed and deprecated in Tycho.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,59 +60,6 @@
         </repository>
       </repositories>
     </profile>
-    <profile>
-      <id>build-feature-source</id>
-      <activation>
-        <!-- Only activated for feature projects-->
-        <file>
-          <exists>feature.xml</exists>
-        </file>
-      </activation>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.eclipse.tycho</groupId>
-            <artifactId>tycho-source-plugin</artifactId>
-            <executions>
-              <execution>
-                <phase>package</phase>
-                <id>feature-source</id>
-                <goals>
-                  <goal>feature-source</goal>
-                </goals>
-                <configuration>
-                  <excludes>
-                    <plugin id="javax.xml" />
-                    <plugin id="org.eclipse.swt" />
-                    <plugin id="org.eclipse.equinox.launcher.cocoa.macosx.x86_64" />
-                    <plugin id="org.eclipse.equinox.launcher.cocoa.macosx.aarch64" />
-                    <plugin id="org.eclipse.equinox.launcher.gtk.linux.ppc64le" />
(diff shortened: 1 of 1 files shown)


merks:
What about these?

image

Is there a risk that we might just pick old ones from the baseline and not noticed that these aren't being built?

It's a little hard to find references to source features in features.  

Do we still build the equinox-sdk product for something?

akurtakov:
They will still build as the config is all through the codebase. 
Equinox: https://github.com/eclipse-equinox/equinox/blob/5c20643d899ff73bbd45baac0d0679e31bbf4c27/pom.xml#L246
P2: https://github.com/eclipse-equinox/p2/blob/c68f680530d11d6a31419e88653d083a2b28ee76/pom.xml#L110

With this PR I'm simply cleaning the build to not produce artifacts that are no longer published. We will see what will be on the line for removal next time there is a slot for this work.

akurtakov:
Failing MacOS test is unrelated. Merging.

ruspl-afed:
Where can I read more about this new approach and what is required from other projects @akurtakov @merks ?

merks:
https://www.eclipse.org/lists/simrel-dev/msg00072.html

> Also Tycho is tentatively planning to remove support producing source features:
> 
> https://github.com/eclipse-tycho/tycho/pull/5137
> 
> So you should make long term plans to accommodate these changes, or provide a compelling reason on the above issue why you cannot live without that Tycho support.
> 
> In general, it's already possible to specify that you want a repository to automatically include corresponding source bundles for all binary bundles. Similarly, for a product you can specify that you want to install all corresponding source bundles (which is probably a bad design but your choice). Hence there is seemingly no compelling need for source features themselves.

---

I think no one is paying attention so they will find that by the time they have a compelling reason for why they need source features the support will already removed.  

The assertion/assumption is that source features are generally (quite/largely) useless because:

- You can specify in Tycho to build a product that includes all available sources (which I did in several EPP package).
- You can specify in Tycho to build a p2 repository that includes all available source of the things otherwise includes in the repositories (which the Platform has been doing for quite some time).

So for what reason does one need features that include bundles sources and in particular also source features themselves?  

- So that someone can install it in an IDE to have selectively the sources for everything in that feature, and then the user of IDE can use the IDE as their target platform!

But that's really not not so compelling.  Or?

ruspl-afed:
Thank you for so comprehensive answer @merks and sorry for missing the corresponding notice in `simrel-dev`

For me it is good news to stop maintaining source features, because it will reduce the support effort.

laeubi:
@ruspl-afed @merks the mojo is now marked as deprecated in Tycho 5 and mentions in the list for removals here:

https://github.com/eclipse-tycho/tycho/blob/tycho-5.0.x/REMOVAL.MD

if you Think we should adjust or enhance the note please fell free to propose a PR.

PR #3132, opened 2025-07-28
Using SubMonitor in WorkbenchActivitySupport and simplify logging

vogella:
WorkbenchActivitySupport uses now SubMonitor and only one message for progress.


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/activities/ws/WorkbenchActivitySupport.java
```diff
@@ -130,16 +131,15 @@ public void run(IProgressMonitor monitor) {
 									openTime = System.currentTimeMillis()
 											+ workbench.getProgressService().getLongOperationTime();
 
+									SubMonitor subMonitor = SubMonitor.convert(monitor, ActivityMessages.ManagerTask,
+											2);
 									// two work units - updating the window bars, and updating view bars
-									monitor.beginTask(ActivityMessages.ManagerTask, 2);
-
-									monitor.subTask(ActivityMessages.ManagerWindowSubTask);
+									subMonitor.beginTask(ActivityMessages.ManagerTask, 2);
```
laeubi: Seems obsolete as you already have init the monitor with that values.

laeubi:
@vogella can you explain the background/advantage for this change?

I can't really see any benefit from using a submonitor here, e.g. no `split()` calls ... instead now we have less information as subtask information is discarded.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/activities/ws/WorkbenchActivitySupport.java
ptziegler: Calling `beginTask(...)` here is not only obsolete, but potentially harmful. If the monitor happens to be a `SubMonitor` already, this method is called twice, which is not allowed.

Once here and once in `SubMonitor.convert(...)`:
```java
public static SubMonitor convert(IProgressMonitor monitor, String taskName, int work) {
    if (monitor == null) {
        monitor = new NullProgressMonitor();
        return new SubMonitor(new RootInfo(monitor), 0, work, SUPPRESS_ALL_LABELS);
    }

    // Optimization: if the given monitor already a SubMonitor, no conversion is
    // necessary
    if (monitor instanceof SubMonitor subMonitor) {
        subMonitor.beginTaskImpl(taskName, work); <-----
        return subMonitor;
    }

    monitor.beginTask(taskName, MINIMUM_RESOLUTION);
    return new SubMonitor(new RootInfo(monitor), MINIMUM_RESOLUTION, work, SUPPRESS_NONE);
}
```

I don't think this happens right now, but if the implementation of the `ProgressMonitorDialog` ever changes, this might then cause problems.

PR #3136, opened 2025-07-30
Add API filters for removal of @noreference SWT classes

### LINKED ISSUE #366: Remove MonitorAwarePoint / MonitorAwareRectangle
The `MonitorAwarePoint` and `MonitorAwareRectangle` classes have been replaced with the `Rectangle.WithMonitor` and `Point.WithMonitor` classes. There does not seem to be a need to keep those classes, but they may just be removed.
However, the PR introducing the replacement classes marked them as deprecated:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2269

While the original PR removed them:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2200

The classes shall be removed.

HeikoKlare:
The classes MonitorAwarePoint and MonitorAwareRectangle have been removed from SWT. They were public inside an API package but marked as `@noreference` as they are not intended to be used as part of an API, which is why API tooling considers there removal a breaking API change. The bundles org.eclipse.ui and org.eclipse.jface reexport SWT packages, which is why API tooling also considers the removal as breaking API change of those bundles. This change adds according API filters.

Belongs to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2349

### DIFF
--- a/bundles/org.eclipse.jface/.settings/.api_filters
+++ b/bundles/org.eclipse.jface/.settings/.api_filters
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<component id="org.eclipse.jface" version="2">
+    <resource path="META-INF/MANIFEST.MF" type="org.eclipse.swt.graphics.MonitorAwarePoint">
+        <filter id="305426566">
+            <message_arguments>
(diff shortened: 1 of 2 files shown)


fedejeanne:
The failing test is unrelated (#751) and the failing checks about unnecessary filters are expected since the filters only become necessary after merging https://github.com/eclipse-platform/eclipse.platform.swt/pull/2349

Issue #3042, opened 2025-06-13
[Test] org.eclipse.text.quicksearch.tests.QuickAccessComputerTest.testQuickAccessComputer fails randomly


fedejeanne:
Failed randomly on **Mac** for #3004 : https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/15628923632/job/44028715479?pr=3004#step:5:14405

```
org.eclipse.text.quicksearch.tests.QuickAccessComputerTest.testQuickAccessComputer -- Time elapsed: 0.399 s <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at org.eclipse.text.quicksearch.tests.QuickAccessComputerTest.testQuickAccessComputer(QuickAccessComputerTest.java:60)
```

PR #3147, opened 2025-08-06
Show ProgressMonitorJobsDialog after delay in ProgressManager.run

fedejeanne:
Avoids a "flickering" produced by showing the progress dialog and closing it immediately for short-running tasks, improving the UX.

Before this PR, calling either one of these 2 methods ...
- `PlatformUI.getWorkbench().getProgressService().run(...)`
- `PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().getActivePart().getSite().getAdapter(IWorkbenchSiteProgressService.class).run(...)`

... and passing `false` as the 1st and/or the 2nd parameter (`fork` / `cancelable`) ended up showing a `ProgressMonitorJobsDialog` immediately. On the other hand passing `true` to both parameters showed the dialog after a short delay of **800 ms**.

After this PR, the dialog is always shown after a delay of 800 ms.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
@@ -967,9 +967,10 @@ public class ProgressManager extends ProgressProvider implements IProgressServic
 	@Override
 	public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable)
 			throws InvocationTargetException, InterruptedException {
-		if (!fork || !cancelable) {
+		if (!shouldRunInBackground() && (!fork || !cancelable)) {
 			// Backward compatible code.
-			final ProgressMonitorJobsDialog dialog = new ProgressMonitorJobsDialog(null);
+			final ProgressMonitorJobsDialog dialog = new ProgressMonitorJobsDialog(
+					ProgressManagerUtil.getDefaultParent());
 			dialog.run(fork, cancelable, runnable);
 			return;
 		}


iloveeclipse:
> ... and passing `false` as the 1st and/or the 2nd parameter (`fork` / `cancelable`) ended up showing a `ProgressMonitorJobsDialog` immediately

Which probably was expected / documented in javadocs? Have you checked that?
Behavior change may also break existing code / tests expecting new shell to appear immediately.

fedejeanne:
> Which probably was expected / documented in javadocs? Have you checked that?

The javadocs make no promises about whether or not the code runs in the UI thread.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/4a2727c448db22f5ebf7fad65987b100007de72e/bundles/org.eclipse.jface/src/org/eclipse/jface/operation/IRunnableContext.java#L44-L49

In fact, in this specialization, the javadoc says that it _might run the runnable asynchronously if `fork` is `true`_

https://github.com/eclipse-platform/eclipse.platform.ui/blob/4a2727c448db22f5ebf7fad65987b100007de72e/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/progress/IProgressService.java#L144-L145

>  Behavior change may also break existing code / tests expecting new shell to appear immediately.

The "backwards compatible code" was added in 2004 (28429cf73aba51cd7a49e451f2aa45e718c04f54) and it provides a bad UX (flickering). I'd very much rather improve the UX if possible, even at the cost of letting clients adapt after 20+ years.
But let's wait for the tests before making decissions :-)

fedejeanne:
Scratch my last comment, the JavaDocs talk about whether or not it runs in another thread but not about a delay. There are no promises about when is the dialog shown so this PR doesn't break any contract.

Side note: I'm not targeting this one for M3. I'll draft it as soon as I get to my computer.

fedejeanne:
Failing test is unrelated #3042

fedejeanne:
@iloveeclipse this PR is now ready to be reviewed. I see no problems w.r.t. the API contracts (see my last few comments). 

All good on your side?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
```diff
@@ -970,6 +970,8 @@ public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable
 		if (!fork || !cancelable) {
 			// Backward compatible code.
 			final ProgressMonitorJobsDialog dialog = new ProgressMonitorJobsDialog(null);
+			dialog.setOpenOnRun(false);
```
iloveeclipse: I wonder if:
1) The original code should simpy be wrapped into `if(!shouldRunInBackground()) ...`
2) The original code should use `ProgressManagerUtil.getDefaultParent()` instead of `null` on the line above

iloveeclipse:
@fedejeanne : could you provide steps to reproduce the original problem?

fedejeanne:
> @fedejeanne : could you provide steps to reproduce the original problem?

Either checkout https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2415 , activate _Organize imports_ as a save action and save some java file or use this view:

<details><summary>RunUsingProgressServiceView (ViewPart)</summary>
<p>

```java
import java.lang.reflect.InvocationTargetException;

import org.eclipse.core.runtime.IProgressMonitor;
import org.eclipse.jface.operation.IRunnableWithProgress;
import org.eclipse.swt.SWT;
import org.eclipse.swt.events.SelectionListener;
import org.eclipse.swt.widgets.Button;
import org.eclipse.swt.widgets.Composite;
import org.eclipse.ui.PlatformUI;
import org.eclipse.ui.part.ViewPart;
import org.eclipse.ui.progress.IProgressService;

public class RunUsingProgressServiceView extends ViewPart {

	/**
	 * The ID of the view as specified by the extension.
	 */
	public static final String ID = "repro.views.RunUsingProgressServiceView";

	@Override
	public void createPartControl(Composite parent) {

		createButton(parent, "Work fast", 1);

		createButton(parent, "Work slow", 100);
	}

	private void createButton(Composite parent, String text, int sleepMillis) {
		Button bt = new Button(parent, SWT.PUSH);
		bt.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
			runUsingProgressService(m -> doSomething(m, sleepMillis));
		}));

		bt.setText(text);
	}

	private void doSomething(IProgressMonitor monitor, int sleepMillis) throws InterruptedException {
		int MAX = 15;
		monitor.beginTask("Working...", MAX);

		for (int i = 0; i < MAX; i++) {
			Thread.sleep(sleepMillis);
			monitor.worked(1);
		}
		monitor.done();
	}

	private static void runUsingProgressService(IRunnableWithProgress runnable) {
		try {
			IProgressService progressService = PlatformUI.getWorkbench().getProgressService();
			progressService.run(true, false, runnable);
		} catch (InvocationTargetException | InterruptedException e) {
			e.printStackTrace();
		}
	}

	@Override
	public void setFocus() {
	}

}
```

</p>
</details> 

... and click the button "Work fast". You'll see the dialog appear and disappear in a glimpse.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
fedejeanne: Done

fedejeanne:
Test failure is unrelated: https://github.com/eclipse-platform/eclipse.platform.ui/issues/370

fedejeanne:
@iloveeclipse did you have a chance to validate the behavior? If yes, I'd like to merge.

iloveeclipse:
> ... and click the button "Work fast". You'll see the dialog appear and disappear in a glimpse.

The dialog still appears and disappers if I click "Work fast" fast enough

iloveeclipse:
> > ... and click the button "Work fast". You'll see the dialog appear and disappear in a glimpse.
> 
> The dialog still appears and disappers if I click "Work fast" fast enough

I had **not** enabled "Always run in background". Now enabling this I see no dialog.

Beside this, code looks better now.

iloveeclipse (approved):

fedejeanne:
Thank you for reviewing and merging, @iloveeclipse ! Regarding your previous comment:

> I had not enabled "Always run in background"...

Interesting. Now that you mention it, why does a dialog show up when pressing "Work Slow" and "Always run in background" is enabled? The description says: _Run long running operations in the background **where possible**_

image

Why isn't possible to skip the dialog altogether? Are the `ProgressManager` (obtained via `PlatformUI.getWorkbench().getProgressService()`) and the `WorkbenchSiteProgressService` (obtained via `PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().getActivePart().getSite().getAdapter(IWorkbenchSiteProgressService.class)`) not capable of skipping the dialog at all?

iloveeclipse:
Unfortunately this patch breaks existing PDE code, which means, it is backwards incompatible with existing applications.
See https://download.eclipse.org/eclipse/downloads/drops4/I20250919-0110/testresults/html/org.eclipse.pde.ui.tests_ep438I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

@fedejeanne : please try to fix it (tests in PDE fail reproducible with this change) or we have to revert.

iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/issues/1971

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
```diff
@@ -967,9 +967,10 @@ public void showInDialog(Shell shell, Job job) {
 	@Override
 	public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable)
 			throws InvocationTargetException, InterruptedException {
-		if (!fork || !cancelable) {
+		if (!shouldRunInBackground() && (!fork || !cancelable)) {
```
iloveeclipse, line 970: The `!shouldRunInBackground()` condition is what breaks the tests

iloveeclipse:
Proposed fix: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3298

fedejeanne:
Thank you for the heads-up and for the proposed fix, @iloveeclipse ! 

Unfortunately, #3298 also made the progress dialog disappear completely. I created #3311 to remedy that.

Issue #3141, opened 2025-08-01
latest 2025-09 milestone builds breaks editor cursor navigation keyboard shortcuts


martinlippert:
I am running a custom build that consumes the latest bits from 2025-09 staging area (which seems to be at the M2 level at the moment) and the experience around keyboard shortcuts for cursor navigation seems to be large broken and/or changed from previous versions.

- Eclipse SDK 2025-09 M1 seems to work like the previous versions.
- Eclipse SDK 2025-09 M2 seems to be broken or changed.

For example (on my macOS system)

- Cmd-Left/Right used to jump to the start/end of the line - this is different now
- Alt-Left/Right used to jump to the next position after "." (for example) inside the same line, now it seems to skip "." positions (and sometimes jump to another file)

Any idea what is causing this?

(originally reported to JDT UI via https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2399, but it turns out generic text editors show the same issue.

mickaelistria:
Can you please go to Preferences > General Keys, enable "Show key bindings when command is invoked: through keyboard" and report the command you get for those shurtcuts? And also report which commands you're expecting/getting with older versions?

Locally, I have in the Java Editor
* Ctrl+← Previous word
* Alt+← Backward history

In another older build of the IDE, I'm also seeing
* Ctrl+← Previous word
* Atl+← does nothing

martinlippert:
Sure:

2025-09 M1:

- Cmd+Left/Right: Line Start/End
- Alt+Left/Right: Next/Previous Word

2025-09 M2:

- Cmd+Left/Right: Next/Previous Word
- Alt+Left/Right: Backward/Forward History

martinlippert:
(Ctrl I+ Left/Right is taken my the OS, so traditionally on macOS, you use Cmd-Left/Right for that instead)

mickaelistria:
It seems like you get the linux or regular shortcuts applied instead of the mac-specific ones. Or more simply put, the mac-specific shortcults are either not available or not chosen.

The only commit that might be related I found (I might have miss some) in recent history is https://github.com/eclipse-platform/eclipse.platform.ui/commit/6aaf6b4285bf065b1a0c638fe4e61942fb8ea512#diff-f7d17d4cd3de2d3b83537af06dfeedd5d81088f1c75483ee5cb38b46812a4fe6 . You could try debugging with this reverted to see if this is the cause of a regression

[...]

PR #3148, opened 2025-08-06
Revert "Remove WS_CARBON check"

mickaelistria:
Reverts eclipse-platform/eclipse.platform.ui#3117 as it cause https://github.com/eclipse-platform/eclipse.platform.ui/issues/3141


merks:
Thanks for taking action.

PR #3152, opened 2025-08-07
Add new getPatternText() method to FilteredItemsSelectionDialog

jjohnstn:
- add new protected method to FilteredItemsSelectionDialog so that JDT UI can modify the output to add defaulted wildcards in camel case type names
- bump up org.eclipse.ui.workbench to 3.136.0
- needed for https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2387

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -2148,11 +2148,12 @@ public abstract class FilteredItemsSelectionDialog extends SelectionStatusDialog
 		 */
 		public ItemsFilter(SearchPattern searchPattern) {
 			patternMatcher = searchPattern;
-			String stringPattern = ""; //$NON-NLS-1$
-			if (pattern != null && !pattern.getText().equals("*")) { //$NON-NLS-1$
-				stringPattern = pattern.getText();
+			String stringPattern = getPatternText();
+			if (stringPattern != null && !stringPattern.equals("*")) { //$NON-NLS-1$
+				patternMatcher.setPattern(stringPattern);
+			} else {
+				patternMatcher.setPattern(""); //$NON-NLS-1$
 			}
-			patternMatcher.setPattern(stringPattern);
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


jjohnstn:
This change is required to be able to manage the type name text input in the Open Types dialog which inherits from the FilteredItemsSelectionDialog which holds the text widget and uses it to form search patterns.  The user request is for a way to default wildcards between camel case segments when searching for types rather than adding them manually.  I added the new method because the pattern control was private and the text was used directly in the code without an opportunity to add the wildcards.  It is possible that the preference or option could be moved to platform and then JDT needn't make any changes.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
```diff
@@ -2149,8 +2149,8 @@ public ItemsFilter() {
 		public ItemsFilter(SearchPattern searchPattern) {
 			patternMatcher = searchPattern;
 			String stringPattern = ""; //$NON-NLS-1$
-			if (pattern != null && !pattern.getText().equals("*")) { //$NON-NLS-1$
-				stringPattern = pattern.getText();
+			if (pattern != null && !getPatternText().equals("*")) { //$NON-NLS-1$
+				stringPattern = getPatternText();
 			}
 			patternMatcher.setPattern(stringPattern);
```
laeubi:
```suggestion
String stringPattern = getPatternText();
if (stringPattern != null && !stringPattern.equals("*")) { //$NON-NLS-1$
	patternMatcher.setPattern(stringPattern);
} else {
	patternMatcher.setPattern(""); //$NON-NLS-1$
}
```

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
```diff
@@ -2982,6 +2982,20 @@ public Control getPatternControl() {
...
+	 * @return text from pattern control or empty string if pattern control is null
+	 * @since 3.136
+	 */
+	protected String getPatternText() {
+		String text = ""; //$NON-NLS-1$
+		if (pattern != null) {
+			text = pattern.getText();
+		}
+		return text;
+	}
```
laeubi:
```suggestion
	protected String getPatternText() {
		if (pattern == null) {
			return null;
		}
		return pattern.getText()
	}
```

jjohnstn: Done

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
jjohnstn: Done

jjohnstn:
@laeubi Any further issues?  Can this be merged for M3?  If possible, could you assign this to me?  Thanks. On the JDT UI side I also am willing to scrap the new preference page for activating this and turn on the feature by default.

merks:
Could you rebase, squash to a single commit, and force push that.  I'll keep an eye out and merge this the build finishes.

jjohnstn:
@merks Squashed and rebased.  Thanks.

merks:
I’ll ignore the heap space failure. Monitoring progress of other platforms.

merks:
Test failure on macOS appears unrelated.

jjohnstn:
Thanks @merks

merks:
What goes around comes around. 😁

laeubi:
@merks thanks for taking over this. Sadly I was incapable of working on this the last few days.

Issue #3155, opened 2025-08-10
6 failures in org.eclipse.ui.genericeditor.tests on Java 25


stephan-herrmann:
Are you guys aware that org.eclipse.ui.genericeditor.tests has problems on Java 25?
We constantly see 6 failures during Y-builds, e.g.:
https://download.eclipse.org/eclipse/downloads/drops4/Y20250809-1000/testresults/html/org.eclipse.ui.genericeditor.tests_ep437Y-unit-linux-x86_64-java25_linux.gtk.x86_64_25.html

None of these are observed on Java 21, see the overview at https://download.eclipse.org/eclipse/downloads/drops4/Y20250809-1000/testResults.php

akurtakov:
Now that I-builds run Java 25 tests these can be seen at https://download.eclipse.org/eclipse/downloads/drops4/I20250813-1800/testresults/html/org.eclipse.ui.genericeditor.tests_ep437I-unit-linux-x86_64-java25_linux.gtk.x86_64_25.html .
Tests are quite obscure thus https://github.com/eclipse-platform/eclipse.platform.ui/pull/3175 , https://github.com/eclipse-platform/eclipse.platform.ui/pull/3178 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/3174 are applied. Hopefully making it a bit easier to spot the problem.

iloveeclipse:
No fails on master anymore, no idea why.

PR #3157, opened 2025-08-11
Use a separate ExecutorService in AsyncCompletionProposalPopup

fedejeanne:
> [!TIP]
> I'm targeting this PR for M1 so I'll keep it drafted. **Feel free to review and test anyway**.

## What it does
Use a dedicated `ExecutorService` in `AsyncCompletionProposalPopup` so that canceling the futures works as expected.

Before this PR, the method `AsyncCompletionProposalPopup.cancelFutures()` did not stop running futures, more precisely: it didn't stop the `computationFutures`. This was because the common pool was being used.

This PR introduces a dedicated executor service (with its own pool) that can be shut down immediately when the completion popup is closed.

## How to test
- Check out this PR and also these other 2:
  - (Mandatory) https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302
  - (Optional: avoids flooding the console with exceptions): https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2425
- Add this delay in `Parser::parse`, inside [this `for(;;)` - loop](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/43ac9cab64c667686ac9af41479d87dd40078fcf/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java#L11607) (right at the beginning):

```java
if (Thread.currentThread().getName().contains("org.eclipse.jface.text")) // ... ThreadPoolExecutor		
{
	try {
		System.out.println("Parsing...");
		Thread.sleep(500);
	} catch (InterruptedException e) {
		// Interrupt again so that the actual interruption mechanism (the if below) is triggered
		Thread.currentThread().interrupt();
		System.out.println("Interrupted during sleep!");
	}
}
```
- Start the workbench and start some autocompletion
- Cancel the autocompletion by clicking outside the popup or by pressing the `esc`-key

### Expected outcome
- If you added the delay in `Parser::parse` then you should see this in the output, several times: `Parsing...`
- After pressing `esc`, you should see no more occurrences of `Parsing...` in the console and you should see `Interrupted during sleep!`

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/Activator.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/Activator.java
@@ -1,0 +1,68 @@
+package org.eclipse.jface.text;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+
+/**
+ * @since 3.29
+ */
+public class Activator implements BundleActivator {
+	/**
+	 * The identifier of the descriptor of this plugin in plugin.xml.
+	 */
+	public static final String ID= "org.eclipse.jface.text"; //$NON-NLS-1$
+
+	private static Activator activator;
+
+	private ExecutorService executor;
+
+	@Override
+	public void start(BundleContext context) {
+		activator= this;
+	}
+
+	@Override
+	public void stop(BundleContext context) {
+		activator= null;
+		if (executor != null) {
+			executor.shutdownNow();
+			executor= null;
+		}
+	}
+
+	public static ExecutorService getExecutor() {
+		activator.createExecutor();
+		return activator.executor;
+	}
+
+	private void createExecutor() {
+		if (activator.executor != null)
+			return;
+
+		executor= new ThreadPoolExecutor(
+				Runtime.getRuntime().availableProcessors(),
+				Runtime.getRuntime().availableProcessors(),
+				3L, TimeUnit.SECONDS,
+				new LinkedBlockingQueue<>(),
+				new ThreadFactory() {
+					AtomicInteger count= new AtomicInteger(1);
+
+					@Override
+					public Thread newThread(Runnable r) {
+						// Name the threads numerically for better debugging.
+						Thread t= new Thread(r, ID + "-worker-" + count.getAndIncrement()); //$NON-NLS-1$
+
+						// No need to keep the JVM running just because of the completion proposals
+						t.setDaemon(true);
+						return t;
+					}
+				});
+	}
+}
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
@@ -23,8 +23,11 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -43,6 +46,7 @@ import org.eclipse.core.runtime.SafeRunner;
 
 import org.eclipse.jface.contentassist.IContentAssistSubjectControl;
 
+import org.eclipse.jface.text.Activator;
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.DocumentEvent;
 import org.eclipse.jface.text.IDocument;
@@ -372,7 +376,7 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 		}
 		List<CompletableFuture<List<ICompletionProposal>>> futures = new ArrayList<>(processors.size());
 		for (IContentAssistProcessor processor : processors) {
-			futures.add(CompletableFuture.supplyAsync(() -> {
+			futures.add(submitInterruptible(() -> {
 				AtomicReference<List<ICompletionProposal>> result= new AtomicReference<>();
 				SafeRunner.run(() -> {
 					ICompletionProposal[] proposals= processor.computeCompletionProposals(fViewer, invocationOffset);
@@ -389,9 +393,40 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 					return Collections.emptyList();
 				}
 				return proposals;
-			}));
+			}, Activator.getExecutor()));
 		}
 		return futures;
+	}
+
+	/**
+	 * Submit a task in such a way that it actually reacts to cancellation (i.e. calls to
+	 * {@code future.cancel(true)})
+	 *
+	 * @param executor Do not use the common pool here since that one does not cancel (interrupts)
+	 *            worker threads
+	 * @return an interruptible future.
+	 */
+	private static <T> CompletableFuture<T> submitInterruptible(
+			Callable<T> task, ExecutorService executor) {
+
+		CompletableFuture<T> cf= new CompletableFuture<>();
+
+		Future<?> ft= executor.submit(() -> {
+			try {
+				cf.complete(task.call());
+			} catch (Exception e) {
+				cf.completeExceptionally(e);
+			}
+		});
+
+		// make canceling the CF also cancel the FutureTask
+		cf.whenComplete((r, t) -> {
+			if (cf.isCancelled()) {
+				ft.cancel(true); // this actually interrupts
+			}
+		});
+
+		return cf;
 	}
 
 	private String getTokenContentType(int invocationOffset) throws BadLocationException {
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.text
-Bundle-Version: 3.28.200.qualifier
+Bundle-Version: 3.29.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -40,4 +40,6 @@ Require-Bundle:
 Import-Package: com.ibm.icu.text
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Automatic-Module-Name: org.eclipse.jface.text
+Bundle-Activator: org.eclipse.jface.text.Activator
+Bundle-ActivationPolicy: lazy
 Require-Capability: eclipse.swt;filter:="(image.format=svg)"


laeubi:
@fedejeanne why I would not object to use an own pool, I still wonder why the features can not be canceled? Also using a single threadpool is different from using the common pool because it guarantees some kind of parallelism.

fedejeanne:
@laeubi there seems to be no guarantee that the futures will be canceled immediately if `mayInterruptIfRunning` is set to `true`. This is from the [JavaDocs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ForkJoinTask.html#cancel%28boolean%29):

>     mayInterruptIfRunning - this value has no effect in the default implementation because interrupts are not used to control cancellation.

There are some more details in [this stackoverflow post](https://stackoverflow.com/questions/21320156/forkjoinpool-resets-thread-interrupted-state)

laeubi:
> @laeubi there seems to be no guarantee that the futures will be canceled immediately if `mayInterruptIfRunning` is set to `true`. This is from the [JavaDocs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ForkJoinTask.html#cancel%28boolean%29):
> 
> > ```
> > mayInterruptIfRunning - this value has no effect in the default implementation because interrupts are not used to control cancellation.
> > ```
> 
> There are some more details in [this stackoverflow post](https://stackoverflow.com/questions/21320156/forkjoinpool-resets-thread-interrupted-state)

But there is also no guarantee that the executor is terminating immediately. Request for cancel is always only a _best effort_ and it depends on the code to handle it properly, so the question would be what exact problem is here to be solved? So I suspect some code currently checks for `Thread.interrupted()` should instead check for `CompletableFuture.isCancelled()` instead.

fedejeanne:
@laeubi , sorry, some context is still missing. I was in the process of creating other PRs that are related to this one:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2425
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302

You are right: cancellation is done in a best effort and someone (the task) needs to check for cancellation on its own. That one is done in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302 and I've added some details in there (see _How to test_) explaining how all 3 PRs play together. Does this help?

fedejeanne:
Test failures were already there: #3155

laeubi:
@fedejeanne I'll try to look into this in more details but will be unavailable for a few days now. just some hints here:

- in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2425 you see we have a `ProgressMonitor`, cancellation is usually handled through that pattern in Eclipse code
- in the context of features, one would likely pass a `ProgressMonitor` that in it `isCanceled()` method checks for the cancellation of the feature (similar to this one here that cancel on first match found: https://github.com/eclipse-pde/eclipse.pde/blob/e7b608f63275301b95614581c3462c4bb04b305e/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/search/dependencies/GatherUnusedDependenciesOperation.java#L431-L443
- Similar one might want to interrupt the current thread of an operation running in a feature if it is required on cancellation instead of throw away the whole executor

So I think we should really see if it can be solved without shutdown the whole pool (what makes a pool quite useless here) if we just want to cancel some of the features.

fedejeanne:
@laeubi I also thought of using a progress monitor, as it is the "standard" way of doing it RCP, but the thing is that the method I want to reach (`Parser::parse`) is too deep in the call hierarchy. You'll notice that `org.eclipse.jdt.core.compiler.batch` has no dependency to any other bundles... tricky.

laeubi:
Some one is calling this code and someone is canceling the future :-)
So somewhere there I would expect to cancel the feature and the thread (if not other way is there). But thread cancellation is something that is poorly supported by the platform code anyways so you will probably get some surprise.

fedejeanne:
?

Sorry, you'll have to lay it down a bit differently because you got me confused 😅 

... the code is called and canceled in `AsyncCompletionProposalPopup`:

- Called (started) in: `buildCompletionFuturesOrJobs(...)` (see `CompletableFuture.supplyAsync(() -> { ... }, completionFuturesExecutor));`)
- Canceled in: `cancelFutures()` (when the user clicks outside the completion proposals pop-up)

Isn't that what you mean?

mickaelistria:
I also have concerns with lots of multi-threading tweaks. Can you please clarify which issue you're attempting to fix and provide steps to reproduce? Usually, the best approach is to have the future code have a monitor that allows to check when a query is to be discarded. I would rather suggest to try to have such monitor and use it in the actual future than having a more complex orchestration framework.
Also, for the particular case of Parser, I would even date to say that parser is almost never a problem, it is usually very very fast. Other bits like symbol/binding resolutions are the expensive ones to skip. Basically, it should be better to let the parser complete and interrupt later than to try having an interruptible parser.

fedejeanne:
@laeubi I adapted the PR so that it uses a custom `ThreadPoolExecutor` that can handle cancellation without the need to shutting it down.

@mickaelistria the particular error that motivated this PR is really difficult to reproduce and I only saw it happen in our big Workspace (~2000 plugins/fragments and lots of auto-generated metamodel classes). But the parser is actually the issue since:

1. It wastes too much time trying to recover from a syntax error (see VVM screenshot) and

<details><summary>VVM Screenshot</summary>
<p>

image

</p>
</details> 


2. Regardless of whether or not the parser is slow, the current code already tries to cancel the execution of the futures (see the invocation of [`future.cancel(true)` here](https://github.com/eclipse-platform/eclipse.platform.ui/blob/7d4cca30f4e36dd5d1d6d7ed19d14706cf21c188/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java#L342)), but this is not happening since the common-pool (_i.e._ the `ForkJoinPool`, which is being used implicitly when calling [`CompletableFuture.supplyAsync( () -> ...)`](https://github.com/eclipse-platform/eclipse.platform.ui/blob/7d4cca30f4e36dd5d1d6d7ed19d14706cf21c188/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java#L375C15-L375C45)) does not propagate the cancellation to the worker threads.

What I mean to say is that, even if the error doesn't happen often, this PR (and the other ones too) implement the functionality that was already intended to be delivered, namely: to **actually** cancel the long running operations that calculate the completion proposals. Without this PR, any other `IContentAssistProcessor` that is added by other plugins might simply waste CPU cycles by calculating completion proposals even when the user already closed the proposals pop-up (JDT is just 1 consumer of this API).

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -60,6 +67,22 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 
 	private static final int MAX_WAIT_IN_MS= 50; // TODO make it a preference
 
+	private static final ExecutorService EXECUTOR= new ThreadPoolExecutor(
```
laeubi: I think you should explain a bit here why this is used. and why given values are chosen.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -389,11 +412,46 @@ protected List<CompletableFuture<List<ICompletionProposal>>> buildCompletionFutu
...
 
+	/**
+	 * Submit a task in such a way that it actually reacts to cancellation (i.e. calls to
+	 * {@code future.cancel(true)})
+	 *
+	 * @param executor Do not use the common pool here since that one does not cancel (interrupts)
+	 *            worker threads
+	 * @return an interruptible future.
+	 */
+	private static <T> CompletableFuture<T> submitInterruptible(
```
laeubi, line 409: All this codes makes me wonder why not using a plain `FutureTask` in the first place?

fedejeanne: `FutureTask` doesn't support chaining (`thenAccept` / `thenRun` ...)

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
fedejeanne: Can you be more specific?

The chosen values are thought for CPU-bound tasks and tasks are not to be dropped.

fedejeanne: Done

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
fedejeanne: Ok @laeubi ?

laeubi: > `FutureTask` doesn't support chaining (`thenAccept` / `thenRun` ...)

I know but where is it actually used here? I looked at the code and except some places where features are combined (to be canceled together) and when adding the results to a list the completable future features are not used at all.

So my feeling is more, drop `CompleteableFuture` at all, use a "classic" future task and let the task itself publish its result when done would simply the handling here allow for better cancellation and make the code much clearer.

fedejeanne: If I get rid of the chaining then I have to orchestrate everything manually, which generates more boilerplate code. It's not more readable.

How about you propose a follow-up PR and discuss this in there? :-)

laeubi: > If I get rid of the chaining then I have to orchestrate everything manually

You already "orchestrate" it manually in the method where you create a `FutureTask` and then convert it into a completable future.

> How about you propose a follow-up PR and discuss this in there? :-)

It does not sound useful to commit some new code to then immediately replace that code with something else. So if there are specific things you are unsure how to implement I would rather like to give some hint so we develop a consistent solution in the first place.

If you think we should keep the `Completable` future indirection here its fine but then I think some of the details must be discussed, e.g. why not submit the future task directly. I would be okay with that but just wanted to be sure first we do not make it more complicated/indirect than required.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -389,11 +419,46 @@ protected List<CompletableFuture<List<ICompletionProposal>>> buildCompletionFutu
...
+
+		executor.submit(() -> {
+			try {
+				ft.run(); // executes in this thread
+				cf.complete(ft.get());
+			} catch (CancellationException e) {
+				cf.cancel(true);
+			} catch (Exception e) {
+				cf.completeExceptionally(e);
+			}
```
laeubi: As cancel effective sets the exception to contain a `CancellationException` this can simply become


```suggestion
			} catch (Exception e) {
				cf.completeExceptionally(e);
			}
```

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -389,11 +419,46 @@ protected List<CompletableFuture<List<ICompletionProposal>>> buildCompletionFutu
...
+	private static <T> CompletableFuture<T> submitInterruptible(
+			Callable<T> task, ExecutorService executor) {
+
+		CompletableFuture<T> cf= new CompletableFuture<>();
+		FutureTask<T> ft= new FutureTask<>(task);
+
+		executor.submit(() -> {
+			try {
+				ft.run(); // executes in this thread
+				cf.complete(ft.get());
```
laeubi: Executor submit already returns a `Future` why do we need a separate one here?

One would expect to simply submit a runnable as below that simply calls the callable directly.


```suggestion
		Future<T> ft = executor.submit(() -> {
			try {
				cf.complete(task.call());
```

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
fedejeanne: I appreciate the comments and am ok with trying to find the right solution from the get go. But please, give me some snippet instead of simply making a scarce comment like "_why not..._". I often find that a snippet avoids confusion.

For example, did you mean "_[...] why not ~using~ receiving a plain FutureTask **as a parameter** in the first place?_" ? If you did then the answer is "Because of readability: `FutureTask` is not a functional interface but `Callable` is, so I just liked the code better as it is right now".

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
fedejeanne: Done

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
fedejeanne: Done

fedejeanne:
@laeubi / @mickaelistria , all code propositions have been adopted (thank you for them!) and conversations have been answered. I'd like to merge this one early next week so please let me know if I left something unaddressed :-)

I'm logging off, have a nice weekend!

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -60,6 +66,29 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 
 	private static final int MAX_WAIT_IN_MS= 50; // TODO make it a preference
 
+	/**
+	 * Designed for CPU intensive tasks, unlimited queue (no task-dropping).
+	 *
+	 */
+	private static final ExecutorService EXECUTOR= new ThreadPoolExecutor(
```
laeubi: I wonder if we can use a `Executors.newWorkStealingPool()` here or will it also not allow cancel of `FutureTasks`?

laeubi: One (ugly) detail is the pool is never shut down now.

In other place (I think it was m2e), I solved it in the way that the `Activator` has a `registerShutdownAction` and you can submit a runnable there that is called when the bundle is stopped. then here in a static initializer one would be able to shutdown the pool.

fedejeanne: It doesn't cancel the futures.

I added details on _How to test_ to the description so you can give it a try too ;-)

laeubi: Another way would be of course to add the executor to the activator itself and manage it there...

fedejeanne: Is this really necessary? What are the chances of shutting down the plugin without closing the IDE?

btw I couldn't find the method you mentioned in GitHub.

fedejeanne: @laeubi I don't see the need to do this since the autocompletions are necessary for as long as the IDE is open. Closing the IDE will shut down the pool.

laeubi: Bundles can be started/stopped /refreshed or updated/uninstalled any time. Also the IDE will not shutdown the pool, the JVM will on exit maybe destroy them anyways.

fedejeanne: Sorry for insisting but I still don't understand _what usecase do you try to improve_?

laeubi: Simply in OSGi if you allocate any resources you should free them, at least when the bundle shuts down.

So assume I would _refresh_ the bundle (or _update_ it) and it gets restarted, then:

1. The (old) bundle pool will hang around forever
2. The (old) classloader of the bundle will not be able to be disposed and leaking class instances (with maybe other static data cached somewhere)

Not that Eclipse IDE works well under that condition currently (and that's why we force the user to reboot in many cases) but this does not mean one should make it worse.

fedejeanne: Can you please link the code you mentioned before? I couldn't find any method called `registerShutdownAction` in GH. 

I'd like to preserve the executor in the same class but the issue I'm having is that, being a `static final` field, it's not recreated when I stop + start the bundle `org.eclipse.jface.text`

laeubi: > I'd like to preserve the executor in the same class but the issue I'm having is that, being a static final field, it's not recreated when I stop + start the bundle org.eclipse.jface.text

Yes that's really a problem here and why its probably bad to make it static in the first place. And thats why I mentioned the activator, even though it does not sounds so nice in the first place, this already is handling this in many cases. Also thinking further to have a _JFace Executor Service_ instead of many classes having their own (like we have Images/Fonts registries) might be something to consider.

> Can you please link the code you mentioned before? I couldn't find any method called `registerShutdownAction` in GH.

It seems this all was replaced with using DS a few years ago https://github.com/eclipse-m2e/m2e-core/commit/22a3184d936e61dc04b5e3b241c85bbb1145868b

fedejeanne: I added it in the activator

fedejeanne: @laeubi fine for you?

laeubi (approved):
Looks fine to me.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/Activator.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@Override
+	public void stop(BundleContext context) {
+		if (executor != null) {
+			executor.shutdownNow();
+			executor= null;
+		}
+	}
+
+	public static ExecutorService getExecutor() {
```
laeubi: you might want to lazy init the value here!

fedejeanne: Done

PR #3158, opened 2025-08-12
Fix LineNumberRuler for fixed zoom

### LINKED ISSUE #334: Several wrong font sizes with "-Dswt.autoScale"
See: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3061

## Summary
There are 2 things to extract from the official issue:
1. Most of the fonts do **not** react to the parameter `-Dswt.autoScale`. This is **by choice** because:
  - This was the "normal" behavior before we introduced the improved HiDPI support
  - Changing this behavior is not trivial and may irritate some users
 2. The font size of the line number ruler **does react to the parameter `-Dswt.autoScale`** but the ruler does **not** adapt its width to it. This one is definitely a bug and needs to be fixed.

### LINKED ISSUE #398: Fix LineNumberRuler with fixed zoom with monitor specific scaling

akoch-yatta:
This PR adds a layout call to the callback executed after a zoom change for the Canvas. If a fixed zoom is used and monitor specific scaling is active the font sizes will be correctly adapted after a zoom change, but the size of the Canvas is not, leading to cut off line numbers. Forcing a layout for this rare case of event will ensure a proper resizing.

### How to test

Start a runtime with
```
-Dswt.autoScale=150
```

on two monitors with different zoom than 150, best one bigger, one smaller and move a Text Editor with line numbers between the monitor.

**Before**
Screenshot 2025-08-12 132002

**After**
Screenshot 2025-08-12 132555

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java
@@ -624,6 +624,7 @@ public class LineNumberRulerColumn implements IVerticalRulerColumn {
 					fBuffer.dispose();
 					fBuffer= null;
 				}
+				layout(false);
 			});
 		});
 


HeikoKlare (approved):
Change is sound and works as expected for me.

PR #3159, opened 2025-08-12
Remove unused, outdated and non-working SessionTests

HeikoKlare:
When starting to modernize tests, I came across the `SessionTests` which seem to have been disabled for decades. They do not start at all because they do not find their data directories and even if they are fixed to find them, most of those tests fail. In addition, they introduce dependencies to the `SessionTestSuite` that is about to be removed.

This how the execution results look like. Note that in session tests usually only the second and follow runs are of interest. The first run just loads an existing workspace and does some checks on it, which you could probably even do without a session test:
image

With this change, I propose to remove the unused SessionTests, their data, launch configuration, and reference in the tests.xml, which has been deactivated for decades. Some classes used in other tests are moved to the according packages.

If anyone knows if there is some value in these tests, would be great to know.


iloveeclipse:
Could you please check why the tests were disabled via git blame and add link to the bugzilla ticket? I assume e4 transition was involved here.

HeikoKlare:
> Could you please check why the tests were disabled via git blame and add link to the bugzilla ticket? I assume e4 transition was involved here.

Sure and I think you are right: they were disabled in 2013 (https://bugs.eclipse.org/bugs/show_bug.cgi?id=398869)

HeikoKlare:
@akurtakov @iloveeclipse @vogella do you see any value in keeping these tests or extracting something out of them? Since they have been disabled for 12 years now, no one really seems to be missing them. So they seem to just produce maintenance overhead now. At the same time, I don't want to remove something if there may still be any value in them, thus asking you as long-standing contributors what you think about them.

akurtakov:
IMO if something wasn't run/used for 12 years - it should be removed.

HeikoKlare:
Thank you for the feedback! Then let's remove them. They will still be avaiable from Git history if someone wants to take something over.

PR #3164, opened 2025-08-12
Replace fail() calls in org.eclipse.ui.tests

HeikoKlare:
This replaces JUnit fail() calls in org.eclipse.ui.tests by various means:
- Replace exception catching and fail() statements with assertThrows()
- Remove exception catching and fail() statements by letting the test methods simply throw the exception


akurtakov:
I highly appreciate this sanitization of tests! It's a very underestimated and thankless job but when things go wrong - these are the things that helps us recover faster!

HeikoKlare:
> I highly appreciate this sanitization of tests! It's a very underestimated and thankless job but when things go wrong - these are the things that helps us recover faster!

Fully agree. Always great to see test improvement PRs from everyone around to ensure that we keep getting the necessary and valuable feedback on the changes we make and on anything that break.

PR #3177, opened 2025-08-14
Remove long deactivated DnD tests from org.eclipse.ui.tests

HeikoKlare:
The DnD tests in `org.eclipse.ui.tests` have effectly been deactivated 12 years ago (probably during E4 transition) by (temporarily) changing to a different test suite:
- https://github.com/eclipse-platform/eclipse.platform.ui/commit/b0b57f0bbe35ab94e81eeac824c669432072a9f4

When switching back to the initial test suite, the DnD tests were disabled for that test suite and never reenabled afterwards:
- https://github.com/eclipse-platform/eclipse.platform.ui/commit/328d2f9ad8a06f4c7519c305ae43d008e7c628a0

Currently, the disablement is done by the `@Ignore` annotation at the ´DragTestSuite`.

The tests are/became incompatible since then. They particularly lack the ability to hook into the drop process that was previously possible with some functionality of the `DragUtil` class. In addition to these tests not being used for 12 years, they now introduce maintenance issues:
- They use a JUnit-3-specific setup that is difficult to adapt to JUnit 4/5
- They are the sole remaining users of an AutoTest fragment including specific test suite and logger implementations that are also still tied to JUnit 3, as well as UITestCase specializations

I do not think that it's worth investing effort into technical maintenance (somehow adapting to JUnit 4/5) without the tests being functional at all. That's why I propose to remove them _and_ create an issue that documents that we don't have such DnD tests anymore with a link to this PR that removed them, such that someone could easily restore the current version of the tests for reviving them, if desired.

@akurtakov @iloveeclipse @vogella asking on this again as it is similiar the unused session tests (https://github.com/eclipse-platform/eclipse.platform.ui/pull/3159) but other than the diverse session tests this affects a complete set of functionality (DnD) which does not seem to be covered by tests at all.


akurtakov:
If the number of run tests doesn't decline - this should be ready to do.
No one cared to revive the tests and even if someone does now it would be most likely easier to write new tests than understanding and reviving those.

HeikoKlare:
The test bot reports 1 removed test, which is just the ignored test suite:
image

As you can see in the precise results, the number of executed tests is the same, it's just one failure of a test that is know to randomly fail: https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

akurtakov:
> The test bot reports 1 removed test, which is just the ignored test suite: image
> 
> As you can see in the precise results, the number of executed tests is the same, it's just one failure of a test that is know to randomly fail: #294

Seems to be good in this case. Feel free to merge when you consider it done.

HeikoKlare:
As proposed, I have added an issue as a reminder that such tests are missing:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/3182

Maybe it could be something for newcomers to dig into how different parts of Platform UI work, which is why I marked it as a "good first issue".

PR #3179, opened 2025-08-14
Remove obsolete exception handling in tests

HeikoKlare:
Let the tests throw the exceptions instead to be handled by the JUnit framework. The adapted places do not throw CoreExceptions that would require additional handling for proper formatting.


HeikoKlare:
Failing test is known as randomly failing: https://github.com/eclipse-platform/eclipse.platform.ui/issues/370

PR #3185, opened 2025-08-15
Fix split package signing problem org.eclipse.ui.editors

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3190


merks:
@MohananRahul  @akurtakov 

I would prefer we don't contribute a build with split package signing problems for M3.    So I kicked this build with the changes from this PR:

https://ci.eclipse.org/releng/job/Builds/job/I-build-4.37/108

subyssurendran666:
The promote build job for the M3 has already done(https://download.eclipse.org/eclipse/downloads/drops4/S-4.37M3-202508141800/). So shall we proceed with the M3a.

merks:
Hmmm.  I see.  I don't want to create a lot of extra work either.  😢

subyssurendran666:
> Hmmm. I see. I don't want to create a lot of extra work either. 😢

Could we consider including this fix in RC1 and proceed with the current M3?

merks:
Yes.

PR #3192, opened 2025-08-15
Minor fix to Window.Java.

grissinibread:
Incorrect shell was being called in initial if statement. 
Corrected to 'loopShell' which is the parameter created for this function.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/window/Window.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/window/Window.java
@@ -803,22 +803,20 @@ public abstract class Window implements IShellProvider {
 	}
 
 	/**
-	 * Runs the event loop for the given shell.
+	 * Runs the event loop for shell.
+	 * <p>
+	 * This method assumes that parameter is not null.
+	 * </p>
 	 *
 	 * @param loopShell
-	 *            the shell
+	 * 				the shell
 	 */
-	private void runEventLoop(Shell loopShell) {
+	private static void runEventLoop(Shell loopShell) {
 
-		//Use the display provided by the shell if possible
 		Display display;
-		if (shell == null) {
-			display = Display.getCurrent();
-		} else {
-			display = loopShell.getDisplay();
-		}
+		display = loopShell.getDisplay();
 
-		while (loopShell != null && !loopShell.isDisposed()) {
+		while (!loopShell.isDisposed()) {
 			try {
 				if (!display.readAndDispatch()) {
 					display.sleep();
--- a/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface;singleton:=true
-Bundle-Version: 3.38.0.qualifier
+Bundle-Version: 3.38.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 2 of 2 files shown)


HeikoKlare (commented):
Thank you for the proposal. The code indeed looks wrong. It just seems to work because this method is only called at a single place where `this.shell` is passed as `loopShell`, such that both values are effectively always the same when reaching this line of code.

So I guess this is something you found by accident, but you did not experience any issues because of this, did you?

To improve this correction, I propose to do either of the following:
- Make the method also static, as that's what it effectively is
- Remove the `loopShell` parameter and directly reference `shell`

In addition to those, I propose to remove the null-check (and instead clarify in the contract that the argument is assumed to be non-null). This contract is currently fulfilled by the sole consumer of the method anyway, so no need for unnecessary null-checks and unused fallback mechanisms.

grissinibread:
I did in fact find this by accident while focusing on another issue! I also did not encounter any issues because of this. I went ahead and implemented your corrections/proposal, check it out and let me know if everything is in check 🤓

Thank you for checking out my proposal as well as the suggestions @HeikoKlare ☺️

bundles/org.eclipse.jface/src/org/eclipse/jface/window/Window.java
```diff
@@ -200,7 +200,7 @@ public void handleException(Throwable t) {
 	/**
 	 * Window shell, or <code>null</code> if none.
 	 */
-	private Shell shell;
+	private static Shell shell;
```
HeikoKlare: You must not make the shell field static. That value is specific to the actual `Window` instance. With this change, every `Window` will use the same shell. Since you made the `runEventLoop` method static, you need to still pass a `Shell` argument into it.

HeikoKlare (dismissed):

bundles/org.eclipse.jface/src/org/eclipse/jface/window/Window.java
grissinibread: Got it! I took your suggestions and tried to combine them without thinking, then tried to find a quick solution to my errors. I went ahead and fixed it up!

HeikoKlare:
Failing test is known to be flaky: https://github.com/eclipse-platform/eclipse.platform.ui/issues/370

iloveeclipse:
This change breaks multiple application tests for us (using I20250904-1800 build).
I'm investigating the root cause, interestingly we don't see error reported with the stack trace, just that various tests fail with `Cannot invoke "org.eclipse.swt.widgets.Shell.getDisplay()" because "loopShell" is null`.

iloveeclipse:
Found stack traces (I expected them in the workspace log but they were not handled and were reported in the tests log).

Test does this:
```
WizardDialog dialog = new WizardDialog(PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(), wizard);
dialog.create();
dialog.open();
```

and becomes this:

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.Shell.getDisplay()" because "loopShell" is null
 at org.eclipse.jface.window.Window.runEventLoop(Window.java:817)
 at org.eclipse.jface.window.Window.open(Window.java:799)
 at SomeWizardAction.run(SomeWizardAction.java:58)
 at XYZ.testExport(XYZ.java:156)
 at java.base/java.util.Optional.ifPresent(Optional.java:178)
 at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
 at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939)
 at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)

```

iloveeclipse:
I've created https://github.com/eclipse-platform/eclipse.platform.ui/issues/3242

Issue #3182, opened 2025-08-14
No tests for drag-and-drop functionality


HeikoKlare:
Tests for drag-and-drop functionality have been implemented around 20 years ago, but were disabled during the E4 transition in 2012 via disabling the according test suite (https://github.com/eclipse-platform/eclipse.platform.ui/commit/b0b57f0bbe35ab94e81eeac824c669432072a9f4), and they were never reenabled even after the test suite was used again (https://github.com/eclipse-platform/eclipse.platform.ui/commit/328d2f9ad8a06f4c7519c305ae43d008e7c628a0).

Since the tests were not run for 13 years and became technically and functionally outdated, those tests have finally been removed via:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3177

Still, tests for the drag-and-drop functionality would be useful, so this issue is a reminder to add such tests. The tests removed with the above mentioned PR might be a good indicator for what functionality could be relevant to test (and to a certain degree maybe also for how to do it).

aloowood:
Hi, can I take this?

HeikoKlare:
> Hi, can I take this?

Sure

vogella:
Looks like the work started https://github.com/eclipse-platform/eclipse.platform.ui/pull/3195

vogella:
See also https://github.com/eclipse-platform/eclipse.platform.swt/issues/3192 for a recent issue with DND under Linux https://github.com/eclipse-platform/eclipse.platform.swt/issues/3192

PR #3195, opened 2025-08-16
Add DnD tests issue #3182

aloowood:


features/.project
```diff
@@ -0,0 +1,11 @@
...
+<projectDescription>
+	<name>features</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
```
merks, line 11: Adding a .project file here is not needed and not desired.

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DetachedDropTarget.java
```diff
@@ -0,0 +1,35 @@
...
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
```
merks: I guess you copied this from somewhere.

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DetachedWindowDragTest.java
```diff
@@ -0,0 +1,26 @@
+package org.eclipse.ui.tests.dnd;
```
merks: Every java file needs a "standard" copyright header.

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DragTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    TestDragSource dragSource;
+    TestDropLocation dropTarget;
+
+    static IProject project;
+    static IFile file1, file2, file3;
+
+    IEditorPart editor1, editor2, editor3;
+
+    static WorkbenchWindow window;
```
merks: Can this be an IWorkbenchWindow?  Should any of these things really be static?

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DragTest.java
```diff
@@ -0,0 +1,95 @@
...
+    static WorkbenchPage page;
```
merks: Can this be an IWorkbenchPage?

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DragTestSuite.java
```diff
@@ -0,0 +1,207 @@
...
+	@org.junit.jupiter.api.Test
+	private void addAllCombinations(TestDragSource dragSource,
+			TestDropLocation[] dropTargets) {
+
+		for (TestDropLocation dropTarget : dropTargets) {
+			if (dropTarget == null) {
+				continue;
+			}
+
+			DragTest newTest = new DragTest(dragSource, dropTarget);
```
merks: Avoid warnings.

merks (changes requested):
Hopefully it's possible to use IWorkbenchWindow and IWorkbenchPage instead of non-API implementation classes.

In general, please keep the PR to a single commit.  Using amend and force push for that purpose rather than keep adding commits.

aloowood:
I have applied the requested changes, cleaned up all warnings, and rebased on top of the latest master. Ready for another review.

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DragOperations.java
```diff
@@ -0,0 +1,92 @@
...
+public class DragOperations {
+
+	/**
+	 * Drags the given view OR editor to the given location (i.e. it only cares that we're given
+	 * a 'Part' and doesn't care whether it's a 'View' or an 'Editor'.
+	 * <p>
+	 * This method should eventually replace the original one once the Workbench has been updated
+	 * to handle Views and Editors without distincton.
+	 */
+	public static void drag(IWorkbenchPart part, TestDropLocation target, boolean wholeFolder) {
```
HeikoKlare: I don't think the tests can really work of this essential method still fails, do they?

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DragTestSuite.java
```diff
@@ -0,0 +1,218 @@
...
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.IPageLayout;
+import org.junit.Ignore;
+import org.junit.runner.RunWith;
+import org.junit.runners.AllTests;
+
+import junit.framework.Test;
+
+@RunWith(AllTests.class)
+@Ignore("DND support not implemented yet in E4")
```
HeikoKlare: You need to remove the `@Ignore` and should set the test suite as the class executed for the bundle in the `build.properties` like in the other test bundles, e.g.:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/223dfe7383547fdead8d59f43fa7d8ad6b2e2022/tests/org.eclipse.ui.tests/build.properties#L32

We finally also need to add a proper `test.xml` to have the tests executed in Jenkins.

tests/org.eclipse.ui.tests.dnd/src/org/eclipse/ui/tests/dnd/DragTestSuite.java
```diff
@@ -0,0 +1,218 @@
...
+public class DragTestSuite {
```
HeikoKlare: This pretty much looks like a JUnit-3-style TestSuite. Since we are about to migrate everything to JUnit 5 (I just migrated everything to JUnit 4), reintroducing these tests should also incorporate and up-to-date style of the test suite.

HeikoKlare (commented):
Thank you for the update. Before we can take a deeper look into the proposal, there are some general concerns to address.

aloowood:
I incorporated the feedback, rebased onto the current master, and squashed everything into a single commit. Thanks in advance for taking another look.

HeikoKlare (commented):
Thank you for the latest update. Unfortunately, there is still no proper test suite and no test executed at all.
The `DragTestSuite` class is not a proper test suite (neither JUnit 4 nor JUnit 5). It is just a plain JUnit 5 test class with a paramaterized test method that effectively does nothing but instantiating the `DragTest` class without executing anything at all.
You will also notice that when executing the tests each of them is executed in less than millisecond (at least on my system) which is unlikely for a UI test.

HeikoKlare:
One additional hint to ensure that you first come to a functional state of the tests before pinging us for review: make sure that the tests that you add/update are actually executed by testing it locally. Run the tests in your IDE and in case you are not familiar with how to set up JUnit 4/5 tests / test suites and are in doubt if it works, you can e.g. set a breakpoint in the actual test methods to ensure that they are really executed. You should also run the single test classes individually to see that they also work without running the complete test suite.

aloowood:
Hi @HeikoKlare,
I’m working on the DnD tests and wanted to ask whether I should use SWTBot or the PDE test framework for this. I noticed that SWTBot isn’t fully compatible with Eclipse 4.37, so I’d like to confirm which approach you’d recommend.

HeikoKlare:
What do you mean with the PDE test framework? I don't think we currently have either a dependency to SWTBot or any PDE artifacts in Platform UI. Can you explain why you need either of those?

aloowood:
I'm trying to write DnD tests, but I'm not entirely sure which approach is preferred in Platform UI. What method would you recommend for this? I want to be sure I'm following the correct approach.

PR #3196, opened 2025-08-17
Add Split Editor shortcuts to editor tab context menu

### LINKED ISSUE #3166: Add "Split Editor" (Vertical & Horizontal) options to the "Close & Detach" menu section of the editor
it would be really helpful to have `Split Editor ( Vertical & Horizontal )` options included in `Close & Detach `options from the editor tab

Image

<br>
Image
<br>
<br>
Including these options here will improve the discoverability of the split editor feature, especially for new users.

SougandhS:
This PR introduces context menu entries for the Split Editor feature (Horizontal & Vertical) in the editor tab’s menu. Making it easier for new users to access.

image

<br>

https://github.com/user-attachments/assets/e8ece2c8-36af-4477-b248-0da985e29a66

Fixes : https://github.com/eclipse-platform/eclipse.platform.ui/issues/3166

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
@@ -34,7 +34,10 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
+import org.eclipse.core.commands.ParameterizedCommand;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.e4.core.commands.ECommandService;
+import org.eclipse.e4.core.commands.EHandlerService;
 import org.eclipse.e4.core.contexts.IEclipseContext;
 import org.eclipse.e4.core.di.annotations.Optional;
 import org.eclipse.e4.core.di.extensions.Preference;
@@ -1535,8 +1538,14 @@ public class StackRenderer extends LazyStackRenderer {
 				new MenuItem(menu, SWT.SEPARATOR);
 
 				createMenuItem(menu, SWTRenderersMessages.menuCloseAll, e -> closeSiblingParts(menu, false));
+
+				new MenuItem(menu, SWT.SEPARATOR);
+
 			}
 		}
+
+		createMenuItem(menu, SWTRenderersMessages.splitEditorH, e -> splitEditor(part.getContext(), true));
+		createMenuItem(menu, SWTRenderersMessages.splitEditorV, e -> splitEditor(part.getContext(), false));
 
 		if (isDetachable(part)) {
 			if (closeableElements > 0) {
@@ -1952,5 +1961,13 @@ public class StackRenderer extends LazyStackRenderer {
 		}
 		return Arrays.stream(parent.getChildren()).filter(child -> id.equals(child.getData(ID)))
 				.filter(type::isInstance).map(type::cast).findFirst().orElse(null);
+	}
+
+	private void splitEditor(IEclipseContext ctx, boolean isHorizontal) {
+		ECommandService commandService = ctx.get(ECommandService.class);
+		EHandlerService handlerService = ctx.get(EHandlerService.class);
+		Map<String, String> param = Map.of("Splitter.isHorizontal", String.valueOf(isHorizontal)); //$NON-NLS-1$
+		ParameterizedCommand command = commandService.createCommand("org.eclipse.ui.window.splitEditor", param); //$NON-NLS-1$
+		handlerService.executeHandler(command);
 	}
 }
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/SWTRenderersMessages.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/SWTRenderersMessages.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2015 IBM Corporation and others.
+ * Copyright (c) 2010, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -37,6 +37,10 @@ public class SWTRenderersMessages extends NLS {
 	public static String menuDetach;
 
 	public static String viewMenu;
+
+	public static String splitEditorH;
+
+	public static String splitEditorV;
 
 	static {
 		// load message values from bundle file
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/messages.properties
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/messages.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-#  Copyright (c) 2010, 2015 IBM Corporation and others.
+#  Copyright (c) 2010, 2025 IBM Corporation and others.
 #
 #  This program and the accompanying materials
 #  are made available under the terms of the Eclipse Public License 2.0
@@ -27,3 +27,5 @@ menuCloseRight = Close Tabs to the &Right
 menuCloseLeft = Close Tabs to the &Left
 menuDetach= &Detach
 viewMenu = View Menu
+splitEditorH = Split Editor in Horizontal
+splitEditorV = Split Editor in Vertical


SougandhS:
Hi @fedejeanne 
Could you please check this when you are available ?

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -1953,4 +1963,18 @@ private <T extends Control> T getChild(Composite parent, String id, Class<T> typ
 		return Arrays.stream(parent.getChildren()).filter(child -> id.equals(child.getData(ID)))
 				.filter(type::isInstance).map(type::cast).findFirst().orElse(null);
 	}
+
+
+	private void splitEditor(IEclipseContext ctx, String isVertical) {
+		Display.getDefault().asyncExec(() -> splitCurrentEditor(ctx, isVertical));
```
merks: This doesn't seem like the proper way to get the display.  The method org.eclipse.ui.IWorkbench.getDisplay() wouldn't exist if this was the proper way.  And I'm not suggesting to use that method because here the caller knows the display.  But in the end, is the asyncExec actually necessary?  The other menu actions don't do that.

merks (changes requested):

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
SougandhS: >This doesn't seem like the proper way to get the display

I initially thought `Display.getCurrent()` would be okay here

>But in the end, is the asyncExec actually necessary?

I used `asyncExec` because the command implementation lives in another bundle. Adding the asyncExec was a safeguard to ensure that the invocation happens safely on the UI thread. I’ll remove the asyncExec here and align the implementation with how the other menu actions are handled.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -1953,4 +1963,13 @@ private <T extends Control> T getChild(Composite parent, String id, Class<T> typ
 		return Arrays.stream(parent.getChildren()).filter(child -> id.equals(child.getData(ID)))
 				.filter(type::isInstance).map(type::cast).findFirst().orElse(null);
 	}
+
+	private void splitEditor(IEclipseContext ctx, String isVertical) {
```
merks: I think it would look a little less "strange" to make `isVertical` a boolean and convert it to a String when putting it in the map.

merks (changes requested):
Sorry, I should have remarked on this earlier.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
SougandhS: Sure, I'll update it

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -1535,9 +1539,15 @@ protected void populateTabMenu(final Menu menu, MPart part) {
 				new MenuItem(menu, SWT.SEPARATOR);
 
 				createMenuItem(menu, SWTRenderersMessages.menuCloseAll, e -> closeSiblingParts(menu, false));
+
+				new MenuItem(menu, SWT.SEPARATOR);
+
 			}
 		}
 
+		createMenuItem(menu, SWTRenderersMessages.splitEditorH, e -> splitEditor(part.getContext(), true));
```
merks, line 1547: Oh dear, I keep noticing things... 

This passes true to a parameter called `isVertical` but, according to the message, this is for doing a horizontal split.  So should the parameter be called `isHorizonal`?

SougandhS: My bad, updated it now 👍

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -1953,4 +1963,13 @@ private <T extends Control> T getChild(Composite parent, String id, Class<T> typ
 		return Arrays.stream(parent.getChildren()).filter(child -> id.equals(child.getData(ID)))
 				.filter(type::isInstance).map(type::cast).findFirst().orElse(null);
 	}
+
+	private void splitEditor(IEclipseContext ctx, boolean isHorizontal) {
+		ECommandService commandService = ctx.get(ECommandService.class);
+		EHandlerService handlerService = ctx.get(EHandlerService.class);
+		Map<String, String> param = new HashMap<>();
+		param.put("Splitter.isHorizontal", String.valueOf(isHorizontal)); //$NON-NLS-1$
+		ParameterizedCommand command = commandService.createCommand("org.eclipse.ui.window.splitEditor", param); //$NON-NLS-1$
```
merks: I'm a horrible person.  😢  

`Map.of` would work well here I think?

SougandhS: `Map.of` would be great! Thanks :D

merks (approved):
Looks good now.

merks:
Test failure unrelated.

SougandhS:
Thank you @merks

laeubi:
This does not feel to be properly integrated

1. Why is this added to a general purpose menu for Tab handling instead of tho the regular popup menu we already have, having this on the tab really looks wrong compared to other functionality there.
2. Why don't we use proper E4 model to register / call the menu and actions

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -1953,4 +1962,12 @@ private <T extends Control> T getChild(Composite parent, String id, Class<T> typ
 		return Arrays.stream(parent.getChildren()).filter(child -> id.equals(child.getData(ID)))
 				.filter(type::isInstance).map(type::cast).findFirst().orElse(null);
 	}
+
+	private void splitEditor(IEclipseContext ctx, boolean isHorizontal) {
+		ECommandService commandService = ctx.get(ECommandService.class);
+		EHandlerService handlerService = ctx.get(EHandlerService.class);
+		Map<String, String> param = Map.of("Splitter.isHorizontal", String.valueOf(isHorizontal)); //$NON-NLS-1$
+		ParameterizedCommand command = commandService.createCommand("org.eclipse.ui.window.splitEditor", param); //$NON-NLS-1$
```
laeubi, line 1970: This command is only aviable in E3 Applications!

SougandhS:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/3166#issuecomment-3200865800
I guess we should revert this change and add these options on right click options of editor part itself?

laeubi:
> [#3166 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/issues/3166#issuecomment-3200865800) I guess we should revert this change and add these options on right click options of editor part itself?

Yes right click menu sounds resonable to me and much mor flexible and matching users exspecations.

SougandhS:
Reverting this change - https://github.com/eclipse-platform/eclipse.platform.ui/pull/3204

merks:
A number of things are "hidden" in the Window -> Editor menu.   I wonder if that entire menu item's shouldn't be on the context menu of an editor...

laeubi:
> I wonder if that entire menu item's shouldn't be on the context menu of an editor...

Yes thats what I would suggest. I must confess that in > 10 years of Eclipse I never noticed that menu even though it contains useful stuff (e.g. toggle whitespaces). The toggle whitespace is even shown already but _only for compare editors_. Other like the zooming feature not (it shows me the shortcut is `STRG + +` but at laest under Linux/Gtk3 it seem not working to use the shortcut but the menu item works).

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
SougandhS: Hi @laeubi @iloveeclipse 
May I know its equivalent E4 command ?
and would it be still possible to implement this feature in editor tab itself with this workaround (hack)  + corresponding E4 command ?
image

:)

laeubi: I think it has to be implemented in E4 first where such feature would need to be implemented independent of editors (not sure it can work there anyways), I'm even not sure it makes sense outside any non text based editor it e.g. if I use it for the manifest editor I get the Editor split but it does not looks useful.

In any case I don't think the is suitable in the Editor tab especially if it requires hacks, things like requiring a name in an ID is really not how it is supposed to work.

SougandhS: Thanks for the info 👍

PR #3197, opened 2025-08-18
Show split shortcuts only for editor tabs

SougandhS:
This pr will make sure split shortcuts are shown only to editor tabs

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
@@ -51,6 +51,7 @@ import org.eclipse.e4.ui.internal.workbench.renderers.swt.BasicPartList;
 import org.eclipse.e4.ui.internal.workbench.renderers.swt.SWTRenderersMessages;
 import org.eclipse.e4.ui.internal.workbench.swt.AbstractPartRenderer;
 import org.eclipse.e4.ui.internal.workbench.swt.CSSConstants;
+import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.e4.ui.model.application.ui.MDirtyable;
 import org.eclipse.e4.ui.model.application.ui.MElementContainer;
 import org.eclipse.e4.ui.model.application.ui.MUIElement;
@@ -1543,9 +1544,12 @@ public class StackRenderer extends LazyStackRenderer {
 
 			}
 		}
+		String partId = part.getElementId();
 
-		createMenuItem(menu, SWTRenderersMessages.splitEditorH, e -> splitEditor(part.getContext(), true));
-		createMenuItem(menu, SWTRenderersMessages.splitEditorV, e -> splitEditor(part.getContext(), false));
+		if (!isDetached(part) && partId != null && (partId.contains("editor") || partId.contains("Split"))) { //$NON-NLS-1$ //$NON-NLS-2$
+			createMenuItem(menu, SWTRenderersMessages.splitEditorH, e -> splitEditor(part.getContext(), true));
+			createMenuItem(menu, SWTRenderersMessages.splitEditorV, e -> splitEditor(part.getContext(), false));
+		}
 
 		if (isDetachable(part)) {
 			if (closeableElements > 0) {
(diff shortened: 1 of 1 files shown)


SougandhS:
Hi @merks 
There was an issue with https://github.com/eclipse-platform/eclipse.platform.ui/pull/3196 , The split options were shown on every tabs 
eg. 
image

With this change it will only show in editor tabs

image


could you please check this pr,

merks:
One thing I noticed while testing is I wondered how one can un-split the editor.  It's much more obvious on the menubar action that this is a toggle action:

image

And I don't think the URI tests are valid, e.g., I see URIs like this:
```
bundleclass://org.eclipse.ui.workbench/org.eclipse.ui.internal.e4.compatibility.CompatibilityEditor
bundleclass://org.eclipse.e4.ui.workbench.addons.swt/org.eclipse.e4.ui.workbench.addons.splitteraddon.SplitHost
```
I'm not sure what the proper test would be, but please test with an editor that's already split to unsplit it and also test with an editor that's detatched.

SougandhS:
> test with an editor that's detatched.

Will check this scenario

SougandhS:
Handled the edge cases

https://github.com/user-attachments/assets/143f7b8a-8025-4ead-b287-2181128a55e3

SougandhS:
Missed to add one case's video 
Detached editor with split 

https://github.com/user-attachments/assets/0e2ca031-b571-4947-a504-f4f5c38f3c79

merks:
Somehow it feels like a hack.  I see that org.eclipse.e4.core.commands.EHandlerService.canExecute(ParameterizedCommand) exists so I wondered if we could create an command and test if it's executable but that doesn't really work properly either.  Maybe @laeubi  or @vogella has an idea for a more appropriate test for whether we are on an editor stack?

Maybe better to look at the tags as is done in other places:

image

iloveeclipse:
> Maybe better to look at the tags as is done in other places

Correct. Not sure if there is already something like utility method somewhere?

SougandhS:
> Maybe better to look at the tags as is done in other places:

for split editors there are no appropriate tags

Normal editor
image

Split editor
image

iloveeclipse:
> for split editors there are no appropriate tags

Could you create a bug? I don't think it is intentional.

SougandhS:
> Could you create a bug? I don't think it is intentional.

Sure 👍

laeubi:
Please note that E4 do no longer has a distinction about "Editors" and "Parts" (editors are just parts) and I don't think we should introduce that again. 

So one should ask what a "split" is only usefull for some parts and what makes these parts special, e.g. do the show a generic editor object? Or should it be just a TextViewer or ...

iloveeclipse:
> Please note that E4 do no longer has a distinction about "Editors" and "Parts" (editors are just parts) and I don't think we should introduce that again.

This is a different dimension. e4 is about presentation in UI, editors vs views is about handling content differently. There are editors contributed to editors extension point and there are views contributed to a different extension point. "Split editor" makes sense for most editors, but not for views. So here one wants to know not what kimd of e4 part is it, but whether it is editor of a content or a view on it. And of course if it is easy to know while a part is created, the parts tagged differently to simplify our life.

laeubi:
@SougandhS @iloveeclipse before proceeding on this, I think the general design falw should be fixed as described here:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/3166#issuecomment-3200865800

the tab menu is really not the place for "usefull" things, this is completely wrong and this problem alread steem from the fact that it was done wrong here.

These menus already have proper ways to handling things depending on the right context! This will be a major regression for pure E4 apps, one should never assume that everyon is using the compatibilty laber and all apps are "IDE" like ones.

laeubi:
> This is a different dimension. e4 is about presentation in UI, editors vs views is about handling content differently. There are editors contributed to editors extension point and there are views contributed to a different extension point. "Split editor" makes sense for most editors, but not for views.

This is not correct. Yes E4 is about presentation of a model, but editor is an E3 concept as well as the usage of extension points! There is a compatibilty later that maps some concepts of E3 extesions into the E4 model but a StackRender is not the right place for adding arbitary commands to renderer tabs, these has to use the E4 model. And if we want such thing we need an E4 model extension (that then might be used by the compatibility layer maybe).

`org.eclipse.ui.window.splitEditor` is even (as by its name) only aviable in E3 so this is guranteed to fail in plain E4 RCP applications! Because of that E4 allows to enable/disable content already by the context, so we would not need any special handling here.

PR #3201, opened 2025-08-18
Replace navigator tests hierarchy to UITestCase with test rule

HeikoKlare:


HeikoKlare:

PR #3204, opened 2025-08-19
Revert "Add Split Editor shortcuts to editor tab context menu"

SougandhS:
This reverts commit a7dc4866df7a134d650fc7d8fe2b64e871b73710 to avoid regressions in E4 applications.
Based on : https://github.com/eclipse-platform/eclipse.platform.ui/pull/3197#issuecomment-3200930424


laeubi:
@merks I noticed this was not merged but I think it is important to not be missing in this release can you please take a look here?

merks:
@akurtakov 

I'll make an executive decision to revert his because it's showing up in many places it should not.

PR #3206, opened 2025-08-20
Replace UITestCase with test rule in UI API tests

HeikoKlare:


HeikoKlare:

PR #3208, opened 2025-08-20
Replace UITestCase with test rule in dynamicplugins UI tests

HeikoKlare:


HeikoKlare:

PR #3211, opened 2025-08-21
Make genericeditor CompletionTest code more direct

akurtakov:
Another commit towards fixing
https://github.com/eclipse-platform/eclipse.platform.ui/issues/3155 . Simplify code, more direct calls, less casts, stronger DisplayHelper conditions to ensure desired state is achieved before expecting.


akurtakov:
Tests change only which should reduce(or at least help figure out) failing tests in I-builds thus merging despite freeze.

Issue #2708, opened 2025-01-15
Random failing org.eclipse.jface.text.tests.reconciler.FastAbstractReconcilerTest#testReplacingDocumentWhenClean()


HeikoKlare:
`org.eclipse.jface.text.tests.reconciler.FastAbstractReconcilerTest#testReplacingDocumentWhenClean()` failed in a recent build (on MacOS): https://github.com/eclipse-platform/eclipse.platform.ui/pull/2704/checks?check_run_id=35665179984

```
reconciler never ran in 5 seconds
java.lang.AssertionError: reconciler never ran in 5 seconds
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.jface.text.tests.reconciler.AbstractReconcilerTest$Barrier.await(AbstractReconcilerTest.java:103)
	at org.eclipse.jface.text.tests.reconciler.AbstractReconcilerTest.testReplacingDocumentWhenClean(AbstractReconcilerTest.java:337)
```

vogella:
Also on Linux https://github.com/eclipse-platform/eclipse.platform.ui/pull/3453

1 out of 3 runs failed: testReplacingDocumentWhenClean (JFaceTextTestSuite FastAbstractReconcilerTest)

expected:<[reconcilerReset]> but was:<[process]>
org.junit.ComparisonFailure: expected:<[reconcilerReset]> but was:<[process]>
	at org.eclipse.jface.text.tests.reconciler.AbstractReconcilerTest.testReplacingDocumentWhenClean(AbstractReconcilerTest.java:336)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)

PR #3215, opened 2025-08-23
Re-enable test cases and move ignore from test suites to test classes

HeikoKlare:
Some test suite of org.eclipse.ui.tests are currently disabled/ignored. This will be impossible when migrating to JUnit 5. This change pushes down the ignore to either the included test classes or just some of the test methods of other methods of those test classes work fine. Thus, this also re-enables some tests that actually run fine.

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dynamicplugins/AcceleratorConfigurationsExtensionDynamicTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/dynamicplugins/AcceleratorConfigurationsExtensionDynamicTest.java
@@ -20,6 +20,7 @@ import org.eclipse.core.commands.common.NamedHandleObject;
 import org.eclipse.core.commands.common.NotDefinedException;
 import org.eclipse.ui.internal.registry.IWorkbenchRegistryConstants;
 import org.eclipse.ui.keys.IBindingService;
+import org.junit.Ignore;
 import org.junit.Test;
 
 /**
@@ -28,6 +29,7 @@ import org.junit.Test;
  *
  * @since 3.1.1
  */
+@Ignore("Bug 405296")
 public final class AcceleratorConfigurationsExtensionDynamicTest extends
 		DynamicTestCase {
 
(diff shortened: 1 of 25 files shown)


vogella:
I think the zoon tests can be deleted, AFAIK this functionality does not exists anymore (or never did, not sure).

HeikoKlare:
Thank you for the hint! If appropriate, just deleting the tests would of course be the cleanest approach.

I see that some zoom functionality is currently not implemented, e.g.:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/d0e2112769871aae4686b832e1573bacb92f94e1/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPage.java#L4351-L4354

So would it make sense to deprecate that functionality?

But there is also some functionality such as `toggleZoom` ...
https://github.com/eclipse-platform/eclipse.platform.ui/blob/d0e2112769871aae4686b832e1573bacb92f94e1/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPage.java#L4356-L4366

... that seems to be covered by these tests, which is actually functional to maximize parts:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/d0e2112769871aae4686b832e1573bacb92f94e1/tests/org.eclipse.ui.tests/Eclipse%20UI%20Tests/org/eclipse/ui/tests/zoom/ZoomTestCase.java#L105-L114

When I execute the tests, a bunch of them even succeeds (but I am not sure if they test something useful):
image

Is it still okay to remove them? Since the tests have never been executed, probably no one will miss them.

vogella:
> So would it make sense to deprecate that functionality?

Yes

> ... that seems to be covered by these tests, which is actually functional to maximize parts:

In this case I would leave them in.

HeikoKlare:
> > So would it make sense to deprecate that functionality?
> 
> Yes

Deprecation via:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3225

> > ... that seems to be covered by these tests, which is actually functional to maximize parts:
> 
> In this case I would leave them in.

Reactivation via:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3226

This PR will be adapted on top once the latter is merged.

HeikoKlare:
Failing test is known to be flaky:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2708

PR #3216, opened 2025-08-23
Migrate UI performance tests to JUnit 4

HeikoKlare:
- Replace UIPerformanceTestSetup with UIPerformanceTestRule, which also makes each test class executable individually
- Remove FilteredTestSuite that became obsolete because of the above
- Remove JUnit 3 inheritance hierarchy via UITestCase and PerformanceTestCase (replaced with PerformanceTestCaseJunit4)
- Remove PerformanceTester that became obsolete because of the above
- Annotate test methods for JUnit 4
- Replace test suites with JUnit 4


HeikoKlare:
Failing test is unrelated to the tests touched here and documented as randomly failing: https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #3217, opened 2025-08-24
Remove setUp/tearDown overwrites in UI performance tests

HeikoKlare:
Define separate `@Before`/`@After` methods instead.


vogella:
If the performance tests do not run too long, we could add them to the regular verification run.

HeikoKlare:
> If the performance tests do not run too long, we could add them to the regular verification run.

They run for several minutes, so for the nightly builds that might be okay but for the GH actions runs for every PR it should, if at all, probably be a separate workflow.
However, I am not sure if those tests have real value without storing the results and comparing them across builds.

HeikoKlare:
To have some reference value, I ran the tests again: on my system they take around 12 minutes
image

PR #3218, opened 2025-08-24
Centralize test project access for UI performance tests in test rule

HeikoKlare:
Currently, the test project access in projects is scattered across multiple places. This centralizes the access at the UIPerformanceTestRule, which is responsible with creating and deleting the project anyway.


HeikoKlare:
Failing test is unrelated to the tests affected here and is known to be randomly failing: https://github.com/eclipse-platform/eclipse.platform.ui/issues/294

PR #3220, opened 2025-08-24
Replace obsolete BasicPerformanceTest superclass with utility class

HeikoKlare:
The BasicPerformanceTest only serves as a utility class for its subclasses. This change removes the class from the inheritance hierarchy and makes a propery utility class out of it.

This prepares for a possible subsequent migration to JUnit 5.


HeikoKlare:
The class can be safely removed/replaced as its bundle does not export any package and I've checked that there is also no consumer in the SDK that accesses internals of that bundle.

PR #3221, opened 2025-08-26
Reopen tab works for Editor Tabs and View tabs respectively.

### LINKED ISSUE #2148: Reopen closed tab

## Suggestion

I would like to be able to reopen a closed tab in Eclipse View. You might know this feature already from Browser like Firefox or Edge. If you right click in the tab-bar of the browser you have the following options:

grafik

I would love to have something similar in ADT. I think it would provide a better usability.

grafik

## Community

deepika-u:
Provide "Reopen Closed Tab" option for the user.

Fixes #2148 

There are basically 2 types of tabs in eclipse.

1. Editor tabs.
2. View tabs.

I have attached step by step screen shots for the implementation where this "Reopen Closed Tab" is visible/shown

- When Editor tabs are opened and say atleast 1 tab is closed, then there appears the "Reopen Closed Tab" on a right click of any existing opened tab like in screenshot 1.

Screenshot1

Note : When there is no tab yet closed, this menu option doesnt show up.

- What are all called as View kind of tabs(left, bottom, right set of tabs all considered as view kind of tabs). Rest all contribute to View kind of tabs and same kind of menu shows up as shown in the screenshot.

Screenshot2_views

- When console is 1 such view which is closed now, hence the menu option "Reopen Closed Tab" shows up in the right click menu.

Screenshot3_console_closed

- When the console view is opened, the menu option  "Reopen Closed Tab"  again doesnt show up.

Screenshot_reopen_disabled

Observations 
1. Maximum of only 1 tab can be reopened at a time.
2. Maximum of 15 times the option "Reopen Closed Tab" will be shown. Which means it can right now only remember most recently closed 15 tabs though you would have closed more.
3. If anyone uses the option "Detach" and close a particular tab, then this tab is closed and that specific tab cannot be reopened using the option "Reopen Closed Tab". It needs to be opened freshly like any other tab.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
@@ -24,8 +24,13 @@ import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
 import jakarta.inject.Inject;
 import jakarta.inject.Named;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -48,6 +53,7 @@ import org.eclipse.e4.ui.internal.workbench.renderers.swt.BasicPartList;
 import org.eclipse.e4.ui.internal.workbench.renderers.swt.SWTRenderersMessages;
 import org.eclipse.e4.ui.internal.workbench.swt.AbstractPartRenderer;
 import org.eclipse.e4.ui.internal.workbench.swt.CSSConstants;
+import org.eclipse.e4.ui.model.application.MContribution;
 import org.eclipse.e4.ui.model.application.ui.MDirtyable;
 import org.eclipse.e4.ui.model.application.ui.MElementContainer;
 import org.eclipse.e4.ui.model.application.ui.MUIElement;
@@ -69,6 +75,7 @@ import org.eclipse.e4.ui.workbench.UIEvents;
 import org.eclipse.e4.ui.workbench.UIEvents.EventTags;
 import org.eclipse.e4.ui.workbench.modeling.EModelService;
 import org.eclipse.e4.ui.workbench.modeling.EPartService;
+import org.eclipse.e4.ui.workbench.modeling.EPartService.PartState;
 import org.eclipse.e4.ui.workbench.modeling.ISaveHandler;
 import org.eclipse.jface.action.IContributionItem;
 import org.eclipse.jface.action.LegacyActionTools;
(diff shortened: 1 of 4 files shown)


laeubi:
> There are basically 2 types of tabs in eclipse.

I don't think this is true. "Editor" is a concept of E3 / IDE.

A Tab is actually a view in the `PartStack` and we have a so called "primary stack" where we show things that are used for editing files (what itself are just views).

So I'm not 100% sure from the description what you want to change, but as the `PartStack` is hidden when the last editor is closed, if one wants to restore tabs it must attach a menu to the `Area`

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -130,6 +137,10 @@ public class StackRenderer extends LazyStackRenderer {
 	private static final String ONBOARDING_COMPOSITE = "EditorStack.OnboardingComposite"; //$NON-NLS-1$
 	private static final String ONBOARDING_IMAGE = "EditorStack.OnboardingImage"; //$NON-NLS-1$
 	private static final String ONBOARDING_TEXT = "EditorStack.OnboardingText"; //$NON-NLS-1$
+	public static final String TAB_UNIQUE_PATH = "editor.uniquePath"; //$NON-NLS-1$
+	public static final String VIEW_UNIQUE_ID = "view.uniqueID"; //$NON-NLS-1$
+	public static final String EDITOR_TAB = "CompatibilityEditor"; //$NON-NLS-1$
```
laeubi, line 142: An editor must not be a compatibility  view/editor. this is only when you have E3 editors and using the compatibility layer.

laeubi (changes requested):
I see several problems here:

1. The feature assumes the E4 Compatibility layer is used
2. It depends in some hard coded strings to be present in IDs and names
3. It retains Parts and Context objects possibly forever what will be a source of memory leak that can only be avoided by a restart of Eclipse

Instead I think 

- the distinction between "Editor" and "View" is not needed. The context here is the container (the `PartStack`) so I would probably want to restore a tab for one stack but it does not matter what is in this stack.
- if this should be enabled always, we need a tag to disable this feature for a partstack so that users have an option to opt-out. As an alternative one would maybe want a way to opt-in for the feature using a tag
- The context and parts should not be retained forever. There are already ways to persist and restore state between restarts of Eclipse and we should use these technique here as well so that not objects are retained forever.
- There should be a preference for the user to configure the number of tabs that can be restore

deepika-u:
> > There are basically 2 types of tabs in eclipse.
> 
> I don't think this is true. "Editor" is a concept of E3 / IDE.
> 
> A Tab is actually a view in the `PartStack` and we have a so called "primary stack" where we show things that are used for editing files (what itself are just views).
> 
> So I'm not 100% sure from the description what you want to change, but as the `PartStack` is hidden when the last editor is closed, if one wants to restore tabs it must attach a menu to the `Area`

Thanks for the detailing on E3 and E4 differentiation which i was not aware of. I was referring to the E3 kind of terminology itself like 
- Editors: for working on resources (files, etc.).
- Views: for auxiliary tools (outline, console, problems view, etc.).

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
deepika-u: > An editor must not be a compatibility view/editor. this is only when you have E3 editors and using the compatibility layer.

You’re right — strictly speaking, editors only exist in the E3 compatibility layer. But in practice, this was the only reliable way I found to distinguish editor-like parts from views. Since users interact with editor tabs much more frequently than view tabs, I felt it made sense to treat them differently rather than having a single generic reopen option clubbing both.

PR #3225, opened 2025-08-27
Deprecate IWorkbenchPage#zoomOut() for removal

HeikoKlare:
The method IWorkbenchPage#zoomOut() has never been implemented in the sole interface implementation WorkbenchPage. All existing consumers rely on the #toogleZoom() implementation. This change thus deprecates the method for removal, as it has never been implemented and required and will unlikely ever get implemented.

The related functionality of WorkbenchPage#unzoomAllPerspective() is also unused, and since it's internal this change just removes it.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/3215#issuecomment-3218305271 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/3215#issuecomment-3224348852.


HeikoKlare:
@vogella can you please verify if this deprecation is reasonable according to what we discussed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3215#issuecomment-3224348852?

vogella (approved):

PR #3228, opened 2025-08-27
Fix the window size in the app model when the window is maximized

### LINKED ISSUE #3227: Window size is overwritten when maximized [Regression]
The window size saved in the model is overwritten when the user maximizes the window.

This is a regression introduced by #3055

The change wrongly removed the return after it sets the ShellMaximizedTag.

## Steps to reproduce

 - Resize the window
 - Maximize the window
 - Restart -> the size is lost

Version: 2025-09 (4.37)
Build id: I20250827-0600
Windows 11

wahlbrink:
Fixes: #3227

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/WBWRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/WBWRenderer.java
@@ -537,9 +537,9 @@ public class WBWRenderer extends SWTPartRenderer {
 					// But set the maximized tag so that the user can access the current state
 					if (shell.getMaximized()) {
 						me.getTags().add(ShellMaximizedTag);
-					} else {
-						me.getTags().remove(ShellMaximizedTag);
+						return;
 					}
+					me.getTags().remove(ShellMaximizedTag);
 
 					try {
 						ignoreSizeChanges = true;


laeubi (approved):
Good catch, looks reasonable to me!

merks (approved):
Yes, it works better if the maximized size is never recorded.  It makes no sense to record the maximized size; I might restart without my two monitors attached on a differently sized monitor than the one on which the window was maximized.

akurtakov:
Can a test be added for this one?

laeubi:
> Can a test be added for this one?

Its maybe possible but not trivial as it requires a workbench restart, triggering size changes of shell and reading actual shell sizes (that might vary between OSes).

An alternative would be to check the model for changes but still requires triggering a maximize/restore events.

iloveeclipse:
What about https://github.com/eclipse-platform/eclipse.platform.ui/issues/3227#issuecomment-3232571288 ?
Doesn't look right to me?

laeubi:
@iloveeclipse given there is different behavior and fixing it now becomes non trivial (as described in https://github.com/eclipse-platform/eclipse.platform.ui/issues/3229) I would more lean to completely revert 
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3055

as it seems it was not properly tested on all platforms and would need a test given it seems to trivial to get it right across platforms.

@wahlbrink as you already suggested that, would you like to prepare a revert of the problematic commit?

merks:
> What about [#3227 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/issues/3227#issuecomment-3232571288) ? Doesn't look right to me?

I think there are multiple problems.  The behavior without the fix is definitely much worse (not maximized and not the expected size) than the behavior with the fix on Windows (not maximized but at the restored location and at the restored size)...

iloveeclipse:
It seems that touching this code opens a can of worms and RC2 is not the right time to open this can.
I vote for simple revert of https://github.com/eclipse-platform/eclipse.platform.ui/commit/ebfb7ec36cd3a6b9dc4b208f6824d639fcbacade

merks:
> It seems that touching this code opens a can of worms and RC2 is not the right time to open this can. I vote for simple revert of [ebfb7ec](https://github.com/eclipse-platform/eclipse.platform.ui/commit/ebfb7ec36cd3a6b9dc4b208f6824d639fcbacade)

Yes, I think we are all slowly getting to this same conclusion.

iloveeclipse:
Superceeded by https://github.com/eclipse-platform/eclipse.platform.ui/pull/3231

PR #3230, opened 2025-08-28
Fix caret of raw insert mode of text editor on HiDPI screens

### LINKED ISSUE #314: Ctrl+Shift+Insert in Eclipse key bindings doesn't show insert caret when monitor zoom ≠ 100%
**Description**
Mirror ticket for https://github.com/eclipse-platform/eclipse.platform.ui/issues/3010

### LINKED ISSUE #3010: Ctrl+Shift+Insert in Eclipse key bindings doesn't show insert caret when monitor zoom ≠ 100%

## Steps to reproduce

Toggling insert mode using Ctrl+Shift+Insert in Eclipse key bindings (this is binded to toggle insert mode) doesn’t show the insert cursor (caret). However when using the regular Insert key, the cursor appears as expected since it switches to Smart Insert mode [code link](https://github.com/vi-eclipse/eclipse.platform.ui/blob/bug/arun/199/rangeIndicator/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java#L6609). But Ctrl+Shift+Insert seems to activate a different insert mode where the cursor is currently missing. At 100% zoom, the caret behaves normally with both Insert and Ctrl+Shift+Insert.

* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): Nightly builds from 21/5/2025

## Community

wahlbrink:
Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3010

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -67,6 +67,7 @@ import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
+import org.eclipse.swt.graphics.ImageDataProvider;
 import org.eclipse.swt.graphics.PaletteData;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.RGB;
@@ -6961,32 +6962,27 @@ public abstract class AbstractTextEditor extends EditorPart
 	}
 
 	private Image createRawInsertModeCaretImage(StyledText styledText) {
-
 		PaletteData caretPalette = new PaletteData(new RGB(0, 0, 0), new RGB(255, 255, 255));
 		int width = getCaretWidthPreference();
-		int widthOffset = width - 1;
-
 		// XXX: Filed request to get a caret with auto-height:
 		// https://bugs.eclipse.org/bugs/show_bug.cgi?id=118612
-		ImageData imageData = new ImageData(4 + widthOffset, styledText.getLineHeight(), 1, caretPalette);
-
-		Display display = styledText.getDisplay();
-		Image bracketImage = new Image(display, imageData);
-		GC gc = new GC(bracketImage);
-		gc.setForeground(display.getSystemColor(SWT.COLOR_WHITE));
-		gc.setLineWidth(0); // NOTE: 0 means width is 1 but with optimized performance
-		int height = imageData.height / 3;
-		// gap between two bars of one third of the height
-		// draw boxes using lines as drawing a line of a certain width produces
-		// rounded corners.
-		for (int i = 0; i < width; i++) {
-			gc.drawLine(i, 0, i, height - 1);
-			gc.drawLine(i, imageData.height - height, i, imageData.height - 1);
-		}
-
-		gc.dispose();
-
-		return bracketImage;
+		int height = styledText.getLineHeight();
+		ImageDataProvider imageDataProvider = zoom -> {
+			double scaleFactor = zoom / 100.0;
+			int w = (int) (width * scaleFactor + 0.5);
+			int h = (int) (height * scaleFactor + 0.5);
+			ImageData imageData = new ImageData(w, h, 1, caretPalette);
+			// gap between two bars of one third of the height
+			int h3 = h / 3;
+			for (int x = 0; x < w; x++) {
+				for (int y = 0; y < h3; y++) {
+					imageData.setPixel(x, y, 1);
+					imageData.setPixel(x, h - y - 1, 1);
+				}
+			}
+			return imageData;
+		};
+		return new Image(styledText.getDisplay(), imageDataProvider);
 	}
 
 	private Caret createRawInsertModeCaret(StyledText styledText) {


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -6537,31 +6538,30 @@ private Caret createInsertCaret(StyledText styledText) {
...
-		for (int i= 0; i < width ; i++) {
-			gc.drawLine(i, 0, i, height - 1);
-			gc.drawLine(i, imageData.height - height, i, imageData.height - 1);
-		}
-
-		gc.dispose();
-
-		return bracketImage;
+		int height= styledText.getLineHeight();
+		return new Image(styledText.getDisplay(), new ImageDataProvider() {
```
HeikoKlare: Is there any reason to use an `ImageDataProvider` rather than an `ImageGcDrawer`?
This requires to manually implement proper line width calculation and line rendering instead of just using `gc.drawLine()` like in the original code.

HeikoKlare (commented):
Thank you for proposing this fix! I have one question regarding the proposed solution, as it appears more complex to me than required.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
wahlbrink: This creates a 1-bit image as before.

HeikoKlare: What do you mean with "1-bit image"? Do you mean that the line is 1 pixel wide? That's the intended behavior of `setLineWidth(0)`. Changing that to `setLineWidth(1)` should resolve it.

wahlbrink: 1-bit image =  image with 1-bit color palette = a pixels uses 1-bit = colloquially called black-and-white image

HeikoKlare: I see. But is it really necessary/beneficial to keep it like that?

arunjose696: I also experimented with using ImageGcDrawer. Since the image created this way is no longer 1-bit, I had to adapt the GC drawing slightly to get the desired result. The following variant worked reliably for me:

```
Display display = styledText.getDisplay();
		ImageGcDrawer imageGcDrawer = (gc, imageWidth, imageHeight) -> {
			gc.setForeground(display.getSystemColor(SWT.COLOR_BLACK));
			gc.setLineWidth(1);
			int h3 = imageHeight / 3;
			for (int i = 0; i < imageWidth; i++) {
				gc.drawLine(i, imageHeight - h3, i, imageHeight - 2 * h3);
			}
		};
		final Image image = new Image(display, imageGcDrawer, width, styledText.getLineHeight());
```

I don’t have a strong opinion on ImageDataProvider vs ImageGcDrawer. From my perspective, using ImageGcDrawer allows relying on gc.drawLine(), while the ImageDataProvider approach also seems to work fine.

@wahlbrink do you see any reason why ImageDataProvider might be the better option here, or any trade-offs between the two that we should consider?

arunjose696:
I’ve tested the changes in both dark and light themes, and they work as expected. and I can see the caret at multiple zooms.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
wahlbrink: 1) I don’t know if there are special requirements (color depth) for cursor images on any platform
2) The ImageDataProvider works on the real monitor resolution:
    - It takes advantage of higher resolutions (when dividing the height into thirds)
    - You get exactly what you want at any zoom level (e.g. no rounding by drawing methods)
3) If you want to use ImageGcDrawer, your solution is not yet mature
    - The two visible rectangles does not always have same height 
    - The drawn rectangle is fragmented at some zoom levels
   
    Screenshot (2x): top ImageDataProvider, bottom ImageGcDrawer
    cursor

HeikoKlare: > 2\. The ImageDataProvider works on the real monitor resolution:

That's also the case for the `ImageGcDrawer`.

Couldn't the drawer simply be something like this?
```java
gc.setBackground(display.getSystemColor(SWT.COLOR_BLACK));
gc.fillRectangle(0, imageHeight / 3, imageWidth, imageHeight / 3);
```

wahlbrink: > That's also the case for the `ImageGcDrawer`

The image yes, but not the GC API (when using autoscale). It works with "zoom 100" int coordinates, therefore the positioning of elements using the GC's regular drawing methods is bound to the "zoom 100" grid.

HeikoKlare: > The image yes, but not the GC API (when using autoscale). It works with "zoom 100" int coordinates, therefore the positioning of elements using the GC's regular drawing methods is bound to the "zoom 100" grid.

If I understand correctly, that refers to an issue to be addressed by https://github.com/eclipse-platform/eclipse.platform.swt/pull/2913.

Is that relevant here? At a first glance, I don't see any drawing of the caret image with GC's `drawImage()` methods but only usages of OS methods taking the OS handles for drawing.

HeikoKlare (approved):
I tested with the following implementation, which looks fine on Windows, but turns out that it at least looks weird on MacOS.
```java
int width = getCaretWidthPreference();
Display display = styledText.getDisplay();
ImageGcDrawer imageGcDrawer = (gc, imageWidth, imageHeight) -> {
	gc.setBackground(display.getSystemColor(SWT.COLOR_BLACK));
	gc.fillRectangle(0, imageHeight / 3, imageWidth, imageHeight / 3);
};
return new Image(display, imageGcDrawer, width, styledText.getLineHeight());
```
For some reason MacOS does not rescaling and anti-aliasing of the caret image (without any need), which makes it look strange. This makes the current caret not look nice at all, with the implementation above it's a bit better, but with the one proposed in this PR it's the best we can have.

In consequence, I am in favor of using the current propsal as it properly fixes the issue for zooms != 100 and also improves the apperance on MacOS in general.

HeikoKlare:
Failing test is unrelated: 
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/3581

PR #3232, opened 2025-08-29
UI component being attempted to be disposed in a non-ui thread.

### LINKED ISSUE #2316: CompatibilityPart/IntroPart: tries to dispose in not-SWT thread
locally running CTabItemTest (passes) logs `This is not an UI thread (or Device already disposed)`.
```
java.lang.NullPointerException: This is not an UI thread (or Device already disposed)
	at java.base/java.util.Objects.requireNonNull(Objects.java:259)
	at org.eclipse.jface.resource.JFaceResources.getResources(JFaceResources.java:243)
	at org.eclipse.ui.part.IntroPart.dispose(IntroPart.java:99)
	at org.eclipse.ui.intro.config.CustomizableIntroPart.dispose(CustomizableIntroPart.java:347)
	at org.eclipse.ui.internal.ViewIntroAdapterPart.dispose(ViewIntroAdapterPart.java:150)
	at org.eclipse.ui.internal.e4.compatibility.CompatibilityPart.invalidate(CompatibilityPart.java:264)
	at org.eclipse.ui.internal.e4.compatibility.CompatibilityPart.destroy(CompatibilityPart.java:421)
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:56)
	at org.eclipse.e4.core.internal.di.InjectorImpl.processAnnotated(InjectorImpl.java:977)
	at org.eclipse.e4.core.internal.di.InjectorImpl.processAnnotated(InjectorImpl.java:939)
	at org.eclipse.e4.core.internal.di.InjectorImpl.disposed(InjectorImpl.java:435)
	at org.eclipse.e4.core.internal.di.Requestor.disposed(Requestor.java:161)
	at org.eclipse.e4.core.internal.contexts.ContextObjectSupplier$ContextInjectionListener.update(ContextObjectSupplier.java:83)
	at org.eclipse.e4.core.internal.contexts.TrackableComputationExt.update(TrackableComputationExt.java:103)
	at org.eclipse.e4.core.internal.contexts.TrackableComputationExt.handleInvalid(TrackableComputationExt.java:68)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.dispose(EclipseContext.java:179)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.dispose(EclipseContext.java:164)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.dispose(EclipseContext.java:164)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.dispose(EclipseContext.java:164)
	at org.eclipse.e4.core.internal.contexts.osgi.EclipseContextOSGi.dispose(EclipseContextOSGi.java:102)
	at org.eclipse.e4.core.internal.contexts.osgi.EclipseContextOSGi.bundleChanged(EclipseContextOSGi.java:144)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.dispatchEvent(BundleContextImpl.java:982)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:230)
	at org.eclipse.osgi.framework.eventmgr.ListenerQueue.dispatchEventSynchronous(ListenerQueue.java:151)
	at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEventPrivileged(EquinoxEventPublisher.java:235)
	at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEvent(EquinoxEventPublisher.java:134)
	at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEvent(EquinoxEventPublisher.java:126)
	at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor.publishModuleEvent(EquinoxContainerAdaptor.java:229)
	at org.eclipse.osgi.container.Module.publishEvent(Module.java:531)
	at org.eclipse.osgi.container.Module.doStop(Module.java:693)
	at org.eclipse.osgi.container.Module.stop(Module.java:554)
	at org.eclipse.osgi.container.SystemModule.stop(SystemModule.java:212)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle$EquinoxSystemModule$1.run(EquinoxBundle.java:240)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

sounds like that should be done in SWT thread. i don't know where that should be changed

deepika-u:
Fixes #2316

As suggested tried fixing in destroy(), but looks like we need extra testing to ensure no shutdown regressions considering the below
- In shutdown, syncExec might block if the display is already disposed or unresponsive.
- EventBroker cleanup still runs in non-UI thread, which is fine. But we have 2 different parts happening - part disposed later, eventbroker unsubscribed immediately.

Note : I have no recreate to test this fix, can this be tested with the existing Unit Test infrastructure in place already?

### DIFF
--- a/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
+++ b/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2015 IBM Corporation and others.
+ * Copyright (c) 2006, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -18,8 +18,10 @@ import org.eclipse.core.resources.IResource;
 import org.eclipse.jface.action.IMenuManager;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.SWTException;
 import org.eclipse.swt.dnd.Clipboard;
 import org.eclipse.swt.events.KeyEvent;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IActionBars;
 import org.eclipse.ui.ISharedImages;
@@ -55,9 +57,22 @@ public class EditActionGroup extends ActionGroup {
 
 	@Override
 	public void dispose() {
-		if (clipboard != null) {
-			clipboard.dispose();
-			clipboard = null;
+		Display display = Display.getCurrent();
+		if (display != null && !display.isDisposed()) {
+			if (clipboard != null && !clipboard.isDisposed()) {
+				clipboard.dispose();
+				clipboard = null;
+			}
+		} else {
+			// Non-UI thread or display disposed
+			if (clipboard != null && !clipboard.isDisposed()) {
+				try {
+					clipboard.dispose();
+					clipboard = null;
+				} catch (SWTException e) {
+					// Log a message if required.
+				}
+			}
 		}
 		super.dispose();
 	}
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2016 IBM Corporation and others.
+ * Copyright (c) 2010, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -32,6 +32,7 @@ import org.eclipse.e4.core.services.log.Logger;
 import org.eclipse.e4.ui.di.Focus;
 import org.eclipse.e4.ui.di.Persist;
 import org.eclipse.e4.ui.di.PersistState;
+import org.eclipse.e4.ui.di.UISynchronize;
 import org.eclipse.e4.ui.internal.workbench.Activator;
 import org.eclipse.e4.ui.internal.workbench.Policy;
 import org.eclipse.e4.ui.model.application.ui.basic.MPart;
@@ -418,9 +419,13 @@ public abstract class CompatibilityPart implements ISelectionChangedListener {
 	}
 
 	@PreDestroy
-	void destroy() {
+	void destroy(UISynchronize sync) {
 		if (!alreadyDisposed) {
-			invalidate();
+			sync.syncExec(() -> {
+				if (!alreadyDisposed) {
+					invalidate();
+				}
+			});
 		}
 
 		eventBroker.unsubscribe(widgetSetHandler);
--- a/bundles/org.eclipse.ui.navigator.resources/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.navigator.resources/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.navigator.resources; singleton:=true
-Bundle-Version: 3.9.800.qualifier
+Bundle-Version: 3.9.900.qualifier
 Bundle-Activator: org.eclipse.ui.internal.navigator.resources.plugin.WorkbenchNavigatorPlugin
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin


deepika-u:
@jukzi : can you take a look at this when you have some time please?

merks:
Note that he is "missing in action" so don't expect anything from him.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
```diff
@@ -420,7 +421,12 @@ public void create() {
 	@PreDestroy
 	void destroy() {
 		if (!alreadyDisposed) {
-			invalidate();
+			Display display = Display.getDefault();
+			if (display != null && !display.isDisposed() && Display.getCurrent() != display) {
+				display.syncExec(this::invalidate);
+			} else {
+				invalidate();
+			}
```
laeubi: E4 Has the `UISynchronize` service for this purposes:


```suggestion
void destroy(UISynchronize sync) {
if (!alreadyDisposed) {
sync.syncExec ...
```

I also wonder if not the caller is to blame not calling from the UI thread, so maybe this should better be fixed than on the client side.

laeubi (changes requested):

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
```diff
@@ -420,7 +421,12 @@ public void create() {
 	@PreDestroy
 	void destroy() {
 		if (!alreadyDisposed) {
-			invalidate();
+			Display display = Display.getDefault();
+			if (display != null && !display.isDisposed() && Display.getCurrent() != display) {
```
jukzi: typically the isDisposed() check has to be repeated in Display thread, because dispose could happen in the diplay thread meanwhile

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
deepika-u: yes, thanks for the newer way. updated.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/CompatibilityPart.java
deepika-u: > typically the isDisposed() check has to be repeated in Display thread, because dispose could happen in the diplay thread meanwhile

added another check as well.

deepika-u:
@laeubi @jukzi : all the comments are incorporated, can you check now please. Thanks for your suggestions.

laeubi:
@deepika-u I still think it is better fixed at the caller side than in this particular method, so in what cases can it really happen that it is called from non-ui thread?

deepika-u:
> @deepika-u I still think it is better fixed at the caller side than in this particular method, so in what cases can it really happen that it is called from non-ui thread?

CompatibilityPart.destroy()
 → InjectorImpl.disposed()
 → EclipseContext.dispose()
 → OSGI/Equinox bundle stop thread
That path is not on the UI thread. It’s on the OSGi framework thread during bundle shutdown. So yes, it can and does happen that destroy() can be called from non-UI thread. That’s how we got the NPE in the first place.

In general 
- In normal user-driven part closure, destroy() will indeed be called from the UI thread.
- But during application or bundle shutdown, destroy() can be invoked from OSGi/framework threads (non-UI). That’s exactly how the reported NPE occurred.
- Since IntroPart.dispose() and other cleanup touch SWT/JFace resources are all dealing with UI components, they must be marshaled to the UI thread.

Therefore, the correct place to fix is in CompatibilityPart.destroy, which bridges the generic e4 lifecycle and the UI-thread-sensitive 3.x parts.

This way we can ensure 
- Lifecycle events are still respected.
- invalidate() always runs in UI thread.
- No raw Display usage, just e4-provided UISynchronize.

laeubi:
> But during application or bundle shutdown, destroy() can be invoked from OSGi/framework threads (non-UI). That’s exactly how the reported NPE occurred.

I think you should find out where exactly this enters the system and what is triggering that part. There should be one place where we can ensure it is called inside the UI thread.

So can you share a stacktrace of the problematic call so we can better decide what would be the best place here?

deepika-u:
@laeubi 
> So can you share a stacktrace of the problematic call so we can better decide what would be the best place here?

I tried myself running CTabItemTest  and captured these logs.
I dint find anything related to the error in .log file. But could see the same error in console of eclipse. So just copied the content of console to a file and also sharing the .log file for reference.
[console_output.txt](https://github.com/user-attachments/files/22115442/console_output.txt)
[.log](https://github.com/user-attachments/files/22115449/default.log)

Please have a look when you get some time and let me know your opinion.

deepika-u:
@laeubi 
On further anlaysis, i felt EditActionGroup.dispose() also needs to be updated. So adding this change too to make the fix complete.
Now with the EditActionGroup.dispose() changes in place along with CompatibilityPart.destroy(), i am not able to get the original reported NPE anymore on running the same CTabItemTest. Attaching the console for the same -> 
[Console_output_4thSept.txt](https://github.com/user-attachments/files/22138345/Console_output_4thSept.txt)

Once you get some time, can you review the changes please.

But i see another problem here though not very important but still a problem "org.eclipse.swt.SWTException: Widget is disposed"
For addressing this problem i have a pr attached which adds safe guards but no functionality being updated via it.
Will be adding the pr once raised. 

#3241 is created for this.

laeubi:
In the log I can only see errors about already disposed part / device but not any access from another thread?

deepika-u:
> In the log I can only see errors about already disposed part / device but not any access from another thread?

yes exactly now with these both file changes, i dont see NPE anymore.

laeubi:
> > In the log I can only see errors about already disposed part / device but not any access from another thread?
> 
> yes exactly now with these both file changes, i dont see NPE anymore.

The point is that we should not fix side-effects but the callers who call the method outside the Device thread so that callers stacktrace would be required here to decide where to best fix it. Requiring everyone dispose method (and you already discovered now two places) does not really scale well.

deepika-u:
@laeubi 
> The point is that we should not fix side-effects but the callers who call the method outside the Device thread so that callers stacktrace would be required here to decide where to best fix it. Requiring everyone dispose method (and you already discovered now two places) does not really scale well.

May be i missed your intent and didnt understand you clearly. But what i also tried as updated in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3232#issuecomment-3248623090

I have tried running /org.eclipse.e4.ui.tests.css.swt/src/org/eclipse/e4/ui/tests/css/swt/CTabItemTest.java locally to reproduce the NPE. As said by jukzi in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2316#issue-2542988909

Please do let me know if i have to try out anything else or capture any other logs to be specific.

laeubi:
> I have tried running /org.eclipse.e4.ui.tests.css.swt/src/org/eclipse/e4/ui/tests/css/swt/CTabItemTest.java locally to reproduce the NPE

And was this successful? Sorry to ask but its a bit unclear to me if the problem is currently happens, only happens in the test and how it relates to `EditActionGroup`?

deepika-u:
> And was this successful? Sorry to ask but its a bit unclear to me if the problem is currently happens, only happens in the test and how it relates to `EditActionGroup`?

From this comment -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/3232#issuecomment-3248623090
If you go through the full description in this section i have ended up in adding the change in EditActionGroup. Respective log for reference is also attached.

Now with CompatibilityPart and EditActionGroup changes in place then attempt to run /org.eclipse.e4.ui.tests.css.swt/src/org/eclipse/e4/ui/tests/css/swt/CTabItemTest.java locally i am not seeing the NPE as reported in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2316

But i see errors related to "org.eclipse.swt.SWTException: Widget is disposed" which i am trying to address with pr https://github.com/eclipse-platform/eclipse.platform.ui/pull/3241

Never mind, thanks for asking again.

Observation : Now with both the pr changes in place when i run CTabItemTest locally, i dont see both the errors. It is clean.

laeubi:
@deepika-u now we have

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3241

is this PR actually needed?

deepika-u:
@laeubi 
Thanks for merging #3241 
> is this PR actually needed?

Yes definitely this would be needed. Because both are addressing 2 different problems.
Today i have updated my master to latest code.
Now when i run CTabItemTest on master, i am still seeing the java.lang.RuntimeException caused by CompatibilityPart and EditActionGroup. Attached the console output here -> 
[console_output_8thSept.txt](https://github.com/user-attachments/files/22201749/console_output_8thSept.txt)

Now i have updated my branch to be in sync with master and now when i run CTabItemTest, i dont see any exceptions as such.
For reference also attached console output -> 
[console_output_8thSept_onbranch_3232.txt](https://github.com/user-attachments/files/22201901/console_output_8thSept_onbranch_3232.txt)

So this confirms that this pr is also needed for successful run of CTabItemTest along with #3241. Hope i have clarified. Thanks for the ask.

laeubi:
> Now when i run CTabItemTest on master, i am still seeing the java.lang.RuntimeException caused by CompatibilityPart and EditActionGroup. 

What I'm wondering about is the following:

1. Do we see it anywhere when running Eclipse at all?
2. Why does it not happen in the ibuild but only in the IDE?
3. Should maybe the tests be simply marked to be running in the UI thread instead?

deepika-u:
@laeubi 
We don’t see this issue in day-to-day Eclipse usage - it only shows up in JUnit test runs because shutdown timing differs between the IDE and the ibuild environment. The change we introduced makes disposal more robust (guards against disposed widgets and device access), so it fixes the test failures without negative side effects. Marking the test as “UI-thread only” would only mask the problem; our fix ensures correctness in both runtime and test scenarios with no behavioral change.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
```diff
@@ -55,11 +57,25 @@ public EditActionGroup(Shell aShell) {
...
+			if (clipboard != null && !clipboard.isDisposed()) {
+				try {
+					clipboard.dispose();
+					clipboard = null;
+				} catch (SWTException e) {
+					// Log or ignore safely
+				}
+			}
 		}
-		super.dispose();
```
laeubi, line 62: Why not call super.dispose() anymore?

laeubi:
> our fix ensures correctness in both runtime and test scenarios with no behavioral change.

When disposal happens during call of dispose or async / blocking in UI running alongside could of course seen as behavior change. And if Test + real environments differ there is of course a chance for false positives (or we did not capture problems). Also it might have side-effects and complicates the code.

e.g. in your current change I can see

```
Display display = Display.getDefault();
if (display != null && !display.isDisposed()) {

}
```

if you look at javadoc you will see

> Returns the default display. **One is created (making the thread that invokes this method its user-interface thread) if it did not already exist**.

so your check for `null` will never succeed. If no display exits one will be created here (and maybe never disposed). So its always hard to guess its really "side-effect-free" to add such code.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
```diff
@@ -55,11 +57,25 @@ public EditActionGroup(Shell aShell) {
 
 	@Override
 	public void dispose() {
-		if (clipboard != null) {
-			clipboard.dispose();
-			clipboard = null;
+		Display display = Display.getDefault();
+		if (display != null && !display.isDisposed()) {
+			display.asyncExec(() -> {
```
laeubi: In the otherplace `syncExec` is used, why is this different here?

deepika-u: EditActionGroup.dispose() with asyncExec() because safety during shutdown is more important than ordering.
- Disposing the Clipboard (and similar UI resources) must happen on the UI thread.
- But dispose() might be called from a non-UI thread during shutdown.
- If we used syncExec() here and called it from a non-UI thread while the display is already disposing, it could deadlock (the UI thread might already be blocked, tearing down widgets).
- asyncExec() safely queues the disposal for "later" on the UI thread, avoiding blocking and deadlock risk.
- If the display is already gone, we just try/catch to clean up as best we can.
So in summary asyncExec() => fire and forget cleanup, safe during shutdown.

CompatibilityPart.destroy() with syncExec() because ordering and completion are crucial for consistent teardown.
- destroy() is part of the Eclipse 4 lifecycle (@PreDestroy), which means the framework expects the part to be fully disposed when this method returns.
- If you used asyncExec() here, invalidate() would run later, and there’s a risk that other teardown code(unsubscribes, model cleanup, etc.) would happen before invalidate() ran. That could cause inconsistencies or NPEs.
- By using syncExec(), you ensure that invalidate() actually completes before the method continues and the lifecycle is maintained correctly.
So in summary syncExec() => enforce order, synchronous cleanup before continuing.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
deepika-u: @laeubi 
The problem if we still call "super.dispose();" inside dispose(), it would run immediately - before the clipboard is actually disposed, since disposal happens asynchronously later.
This creates a lifecycle ordering ambiguity:
Context cleared too early.
Asynchronous disposal runs after super.dispose()

If that async runnable references anything indirectly tied to the context (logging, resources, actions), you can get into NPEs or disposed-object access.

So this is Why super.dispose() was dropped.
- super.dispose() is trivial (just setContext(null)).
- Avoided calling it before async disposal finishes, to prevent race conditions.
- In practice, leaving context uncleared is not a big resource leak compared to crashing the UI.

So it was sacrificed for safety.

But if we still want to be correct, the better solution is to still call super.dispose() inside the async block. That way we don’t silently skip the context cleanup from ActionGroup.

We can do this way ::
```
@Override
public void dispose() {
    Display display = Display.getDefault();
    if (display != null && !display.isDisposed()) {
        display.asyncExec(() -> {
            if (clipboard != null && !clipboard.isDisposed()) {
                clipboard.dispose();
                clipboard = null;
            }
            EditActionGroup.super.dispose();
        });
    } else {
        // Display already gone, dispose synchronously as a fallback
        if (clipboard != null && !clipboard.isDisposed()) {
            try {
                clipboard.dispose();
                clipboard = null;
            } catch (SWTException e) {
                // Log or ignore safely
            }
        }
        super.dispose();
    }
}
```

deepika-u:
@laeubi 
> Display display = Display.getDefault();
> if (display != null && !display.isDisposed()) {
> 
> }
> ```
> if you look at javadoc you will see
> 
> > Returns the default display. **One is created (making the thread that invokes this method its user-interface thread) if it did not already exist**.

Thanks for the feedback! You're absolutely right about the side effects of Display.getDefault(). I've updated the code to use Display.getCurrent() instead, which avoids creating a new display and ensures we're only disposing UI resources on the UI thread. This should reduce complexity and avoid unintended behavior in both test and runtime environments.

When you get some time, could you take a look at it now please?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
```diff
@@ -55,9 +57,22 @@ public EditActionGroup(Shell aShell) {
 
 	@Override
 	public void dispose() {
-		if (clipboard != null) {
-			clipboard.dispose();
-			clipboard = null;
+		Display display = Display.getCurrent();
```
laeubi, line 60: why not using `shell.getDisplay()` it seems obvious here?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
```diff
@@ -55,9 +57,22 @@ public EditActionGroup(Shell aShell) {
...
+		if (display != null && !display.isDisposed()) {
+			if (clipboard != null && !clipboard.isDisposed()) {
+				clipboard.dispose();
+				clipboard = null;
+			}
+		} else {
+			// Non-UI thread or display disposed
+			if (clipboard != null && !clipboard.isDisposed()) {
+				try {
+					clipboard.dispose();
```
laeubi, line 70: Can the clipboard really be disposed outside the UI thread?!?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
deepika-u: Yeah shell.getDisplay() is also a good choice. But i need to ensure that the Shell reference is not null and not disposed before calling shell.getDisplay(). Can i go ahead with this way?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
deepika-u: Ideally not supposed to be which might result into SWTException or resource leaks if disposal fails silently or false positives. 
In my opinion - If we are not on the UI thread and can't guarantee access to it, it's safer to skip disposal or log a warning.

Do you mean i can skip this line `clipboard.dispose();` at line 70?

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
laeubi: Thinking more can you use clipboard itself, it is already is disposed what is the point doing more work here?

deepika-u: @leubi 
Sorry, was not able to get your intent.  You meant to use Display.getCurrent() or shell.getDisplay()?

merks:
I hate to be a party pooper here  but I think a call to dispose on UI part on a non UI thread is wrong and the caller should not do that.  And this doesn't happen in a normal IDE when I exit.  But we're changing code here that will affect the normal behavior.  Normally this happens before org.eclipse.e4.core.internal.contexts.osgi.EclipseContextOSGi.dispose() is called:

image

I really don't think we should add defensive code in the UI to deal with a situation that arises only in a test.

I highly recommend finding a way to fix this by modifying the test to dispose the UI (close the workbench) before this destroy handling kicks in.

bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/actions/EditActionGroup.java
deepika-u: If we dont have this pr in place, while running CTabItemTest locally we'll again endup into original problem reshaped as( post pr 3241)  
this problem shown in the console output below -> 
[Console_without_pr3232.txt](https://github.com/user-attachments/files/22402471/Console_without_pr3232.txt)

laeubi: @deepika-u where do you get this is it windows? 
I think it makes sense to go through each exception step by step, the first I see in your log is:

```
org.eclipse.swt.SWTException: Device is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4946)
	at org.eclipse.swt.SWT.error(SWT.java:4832)
	at org.eclipse.swt.widgets.Display.error(Display.java:1332)
	at org.eclipse.swt.widgets.Display.getThread(Display.java:2696)
	at org.eclipse.swt.dnd.Clipboard.dispose(Clipboard.java:219)
	at org.eclipse.ui.internal.navigator.resources.actions.EditActionGroup.dispose(EditActionGroup.java:59)
```

This looks like a bug in SWT (windows), because calling dispose on an already disposed component should be a no-op (and if device is disposed clipboard is gone anyways...).

Looking into other widgets implementations the do

`if (display.thread != Thread.currentThread ()) error (SWT.ERROR_THREAD_INVALID_ACCESS);`

so calling `getThread` instead of access the field directly seems the culprit here.

PR #3238, opened 2025-09-01
Remove notification pop-up informing about find/replace overlay UI

HeikoKlare:
The find/replace overlay UI inside an editor replacing the existing find/replace dialog UI (or serving as an enabled-by-default alternative) was released together with a notification pop-up that informs about the new UI and the ability to switch back to the existing UI when the product is started for the first time. After five releases, the notification became obsolete as the UI is actually not new anymore and most users starting a product for the first time already know it.
Thus, this change removes the notification pop-up for the find/replace overlay UI.


vogella (approved):

vogella:
Yesterday I was thinking the same, strange that we still show this notification.

PR #3241, opened 2025-09-04
Added guards to safely handle cases when Display is null or disposed.

deepika-u:
getActiveWindow()
getSplashShell(Display display)

- Added guards to safely handle cases where the Display is null or disposed.
- Prevents SWTException: Widget is disposed when querying the active shell during shutdown avoiding crashes.

These changes make both methods more robust during application startup and shutdown, eliminating crashes caused by accessing disposed SWT resources.

Raised this pr as discussed in -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/3232#issuecomment-3253487291

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPlugin.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPlugin.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -1175,15 +1175,25 @@ public class WorkbenchPlugin extends AbstractUIPlugin {
 	 */
 	public static Shell getSplashShell(Display display)
 			throws NumberFormatException, IllegalArgumentException {
+		if (display == null || display.isDisposed()) {
+			return null;
+		}
 		Shell splashShell = (Shell) display.getData(DATA_SPLASH_SHELL);
-		if (splashShell != null)
+		if (splashShell != null) {
+			if (splashShell.isDisposed()) {
+				return null;
+			}
 			return splashShell;
+		}
 
 		String splashHandle = System.getProperty(PROP_SPLASH_HANDLE);
 		if (splashHandle == null) {
 			return null;
 		}
 
+		if (display.isDisposed()) {
+			return null;
+		}
 		splashShell = Shell.internal_new(display, Long.parseLong(splashHandle));
 
 		display.setData(DATA_SPLASH_SHELL, splashShell);
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/services/WorkbenchSourceProvider.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/services/WorkbenchSourceProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2015 IBM Corporation and others.
+ * Copyright (c) 2009, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPlugin.java
```diff
@@ -1175,15 +1175,21 @@ public static boolean isSplashHandleSpecified() {
...
-		if (splashShell != null)
+		if (splashShell != null && !splashShell.isDisposed())
 			return splashShell;
 
 		String splashHandle = System.getProperty(PROP_SPLASH_HANDLE);
 		if (splashHandle == null) {
 			return null;
 		}
 
+		if (display.isDisposed()) {
```
laeubi, line 1194: Should this not already be checked in line 1178?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPlugin.java
```diff
@@ -1175,15 +1175,21 @@ public static boolean isSplashHandleSpecified() {
 	 */
 	public static Shell getSplashShell(Display display)
 			throws NumberFormatException, IllegalArgumentException {
+		if (display == null || display.isDisposed()) {
+			return null; // display is gone, don’t touch it
+		}
 		Shell splashShell = (Shell) display.getData(DATA_SPLASH_SHELL);
-		if (splashShell != null)
+		if (splashShell != null && !splashShell.isDisposed())
```
laeubi: If the shell is disposed already, should we probably return `null` here?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPlugin.java
deepika-u: Since this being a timing issue, have added the check before its being used again.

In shutdown scenarios, it is possible that the Display was disposed in between (e.g., another thread closed the workbench).

Without the second check, Shell.internal_new() would throw an SWTException: Widget is disposed.

So this second check is a race-condition guard. It’s defensive programming against the Display being disposed after the first check.

laeubi: But who should dispose the display in the meanwhile? This looks a bit fishy here so is this really called outside the event thread?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPlugin.java
deepika-u: Did you mean this way?
```
Shell splashShell = (Shell) display.getData(DATA_SPLASH_SHELL);
if (splashShell != null) {
    if (!splashShell.isDisposed()) {
        return splashShell;
    } else {
        return null; 
    }
}
```

laeubi: Yes but it depends on the error case I think we like to handle.  So in what cases the shell is disposed? Is it only during shutdown? Then it should be save to early exit.

deepika-u: Agree with early exit. So updated as per that.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPlugin.java
deepika-u: You’re right - under normal circumstances, the Display is only disposed on the UI thread at shutdown. If we can guarantee that getSplashShell() is always called from the UI thread, then the second check is redundant.

However, since this is a static utility used during startup/shutdown, I was not 100% sure whether it’s always invoked on the UI thread. If there is no non UI thread caller, I can drop the second check. Otherwise, it serves as a defensive guard against race conditions.

PR #3245, opened 2025-09-05
Remove useless p2.inf from o.e.e4.ui.workbench.renderers.swt.cocoa

akurtakov:
It was introduced via
https://github.com/eclipse-platform/eclipse.platform.ui/commit/d321fb9a33ced720c196d7a1aa1bee8d1dd3bd06 most likely to workaround org.eclipse.swt host not mandating dependency on the correct os/arch specific fragment. As SWT does that this is useless and just generates issues with maven publishing. Pointed out in eclipse-platform/eclipse.platform.releng.aggregator#3264 (comment)


akurtakov:
Failing tests are unrelated. Merging.

PR #3246, opened 2025-09-06
Migrate test suites in org.eclipse.ui.tests to JUnit 5

HeikoKlare:
Also makes the suite classes auto-formatting compliant.


HeikoKlare:
@vogella I just noticed that you started a similar attempt in
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2993

Do you want to finalize that one or should we with this one?

vogella:
It's completely fine for me if you migrate the tests. I'm currently not able to work on them due to other commitments

HeikoKlare:
Seems like https://github.com/eclipse-platform/eclipse.platform.ui/pull/3353 was merged in favor of this one.

PR #3249, opened 2025-09-11
Hide commands from disabled activities in onboarding

### LINKED ISSUE #3248: Hide platform commands on onboarding area from disabled activities
If I add e.g. Copilot to Eclipse, but disable it via activities:

```
   <extension
        point="org.eclipse.ui.activities">

     <activity
           name="Github Copilot"
           description="Requires a valid Github Copilot enterprise subscription and an Internet connection."
           id="com.advantest.copilot">
     </activity>

     <activityPatternBinding
           activityId="com.advantest.copilot"
           pattern="com\.microsoft\.copilot.*">
     </activityPatternBinding>

     <category
           name="AI Integration"
           description="Requires a valid Github Copilot enterprise subscription and an Internet connection."
           id="com.advantest.copilot4eclipseCategory">
     </category>

     <categoryActivityBinding
           activityId="com.advantest.copilot"
           categoryId="com.advantest.copilot4eclipseCategory">
     </categoryActivityBinding>

   </extension>
```

In the welcome area we still see a command from Copilot:

Image

What we want instead is:

Image

trancexpress:
This change adjusts ModeledPageLayout.addEditorOnboardingCommandId() to not show commands from activities that are not enabled.

Fixes: #3248

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/ModeledPageLayout.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/ModeledPageLayout.java
@@ -24,6 +24,7 @@ import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.e4.core.services.contributions.IContributionFactory;
 import org.eclipse.e4.ui.internal.workbench.PartStackUtil;
 import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.e4.ui.model.application.commands.MBindingTable;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/e4/compatibility/ModeledPageLayout.java
```diff
@@ -644,6 +645,10 @@ public void addEditorOnboardingCommandId(String commandId) {
 		if (numberOfOnboardingCommands >= 5)
 			return;
 
+		IContributionFactory contributionFactory = application.getContext().get(IContributionFactory.class);
```
laeubi, line 648: should this not be part of the filtering? There it seem to try find some kind of first matching item for a commandhandler, so should really the commandId be filtered or not more then handler for it?

trancexpress: I've done the filtering as in this PR: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2801/files#diff-f7d17d4cd3de2d3b83537af06dfeedd5d81088f1c75483ee5cb38b46812a4fe6R378

See: `BindingTable.isActive()`

laeubi: I think here contributor is the same so it might not matter, I still wonder what one really wants to filter is it the command provider or the handler provider... possibly both?

iloveeclipse: This change is only about **visual representation** of the command shown on the onboarding editor area contributed via `editorOnboardingCommand` to the perspective, execution of this command is already impossible via https://github.com/eclipse-platform/eclipse.platform.ui/pull/2801.

iloveeclipse (approved):

PR #3251, opened 2025-09-11
Fix FormEditor.dispose() calling super.dispose() first

iloveeclipse:
Common pattern for init/dispose, setup/teardown is that on init, super is called first and on dispose last.

This is always done this way to allow super class initialize whatever needed first and subclass work on that, and the dispose should always perform cleanup in the opposite direction. Otherwise super code might cleanup things that are still needed for the subclass to properly cleanup.

Probably related to
https://github.com/eclipse-pde/eclipse.pde/issues/1963


iloveeclipse:
The Mac build failed downloading something from eclipse.org infra.
M1 is the best time for such changes, merging.

PR #3283, opened 2025-09-17
Don't increase scrollable size for zero width tables

iloveeclipse:
This seem to fix StackOverflow on "Trust" table on Linux, see https://github.com/eclipse-equinox/p2/issues/946.

No idea how and if this affect everything else.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
@@ -254,7 +254,7 @@ public abstract class AbstractColumnLayout extends Layout {
 		int width = Math.max(0, area.width - trim);
 
 		if (width > 1)
-			layoutTableTree(table, width, area, tableWidth < area.width);
+			layoutTableTree(table, width, area, tableWidth > 0 && tableWidth < area.width);
 
 		// For the first time we need to relayout because Scrollbars are not
 		// calculate appropriately


bundles/org.eclipse.jface/src/org/eclipse/jface/layout/AbstractColumnLayout.java
```diff
@@ -254,7 +254,7 @@ protected void layout(Composite composite, boolean flushCache) {
 		int width = Math.max(0, area.width - trim);
 
 		if (width > 1)
-			layoutTableTree(table, width, area, tableWidth < area.width);
+			layoutTableTree(table, width, area, tableWidth > 0 && tableWidth < area.width);
```
laeubi, line 257:
```suggestion
if (width > 1 && tableWidth > 0)
			layoutTableTree(table, width, area, tableWidth < area.width);
```

This seems more logical to me, we already check if the areas width is larger than 1 (why not zero? anyways ...) so it seems to already want to skip under this condition, having the table has no size at all seems a valid extension of such checks.

iloveeclipse: Assuming we merge this one and it will introduce regressions, would you then agree to merge https://github.com/eclipse-equinox/p2/pull/947 instead? 
I honestly don't have more time to spend on debugging the "Trust" preference page.

merks: I can relate to not having time, so I empathize.  I wold much prefer to see this cycle broken in this part of the code.

@laeubi 

You made a suggestion but I don't really understand it.

Is the indentation off on the new line 256?

iloveeclipse: > This seems more logical to me,

I haven't noticed that you have proposed different change. No, your proposal will not work, `layoutTableTree()` call is **really** needed. My proposal avoids some handling inside that method.

merks: I reviewed it in my workspace and it looks fine to me.  Please proceed.  (And thank you!).

merks (approved):

PR #3284, opened 2025-09-17
Install latest possible lsp4e

iloveeclipse:


iloveeclipse:
Merging, as this doesn't affect builds

PR #3290, opened 2025-09-18
Improve ImageBasedFrame scaling

### LINKED ISSUE #451: Improve ImageBasedFrame scaling logic to make it more dynamic
Since ImageBasedFrame is a Control defined outside SWT, we do not want to make it responsible for zoomChange since the controls inside it already adapt themselves to the zoom, but rather it should react to when it's parent is resizing making sure the children are wrapped properly.

amartya4256:
With both Async and Sync DPI scaling, ImageBasedFrame scales itself on ZoomChanged event to pack the control to wrap in it. However, with the current approach of scaling on DPI_CHANGED it works fine as it only receives the event after the children have been resized. However, since ImageBasedFrame is a Control wrapped inside a Canvas, those widgets are already scaled implicitly and there's no need to pack them on ZoomChanged event but rather when a Resize event is received at the parent, allowing the ImageBasedFrame to only recalibrate its size when the parent is resizing.

This approach moves the responsibility of handling ZoomChanged event from outside SWT for ImageBasedFrame and also provides a possibility to scale it properly when the scaling is Async, i.e. the children might not have scaled when ZoomChanged is received.

###How to test
Simply A/B testing with master to check if the toolbar images scale properly since they are all wrapped in ImageBasedFrames, especially the perspective switcher in the top right.

Before:
image

After: 
image

This PR improves the state of https://github.com/eclipse-platform/eclipse.platform.swt/pull/2520, although not dependent on it. Must be merged before it.

### DIFF
--- a/bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
+++ b/bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
@@ -14,6 +14,8 @@
  ******************************************************************************/
 package org.eclipse.e4.ui.widgets;
 
+import java.util.concurrent.atomic.AtomicReference;
+
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.ControlEvent;
 import org.eclipse.swt.events.ControlListener;
@@ -24,6 +26,7 @@ import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.widgets.Canvas;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.ToolBar;
 
 
@@ -66,9 +69,19 @@ public class ImageBasedFrame extends Canvas {
 			ImageBasedFrame frame = (ImageBasedFrame) event.widget;
 			frame.setCursor(null);
 		});
+
 		toWrap.addListener(SWT.ZoomChanged, event -> {
-			toWrap.pack(true);
-			setFramedControlLocation();
+			final AtomicReference<Listener> scaleOnParentResize = new AtomicReference<>();
+			scaleOnParentResize.set(e -> {
+				if (isDisposed()) {
+					return;
+				}
+				toWrap.pack(true);
+				setFramedControlLocation();
+				parent.layout();
+				parent.removeListener(SWT.Resize, scaleOnParentResize.get());
+			});
+			parent.addListener(SWT.Resize, scaleOnParentResize.get());
 		});
 
 		addMouseMoveListener(e -> {


fedejeanne:
How does one test this?

amartya4256:
> How does one test this?

It's just an improvement in how it ImageBasedFrame implemenation for this case. I would expect ImageBasedFrame to behave the same as it does right now on master, with resize of Parent and on DPI change.

HeikoKlare (commented):
This will now react to every resize event and on every OS with a relayouting of the parent and other adaptations to execute. Since this kind of adaptation only seems to be necessary on a DPI change, is that really the right thing to do? I would expect these adaptations to then be executed far too often.

fedejeanne:
> > How does one test this?
> 
> It's just an improvement in how it ImageBasedFrame implemenation for this case. I would expect ImageBasedFrame to behave the same as it does right now on master, with resize of Parent and on DPI change.

Then please provide a snippet (or name an existing one) so that we can validate the behavior and see if your expectations are met :-)

amartya4256:
> This will now react to every resize event and on every OS with a relayouting of the parent and other adaptations to execute. Since this kind of adaptation only seems to be necessary on a DPI change, is that really the right thing to do? I would expect these adaptations to then be executed far too often.

I agree. We can make this resize handler dependent on after ZoomChanged event. How do you like such idea?

```
AtomicBoolean isPostZoomChangedEvent = new AtomicBoolean(false);

toWrap.addListener(SWT.ZoomChanged, event -> isPostZoomChangedEvent.set(true));

parent.addListener(SWT.Resize, event -> {
	if (isDisposed() || !isPostZoomChangedEvent.get())
		return;
	toWrap.pack(true);
	setFramedControlLocation();
	parent.layout();
	isPostZoomChangedEvent.set(false);
});
```

HeikoKlare:
That would definitely be better. Couldn't even the `SWT.ZoomChanged` listener register the `SWT.Resize` listener as a one-time listener? I.e., the `SWT.ZoomChanged` listener would register the `SWT.Resize` listener that deregisters itself once it was executed.

amartya4256:
> That would definitely be better. Couldn't even the `SWT.ZoomChanged` listener register the `SWT.Resize` listener as a one-time listener? I.e., the `SWT.ZoomChanged` listener would register the `SWT.Resize` listener that deregisters itself once it was executed.

Adapted the suggested change.

HeikoKlare:
> Adapted the suggested change.

@amartya4256 is the latest change as intended? My proposal was to keep the `SWT.Resize` listener but only register and execute it once, but now we are back to only having an (enhanced) `SWT.ZoomChanged` listener. If that works, it's of course fine.

amartya4256:
My bad, I pushed without adding the latest change ;D

bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
```diff
@@ -66,9 +67,18 @@ public ImageBasedFrame(Composite parent, Control toWrap, boolean vertical,
 			ImageBasedFrame frame = (ImageBasedFrame) event.widget;
 			frame.setCursor(null);
 		});
+
 		toWrap.addListener(SWT.ZoomChanged, event -> {
-			toWrap.pack(true);
-			setFramedControlLocation();
+			final Listener[] scaleOnParentResize = new Listener[1];
```
HeikoKlare: Maybe use `AtomicReference` instead of this magic one-element array?

bundles/org.eclipse.e4.ui.widgets/src/org/eclipse/e4/ui/widgets/ImageBasedFrame.java
```diff
@@ -66,9 +67,18 @@ public ImageBasedFrame(Composite parent, Control toWrap, boolean vertical,
...
+			scaleOnParentResize[0] = e -> {
+				if (isDisposed())
+					return;
```
HeikoKlare: Wrap into block:
```suggestion
				if (isDisposed()) {
					return;
			    }
```

HeikoKlare (commented):
The change works fine for me. I only have minor comments.

HeikoKlare (approved):

PR #3315, opened 2025-09-23
Stackoverflow fix in ImageDescriptor

akurtakov:
Fixes getImageData(100) calling itself endlessly (introduced via https://github.com/eclipse-platform/eclipse.platform.ui/commit/c2fe1447c499d0de02b24e5bd07e18d559105e0e ) by reintroducing the bigger getImageData(100) and getImageData() endlessly (introduced in 2017 via
https://github.com/eclipse-platform/eclipse.platform.ui/commit/e6d12ca0ae4762273680dc65ae8eb6f211778acc ) that wasn't experienced thanks to the *happy accident* that it was asbtract method and all implementations before that had their own and every implementor since then most likely figured the endless loop while developing.
Enhanced documentation to mention that.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
@@ -424,30 +424,32 @@ public abstract class ImageDescriptor extends DeviceResourceDescriptor<Image> {
 	 */
 	public ImageData getImageData(int zoom) {
 		if (zoom == 100) {
-			return getImageData(100);
+			return getImageData();
 		}
 		return null;
 	}
 
 	/**
-	 * Creates and returns a new SWT <code>ImageData</code> object
-	 * for this image descriptor.
-	 * Note that each call returns a new SWT image data object.
+	 * Creates and returns a new SWT <code>ImageData</code> object for this image
+	 * descriptor. Note that each call returns a new SWT image data object.
 	 * <p>
-	 * This framework method is declared public so that it is
-	 * possible to request an image descriptor's image data without
-	 * creating an SWT image object.
+	 * This framework method is declared public so that it is possible to request an
+	 * image descriptor's image data without creating an SWT image object.
 	 * </p>
 	 * <p>
 	 * Returns <code>null</code> if the image data could not be created.
 	 * </p>
 	 * <p>
-	 * This method was abstract until 3.13. Clients should stop re-implementing
-	 * this method and should re-implement {@link #getImageData(int)} instead.
+	 * This method was abstract until 3.13. Clients should stop re-implementing this
+	 * method and should re-implement {@link #getImageData(int)} instead.
 	 * </p>
 	 *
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@akurtakov, do I miss something? PR description says about StackOverflow fix, but code change shows only javadoc comment changed?

merks (approved):
FYI, the installer was broken from stack overflow in org.eclipse.emf.edit.ui.provider.ImageWrapperImageDescriptor

akurtakov:
> @akurtakov, do I miss something? PR description says about StackOverflow fix, but code change shows only javadoc comment changed?

I forgot to push that part initially. Should be good now.

merks:
FYI, this was the stack overflow

image

iloveeclipse:
> I forgot to push that part initially. Should be good now.

Good, but if that was automated cleanup that redirects from deprecated to non deprecated methods, wouldn't that hit again on next cleanup round?

merks:
> > I forgot to push that part initially. Should be good now.
> 
> Good, but if that was automated cleanup that redirects from deprecated to non deprecated methods, wouldn't that hit again on next cleanup round?

Yes, I had the same question. But apparently the cleanup considers the Javadoc to determine the replacement call and @akurtakov tested that the changes to the Javadoc prevents the save action (automated by the cleanup bot) from making that change again.  We should keep and eye on this nevertheless!

akurtakov:
Save action runs that changes deprecated calls relies on deprecated javadoc tag content (https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/210f65acc02641c0e26a081aa88ca119c4c2b2ca/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/text/correction/QuickAssistProcessorUtil.java#L591 ) which explicitly kicks in only for "use" and "replace by" thus I changed the javadoc to "replace with" for the sake of differing.

iloveeclipse:
> thus I changed the javadoc to "replace with" for the sake of differing.

Hmm... 

Next time someone will change javadoc & face same issue again.

@jjohnstn : would it be possible to exclude automated replacement of deprecated methods via some kind of "do_not_fix_me" tag on the code that uses deprecated method?

laeubi:
@akurtakov javaodc of the class says:

>  Using this abstract class involves defining a concrete subclass
>  and re-implementing the {`@link #getImageData(int)`} method.
>  Legacy subclasses that are not HiDPI-aware used to override the
>  deprecated {`@link #getImageData()`} method. **Subclasses <b>must</b>
>  re-implement exactly one of the {`@code getImageData`} methods**
>  and they should not re-implement both.

So does this javadoc needs fixing?

akurtakov:
IMO one could drop the note about reimplement exactly one but I haven't given it enough thought. I'm merging this one to prevent the regression and starting new I-build. Documentation enhancement can happen in follow up PRs.

akurtakov:
Ongoing build is https://ci.eclipse.org/releng/job/Builds/job/I-build-4.38/40/ .

merks:
Thank you.  ❣️

PR #3328, opened 2025-09-24
Improve Quick Search Dialog UI

deepika-u:
This pr basically addresses the below issues

- [x] 1. When 1 or more search results exist, by default no entry is shown till user selects but now 1st entry is showed in preview section.
- [x] 2. After a search, when search pattern is cleared, the preview remains which is no more valid. Now the preview is cleared.

1. When there are 1 or more multiple search results, by default nothing is selected and shown in preview till user selects an entry. Now the 1st entry is shown by default.

Multiple search results
Before : 
image

After : 
image

1 Search result
Before : 
image

After : 
image

2. Now when the search pattern is cleared but preview section still showing up stale entries which are no more valid. Even the ruler section for the preview needs to be cleared.

Before : 
image

After : 
image

Please do let me know if you have any suggestions in the same context.

### DIFF
--- a/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
+++ b/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
@@ -58,6 +58,7 @@ import org.eclipse.jface.preference.PreferenceConverter;
 import org.eclipse.jface.resource.JFaceResources;
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.CursorLinePainter;
+import org.eclipse.jface.text.Document;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.IRegion;
 import org.eclipse.jface.text.source.CompositeRuler;
@@ -1173,6 +1174,9 @@ public class QuickSearchDialog extends SelectionStatusDialog {
 							viewer.revealRange(rangeStart, rangeEnd - rangeStart);
 
 							var targetLineFirstMatch = getQuery().findFirst(document.get(item.getOffset(), contextLenght - (item.getOffset() - start)));
+							if (targetLineFirstMatch == null) {
+							    return; // nothing to refresh
+							}
 							int targetLineFirstMatchStart = item.getOffset() + targetLineFirstMatch.getOffset();
 							// sets caret position
 							viewer.setSelectedRange(targetLineFirstMatchStart, 0);
@@ -1289,6 +1293,10 @@ public class QuickSearchDialog extends SelectionStatusDialog {
 				//element is available in the list.
 				openButton.setEnabled(itemCount>0);
 			}
+			//Auto-select the first search result for preview to be shown.
+			if (itemCount >= 1 && list.getSelection().isEmpty()) {
+	            list.getTable().select(0);
+	        }
 			refreshDetails();
 		}
 	}
@@ -1484,6 +1492,14 @@ public class QuickSearchDialog extends SelectionStatusDialog {
 		} else {
 			//The QuickTextSearcher is already active update the query
 			this.searcher.setQuery(newFilter, force);
+			if(newFilter.getPatternString().trim().isEmpty()) {
+				//When pattern is cleared, clear the preview section
+				viewer.setDocument(new Document("")); //$NON-NLS-1$
+			    if (lineNumberColumn != null) {
+			        viewer.removeVerticalRulerColumn(lineNumberColumn);
+			        viewer.addVerticalRulerColumn(lineNumberColumn);
+			    }
+			}
 		}
 		if (progressJob!=null) {
 			progressJob.schedule();


BeckerWdf:
Nice work. Always good so see such small but very helpfull improvement.

If am unsure about this:
> * Refresh button is a simple button. Now it is shown like an icon as in any other Git Repo view or Search view.

Why do you think that the icon is better then text button? Couldn't we also keep the button but add the icon to the button?

BeckerWdf:
> Nice work. Always good so see such small but very helpfull improvement.
> 
> If am unsure about this:
> 
> > * Refresh button is a simple button. Now it is shown like an icon as in any other Git Repo view or Search view.
> 
> Why do you think that the icon is better then text button? Couldn't we also keep the button but add the icon to the button?

In addition: For refresh I also see this icon in the IDE:
image
and this:
image

When to use which one?

BeckerWdf:
I have the feeling that the "icon" only would be correct if it's in a toolbar. But here we have a dialog that does not have a toolbar and for me it feels more correct to have a button.

deepika-u:
> Nice work. Always good so see such small but very helpful improvement.

Would love to do. Thanks. Anything which basically improves the user experience - i am happy to contribute there :)

> If am unsure about this:
> 
> > * Refresh button is a simple button. Now it is shown like an icon as in any other Git Repo view or Search view.
> 
> Why do you think that the icon is better then text button? Couldn't we also keep the button but add the icon to the button?

Let me check on it and update you.

deepika-u:
> In addition: For refresh I also see this icon in the IDE: 

image 1: used specifically for file refresh (e.g., synchronizing a file with the filesystem).

> and this: 

image 2: the more generic refresh icon, used in contexts like search results, project refresh, or view updates.
> When to use which one?

Since in this dialog the action is to refresh the search results, I chose the generic refresh icon.

BeckerWdf:
> Since in this dialog the action is to refresh the search results, I chose the generic refresh icon.

Yes. I agree.

deepika-u:
@BeckerWdf 
How does this look?
image

Please do share me an opinion on this? Alignment and spacing needs to be set, nothing finalized, just tried.

BeckerWdf:
I just had a chat with our UX-Expert on this topic.
She recommends not to use the "refresh" icon for that. See says that the "Referesh" button is totally fine as it is.
But one additional thing. Maybe we can move the "Refresh" button un into the same row as the text input fields. So in the same location has you have shown it here:
image
But keeping it as a button.

Just out of curiosity: Why is this button needed at all? I have had the need to press it. Can you explain that to me?

deepika-u:
@BeckerWdf 
I have checked and updated as below. Sorry for the delayed response as i got occupied with another issue.
image


Thanks for checking with UI experts.

Regarding why the button is needed:

- The search results don’t always auto-update if resources change externally (e.g. file changes from Git or another tool).
- Having an explicit refresh gives users a safe way to make sure results are up to date without closing/reopening the dialog.
- This also matches the pattern in other Eclipse views, where a manual refresh coexists with auto-updates.

So while it may not be used often, it provides an escape hatch when results are stale. Hope i have answered all your questions. Thanks.

BeckerWdf:
> @BeckerWdf I have checked and updated as below. Sorry for the delayed response as i got occupied with another issue. 

No worries. This happens to all of us.

> Thanks for checking with UI experts. 

Sure. I am always happy to help in the UX area there it's possible

> Regarding why the button is needed:
> 
> * The search results don’t always auto-update if resources change externally (e.g. file changes from Git or another tool).
> * Having an explicit refresh gives users a safe way to make sure results are up to date without closing/reopening the dialog.
> * This also matches the pattern in other Eclipse views, where a manual refresh coexists with auto-updates.
> 
> So while it may not be used often, it provides an escape hatch when results are stale.

Understood. I expected something like this.

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
```diff
@@ -813,14 +813,20 @@ public void getName(AccessibleEvent e) {
...
-		GridDataFactory.fillDefaults().span(4,1).grab(true, false).applyTo(searchInComposite);
+		GridDataFactory.fillDefaults().span(3,1).grab(true, false).applyTo(searchInComposite);
 		Label searchInLabel = new Label(searchInComposite, SWT.NONE);
 		searchInLabel.setText(Messages.QuickSearchDialog_In);
 		GridDataFactory.swtDefaults().indent(5, 0).applyTo(searchInLabel);
 		searchIn = new Combo(searchInComposite, SWT.SINGLE | SWT.BORDER | SWT.ICON_CANCEL);
 		searchIn.setToolTipText(Messages.QuickSearchDialog_InTooltip);
 		GridDataFactory.fillDefaults().grab(true, false).indent(5, 0).applyTo(searchIn);
 
+		// Refresh button next to pattern field towards extreme right
```
BeckerWdf: can't we use something like:

`
		Button refreshButton = createButton(inputRow, REFRESH_BUTTON_ID, Messages.QuickSearchDialog_Refresh, false);
		GridDataFactory.swtDefaults().align(SWT.CENTER, SWT.CENTER).grab(false, false).applyTo(refreshButton);
		refreshButton.addListener(SWT.Selection, e -> refreshButtonPressed())
`

BeckerWdf (changes requested):

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
deepika-u: > Button refreshButton = createButton(inputRow, REFRESH_BUTTON_ID, Messages.QuickSearchDialog_Refresh, false); GridDataFactory.swtDefaults().align(SWT.CENTER, SWT.CENTER).grab(false, false).applyTo(refreshButton); refreshButton.addListener(SWT.Selection, e -> refreshButtonPressed())

I just tried your suggestion as is and thinking this is more lesser lines of code(looks more appealing and readable for me too.). But when i tested i am seeing this small behavior change where alignment is lost.

quick_search

In the image, 1st section corresponds to createButton() kind of creation of button.
2nd section as per the existing pr change.

But thanks for letting me know various ways of thinking to address the same issue. Hope i have answered all the concerns raised. If i have missed any please do let me know.

Thanks alot @BeckerWdf in advance.

BeckerWdf: ok undestood.
pls. check what was done with the "REFRESH_BUTTON_ID" ?

deepika-u:
@BeckerWdf 
Could you take a look at this when you get some time please?

deepika-u:
@BeckerWdf 
Sorry to bother you again, could you take a look at this when you get some time please?

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/QuickSearchDialog.java
BeckerWdf: See the usage of "REFRESH_BUTTON_ID" in line 1362. As you don't do
refreshButton.setData(Integer.valueOf(REFRESH_BUTTON_ID));
code in line 1362 is now dead.

I think you don't need the REFRESH_BUTTON_ID stuff and the buttonPressed implementation when you do:
`refreshButton.addListener(SWT.Selection, e -> refreshButtonPressed());`
So I think you can remove that.

BeckerWdf (changes requested):

BeckerWdf:
when I open the dialog for the first time (a new workspace) it looks like this:
image

it is too small (related to the width). 

Without your change this looks good:

image

BeckerWdf:
Also this layout somehow is messed up with your change:
image

See that the drop down now has the wrong hight / alignment.

deepika-u:
@BeckerWdf 
Sorry to take abit long to get back on this issue.

I have rechecked further on this issue.
**On windows 11**, when i apply the pr and launch a child - i dont see the problem you reported.
image

**On mac**, when i apply the pr and launch a child - i am seeing this problem what you are reporting
image

Let me check if i can come up with a fix for it. (Will also check on first time look on mac as well when pr is applied.)

Otherwise option 2, we'll leave the refresh button as-is and proceed with fixes 1 and 2 sub issues - which is my last take on it. 
Anyways will keep you posted, thanks for your inputs.

BeckerWdf:
> Otherwise option 2, we'll leave the refresh button as-is and proceed with fixes 1 and 2 sub issues - which is my last take on it.

Sure.

deepika-u:
Sticking back to 1 and 2 sub issues. 
Tried multiple ways of setting the alignment correct but somehow its not helping but also seeing new problems. So revoking the 3rd sub issue - Move Refresh button to the search pattern line towards the right end.

@BeckerWdf - You may take a look when you get some time please. Updating the description of pr accordingly.

BeckerWdf (approved):

deepika-u:
@BeckerWdf 
Thanks alot for the approval.
Can you please take it forward as i dont have merge access when you get some time.

elsazac:
Test looks good. I would like to merge this by EOD, If there are no outstanding concerns on this PR.

elsazac:
Test looks good; merging this now.

PR #3329, opened 2025-09-26
Scale IBF on shell:resize after async ZoomChanged

akoch-yatta:
This PR adapts ImageBasedFrame scaling to be triggered when it's Shell receives a resize event after ZoomChanged event on its wrapped control. This will correct an inconsistent refresh when the ZoomChanged events are handled asynchronously.

This PR does not fix an existing visible issue, but will prevent an issue, after the asynchronous handling in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2520 is merged.


HeikoKlare (approved):
Looks good an works fine with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2520

Issue #194, opened 2022-07-05
Support pattern matching in the editor drop-down list


vogella:
Would be great to have pattern matching and highlighting and auto-insert from * in the editor drop-down also 

![image](https://user-images.githubusercontent.com/139910/177365252-2b07af46-26ff-48c0-bb92-33f1296f11c4.png)

@[pbodnar](https://github.com/issues?q=is%3Apr+is%3Aopen+author%3Apbodnar) maybe something for you too? It might be possible to reuse your work for open resource.

PR #3336, opened 2025-09-30
Optional filter text autocomplete support in chevron pop up

elsazac:
Refs : https://github.com/eclipse-platform/eclipse.platform.ui/issues/194

This PR address one of the feature suggestion in https://github.com/eclipse-platform/eclipse.platform.ui/issues/194.

This PR includes an autocomplete feature in the chevron popup filter, making it easier to navigate when too many editors are opened in the workbench. As the user types, matching editor names are suggested in the filter text area and the remaining part is auto-filled, improving speed and efficiency in locating editors.

This enhancement is implemented as an optional feature for the filter text in chevron pop up ie, users can enable or disable the autocomplete option based on their preference. A checkbox option to enable and disable the autocomplete behaviour is given in the **`Eclipse -> Settings -> General`** page. The feature is controlled through a new preference flag, **`ENABLE_AUTOCOMPLETE_IN_CHEVRON`**, stored under `org.eclipse.ui.workbench`. Users can enable or disable autocomplete in the Workbench Preferences, and when disabled, the chevron shows the normal standard filtering behaviour.

Key changes:

1. Introduced a new optional autocomplete behaviour in the chevron popup to provide suggestions and auto completion.
2. Normal text filtering is still available.
3. Added a **`checkbox`** in Preference page to let users enable or disable the autocomplete option depending upon their preference. 
4. By default the option is disabled, but if the feature is to be enabled it can be done via `Eclipse -> Settings -> General -> Enable autocomplete option in chevron popup`

https://github.com/user-attachments/assets/5c9183b7-ee77-474b-8cad-b8d5e97dde3e

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2018 IBM Corporation and others.
+ * Copyright (c) 2003, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.e4.ui.internal.workbench.renderers.swt;
 
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.e4.ui.workbench.swt.internal.copy.SearchPattern;
 import org.eclipse.e4.ui.workbench.swt.internal.copy.WorkbenchSWTMessages;
 import org.eclipse.jface.preference.JFacePreferences;
@@ -25,6 +27,8 @@ import org.eclipse.jface.viewers.TableViewer;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.jface.viewers.ViewerFilter;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.KeyAdapter;
+import org.eclipse.swt.events.KeyEvent;
 import org.eclipse.swt.events.KeyListener;
 import org.eclipse.swt.events.MouseEvent;
 import org.eclipse.swt.events.MouseListener;
@@ -312,6 +316,19 @@ public abstract class AbstractTableInformationControl {
 		return fTableViewer;
 	}
 
+	private TableItem suggestCompletionFromTable(String inputText) {
+		if (inputText == null || inputText.isEmpty()) {
+			return null;
+		}
+		String lowerCaseText = inputText.toLowerCase();
+		for (TableItem item : fTableViewer.getTable().getItems()) {
+			if (item.getText().toLowerCase().startsWith(lowerCaseText)) {
+				return item;
+			}
+		}
+		return null;
+	}
+
 	protected Text createFilterText(Composite parent) {
 		fFilterText = new Text(parent, SWT.NONE);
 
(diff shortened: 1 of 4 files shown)


elsazac:
@merks Would you be able to take a look at this?

elsazac:
@merks Would you mind reviewing this PR when you have a chance?

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
```diff
@@ -312,6 +316,19 @@ public TableViewer getTableViewer() {
 		return fTableViewer;
 	}
 
+	private String suggestCompletionFromTable(String Text) {
+		if (Text == null || Text.isEmpty()) {
```
merks: Don’t use a name that starts with a capital letter

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
```diff
@@ -312,6 +316,19 @@ public TableViewer getTableViewer() {
...
+			return null;
+		}
+		for (TableItem item : fTableViewer.getTable().getItems()) {
+			String textTable = item.getText();
+			if (textTable.toLowerCase().startsWith(Text.toLowerCase())) {
+				return textTable;
```
merks: That’s a lot of calls to lower case the parameter in the loop.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WorkbenchPreferencePage.java
```diff
@@ -388,6 +393,25 @@ protected IPreferenceStore doGetPreferenceStore() {
...
+		IEclipsePreferences prefs = InstanceScope.INSTANCE.getNode("org.eclipse.ui.workbench"); //$NON-NLS-1$
+		boolean enabled = prefs.getBoolean(WorkbenchMessages.Preference_Autocomplete, false);
+		enableAutocompleteButton.setSelection(enabled);
+
+		enableAutocompleteButton.addListener(SWT.Selection, e -> {
+			prefs.putBoolean(WorkbenchMessages.Preference_Autocomplete, enableAutocompleteButton.getSelection());
+			try {
+				prefs.flush();
+			} catch (Exception ex) {
+				ex.printStackTrace();
```
merks: No one will see this. Log or ignore.

merks (changes requested):

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
elsazac: Yes, corrected them.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WorkbenchPreferencePage.java
elsazac: Done, I have added proper logging instead.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
elsazac: Thanks, I have updated it now. The lowercase value is now computed only once before iterating.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WorkbenchPreferencePage.java
```diff
@@ -388,6 +393,25 @@ protected IPreferenceStore doGetPreferenceStore() {
...
+		enableAutocompleteButton = new Button(parent, SWT.CHECK);
+		enableAutocompleteButton.setText(WorkbenchMessages.Preference_Enable_Autocomplete_ChevronPopUp);
+		enableAutocompleteButton.setToolTipText(WorkbenchMessages.Preference_Enable_Autocomplete_ChevronPopUp_ToolTip);
+
+		IEclipsePreferences prefs = InstanceScope.INSTANCE.getNode("org.eclipse.ui.workbench"); //$NON-NLS-1$
+		boolean enabled = prefs.getBoolean(WorkbenchMessages.Preference_Autocomplete, false);
+		enableAutocompleteButton.setSelection(enabled);
+
+		enableAutocompleteButton.addListener(SWT.Selection, e -> {
+			prefs.putBoolean(WorkbenchMessages.Preference_Autocomplete, enableAutocompleteButton.getSelection());
```
merks: I'm trying to understand the context for this immediate reaction to the state change.  I.e., it seems to me this immediately changes the preference and saves the preference but if this is a preference page, nothing should be saved (or changed) until the user presses apply or OK.

merks (commented):
I'm concerned that the preference happens too quickly and not when the changes are applied by hitting Apply or OK.  Please clarify the details around that.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/WorkbenchPreferencePage.java
elsazac: yes, you're right, I shouldn't have added immediate persistence in the selection listener as it bypasses the Apply/Cancel options. i have now removed it.

merks:
I wanted to actually test it...

One thing I realized is had no idea about with the "chevron poup" actually refers to.   I also wondered why anyone would ever not want this, i.e., why an option (and one where the user is unlikely to map the label of the option to something concrete in the UI.

Maybe @iloveeclipse, @HeikoKlare, or @sratz have opinions about whether this should be an option on just be the new default behavior always?

It seems cool and convenient, but the behavior is not exactly what I expected.  I.e., when I was in this state below, I though hitting enter would choose the auto-complete value:

image

But it chooses whatever is selected down below, which is not unreasonable, but somehow inconvenient because I don't actually know how to "choose" the autocompleted value.  There seem to be no way which somehow detracts from the usefulness...

iloveeclipse:
I've checked the feature and I think the extra code doesn't improve usability, it confuses more then it helps.

The (probably AI generated) feature desription says:

>  As the user types, matching editor names are suggested in the filter text area and the remaining part is auto-filled, improving speed and efficiency in locating editors.

Current behavior is that the popup list already filters the list and preselects first matching item "as the user types", so after typing few letters all what user needs is to hit "enter" - it is already as simple as it could be. 

With the proposed extra code we *also* see the text for the already selected element in the text box, and that's it. Still hitting "enter" does same. 

So what is the benefit of filling in the text in the text box? IMO this PR can be closed.

elsazac:
Thanks @merks for testing this. apologies for the delayed response. I had been away for a while and am only now getting around to addressing everything.

> and one where the user is unlikely to map the label of the option to something concrete in the UI.

I agree that the name "chevron popup" isn't very understandable from a user POV as many might not be even be familiar with the name. If you have a suggestion for a more relevant one, i can update that. 

> But it chooses whatever is selected down below, which is not unreasonable, but somehow inconvenient because I don't actually know how to "choose" the autocompleted value. There seem to be no way which somehow detracts from the usefulness...

What I had been doing was pressing the right arrow key to accept the autocomplete text first, and then pressing enter to get the item. I hadn't really considered the possibility of using enter directly to accept the autocomplete suggestion, but your comment made me realize that it makes sense and is probably what users would naturally expect. so I have updated the code accordingly. I would appreciate it if you could give the updated code another try. Thanks!

merks:
The very first thing I try already bothers me.  I try to type `i` and it turns it into `I`:

image

And the thing I see showing is still not what's opened if I hit enter.  I don't know if most people would know to hit the right arrow first or most people would expect enter to open what's showing in the field where I'm typing...

iloveeclipse:
> So what is the benefit of filling in the text in the text box? IMO this PR can be closed.

This was not answered yet, and I still believe we are fixing not existing issue here.

PR #3339, opened 2025-09-30
Regenerate the EMF models of org.eclipse.e4.ui.model.workbench

merks:
- Fix spelling mistakes in the Ecore model so they don't reappear in the generated code.
- Change the GenModel compliance level to 21.
- Configure the project-specific cleanup action to include instanceof cleanup which is not part of the `Eclipse [built-in]` profile so that the cleanup of the generated code includes that action.


merks:
While looking at https://github.com/eclipse-pde/eclipse.pde/pull/1974 I noticed that other GenModels, the ones in this PR, could no longer regenerate the code and they generated results that undid the cleanups.

With this PRs changes, the generator will do the same cleanups, including the instanceof cleanup, and is able to generate clean code.  I had to change the project-specific preferences in a "not so nice way".  Most of the generated changes are whitespace/formatting changes except for the two files that @iloveeclipse modified by hand.

@iloveeclipse 

It is possible to preserver your changes to PartImpl and UIElementImpl.  One might change the `@generated` annotation to `@generated NOT`.  But then in the future if the model is changed the `toString` method will not pick up the changes.  One might in addition to that introduce a toStringGen method, and not call it, but leave it in the code so that if the model changes and that toStringGen method then also changes, maybe someone will think hard enough to modify the hand modified methods with the corresponding changes.  Please let me know your thoughts about this; I don't want to just overwrite your changes without your involvement.

iloveeclipse:
> It is possible to preserver your changes to PartImpl and UIElementImpl. One might change the `@generated` annotation to `@generated NOT`. But then in the future if the model is changed the `toString` method will not pick up the changes. One might in addition to that introduce a toStringGen method, and not call it, but leave it in the code so that if the model changes and that toStringGen method then also changes, maybe someone will think hard enough to modify the hand modified methods with the corresponding changes. Please let me know your thoughts about this; I don't want to just overwrite your changes without your involvement.

OK, the commit was 168689f533787c40783eea7b4b4fad741a7a8431 and the aim was to reduce "noise" from `toString()`.

I would like to "freeze" the code "as is", assuming if someone would need to see new attributes in debug output it is an easy task to add them manually.

So what is the simplest solution to keep the code "as is"?

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/application/ui/basic/impl/PartImpl.java
```diff
@@ -1334,8 +1334,9 @@ public Object eInvoke(int operationID, EList<?> arguments) throws InvocationTarg
 
 	/**
 	 * <!-- begin-user-doc -->
+	 * If the method {@link #toStringGen()} below changes, please ensure that this is updated.
 	 * <!-- end-user-doc -->
-	 * @generated
+	 * @generated NOT
 	 */
```
merks, line 1340: The `@generated NOT` ensures that this method is not overwritten during regeneration.

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/application/ui/basic/impl/PartImpl.java
```diff
@@ -1382,4 +1383,41 @@ public String toString() {
...
 
-} //PartImpl
+	/**
+	 * <!-- begin-user-doc -->
+	 * If the method below changes, please ensure that {@link #toString()} is updated.
+	 * <!-- end-user-doc -->
+	 * @generated
+	 */
+	@Override
+	public String toStringGen() {
```
merks, line 1393: This ensures that what's normally generated for toString() is generated here instead so that if the model changes, hopefully someone pays enough attention to read the documentation and to update the toString method above.

merks:
> So what is the simplest solution to keep the code "as is"?

I described the changes I made to preserve your toString changes.

OK?

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/application/ui/basic/impl/PartImpl.java
iloveeclipse: thanks.

bundles/org.eclipse.e4.ui.model.workbench/src/org/eclipse/e4/ui/model/application/ui/basic/impl/PartImpl.java
iloveeclipse: I see.

PR #3342, opened 2025-10-01
Revert ImageBasedFrame zoom change listener

akoch-yatta:
This PR reverts the change to the zoom changed listener of the ImageBasedFrame. The async DPI change handling of the windows implementation does not need the complex event chaining anymore.

It is reverting the changes from 44f830b8bd43aa6a96046f77830d2a20b642cf62 and 2f6a390496172bccdcbc3e696d3e68601f639773 keeping only the disposal check

Is supposed to not cause regressions with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2520


HeikoKlare (approved):
Works fine with (and without): https://github.com/eclipse-platform/eclipse.platform.swt/pull/2520
And of course it's much better to have the logic as simple as possible.

PR #3346, opened 2025-10-02
Migrate org.eclipse.tests.urischeme from JUnit4 to JUnit5

vogella:


vogella:
@HeikoKlare I going to send more AI generated PR with the updatess in case you want to watch them (I believe that the modernization  of the tests is an important topic for you), I will not ping you anymore directly in them.

PR #3347, opened 2025-10-02
Migrate org.eclipse.ui.tests.views.properties.tabbed from JUnit4 to J…

vogella:
…Unit5


vogella:
cc @HeikoKlare this is the first one from the AI generated merges

PR #3355, opened 2025-10-02
Refactor debug conditionals for improved readability in content assist

vogella:
classes

Using JDT clean-up "Convert control statement bodies to block"


vogella:
See #3293

PR #3356, opened 2025-10-02
Refactor debug conditionals for improved readability in content assist

vogella:
classes

Using JDT clean-up "Convert control statement bodies to block"

See #3293


HeikoKlare:
Please note that there is already
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3293

akurtakov:
It is actually an effort to make that one reviewable as even GH gives us "Due to the large number of changes in this pull request, only one file is being shown at a time." on the reffered PR. Hopefully, after few such PRs are merged, GH will make it actually reviewable in normal way.

PR #3363, opened 2025-10-02
Migrate org.eclipse.text.tests from JUnit4 to JUnit5

vogella:


akurtakov:
New warnings are caused by https://github.com/eclipse-platform/eclipse.platform.swt/pull/2563. Merging.

PR #3345, opened 2025-10-02
Add syntax highlight to code blocks

Mailaender:
This creates nicer previews on GitHub and when reading documentation with @eclipse-theia.

PR #3371, opened 2025-10-02
Add syntax highlight to code blocks part 2

Mailaender:
I forgot some in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3345.


akurtakov:
Merging as verification will not test this.

PR #3378, opened 2025-10-07
Add API filters for removed export of SVG package in SWT

HeikoKlare:
This is necessary once
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2602

is merged. JFace and UI reexport SWT, such that the removal of a package export in SWT is effectively a breaking API change of these bundles. So this needs to be merged right after the SWT PR is merged.


HeikoKlare:
Merging this together with
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2602

to ensure that with the next I-Build no API error occur. The API errors about unused problem filters in this PR are expected as it's built against SWT without https://github.com/eclipse-platform/eclipse.platform.swt/pull/2602 and will disappear with next I-Build.

PR #3379, opened 2025-10-07
Find/Replace Overlay: Restore cursor position when search input is cleared

### LINKED ISSUE #1944: find/replace overlay: return to last cursor position when emptying the search

deepika-u:
Restores the editor caret to its original position when the Find/Replace overlay search field is cleared.

Fixes #1944 

This PR introduces logic to restore the original caret/selection position when the search input field becomes empty, but only when incremental search mode is active. This ensures consistent and expected behavior for users typing in the Find/Replace overlay.

Key Changes
1. Updated performSearch()
  - Added a conditional check for SearchOptions.INCREMENTAL when the find string is empty.
  - Calls restoreSelectionIfEmpty() to restore the selection only in incremental mode.
  
2. Added restoreSelectionIfEmpty() method
  - Restores the selection to incrementalBaseLocation if incremental mode is active and a valid base location exists.
  - Skips restoration when incremental mode is not active to avoid unexpected behavior in non-incremental searches.

Accordingly the below test cases are updated/added.
- testResetIncrementalBaseLocation()
- testRestoreSelectionOnEmptyFindStringInIncrementalMode()
- testNoRestoreOnEmptyFindStringWhenNotIncremental()
- testSetFindString_incrementalActive()

@HeikoKlare : 
Can you have a look into this when you get some time please.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
@@ -324,6 +324,9 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	private boolean performSearch(boolean updateFromIncrementalBaseLocation) {
 		resetStatus();
 		if (findString.isEmpty()) {
+			if (isAvailableAndActive(SearchOptions.INCREMENTAL)) {
+				restoreSelectionIfEmpty();
+			}
 			return false;
 		}
 
@@ -336,6 +339,19 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 			// we don't keep state in this dialog
 		}
 		return somethingFound;
+	}
+
+	/**
+	 * Restores the original caret/selection position when the search field becomes
+	 * empty.
+	 */
+	private void restoreSelectionIfEmpty() {
+		if (incrementalBaseLocation == null) {
+			return;
+		}
+		if (target instanceof IFindReplaceTargetExtension extension) {
+			extension.setSelection(incrementalBaseLocation.x, incrementalBaseLocation.y);
+		}
 	}
 
 	/**
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
@@ -833,6 +833,53 @@ public class FindReplaceLogicTest {
 		findReplaceLogic.resetIncrementalBaseLocation();
 		findReplaceLogic.performSearch();
 		assertThat(textViewer.getSelectedRange(), is(new Point(5, 4)));
+		textViewer.setSelectedRange(7, 0);
+		findReplaceLogic.resetIncrementalBaseLocation();
+		findReplaceLogic.performSearch();
+		assertThat(textViewer.getSelectedRange(), is(new Point(10, 4)));
+		textViewer.setSelectedRange(10, 0);
+		findReplaceLogic.resetIncrementalBaseLocation();
+		findReplaceLogic.performSearch();
+		assertThat(textViewer.getSelectedRange(), is(new Point(10, 4)));
+		// Verify that after clearing the search, caret goes back to last active location
+		findReplaceLogic.setFindString("");
+		findReplaceLogic.performSearch();
+		assertThat(textViewer.getSelectedRange(), is(new Point(10, 4)));
+	}
+
+	@Test
+	public void testRestoreSelectionOnEmptyFindStringInIncrementalMode() {
+		String setupString= "test\ntest\ntest";
+		TextViewer textViewer= setupTextViewer(setupString);
+		textViewer.setSelectedRange(5, 0); // Set initial selection
+		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
+		findReplaceLogic.activate(SearchOptions.FORWARD);
+		findReplaceLogic.activate(SearchOptions.WRAP);
+		findReplaceLogic.activate(SearchOptions.INCREMENTAL);
+
+		findReplaceLogic.resetIncrementalBaseLocation(); // Set base location
+		findReplaceLogic.setFindString("test");
+		assertThat(textViewer.getSelectedRange(), is(new Point(5, 4)));
+		findReplaceLogic.setFindString(""); // Clear the find string
+		assertThat(textViewer.getSelectedRange(), is(new Point(5, 0))); // Restored to base location
+	}
+
+	@Test
+	public void testNoRestoreOnEmptyFindStringWhenNotIncremental() {
+		String setupString= "test\ntest\ntest";
+		TextViewer textViewer= setupTextViewer(setupString);
+		textViewer.setSelectedRange(5, 0); // Set initial selection
+		IFindReplaceLogic findReplaceLogic= setupFindReplaceLogicObject(textViewer);
+		findReplaceLogic.activate(SearchOptions.FORWARD);
+		findReplaceLogic.activate(SearchOptions.WRAP);
+		// Do NOT activate incremental mode
+
+		findReplaceLogic.resetIncrementalBaseLocation();
+		findReplaceLogic.setFindString("test");
+		assertThat(textViewer.getSelectedRange(), is(new Point(5, 0)));
+		findReplaceLogic.setFindString(""); // Clear the find string
+		// Will not restore selection since incremental mode is not active
+		assertThat(textViewer.getSelectedRange(), is(new Point(5, 0))); // Remains unchanged
 	}
 
 	@Test
@@ -866,8 +913,8 @@ public class FindReplaceLogicTest {
 		findReplaceLogic.setFindString("Te");
 		assertEquals(new Point(0, 2), findReplaceLogic.getTarget().getSelection());
 
-		findReplaceLogic.setFindString(""); // this clears the incremental search, but the "old search" still remains active
-		assertEquals(new Point(0, 2), findReplaceLogic.getTarget().getSelection());
+		findReplaceLogic.setFindString(""); // clears the incremental search input, restoring the caret to the base location
+		assertEquals(new Point(0, 0), findReplaceLogic.getTarget().getSelection());
 	}
 
 	@Test
--- a/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor; singleton:=true
-Bundle-Version: 3.19.400.qualifier
+Bundle-Version: 3.19.500.qualifier
 Bundle-Activator: org.eclipse.ui.internal.texteditor.TextEditorPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor.tests
-Bundle-Version: 3.14.1000.qualifier
+Bundle-Version: 3.14.1100.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: 


HeikoKlare (commented):
@deepika-u thank you for working on this!

We already have a bunch of tests for the `FindReplaceLogic` as well as for the dialog and overlay UIs (in `FindReplaceUITest`). Could you please extend them with test methods to cover the use cases you address? That would help to verify the behavior (now and against regressions in the future), and it would also make it easier to understand how the additions are supposed to behave when one can just read the use cases in the form of tests.

deepika-u:
> @deepika-u thank you for working on this!
> 
> We already have a bunch of tests for the `FindReplaceLogic` as well as for the dialog and overlay UIs (in `FindReplaceUITest`). Could you please extend them with test methods to cover the use cases you address? That would help to verify the behavior (now and against regressions in the future), and it would also make it easier to understand how the additions are supposed to behave when one can just read the use cases in the form of tests.

Let me check on them and also test failures. Thanks for your time.

deepika-u:
@HeikoKlare 
As suggested added and updated junit test cases for the same. Have a look when you get some time please.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -833,6 +833,77 @@ public void testResetIncrementalBaseLocation() {
...
+		TextViewer textViewer= setupTextViewer(setupString);
+		textViewer.setSelectedRange(6, 0); // caret after 'alpha '
+		IFindReplaceLogic logic= setupFindReplaceLogicObject(textViewer);
+		logic.activate(SearchOptions.FORWARD);
+		logic.activate(SearchOptions.INCREMENTAL);
+
+		// Initially empty find string
+		logic.setFindString("");
+		// Now start a new search (transition empty -> non-empty)
+		logic.setFindString("beta");
```
HeikoKlare: Wouldn`t it be better to use, e.g., "gamma" here, as (6,0) will now be both the position of the current incremental match as well as the restoreBaseLocation?

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -833,6 +833,77 @@ public void testResetIncrementalBaseLocation() {
...
+		// Expect the stored restore location to match the caret before search started
+		assertThat(((FindReplaceLogic) logic).getRestoreBaseLocation(), is(new Point(6, 0)));
```
HeikoKlare: It seems like this cast and exposing `getRestoreBaseLocation()` is only necessary for testing purposes. Since these tests are supposed to test the API, wouldn't it be better to test use cases of the API? E.g., a user is not interested in what the internal restoreBaseLocation value is (that's an implementation detail) but they are interested in the text viewer selection restoring to the original position after doing some search and then removing the find string again.
As an example, a test flow could be like this:
- Open document "alpha beta gamma"
- Set lection position in document to 6 (before "beta")
- Search for "gamma"
- Remove search input
- Expectation: selection position is 6 (whlie before implementing this feature it was 11)

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -833,6 +833,77 @@ public void testResetIncrementalBaseLocation() {
...
+
+		// Clear the search (simulate reopen of overlay)
+		logic.setFindString("");
+		logic.performSearch();
+
+		// Move caret and start new search session
+		textViewer.setSelectedRange(6, 0);
+		logic.setFindString("beta");
+		// Verify that restoreBaseLocation updated to the new caret position
+		assertThat(((FindReplaceLogic) logic).getRestoreBaseLocation(), is(new Point(6, 0)));
```
HeikoKlare: I tested a similar scenario which does not currently not work, and I am not sure if it's intended or not:
- Open document "alpha beta gamma"
- Set selection in document to 0 (before "alpha")
- Search for "gamma"
- Set selection in document to 6 (before "beta")
- Remove the search input

I expect the cursor to move to the last selected position 6 (before "beta"), but I get the cursor at position 0, the one when find string was empty for the last time.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -165,6 +171,10 @@ private static boolean isWord(String str) {
 	public void resetIncrementalBaseLocation() {
 		if (target != null && shouldInitIncrementalBaseLocation()) {
 			incrementalBaseLocation = target.getSelection();
+			if (isSetRestoreBaseLocation) {
+				restoreBaseLocation = incrementalBaseLocation;
+				isSetRestoreBaseLocation = false;
+			}
```
HeikoKlare: Can you explain the purpose of this code?
`isSetRestoreBaseLocation` seems to be initalized with true and is then set to false here, thus it's a one-time execution in the lifecycle of an instance of `FindReplaceLogic`. But to the best of my knowledge such an instance will live as long as an editor lives, which means that this code is just executed on first opening of a find/replace UI in that editor, and then it will never be executed again. And this first time happens right on initialization, as resetting the incremental base location happens when activating the `INCREMENTAL` search option. The `restoreBaseLocation` will only become relevant when a search string was entered and that will in every case happen after the incremental base position was set for the first time.
So I wonder if this code and the `isSetRestoreBaseLocation` field is necessary at all?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -60,6 +61,11 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 
 	@Override
 	public void setFindString(String findString) {
+		if (this.findString.isEmpty() && !findString.isEmpty()) {
+			// User just started a new search after clearing previous search.
+			if (target != null)
+				restoreBaseLocation = target.getSelection();
```
HeikoKlare: Please always put expression blocks into braces (applies in the same way also to other places in the PR).
```suggestion
			if (target != null) {
				restoreBaseLocation = target.getSelection();
			}
```

HeikoKlare (commented):
Thank you for the contribution and the addition of tests!
I tested the improvement a bit and it generally it works quite fine for me. I found one use case that behaves different than what I would expect (see the comments for it). And currently the change looks more complex to me than it needs to be, but maybe I miss something when trying to understand the purpose of that code (also see one of the comments for this).

deepika-u:
Sorry, i couldnt reply on your comments as i had some workspace issues. Now all set, will check and reply on this.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
deepika-u: Taken care, thanks.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
deepika-u: Thanks for the thoughtful review! You're absolutely right that isSetRestoreBaseLocation is currently used as a one-time flag, and its lifecycle is tied to the first activation of incremental search.
The intent behind this logic was to ensure that when a user clears the search input, the cursor position is restored to where it was before the new search began. However, I now realize that setting restoreBaseLocation during resetIncrementalBaseLocation() (which happens early) may not align with the actual user action of starting a new search.
I agree that the isSetRestoreBaseLocation flag adds complexity and may not be necessary. Instead, I can simplify the logic by capturing the selection directly in setFindString() when transitioning from an empty to a non-empty string — which is the actual moment a new search begins. This way, restoreBaseLocation is set only when needed, and there's no need for a lifecycle flag.
I have revised the implementation accordingly to remove isSetRestoreBaseLocation and updated the logic.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
deepika-u: The logic to capture restoreBaseLocation only triggers when transitioning from an empty to a non-empty find string (setFindString("") → setFindString("non-empty")).
In your case, you moved the caret after the search was already active, so the transition logic didn’t capture the new caret position.
In summary, also updated the test case. Your scenario fails because the caret must be moved before starting a new search string to capture the new base location. This is the behavior as per present design, based on the implementation in setFindString().

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
deepika-u: Sounds reasonable, so updated the test case.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
deepika-u: Even removed getRestoreBaseLocation(). Updated the test case in detail. Thanks.

deepika-u:
@HeikoKlare 
Can you recheck now when you get some time please? Mac failures are unrelated as i have seen. Thanks alot for your time.

deepika-u:
Mac failures cleared now. But seeing this issue now "05:42:31.581 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-source-plugin:5.0.1-SNAPSHOT:plugin-source (plugin-source) on project org.eclipse.ui.workbench: Error creating source archive: Problem creating jar: Execution exception: Java heap space -> [Help 1]

05:42:31.581 [ERROR] ".
@HeikoKlare  - Thanks alot for your time.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -50,7 +50,7 @@ public class FindReplaceLogic implements IFindReplaceLogic {
 	private IFindReplaceStatus status;
 	private IFindReplaceTarget target;
 	private Point incrementalBaseLocation;
-
+	private Point restoreBaseLocation = new Point(0, 0);
```
HeikoKlare: Why initialize with (0,0) and then later check for null (which it can never be)? This leads to setting the cursor location to (0,0) when performing a search while no find string was entered.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -338,6 +345,18 @@ private boolean performSearch(boolean updateFromIncrementalBaseLocation) {
 		return somethingFound;
 	}
 
+	/**
+	 * Restores the original caret/selection position when the search field becomes
+	 * empty.
+	 */
+	private void restoreSelectionIfEmpty() {
+		if (restoreBaseLocation == null)
+			return;
```
HeikoKlare: As noted in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3379#discussion_r2421823296: please always put expression blocks into braces (applies in the same way also to other places in the PR).

HeikoKlare (commented):
Thank you for the update. With the simplification, the PR is much more streamlined. I still have some comments and I wonder if it cannot be further simplified:
- The restore behavior is currently applied only depending on the find string state. However, this should probably only be a feature used when incremental mode is active. For example, what happens is that if you are using the find/replace dialog (without incremental model) and press ENTER when the find field is empty, it unexpectedly resets the cursor location to (0,0). So I think the behavior should be limited to incremental search and that should also be reflected in tests (either for the logic or maybe as system tests for the FindReplaceOverlay/FindReplaceDialog.
- Why do we need to store another location at all? We already have the `incrementalBasePosition`, which is the position where incremental search starts. It will be reset when you perform an actual search (clicking the search button). After that, every incremental search starts from that search results. So that's also the position the user has in mind when typing and results become highlighted. Why should we restore to an even different position if we clear the input field? That both adds technical complexity and also more reference points for the user to consider, making the functionality less comprehensible (in my opinion).

So I propose to try to completely remove the new base location and only use the incremental base position to restore to when clearing the input field. In addition, this should only be done incremental mode is active (or at least no unexpected behavior should occur in case incremental mode is deactivated).

deepika-u:
> So I propose to try to completely remove the new base location and only use the incremental base position to restore to when clearing the input field. In addition, this should only be done incremental mode is active (or at least no unexpected behavior should occur in case incremental mode is deactivated).

Hi Heiko, 
I have tried checking further on your inputs, updated the changes accordingly.  Can you take a look when you get some time please?

deepika-u:
@HeikoKlare 
Sorry to bother you again, can you take a look at it when you get some time please?

HeikoKlare:
I will have a look as soon as I find the time. In my opinion, it should wait for the next release cycle anyway since the fix is not urgently needed but affects a central UI/UX-related element with potentially unforseeable side effects. That's why I have to consider it with low priority as long as we have things for the upcoming release to do.

deepika-u:
Appreciate the heads-up. I’ll proceed with low-impact investigation and stay aligned with your timeline.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
```diff
@@ -867,7 +914,7 @@ public void testSetFindString_incrementalActive() {
 		assertEquals(new Point(0, 2), findReplaceLogic.getTarget().getSelection());
 
 		findReplaceLogic.setFindString(""); // this clears the incremental search, but the "old search" still remains active
-		assertEquals(new Point(0, 2), findReplaceLogic.getTarget().getSelection());
+		assertEquals(new Point(0, 0), findReplaceLogic.getTarget().getSelection()); // after #3379 pr
```
HeikoKlare: It would be good to update the comment, as the behavior changed so that it does not fit anymore. Then with the proper expected behavior described in the comment, there is also not need for the additional comment referring to the PR, as the description itself is comprehensive enough.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
```diff
@@ -338,6 +340,23 @@ private boolean performSearch(boolean updateFromIncrementalBaseLocation) {
...
 
+	/**
+	 * Restores the original caret/selection position when the search field becomes
+	 * empty.
+	 */
+	private void restoreSelectionIfEmpty() {
+	    if (!isAvailableAndActive(SearchOptions.INCREMENTAL)) {
+			// Incremental mode not active, so skipping restore
+	        return;
+	    }
```
HeikoKlare: This is obsolete as the caller already does the check. Either of the two checks may be removed.

HeikoKlare (commented):
I really like the current proposal. The logic is greatly simple and straightforward now. At least for me, the behavior also fits to what I as a user would expect.
Thank you also for updating the tests again.

I only have minor comments left. Then I would consider this ready to be merged right after master opens for the next release cycle.

tests/org.eclipse.ui.workbench.texteditor.tests/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogicTest.java
deepika-u: done.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceLogic.java
deepika-u: removed.

deepika-u:
> I really like the current proposal. The logic is greatly simple and straightforward now. At least for me, the behavior also fits to what I as a user would expect. Thank you also for updating the tests again.
> 
> I only have minor comments left. Then I would consider this ready to be merged right after master opens for the next release cycle.

Thanks alot once again for your time.
All the changes are done now. Commits are squashed.

HeikoKlare (approved):
Thank you! Let's merge this once `master` branch is open for 4.39 development.

deepika-u:
Thanks alot.

PR #3387, opened 2025-10-09
Improve Contrast of Line Number Color

### LINKED ISSUE #3385: Color Contrast of Line Number Ruler Not Sufficient

## Steps to reproduce

Check the contrast of the line number ruler

I expected: A contrast ration of 4.5:1

But got: A contrast ration of 4.42:1

See: https://coolors.co/contrast-checker/787878-ffffff

Changing the line number color only a bit improves this a lot. See: https://coolors.co/contrast-checker/767676-ffffff

## Community

BeckerWdf:
Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3385

### DIFF
--- a/bundles/org.eclipse.ui.editors/plugin.xml
+++ b/bundles/org.eclipse.ui.editors/plugin.xml
@@ -1077,7 +1077,7 @@
           id="org.eclipse.ui.editors.lineNumberRulerColor"
           isEditable="false"
           label="%dummy"
-          value="120,120,120">
+          value="115,115,115">
        </colorDefinition>
        <colorDefinition
           id="org.eclipse.ui.editors.hyperlinkColor"


BeckerWdf:
You can see the difference here: 

Screenshot 2025-10-09 at 15 28 53

I made screenshots and zoomed it so that the difference more visible.

bundles/org.eclipse.ui.editors/plugin.xml
```diff
@@ -1077,7 +1077,7 @@
           id="org.eclipse.ui.editors.lineNumberRulerColor"
           isEditable="false"
           label="%dummy"
-          value="120,120,120">
+          value="115,115,115">
```
mickaelistria, line 1080: Couldn't we instead use a default system color constant instead, like `COLOR_WIDGET_NORMAL_SHADOW_RGBA` ? The OS usually makes good choice regarding accessibility. Relying on them by default would probably prevent current, and future, trouble.

BeckerWdf: What's the value of COLOR_WIDGET_NORMAL_SHADOW_RGBA on the different OSes?

akurtakov: The value depends on the theme used. One can run SWT's ControlExample  - it has a tab that shows all these colors.

mickaelistria: On Linux it's 172,172,171 so actually brighter... But still, I don't think any color should be hardcoded, and we should instead rely on the OS, which may modify these values according to accessibility settings.

BeckerWdf: 177,172,171 is even worse:
image

So this does not help.

mickaelistria: > So this does not help.

The question is also "help in what?" Fixing a problem that doesn't exist on the whole OS or for which OS already has solutions?

IMO, hardcoding accessibility or even custom widget logic in Platform is overall a source of problem. Relying on the OS often brings much better results, more reactive with OS theme settings, new trends and so on. I don't think that Eclipse should focus on hardcoding local solutions to problems that the OS itself is not fixing. Concretely, my experience having learned a bit how Gnome does its UI/UX decisions with real experts; and how we have developed UI decisions in Eclipse Platform for years, I have come to the personal conclusion that we are clearly multiple leagues behind the OS-level UI teams and I believe blindly trusting what they do is going to a better and more sustainable outcome than trying to improve things "locally" by ourselves in Platform.

Compared with standard apps on gnome: gedit has line numbers in same color as text (but in a ruler with light grey background), and gnome-builder uses a light grey (probably LIGHT_SHADOW).

vogella: The new hard coded color looks better than the old hard-coded color, so +1 from me.

The discussion if system colors are better is IMHO independent of this, we currently have a hard-coded non system color.

BeckerWdf: I ran the control example on macOS and none of the greyish colors (even not COLOR_WIDGET_NORMAL_SHADOW) has a sufficient contrast.

I also agree with @vogella that the discussion on system colors should be independent from this concrete issue. Maybe even in the broader scope the the "SWT modernization / Project 31" topic.

I think this PR really improves the situation for our users. So pls. let's agree on merging this one.

BeckerWdf: > Concretely, my experience having learned a bit how Gnome does its UI/UX decisions with real experts;

By the way: I was also working with a real (UX) expert from my team on that topic.

mickaelistria: > let's agree on merging this one.

Let's say, I don't disagree on merging it. It's indeed not worse from the screenshot.
It looks like this color is not overridden by the dark theme. Have you checked it with dark theme? Should the dark theme color be adjusted too?

> By the way: I was also working with a real (UX) expert from my team on that topic.

Then (out of curiosity), can you please ask them how come that this line numbers are a problem for other Gnome-based apps seem happy with a lighter color for line numbers?

Also, 1 other thing maybe worth challenging in the choice of color can be "why sticking to greyscale"? VSCode or Visual Studio seem to be using some dark cyan color and it looks nice, maybe the color is even derived from the  COLOR_LINK_FOREGROUND system color.
That can be a good inspiration.

BeckerWdf: > It looks like this color is not overridden by the dark theme. Have you checked it with dark theme? Should the dark theme color be adjusted too?

I did check. It's overridden in the dark theme and contrast in the dark theme is ok.

BeckerWdf: > Also, 1 other thing maybe worth challenging in the choice of color can be "why sticking to greyscale"? VSCode or Visual Studio seem to be using some dark cyan color and it looks nice, maybe the color is even derived from the COLOR_LINK_FOREGROUND system color.
> That can be a good inspiration.

I also checked VSCode. For VSCode there are a ton of themes and different themes do that differently. I also so that cyan-ish color but there are also themes that use a gresh-ish color. 

I just wanted to stick to the "Eclipse way" and only make minimal changes.

BeckerWdf: > Then (out of curiosity), can you please ask them how come that this line numbers are a problem for other Gnome-based apps seem happy with a lighter color for line numbers?

What do you mean? You you please explain your question again?

PR #3397, opened 2025-10-13
Fix for multi threading issues in AnnotationMap

mehmet-karaman:
- This PR will fix multithreading issues described in https://github.com/eclipse-xtext/xtext/issues/3524
- added synchronize block to methods in AnnotationModel, wherever reads / writes could happen in multiple threads.
- changed execution order, to be able to synchronize only necessary code blocks.
- Removed dead code in AnnotationModel.cleanup because mapLock can't be null.
- Added new javadoc to IAnnotationMap.getLockObject().
- added assert.isLegal check to AnnotationMAp.setLockObject()

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
@@ -451,18 +451,17 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 	 */
 	protected void addAnnotation(Annotation annotation, Position position, boolean fireModelChanged) throws BadLocationException {
 		IAnnotationMap annotations= getAnnotationMap();
-		if (!annotations.containsKey(annotation)) {
-
-			addPosition(fDocument, position);
-			annotations.put(annotation, position);
+		synchronized (getLockObject()) {
+			if (annotations.containsKey(annotation)) {
+				return;
+			}
 			fPositions.put(position, annotation);
-			synchronized (getLockObject()) {
-				getAnnotationModelEvent().annotationAdded(annotation);
-			}
-
-			if (fireModelChanged) {
-				fireModelChanged();
-			}
+			annotations.put(annotation, position);
+			getAnnotationModelEvent().annotationAdded(annotation);
+		}
+		addPosition(fDocument, position);
+		if (fireModelChanged) {
+			fireModelChanged();
 		}
 	}
 
@@ -679,9 +678,8 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 
 			ArrayList<Annotation> deleted= new ArrayList<>();
 			IAnnotationMap annotations= getAnnotationMap();
-			Object mapLock = annotations.getLockObject();
 
-			synchronized (mapLock) {
+			synchronized (getLockObject()) {
 				annotations.forEach((a, p) -> {
 					if (p == null || p.isDeleted()) {
 						deleted.add(a);
@@ -693,10 +691,12 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 				removeAnnotations(deleted, false, false);
 				synchronized (getLockObject()) {
 					if (fModelEvent != null) {
+						final AnnotationModelEvent event = fModelEvent;
+						fModelEvent = null;
 						new Thread() {
 							@Override
 							public void run() {
-								fireModelChanged();
+								fireModelChanged(event);
 							}
 						}.start();
 					}
@@ -760,7 +760,11 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 
 		try {
 			Position[] positions= document.getPositions(IDocument.DEFAULT_CATEGORY, offset, length, canStartBefore, canEndAfter);
-			return new AnnotationsInterator(positions, fPositions);
+			IdentityHashMap<Position, Annotation> positionsMap;
+			synchronized (getLockObject()) {
+				positionsMap= new IdentityHashMap<>(fPositions);
+			}
+			return new AnnotationsInterator(positions, positionsMap);
 		} catch (BadPositionCategoryException e) {
 			// can happen if e.g. the document doesn't contain such a category, or when removed in a different thread
 			return Collections.<Annotation>emptyList().iterator();
@@ -784,11 +788,14 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 			return iter;
 		}
 
-		List<Iterator<Annotation>> iterators= new ArrayList<>(fAttachments.size() + 1);
-		iterators.add(iter);
-		Iterator<Object> it= fAttachments.keySet().iterator();
-		while (it.hasNext()) {
-			iterators.add(fAttachments.get(it.next()).getAnnotationIterator());
+		List<Iterator<Annotation>> iterators;
+		synchronized (getLockObject()) {
+			iterators= new ArrayList<>(fAttachments.size() + 1);
+			iterators.add(iter);
+			Iterator<Object> it= fAttachments.keySet().iterator();
+			while (it.hasNext()) {
+				iterators.add(fAttachments.get(it.next()).getAnnotationIterator());
+			}
 		}
 
 		return new MetaIterator<>(iterators.iterator());
@@ -837,22 +844,21 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 	 */
 	protected void removeAllAnnotations(boolean fireModelChanged) {
 		IAnnotationMap annotations= getAnnotationMap();
-		if (fDocument != null) {
+		List<Position> positionsToRemove = new ArrayList<>();
+		synchronized (getLockObject()) {
 			Iterator<Annotation> e= getAnnotationMap().keySetIterator();
 			while (e.hasNext()) {
 				Annotation a= e.next();
 				Position p= annotations.get(a);
-				removePosition(fDocument, p);
-//				p.delete();
-				synchronized (getLockObject()) {
-					getAnnotationModelEvent().annotationRemoved(a, p);
-				}
+				positionsToRemove.add(p);
+				getAnnotationModelEvent().annotationRemoved(a, p);
 			}
+			annotations.clear();
+			fPositions.clear();
 		}
-
-		annotations.clear();
-		fPositions.clear();
-
+		for (Position position : positionsToRemove) {
+			removePosition(fDocument, position);
+		}
 		if (fireModelChanged) {
 			fireModelChanged();
 		}
@@ -872,24 +878,22 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 	 */
 	protected void removeAnnotation(Annotation annotation, boolean fireModelChanged) {
 		IAnnotationMap annotations= getAnnotationMap();
-		if (annotations.containsKey(annotation)) {
-
-			Position p= null;
+		Position p;
+		synchronized (getLockObject()) {
 			p= annotations.get(annotation);
-			if (fDocument != null) {
-				removePosition(fDocument, p);
-//				p.delete();
+			if (p == null) {
+				return;
 			}
-
 			annotations.remove(annotation);
 			fPositions.remove(p);
-			synchronized (getLockObject()) {
-				getAnnotationModelEvent().annotationRemoved(annotation, p);
-			}
+			getAnnotationModelEvent().annotationRemoved(annotation, p);
+		}
+		if (fDocument != null) {
+			removePosition(fDocument, p);
+		}
 
-			if (fireModelChanged) {
-				fireModelChanged();
-			}
+		if (fireModelChanged) {
+			fireModelChanged();
 		}
 	}
 
@@ -982,13 +986,20 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 	@Override
 	public void addAnnotationModel(Object key, IAnnotationModel attachment) {
 		Assert.isNotNull(attachment);
-		if (!fAttachments.containsValue(attachment)) {
-			fAttachments.put(key, attachment);
-			for (int i= 0; i < fOpenConnections; i++) {
-				attachment.connect(fDocument);
+		Assert.isNotNull(key);
+		synchronized (getLockObject()) {
+			if (fAttachments.containsValue(attachment)) {
+				return;
 			}
-			attachment.addAnnotationModelListener(fModelListener);
+			fAttachments.put(key, attachment);
 		}
+		IDocument document= fDocument;
+		if (document != null) {
+			for (int i= 0; i < fOpenConnections; i++) {
+				attachment.connect(document);
+			}
+		}
+		attachment.addAnnotationModelListener(fModelListener);
 	}
 
 	/*
@@ -1006,13 +1017,21 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 	 */
 	@Override
 	public IAnnotationModel removeAnnotationModel(Object key) {
-		IAnnotationModel ret= fAttachments.remove(key);
-		if (ret != null) {
-			for (int i= 0; i < fOpenConnections; i++) {
-				ret.disconnect(fDocument);
+		Assert.isNotNull(key);
+		IAnnotationModel ret;
+		synchronized (getLockObject()) {
+			ret= fAttachments.remove(key);
+			if (ret == null) {
+				return null;
 			}
-			ret.removeAnnotationModelListener(fModelListener);
 		}
+		IDocument document= fDocument;
+		if (document != null) {
+			for (int i= 0; i < fOpenConnections; i++) {
+				ret.disconnect(document);
+			}
+		}
+		ret.removeAnnotationModelListener(fModelListener);
 		return ret;
 	}
 


laeubi:
@mehmet-karaman can you please more explain the rationale behind the changes? I think we can assume code is there for a reason so if we change fundamental things we should carefully describe the reasons and causes and why it is not a bug of the caller for example as otherwise we might run into deadlocks if code that previously run outside locks now run under lock conditions.

iloveeclipse:
The change here is coming from https://github.com/eclipse-xtext/xtext/issues/3524 investigation.

iloveeclipse:
@szarnekow : if you have time, would be good if you could check this PR.

laeubi:
It still would be good to more explain the individual changes and why they are required / needed. Also I think some changes are better a separate PR (as they are easier to review then e.g. "Removed dead code" or API documentation enhancements)

iloveeclipse:
> Also I think some changes are better a separate PR

Could you please point which exactly?

laeubi:
Each of those can be an own PR

- Removed dead code in AnnotationModel.cleanup because mapLock can't be null.
- Added new javadoc to IAnnotationMap.getLockObject()
- added assert.isLegal check to AnnotationMAp.setLockObject()

All of these seem independent and local enough to be quickly reviewed and merged and likely will improve things without risk for regression. And even if they are easier to revert in isolation.

Then might be the next thing

- changed execution order, to be able to synchronize only necessary code blocks.

and finally

- added synchronize block to methods in AnnotationModel, wherever reads / writes could happen in multiple threads.


This will make each PR small, focused and likely better to understand the implications and its easier to write a concise commit message.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -827,23 +824,25 @@ public void removeAllAnnotations() {
 	 * @param fireModelChanged indicates whether to notify all model listeners
 	 */
 	protected void removeAllAnnotations(boolean fireModelChanged) {
+		if (fDocument == null) {
+			return;
```
iloveeclipse: I believe we should remove this check completely to make sure model is "cleaned  up". 
`removePosition(fDocument, position);` already checks for the document not being null, so it should be fine.

iloveeclipse:
> Each of those can be an own PR
> 
> * Removed dead code in AnnotationModel.cleanup because mapLock can't be null.
> * Added new javadoc to IAnnotationMap.getLockObject()
> * added assert.isLegal check to AnnotationMAp.setLockObject()

I believe these could be seperated from the main PR, but should go in one PR, because they all require each other.

laeubi:
> I believe these could be seperated from the main PR, but should go in one PR, because they all require each other.

Sure at laest I think these are more "cleanup" and currently make the PR harder to review than it should.

iloveeclipse:
I've rebased on https://github.com/eclipse-platform/eclipse.platform.ui/pull/3399 state, fixed merge conflicts & my own comment above.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
iloveeclipse: Done

iloveeclipse:
Test failures on Windows are unrelated.

@mehmet-karaman, @laeubi : please feel free to review.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -680,7 +679,7 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 			ArrayList<Annotation> deleted= new ArrayList<>();
 			IAnnotationMap annotations= getAnnotationMap();
 			Object mapLock = annotations.getLockObject();
-
+			Assert.isNotNull(mapLock);
```
laeubi: Seems obsolete now we require to always return non null value

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -693,10 +692,12 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 				removeAnnotations(deleted, false, false);
 				synchronized (getLockObject()) {
```
laeubi, line 692:
```suggestion
				synchronized (mapLock) {
```

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -760,7 +761,11 @@ private Iterator<Annotation> getRegionAnnotationIterator(int offset, int length,
 
 		try {
 			Position[] positions= document.getPositions(IDocument.DEFAULT_CATEGORY, offset, length, canStartBefore, canEndAfter);
-			return new AnnotationsInterator(positions, fPositions);
+			IdentityHashMap<Position, Annotation> positionsMap;
+			synchronized (getLockObject()) {
+				positionsMap= new IdentityHashMap<>(fPositions);
```
laeubi, line 765:
```suggestion
				positionsMap= Map.copyOf(fPositions);
```

laeubi (approved):
It looks ok in general but I have not testes it.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
merks: The `positionsMap` variable is of type `IdentityHashMap<Position, Annotation>` so this suggestion won't compile and `fPositions` is also of type `IdentityHashMap<Position, Annotation>.  So I think it needs to remain like this.

iloveeclipse: Yes

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
mehmet-karaman: removed that

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
mehmet-karaman: done.

mehmet-karaman:
Resolved two remaining comments and rebased on latest master.

iloveeclipse:
@mehmet-karaman : you have now two commits, I assume by mistake. Can you please have **one**?

mehmet-karaman:
squashed the two commits.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -691,12 +690,14 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 
 			if (fireModelChanged && forkNotification) {
 				removeAnnotations(deleted, false, false);
-				synchronized (getLockObject()) {
+				synchronized (mapLock) {
```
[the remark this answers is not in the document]
iloveeclipse: This was a deliberate "safe" choice in commit 1a1eb948c1ede602fb7474bbbd8162ea7d11a14b. Now we can switch to `getLockObject()`.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -1006,13 +1018,21 @@ public IAnnotationModel getAnnotationModel(Object key) {
...
+		synchronized (getLockObject()) {
+			ret= fAttachments.remove(key);
+			if (ret == null) {
+				return null;
+			}
+		}
+		IDocument document= fDocument;
+		if (document != null) {
 			for (int i= 0; i < fOpenConnections; i++) {
 				ret.disconnect(fDocument);
```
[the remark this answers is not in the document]
iloveeclipse: Correct.

iloveeclipse (changes requested):
Funny enough, the very first time I've started IDE from debugger with this patch, IDE was frozen immediately with this deadlock:

```
"main":
	at org.eclipse.jface.text.source.AnnotationModel.getRegionAnnotationIterator(AnnotationModel.java:765)
	- waiting to lock <0x0000000088000000> (a java.lang.Object)
	at org.eclipse.jface.text.source.AnnotationModel.getAnnotationIterator(AnnotationModel.java:722)
	at org.eclipse.jface.text.source.inlined.InlinedAnnotationSupport$UpdateStylesWidth.applyTextPresentation(InlinedAnnotationSupport.java:108)
	at org.eclipse.jface.text.TextViewer.changeTextPresentation(TextViewer.java:4828)
	at org.eclipse.jdt.internal.ui.javaeditor.SemanticHighlightingPresenter.updatePresentation(SemanticHighlightingPresenter.java:245)
	at org.eclipse.jdt.internal.ui.javaeditor.SemanticHighlightingPresenter.lambda$0(SemanticHighlightingPresenter.java:149)
	at org.eclipse.jdt.internal.ui.javaeditor.SemanticHighlightingPresenter$$Lambda/0x000000001ca289b0.run(Unknown Source)
	at org.eclipse.jdt.internal.ui.javaeditor.SemanticHighlightingReconciler.lambda$3(SemanticHighlightingReconciler.java:670)
	- locked <0x0000000088034398> (a java.lang.Object)
	at org.eclipse.jdt.internal.ui.javaeditor.SemanticHighlightingReconciler$$Lambda/0x000000001ca28be0.run(Unknown Source)
	- locked <0x0000000088000010> (a org.eclipse.swt.widgets.RunnableLock)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5079)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4548)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:824)
	at org.eclipse.jface.window.Window.open(Window.java:804)
	at org.eclipse.search.internal.ui.OpenSearchDialogAction.run(OpenSearchDialogAction.java:60)
	at org.eclipse.search.ui.NewSearchUI.openSearchDialog(NewSearchUI.java:304)
	at org.eclipse.search.internal.ui.OpenFileSearchPageAction.run(OpenFileSearchPageAction.java:51)
	at org.eclipse.ui.internal.PluginAction.runWithEvent(PluginAction.java:239)
	at org.eclipse.ui.internal.WWinPluginAction.runWithEvent(WWinPluginAction.java:218)
	at org.eclipse.jface.action.ActionContributionItem.handleWidgetSelection(ActionContributionItem.java:581)
	at org.eclipse.jface.action.ActionContributionItem.lambda$4(ActionContributionItem.java:415)
	at org.eclipse.jface.action.ActionContributionItem$$Lambda/0x000000001c4815b0.handleEvent(Unknown Source)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5889)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1656)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5104)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4545)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1147)
	at org.eclipse.ui.internal.Workbench$$Lambda/0x000000001c1b7658.run(Unknown Source)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:185)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:219)
	at java.lang.invoke.DirectMethodHandle$Holder.invokeStatic(java.base@25/DirectMethodHandle$Holder)
	at java.lang.invoke.LambdaForm$MH/0x000000001c04f800.invoke(java.base@25/LambdaForm$MH)
	at java.lang.invoke.LambdaForm$MH/0x000000001c04fc00.invokeExact_MT(java.base@25/LambdaForm$MH)
	at jdk.internal.reflect.DirectMethodHandleAccessor.invokeImpl(java.base@25/DirectMethodHandleAccessor.java:156)
	at jdk.internal.reflect.DirectMethodHandleAccessor.invoke(java.base@25/DirectMethodHandleAccessor.java:104)
	at java.lang.reflect.Method.invoke(java.base@25/Method.java:565)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:615)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:563)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1415)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1387)
"Worker-10: Override indicator installation job":
	at org.eclipse.core.internal.filebuffers.SynchronizableDocument.addPosition(SynchronizableDocument.java:306)
	- waiting to lock <0x0000000088034398> (a java.lang.Object)
	at org.eclipse.jface.text.AbstractDocument.addPosition(AbstractDocument.java:364)
	at org.eclipse.jface.text.source.AnnotationModel.addPosition(AnnotationModel.java:492)
	at org.eclipse.jface.text.source.AnnotationModel.addAnnotation(AnnotationModel.java:462)
	at org.eclipse.jface.text.source.AnnotationModel.replaceAnnotations(AnnotationModel.java:432)
	at org.eclipse.jface.text.source.AnnotationModel.replaceAnnotations(AnnotationModel.java:401)
	at org.eclipse.jdt.internal.ui.javaeditor.OverrideIndicatorManager.updateAnnotations(OverrideIndicatorManager.java:218)
	- locked <0x0000000088000000> (a java.lang.Object)
	at org.eclipse.jdt.internal.ui.javaeditor.OverrideIndicatorManager.reconciled(OverrideIndicatorManager.java:261)
	at org.eclipse.jdt.internal.ui.javaeditor.ClassFileEditor$1.run(ClassFileEditor.java:796)

Found 1 deadlock.
```

So having all the tests green doesn't mean we are fine :-)

iloveeclipse:
This is javadoc from `org.eclipse.jface.text.source.AnnotationModel`

>  All modifications of the model's internal **annotation map** are synchronized using the model's lock object.

This is javadoc from `org.eclipse.jface.text.ISynchronizable`:

> In order to reduce the probability of dead locks clients should synchronize their access to these objects by using the provided lock object

So the code in `AnnotationModel` "can" use lock object returned by `getLockObject()` only to protect the  **annotation map** but **not the model itself**! OMG!

So existing clients (like seen in the deadlock above) can freely use model lock object to lock at any time on any thread **outside** the model code. The problem is built-in "by design". No one should use object **internal** lock from outside to synchronize on it as `org.eclipse.jface.text.ISynchronizable` prpoposes, because there is no guarantee what the client code does and which locks it will acquire while hoding object internal lock...

I don't know why that was designed this way, but this is an invite for trouble, anti-pattern of MT code.

In the example deadlock we see that
- `OverrideIndicatorManager.updateAnnotations()` obtains a lock on the model, model calls `document.addPosition()` that waits on a lock on document
- `SemanticHighlightingReconciler.updatePresentation()` obtains a lock on the document object and waits on a lock to the model in `getRegionAnnotationIterator()`.

The problem is that client code (as seen in the deadlock stack) can hold other locks while calling into model, and model code can request locks to code "outside" in multiple places like interacting with document or sending change events.

In the PR beside other changes we've added extra synchronization (via `getLockObject()`) in following methods that didn't acquired lock before:
- `getRegionAnnotationIterator()` 
- `getAnnotationIterator()`
- `addAnnotationModel()`
- `removeAnnotationModel()`

These locks were added because code inside can be affected by MT changes on internal data `fAttachments` and `fPositions`. 

The deadlock we see happened in `getRegionAnnotationIterator()`. However, it could happen in any of hte four methods above, and it is just matter of time we might see another deadlock if using  `getLockObject()`. 

One solution would be to change (break) API and **do not expose internal object lock** - but this would require refactoring of **all** clients code, which is simply not possible.

Another solution would be to introduce **yet another** lock object to protect `AnnotationModel` itself from concurrent modifications. This was avoided in 2619fb4e18c333733171f051e12c442d7be4dbf6 that tried to deal with same problems by using `ConcurrentHashMap`, but the `ConcurrentHashMap` only garantees **internal** consistency, not the consistency of iteration operations on the map like `fAttachments.get(it.next()`.

But intoducing **yet another** lock object would not reduce MT problems, it would rather increase the deadlock possibility, given the current code state.

**So if we would make the code in `AnnotationModel` properly synchronized it will deadlock soooner or later!**

So back to the original problem with exceptions like in https://github.com/eclipse-xtext/xtext/issues/3524 - if we can't make `AnnotationModel` thread safe by design, which other choises we have?

This leads us to possible **workaround** (I can't call it solution anymore): while using `fAttachments` and `fPositions`, don't use iterators but iterate over the map only via one from multiple `ConcurrentHashMap.forEach...` methods, which probably could fix https://github.com/eclipse-xtext/xtext/issues/3524 but of course would never guarantee any real consistency of the underlined data (which is also the case already). Also `addAnnotationModel()` is "special" as it checks for `containsValue()` before inserting something, so the code probably should be rewritten via `forEachValue()` ... Non trivial too.

And the last workaround I can imagine would be: since we know the code in twe methods above that requires locks needs locks or otherwise would run into troubles, we could try to **retry** operations inside the methods if an exceptions like `NoSuchElementException` or  `NullPointerException`  happens. something like:

```
	private Iterator<Annotation> getAnnotationIterator(boolean cleanup, boolean recurse) {
		Iterator<Annotation> iter= getAnnotationIterator(cleanup);
		if (!recurse || fAttachments.isEmpty()) {
			return iter;
		}
		List<Iterator<Annotation>> iterators = null;
		while (iterators == null) {
			try {
				iterators= getAnnotationIterator(iter);  // can't use lock here...
			} catch (NullPointerException | NoSuchElementException e) {
				// ignore and retry
			}
		}
		return new MetaIterator<>(iterators.iterator());
	}
	private List<Iterator<Annotation>> getAnnotationIterator(Iterator<Annotation> iter) {
		List<Iterator<Annotation>> iterators= new ArrayList<>(fAttachments.size() + 1);
		iterators.add(iter);
		Iterator<Object> it= fAttachments.keySet().iterator();
		while (it.hasNext()) {
			iterators.add(fAttachments.get(it.next()).getAnnotationIterator());
		}
		return iterators;
	}
```

It will not make anything thread safe, but clients already deal with not thread safe implemantation of `AnnotationModel`, they just will not need to catch exceptions like `catch (NullPointerException | NoSuchElementException e)` in their own code...

So there are now two workarounds possible ut no real solution in sight. I would probably first look into removing added locks in the methods below and replacing whatever operations were used to concurrent operations provided by one of  `ConcurrentHashMap.forEach...` methods and similar.
- `getRegionAnnotationIterator()` 
- `getAnnotationIterator()`
- `addAnnotationModel()`
- `removeAnnotationModel()`

I **believe** other code that is changed in this PR could stay changed as proposed, but I need more time to understand all implications considering the four methods above can't be made MT safe anymore.

laeubi:
It feels from the description that one needs a complete rework of the data model instead... in any case I think the idea with the lock object is/was that one can prevent modifications or inhibit events accross multiple objects.

Regarding the Xtext case, maybe the *caller* should already lock on the lock object itself when it enters a critical section then and be carefull to not hold any additional locks?

mehmet-karaman:
Seems this wasn't the first approach to improve the AnnotationModel.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/892

PR #3399, opened 2025-10-13
Code cleanup in AnnotationModel, AnnotationMap, Javadoc inherit adjustment

mehmet-karaman:
- split of the PR: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3397
- Removed dead code in AnnotationModel.cleanup because mapLock can't be null.
- Added new javadoc to IAnnotationMap.getLockObject()
- added assert.isLegal check to AnnotationMAp.setLockObject()

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
@@ -21,6 +21,8 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 
+import org.eclipse.core.runtime.ILog;
+
 import org.eclipse.jface.text.Position;
 
 
@@ -54,11 +56,38 @@ class AnnotationMap implements IAnnotationMap {
 		fInternalMap= new HashMap<>(capacity);
 	}
 
+	/**
+	 * Sets the lock object for this object. Subsequent calls to specified methods of this object
+	 * are synchronized on this lock object. Which methods are synchronized is specified by the
+	 * implementer.
+	 * <p>
+	 * If the lock object is not explicitly set by this method, the annotation map uses its internal
+	 * lock object for synchronization.
+	 * </p>
+	 * 
+	 * <p>
+	 * <em>You should not override an existing lock object unless you own that lock object yourself.
+	 * Use the existing lock object instead.</em>
+	 * </p>
+	 *
+	 * @param lockObject the lock object. If <code>null</code> is given, internal lock object is
+	 *            used for locking.
+	 */
 	@Override
 	public synchronized void setLockObject(Object lockObject) {
+		if(fLockObject != null && fLockObject != lockObject) {
+			ILog log= ILog.of(AnnotationMap.class);
+			String message = "Attempt to override existing lock object of AnnotationMap."; //$NON-NLS-1$
+			log.warn(message, new IllegalStateException(message));
+		}
 		fLockObject= lockObject;
 	}
 
+	/**
+	 * Lock object used to synchronize concurrent access to the internal map data.
+	 * 
+	 * @return <b>never</b> returns null
+	 */
 	@Override
 	public synchronized Object getLockObject() {
 		if (fLockObject == null) {
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.core.runtime.Assert;
@@ -338,11 +339,30 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 		return (IAnnotationMap) fAnnotations;
 	}
 
+	/**
+	 * Subclasses should <b>never</b> return null. Clients should use the lock
+	 * object in order to synchronize concurrent access to the internal map data.
+	 * 
+	 * @return <b>never</b> returns null
+	 */
 	@Override
 	public Object getLockObject() {
-		return getAnnotationMap().getLockObject();
+		return Objects.requireNonNull(getAnnotationMap().getLockObject());
 	}
 
+	/**
+	 * Sets the lock object for this object. Subsequent calls to specified methods of this object
+	 * are synchronized on this lock object. Which methods are synchronized is specified by the
+	 * implementer.
+	 * 
+	 * <p>
+	 * <em>You should not override an existing lock object unless you own that lock object yourself.
+	 * Use the existing lock object instead.</em>
+	 * </p>
+	 *
+	 * @param lockObject the lock object. If <code>null</code> is given, default implementation uses
+	 *            the internal lock object for locking.
+	 */
 	@Override
 	public void setLockObject(Object lockObject) {
 		getAnnotationMap().setLockObject(lockObject);
@@ -661,23 +681,12 @@ public class AnnotationModel implements IAnnotationModel, IAnnotationModelExtens
 			IAnnotationMap annotations= getAnnotationMap();
 			Object mapLock = annotations.getLockObject();
 
-			if (mapLock == null) {
-				Iterator<Annotation> e= annotations.keySetIterator();
-				while (e.hasNext()) {
-					Annotation a= e.next();
-					Position p= annotations.get(a);
+			synchronized (mapLock) {
+				annotations.forEach((a, p) -> {
 					if (p == null || p.isDeleted()) {
 						deleted.add(a);
 					}
-				}
-			} else {
-				synchronized (mapLock) {
-					annotations.forEach((a, p) -> {
-						if (p == null || p.isDeleted()) {
-							deleted.add(a);
-						}
-					});
-				}
+				});
 			}
 
 			if (fireModelChanged && forkNotification) {
--- a/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
@@ -79,4 +79,28 @@ public interface IAnnotationMap extends Map<Annotation, Position>, ISynchronizab
 	 */
 	@Override
 	Collection<Position> values();
+	
+	/**
+	 * Implementers of this interface should <b>never</b> return null. Clients should use the lock
+	 * object in order to synchronize concurrent access to the implementer.
+	 * 
+	 * @return the lock object to be used, if no explicit lock object is set, internal one should
+	 *         be used, therefore never <code>null</code>
+	 */
+	@Override
+	Object getLockObject();
+	
+	/**
+	 * Sets the lock object for this object. Subsequent calls to specified methods of this object
+	 * are synchronized on this lock object. Which methods are synchronized is specified by the
+	 * implementer.
+	 * <p>
+	 * <em>You should not override an existing lock object unless you own that lock object yourself.
+	 * Use the existing lock object instead.</em>
+	 * </p>
+	 *
+	 * @param lockObject the lock object. Never <code>null</code>.
+	 */
+	@Override
+	void setLockObject(Object lockObject);
 }


mehmet-karaman:
@laeubi @iloveeclipse this is the first split of the original PR.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -661,23 +661,12 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 			IAnnotationMap annotations= getAnnotationMap();
 			Object mapLock = annotations.getLockObject();
```
laeubi, line 682:
```suggestion
			Object mapLock = Objects.requireNonNull( annotations.getLockObject(), "Lock object must not be null");
```

iloveeclipse:
Mehmet, please don't hesitate to **link** to original tickets or PR's. It is always an extra work to find out what is related.
In current case: original PR is https://github.com/eclipse-platform/eclipse.platform.ui/pull/3397

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
```diff
@@ -79,4 +79,12 @@ public interface IAnnotationMap extends Map<Annotation, Position>, ISynchronizab
...
 	@Override
 	Collection<Position> values();
+	
+	/**
+	 * {@inheritDoc}
+	 * Implementers of this interface should <b>never</b> return null. 
+	 * @return <b>never</b> returns null 
+	 */
+	@Override
+	Object getLockObject();
```
laeubi, line 91: Where is this coming from? The overridden API here allows explicitly `null` and I can not see from the current API contract that his was anywhere required.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
```diff
@@ -56,6 +58,7 @@ public AnnotationMap(int capacity) {
 
 	@Override
 	public synchronized void setLockObject(Object lockObject) {
+		Assert.isLegal(fLockObject == null || fLockObject == lockObject, "Lock was already set"); //$NON-NLS-1$
```
laeubi: Same here I can't find any reasoning why this is now required (and if javadoc needs to be updated)

laeubi (changes requested):
I miss a reasoning why we can change the API contract of these methods, it looks like a API breaking change.

Also commit message should clearly state that e.g.

```
Require non null LockObject in IAnnotationMap

- describe the reasoning here -
```

this is obviously not a cleanup / javadoc only thing!

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
mehmet-karaman: This comes from org.eclipse.jface.text.ISynchronizable which usually can return null.. But in case of the IAnnotationMap it returns either the lockObject (when != null or the fInternalLockObject).. There is only one callee in Annotation part which checks if its null or not and all use it as granted that it won't be null.. So this is just a comment to clarify the use in IAnnotationMap implementors. 

This is how we found the dead code

laeubi: 1. Maybe callers are implemented wrong
2. `IAnnotationMap` is an interface so other implementations are possible that behave differently and the interface is neither internal nor marked as noimplement.

So if these assertion might be true for the one implementations, they are not generally true for the APi contract.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
laeubi: This is even a different story, it is not forbidden to set the object to something else but actually discouraged.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
iloveeclipse: > The overridden API here allows explicitly null and I can not see from the current API contract that his was anywhere required.

The code in platform in most places assumed that the lock from `IAnnotationMap` is never null. 
If any client *would* use null, `AnnotationModel` as `Standard implementation of {@link IAnnotationModel} and its extension interfaces` would fail with NPE in multiple places.

So the change in  `AnnotationModel` is fine IMO. 

The only questionable point would be that years after `IAnnotationMap` API was introduced (and implemented), we are updating the API javadoc to by in-sync with the actual (original) behavior implementation. The change in `IAnnotationMap` clarifyes the overridden `getLockObject()` interface method behavior. This can be considered as changed spec (even if changed spec just follows implemented behavior), so we might bump the minor version segment on `org.eclipse.text` as well.

**If** we would do so, this would require minor version bump to be done in `org.eclipse.jface.text` as well (because it re-exports `org.eclipse.text`). I'm not sure the spec update here would justify that.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
iloveeclipse: The lock is used in synchronized blocks in the AnnotationModel. Today, if lock would change to be a different object, it would cause unpredictable side effects at runtime because it would allow two threads entering same "synchronized" block because it would lock on two different objects.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
laeubi: I think it is okay to state that not `null` is to be expected here, still its a bit weird to document implementation details. So if we go that direction it should be consistent and mention that it returns the user set lock object or an internal default object if no such thing exits.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
laeubi: But only if used inconsistently. The overridden method already states this is possible but discouraged, but there seem to be use-cases then.

Given that, one might ask why not set the object in the constructor then and make it final anyways?

mehmet-karaman: It is more a flavor.. At the end I want a solution what works.. What shall we do? Remove that line? Empty the body of that method?

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
laeubi: I have now checked usages of `org.eclipse.jface.text.ISynchronizable.getLockObject()` and all of them seem to contain proper null checks. The only one is the _internal implementation_ of `AnnotationModel` where it is assumed to be non-null what is because of how it is implemented.

So I think the javadoc here is misleading and we should just remove that completely (unless someone can show me a place where it is really used). I also don't see any advantage in this as `IAnnotationMap` seems not widely used in public code and where it is used we do not use getLockObject() at all.... so even if there where a guarantee one likely can not take advantage of it.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
laeubi: Looking through callsite of this I can't see any problem with that, is there any actual caller / bug / ... related to this?

At all places it seems only be used for init a document and ensure the same lock is e.g. used for some other `ISyncronizable`, so again all seems save to me and not needed to further restrict it for this use-case.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
laeubi: `getAnnotationMap()`; is a protected method in a public class, so I don't think there is a guarantee we always get our internal implementation!

Also the pattern here used matches the one in other places in platform so I would also not change this here even though the code likely is never used.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
iloveeclipse: I didn't get it, we are not discussing `ISynchronizable.getLockObject()` general use, only `IAnnotationMap` specification.

In order to remove the "dead" code we noticed in the `cleanup()`, we must be sure `IAnnotationMap.getLockObject()` returns non-null. If this is possible,  code in `cleanup()` is not dead code and we can't remove that but have to change all other code assuming the opposite (which is definitely not what we intent to do).

We also can't change the code in `org.eclipse.jface.text.source.AnnotationModel.getAnnotationMap()` to return something else and not `IAnnotationMap`, because this would break API.

So why can't we state that `IAnnotationMap.getLockObject()` is supposed to return non-null?

> So I think the javadoc here is misleading and we should just remove that completely (unless someone can show me a place where it is really used). I also don't see any advantage in this as IAnnotationMap seems not widely used in public code and where it is used we do not use getLockObject() at all.... so even if there where a guarantee one likely can not take advantage of it.

The piece of code that uses that is in this PR, it is the removed "dead" code.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
iloveeclipse: > Given that, one might ask why not set the object in the constructor then and make it final anyways?

Because we would break client code that sets the lock after object creation. See `org.eclipse.jface.text.source.SourceViewer.createVisualAnnotationModel(IAnnotationModel)` for creation and `org.eclipse.jface.text.source.SourceViewer.setDocument(IDocument, IAnnotationModel, int, int)` for the lock setting.

iloveeclipse: I would propose to add here a system property and only "assert" if the system property is set (not by default).

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
iloveeclipse: > Also the pattern here used matches the one in other places in platform

I'm guilty for this particular case, see  1a1eb948c1ede602fb7474bbbd8162ea7d11a14b, simply because I didn't wanted to rewrite the whole class synchronization at that time and wanted to have a "safe" fix, also because of the `IAnnotationMap.getLockObject()` was not properly specified / documented from the beginning.

laeubi:
In general I'm not sure it pays of to change anything here. Even if I can see that removing some dead code might be a good thing I wonder if it is worth the hassle here if it is "really dead".

In contrast we introduce some asymmetry here to other places that use the ISycnronizer API... so I'm not totally against this but would propose to simply leave it as is.

iloveeclipse:
The point is: for the reader of the `IAnnotationModel` & `AnnotationModel` API's it is not clear at all what happens here. Why there is one piece of code checking lock for null andthe rest of the code not, when the `ISynchronizable.getLockObject()` explicitly says it can be null. Is all the code buggy? Surely not, something seem to work somehow. Can the new code rely on lock being not null or not? No idea. 

So this PR clarifies that mess and helps the reader of next one (https://github.com/eclipse-platform/eclipse.platform.ui/pull/3397) to assume we are always working with non null lock object.

iloveeclipse:
> I'm not totally against this but would propose to simply leave it as is.

Is there anything in this PRT to be changed? If yes, please point which exact changes are needed so we can continue.

Otherwise remove the negative review, because from our understanding this PR makes sense in the context of the next one.

laeubi:
> he point is: for the reader of the `IAnnotationModel` & `AnnotationModel` API's it is not clear at all what happens here. Why there is one piece of code checking lock for null andthe rest of the code not

This is I think what is mainly a confusing of different levels of code and it is claimed without a proof.

e.g. I have checked that all *public* access to `IAnnotationModel` is doing proper `null` checks... `AnnotationModel` is not API and do not do null checks because its implementation ensures null is never returned.

This is completely common way and I do not see why implementation details need to show up in the public API ... so this is my primary concern.

iloveeclipse:
> `AnnotationModel` is not API

Why that?

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
mickaelistria: As a 1st iteration, that wouldn't require configuration, could the erroneous case be logged? If it is logged with a nice enough message, anyone could even add a LogListener in their app to capture the occurrences.
Would it be enough to troubleshot the initial problem?

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
```diff
@@ -56,6 +58,7 @@ public AnnotationMap(int capacity) {
 
 	@Override
```
mickaelistria, line 76: I think it would be worth adding here some `@implNote` javadoc explaining that for this particular implementation, method is expected to be called only once with a value (or allowing resetting to null).

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -661,23 +661,12 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
...
 
-			if (mapLock == null) {
-				Iterator<Annotation> e= annotations.keySetIterator();
-				while (e.hasNext()) {
-					Annotation a= e.next();
-					Position p= annotations.get(a);
+			synchronized (mapLock) {
```
mickaelistria, line 684: AFAIK, if the implementation is "incorrect" in term of expected but non-specified behavior (imagine a legacy incorrect impl) the result here would replace some non-breaking possible bug, by some NPE here.
What' the benefit in getting rid of the `mapLock == null` case if we cannot ensure from API and legacy that this case is non-existent?

mickaelistria:
> >  AnnotationModel is not API
>
> Why that?

It's package-visible only.

iloveeclipse:
> > > AnnotationModel is not API
> > 
> > 
> > Why that?
> 
> It's package-visible only.

**public** class AnnotationModel

mickaelistria:
My bad, I had (mis)read AnnotationMap...

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
iloveeclipse: > What' the benefit in getting rid of the mapLock == null case if we cannot ensure from API and legacy that this case is non-existent?

Looking around, the return from `AnnotationModel.getLockObject()` is dereferenced without null check 10 from 11 times in this class. This alone would be enough to clean up the code. But why do we have this one call? Because it was not properly documented  and the parent interface method `ISynchronizable.getLockObject()` **explicitly** says that `Returns the lock object or **null** if there is none.`

So for a responsible developer there is a dilemma: either interface contract in `ISynchronizable` or statistics in this class. As I already wrote, I had no interest to solve that particular dilemma while looking at unrelated preformence issue, so this particular "if" branch was added by me in 1a1eb94 as a "safe" fix.

With the current PR this dilemma is supposed to be fixed by updating Interface documentation & removing unnecessary "if" branch.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
iloveeclipse: OK, let by default do not "assert" but log a warning to the log.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationMap.java
iloveeclipse: Yes, good point.

iloveeclipse:
> This is I think what is mainly a confusing of different levels of code and it is claimed without a proof.

The proof is the current code in the  `AnnotationModel` class that dereferences the "possible null" lock objects 10 times out of 11 without null check.

> e.g. I have checked that all public access to IAnnotationModel is doing proper null checks... 

You have most likely checked `ISynchronizable.getLockObject()` callers, not `AnnotationModel.getLockObject()` callers.
This calls aren't doing null checks:
```
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.CompilationUnitAnnotationModel.addAnnotation(Annotation, Position, boolean)
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.CompilationUnitAnnotationModel.getJavaMarkerAnnotations(Position)
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.CompilationUnitAnnotationModel.internalBeginReporting(boolean)
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.CompilationUnitAnnotationModel.internalEndReporting(ProblemRequestorState)
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.CompilationUnitAnnotationModel.removeAllAnnotations(boolean)
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.CompilationUnitAnnotationModel.removeAnnotation(Annotation, boolean)
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.CompilationUnitAnnotationModel.reportProblems(List<IProblem>)
org.eclipse.pde.internal.ui.editor.text.XMLReconcilingStrategy.SpellingProblemCollector.endCollecting()
org.eclipse.ui.texteditor.spelling.SpellingReconcileStrategy.SpellingProblemCollector.endCollecting()
org.eclipse.cdt.internal.ui.editor.CDocumentProvider.TranslationUnitAnnotationModel.addAnnotation(Annotation, Position, boolean)
org.eclipse.cdt.dsf.debug.internal.ui.disassembly.model.BreakpointsAnnotationModel.breakpointChanged(IBreakpoint, IMarkerDelta)
org.eclipse.cdt.internal.ui.editor.CDocumentProvider.TranslationUnitAnnotationModel.getAnnotations(Position)
org.eclipse.cdt.internal.ui.editor.CDocumentProvider.TranslationUnitAnnotationModel.internalBeginReporting(boolean)
org.eclipse.cdt.internal.ui.editor.CDocumentProvider.TranslationUnitAnnotationModel.internalEndReporting(ProblemRequestorState)
org.eclipse.xtext.ui.editor.quickfix.XtextResourceMarkerAnnotationModel.queueAnnotationChanged(Annotation)
org.eclipse.cdt.internal.ui.editor.CDocumentProvider.TranslationUnitAnnotationModel.removeAllAnnotations(boolean)
org.eclipse.cdt.internal.ui.editor.CDocumentProvider.TranslationUnitAnnotationModel.removeAnnotation(Annotation, boolean)
org.eclipse.cdt.internal.ui.editor.CDocumentProvider.TranslationUnitAnnotationModel.reportProblems(List<IProblem>)
```

> AnnotationModel is not API and do not do null checks because its implementation ensures null is never returned.

It **is** API and it **should** do null checks on `IAnnotationMap getAnnotationMap()` return value because it could be overridden, but it never did.


OK, I will push a PR update that will:

1) log warning instead of assert
2) Improve javadoc of the `IAnnotationModel.getLockObject()` method to mention that changing the lock object after setting it is not expected and mght lead to synchronization problems
3) Add javadoc to `AnnotationModel.getLockObject()` and `AnnotationModel.setLockObject()` methods matching the javadoc of `IAnnotationModel.getLockObject()`.
3) Rebase on master
4) Update commit message

iloveeclipse:
@laeubi , @mickaelistria : could you please re-review?

iloveeclipse:
> @laeubi , @mickaelistria : could you please re-review?

Ping.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -338,11 +338,30 @@ protected IAnnotationMap getAnnotationMap() {
...
 
+	/**
+	 * Subclasses should <b>never</b> return null. Clients should use the lock
+	 * object in order to synchronize concurrent access to the internal map data.
+	 * 
+	 * @return <b>never</b> returns null
+	 */
 	@Override
 	public Object getLockObject() {
 		return getAnnotationMap().getLockObject();
```
laeubi:
```suggestion
		return Objects.requireNonNull(getAnnotationMap().getLockObject());
```

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -338,11 +338,30 @@ protected IAnnotationMap getAnnotationMap() {
 		return (IAnnotationMap) fAnnotations;
 	}
 
+	/**
+	 * Subclasses should <b>never</b> return null. Clients should use the lock
+	 * object in order to synchronize concurrent access to the internal map data.
+	 * 
+	 * @return <b>never</b> returns null
```
laeubi, line 346: do we need the replicate the javadoc from the interface here?

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
```diff
@@ -79,4 +79,27 @@ public interface IAnnotationMap extends Map<Annotation, Position>, ISynchronizab
 	 */
 	@Override
 	Collection<Position> values();
+	
+	/**
+	 * Implementers of this interface should <b>never</b> return null. Clients should use the lock
+	 * object in order to synchronize concurrent access to the implementer.
+	 * 
+	 * @return <b>never</b> returns null
```
laeubi:
```suggestion
	 * @return the lock object to be used, if no explicit lock object is set and internal one will be used, therefore never <code>null</code>
```

laeubi (approved):
Give the limited scope of this it looks "safe" to do, I added some style comments that might or might not be considered.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
iloveeclipse: OK

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
iloveeclipse: Yes, bacause the `IAnnotationMap` interface isn't implemented here.

bundles/org.eclipse.text/src/org/eclipse/jface/text/source/IAnnotationMap.java
iloveeclipse: OK.

iloveeclipse (approved):

PR #3402, opened 2025-10-14
Add a command to persist workbench state

### LINKED ISSUE #3398: A command to store/persist current workbench state

## Suggestion

* I tweak some interesting things in the workspace (open files, breakpoints...)

Then for some (external, OS related, not topic of this issue) reason, my workbench crashes hard, and I restart after a system KILL operation.

* Many pieces my former workbench tweaks are lost (but I cannot expect them to be persisted after the crash/KILL)

I know I can reduce the "workbench save interval from the preferences, and that does already help.
But as I know when such cases are more likely to happen, I'd like to be able to manually trigger a saved state of the workbench (at least with a command accessible via Ctrl+3).

## Community

mickaelistria:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3398

This command can be tried the following way:

Base

0. Start an IDE with this patch
1. Close some views
2. Kill the IDE (brutally, not with the nice "exit" button)
3. Restart: ❌ Closed views are here again

Diff

0. Start an IDE with this patch
1. Close some views
2. Ctrl+3 > "Persist workbench state"
3. Kill the IDE (brutally, not with the nice "exit" button)
4. Restart: ✔️ Closed views are not there

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/PersistWorkbenchHandler.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/PersistWorkbenchHandler.java
@@ -1,0 +1,29 @@
+/*******************************************************************************
+ * Copyright (c) 2025, Red Hat, Inc. and others.
+ *
+ * This program and the accompanying materials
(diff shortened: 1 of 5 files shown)


BeckerWdf:
I think this is a nice an pragmatic addition

PR #3407, opened 2025-10-17
Enable save actions for o.e.text

akurtakov:
PR https://github.com/eclipse-platform/eclipse.platform.ui/pull/3399 introduced trailing whitespaces and PR
https://github.com/eclipse-platform/eclipse.platform.ui/pull/3406 was created by the bot to clean this up.
With save actions we should prevent such cases.


akurtakov:
Windows test failures are unstable tests as enabling save actions is not code change.

HeikoKlare:
> Windows test failures are unstable tests as enabling save actions is not code change.

Still a bit strange, as those tests are not known to be flaky and I have never seen exactly these fail before. But since yesterday they seem to fail quite constantly in every build.

PR #2641, opened 2024-12-20
Add CloseTestWindowsRule

akurtakov:
To remaining tests identified. Fixes
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2615

PR #3409, opened 2025-10-17
Migrate JUnit 4 tests to JUnit 5 in org.eclipse.jface.text.tests

vogella:


vogella:
I assume the test failure are related to

[#2636](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2636) or [#2641](https://github.com/eclipse-platform/eclipse.platform.swt/issues/2641)

vogella:
@fedejeanne @akurtakov @HeikoKlare @HannesWell 

Migration was done with Claude Code (which IMHO is currently the best AI engine for coding tasks). Changes look fine to me and the number of tests is still the same. Any concerns? Otherwise I start merging such changes if the build is green and the test number is the same (of course after manually checking for "strange" changes).

HeikoKlare:
In general if the changes stand up to manual review, I don't see any reason for not merging them.
I see some situations in which I would be more hesitant:
- Everything related to session tests need a very thorough analysis and is probably better done manually as we have a completely different testing concept/framework for them with JUnit 5.
- Whenever test class hierarchies are involved, as I have seen very few hierarchies that were really appropriate and most of them were better replaced by utilities, delegation, test rules/extensions etc.
- Whenever test rules are involved, they need to be properly migrated to test extensions (or some other concept).

There might of course be different on opinions on how much cleanup such a migration should involve. This pretty much depends on the incentive or goal for doing a migration. For me, migrating to JUnit 5 was never (only) motivated by moving to a more modern framework, but I always took this as a chance to improve the test design and in particular test isolation. In particular in the Platform repository, test isolation was pretty bad before I migrated most of the stuff from JUnit 3/4 to JUnit 5 and the biggest win (and also my essential goal) were more stable tests due to improved isolation and not the migration to JUnit 5 itself.

vogella:
Thanks @HeikoKlare for the detailed feedback. I agree with your goals and think AI can help with these too. In addition I see the following goals for me:

* Using JUnit5 / JUnit6 should help with isolating and improving the tests in the future
* Using modern test frameworks make the code base more interesting to new contributors
* Migrating tests which can be easily migrated by AI tools makes the remaining problems easier to spot to us
* Rule migration or test isolation should also be possible with AI tools 

 See also https://github.com/eclipse-platform/eclipse.platform.ui/issues/3414 for some background analysis Claude did before starting to work on the task.

vogella:
Change LGTM, planning to merge early next week

fedejeanne (approved):

vogella (commented):
The migration to JUnit 5 in the touched files looks correct. The assertion arguments were correctly swapped, and the lifecycle annotations were updated properly.

Note: I verified that there are still other tests in this package (e.g., `TextViewerTest.java`) using JUnit 4, so this is a partial migration.

Issue #196, opened 2022-07-07
[Test] org.eclipse.ui.tests.navigator.resources.NestedResourcesTests.testProjectHierarchy fails on linux intermittently


ktatavarthi:
The test  org.eclipse.ui.tests.navigator.resources.NestedResourcesTests.testProjectHierarchy failed on linux in the I-Build [I20220706-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220706-1800/testresults/html/org.eclipse.ui.tests.navigator_ep425I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html) with the below error

```
expected:<2> but was:<1>

java.lang.AssertionError: expected:<2> but was:<1>
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.failNotEquals(Assert.java:835)
at org.junit.Assert.assertEquals(Assert.java:647)
at org.junit.Assert.assertEquals(Assert.java:633)
at org.eclipse.ui.tests.navigator.resources.NestedResourcesTests.testProjectHierarchy(NestedResourcesTests.java:98) 
```

fedejeanne:
It failed on Mac too

PR #3410, opened 2025-10-17
Migrate JUnit 4 tests to JUnit 5 in org.eclipse.jface.text.tests

vogella:


tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/FilteringAsyncContentAssistTests.java
```diff
@@ -288,7 +288,8 @@ public void testCA_WithFirstDelayedThenImmediateProposals() throws Exception {
 	 *
 	 * @throws Exception exception
 	 */
-	@Test @Ignore
+	@Test
+	@Disabled("Bug: filtering only applied after all CA processors have completed")
```
fedejeanne, line 292: The text inside `@Disabled` is new. Is this intentional?

tests/org.eclipse.text.tests/META-INF/MANIFEST.MF
```diff
@@ -15,6 +15,7 @@ Require-Bundle:
  org.eclipse.text;bundle-version="[3.6.3,4.0.0)",
  org.junit;bundle-version="4.12.0"
 Import-Package: org.junit.jupiter.api;version="[5.14.0,6.0.0)",
+ org.junit.jupiter.api.function;version="[5.14.0,6.0.0)",
```
fedejeanne: Is this one really needed? I don't see it imported anywhere in the plugin. I removed it and everything compiles. 
```suggestion
```

fedejeanne (commented):
Looks pretty well, I only spotted 2 minor things

tests/org.eclipse.text.tests/META-INF/MANIFEST.MF
merks: Yes but do the tests run properly without it?  @HannesWell FYI

fedejeanne: Hm, good question, Ed. It seems this is not the only plugin that "magically" needs this import. I only see 1 plugin in my WS for which the package is both imported _and_ used in the code. The rest merely imports it in the manifest:

image

merks:
Yes, I fixed some non-running tests to add the range.  But now the JDT's junit runner also has ranges so not entirely clearly what is needed where.  I though it would be good to mention it...

tests/org.eclipse.text.tests/META-INF/MANIFEST.MF
HeikoKlare: This import is always required when `assertThrows` is used, because that method expects an `org.junit.jupiter.api.function.Executable`. Otherwise test execution will yield:
```
java.lang.NoClassDefFoundError: org/junit/jupiter/api/function/Executable
   ...
```

fedejeanne: Ah, that makes sense. I didn't see the usages in the code (no `import` declarations) because the method is fed a lambda expression 👍

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/FilteringAsyncContentAssistTests.java
vogella: Yes, the AI tooling tried to identify the reason why this test is disabled and the reasoning seemed ok for me.

vogella:
Update this PR to only include changes for org.eclipse.jface.text.tests

vogella:
No idea what this error message means:


Tests run: 32, Failures: 0, Errors: 0, Skipped: 0

CoreSVG has logged an error. Set environment variabe "CORESVG_VERBOSE" to learn more.

fedejeanne (approved):
Test failures are known:
- #1517 
- #196

vogella:
> Test failures are known:
> 
> * #1517
> * #196

Use Claude code to implement fixes for both old issues, the fixes look good to me.

* https://github.com/eclipse-platform/eclipse.platform.ui/pull/3429
* https://github.com/eclipse-platform/eclipse.platform.ui/pull/3430

PR #3416, opened 2025-10-20
Fix #3405 Fix incorrect reuse of code mining annotations with mismatched types

### LINKED ISSUE #3405: Code minings are not correctly redrawn when switching between in-line and line header mode
## Issue description

I'm working on an Eclipse plug-in that creates code minings. While testing these code minings, I found out that switching between _in-line code minings_ and _line header code minings_ (in my case by changing preference settings) causes some minings not to be redrawn.

In more detail:

1. case: When I have in-line code minings in my editor and I switch preferences to use line header code minings instead, some code minings are redrawn as line header code minings, but some stay beeing drawn as in-line code minings, instead of redrawing all code minings as line header code minings.
2. case: When I have line header code minings in my editor and I switch preferences to use in-line code minings instead, some code minings are redrawn as in-line code minings, but some stay beeing drawn as line header code minings, instead of redrawing all code minings as in-line code minings.

## First analysis

I debugged the executed code and found out that switching preferences between in-line and line header code minings resulted in correctly re-creating all code minings, but the `org.eclipse.jface.internal.text.codemining.CodeMiningManager` mixes up `CodeMiningLineContentAnnotation`s with `CodeMiningLineHeaderAnnotation`s. If the manager finds an existing code mining annotation for some position, it reuses the annotation without checking if it has the right type. It seems, the `CodeMiningManager` should instead replace `CodeMiningLineContentAnnotation`s with `CodeMiningLineHeaderAnnotation`s when switching from in-line code minings to line header code minings and vice versa.

## Steps to reproduce

Unfortunately, the plug-in I'm working on is not open source. Thus, I cannot provide simple steps to reproduce the issue.

In my case, I have a file with plenty code minings, often multiple code minings per line and often several consecutive lines with at least one code mining per line. Having this file open in an editor and then switching the code mining preferences for the custom editor causes the issue.

* Eclipse 2025-03 (started via launch configuration from another instance of the IDE)
* Windows 11 Pro

travkin79:
There are three types of code minings and three corresponding types of annotations. Issue #3405 describes a problem with updating the annotations. In certain cases the annotations are not re-created if the code mining type changes, e.g. from in-line code minings to line header code minings. This PR fixes that issue.

In order to make reviews easy, I left three commits for three different changes. The commits can be squashed if you prefer having only one commit.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
@@ -10,6 +10,7 @@
  *
  *  Contributors:
  *  Angelo Zerr <angelo.zerr@gmail.com> - [CodeMining] Provide CodeMining support with CodeMiningManager - Bug 527720
+ *  Dietrich Travkin <dietrich.travkin@solunar.de> - Fix code mining redrawing - Issue 3405
  */
 package org.eclipse.jface.internal.text.codemining;
 
@@ -230,6 +231,41 @@ public class CodeMiningManager implements Runnable {
 				Collectors.mapping(Function.identity(), Collectors.toList())));
 	}
 
+	private static enum CodeMiningMode {
+		InLine(LineContentCodeMining.class, CodeMiningLineContentAnnotation.class),
+		HeaderLine(LineHeaderCodeMining.class, CodeMiningLineHeaderAnnotation.class),
+		FooterLine(DocumentFooterCodeMining.class, CodeMiningDocumentFooterAnnotation.class);
+
+		public final Class<? extends ICodeMining> codeMiningType;
+		public final Class<? extends AbstractInlinedAnnotation> annotationType;
+
+		CodeMiningMode(Class<? extends ICodeMining> codeMiningType,
+				Class<? extends AbstractInlinedAnnotation> annotationType) {
+			this.codeMiningType= codeMiningType;
+			this.annotationType= annotationType;
+		}
+
+		public static CodeMiningMode createFor(List<ICodeMining> minings) {
+			Assert.isNotNull(minings);
+
+			CodeMiningMode mode= CodeMiningMode.HeaderLine;
+			if (!minings.isEmpty()) {
+				ICodeMining first= minings.get(0);
+
+				if (CodeMiningMode.InLine.codeMiningType.isInstance(first)) {
+					mode= CodeMiningMode.InLine;
+				} else if (CodeMiningMode.HeaderLine.codeMiningType.isInstance(first)) {
+					mode= CodeMiningMode.HeaderLine;
+				} else if (CodeMiningMode.FooterLine.codeMiningType.isInstance(first)) {
+					mode= CodeMiningMode.FooterLine;
+				} else {
+					mode= CodeMiningMode.InLine;
+				}
+			}
+			return mode;
+		}
+	}
+
 	/**
 	 * Render the codemining grouped by line position.
 	 *
@@ -257,11 +293,13 @@ public class CodeMiningManager implements Runnable {
 			Position pos= new Position(g.getKey().offset, g.getKey().length);
 			List<ICodeMining> minings= g.getValue();
 			ICodeMining first= minings.get(0);
-			boolean inLineHeader= !minings.isEmpty() ? (first instanceof LineHeaderCodeMining) : true;
+
+			CodeMiningMode mode= CodeMiningMode.createFor(minings);
+
 			// Try to find existing annotation
 			AbstractInlinedAnnotation ann= fInlinedAnnotationSupport.findExistingAnnotation(pos);
-			if (ann == null) {
-				// The annotation doesn't exists, create it.
+			if (ann == null || !mode.annotationType.isInstance(ann)) {
+				// The annotation doesn't exists or has wrong type => create a new one.
 				boolean afterPosition= false;
 				if (first instanceof LineContentCodeMining m) {
 					afterPosition= m.isAfterPosition();
@@ -274,16 +312,14 @@ public class CodeMiningManager implements Runnable {
 					mouseOut= first.getMouseOut();
 					mouseMove= first.getMouseMove();
 				}
-				if (inLineHeader) {
-					ann= new CodeMiningLineHeaderAnnotation(pos, viewer, mouseHover, mouseOut, mouseMove);
-				} else {
-					boolean inFooter= !minings.isEmpty() ? (first instanceof DocumentFooterCodeMining) : false;
-					if (inFooter) {
-						ann= new CodeMiningDocumentFooterAnnotation(pos, viewer, mouseHover, mouseOut, mouseMove);
-					} else {
-						ann= new CodeMiningLineContentAnnotation(pos, viewer, afterPosition, mouseHover, mouseOut, mouseMove);
-					}
-				}
+
+				ann= switch (mode) {
+					case InLine -> new CodeMiningLineContentAnnotation(pos, viewer, afterPosition, mouseHover, mouseOut, mouseMove);
+					case HeaderLine -> new CodeMiningLineHeaderAnnotation(pos, viewer, mouseHover, mouseOut, mouseMove);
+					case FooterLine -> new CodeMiningDocumentFooterAnnotation(pos, viewer, mouseHover, mouseOut, mouseMove);
+
+					default -> throw new IllegalStateException("Found unexpected code mining display mode: " + mode); //$NON-NLS-1$
+				};
 			} else if (ann instanceof ICodeMiningAnnotation && ((ICodeMiningAnnotation) ann).isInVisibleLines()) {
 				// annotation is in visible lines
 				annotationsToRedraw.add((ICodeMiningAnnotation) ann);
--- a/examples/org.eclipse.jface.text.examples/src/org/eclipse/jface/text/examples/codemining/CodeMiningDemo.java
+++ b/examples/org.eclipse.jface.text.examples/src/org/eclipse/jface/text/examples/codemining/CodeMiningDemo.java
@@ -10,6 +10,7 @@
  *
  *  Contributors:
  *  Angelo Zerr <angelo.zerr@gmail.com> - [CodeMining] Add CodeMining support in SourceViewer - Bug 527515
+ *  Dietrich Travkin <dietrich.travkin@solunar.de> - Fix code mining redrawing - Issue 3405
  */
 package org.eclipse.jface.text.examples.codemining;
 
@@ -34,7 +35,9 @@ import org.eclipse.jface.text.source.ISourceViewer;
 import org.eclipse.jface.text.source.ISourceViewerExtension5;
 import org.eclipse.jface.text.source.SourceViewer;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swt.widgets.Text;
@@ -45,13 +48,21 @@ import org.eclipse.swt.widgets.Text;
 public class CodeMiningDemo {
 
 	private static boolean showWhitespaces = false;
+	private static AtomicReference<Boolean> useInLineCodeMinings = new AtomicReference<>(false);
+
+	private static String LINE_HEADER = "Line header";
+	private static String IN_LINE = "In-line";
 
 	public static void main(String[] args) throws Exception {
 
 		Display display = new Display();
 		Shell shell = new Shell(display);
-		shell.setLayout(new GridLayout());
+		shell.setLayout(new GridLayout(2, false));
 		shell.setText("Code Mining demo");
+
+		Button toggleInLineButton = new Button(shell, SWT.PUSH);
+		toggleInLineButton.setText(LINE_HEADER);
+		GridDataFactory.fillDefaults().align(SWT.BEGINNING, SWT.FILL).grab(false, false).applyTo(toggleInLineButton);
 
 		AtomicReference<String> endOfLineString = new AtomicReference<>("End of line");
 		Text endOfLineText = new Text(shell, SWT.NONE);
@@ -70,7 +81,9 @@ public class CodeMiningDemo {
 						+ "// Name class with a number N to emulate Nms before resolving the references CodeMining\n"
 						+ "// Empty lines show a header annotating they're empty.\n"
 						+ "// The word `echo` is echoed.\n"
-						+ "// Lines containing `end` get an annotation at their end\n\n"
+						+ "// Lines containing `end` get an annotation at their end\n"
+						+ "// Press the toggle button in the upper left  corner to switch between\n"
+						+ "// showing reference titles in-line and showing them in additional lines.\n\n"
 						+ "class A\n" //
 						+ "new A\n" //
 						+ "new A\n\n" //
@@ -79,29 +92,39 @@ public class CodeMiningDemo {
 						+ "class 5\n" //
 						+ "new 5\n" //
 						+ "new 5\n" //
-						+ "new 5\n" //
+						+ "new 5\n\n" //
+						+ "Text with some references like [REF-X]\n" + "and [REF-Y] in it.\n\n"
 						+ "multiline \n" //
 						+ "multiline \n\n" //
 						+ "suffix \n"),
 				new AnnotationModel());
-		GridDataFactory.fillDefaults().grab(true, true).applyTo(sourceViewer.getTextWidget());
+		GridDataFactory.fillDefaults().span(2, 1).grab(true, true).applyTo(sourceViewer.getTextWidget());
 		// Add AnnotationPainter (required by CodeMining)
 		addAnnotationPainter(sourceViewer);
+
+		toggleInLineButton.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
+			useInLineCodeMinings.set(!useInLineCodeMinings.get());
+			toggleInLineButton.setText(useInLineCodeMinings.get() ? IN_LINE : LINE_HEADER);
+			sourceViewer.updateCodeMinings();
+		}));
+
 		// Initialize codemining providers
-		((ISourceViewerExtension5) sourceViewer).setCodeMiningProviders(new ICodeMiningProvider[] {
+		sourceViewer.setCodeMiningProviders(new ICodeMiningProvider[] {
 				new ClassReferenceCodeMiningProvider(), //
 				new ClassImplementationsCodeMiningProvider(), //
 				new ToEchoWithHeaderAndInlineCodeMiningProvider("echo"), //
 				new MultilineCodeMiningProvider(), //
 				new EmptyLineCodeMiningProvider(), //
 				new EchoAtEndOfLineCodeMiningProvider(endOfLineString), //
-				new LineContentCodeMiningAfterPositionProvider() });
+				new LineContentCodeMiningAfterPositionProvider(), //
+				new ReferenceCodeMiningProvider(useInLineCodeMinings) });
+
 		// Execute codemining in a reconciler
 		MonoReconciler reconciler = new MonoReconciler(new IReconcilingStrategy() {
 
 			@Override
 			public void setDocument(IDocument document) {
-				((ISourceViewerExtension5) sourceViewer).updateCodeMinings();
+				sourceViewer.updateCodeMinings();
 			}
 
 			@Override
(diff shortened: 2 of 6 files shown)


bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
```diff
@@ -230,6 +230,20 @@ private static Map<Position, List<ICodeMining>> groupByLines(List<? extends ICod
 				Collectors.mapping(Function.identity(), Collectors.toList())));
 	}
 
+	private enum CodeMiningMode {
```
tobiasmelcher: please make enum "static" -> `private static enum CodeMiningMode`

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
```diff
@@ -257,11 +271,24 @@ private void renderCodeMinings(Map<Position, List<ICodeMining>> groups, ISourceV
 			Position pos= new Position(g.getKey().offset, g.getKey().length);
 			List<ICodeMining> minings= g.getValue();
 			ICodeMining first= minings.get(0);
-			boolean inLineHeader= !minings.isEmpty() ? (first instanceof LineHeaderCodeMining) : true;
+
+			CodeMiningMode mode= CodeMiningMode.InLine;
+			if (!minings.isEmpty()) {
+				if (CodeMiningMode.InLine.codeMiningType.isInstance(first)) {
```
tobiasmelcher: could this logic be moved to a static helper method to CodeMiningMode. Method name could be `createFor` or similar.

examples/org.eclipse.jface.text.examples/src/org/eclipse/jface/text/examples/codemining/CodeMiningDemo.java
```diff
@@ -45,14 +48,22 @@
 public class CodeMiningDemo {
 
 	private static boolean showWhitespaces = false;
+	private static AtomicReference<Boolean> useInLineCodeMinings = new AtomicReference<>(false);;
```
tobiasmelcher: remove the second semicolon

tobiasmelcher (commented):
Thanks a lot for the fixes, great work.
Could you please also create a unit test which documents the problem scenario where the CodeMiningType for a given offset is changed?

travkin79:
> Could you please also create a unit test which documents the problem scenario where the CodeMiningType for a given offset is changed?

Hi @tobiasmelcher,
Thank you for the feedback. I'll incorporate corresponding changes.

I was thinking about how to implement a test for my case, but for some reason, some tests, esp. CodeMining test, fail on my machine on the master branch (without my changes), commit hash 0ce143000f851b4b90f3f1d1103d17a0add37292. I didn't find the reason yet.

image

travkin79:
Hi @tobiasmelcher and @BeckerWdf,

I'm trying to implement a test for issue #3405, but I'm struggling with various issues with running the tests on my machine(s). I might need some help.

Some tests fail for the unmodified Eclipse platform UI master branch, commit 0ce143000f851b4b90f3f1d1103d17a0add37292, on MacOS and on Windows 11 as well. Due to the failing tests, I cannot assure that my new test will behave as expected.

In addition, the SWT widgets or the tests seem to behave strange. For example, when debugging my new test case on MacOS, the first line's vertical line indentation is reported to be 16, but the code mining is being drawn onto the source viewer's content instead of drawing it into an extra line (see screenshot 1). The in-line code mining is sometimes also drawn into the source viewer's text (see screeshot 2). The method `getStyleRangeAtOffset(...)` seems to always return `null`, on both operating systems (Mac & Win). Thus, I don't know which conditions to choose in my tests in order to check if the code minings are correctly drawn.

Screenshot 1:
image

Screenshot 2:
image


I've extended the code mining demo where the code minings seem to behave as expected, i.e. drawing a "title" either above a reference (e.g. "REF-X" in my example) in a line header or in the same line in front of the referencing text:

Line Header:
image

In-line:
image

It would be great if someone could take a look at my PR draft and help me fixing the tests.

tobiasmelcher:
@travkin79 sure, I'll take a look and make a proposal. Please give me some time. I will also check if I can reproduce the other test errors you have mentioned.

travkin79:
Thank you @tobiasmelcher,

I managed to finish a first version of a test. It turns out, `getStyleRangeAtOffset(...)` does return something `!= null`, but for an unexpected index. For a text "Here REF-X..." where a code mining label is placed after "Here ", I expected `getStyleRangeAtOffset(5)` to be non-null, but instead I get a non-null result for `getStyleRangeAtOffset(4)`, i.e. at the position of the space character. That looks wrong to me.

Besides that, there might be an issue with my PR in case that someone is using `AbstractCodeMining` class instead of sub-classing one of `LineContentCodeMining`, `LineHeaderCodeMining`, or `DocumentFooterCodeMining`. If none of the latter classes is sub-classed, we cannot determine which of the three corresponding code mining modes we have. I saw a few tests fail because of using anonymous sub-classes of `AbstractCodeMining` and fixed these, but there might be some client code doing something similar. I guess, moving `AbstractCodeMining` to an internal package is not an option.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
```diff
@@ -230,6 +231,40 @@ private static Map<Position, List<ICodeMining>> groupByLines(List<? extends ICod
 				Collectors.mapping(Function.identity(), Collectors.toList())));
 	}
 
+	private static enum CodeMiningMode {
```
iloveeclipse, line 234: Could you please apply this patch to this enum?

```
diff --git a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
index 3427cf7..d69aee6 100644
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
@@ -234,7 +234,8 @@
 	private static enum CodeMiningMode {
-		InLine(LineContentCodeMining.class, CodeMiningLineContentAnnotation.class), HeaderLine(LineHeaderCodeMining.class,
-				CodeMiningLineHeaderAnnotation.class), FooterLine(DocumentFooterCodeMining.class, CodeMiningDocumentFooterAnnotation.class);
+		InLine(LineContentCodeMining.class, CodeMiningLineContentAnnotation.class), 
+		HeaderLine(LineHeaderCodeMining.class, CodeMiningLineHeaderAnnotation.class), 
+		FooterLine(DocumentFooterCodeMining.class, CodeMiningDocumentFooterAnnotation.class);
 
-		public Class<? extends ICodeMining> codeMiningType;
-		public Class<? extends AbstractInlinedAnnotation> annotationType;
+		public final Class<? extends ICodeMining> codeMiningType;
+		public final Class<? extends AbstractInlinedAnnotation> annotationType;
 
```

iloveeclipse (commented):
@travkin79 : Please 
1) squash all the changes from 6 commits to one or two.
2) Provide meaningful commit message, ideally following https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations
3) featch latest master state and rebase your branch on it
4) Force push so PR is updated

tobiasmelcher:
> there might be an issue with my PR in case that someone is

The tests of CodeMiningTest also fail on windows on my machine. 

image

I assume the test errors appear if you have monitor scaling factory larger than 1.0.
I will take care of this issue.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
travkin79: Hi @iloveeclipse,
Sure. It seems, the Oomph installed Eclipse IDE (SDK) has some save actions that reformat the code this way. I was not sure if that's really desired behavior since the automated formatting made the code less readable. Maybe someone should check the save actions that are set by oomph.

By the way, applying a patch file worked without reformatting the code.

merks: The setup sets very few preferences:

image

So any save actions are either the default, which is none by default,

image

Or some project-specific setting, e.g., these:

image

travkin79:
> @travkin79 : Please
> 
> 1. squash all the changes from 6 commits to one or two.
> 2. Provide meaningful commit message, ideally following https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations
> 3. featch latest master state and rebase your branch on it
> 4. Force push so PR is updated

Done. Thank you for looking at my PR

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
iloveeclipse: > Sure. It seems, the Oomph installed Eclipse IDE (SDK) has some save actions that reformat the code this way. I was not sure if that's really desired behavior since the automated formatting made the code less readable. Maybe someone should check the save actions that are set by oomph.
> 
> By the way, applying a patch file worked without reformatting the code.

It's because of save actions configured to reformat code. In this case the code reformatting rules make the enum definition not readable.

travkin79:
Hi @tobiasmelcher,

> I assume the test errors appear if you have monitor scaling factory larger than 1.0.
I will take care of this issue.

I think, fixing these tests should be done in a separate issue / PR. I just wanted to be sure, that I'm not doing something wrong when running the existing tests.

In addition to the tests already mentioned, the following tests do also fail on my machine. I guess, it might be for the same reason.
image

tobiasmelcher:
> Thank you @tobiasmelcher,
> 
> I managed to finish a first version of a test. It turns out, `getStyleRangeAtOffset(...)` does return something `!= null`, but for an unexpected index. For a text "Here REF-X..." where a code mining label is placed after "Here ", I expected `getStyleRangeAtOffset(5)` to be non-null, but instead I get a non-null result for `getStyleRangeAtOffset(4)`, i.e. at the position of the space character. That looks wrong to me.
> 

@travkin79 : If you want to work with offset 5 inside the unit test implementation, you need to set `afterPosition` of `LineContentCodeMining` to `true`. Then there is no `- 1` necessary.

```
private static class ReferenceInLineCodeMining extends LineContentCodeMining {

		public ReferenceInLineCodeMining(String label, int positionOffset, ICodeMiningProvider provider) {
			super(new Position(positionOffset, 1), true, provider, null); // <!----- second argument needs to be set to true (default is false)
			this.setLabel(label);
		}

	}
```

tobiasmelcher:
> just wanted to be sure, that I'm not doing something wrong when running the existing tests.
> 
> In addition to the tests already mentione

Sure, the tests will be fixed via a separate pull request.

tobiasmelcher:
> Besides that, there might be an issue with my PR in case that someone is using `AbstractCodeMining` class instead of sub-classing one of `LineContentCodeMining`, `LineHeaderCodeMining`, or `DocumentFooterCodeMining`. If none of the latter classes is sub-classed, we cannot determine which of the three corresponding code mining modes we have. I saw a few tests fail because of using anonymous sub-classes of `AbstractCodeMining` and fixed these, but there might be some client code doing something similar. I guess, moving `AbstractCodeMining` to an internal package is not an option.

@travkin79 : This is an important topic. We need to support the case that implementations return only an instance of `AbstractCodeMining`. The old implementation then used `CodeMiningLineContentAnnotation`. Instead of raising an `IllegalStateException` inside `CodeMiningMode#createFor` we need to set the mode to `InLine`. Could you please update the code? Thanks a lot.

travkin79:
> > Besides that, there might be an issue with my PR in case that someone is using `AbstractCodeMining` class instead of sub-classing one of `LineContentCodeMining`, `LineHeaderCodeMining`, or `DocumentFooterCodeMining`. If none of the latter classes is sub-classed, we cannot determine which of the three corresponding code mining modes we have. I saw a few tests fail because of using anonymous sub-classes of `AbstractCodeMining` and fixed these, but there might be some client code doing something similar. I guess, moving `AbstractCodeMining` to an internal package is not an option.
> 
> @travkin79 : This is an important topic. We need to support the case that implementations return only an instance of `AbstractCodeMining`. The old implementation then used `CodeMiningLineContentAnnotation`. Instead of raising an `IllegalStateException` inside `CodeMiningMode#createFor` we need to set the mode to `InLine`. Could you please update the code? Thanks a lot.

Hi @tobiasmelcher,
I did exactly that before, but faced another problem. I tested my implementation (without throwing an exception) with the new Code Mining Demo example and with the existing code mining tests. In some cases using `CodeMiningLineContentAnnotation` as a default lead to the source viewer being re-painted over and over again and the scroll bar was flickering (getting longer and then shorter over and over again). Thus, I think, we need to check that case in more detail.

I could create a separate git branch, where you could reproduce and debug that case on your machine if you like.

tobiasmelcher:
> > > Besides that, there might be an issue with my PR in case that someone is using `AbstractCodeMining` class instead of sub-classing one of `LineContentCodeMining`, `LineHeaderCodeMining`, or `DocumentFooterCodeMining`. If none of the latter classes is sub-classed, we cannot determine which of the three corresponding code mining modes we have. I saw a few tests fail because of using anonymous sub-classes of `AbstractCodeMining` and fixed these, but there might be some client code doing something similar. I guess, moving `AbstractCodeMining` to an internal package is not an option.
> > 
> > 
> > @travkin79 : This is an important topic. We need to support the case that implementations return only an instance of `AbstractCodeMining`. The old implementation then used `CodeMiningLineContentAnnotation`. Instead of raising an `IllegalStateException` inside `CodeMiningMode#createFor` we need to set the mode to `InLine`. Could you please update the code? Thanks a lot.
> 
> Hi @tobiasmelcher, I did exactly that before, but faced another problem. I tested my implementation (without throwing an exception) with the new Code Mining Demo example and with the existing code mining tests. In some cases using `CodeMiningLineContentAnnotation` as a default lead to the source viewer being re-painted over and over again and the scroll bar was flickering (getting longer and then shorter over and over again). Thus, I think, we need to check that case in more detail.
> 
> I could create a separate git branch, where you could reproduce and debug that case on your machine if you like.

That would be very helpful, thanks. Could you please also tell me then the steps how to reproduce the problem scenario?

travkin79:
> > Thank you @tobiasmelcher,
> > I managed to finish a first version of a test. It turns out, `getStyleRangeAtOffset(...)` does return something `!= null`, but for an unexpected index. For a text "Here REF-X..." where a code mining label is placed after "Here ", I expected `getStyleRangeAtOffset(5)` to be non-null, but instead I get a non-null result for `getStyleRangeAtOffset(4)`, i.e. at the position of the space character. That looks wrong to me.
> 
> @travkin79 : If you want to work with offset 5 inside the unit test implementation, you need to set `afterPosition` of `LineContentCodeMining` to `true`. Then there is no `- 1` necessary.
> 
> ```
> private static class ReferenceInLineCodeMining extends LineContentCodeMining {
> 
> 		public ReferenceInLineCodeMining(String label, int positionOffset, ICodeMiningProvider provider) {
> 			super(new Position(positionOffset, 1), true, provider, null); // <!----- second argument needs to be set to true (default is false)
> 			this.setLabel(label);
> 		}
> 
> 	}
> ```

Hi @tobiasmelcher,
Thank you for the hint. I adapted my test and code mining demo code (for now in a separate commit that can be squashed later).

But checking the implementation in `CodeMiningManager` I looks to me as if `afterPosition` should always be `true` for `CodeMiningLineContentAnnotation` and `LineContentCodeMining`. If I am write, then the constructors should be adapted accordingly.

tobiasmelcher:
> > > Thank you @tobiasmelcher,
> > > I managed to finish a first version of a test. It turns out, `getStyleRangeAtOffset(...)` does return something `!= null`, but for an unexpected index. For a text "Here REF-X..." where a code mining label is placed after "Here ", I expected `getStyleRangeAtOffset(5)` to be non-null, but instead I get a non-null result for `getStyleRangeAtOffset(4)`, i.e. at the position of the space character. That looks wrong to me.
> > 
> > 
> > @travkin79 : If you want to work with offset 5 inside the unit test implementation, you need to set `afterPosition` of `LineContentCodeMining` to `true`. Then there is no `- 1` necessary.
> > ```
> > private static class ReferenceInLineCodeMining extends LineContentCodeMining {
> > 
> > 		public ReferenceInLineCodeMining(String label, int positionOffset, ICodeMiningProvider provider) {
> > 			super(new Position(positionOffset, 1), true, provider, null); // <!----- second argument needs to be set to true (default is false)
> > 			this.setLabel(label);
> > 		}
> > 
> > 	}
> > ```
> 
> Hi @tobiasmelcher, Thank you for the hint. I adapted my test and code mining demo code (for now in a separate commit that can be squashed later).
> 
> But checking the implementation in `CodeMiningManager` I looks to me as if `afterPosition` should always be `true` for `CodeMiningLineContentAnnotation` and `LineContentCodeMining`. If I am write, then the constructors should be adapted accordingly.

That's a compatibility issue. We cannot change the default. Code minings were initially implemented to support the parameter hints where the mining should be drawn before a given parameter.
image

I introduced at a later point in time the `afterPosition` argument in the Code Mining API which had to be introduced in a compatible way. The `afterPosition` mechanism is used to realize AI based code completions (the Copilot plugin is one user of this argument).

travkin79:
Thank you for the explanation @tobiasmelcher,
I created a git branch for testing purposes, where I no longer throw an exception if the `CodeMiningManager` doesn't find one of the three expected sub-types of `AbstractCodeMining`. I also set the default code mining mode back to line header, since that was originally the case and this way we can reproduce the permanent redrawing issue. In addition, I reverted replacing `AbstractCodeMining` with `LineContentCodeMining` in some tests and examples.

Surprisingly, setting the default to `CodeMiningMode.InLine` in `CodeMiningMode#createFor` does not produce the flickering scroll bar / redrawing code minings issue. But I'm not sure if the issue only disappears for our code mining demo example. It could still occur in other cases.

git branch: https://github.com/travkin79/eclipse.platform.ui/tree/test-code-minings

**Steps to reproduce**
Run the `org.eclipse.jface.text.examples.codemining.CodeMiningDemo` as a Java application (Run as...)

**Result**
You'll see a window with some code where the scroll bar is flickering, becoming bigger and smaller again, over and over again. (see attached 5 sec video)

https://github.com/user-attachments/assets/29526322-54a1-4a03-9501-a2ba476084ff

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningManager.java
travkin79: Yes, the project-specific settings for save actions are the cause. Though, I think, not re-formatting edited lines is the real problem, but some project-specific formatter settings. I'm pretty sure, there is some formatter settings that would keep user-added line breaks (e.g. in enums, switch expressions / statements, streams,...). I think, the formatter settings should be changed in this case.

image

tobiasmelcher:
> `CodeMiningDemo`

Hi @travkin79 ,
thanks a lot for providing all the details, that helped a lot. I was able to reproduce the flickering scenario.
I compared the old decision logic what kind of annotation should be created `boolean inLineHeader= !minings.isEmpty() ? (first instanceof LineHeaderCodeMining) : true;` with the new one `CodeMiningMode#createFor` and found a small difference. If `minings` list is not empty and first entry is not a `LineHeaderCodeMining` then a `CodeMiningLineContentAnnotation` was created in the old logic. This is not the case for the new implementation.

I changed `CodeMiningMode#createFor` the following way and flickering no longer occurs.
```
public static CodeMiningMode createFor(List<ICodeMining> minings) {
	Assert.isNotNull(minings);

	CodeMiningMode mode= CodeMiningMode.HeaderLine;
	if (!minings.isEmpty()) {
		ICodeMining first= minings.get(0);

		if (CodeMiningMode.InLine.codeMiningType.isInstance(first)) {
			mode= CodeMiningMode.InLine;
		} else if (CodeMiningMode.HeaderLine.codeMiningType.isInstance(first)) {
			mode= CodeMiningMode.HeaderLine;
		} else if (CodeMiningMode.FooterLine.codeMiningType.isInstance(first)) {
			mode= CodeMiningMode.FooterLine;
		} else {
			mode= CodeMiningMode.InLine; // <!------------ add this else branch
		}
	}
	return mode;
}
```
Can we go with this approach? I did not further analyze the flickering issue and cannot explain what happens. I think we are safe if we can ensure that the new implementation creates the same annotations as before.

With best regards,
Tobias

travkin79:
Hi @tobiasmelcher,
Now I see the case that I didn't consider. Thank you very much for analyzing. Yes, I agree with that solution. If there are no objections, we can go with this solution. Can you add your fix while merging?

Best regards,
Dietrich

tobiasmelcher:
> Hi @tobiasmelcher, Now I see the case that I didn't consider. Thank you very much for analyzing. Yes, I agree with that solution. If there are no objections, we can go with this solution. Can you add your fix while merging?
> 
> Best regards, Dietrich

I hope I updated the pull request the correct way? If everything is fine, we need to squash the changes. I can do it if you want.

travkin79:
Hi @tobiasmelcher,
It would be great if you could squash the commits and rebase on master.

tobiasmelcher:
Test error in LinkHelperTest.testLinkHelperEditorActivation not related to this change

BeckerWdf:
So is this good to be merged now?

tobiasmelcher:
> So is this good to be merged now?

Yes, from my point of view this pull request is ready to be merged.

BeckerWdf:
@travkin79: Thanks for providing a PR.

travkin79:
Thank you all for the reviews, analysis, and merging. 👍

PR #3074, opened 2025-06-29
Allow using visible regions with projections

danthe1st:
This is my attempt to fix #3073 in order to implement https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2264

### Description

`ProjectionViewer` currently cannot use visible regions and projections together (except by calling `enableProjections()` after `setVisibleRegions` in which case it will show a wrong region). This PR changes the implementation of visible regions to collapse everything outside the visible region.

### Note

This PR does not allow collapsing or expanding any projection/folding regions outside of the declared _visible region_ (e.g. by other code modifying the document in some way or if a projection region is partially in the visible region).

Aside from that, this implementation has the following effect: If `setVisibleRegion` is called with a region that is fully collapsed, the editor is basically empty. If another behavior is desired, it would be good to know what the better behavior is.

### testing with JDT

To test this with JDT as requested in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2264, do the following:
- Create a workspace with eclipse.platform.ui and JDT-UI set up (I did so by creating a Target Platform and added whatever I thought was necessary until I no longer got any errors but I have no idea how to properly do that)
- Use the changes in eclipse.platform.ui
- Apply https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2302 with JDT-UI:
- Create an "Eclipse Application" run configuration with this version of Eclipse Platform and the patched version of JDT-UI and start it
- Enable Window > Preferences > Java > Editor > "Only show the selected Java Element"
- Make sure folding is enabled at Window > Preferences > Java > Editor > Folding
- Open a Java class
- Select elements in the outline and observe that only the "selected element" is shown in the editor and folding still works (whether it's normal folding, custom folding regions, "Extended Folding", etc shouldn't make a difference)
image
- If "Enhanced Folding" is active, make sure that https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2410 is available as well

PR #3423, opened 2025-10-21
hide projection annotations outside of the current visible region

danthe1st:
In #3074, I prevented collapsing and expanding projection regions that overlap with parts of the file outside of the current visible region.

However, this does not prevent the projection annotation from being shown in the editor. This PR changes that behavior and ensures these annotations (that cannot be collapsed or expanded) cannot be shown.

To prevent showing empty "files" in the editor which cannot be expanded, I also changed the code for setting the visible region to expand any surrounding or "bordering" projection regions.

### manual Testing

This can be tested by setting up an editor with content similar to the following:
```java

visible_region_start

projection_start

visible_region_end

projection_end

```

If the visible region is configured from `visible_region_start` to `visible_region_end` and a projection annotation is created between `projection_start` and `projection_end` (in both cases, these texts are inclusive), there is an annotation at `projection_start` which cannot be collapsed.

If I figure out how to test whether an annotation is drawn, I can provide a proper test case.

Here is a test case for manual inspection:
```java
	@Test
	public void testProjectionRegionsShownOnlyInVisibleRegion() throws BadLocationException, InterruptedException {
		Shell shell= new Shell(Display.getCurrent());
		shell.setLayout(new FillLayout());
		TestProjectionViewer viewer= new TestProjectionViewer(shell, null, null, true, SWT.ALL);
                // This should hide the folding annotation
		String documentContent= """

				visible_region_start

				projection_start

				visible_region_end

				projection_end

				""";

                // This should show it
//		String documentContent= """
//
//				visible_region_start
//
//				projection_start
//
//				projection_end
//
//				visible_region_end
//
//
//				""";
		Document document= new Document(documentContent);
		viewer.setDocument(document, new AnnotationModel());
		int visibleRegionStart = documentContent.indexOf("visible_region_start");
		int visibleRegionEnd = documentContent.indexOf("\n", documentContent.indexOf("visible_region_end"))+1;

		int projectionStart= documentContent.indexOf("projection_start");
		int projectionEnd= documentContent.indexOf("\n", documentContent.indexOf("projection_end")) + 1;

		viewer.setVisibleRegion(visibleRegionStart, visibleRegionEnd - visibleRegionStart);
		viewer.enableProjection();
		ProjectionAnnotation annotation= new ProjectionAnnotation();
		viewer.getProjectionAnnotationModel().addAnnotation(annotation, new Position(projectionStart, projectionEnd - projectionStart));
		shell.setVisible(true);
		Canvas canvas= new Canvas(shell, SWT.ALL);
		try {
			canvas.addPaintListener(new PaintListener() {

				@Override
				public void paintControl(PaintEvent e) {
					annotation.paint(e.gc, canvas, new Rectangle(e.x, e.y, e.width, e.height));
				}
			});

			Display display= shell.getDisplay();
			while (!shell.isDisposed()) {
				if (!display.readAndDispatch()) {
					display.sleep();
				}
			}
		} finally {
			canvas.dispose();
			shell.dispose();
		}
	}
```

This is what it should look like:
image

With the other version (or without this change), there is a projection annotation shown on the right:
image

I also added automated tests for it.

### JDT example (for context)

More or less https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2302#issuecomment-3423803023, this can be seen with JDT using https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2302 by enabling Window > Preferences > Java > Editor > Only show the selected Java Element as well as Window > Preferences > Java > Editor > Folding > Enable folding of custom regions

Open a Java file with the following code (adapted from https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2302#issuecomment-3423803023):
```java
public class SomeClass {
    // region

    void a() {

    }

    void b() {

    }

    // endregion
}
```

When selecting `a()` in the editor, it only shows
```java
    // region

    void a() {

    }
```
Without this PR, the `// region` has a folding annotation on the right which cannot be used. With this PR, this region is used. However, it is still present when the whole class is shown.

### DIFF
--- a/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionAnnotation.java
+++ b/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionAnnotation.java
@@ -73,6 +73,8 @@ public class ProjectionAnnotation extends Annotation implements IAnnotationPrese
 	/** Indicates whether this annotation should be painted as range */
 	private boolean fIsRangeIndication= false;
 
+	private boolean hidden= false;
+
 	/**
 	 * Creates a new expanded projection annotation.
 	 */
@@ -115,6 +117,9 @@ public class ProjectionAnnotation extends Annotation implements IAnnotationPrese
 
 	@Override
 	public void paint(GC gc, Canvas canvas, Rectangle rectangle) {
+		if (hidden) {
+			return;
+		}
 		Image image= getImage(canvas.getDisplay());
 		if (image != null) {
 			ImageUtilities.drawImage(image, gc, canvas, rectangle, SWT.CENTER, SWT.TOP);
@@ -126,6 +131,10 @@ public class ProjectionAnnotation extends Annotation implements IAnnotationPrese
 				drawRangeIndication(gc, canvas, rectangle);
 			}
 		}
+	}
+
+	void setHidden(boolean hidden) {
+		this.hidden= hidden;
 	}
 
 	@Override
--- a/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
+++ b/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
@@ -124,6 +124,7 @@ public class ProjectionViewer extends SourceViewer implements ITextViewerExtensi
 					fProjectionSummary.updateSummaries();
 				}
 				processCatchupRequest(event);
+				correctChangedAnnotationVisibility(event);
 
 			} else if (model == getAnnotationModel() && fProjectionSummary != null) {
 				fProjectionSummary.updateSummaries();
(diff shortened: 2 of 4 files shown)


iloveeclipse:
In general works as advertised, thanks. If you could figure out how to add test for that, it would be great.

jjohnstn:
@danthe1st There is a problem.  If one collapses a region around an element (e.g. method) and then selects the method, it shows the collapsed region comment but no annotation.  If one then selects the class, the region comment has a plus, though it is already expanded and the weird text at end of folded lines is all the place.  Clicking on the plus resets everything back to where it is supposed to be.

```
public class TestFolding {

	// region

	void test() {

	}
	void test2() {

	}

	void test2a() {

	}
	void test3() {}

	// endregion

	// region2

	void test4() {

	}

	// endregion2
}

```

danthe1st:
> If one collapses a region around an element (e.g. method) and then selects the method, it shows the collapsed region comment but no annotation.

I see. It might be a possibility to automatically expand regions like this in such a situation (I guess selecting a visible region that is practically collapsed doesn't make much sense).

danthe1st:
@jjohnstn I attempted to provide a fix for it by automatically expanding regions that border the current folding region. However, I didn't get to test it properly yet (I plan to do that tomorrow).

I think this should be the best behavior for anything running into that behavior (not just JDT).

danthe1st:
I improved and tested my fix to the problem mentioned by @jjohnstn and I adjusted this PR (both the code/tests and the PR description).

While testing, I think I found another issue but I think that's unrelated to this PR and should probably be fixed on its own.

iloveeclipse:
Thanks.

iloveeclipse:
This PR caused regression, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/3459.

PR #3429, opened 2025-10-22
CountDownLatch Synchronization for RCPTestWorkbenchAdvisor

### LINKED ISSUE #1517: Random failing PlatformUITest#testCreateAndRunWorkbench()

vogella:
The flaky test was caused by a race condition in
RCPTestWorkbenchAdvisor.java. The test spawns 4 background threads during preStartup() that schedule async/sync runnables on the display.

However, there was no guarantee these runnables would execute before postStartup() marked the workbench as "started" (by setting started = true).

This meant asyncWithDisplayAccess could still be null when the test assertions ran, causing intermittent failures.

This change adds a CountDownLatch to ensure all 4 async/sync operations complete before marking the workbench as started:

  Key changes to
/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java:

  1. Added latch field (line 48):
    - private static java.util.concurrent.CountDownLatch asyncLatch = null;
  2. Initialize latch in preStartup() (line 130):
    - asyncLatch = new java.util.concurrent.CountDownLatch(4); — tracks 4 operations
  3. Count down after each operation completes:
    - In setupSyncDisplayThread(): Added finally block (lines 179-183) that calls asyncLatch.countDown()
    - In setupAsyncDisplayThread(): Added finally block (lines 205-209) that calls asyncLatch.countDown()
  4. Wait for operations in postStartup() (lines 222-235):
    - Waits up to 5 seconds for all operations to complete
    - Logs warnings if timeout occurs or thread is interrupted
    - Only marks started = true after all operations complete

Fixes #1517

### DIFF
--- a/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
+++ b/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
@@ -13,6 +13,10 @@
  *******************************************************************************/
 package org.eclipse.ui.tests.harness.util;
 
+import java.util.concurrent.Phaser;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.swt.SWTException;
 import org.eclipse.swt.widgets.Display;
@@ -43,6 +47,10 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 	public static volatile Boolean asyncWithoutDisplayAccess = null;
 
 	private static boolean started = false;
+
+	// Use a Phaser to ensure async operations are scheduled before postStartup
+	// Each thread registers itself and arrives when the operation is scheduled
+	private static final Phaser asyncPhaser = new Phaser(1); // 1 for the main thread
 
 	public static boolean isSTARTED() {
 		synchronized (RCPTestWorkbenchAdvisor.class) {
@@ -122,12 +130,10 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 	public void preStartup() {
 		super.preStartup();
 		final Display display = Display.getCurrent();
+
 		if (display != null) {
 			display.asyncExec(() -> {
-				if (isSTARTED())
-					asyncDuringStartup = Boolean.FALSE;
-				else
-					asyncDuringStartup = Boolean.TRUE;
+				asyncDuringStartup = !isSTARTED();
 			});
 		}
 
@@ -151,6 +157,8 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 	}
 
 	private void setupSyncDisplayThread(final boolean callDisplayAccess, final Display display) {
+		// Register this thread with the phaser
+		asyncPhaser.register();
 		Thread syncThread = new Thread() {
 			@Override
 			public void run() {
@@ -158,17 +166,19 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 					DisplayAccess.accessDisplayDuringStartup();
 				try {
 					display.syncExec(() -> {
-						synchronized (RCPTestWorkbenchAdvisor.class) {
-							if (callDisplayAccess)
-								syncWithDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
-							else
-								syncWithoutDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
+						if (callDisplayAccess) {
+							syncWithDisplayAccess = !isSTARTED();
+						} else {
+							syncWithoutDisplayAccess = !isSTARTED();
 						}
 					});
 				} catch (SWTException e) {
 					// this can happen because we shut down the workbench just
 					// as soon as we're initialized - ie: when we're trying to
 					// run this runnable in the deferred case.
+				} finally {
+					// Signal that this operation has completed
+					asyncPhaser.arriveAndDeregister();
 				}
 			}
 		};
@@ -177,19 +187,26 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 	}
 
 	private void setupAsyncDisplayThread(final boolean callDisplayAccess, final Display display) {
+		// Register this thread with the phaser
+		asyncPhaser.register();
 		Thread asyncThread = new Thread() {
 			@Override
 			public void run() {
 				if (callDisplayAccess)
 					DisplayAccess.accessDisplayDuringStartup();
-				display.asyncExec(() -> {
-					synchronized (RCPTestWorkbenchAdvisor.class) {
-						if (callDisplayAccess)
-							asyncWithDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
-						else
-							asyncWithoutDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
-					}
-				});
+				try {
+					display.asyncExec(() -> {
+						if (callDisplayAccess) {
+							asyncWithDisplayAccess = !isSTARTED();
+						} else {
+							asyncWithoutDisplayAccess = !isSTARTED();
+						}
+					});
+				} finally {
+					// Signal that this operation has been scheduled (not necessarily executed yet)
+					// This avoids deadlock since we're not waiting for execution on the UI thread
+					asyncPhaser.arriveAndDeregister();
+				}
 			}
 		};
 		asyncThread.setDaemon(true);
@@ -199,6 +216,29 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 	@Override
 	public void postStartup() {
 		super.postStartup();
+
+		// Wait for all async/sync operations to be scheduled (not blocking UI thread)
+		try {
+			// Wait up to 5 seconds for all operations to be scheduled
+			// The main thread arrives and deregisters, waiting for all other registered threads
+			asyncPhaser.awaitAdvanceInterruptibly(asyncPhaser.arrive(), 5, TimeUnit.SECONDS);
+		} catch (TimeoutException e) {
+			// Log warning but don't throw - we need to mark as started to avoid breaking subsequent tests
+			System.err.println("WARNING: Not all async/sync operations were scheduled within timeout");
+			e.printStackTrace();
+		} catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			System.err.println("WARNING: Interrupted while waiting for async/sync operations");
+		}
+
+		// Pump the event loop to ensure async runnables execute before marking as started
+		// This prevents the original race condition where async variables might not be set yet
+		// Wait until the variables that should be set during startup are actually set to TRUE
+		UITestUtil.processEventsUntil(() -> Boolean.TRUE.equals(syncWithDisplayAccess) && Boolean.TRUE.equals(asyncWithDisplayAccess), 5000);
+		// Process any remaining events to allow variables that should NOT be set during startup
+		// to accidentally execute (to detect regression)
+		UITestUtil.processEvents();
+
 		synchronized (RCPTestWorkbenchAdvisor.class) {
 			started = true;
 		}
--- a/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
+++ b/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
@@ -13,9 +13,8 @@
  *******************************************************************************/
 package org.eclipse.ui.tests.harness.util;
 
-import java.util.concurrent.Phaser;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.swt.SWTException;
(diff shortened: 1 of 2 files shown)


tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -122,6 +125,10 @@ public void eventLoopIdle(final Display display) {
 	public void preStartup() {
 		super.preStartup();
 		final Display display = Display.getCurrent();
+
+		// Initialize a latch to track completion of the 4 async/sync operations
+		asyncLatch = new java.util.concurrent.CountDownLatch(4);
```
fedejeanne: Use an import instead of the fully qualified name.

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -199,6 +218,22 @@ public void run() {
 	@Override
 	public void postStartup() {
 		super.postStartup();
+
+		// Wait for all async/sync operations to complete before marking as started
+		// This prevents a race condition where asyncWithDisplayAccess might not be set yet
+		if (asyncLatch != null) {
```
fedejeanne: Can this condition ever be false? I don't see how.

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -183,11 +195,18 @@ public void run() {
...
+					try {
+						synchronized (RCPTestWorkbenchAdvisor.class) {
+							if (callDisplayAccess)
+								asyncWithDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
+							else
+								asyncWithoutDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
+						}
+					} finally {
+						// Signal that this operation has completed
+						if (asyncLatch != null) {
```
fedejeanne: Can this condition ever be false? I don't see how.

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -169,6 +176,11 @@ public void run() {
 					// this can happen because we shut down the workbench just
 					// as soon as we're initialized - ie: when we're trying to
 					// run this runnable in the deferred case.
+				} finally {
+					// Signal that this operation has completed
+					if (asyncLatch != null) {
```
fedejeanne: Can this condition ever be false? I don't see how.

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -44,6 +44,9 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 
 	private static boolean started = false;
 
+	// Use a CountDownLatch to ensure async operations complete before postStartup
+	private static java.util.concurrent.CountDownLatch asyncLatch = null;
```
fedejeanne: Use an import instead of the fully qualified name.

Additionally, if the latch is not optional (and I don't think it is) then you could create a constant and initialize it right here. That way you wouldn't have to check for null anywhere.

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -183,11 +195,18 @@ public void run() {
...
-						if (callDisplayAccess)
-							asyncWithDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
-						else
-							asyncWithoutDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
+					try {
+						synchronized (RCPTestWorkbenchAdvisor.class) {
+							if (callDisplayAccess)
+								asyncWithDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
+							else
+								asyncWithoutDisplayAccess = !isSTARTED() ? Boolean.TRUE : Boolean.FALSE;
```
fedejeanne:
```suggestion
							if (callDisplayAccess)
								asyncWithDisplayAccess = !isSTARTED();
							else
								asyncWithoutDisplayAccess = !isSTARTED();
```

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -199,6 +218,22 @@ public void run() {
...
+		if (asyncLatch != null) {
+			try {
+				// Wait up to 5 seconds for all operations to complete
+				boolean completed = asyncLatch.await(5, java.util.concurrent.TimeUnit.SECONDS);
+				if (!completed) {
+					System.err.println("Warning: Not all async/sync operations completed within timeout");
+				}
+			} catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				System.err.println("Warning: Interrupted while waiting for async/sync operations");
```
fedejeanne: It's probably safer to simply re-throw the exception and let it fail too. Otherwise the test may fail afterwards and one will have to look into the logs to see why.
```suggestion
throw e;
```

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -199,6 +218,22 @@ public void run() {
...
+
+		// Wait for all async/sync operations to complete before marking as started
+		// This prevents a race condition where asyncWithDisplayAccess might not be set yet
+		if (asyncLatch != null) {
+			try {
+				// Wait up to 5 seconds for all operations to complete
+				boolean completed = asyncLatch.await(5, java.util.concurrent.TimeUnit.SECONDS);
+				if (!completed) {
+					System.err.println("Warning: Not all async/sync operations completed within timeout");
+				}
```
fedejeanne, line 228: I would let it fail. I'd rather have a test that fails consistently than one that _might_ fail even if `completed == false`.

```suggestion
				assertTrue(completed, "Not all async/sync operations completed within timeout");
```

fedejeanne: Hm, it seems I missed an important part: when this fails, the workbench is already created and therefore subsequent attempts to create it will fail too.

After the latest changes, the logs show that once this happens, a bunch of tests fail with the error `Workbench already exists and cannot be created again`

Here are the first 3 failures.

<details><summary>Stack traces</summary>
<p>

```java
org.eclipse.ui.tests.rcp.PlatformUITest.testCreateAndRunWorkbench -- Time elapsed: 7.070 s <<< FAILURE!
java.lang.AssertionError: Not all async/sync operations completed within timeout
	at org.eclipse.ui.tests.harness.util.RCPTestWorkbenchAdvisor.postStartup(RCPTestWorkbenchAdvisor.java:221)
	at org.eclipse.ui.tests.rcp.util.WorkbenchAdvisorObserver.postStartup(WorkbenchAdvisorObserver.java:140)
	at org.eclipse.ui.internal.Workbench.lambda$18(Workbench.java:2901)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1104)
	at org.eclipse.ui.tests.rcp.PlatformUITest.testCreateAndRunWorkbench(PlatformUITest.java:57)

org.eclipse.ui.tests.rcp.PlatformUITest.testCreateAndRunWorkbenchWithExceptionOnStartup skipped
Running RcpTestSuite WorkbenchAdvisorTest
Tests run: 9, Failures: 0, Errors: 9, Skipped: 0, Time elapsed: 0.428 s <<< FAILURE! -- in RcpTestSuite WorkbenchAdvisorTest
org.eclipse.ui.tests.rcp.WorkbenchAdvisorTest.testEmptyProgressRegion -- Time elapsed: 0.191 s <<< ERROR!
java.lang.IllegalStateException: Workbench already exists and cannot be created again.
	at org.eclipse.ui.internal.Workbench.<init>(Workbench.java:481)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:609)
	at org.eclipse.ui.tests.rcp.WorkbenchAdvisorTest.testEmptyProgressRegion(WorkbenchAdvisorTest.java:293)

org.eclipse.ui.tests.rcp.WorkbenchAdvisorTest.testFillAllActionBar -- Time elapsed: 0.031 s <<< ERROR!
java.lang.IllegalStateException: Workbench already exists and cannot be created again.
	at org.eclipse.ui.internal.Workbench.<init>(Workbench.java:481)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:609)
	at org.eclipse.ui.tests.rcp.WorkbenchAdvisorTest.testFillAllActionBar(WorkbenchAdvisorTest.java:279)
```

</p>
</details> 

The 1st failure is due to the timeout (`Not all async/sync operations completed within timeout`) and the workbench is not cleaned-up/shutdown. The following failing tests also fail with `Workbench already exists and cannot be created again.`.

The way I see it, we need 2 things:
1. See why the timeout occurs (is it "just" a matter of waiting a bit longer since the hardware is old and slow or is it something else?)
2. Clean-up if the timeout occurs so that the following tests may start the workbench again.

I don't have any novel ideas at the moment :-\ and you @vogella ?

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -199,6 +212,38 @@ public void run() {
...
+				// Small yield to allow any final async operations to complete
+				try {
+					Thread.sleep(10);
+				} catch (InterruptedException e) {
+					Thread.currentThread().interrupt();
+					break;
+				}
+			}
+		}
```
HeikoKlare: Is this quite complex event processing really necessary? All relevant runnables are guaranteed to be scheduled here, so `display.readAndDispatch()` should not yield `false` until they all have been processed. You just need to make sure that those runnables that are _not_ supposed to be executed during startup had the chance to accidentally be executed (to detect regression). So wouldn't be something like this be sufficient?
```suggestion
		UITestUtil.processEventsUntil(() -> syncWithDisplayAccess && asyncWithDisplayAccess, 5000);
		UITestUtil.processEvents();
```

tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/RCPTestWorkbenchAdvisor.java
```diff
@@ -44,6 +47,9 @@ public class RCPTestWorkbenchAdvisor extends WorkbenchAdvisor {
 
 	private static boolean started = false;
 
+	// Use a CountDownLatch to ensure async operations complete before postStartup
+	private static final CountDownLatch asyncLatch = new CountDownLatch(4);
```
HeikoKlare: nit: wouldn't it be cleaner to use something like a `Phaser` and make every thread register itself and then deregister/arrive when being executed (like now done with `asyncLatch.countDown()`), so that the latch's count does not need to magically match the number of thread created?

HeikoKlare (commented):
The change looks sound. Thank you for taking care of that long-standing failing test. I just have minor comments regarding a potential improvement/simplification.

vogella:
## Updated with CountDownLatch fix

Replaced the failing Phaser approach with a simpler and more reliable CountDownLatch solution.

### What was wrong with Phaser?
The Phaser approach counted arrivals when operations were **scheduled**, not when they **executed**. This meant the test could mark `started=true` before the async runnables actually ran on the UI thread, causing test failures.

### What's the fix?
The CountDownLatch counts down **inside** the runnables after they execute, ensuring proper synchronization:
- Wait for 2 operations to complete (asyncWithDisplayAccess and syncWithDisplayAccess)
- Count down happens after variable assignment in the runnable
- `postStartup()` blocks until latch reaches zero
- Then processes remaining events before marking started=true

### Test Results
✅ Test passes consistently - verified with 5 consecutive local runs:
```
Run 1: Tests run: 4, Failures: 0, Errors: 0, Skipped: 1
Run 2: Tests run: 4, Failures: 0, Errors: 0, Skipped: 1
Run 3: Tests run: 4, Failures: 0, Errors: 0, Skipped: 1
Run 4: Tests run: 4, Failures: 0, Errors: 0, Skipped: 1
Run 5: Tests run: 4, Failures: 0, Errors: 0, Skipped: 1
```

The fix is now ready for CI verification.

vogella:
@HeikoKlare any additional feedback?

HeikoKlare:
I have to admit that I lost track of the changes and currently don't have further feedback. At least I do not see anything suspicious in the changes. I would propose to just give it a try.

PR #3431, opened 2025-10-22
Detect the desired size from the path or Query parameters.

laeubi:
Currently we have the problem that I want to replace some icons with a SVG variant. Currently Icons can be organized in folders like `/icons/16x16/search.png`, `/icons/32x32/search.png` and so on for high resolution support. Currently if one want to replace such structure with SVG it is required to scale down the SVG to 16x16 pixel document size as otherwise they get rendered at there native size (what usually is much larger).

As it is not really desirable to restrict the size of the SVG design for technical reasons, JFace now can detect two cases:

1) the SVG is places in a folder with "classic" folder layout the size is extracted and passed down as a hint for dynamic sizable icons 2) one can additionally add a query parameter, e.g. if I have an icon like `/icons/obj16/search.svg` and I have two places where I want to use it one for a toolbar (16x16) and once for a Wizard Images (usually 128x128) I can use the url
`bundle:/example.id/icons/obj16/search.svg?size=16x16` and` bundle:/example.id/icons/obj16/search.svg?size=128x128` to accomplish this task without the need to even store two SVGs

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLHintProvider.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLHintProvider.java
@@ -1,0 +1,50 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jface.resource;
+
+import java.net.URL;
+import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.swt.graphics.Point;
+
+class URLHintProvider implements Supplier<Point> {
+
+	private static final Pattern QUERY_PATTERN = Pattern.compile("&size=(\\d+)x(\\d+)"); //$NON-NLS-1$
+	private static final Pattern PATH_PATTERN = Pattern.compile("/(\\d+)x(\\d+)/"); //$NON-NLS-1$
+
+	private URL url;
+
+	public URLHintProvider(URL url) {
+		this.url = url;
+	}
+
+	@Override
+	public Point get() {
+		String query = url.getQuery();
+		Matcher matcher;
+		if (query != null && !query.isEmpty()) {
+			matcher = QUERY_PATTERN.matcher("&" + query); //$NON-NLS-1$
+		} else {
+			String path = url.getPath();
+			matcher = PATH_PATTERN.matcher(path);
+		}
+		if (matcher.find()) {
+			return new Point(Integer.parseInt(matcher.group(1)), Integer.parseInt(matcher.group(2)));
+		}
+		return null;
+	}
+
+}
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
@@ -25,6 +25,7 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -42,6 +43,7 @@ import org.eclipse.swt.graphics.ImageData;
 import org.eclipse.swt.graphics.ImageDataProvider;
 import org.eclipse.swt.graphics.ImageFileNameProvider;
 import org.eclipse.swt.graphics.ImageLoader;
+import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.internal.DPIUtil.ElementAtZoom;
 import org.eclipse.swt.internal.NativeImageLoader;
 import org.eclipse.swt.internal.image.FileFormat;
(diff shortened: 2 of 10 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLHintProvider.java
```diff
@@ -0,0 +1,50 @@
...
+		String query = url.getQuery();
+		Matcher matcher;
+		if (query != null && !query.isEmpty()) {
+			matcher = QUERY_PATTERN.matcher("&" + query); //$NON-NLS-1$
+		} else {
+			String path = url.getPath();
+			matcher = PATH_PATTERN.matcher(path);
+		}
+		if (matcher.find()) {
+			return new Point(Integer.parseInt(matcher.group(1)), Integer.parseInt(matcher.group(1)));
```
arunjose696: shouldnt one of them be matcher.group(2) ?

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java
```diff
@@ -147,8 +149,10 @@ private static ImageData getImageData(URL url, int fileZoom, int targetZoom) {
 	}
 
 	@SuppressWarnings("restriction")
-	private static ImageData loadImageFromStream(InputStream stream, int fileZoom, int targetZoom) {
-		return NativeImageLoader.load(new ElementAtZoom<>(stream, fileZoom), new ImageLoader(), targetZoom).get(0)
+	private static ImageData loadImageFromStream(InputStream stream, int fileZoom, int targetZoom,
+			Supplier<Point> hintProvider) {
+		return NativeImageLoader
```
arunjose696: NativeImageLoader already provides a method for loading images with a target width/height, which works for SVGs, couldnt this be used?

I am not clear why the zoom and the desired size both should be passed down, and then is the expectation that both this values should be always consistent. Wouldnt it make sense to pass the target size for svgs as query params and use zoom based `NativeImageLoader::load `method if the query params are null?

HeikoKlare: Please note that there is also a discussion going on in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2656, just to avoid that the same arguments are discussed twice at different places :-)

laeubi:
Some test are failing here because Equinox is not parsing query strings correctly, I tried to address this here:

- https://github.com/eclipse-equinox/equinox/pull/1217

HeikoKlare (approved):
The proposal looks fine for me now. I approve despite the currently failing tests since I will be on vacation soon and just want to document my opinion before that. I know that you will take of the tests and not merge before they run fine :-)

The SWT version constraint obviously makes sense because of the access to SWT internals which can change in any future release. That means we now have to update that dependency (bumping the upper bound if appropriate) for every new SWT release, right?

laeubi:
> That means we now have to update that dependency (bumping the upper bound if appropriate) for every new SWT release, right?

At least unless we have an official (stable) API for that.

selundqma:
Dear developers. Is this commit targeted to be included in the next Eclipse release (4.39)? BR

Dani-Hub:
@laeubi I have a question regarding your pull request: Can you confirm that evaluation of a possible query part of an image URI is restricted to protocols known by Eclipse (Such as "file", "bundle", etc)? If not, I'm worried, since in our company we have registered our own image URI schemes (which may or may not provide query parts) and your logic shouldn't affect those user-provided URIs because that could possibly lead to unintended code path changes within URLImageDescriptor and his friends.

laeubi:
Evaluation is not limited to any specific types because this would contradict the purpose. It works the same as the magic `NxN` and `@Z` what also is not restricted in any way. Of course it is limited to the SVG feature itself so I assume your custom URL will handle it differently anyways.

Dani-Hub:
> Evaluation is not limited to any specific types because this would contradict the purpose. It works the same as the magic `NxN` and `@Z` what also is not restricted in any way. Of course it is limited to the SVG feature itself so I assume your custom URL will handle it differently anyways.

Consider the following URI example (assuming that the scheme `scaled-image-id` has been registered): `scaled-image-id:bla-blub?w=32&h=32`. This URI has a query element and it will be handled by this new logic:
https://github.com/laeubi/eclipse.platform.ui/blob/a18db46e9032da33a8e1aec739b3d5e7e186de69/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java#L66
I'm yet not sure what the implications would be. I need to check this.

Dani-Hub:
> > Evaluation is not limited to any specific types because this would contradict the purpose. It works the same as the magic `NxN` and `@Z` what also is not restricted in any way. Of course it is limited to the SVG feature itself so I assume your custom URL will handle it differently anyways.
> 
> Consider the following URI example (assuming that the scheme `scaled-image-id` has been registered): `scaled-image-id:bla-blub?w=32&h=32`. This URI has a query element and it will be handled by this new logic: https://github.com/laeubi/eclipse.platform.ui/blob/a18db46e9032da33a8e1aec739b3d5e7e186de69/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java#L66 I'm yet not sure what the implications would be. I need to check this.

OK, I checked this. It looks fortunately harmless, because in this particular method we would otherwise enter

https://github.com/laeubi/eclipse.platform.ui/blob/a18db46e9032da33a8e1aec739b3d5e7e186de69/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java#L274

and then here

https://github.com/laeubi/eclipse.platform.ui/blob/a18db46e9032da33a8e1aec739b3d5e7e186de69/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLImageDescriptor.java#L289

but since our URI isn't a file URI, this would end up returning null anyway.

I'm still nervous about the genericy of the regexes in `URLHintProvider`. Could we possibly restrict them to be applied only to Eclipse-known URIs such as file, bundle, etc?

I think our `scaled-image-id` also can contain URIs such as `scaled-image-id:fasel-32x32`, which would unintentionally enter this branch:

https://github.com/laeubi/eclipse.platform.ui/blob/a18db46e9032da33a8e1aec739b3d5e7e186de69/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/URLHintProvider.java#L42

PR #3441, opened 2025-10-23
Mark removal for unused deprecated methods as well as some constants

deepika-u:


deepika-u:
@vogella 
Could you take a look at it when you get some time please?

vogella:
LGTM, planning to merge tomorrow, if no one objects

merks:
@vogella 

I think tonight's build will be M2 so probably it's better to do this today.

PR #3445, opened 2025-10-23
Add CLAUDE.md for repository guidance and architecture overview

vogella:


laeubi:
Can't claude reuse exiting github instructions? do we need an won file for every tool now? Just wondering...

vogella:
> Can't claude reuse exiting github instructions? do we need an won file for every tool now? Just wondering...

It usted it as input but it needs its own file. Btw. I tested both Copilot and Claude and IMHO results with Claude are much better. Usage limits are way stricter for Claude, so in case you want to do lot of stuff Copilot is still better, while quality-wise I favor Claude. Next month I plan to try out ChatGPT codex, which I hear also great things about (it has also a free month).

vogella:
Merging as it is just documentation.

laeubi:
Maybe one can just referenc in one file the other documentation? It seems a bit strange to have two "doumentations" that should be similar and could get out of sync.

Beside that for compilote you can enable to use more models, have not tried it yet but maybe it helps: 

https://docs.github.com/en/copilot/reference/ai-models/supported-models

vogella:
The quality difference (and the token usage) is not only based on the model but primary based on the system prompt which instructs the model to work differently. See here: https://github.com/x1xhlol/system-prompts-and-models-of-ai-tools

That is the reason why the same LLM results in different output based on the tooling you use

sebthom:
@vogella Why didn't you add an [AGENTS.md](https://agents.md/) that can be used by all AI tools and import it into an empty CLAUDE.md file via the `@AGENTS.md` directive?

vogella:
Because I didn't know about it @sebthom  :-) Can you send a PR for this?

PR #3457, opened 2025-10-26
disable redrawing when setting visible region to prevent flickering

danthe1st:
When `setVisibleRegion` is repeatedly [called with projection regions enabled](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3074), it first expands everything outside of the old visible region and then collapses everything inside the new visible region. This can result in Eclipse showing unwanted text for a fraction of a second.

Without this change:

[recording with flickering](https://github.com/user-attachments/assets/c25211aa-e9cb-4a49-bbfa-23cfa9168fcc)

With this change:

[recording without flickering](https://github.com/user-attachments/assets/e11873e3-c25d-47dc-8a1d-3810e22b257d)

(I made the recordings with https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2302 active to get the visible regions logic in JDT)

I don't think an automated test is reasonable for this but I can try to provide a way to reproduce this manually without JDT if necessary.

### DIFF
--- a/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
+++ b/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -767,6 +767,8 @@ public class ProjectionViewer extends SourceViewer implements ITextViewerExtensi
 		if (document == null) {
 			return;
 		}
+		int topIndex= redraws() ? getTopIndex() : -1;
+		setRedraw(false);
 		try {
 			// If the visible region changes, make sure collapsed regions outside of the old visible regions are expanded
 			// and collapse everything outside the new visible region
@@ -782,6 +784,8 @@ public class ProjectionViewer extends SourceViewer implements ITextViewerExtensi
 		} catch (BadLocationException e) {
 			ILog log= ILog.of(getClass());
 			log.log(new Status(IStatus.WARNING, getClass(), IStatus.OK, null, e));
+		} finally {
+			setRedraw(true, topIndex);
 		}
 	}
 


iloveeclipse (approved):

bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
```diff
@@ -768,6 +768,7 @@ public void setVisibleRegion(int start, int length) {
 			return;
 		}
 		try {
+			setRedraw(false);
```
iloveeclipse: I believe better to first get current "redraw" state, and restore to that state in finally. We had similar fix before somewhere else and that caused other issues because it switched redraw ON while it was already disabled before.

iloveeclipse (dismissed):

bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
danthe1st: @iloveeclipse 
It seems like `TextViewer#redraw` [already handles that](https://github.com/eclipse-platform/eclipse.platform.ui/blob/3a441d9e970383bfda58f254ccb794aa4942f490/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewer.java#L5262-L5280). For example, I could add the following code and get the output from the comments:
```java
System.out.println(redraws());//true
setRedraw(false);
System.out.println(redraws());//false
setRedraw(false);
System.out.println(redraws());//false
setRedraw(true);
System.out.println(redraws());//false
setRedraw(true);
System.out.println(redraws());//true
```

However, I could change the `setRedraw(true)` to `setRedraw(true, topIndex);` with `topIndex` being `redraws() ? getTopIndex() : -1` from before `setRedraw(false)` as done in `catchupWithProjectionAnnotationModel`

danthe1st: Done.

danthe1st:
(only rebased/updated copyright headers, still waiting for re-review)

vogella (approved):
The changes look good to me.

The logic correctly handles the redraw state and ensures the scroll position is maintained (or ignored if redraw was already disabled) using the `TextViewer.setRedraw(boolean, int)` API. The `try...finally` block ensures that redrawing is always restored.

Also, thanks for updating the copyright year.

vogella:
Thanks @danthe1st

PR #3464, opened 2025-10-28
Adapt size of Dialogs based on default size upon zoom change

### LINKED ISSUE #400: (Some) Dialogs are smaller than expected

HeikoKlare:
When moved between monitors of different zoom, `Dialog`s may not show their full contents anymore. This is caused by their shell and the used fonts being linearly scaled in size according to the zoom, even though font size and required area for a font of that size are not linearly related. I.e., a font of double the size requires more than double of the width and height.

Since there are many ways to customize `Dialog`s and their sizes/layouts during and after initialization, it's difficult to solve that problem in general (also see https://github.com/eclipse-platform/eclipse.platform.ui/pull/3212). The problem is, however, most severe for `Dialog`s that use the default calculated size and are not resizable, as those `Dialog`s do not have sophisticated layouts that adapt to the limited space in a shell and may easily lead to cut offs the user cannot work around by resizing the dialog. Thus, this change adds according zoom change and resize listeners to identify if exactly such `Dialog`s that use the default computed size and, in that case, recomputes it upon a zoom change.

This is supposed to be a conservative alternative to https://github.com/eclipse-platform/eclipse.platform.ui/pull/3212

### How to test

On a Windows system, the behavior can particularly be seen in any dialog in which the number of text lines increases when moving the dialog to a monitor with higher zoom.
For example, a `PlainMessageDialog` with an according message looks like this when moved from 100% monitor to 150% monitor without this change:
image

And it looks like this with the proposed change:
image

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/Dialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/Dialog.java
@@ -16,6 +16,7 @@ package org.eclipse.jface.dialogs;
 
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
@@ -40,6 +41,7 @@ import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Shell;
 
 /**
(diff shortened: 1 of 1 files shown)


arunjose696:
There was a comment in (https://github.com/eclipse-platform/eclipse.platform.ui/issues/3212#issuecomment-3271535000_) by @laeubi, I think even this might leave unresolved.

I had originally created a snippet to test this, could see my snippet also breaks with this branch when the dialogue has a table with scrollbars and it is statrted at 100% monitor and moved to 200% monitor.


```java
package org.eclipse.swt.snippets;
import org.eclipse.jface.dialogs.*;
import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;
import org.eclipse.swt.layout.*;

public class MessageDialogWithTable {

        shell.setText("Parent Shell");

        // Create the MessageDialog
        MessageDialog dialog = new MessageDialog(shell,
                "Large Table Dialog",   // title
                null,                   // dialog title image
                "This dialog contains a large scrollable table:", // message
                MessageDialog.INFORMATION,
                new String[] {"OK"},    // button labels
                0) {                    // default index

        	  @Override
              protected int getShellStyle() {
                  // Add RESIZE and MAX style bits
                  return super.getShellStyle() | SWT.RESIZE | SWT.MAX;
              }

            @Override
            protected Control createCustomArea(Composite parent) {
                Composite container = new Composite(parent, SWT.NONE);
                container.setLayout(new GridLayout(1, false));
                container.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));

                // Scrolled composite for the table
                ScrolledComposite scrolledComposite = new ScrolledComposite(container, SWT.BORDER | SWT.H_SCROLL | SWT.V_SCROLL);
                scrolledComposite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
                scrolledComposite.setExpandHorizontal(true);
                scrolledComposite.setExpandVertical(true);

                Composite tableContainer = new Composite(scrolledComposite, SWT.NONE);
                tableContainer.setLayout(new FillLayout());

                // Create the table
                Table table = new Table(tableContainer, SWT.BORDER | SWT.FULL_SELECTION);
                table.setHeaderVisible(true);
                table.setLinesVisible(true);

                // Add many columns
                for (int c = 0; c < 60; c++) {
                    TableColumn column = new TableColumn(table, SWT.NONE);
                    column.setText("Column " + (c + 1));
                    column.setWidth(100);
                }

                // Add many rows
                for (int r = 0; r < 100; r++) {
                    TableItem item = new TableItem(table, SWT.NONE);
                    for (int c = 0; c < 60; c++) {
                        item.setText(c, "Item " + (r + 1) + "," + (c + 1));
                    }
                }

                scrolledComposite.setContent(tableContainer);
                //scrolledComposite.setMinSize(tableContainer.computeSize(SWT.DEFAULT, SWT.DEFAULT));

                return container;
            }
        };

        dialog.open();
        display.dispose();
    }
}

```

akoch-yatta (approved):
The code to detect whether the dialog should be automatically resized or not is quite complex, but I do not see an easier way to achieve that. I would weigh the improved UX higher and approve it.
I tested the approach with different dialogs and it improves the UX for all dialogs I tested for that are affected by cut-off text. I did not experience any unwanted dialog resizing.

PR #3467, opened 2025-10-29
Mark several unused deprecated methods/constants/classes in org.eclipse.jface package for removal

deepika-u:


bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/InputDialog.java
```diff
@@ -171,7 +171,7 @@ protected Control createDialogArea(Composite parent) {
 	 * @return the error message label
 	 * @deprecated use setErrorMessage(String) instead
 	 */
-	@Deprecated
+	@Deprecated(forRemoval = true, since = "2025-12")
 	protected Label getErrorMessageLabel() {
```
merks, line 175: This is kind of odd because it doesn't actually specify a replacement and there is no replacement.

merks (changes requested):
I don't think we should remove a method for which there is no replacement.  The original deprecation seems bogus.

bundles/org.eclipse.jface/src/org/eclipse/jface/dialogs/InputDialog.java
deepika-u: oh ok. So reverting this change.
Some pointers if possible so that i shouldnt do similar updates once again.

deepika-u:
@merks Could you take a look at this when you get some time please.

deepika-u:
@merks 
Can you take a look at this when you get some time please.

vogella:
@merks can you see if the change is not good for you? @deepika-u can you re-trigger the build?

merks (changes requested):
When I see existing calls to newly to-be-removed methods I think we're going down the wrong path.

Image

I'm quite sure that screwing around with ImageDescriptor methods will be a bad thing downstream, and we have evidence of that even here.

merks:
I wanted to make a general point here as well.  If you mark something for removal, please ensure you have the full SDK IDE setup and that you look at the call hierarchy of each and every field, method, or class you plan to mark for deletion.  If you find even one use of it, you can assume that you will find 1000 uses of it out there in the ecosystem such that if you delete that thing, there will be things that simply stop working.  Then you should ask yourself, is there significant value in the removal that offsets the broken functionality downstream, keeping in mind that any broken functionality in any IDE for any component will generally create a bad impression overall; one that generally comes back to roost in the Platform.  Is it worth it?

laeubi:
Especially at least Platform itself **must** be free from any references first, as otherwise builds in the sub repositories will start to fail!

deepika-u:
If i run this QuickAccessComputerTest locally, with and without pr - in both cases test is passing.

deepika-u:
If i run locally the ProgressContantsTest, all the tests are passing locally with and without the pr.
image

vogella:
1 out of 3 runs failed: testBinaryContentTypeWithDescriberParallel (AllSearchTests AllFileSearchTests FileSearchTests) is known to be flaky. I have a look.

deepika-u:
> 1 out of 3 runs failed: testBinaryContentTypeWithDescriberParallel (AllSearchTests AllFileSearchTests FileSearchTests) is known to be flaky. I have a look.

Thanks for your input.

deepika-u:
@merks : can you take a look at this once you get some time please?

merks (approved):
Looks okay.

PR #2219, opened 2024-08-26
Fix StickyScrollingHandlerTest.testThrottledExecution on Mac

Christopher-Hermann:
The unit test StickyScrollingHandlerTest.testThrottledExecution was failing in the I build on MacOS. With this change, the test should run again.

Fixes #2190

PR #3474, opened 2025-10-30
Don't log error if info for deleted marker is not found

trancexpress:
Its possible that a marker is deleted, while its info is being retrieved. This will produce a logged error such as: "Marker id 6922 not found.", while the caller continues working without the info.

In this case, the logged error is not necessary - the marker is gone, not being able to retrieve its info is expected behavior.

This change adjusts MarkerUtilities to not log an error for a not found Marker, if the marker is not existing at the time of catching the respective exception.

Fixes: #2219

### DIFF
--- a/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/MarkerUtilities.java
+++ b/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/MarkerUtilities.java
@@ -31,6 +31,7 @@ import org.eclipse.core.runtime.Platform;
 
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceStatus;
 import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.IWorkspaceRunnable;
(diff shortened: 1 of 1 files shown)


laeubi:
@trancexpress :+1:  from me for getting rid of usless errors, removing a marker that is removed should simply be silent.

iloveeclipse (approved):

iloveeclipse:
Jenkims build error is unrelated.

```
21:47:37.862 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.12.0:jar (attach-javadocs) on project org.eclipse.ui.tests.navigator: MavenReportException: Error while generating Javadoc: 
21:47:37.862 [ERROR] Exit code: 1
21:47:37.862 [ERROR] error: Error fetching URL: https://docs.osgi.org/javadoc/osgi.core/8.0.0/ (java.net.ConnectException: Connection timed out)
21:47:37.862 [ERROR] 1 error
21:47:37.862 [ERROR] Command line was: /opt/tools/java/temurin/jdk-21/jdk-21.0.8+9/bin/javadoc -J-Duser.language
```

trancexpress:
The fixed issue here is actually: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3572

PR #3475, opened 2025-10-30
Limit monitor-specific scaling to supported autoscale modes

### LINKED ISSUE #399: Limit monitor-specific scaling to supported autoscale modes

ShahzaibIbrahim:
These changes have following affects:

- Upon starting eclipse product when monitor-specific scaling is set with unsupported autoscale mode, it will show a error dialog and application won't start.
- User won't be able to enable monitor-specific scaling from appearance page if unsupported autoscale mode is set from property or ini file.
- User can then either force the unsupported autoscale mode with monitor specific-scaling by setting swt.autoScale.force to true or switch to supported autoscale mode.

### Testing

___

**Scenario#1:** Starting application with  `-Dswt.autoScale=int200`

Result: App should start (if monitor-specific scaling is not checked in Appearance page), but option to turn on monitor-specific scaling in **Appearance** page should be disabled.

image

___

**Scenario#2:** Starting application with  `-Dswt.autoScale=int200` and  `-Dswt.autoScale.updateOnRuntime=true`

Result: App will start, user will see an error dialog stating the incompatibility of an option and monitor specific-scaling will be disabled.

image

___

**Scenario#3:** Starting application with  `-Dswt.autoScale=int200` and  `-Dswt.autoScale.updateOnRuntime=true` 
 and `-Dswt.autoScale.force=true`

Result: makes more sense with swt applications. if user really wants to use other scaling mode with monitor-specific scaling.

```

public class AutoScaleSnippet {

public static void main (String [] args) {
	System.setProperty("swt.autoScale.updateOnRuntime", "true");
	System.setProperty("swt.autoScale", "int200");
	System.setProperty("swt.autoScale.force", "true");
	Display display = new Display ();
	shell.setText("Autoscale snippet");
	shell.open ();
	while (!shell.isDisposed ()) {
		if (!display.readAndDispatch ()) display.sleep ();
	}
	display.dispose ();
}
```
___

**Scenario#4:** Starting application with  `-Dswt.autoScale=quarter` 

Result: App should start, and user should be able to turn on/off the monitor-specific scaling in **Appearance** page.

image

___

**Note: following is the list of allowed autoscale values to be paired with monitor-specific scaling: ["false", "quarter", "exact"]** or ***any concrete zoom value e.g. 225***

### Dependency

- [x] Requires: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2709

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
@@ -669,6 +669,13 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 				// run the legacy workbench once
 				returnCode[0] = workbench.runUI();
 
+				if (AUTOSCALE_ADAPTATION.isMonitorSpecificScalingDisabledForIncompatibility()) {
+					display.asyncExec(() -> {
+						MessageDialog.openError(null, WorkbenchMessages.RescaleAtRuntimeIncompatibilityTitle,
+								NLS.bind(WorkbenchMessages.RescaleAtRuntimeIncompatibilityDescription));
+					});
+				}
+
 				if (returnCode[0] == PlatformUI.RETURN_OK) {
 					// run the e4 event loop and instantiate ... well, stuff
 					if (serviceListener.get() != null) {
@@ -692,18 +699,6 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 			}
 		});
 		return returnCode[0];
-	}
-
-	private static void setRescaleAtRuntimePropertyFromPreference() {
-		if (System.getProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY) != null) {
-			WorkbenchPlugin.log(Status.warning(SWT_RESCALE_AT_RUNTIME_PROPERTY
-					+ " is configured (e.g., via the INI), but the according preference should be preferred instead." //$NON-NLS-1$
-			));
-		} else {
-			boolean rescaleAtRuntime = ConfigurationScope.INSTANCE.getNode(WorkbenchPlugin.PI_WORKBENCH)
-					.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, true);
-			System.setProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY, Boolean.toString(rescaleAtRuntime));
-		}
 	}
 
 	private static void setSearchContribution(MApplication app, boolean enabled) {
@@ -771,7 +766,7 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 			Display.setAppName(applicationName);
 		}
 
-		setRescaleAtRuntimePropertyFromPreference();
+		AUTOSCALE_ADAPTATION.setRescaleAtRuntimePropertyFromPreference();
 
 		// create the display
 		Display newDisplay = Display.getCurrent();
@@ -3673,6 +3668,8 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 	private static class AutoscaleAdaptation {
 		private static final String SWT_AUTOSCALE = "swt.autoScale"; //$NON-NLS-1$
 
+		private boolean incompatibleMonitorSpecificScalingDisabled;
+
 		private final String initialAutoScaleValue;
 
 		public AutoscaleAdaptation() {
@@ -3681,6 +3678,27 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 
 		public void runWithInitialAutoScaleValue(Runnable runnable) {
 			DPIUtil.runWithAutoScaleValue(initialAutoScaleValue, runnable);
+		}
+
+		public void setRescaleAtRuntimePropertyFromPreference() {
+			if (System.getProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY) != null) {
+				WorkbenchPlugin.log(Status.warning(SWT_RESCALE_AT_RUNTIME_PROPERTY
+						+ " is configured (e.g., via the INI), but the according preference should be preferred instead." //$NON-NLS-1$
+				));
+			} else {
+				boolean rescaleAtRuntime = ConfigurationScope.INSTANCE.getNode(WorkbenchPlugin.PI_WORKBENCH)
+						.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, true);
+				System.setProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY, Boolean.toString(rescaleAtRuntime));
+			}
+
+			if (DPIUtil.isMonitorSpecificScalingActive() && !DPIUtil.isSetupCompatibleToMonitorSpecificScaling()) {
+				incompatibleMonitorSpecificScalingDisabled = true;
+				System.setProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY, Boolean.toString(false));
+			}
+		}
+
+		public boolean isMonitorSpecificScalingDisabledForIncompatibility() {
+			return incompatibleMonitorSpecificScalingDisabled;
 		}
 
 	}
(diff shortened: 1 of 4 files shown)


ShahzaibIbrahim:
Test failures are due to dependent PR: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2709 not being merged yet.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
```diff
@@ -222,7 +225,16 @@ private void createRescaleAtRuntimeCheckButton(Composite parent) {
...
 				initialStateRescaleAtRuntime);
-		rescaleAtRuntime.setToolTipText(WorkbenchMessages.RescaleAtRuntimeDescription);
+		if (!DPIUtil.isSetupCompatibleToMonitorSpecificScaling()) {
+			rescaleAtRuntime.setEnabled(false);
+			Font font = parent.getFont();
+			Composite note = createNoteComposite(font, parent, WorkbenchMessages.Preference_note,
+					WorkbenchMessages.RescaleAtRuntimeDisabledDescription);
+			note.setLayoutData(GridDataFactory.swtDefaults().span(2, 1).create());
+		} else {
+			rescaleAtRuntime.setEnabled(true);
```
HeikoKlare: This can probably be removed as it should be enabled by default.

amartya4256 (approved):
I have tested the PR and it seems good to me. Approving.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -704,6 +704,13 @@ private static void setRescaleAtRuntimePropertyFromPreference() {
 					.getBoolean(IWorkbenchPreferenceConstants.RESCALING_AT_RUNTIME, true);
 			System.setProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY, Boolean.toString(rescaleAtRuntime));
 		}
+
+		if (DPIUtil.isMonitorSpecificScalingActive() && !DPIUtil.isSetupCompatibleToMonitorSpecificScaling()) {
+			System.setProperty(SWT_RESCALE_AT_RUNTIME_PROPERTY, Boolean.toString(false));
+			MessageDialog.openError(new Shell(Display.getDefault()),
+					WorkbenchMessages.RescaleAtRuntimeIncompatibilityTitle,
```
HeikoKlare: This should be deferred to when the workbench has been started, so that the dialog can be created on top of the workbench instead of creating a new shell on the default display. In particular because at this point in time the display for the workbench is not initialized yet, so this may have unintended side effects.

ShahzaibIbrahim: But we wanted to show the error dialog and **not open the workbench** due to incompatibility, right? 

If we do it after **workbench is created** then user can still use the IDE. Could you point me to where can it be deferred?

HeikoKlare: > But we wanted to show the error dialog and **not open the workbench** due to incompatibility, right?

No, that's also not how it is currently implement. In case of an incompatibility, monitor-specific scaling is simply disabled and the user should be informed about it, but the product will start anyway.

I took a look at the `Workbench` implementation and see that other dialogs are opened without specifying a `parent` at all. Maybe that could be a proper way? In any case, instantiating a `Display` at this point in time is invalid.

ShahzaibIbrahim: I have added a second commit to address this comment. Although I am not 100% sure If it address the issue you've mentioned. 

I have deferred showing the error dialog by using asyncExec instead of creating a new Shell immediately. This ensures the dialog is created after the display is initialized (and used for workbench), avoiding early-startup side effects and temporary detached shells. 

Side note: In setRescaleAtRuntimePropertyFromPreference method we set the updateOnRuntime value to **false** if it's not compatible and we cannot use the same conditions to check if unsupported autoscale combination was used. That's why I return true or false to preserve isIncompatibleAutoScaleValue.

HeikoKlare: @ShahzaibIbrahim I have added a proposal via a commit on top of the PR, which ...
- moves the call for opening a dialog to a point in time where the workbench window is already shown, such that the dialog appears on top of the workbench window, and ...
- embeds the functionality into the `AutoscaleAdaptation` class, such that the "error" state can be properly stored and retrieved

Can you please review and test this and if it is fine for you merge everything together?

HeikoKlare:
The many test failures seem to be because actually the `DPIUtil#isSetupCompatibleToMonitorSpecificScaling()` implementation is currently wrong. It erroneously considers the default autoScale value as incompatible:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/407cb9a1672e429a9ea3272e74bc5588b90e3f12/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/DPIUtil.java#L142-L145

That will be implicitly fixed with:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2771

For that reason, I propose to process that PR first. I wanted to merge it after this one to avoid that a consumer with a to-be-made-incompatible `swt.autoScale` value will not see the application start anymore instead of gracefully dealing with the situation as implemented in this PR, but given that's it will be integrated with just one day of delay and hopefully no up-to-date I-Build consumers use such outdated autoscale configurations anyway, it's probably okay to do it like that.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
ShahzaibIbrahim: @HeikoKlare I have tested it with different autoscale values (int200, quarter) with updateOnRuntime combinations (true, false, force). Everything seems to work as expected. This PR is good to be merged for me.

HeikoKlare (commented):
I have squashed the commits and adapted the commit message to only reflect the changes to Platform UI and not the ones to SWT.

PR #3496, opened 2025-11-04
Fix flaky testItemOrder in ProgressViewTests

### LINKED ISSUE #195: [Tests] ProgressViewTests.testItemOrder is flaky

vogella:
## Summary
- Fixes the race condition causing `ProgressViewTests.testItemOrder` to fail randomly
- Addresses the timing issues with `ProgressViewerComparator`'s stable sorting mechanism

## Root Cause
The test was failing due to a race condition with the `ProgressViewerComparator`'s `lastIndexes` HashMap:
- The comparator maintains previous positions to provide visual stability (preventing jobs from jumping in the UI)
- Throttled updates during job scheduling caused jobs to be added to `lastIndexes` in schedule order
- This created an inconsistent baseline where stable sorting preserved schedule order instead of priority order

## Solution
- Reopen the progress view after all jobs are scheduled and running to reset the comparator's state
- This ensures all jobs are sorted by priority from a clean `lastIndexes` state
- Removed the unreliable timing-based approach that tried to manipulate throttled updates with a dummy job

## Testing
- Test now passes reliably (verified with 3 consecutive runs)
- Works with the stable sorting behavior rather than against it

## Test Plan
- [x] Run `ProgressViewTests.testItemOrder` multiple times - all pass
- [x] Verified fix addresses the race condition described in issue

Fixes #195

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/progress/ProgressViewTests.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/progress/ProgressViewTests.java
@@ -25,7 +25,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.concurrent.TimeUnit;
 
-import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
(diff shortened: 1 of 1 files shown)


vogella:
@HeikoKlare any concern? I think this makes sense

PR #3497, opened 2025-11-04
Fix flaky test FileSearchTests.testBinaryContentTypeWithDescriber

### LINKED ISSUE #3490: Flacky Test: org.eclipse.search.tests.filesearch.FileSearchTests fails on windows
Tests run: 19, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 7.935 s <<< FAILURE! -- in AllSearchTests AllFileSearchTests FileSearchTests
org.eclipse.search.tests.filesearch.FileSearchTests.testDerivedFilesSerial -- Time elapsed: 0.966 s
org.eclipse.search.tests.filesearch.FileSearchTests.testBinaryContentTypeWithDescriberParallel -- Time elapsed: 1.828 s <<< FAILURE!
org.opentest4j.AssertionFailedError: Number of total results ==> expected: <1> but was: <2>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:563)
	at org.eclipse.search.tests.filesearch.FileSearchTests.testBinaryContentTypeWithDescriber(FileSearchTests.java:614)
	at org.eclipse.search.tests.filesearch.FileSearchTests.testBinaryContentTypeWithDescriberParallel(FileSearchTests.java:565)

vogella:
The test was failing intermittently on Windows with "expected: <1> but was: <2>" because it was searching the entire project scope where files from other concurrent tests could interfere.

Changes:
- Use unique timestamped folder name to avoid conflicts
- Narrow search scope to only the test's specific folder
- Add comments explaining the fix

This ensures test isolation when running in parallel.

Fixes #3490

### DIFF
--- a/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/FileSearchTests.java
+++ b/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/FileSearchTests.java
@@ -601,13 +601,16 @@ public class FileSearchTests {
 						.forEach(extension -> registry.removeExtension(extension, masterToken));
 			}) {
 
-				IFolder folder= ResourceHelper.createFolder(fProject.getFolder("folder1"));
+				// Use unique folder name to avoid conflicts with other tests running in parallel
+				String uniqueFolderName= "binaryContentTypeTest-" + java.util.UUID.randomUUID().toString();
+				IFolder folder= ResourceHelper.createFolder(fProject.getFolder(uniqueFolderName));
 				IFile textfile= ResourceHelper.createFile(folder, "textfile", "text hello");
 				IFile binaryfile= ResourceHelper.createFile(folder, "binaryfile", "binary hello");
 
 				Pattern searchPattern= PatternConstructor.createPattern("hello", true, false);
 
-				FileTextSearchScope scope= FileTextSearchScope.newSearchScope(new IResource[] { fProject }, (String[]) null, false);
+				// Search only in the unique folder to avoid interference from other tests
+				FileTextSearchScope scope= FileTextSearchScope.newSearchScope(new IResource[] { folder }, (String[]) null, false);
 				TextSearchEngine.create().search(scope, collector, searchPattern, null);
 
 				TestResult[] results= collector.getResults();


vogella:
@HeikoKlare @fedejeanne WDYT?

vogella:
@HeikoKlare any concerns?

fedejeanne (approved):

tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/FileSearchTests.java
```diff
@@ -601,13 +601,16 @@ private void testBinaryContentTypeWithDescriber(TestResultCollector collector) t
 						.forEach(extension -> registry.removeExtension(extension, masterToken));
 			}) {
 
-				IFolder folder= ResourceHelper.createFolder(fProject.getFolder("folder1"));
+				// Use unique folder name to avoid conflicts with other tests running in parallel
+				String uniqueFolderName= "binaryContentTypeTest-" + System.currentTimeMillis();
+				IFolder folder= ResourceHelper.createFolder(fProject.getFolder(uniqueFolderName));
```
HeikoKlare: These tests are JUnit 5, so wouldn't it be better to use the TempDirectory extension to create unique temporary directories? Using `currentTimeMillis()` to avoid conflicts seems to be prone to errors as well, as no one can guarantee that two tests methods do not call this method at the same time (same time = even sequential but at same system time millis). Note that this private methods is called from two methods which could thus both produce the exact same folder name.

fedejeanne: Just a practical thought: how big are the chances of a name collision happening? We have other tests doing the same trick and they are stable.

sratz: If it has to be under the `fProject` folder, then maybe just use `java.util.UUID.randomUUID().toString()`?

HeikoKlare: I had many tests failing for such a reason when stabilizing the Platform tests two years ago. Tests might be stable with such a hack, some are like that forever, some are until you move to a different infrastructure (which was the reason for the Platform tests starting to get flaky because they were suddently executed on random machines via GHA instead of dedicated servers running Jenkins).
And in any case, when investing the time into fixing a test anyway, I would be in favor of doing it properly rather than doing something that is prone to fail later on again. But still, better have a fixed test for now than not having any improvement at all, so please take it as a hint/proposal and not as a request from my side.

vogella: Thanks @HeikoKlare @fedejeanne and @sratz Proposal looks great to me, I will ask Claude AI to update the PR with the proposal.

vogella: Isn't it nice that Claude code takes the same amount of time to think about @sratz proposal, agree to it and adjusting the PR, them me writing the above statement?

HeikoKlare (approved):

vogella:
Thanks everybody for the review and help.

PR #3498, opened 2025-11-04
Revert "Workaround for unexpected scrollbars in MultipleHyperlink"

### LINKED ISSUE #355: Replace Workaround for MultipleHyperlinkPresenter
**Description**
In [this](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2998) PR we added a workaround to get rid of unnecessary scrollbars. With the introduction of float precision in some rectangle, we should evaluate how a proper solution (perhaps with API changes) would look for this issue

**Reproduction**
**Without the workaround** with certain zoom settings (e.g., 125%, 175% or 225%, usually primary must be smaller than the secondary), the calculated size may be too small, causing the Composite to show Scrollbars, although it calculated to not need scrollbars. 
Image
**With the workaround** the Shell can be way too big (Primary 150%, secondary 100%):
Image

**Expected Behavior**
No workaround added in the code
No unnecessary scrollbars in the Shell
Shell not wider than expected

**Necessary configuration:**

Depending on the scenario, monitor specific scaling is active + different zoom on primary and secondary

**Additional knowledge**

ShahzaibIbrahim:
The workaround was introduced to resolved the unexpected scrollbars in MultiHyperLinks but later it was fixed as a result of fix in https://github.com/eclipse-platform/eclipse.platform.swt/commit/56379d5edf29a33b9b7cd4d639a921f2b2465820

### How to test
Use this Snippet and open it on a 225% monitor.

```
package test;

public class TestClass {
		// TODO Auto-generated method stub
	}
}
```

Hover over main and press control key. 

### Result

I have tested it with primary monitor = 150% and secondary = 125, 150, 175, 200, 225, 250, 300.

### DIFF
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.text
-Bundle-Version: 3.29.0.qualifier
+Bundle-Version: 3.29.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/hyperlink/MultipleHyperlinkPresenter.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/hyperlink/MultipleHyperlinkPresenter.java
@@ -153,14 +153,6 @@ public class MultipleHyperlinkPresenter extends DefaultHyperlinkPresenter implem
 			fForegroundColor= foregroundColor;
 			fBackgroundColor= backgroundColor;
 			create();
-			getShell().addListener(SWT.ZoomChanged, event -> {
-				final Shell shell= getShell();
-				shell.getDisplay().asyncExec(() -> {
-					if (!shell.isDisposed()) {
-						shell.setSize(computeSizeHint());
-					}
-				});
-			});
 		}
 
 		@Override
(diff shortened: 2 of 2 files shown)


HeikoKlare (commented):
Can you please explain the relation to https://github.com/eclipse-platform/eclipse.platform.swt/pull/2324 and how to test for regression against ist?
I just reverted that change and applied this PR, which should, according to the PR description, reintroduce the mentioned issue. However, everything looks fine for me (with 150% primary monitor and 225% secondary monitor on both of them).

arunjose696:
Sorry for any confusion. Referring to my comment in [vi-eclipse/Eclipse-Platform#355](https://github.com/vi-eclipse/Eclipse-Platform/issues/355#issuecomment-3135153483)

The issue fixed by [eclipse-platform/eclipse.platform.swt#2324](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2324)  is a completely different one  it addressed scrollbars appearing even when the workaround was applied.

The issue that was originally solved by the workaround from #2998  appears to be fixed now due to [56379d5edf29a33b9b7cd4d639a921f2b2465820](https://github.com/eclipse-platform/eclipse.platform.swt/commit/56379d5edf29a33b9b7cd4d639a921f2b2465820). Since pixelToPoint now rounds up instead of down so it adds one pixel.

However, I think this issue should be kept in mind if we plan to modify rounding behavior  in pixelToPoint

ShahzaibIbrahim:
@arunjose696 is right in this case. I tried reverting https://github.com/eclipse-platform/eclipse.platform.swt/commit/56379d5edf29a33b9b7cd4d639a921f2b2465820 and at 225% I got these scroll bars again. So this commit what fixed this issue and workaround is not needed anymore. 

image

HeikoKlare:
The mentioned commit was basically just a partial revert of what was done 2 week before: https://github.com/eclipse-platform/eclipse.platform.swt/commit/55b5adf00dc0662989eac47d31bff42680c12aae

So that one is not the original fix for the scrollbar issue. Without knowing which original commit properly fixed the scrollbar issue, I am a bit reluctant to revert the workaround, as we do not know how and whether it was properly fixed.

ShahzaibIbrahim:
I agree, In this case I am also not sure what commit really fixed it and why we don't need the workaround anymore.

HeikoKlare (approved):
As the workaround should be obsolete with all the SWT improvements in the recent months and since we are right at the beginning of the development cycle with plenty of time to test, let's remove this workaround now to clean up the code and see if anything breaks again.

PR #3506, opened 2025-11-06
Remove unused method

elsazac:
The `loadFilePatternDefaults()` method has been commented out since 2006 and is no longer referenced or used anywhere in the codebase. Hence removing it as a part of cleanup.


vogella:
LGTM, will merge if the build is green. Thanks @elsazac

vogella:
Thanks @elsazac

PR #3511, opened 2025-11-07
Remove unnecessary quotes from link text

elsazac:
The single quotes around the placeholder text in the preference page were likely added following older UI conventions. Quotes often differentiate literal text or placeholders from the rest of the message. Since the text already appears as a blue clickable link, the additional quotes feels unnecessary. Removing them keeps the link still clickable while making the message cleaner and consistent with other preference links across the UI, where such quotes are not used.

eg : most of the links in preference page appear this way ->
Screenshot 2025-11-06 at 7 34 26 PM

Similarly this brings a constant behaviour across UI,

Before:
image

After:
image


BeckerWdf:
This UI cleanup.

vogella:
Thanks @elsazac

PR #3515, opened 2025-11-07
Performance: Cache method calls in loops and optimize string concaten…

vogella:
Apply low-risk performance optimizations across multiple bundles.

After review, this PR contains only the legitimate performance improvements:

**Repeated method call caching:**
- MenuManagerShowProcessor: Cache menuModel.getChildren() to avoid repeated method calls in loop condition
- WorkbenchPage: Cache getChildren() results to avoid repeated method calls in loops

**String concatenation optimizations:**
- WizardsRegistryReader: Use StringBuilder for path construction in loop instead of repeated string concatenation

All changes are non-API breaking and maintain identical behavior while reducing unnecessary object allocations and method calls.

**Note:** Changes that provided no actual benefit (caching simple field accesses like List.size(), String.length(), or unnecessary variable extractions) have been removed.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/MenuManagerShowProcessor.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/MenuManagerShowProcessor.java
@@ -123,8 +123,8 @@ public class MenuManagerShowProcessor implements IMenuListener2 {
 	 * {@link MDynamicMenuContribution} application model elements
 	 */
 	private void processDynamicElements(MMenu menuModel, MenuManager menuManager) {
-		MMenuElement[] menuElements = menuModel.getChildren().toArray(
-				new MMenuElement[menuModel.getChildren().size()]);
+		List<MMenuElement> children = menuModel.getChildren();
+		MMenuElement[] menuElements = children.toArray(new MMenuElement[children.size()]);
 		for (MMenuElement currentMenuElement : menuElements) {
 
 			if (currentMenuElement instanceof MDynamicMenuContribution dmc) {
@@ -154,9 +154,9 @@ public class MenuManagerShowProcessor implements IMenuListener2 {
 				if (mel.size() > 0) {
 
 					int position = 0;
-					while (position < menuModel.getChildren().size()) {
-						if (currentMenuElement == menuModel.getChildren().get(
-								position)) {
+					children = menuModel.getChildren();
+					while (position < children.size()) {
+						if (currentMenuElement == children.get(position)) {
 							position++;
 							break;
 						}
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPage.java
@@ -3423,16 +3423,18 @@ public class WorkbenchPage implements IWorkbenchPage {
 		// Hide placeholders for parts that exist in the 'global' areas
 		modelService.hideLocalPlaceholders(window, dummyPerspective);
 
-		int dCount = dummyPerspective.getChildren().size();
-		while (!dummyPerspective.getChildren().isEmpty()) {
-			MPartSashContainerElement dChild = dummyPerspective.getChildren().remove(0);
-			persp.getChildren().add(dChild);
+		List<MPartSashContainerElement> dummyChildren = dummyPerspective.getChildren();
+		List<MPartSashContainerElement> perspChildren = persp.getChildren();
+		int dCount = dummyChildren.size();
+		while (!dummyChildren.isEmpty()) {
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/action/ContributionManager.java
```diff
@@ -272,7 +272,8 @@ protected boolean hasDynamicItems() {
 	 * @return <code>int</code> the index or -1 if the item is not found
 	 */
 	public int indexOf(String id) {
-		for (int i = 0; i < contributions.size(); i++) {
+		int size = contributions.size();
```
BeckerWdf: isn't this something the compilier anyway optimizes?

bundles/org.eclipse.jface/src/org/eclipse/jface/bindings/BindingManager.java
```diff
@@ -1339,8 +1339,9 @@ public TriggerSequence getBestActiveBindingFor(final ParameterizedCommand comman
 			}
 
 			// If this is still a tie, then just chose the shortest text.
-			compareTo = bestTriggerSequence.format().length()
-					- currentTriggerSequence.format().length();
+			int bestLength = bestTriggerSequence.format().length();
+			int currentLength = currentTriggerSequence.format().length();
+			compareTo = bestLength - currentLength;
```
BeckerWdf: why should this be faster? the number of method calls does not change.

bundles/org.eclipse.jface/src/org/eclipse/jface/action/ContributionManager.java
merks: Not likely, the size could change in the loop.   I personally prefer this style:
```
for (int i = 0, size = contributions.size(); i < size; i++) {
```
**Assuming** the intent is not to deal with the size changing in the loop.

vogella:
Thanks for the early feedback but I would prefer if draft reviews are not pro-actively reviewd. For example, the  currentTriggerSequence.format().length(); change is obviously non-sense, I only did not yet find the time to update the PR.

So better spend time in reviewing non-draft PR to avoid spending time on something that is not yet ready to be reviewed.

vogella:
These changes won't be noticeable to end users. They're technically correct optimizations, but the performance gain is likely unmeasurable. The main value is in code quality and avoiding unnecessary allocations, not in user-perceivable performance improvement.

laeubi:
> WizardsRegistryReader: Use StringBuilder for path construction in loop instead of repeated string concatenation

We have 

grafik

just as you asked here https://github.com/eclipse-platform/eclipse.platform.ui/issues/3517 I would suggest to enable such cleanups on a more general scope.

vogella:
> > WizardsRegistryReader: Use StringBuilder for path construction in loop instead of repeated string concatenation
> 
> We have
> 
> grafik
> just as you asked here #3517 I would suggest to enable such cleanups on a more general scope.

I tried this once in the past and it create tons of silly changes. I don't think this is one of the clean-up I would like to activate. The changes here are more dedicated.

laeubi:
BTW we also have

grafik

vogella:
Please put suggestions for new clean-up into this issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3517

laeubi:
> I tried this once in the past and it create tons of silly changes. 

Pick one example, report it to JDT and it can be fixed (maybe even with the help of the AI).

@jjohnstn did a great job lately, but without _using_ the cleanups and _reporting_ issues things will never improve.

> The changes here are more dedicated.

Dedicated to what? You said its all likely not noticeable by the user so claim they improve "performance" is maybe bit to big thing to claim.

But if the real goal is as claimed here

> The main value is in code quality and avoiding unnecessary allocations

then it would be much more valuable to:

1. Use this particular example
2. Apply the cleanup and see if it works or not
3. report any issues and let them be fixed
4. loop to 1 until happy
5. 5. enable the cleanup to improve on a larger scale.

vogella:
> > I tried this once in the past and it create tons of silly changes.
> 
> Pick one example, report it to JDT and it can be fixed (maybe even with the help of the AI).

The conversion was not wrong but for places which not performance relevant  like debug output and increased the code complexity IMHO. Like always you are free to investigate yourself (to which you usually reply you do not have the time for that which is valid and also the reason why I do not plan to work on this as I also do not have the time for this).

laeubi:
> The conversion was not wrong but for places which not performance relevant like debug output and increased the code complexity IMHO

This is a valid concern that could probably be addressed

> Like always you are free to investigate yourself 

You asked for where you are could have used cleanups and I just gave you examples, apart from that I always have incrementally enabled cleanups and investigate what are useful and reporting bugs:

- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues?q=is%3Aissue%20state%3Aclosed%20author%3Alaeubi%20cleanup

there is even one (open) enhancement request:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2232

So yes everyone is busy but especially then we should not have 100'd of fragmented micro optimizations.

vogella:
Trying to address the flaky Mac test here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3532

PR #3518, opened 2025-11-08
Prompt user when switching to a locked workspace in Eclipse

SougandhS:
This change prompts the user when attempting to switch to a workspace in Eclipse that is currently locked by another instance, preventing Eclipse from restarting into the locked workspace. Previously, if Eclipse restarted with a locked workspace, the workspace lock prompt would appear, and clicking “OK” would exit the application, requiring the user to reopen Eclipse to select an active workspace. With this update, the user is warned before switching, informed about the lock via a dialog, and prevented from restarting into the locked workspace, improving workflow and avoiding unnecessary restarts.

Before :

https://github.com/user-attachments/assets/b719d795-ad60-4189-a31c-bb547c67d4f9

After : 
(commented out the initial development mode check for this video)

https://github.com/user-attachments/assets/da95abb1-119f-463f-8407-5f791f3bb0f5

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
@@ -2694,7 +2694,13 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 	 * as the workspace location.
 	 *
 	 * @param workspacePath the new workspace location
-	 * @return {@link IApplication#EXIT_OK} or {@link IApplication#EXIT_RELAUNCH}
+	 * @return <code>null</code> if application was started in development mode and
+	 *         the specified workspace path is different from the current one,
+	 *         otherwise {@link IApplication#EXIT_RELAUNCH}.
+	 *         <p>
+	 *         In case of <code>null</code> return, a message dialog will be shown,
+	 *         so the caller doesn't need to handle the error case via dialog.
+	 *         </p>
 	 */
 	@SuppressWarnings("restriction")
 	public static Object setRestartArguments(String workspacePath) {
@@ -2706,10 +2712,6 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 		}
 
 		String command_line = Workbench.buildCommandLine(workspacePath);
-		if (command_line == null) {
-			return IApplication.EXIT_OK;
-		}
-
 		System.setProperty(Workbench.PROP_EXIT_CODE, IApplication.EXIT_RELAUNCH.toString());
 		System.setProperty(IApplicationContext.EXIT_DATA_PROPERTY, command_line);
 		return IApplication.EXIT_RELAUNCH;
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkspaceLock.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkspaceLock.java
@@ -1,0 +1,143 @@
+/*******************************************************************************
+ * Copyright (c) 2025 IBM Corporation.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.ui.internal;
+
+import java.io.InputStream;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Properties;
+import org.eclipse.core.runtime.URIUtil;
+import org.eclipse.jface.dialogs.MessageDialog;
+import org.eclipse.osgi.util.NLS;
+import org.eclipse.swt.widgets.Shell;
+
+/**
+ * Utility class for reading and presenting workspace lock information.
+ *
+ * <p>
+ * This class is used during two different phases of the Eclipse application
+ * lifecycle:
+ * </p>
+ * <ul>
+ * <li>before the Workbench is created (no workbench windows exist, such as when
+ * the Workspace Launcher dialog is displayed)</li>
+ * <li>after the Workbench has been created and workbench windows are
+ * available</li>
+ * </ul>
+ *
+ * <p>
+ * To support both environments, this class does not rely on workbench-specific
+ * APIs such as {@code PlatformUI.getWorkbench()} or {@code IWorkbenchWindow},
+ * nor on any API that requires an initialized workbench window. Only SWT-level
+ * constructs (for example, {@link org.eclipse.swt.widgets.Display} and
+ * {@link org.eclipse.swt.widgets.Shell}) and core/runtime APIs are used.
+ * </p>
+ *
+ */
+public class WorkspaceLock {
+
+	public static final String PROCESS_ID = "process-id"; //$NON-NLS-1$
+
+	public static final String DISPLAY = "display"; //$NON-NLS-1$
+
+	public static final String HOST = "host"; //$NON-NLS-1$
+
+	public static final String USER = "user"; //$NON-NLS-1$
+
+	/**
+	 * Read workspace lock file of the selected workspace if it is locked by another
+	 * Eclipse application and parse all the properties present. Based on the
+	 * Eclipse version and operating system some or all the properties may be not
+	 * present. In such scenario empty string may be returned.
+	 *
+	 * @param workspaceUrl the <code>URL</code> of workspace to check for lock
+	 *                     details
+	 * @return <code>null</code> if workspace is not locked, empty string if
+	 *         workspace is locked but no details are available (or lock file is not
+	 *         accessible), or a formatted string with lock details
+	 */
+	public static String getWorkspaceLockDetails(URL workspaceUrl) {
+		Path lockFile = getLockInfoFile(workspaceUrl);
+		if (lockFile == null || !Files.exists(lockFile)) {
+			return null;
+		}
+		StringBuilder lockDetails = new StringBuilder();
+		try (InputStream is = Files.newInputStream(lockFile)) {
+			Properties properties = new Properties();
+			properties.load(is);
+			String prop = properties.getProperty(USER);
+			if (prop != null) {
+				lockDetails.append(NLS.bind(WorkbenchMessages.IDEApplication_workspaceLockOwner, prop));
+			}
+			prop = properties.getProperty(HOST);
+			if (prop != null) {
+				lockDetails.append(NLS.bind(WorkbenchMessages.IDEApplication_workspaceLockHost, prop));
+			}
+			prop = properties.getProperty(DISPLAY);
+			if (prop != null) {
+				lockDetails.append(NLS.bind(WorkbenchMessages.IDEApplication_workspaceLockDisplay, prop));
+			}
+			prop = properties.getProperty(PROCESS_ID);
+			if (prop != null) {
+				lockDetails.append(NLS.bind(WorkbenchMessages.IDEApplication_workspaceLockPID, prop));
+			}
+		} catch (Exception e) {
+			WorkbenchPlugin.log("Could not read lock info file: " + lockFile, e); //$NON-NLS-1$
+		}
+		return lockDetails.toString();
+	}
+
+	/**
+	 * Returns the lock file.
+	 *
+	 * @param workspaceUrl the <code>URL</code> of selected workspace
+	 * @return the path to the <code>.lock_info</code> file within the specified
+	 *         workspace, or <code> null</code> if the workspace URL cannot be
+	 *         converted to a valid URI
+	 */
+	public static Path getLockInfoFile(URL workspaceUrl) {
+		Path lockFile = Path.of(".metadata", ".lock_info"); //$NON-NLS-1$ //$NON-NLS-2$
+		try {
+			return Path.of(URIUtil.toURI(workspaceUrl)).resolve(lockFile);
+		} catch (URISyntaxException e) {
+			return null;
+		}
+	}
+
+	/**
+	 * Opens an error dialog indicating that the selected workspace is locked by
+	 * another Eclipse instance.
+	 * <p>
+	 * This method works in both early startup (before the Workbench is created) and
+	 * in normal runtime (after Workbench windows exist).
+	 * </p>
+	 *
+	 * @param shell              the parent {@link Shell} for the dialog, or
+	 *                           {@code null} if no workbench window is available
+	 * @param workspacePath      the absolute path of the workspace that could not
+	 *                           be locked
+	 * @param workspaceLockOwner a formatted description of the existing lock owner
+	 */
+	public static void showWorkspaceLockedDialog(Shell shell, String workspacePath, String workspaceLockOwner) {
+		String lockMessage = NLS.bind(WorkbenchMessages.IDEApplication_workspaceCannotLockMessage2, workspacePath);
+		String wsLockedError = lockMessage + System.lineSeparator() + System.lineSeparator()
+				+ NLS.bind(WorkbenchMessages.IDEApplication_workspaceLockMessage, workspaceLockOwner);
+
+		MessageDialog.openError(shell,
+				WorkbenchMessages.IDEApplication_workspaceCannotLockTitle, wsLockedError);
+	}
+
+}
--- a/bundles/org.eclipse.ui.ide.application/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.ide.application/META-INF/MANIFEST.MF
@@ -8,7 +8,7 @@ Bundle-Localization: plugin
 Require-Bundle: org.eclipse.ui.ide;bundle-version="[3.21.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.core.resources;bundle-version="[3.19.0,4.0.0)",
- org.eclipse.ui;bundle-version="[3.208.0,4.0.0)",
+ org.eclipse.ui;bundle-version="[3.209.0,4.0.0)",
  org.eclipse.ui.navigator.resources;bundle-version="[3.9.0,4.0.0)",
  org.eclipse.core.net;bundle-version="[1.5.0,2.0.0)",
  org.eclipse.core.filesystem;bundle-version="1.10.0",
--- a/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
+++ b/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2020 IBM Corporation and others.
+ * Copyright (c) 2003, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -24,11 +24,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.InetAddress;
 import java.net.MalformedURLException;
-import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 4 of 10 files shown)


SougandhS:
Hi @iloveeclipse, could you please check this PR when you get some time ?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -630,8 +635,7 @@ public void update() {
 					}
 				};
 				registration[0] = FrameworkUtil.getBundle(WorkbenchPlugin.class).getBundleContext()
-						.registerService(StartupMonitor.class.getName(), startupMonitor,
-						properties);
+						.registerService(StartupMonitor.class.getName(), startupMonitor, properties);
```
iloveeclipse: Could you please undo all unrelated code changes, formatting and whatever else not connected to the problem?

iloveeclipse (commented):
I can check on Monday. The idea itself is good.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
merks: it's often annoying that the save actions do things to the entire file and it does that every time one edits anything in the file (except via the compare editor) which make it hard to commit "just the changes".

SougandhS: Done

BeckerWdf:
would it be possible to bring up the exact same dialog (with the same information) in both situations?

SougandhS:
> would it be possible to bring up the exact same dialog (with the same information) in both situations?

Sure 👍

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2699,7 +2704,8 @@ private static String buildCommandLine(String workspace) {
 	 * as the workspace location.
 	 *
 	 * @param workspacePath the new workspace location
-	 * @return {@link IApplication#EXIT_OK} or {@link IApplication#EXIT_RELAUNCH}
+	 * @return {@link IApplication#EXIT_OK} or {@link IApplication#EXIT_RELAUNCH} or
+	 *         <code>null</code>
```
iloveeclipse: `null` breaks logic in `org.eclipse.ui.internal.ide.application.IDEApplication.checkInstanceLocation(Shell, Map)`, however I see the `null` was also added in be51d14c6f825ce5f072dcbf771449620e789b6a and javadoc was not updated too. See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1622 + https://github.com/eclipse-platform/eclipse.platform.ui/pull/1622.

SougandhS:
> would it be possible to bring up the exact same dialog (with the same information) in both situations?

Updated as suggested 

https://github.com/user-attachments/assets/dc8c411b-8974-4a36-88f6-79bf434d9afc

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2714,12 +2721,88 @@ public static Object setRestartArguments(String workspacePath) {
...
+				String lockMessage = NLS.bind(WorkbenchSWTMessages.IDEApplication_workspaceCannotLockMessage2,
+						workspacePath);
+				String wsLockedError = lockMessage + System.lineSeparator() + System.lineSeparator()
+						+ NLS.bind(WorkbenchSWTMessages.IDEApplication_workspaceLockMessage,
+						workspaceLock);
+
+				MessageDialog.openError(PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
+						WorkbenchSWTMessages.IDEApplication_workspaceCannotLockTitle,
+						wsLockedError);
+				return null;
```
iloveeclipse: This method is used also in `IDEApplication.checkInstanceLocation()` which uses `null` internally as "valid workspace is selected".

Looks like `null` breaks logic in `org.eclipse.ui.internal.ide.application.IDEApplication.checkInstanceLocation(Shell, Map)`, however I see the `null` was also added in be51d14c6f825ce5f072dcbf771449620e789b6a and javadoc was not updated too. See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1307 + https://github.com/eclipse-platform/eclipse.platform.ui/pull/1622.

The use of `null` should have been documented & validated for  `IDEApplication.checkInstanceLocation()` usage in be51d14c6f825ce5f072dcbf771449620e789b6a, but latest now it is a must have.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2714,12 +2721,88 @@ public static Object setRestartArguments(String workspacePath) {
...
+		try {
+			String workspaceLock = getWorkspaceLockDetails(selectedWorkspace.toUri().toURL());
+			if (workspaceLock != null && !workspaceLock.isEmpty()) {
+				String lockMessage = NLS.bind(WorkbenchSWTMessages.IDEApplication_workspaceCannotLockMessage2,
+						workspacePath);
+				String wsLockedError = lockMessage + System.lineSeparator() + System.lineSeparator()
+						+ NLS.bind(WorkbenchSWTMessages.IDEApplication_workspaceLockMessage,
+						workspaceLock);
+
+				MessageDialog.openError(PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
```
iloveeclipse: The code is also used from `IDEApplication.checkInstanceLocation()` where no active workbench window exists.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2714,12 +2721,88 @@ public static Object setRestartArguments(String workspacePath) {
...
+	/**
+	 * Extract the lock details of the selected workspace if it is locked by another
+	 * Eclipse application
+	 *
+	 * @param workspaceUrl the <code>URL</code> of selected workspace
+	 * @return <code>String</code> details of lock owned workspace,
+	 *         <code>null or Empty</code> if not locked
+	 */
+	@SuppressWarnings("restriction")
+	private static String getWorkspaceLockDetails(URL workspaceUrl) {
```
iloveeclipse: COpying code from `org.eclipse.ui.internal.ide.application.IDEApplication.getWorkspaceLockInfo(URL)` is the works thing one can do.

Please refactor common code, not copy it around with slightly changed logic.

iloveeclipse (changes requested):

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
SougandhS: `org.eclipse.ui.internal.ide.application.IDEApplication` is not available in `o.e.workbench.ui` , when I made `IDEApplication.getWorkspaceLockInfo(URL)` as static and did a Import/required bundle in `o.e.workbench.ui` `Manifest`, Eclipse then behaves strange, it executes the `workbench.java` changes of master rather than my local changes. - is this a bug or am I missing something ?

iloveeclipse: ide can use & uses workbench, other way around it is not possible

SougandhS: Thanks, I have refactored `IDEApplication`

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
SougandhS: refactored

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
SougandhS: i have updated the javadoc a bit

SougandhS:
@iloveeclipse Could u please re-check this ?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -3684,4 +3701,72 @@ public void runWithInitialAutoScaleValue(Runnable runnable) {
...
+		}
+	}
+
+	@SuppressWarnings("restriction")
+	public static void showWorkspaceLockedDialog(String workspacePath, String workspaceLock) {
+		String lockMessage = NLS.bind(WorkbenchSWTMessages.IDEApplication_workspaceCannotLockMessage2, workspacePath);
+		String wsLockedError = lockMessage + System.lineSeparator() + System.lineSeparator()
+				+ NLS.bind(WorkbenchSWTMessages.IDEApplication_workspaceLockMessage, workspaceLock);
+
+		MessageDialog.openError(PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
```
iloveeclipse: I've already commented before, if the workspace selection dialog is opened, there is **NO** active workbench window, not a single window is opened!

I would recommend to move all this workspace locking code to extra class and add a bold statement on the class javadoc that it is used in both workbench / window - free and workbench - created use cases.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2699,7 +2705,8 @@ private static String buildCommandLine(String workspace) {
 	 * as the workspace location.
 	 *
 	 * @param workspacePath the new workspace location
-	 * @return {@link IApplication#EXIT_OK} or {@link IApplication#EXIT_RELAUNCH}
+	 * @return {@link IApplication#EXIT_OK} or {@link IApplication#EXIT_RELAUNCH} or
+	 *         <code>null</code>
```
iloveeclipse: While the returning constants is self explanatory, what does "return null" mean for the caller? NO exit? Don't care? Leave me alone with my problems? 
Just imagine you would see this javadoc and not see the implementation. What you would think about this API?

iloveeclipse (changes requested):
I believe M3 is the bad time for changing workspace locking code. Take time, **test** the change also please with the workspace selection dialog that is shown on Eclipse startup, and if everything will work, it could land in the next release.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
SougandhS: Will add few more details then

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
SougandhS: handled as suggested 👍

SougandhS:
> I believe M3 is the bad time for changing workspace locking code. Take time, **test** the change also please with the workspace selection dialog that is shown on Eclipse startup, and if everything will work, it could land in the next release.

I have tested the change with a new SDK build and is even working with Eclipse startup case
image

<br>

image

iloveeclipse:
I've moved messages to the right bundle & rebased on master.

TODO (for me)
* Update bundle version requirements for ide and application bundles on latest workbench
* Code review
* Behavior validation

@SougandhS : would it be OK if I would squash everything to one commit (except bundle version updates)?

iloveeclipse:
> would it be OK if I would squash everything to one commit (except bundle version updates)?

We must do that, the API tools sees "touched" code which was reverted and complains about missing version bump.

```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.3-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.ui.workbench.swt: Only qualifier changed for (org.eclipse.e4.ui.workbench.swt/0.18.0.v20260317-1600). Expected to have bigger x.y.z than what is available in baseline (0.18.0.v20260119-0835) -> [Help 1]
```
There is no reason to bump that version for not really changed code.

SougandhS:
> @SougandhS : would it be OK if I would squash everything to one commit (except bundle version updates)?

Please go ahead 👍

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkspaceLock.java
```diff
@@ -0,0 +1,142 @@
...
+
+	public static final String USER = "user"; //$NON-NLS-1$
+
+	/**
+	 * Extract the lock details of the selected workspace if it is locked by another
+	 * Eclipse application
+	 *
+	 * @param workspaceUrl the <code>URL</code> of selected workspace
+	 * @return <code>String</code> details of lock owned workspace,
+	 *         <code>null or Empty</code> if not locked
```
iloveeclipse: `<code>null or Empty</code> if not locked` is not clear what is the difference. We should better handle the case where the lock is there or it is there but not readable.

I would also take over original javadoc from `IDEApplication` which better explained what we do here.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkspaceLock.java
```diff
@@ -0,0 +1,142 @@
...
+	 */
+	public static String getWorkspaceLockDetails(URL workspaceUrl) {
+		Path lockFile = getLockInfoFile(workspaceUrl);
+		if (lockFile != null && Files.exists(lockFile)) {
```
iloveeclipse: I would invert the if condition to have cllearer code returning null if there is no file.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkspaceLock.java
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * <p>
+ * To support both environments, this class does not rely on workbench-specific
+ * APIs such as {@code PlatformUI.getWorkbench()} or {@code IWorkbenchWindow},
+ * nor on any API that requires an initialized workbench window. Only SWT-level
+ * constructs (for example, {@link org.eclipse.swt.widgets.Display} and
+ * {@link org.eclipse.swt.widgets.Shell}) and core/runtime APIs are used.
+ * </p>
+ *
+ * @since 3.5
```
iloveeclipse: since version is wrong, but `since` is not needed anyway here

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2709,6 +2712,16 @@ public static Object setRestartArguments(String workspacePath) {
...
 		}
+		Path selectedWorkspace = Path.of(workspacePath);
+		try {
+			String workspaceLock = WorkspaceLock.getWorkspaceLockDetails(selectedWorkspace.toUri().toURL());
+			if (workspaceLock != null) {
+				WorkspaceLock.showWorkspaceLockedDialog(null, workspacePath, workspaceLock);
+				return null;
+			}
+		} catch (MalformedURLException e) {
+			return null;
```
iloveeclipse: this return null here is wrong, because we simply can't get workspace URL, but have no idea if workspace is locked or not. Probably exception should be propagated.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2694,7 +2696,8 @@ private static String buildCommandLine(String workspace) {
 	 * as the workspace location.
 	 *
 	 * @param workspacePath the new workspace location
-	 * @return {@link IApplication#EXIT_OK} or {@link IApplication#EXIT_RELAUNCH}
+	 * @return {@link IApplication#EXIT_OK} or {@link IApplication#EXIT_RELAUNCH} or
+	 *         <code>null</code> if workspace is locked
```
iloveeclipse: Unfortunately code doesn't match the javadoc, see retrurn null at line 2708 after be51d14c6f825ce5f072dcbf771449620e789b6a.

This all needs careful rework.

iloveeclipse (changes requested):
I've fixed version constraints needed, refactored WorkspaceLock and started to look into Workbench code changes needed. Unfortunately be51d14c6f825ce5f072dcbf771449620e789b6a changes were not consistent, we have to understand how that can be fixed in context of the required changes here.

I will push updated PR now, but I plan to continue looking into Workbench changes needed.

iloveeclipse:
Arrgh, no explanation whatsever regarding specific change in `Workbench.setRestartArguments(String)`.

Related changes are for https://github.com/eclipse-equinox/p2/issues/166 and all have same commit message copy/pasted 1:1 for every change in 3 different PRs.
- https://github.com/eclipse-equinox/equinox/pull/401
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1307
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1622

The main problem is that `Workbench.setRestartArguments(String)` now starts to return **null** after be51d14c6f825ce5f072dcbf771449620e789b6a, but the caller of that API 
`org.eclipse.ui.internal.ide.application.IDEApplication.checkInstanceLocation(Shell, Map)` never checked return value (because javadoc says it will never be null) and forwarded that to `org.eclipse.ui.internal.ide.application.IDEApplication.start(IApplicationContext)`, where **null** was meant to be OK, let's continue with startup!!! And that is the opposite case of what "null" answer from `Workbench.setRestartArguments(String)` is supposed to say.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
```diff
@@ -2694,21 +2696,41 @@ private static String buildCommandLine(String workspace) {
...
 			return null;
 		}
 
 		String command_line = Workbench.buildCommandLine(workspacePath);
+		// TOODO: this below will never happen, buildCommandLine() never returns null
+		// since be51d14c6f825ce5f072dcbf771449620e789b6a
 		if (command_line == null) {
 			return IApplication.EXIT_OK;
 		}
+		Path selectedWorkspace = Path.of(workspacePath);
```
iloveeclipse: OK, I finally got time to understand & debug this all. This part should be removed from here and moved to `OpenWorkspaceAction`.

The rationale is: we have two consumers.  `OpenWorkspaceAction` (IDE is running) and `IDEApplication` (IDE is not running). Later one has very complex and comprehensive logic/flow where the workspace lock is checked before call to `setRestartArguments()` here. 

So showing a dialog from here as proposed would "break" or overcomplicate the internal logic that needs something like yes/no/retry and not "null" for different error scenario.

Therefore I would move the new code that opens extra dialog to the caller (`OpenWorkspaceAction`).

iloveeclipse:
I believe I have some useful state now.

@SougandhS : could you please check both code & behavior?

I will do some more manual checks for the workspace selection dialog which is shown before IDE starts tomorrow, this area is a mine field with all the corner cases...

SougandhS:
> I believe I have some useful state now.
> 
> @SougandhS : could you please check both code & behavior?
> 
> I will do some more manual checks for the workspace selection dialog which is shown before IDE starts tomorrow, this area is a mine field with all the corner cases...

Looks good from testing

iloveeclipse:
Just small cleanup in org.eclipse.ui.internal.ide.actions.OpenWorkspaceAction.canRestartWithWorkspace(String) and rebase. I believe that is OK to merge now.

iloveeclipse (approved):

SougandhS:
Thank you @iloveeclipse for the additional fixes and cleanups 🚀

iloveeclipse:
> I will do some more manual checks for the workspace selection dialog which is shown before IDE starts tomorrow, this area is a mine field with all the corner cases...

As usually with workspace selection code, it went wrong, but now by a stupid mistake.

I've tested mostly the use cases for workspace selection dialog, and just briefly the switch from IDE, and exact this is now broken. One can't switch to a different workspace from the running IDE now...

The problem is that we check for the *existence* of the `.lock_info` file, which is however not same as the actual `.lock` file, where only checking for the actual **lock** is the definitive answer for the workspace being locked or not. The reason for that is that if process is killed / crashed, both files are kept on disk, but not the file **lock** which is released automatically if process dies.

I will try to fix that ASAP.

iloveeclipse:
> I will try to fix that ASAP.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/3827

PR #3526, opened 2025-11-10
Fix Project Explorer not refreshing source containers after drag-and-drop move

### LINKED ISSUE #3525: Project explorer doesn't refresh correctly after moving files

## Steps to reproduce

- Create two projects "Project A" and "Project B"
- Create a "test" text file in "Project A"
- Move "test"  to "Project B" via drag & drop

Image

* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): 2025-12

## Community

laeubi:
Fixes eclipse-platform#3525

FYI @ptziegler does this fix the issue you see?

### DIFF
--- a/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/workbench/ResourceExtensionContentProvider.java
+++ b/bundles/org.eclipse.ui.navigator.resources/src/org/eclipse/ui/internal/navigator/resources/workbench/ResourceExtensionContentProvider.java
@@ -24,15 +24,11 @@ import java.util.TreeSet;
 import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceDelta;
-import org.eclipse.core.runtime.CoreException;
-import org.eclipse.core.runtime.IStatus;
-import org.eclipse.core.runtime.Status;
 import org.eclipse.jface.viewers.AbstractTreeViewer;
 import org.eclipse.jface.viewers.StructuredViewer;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.ui.internal.navigator.resources.nested.PathComparator;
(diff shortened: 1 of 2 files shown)


ptziegler:
This makes things even worse now:

image

I also don't think refreshing the UI is the correct solution here. The content provider is simply inconsistent.

ptziegler:
As an additional information: This entry shows up in the log while moving the file

> !ENTRY org.eclipse.ui.navigator 4 0 2025-11-10 12:40:20.423
> !MESSAGE Cannot find navigator content extension (using triggerPoints) for object "\<null\>" parent path: "P/Project". Check that you have an expression for this object in a navigatorContent triggerPoints or enablement.

laeubi:
Hm... so maybe like this?

ptziegler:
> Hm... so maybe like this?

Still the same issue. I did some more debugging and came across this method here. Which is called twice (on the same viewer instance), once by the `JavaNavigatorContentProvider` and once by the `ResourceExtensionContentProvider`, both times with the same elements.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/39d028ea57b25aec286fd8c7c3dda24d0677303b/bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonViewer.java#L334-L345

The second time, this element obviously no longer exists. So instead a "null" tree item is inserted.

If I use this hack here, the viewer is updated correctly again:

```java
@Override
public void remove(Object... elements) {
	var e = new Exception();
	var st = e.getStackTrace();
	if (st[2].toString().contains("org.eclipse.jdt.")) { //$NON-NLS-1$
		return;
	}

	NavigatorPipelineService pipeDream = (NavigatorPipelineService) contentService
			.getPipelineService();

	PipelinedShapeModification modification = new PipelinedShapeModification(
			null, new ContributorTrackingSet(contentService, elements));

	pipeDream.interceptRemove(modification);

	super.remove(modification.getChildren().toArray());
}
```

laeubi:
So it seems we maybe need some kind of asyncExec here...

ptziegler:
@laeubi I think this one is obsolete now?

PR #3527, opened 2025-11-10
Skip dummy item creation when removing element from viewer twice #3525

### LINKED ISSUE #3525: Project explorer doesn't refresh correctly after moving files

ptziegler:
Due to the fix for Bug 210747, a dummy tree-item is added to the viewer if the same element is removed twice. The intend of the original fix is to remove the "expand" icon from a tree item, when its still collapsed children are removed from the viewer (but not the model).

The problem with this approach is that it doesn't take the situation into consideration, where the children of the item have already been removed.

Given a tree item with at least one child in the data model but where all child-widgets have been removed in the viewer. When removing the children again, this internally calls `updatePlus(Item,Object)`.

But because the number of child-widgets doesn't match the number of children, this method now creates a dummy item, to correct this inconsistency. This problem doesn't occur when calling `internalRemove(Object,Object[])`, because here the call to
`updatePlus(Item,Object)` is guarded by a check to make sure the tree item has at least one child-widget.

To reproduce:

1) Make sure the ".* resources" filter is enabled
2) Create a new project "Project A" and "Project B"
3) Create a new text file "test" in "Project A"
4) Move "test" to "Project B"

Closes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3525

### DIFF
--- a/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface;singleton:=true
-Bundle-Version: 3.38.100.qualifier
+Bundle-Version: 3.38.200.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.jface,
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -2095,12 +2095,7 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 				Object parent = getParentElement(element);
 				if (parent != null && !equals(parent, getRoot())
 						&& !(parent instanceof TreePath tp && tp.getSegmentCount() == 0)) {
-					Widget[] parentItems = internalFindItems(parent);
-					for (Widget parentItem : parentItems) {
-						if (parentItem instanceof Item it) {
-							updatePlus(it, parent);
-						}
-					}
+					internalRemove(parent, new Object[] { element });
 				}
 			}
 		}
--- a/tests/org.eclipse.jface.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.jface.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.jface.tests
-Bundle-Version: 1.4.1100.qualifier
+Bundle-Version: 1.4.1200.qualifier
 Automatic-Module-Name: org.eclipse.jface.tests
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Require-Bundle: org.junit;bundle-version="4.12.0",
--- a/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
+++ b/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/viewers/TreeViewerTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2006 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -14,7 +14,10 @@
 package org.eclipse.jface.tests.viewers;
 
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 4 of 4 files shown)


ptziegler:
@laeubi This fixes the issue with the project explorer. Though I'd still like to add a proper test case.

laeubi:
@ptziegler good finding! I can guess its hard to come of with a simple testcase here given JDT is involved and we likely do not want to include it as a dependency.

iloveeclipse:
Considering M3 is this week, I would prefer to postpone this change to 4.39 (assuming it works() in general and not only for specific use case). 

This change (which affects org.eclipse.jface.viewers.AbstractTreeViewer.remove(Object...)` behavior must be tested for a longer time before release.

ptziegler:
Aaah, now everything makes sense. The dummy tree item is created in `updatePlus(...)`, in response to Bug [210747]

image

>  I can guess its hard to come of with a simple testcase here given JDT is involved and we likely do not want to include it as a dependency.

The `remove()` method might be called from anywhere. It just happens to be JDT for the Project Explorer. The difficult part is finding a good test case that can be adapted. 😄

ptziegler:
I had a closer look at Bug [210747](https://bugs.eclipse.org/bugs/show_bug.cgi?id=210747) and whatever was done with f1899ad2abb3cb64959103464c21227d59b4f994 doesn't work, regardless of my change.

> Steps To Reproduce:
> 1. The content provider returns three elements (a), (b) and (c), while (b) is a child of (a) and (c) is a child of (b)
> 2. programatically expand item (a)
> observe: item (b) is rendered with a plus sign
> 
> 3. Call AbstractTreeViewer.remove() to remove item (c)
> observe: the plus sign is not removed

As it is, `item (b)` continues to have a dummy tree-item as its child and would therefore show a plus sign. This is because the element is only removed from the UI and the call to `isExpandable(...)` queries the content provider (i.e. the model). So `needsPlus` returns `true`, even if its only child has just been removed.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/39d028ea57b25aec286fd8c7c3dda24d0677303b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java#L3125-L3127

What a mess...

ptziegler:
I've adapted the check so that for the use-case described in Bug 210747, where an element was removed from the tree that hasn't been realized yet, the `updatePlus()` method is still called. In my use-case, where an element is removed twice, nothing happens the second time.

I'm still not sure what the reasoning behind the initial change was, but I'm fairly confident that what was described in the bug was never fixed and probably can't be reasonbly fixed.

I've added a test case for this bug, which will obviously fail. Perhaps someone who is more familiar with the TreeViewer can have a look, otherwise I'll remove it again for the next M1.

> Considering M3 is this week, I would prefer to postpone this change to 4.39 (assuming it works() in general and not only for specific use case).
> 
> This change (which affects org.eclipse.jface.viewers.AbstractTreeViewer.remove(Object...)` behavior must be tested for a longer time before release.

That's fine by me. This problem has existed for almost two decades, so three more months won't make much of a difference.

ptziegler:
I'm a little bit concerned about this code block here, which is also responsible for the test failure and which was added with #810

https://github.com/eclipse-platform/eclipse.platform.ui/blob/601b0d3006dab3d8612defae3b97d3bff53debe2/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java#L2065-L2084

In short, if `setDisplayIncrementally()` is called, you are no longer able to remove elements from the viewer without first removing them from the model.

ptziegler:
If there are no objections (and if the remaining checks succeed), I'll plan on merging this within the next few days.

iloveeclipse:
@ptziegler : not sure if the commit title & message, PR title & message are reflecting the actual fix?
It would be good to update everything to the latest change state.

iloveeclipse:
> To reproduce:
> 
> * Create a new project "Project A" and "Project B"
> * Create a new text file "test" in "Project A"
> * Move "test" to "Project B"

I can't reproduce on master, tried with Package and Project Explorers. Anything missing, like projects should be expanded / collapsed etc? Or is this only Windows specific (I'm on RHEL Linux/X11).

iloveeclipse:
> I can't reproduce on master, tried with Package and Project Explorers. Anything missing, like projects should be expanded / collapsed etc?

OK, following is needed:
1) Create a new project "X" and "Y".
2) Make sure in view preferences there is no other files or directories (like .settings, .project etc) are shown, so both projects are shown as "empty".
3) Create a new text file "hello.txt" in "X". It should be the only visible file in the project.
4) Move "hello.txt" to "Y".
5) kaboom: the Package Explorer doesn't show "hello.txt", and Project Explorer shows strange empty node
image

iloveeclipse:
This patch doesn't fix for me the problem with the [steps above](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3527#issuecomment-3595577149).

ptziegler:
> Make sure in view preferences there is no other files or directories (like .settings, .project etc) are shown, so both projects are shown as "empty".

Yes, the resource filter for files starting with a '.' needs to be enabled. Apologies for the inaccuracy.

> This patch doesn't fix for me the problem with the https://github.com/eclipse-platform/eclipse.platform.ui/pull/3527#issuecomment-3595577149.

On Linux, this block is entered, which then causes the creation of an additional tree item. But not on Windows. Commenting this block out makes the viewer behave as expected, but that shouldn't be the solution.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/8b402ec5b05563d5222e30acba8b8cb4666505ac/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java#L279-L299

The call to `getExpanded()` is simply delegated to the `TreeItem` widget. This looks like in inconsistency in SWT? Both on Linux and Windows, this method should return `true`.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/8b402ec5b05563d5222e30acba8b8cb4666505ac/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/TreeViewer.java#L176-L179

iloveeclipse:
> This looks like in inconsistency in SWT?

So it looks like the patch works on Windows only? Have you Mac access, any idea how it looks there?

Could you please investigate that Windows/Linux inconsistency before merging the patch?

ptziegler:
The `TreeItem` has a local variable that keeps track of the "expanded" state. However, this variable is not used inside `getExpanded()`, which instead calls the GTK API directly. The result is two conflicting values: `true` from the local variable and `false` from GTK. This doesn't look right...

image

ptziegler:
> So it looks like the patch works on Windows only?

That seems to be the cases, yes.

> Have you Mac access, any idea how it looks there?

No, only Windows and Linux.

> Could you please investigate that Windows/Linux inconsistency before merging the patch?

That's my current plan. If this really is an SWT issue, it should be possible to derive a small snippet with the same, inconsistent "expanded" state.

iloveeclipse:
> @ptziegler : not sure if the commit title & message, PR title & message are reflecting the actual fix? It would be good to update everything to the latest change state.

Can you please update commit message & rebase?

ptziegler:
@iloveeclipse I've updated the commit/pr message and elaborated a little bit on what exactly is going wrong and why.

iloveeclipse (approved):

iloveeclipse:
Thanks @ptziegler. Sorry for delay, as usually totally overloaded...

ptziegler:
No worries. Thanks for also taking a look at this change. Otherwise I completely would've missed the weird behavior on Linux.

PR #3531, opened 2025-11-11
Fix flaky DynamicTest.testDynamicRegistry race condition

### LINKED ISSUE #2458: DynamicTest.testDynamicRegistry fails randomly

vogella:
## Summary

Fixes the random failures in `DynamicTest.testDynamicRegistry` by replacing the plain boolean array with `AtomicBoolean` instances for proper thread-safe visibility across listener callbacks.

## Root Cause

The test was failing randomly with "Activity Listener not called" or "Category Listener not called" errors because:
- The test uses a plain `boolean[]` to track listener invocations
- Extension registry listeners may be invoked on different threads
- Without proper synchronization (volatile/atomic), changes to the array made by listener threads may not be visible to the test thread
- This causes `DisplayHelper.waitForCondition()` to timeout even when listeners were actually called

## Changes

- Replace `boolean[] registryChanged` with two `AtomicBoolean` instances (`activityChanged` and `categoryChanged`)
- Update listeners to use `.set(true)` instead of array assignment
- Update condition checks and assertions to use `.get()` instead of array access
- Add import for `java.util.concurrent.atomic.AtomicBoolean`

## Test Plan

- Ran the test 5 times consecutively - all passed ✅
- The fix ensures proper memory barriers between listener threads and test thread

## Related Issues

Fixes #2458

### DIFF
--- a/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/HoverTest.java
+++ b/tests/org.eclipse.ui.genericeditor.tests/src/org/eclipse/ui/genericeditor/tests/HoverTest.java
@@ -33,6 +33,7 @@ import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.StyledText;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.Link;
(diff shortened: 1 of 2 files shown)


vogella:
@fedejeanne WDYT?

fedejeanne:
Sorry for the late reply. The changes look OK. Based on the description of the issue, this additional wait shouldn't be necessary (or at least, I don't see _why_ it is necessary in the description):

```java
DisplayHelper.waitForCondition(display, 3000, () -> firstShell.isDisposed());
```

But since that operation doesn't slow down the test if the shell is disposed, I'd say it's safe to use 👍 

Thank you for fixing the flaky tests, @vogella ! 🚀 👍

PR #3533, opened 2025-11-11
Adjust selection bounds for CTabRenderring

### LINKED ISSUE #517: CTabFolder highlight not positioned/sized correctly on zoom != 100%
**Description**
The highlight line of the active/selected CTabFolder item is not positioned correctly when the monitor zoom is not 100%.
For example, on 150% it looks like this (with and without theming):
Image
Image

You see the white area left to above the line, which should not be there.

**Reproduction**
To reproduce, simply use a 150% monitor, start an SDK and look at the editor tabs.

**Expected Behavior**
The highlight line shall be properly drawn for every zoom. This applies to both the Platform UI's enhanced `CTabRendering` as well as when only using SWT's `CTabFolderRenderer` (when theming is disabled). So the behavior should be checked with both renderers, i.e., with and without theming.
In particular, it shall be ensured that tab icons do not touch the highlight line, such as icon for a Manifest editor tab does when theming is disabled:
Image

**Necessary configuration:**
150% monitor (or higher)

ShahzaibIbrahim:
Selected tabs highlight look slightly off and more noticeable when zoom is not 100%. These adjusted value shows no gaps from top and better aligned highlights for tabs (Theme is enabled)

### Results:
image

image

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -18,6 +18,7 @@
 package org.eclipse.e4.ui.workbench.renderers.swt;
 
 import java.lang.reflect.Field;
+import java.util.Arrays;
 import java.util.Objects;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
@@ -621,17 +622,47 @@ public class CTabRendering extends CTabFolderRenderer implements ICTabRendering,
 		}
 
 		if (selectedTabHighlightColor != null) {
+			gc.setForeground(selectedTabHighlightColor);
 			gc.setBackground(selectedTabHighlightColor);
+			int originalLineWidth = gc.getLineWidth();
+			int highlightHeight = 2;
 			boolean highlightOnTop = drawTabHighlightOnTop;
 			if (onBottom) {
 				highlightOnTop = !highlightOnTop;
 			}
-			int highlightHeight = 2;
-			int verticalOffset = highlightOnTop ? 0 : bounds.height - (highlightHeight - 1);
-			int horizontalOffset = itemIndex == 0 || cornerSize == SQUARE_CORNER ? 0 : 1;
-			int widthAdjustment = cornerSize == SQUARE_CORNER ? 0 : 1;
-			gc.fillRectangle(bounds.x + horizontalOffset, bounds.y + verticalOffset, bounds.width - widthAdjustment,
-					highlightHeight);
+
+			if (highlightOnTop != onBottom && cornerSize != SQUARE_CORNER) {
+				// When using round tabs, extract the part of the outline that covers the
+				// highlight and draw it with adapted height
+				int[] highlightShape = Arrays.copyOfRange(tabOutlinePoints, 14, tabOutlinePoints.length - 14);
+				int yEnd = !highlightOnTop
+						? outlineBoundsForOutline.y + outlineBoundsForOutline.height - highlightHeight + 1
+						: highlightHeight;
+				highlightShape[1] = highlightShape[highlightShape.length - 1] = yEnd;
+				boolean gcAdvanced = gc.getAdvanced();
+				gc.setAdvanced(false);
+				gc.fillPolygon(highlightShape);
+				gc.setAdvanced(gcAdvanced);
+			} else {
+				// When using square tab or drawing the highlight at the bottom, simply draw a
+				// line
+				int highlightY = highlightOnTop ? outlineBoundsForOutline.y
+						: bounds.y + bounds.height - highlightHeight + (!onBottom ? 1 : 0);
+				// When the rectangle is drawn at the bottom, the outline may not fully cover
+				// the left pixel such that we should start one point to the right (even though
+				// we produce a slight gap then)
+				int xOffset = highlightOnTop ? 0 : 1;
+				gc.fillRectangle(outlineBoundsForOutline.x + xOffset, highlightY,
+						outlineBoundsForOutline.width - xOffset, highlightHeight);
+				if (highlightOnTop && !onBottom) {
+					// Compensate for the outline being draw on top of the filled region by
+					// extending the highlight with an equally wide line next to the filled region
+					gc.setLineWidth(1);
+					gc.drawLine(outlineBoundsForOutline.x, highlightY + highlightHeight,
+							outlineBoundsForOutline.x + outlineBoundsForOutline.width, highlightY + highlightHeight);
+				}
+			}
+			gc.setLineWidth(originalLineWidth);
 		}
 
 		if (backgroundPattern != null) {


bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -630,7 +630,8 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
 			int verticalOffset = highlightOnTop ? 0 : bounds.height - (highlightHeight - 1);
 			int horizontalOffset = itemIndex == 0 || cornerSize == SQUARE_CORNER ? 0 : 1;
 			int widthAdjustment = cornerSize == SQUARE_CORNER ? 0 : 1;
-			gc.fillRectangle(bounds.x + horizontalOffset, bounds.y + verticalOffset, bounds.width - widthAdjustment,
+			gc.fillRectangle(bounds.x + horizontalOffset - 1, bounds.y + verticalOffset - 1,
```
HeikoKlare: Why is it necessary/correct to remove 1 from the vertical offset? When drawing the line on top, this will place the line at `bounds.y - 1`, which sounds odd. In addition, when the line is not drawn on top, the `verticalOffset` calculation adds `1`, which is now substracted again. Is that correct? Should that new offset only be applied when `hightlightOnTop`? And shouldn't it then be included in the offset calculation rather than the call to `fillRectangle` for the sake of comprehensibility?
Same applies to the horizontal offset, which takes `SQUARE_CORNER` into account. Is that properly reflected in the adapted calculation? If yes, how to best test this?

HeikoKlare (commented):
With this change, the highlight line looks very thin (on 100%):
image

This is how it used to look:
image

ShahzaibIbrahim:
@HeikoKlare I found the problem. Here were drawing the tab from (0,0) pixel but still using initial item location (1,1) and hence the gap was visible. I also updated the height to 3 when we draw from (0,0) so we have enough height for the selection marker.

The code that caused the gap: 
https://github.com/vi-eclipse/eclipse.platform.ui/blob/8be04038fd7a1dd5cf4f7972e7813d0925181315/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java#L572C2-L577

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -626,11 +626,11 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
 			if (onBottom) {
 				highlightOnTop = !highlightOnTop;
 			}
-			int highlightHeight = 2;
-			int verticalOffset = highlightOnTop ? 0 : bounds.height - (highlightHeight - 1);
+			int highlightHeight = superimposeKeylineOutline ? 3 : 2;
```
HeikoKlare: Shouldn't this rather be `2 + superimposeKeylineOutine ? OUTER_KEYLINE_WIDTH : 0`?

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -626,11 +626,11 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
...
+			int verticalOffset = highlightOnTop ? 0 : outlineBoundsForOutline.height - (highlightHeight - 1);
```
HeikoKlare: With `superImposeKeyline == true` this will now draw a line at the bottom that is one pixel higher than expected, won't it?

HeikoKlare (commented):
The appearance is in my opinion much better than before. I have some questions/remarks on the current propose.
And the appearance with round tabs now became worse (than it already was before):
image

We may probably do something similar than in SWT to draw a proper shape. The current adjustments by 1 point based on `cornerSize` do not look sufficient.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
ShahzaibIbrahim: I have fixed the condition to have an extra pixel when highlight is on top only. 

`int highlightHeight = 2 + (superimposeKeylineOutline && highlightOnTop ? OUTER_KEYLINE_WIDTH : 0);`

ShahzaibIbrahim:
> The appearance is in my opinion much better than before. I have some questions/remarks on the current propose. And the appearance with round tabs now became worse (than it already was before): image
> 
> We may probably do something similar than in SWT to draw a proper shape. The current adjustments by 1 point based on `cornerSize` do not look sufficient.

While I agree the rounded tabs should be adapted. But I wouldn't say it looks "worse" than master. Here's the SS comparing this PR and master state. 

image

I would suggest that we fix the shape in another PR. This PR fixes the gap that we had from the top.

amartya4256:
I pulled the branch and ran it on 250%. This is how it looks for me:
image

I can see a small gap in the left and at the top. Is it supposed to look like this?

HeikoKlare:
> I can see a small gap in the left and at the top. Is it supposed to look like this?

No, it should be fully aligned at the top, left and right without any gaps at every zoom.

ShahzaibIbrahim:
> I pulled the branch and ran it on 250%. This is how it looks for me: image
> 
> I can see a small gap in the left and at the top. Is it supposed to look like this?

Are you sure you checked out master-517 from platform.ui? We also have master-517 for SWT. I checked again and there is no gap for the highlight. 

image

HeikoKlare:
@amartya4256 can you please retest this? Unfortunately, I have currently monitor available that I can set to 250%. Everything up to 225% looks fine to me.

HeikoKlare:
> I would suggest that we fix the shape in another PR. This PR fixes the gap that we had from the top.

@ShahzaibIbrahim But maybe that would require another readaptation of the calculations. Shouldn't we just fix it right away? Then it would also make more sense to take into account the `outlineBoundsForOutline` as those are used by the calculations for the (potentially rounded) borders.

ShahzaibIbrahim:
> > I would suggest that we fix the shape in another PR. This PR fixes the gap that we had from the top.
> 
> @ShahzaibIbrahim But maybe that would require another readaptation of the calculations. Shouldn't we just fix it right away? Then it would also make more sense to take into account the `outlineBoundsForOutline` as those are used by the calculations for the (potentially rounded) borders.

I have added a commit to put highlight on rounded tabs. Although it looks big jagged as original one (non-selected rounded tabs) also bit jagged. I have reused the shape. 

Here's the comparison of the selected and unselected tabs: 

image

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -621,17 +622,29 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
...
-			int verticalOffset = highlightOnTop ? 0 : bounds.height - (highlightHeight - 1);
-			int horizontalOffset = itemIndex == 0 || cornerSize == SQUARE_CORNER ? 0 : 1;
-			int widthAdjustment = cornerSize == SQUARE_CORNER ? 0 : 1;
-			gc.fillRectangle(bounds.x + horizontalOffset, bounds.y + verticalOffset, bounds.width - widthAdjustment,
-					highlightHeight);
+			if (cornerSize == SQUARE_CORNER) {
+				int highlightHeight = 2 + (superimposeKeylineOutline && highlightOnTop ? OUTER_KEYLINE_WIDTH : 0);
+				int verticalOffset = highlightOnTop ? 0 : outlineBoundsForOutline.height - (highlightHeight - 1);
+				int horizontalOffset = itemIndex == 0 || cornerSize == SQUARE_CORNER ? 0 : 1;
+				int widthAdjustment = cornerSize == SQUARE_CORNER ? 0 : 1;
```
HeikoKlare: Since `cornerSize == SQUARE_CORNER` in this branch, these variable can be completely removed, can't they?

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -621,17 +622,29 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
...
+				gc.fillRectangle(outlineBoundsForOutline.x + horizontalOffset,
+						outlineBoundsForOutline.y + verticalOffset, outlineBoundsForOutline.width - widthAdjustment,
+						highlightHeight);
+			} else {
+				int[] highlightShape = Arrays.copyOf(tabOutlinePoints, tabOutlinePoints.length);
+				// Update Y coordinates in shape to apply highlight thickness
+				int thickness = 2;
```
HeikoKlare: Why don't we need to apply the more complex calculation logic used for the `SQUARE_CORNER` case here? This seems to lead to hightlights that are far too thick:
image
image
Also note that there are blue-colored fragments remaining to the left and right. Seems like some more tweaking of the shape points is necessary.

Screenshots are taken at 175%.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -621,17 +622,29 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
...
+				int highlightY = onBottom ? bottomY - thickness : thickness + 1;
```
HeikoKlare: This introduces a regression, as it uses `onBottom` instead of the `highlightOnTop`. I think there is even a mistake in the general distinction just based on `cornerSize == SQUARE_CORNER`. You will also need to draw a rectangle in case of round corners but when the highlight is not at the end with the rounding, i.e., if `highlightOnTop != onBottom`. So the condition for drawing a rectangle probably needs to be `cornerSize == SQUARE_CORNER || highlightOnTop == onBottom`.
image
image

HeikoKlare (dismissed):

ShahzaibIbrahim:
After latest adaptation, there is no more "Bleeding" on the very left and right. The selection highlight looks fine to me in different zoom levels. Also addressed the comments from you @HeikoKlare. Let me know if I miss something. 

image

HeikoKlare (commented):
We seem to be coming closer. The code became much simpler and many scenarios now look good.
But with rounded corner, the highlight is still far too large. At 150%, it looks like this (compare to the "correct" hight for the non-focused "problem" view below):
image

When reducing the hight by 1, I see the rendering artifacts mentioned before again.
The screenshots in the previous post seem to be taken with some custom auto-scaling settings. The tab icons are not scaled according to the text, so probably monitor-specific scaling is disabled or some custom auto-scale configuration is active.

ShahzaibIbrahim:
> We seem to be coming closer. The code became much simpler and many scenarios now look good. But with rounded corner, the highlight is still far too large. At 150%, it looks like this (compare to the "correct" hight for the non-focused "problem" view below): 

I reduced the height by 1 and fixed the bleeding and this is how it looks now. For me its too small now. But if it feels okay to you I can push the changes. 

image

ShahzaibIbrahim:
As discussed earlier, the highlight height for the curved and non-curved tab folders should be same. I have adjusted the height now. This is the final result. 

image

HeikoKlare (commented):
I looks almost fine now. But the lines still do not have the same height, such that the highlight bar with rounded corners appear a bit too thick and prominent. These are the heights I get:
- 100%: round corners 3px, square corners 2px
- 175%: round corners 5px, square corners 4px

Here is the comparison of the highlight at 100% with round corners before (top) and after (bottom) this change:
image

The overall appearance is definitely much better, but the line is thicker now.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -621,17 +622,27 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
 		}
 
 		if (selectedTabHighlightColor != null) {
+			Color oldBackground = gc.getBackground();
```
HeikoKlare: Why do we (re-)store the background color. This was not necessary before and I don't see that this changed.

ShahzaibIbrahim:
> I looks almost fine now. But the lines still do not have the same height, such that the highlight bar with rounded corners appear a bit too thick and prominent. These are the heights I get:
> 
> * 100%: round corners 3px, square corners 2px
> * 175%: round corners 5px, square corners 4px


The final fix meant we have consistent size for highlighting in CTabFolders for both square and rounded tabs. See. 

image

akoch-yatta:
I still see a difference in upper and lower height of the highlight in 150%
image

ShahzaibIbrahim:
> I still see a difference in upper and lower height of the highlight in 150% image

I tried with power toys and the lower highlight and upper highlight both are 3px each.

akoch-yatta:
Just measured it in my Runtime, the lower one is 3px, the upper one is 4px

Doesn't that make sense as `final int highlightHeight = 1 + (superimposeKeylineOutline && highlightOnTop ? OUTER_KEYLINE_WIDTH : 0);` does draw the rectangle with a different height because of that?

HeikoKlare:
> Just measured it in my Runtime, the lower one is 3px, the upper one is 4px

I can confirm this observation.

arunjose696:
I have tested with the new changes, I can observe the lower one and upper one are now both 3px, but I observe a different issue.  The boarder that passes through the bottom highlight are now missing

This change
image

Master
image

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -626,12 +627,24 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
...
-			int widthAdjustment = cornerSize == SQUARE_CORNER ? 0 : 1;
-			gc.fillRectangle(bounds.x + horizontalOffset, bounds.y + verticalOffset, bounds.width - widthAdjustment,
-					highlightHeight);
+			final int highlightHeight = 2;
+			if (cornerSize == SQUARE_CORNER || highlightOnTop == onBottom) {
+				int verticalOffset = highlightOnTop ? 0 : outlineBoundsForOutline.height - (highlightHeight - 1);
+				gc.fillRectangle(outlineBoundsForOutline.x,
+						outlineBoundsForOutline.y + verticalOffset, outlineBoundsForOutline.width, highlightHeight);
+			} else {
+				int[] highlightShape = Arrays.copyOfRange(tabOutlinePoints, 12, tabOutlinePoints.length - 12);
```
arunjose696: I was not able to follow what was 12, I think it would be better if it could be given a variable name for readablity,

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -626,12 +627,24 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
...
+				int highlightY = highlightOnTop ? highlightHeight
+						: outlineBoundsForOutline.height - highlightHeight;
+				highlightShape[1] = highlightShape[3] = highlightShape[highlightShape.length
+						- 1] = highlightShape[highlightShape.length - 3] = highlightY;
+				gc.setAdvanced(false);
```
arunjose696, line 643: why do we unset the advanced mode only in this branch, also does it make sense to reset the advanced mode to the previous value?

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
ShahzaibIbrahim: I think it's self-explanatory. We are taking part of tabOutlinePoints to create a highlight shape. I would recommend to keep it like this.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
ShahzaibIbrahim: When passing highlight shape to fillPolygon, the advance mode output differs to what we want to draw here (exact Y pixels). Resetting to previous state is preserved now.

ShahzaibIbrahim:
> I have tested with the new changes, I can observe the lower one and upper one are now both 3px, but I observe a different issue. The boarder that passes through the bottom highlight are now missing

It's resolved now.

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -620,18 +621,33 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
...
 				highlightOnTop = !highlightOnTop;
 			}
-			int highlightHeight = 2;
-			int verticalOffset = highlightOnTop ? 0 : bounds.height - (highlightHeight - 1);
-			int horizontalOffset = itemIndex == 0 || cornerSize == SQUARE_CORNER ? 0 : 1;
-			int widthAdjustment = cornerSize == SQUARE_CORNER ? 0 : 1;
-			gc.fillRectangle(bounds.x + horizontalOffset, bounds.y + verticalOffset, bounds.width - widthAdjustment,
-					highlightHeight);
+			final int highlightHeight = 2;
+			if (cornerSize == SQUARE_CORNER || highlightOnTop == onBottom) {
```
arunjose696: @ShahzaibIbrahim I’m just slightly confused about the second part of the if condition (highlightOnTop == onBottom). I’m curious if there’s a scenario you’ve encountered where both highlightOnTop and onBottom are true. It would be interesting to test if this is something commonly reproducible in the IDE. I could see this complete flow is not triggered when running CustomControlExample so could not use it for reproduction

Apart from that, I’ve tested the changes at 100%, 150%, 200%, and 350% zoom levels, and the updates look good for both square and rounded tabs. The appearance is noticeably improved compared to the current master.

arunjose696 (approved):
Approving this as the changes look fine on all tested zooms

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
```diff
@@ -620,18 +621,33 @@ void drawSelectedTab(int itemIndex, GC gc, Rectangle bounds) {
 			gc.drawLine(startX, 0, endX, 0);
 		}
 
+		if (TAB_OUTLINE_WIDTH > 0) {
+			gc.drawPolyline(tabOutlinePoints);
+		}
```
HeikoKlare: I was wondering why the outline is now drawn first instead of after painting the highlight.
When taking a look at how it looks, it turns out that this PR now changes the placement of the highlight, such that it overwrites the outline:
image

While you do not notice that much with the default gray outline, this would look really odd if you have a colored outline. Since the `gc.setForeground(tabOutlineColor)` call has been accidentally removed, you will not notice as the color is simply missing.

Please move this call back or ensure by any other means that the outline remains properly drawn.

HeikoKlare (dismissed):
This PR currently breaks some existing behavior (also see the detailed comment).

The highlights are misplaced (also see the detailed comment). This can directly be seen at the highlights displayed at the bottom, which are too small now:
image

This is how they used to look like:
image

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
ShahzaibIbrahim: I have reverted the outline to its original position. However as discussed to accommodate outline we had to increase the highlight height for squared tabs. Here is how it compares to rounded and bottom tabs.

Image

This is the expected highlight size chart for different zooms

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/SHAHZA~1.IBR/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/SHAHZA~1.IBR/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#467886" vlink="#96607D">


Highlight   Tab | Formula | In Points | 100 | 150 | 175
-- | -- | -- | -- | -- | --
Square Top | Round(pts * (zoom / 100) - 1(outline)) | 3pt | 3 | 4 | 4
Bottom | Round(pts * (zoom / 100)) | 2pt | 2 | 3 | 4
Rounded | Round(pts * (zoom / 100)) | 2pt | 2 | 3 | 4



</body>

</html>

HeikoKlare (commented):
Almost everything looks fine now. But the highlight bar is misplaced when shown at the bottom and using round tabs (e.g., at 150%):
image

ShahzaibIbrahim:
> Almost everything looks fine now. But the highlight bar is misplaced when shown at the bottom and using round tabs (e.g., at 150%): image

You're right. The bottom selection highlight shouldn't care if they are rounded or squared tabs. I have fixed it. 

image

HeikoKlare (commented):
Now with the latest changes it is too small again when placed at the bottom. In this screenshot, the left is the state now and the right is the state with this PR (at 150% zoom):
image

This is also somehow expected as the offset calculation used `bounds.height - (highlightHeight - 1)` before andnow uses `bounds.height - (highlightHeight)`.

May I also ask to test every combination (rounded/square, bottom/top, different zooms) for the next review iteration yourself first?

ShahzaibIbrahim:
> Now with the latest changes it is too small again when placed at the bottom. In this screenshot, the left is the state now and the right is the state with this PR (at 150% zoom): image
> 
> This is also somehow expected as the offset calculation used `bounds.height - (highlightHeight - 1)` before andnow uses `bounds.height - (highlightHeight)`.
> 

If you look the screenshot, the current state highlight is below the outline 
image

And the after change, the size is as per the chart that I mentioned here in the comment (i.e. 3px at 150%): https://github.com/eclipse-platform/eclipse.platform.ui/pull/3533#discussion_r2664465853

HeikoKlare:
Okay, I see the misplacement in the existing behavior. But it's misplaced with this PR as well, just in the opposite direction (it is placed too high instead too low). Note that the zoom I posted (150%) seems to actually have been incorrect for the screenshot. It must have been taken at 125%. 

> And the after change, the size is as per the chart that I mentioned here in the comment (i.e. 3px at 150%): [#3533 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3533#discussion_r2664465853)

So why do we do it like that? While the highlight at the bottom was just misplaced before (at 125%), it is now misplaced and inconsistent to other highlights being show at the top (if not using round tabs, which is rather uncommon). Couldn't we draw an additional line (in addition to filling the rect) to compensate for the rounding error here? Since the outline is drawn at the y coordinate where the highlight is missing at 125%, it might be possible to compensate with that.

ShahzaibIbrahim:
> Okay, I see the misplacement in the existing behavior. But it's misplaced with this PR as well, just in the opposite direction (it is placed too high instead too low). Note that the zoom I posted (150%) seems to actually have been incorrect for the screenshot. It must have been taken at 125%.
> 
> > And the after change, the size is as per the chart that I mentioned here in the comment (i.e. 3px at 150%): [#3533 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3533#discussion_r2664465853)
> 
> So why do we do it like that? While the highlight at the bottom was just misplaced before (at 125%), it is now misplaced and inconsistent to other highlights being show at the top (if not using round tabs, which is rather uncommon). Couldn't we draw an additional line (in addition to filling the rect) to compensate for the rounding error here? Since the outline is drawn at the y coordinate where the highlight is missing at 125%, it might be possible to compensate with that.

I have now drawn a line on very bottom (the filler space), although It will make it look slightly darker (I have updated the values in the chart, see description). Also I have tested on every zoom which to me looks good now. The highlight is right on the line. The latest push doesn't affect top highlights (square or rounded). Attaching the screenshot for bottom highlight for all the zoom: 

image

Bottom Highlight with rounded tabs on top

image

HeikoKlare:
Let's take a step back here, as we seem to be running in circles. In my understanding, the goal of this change should be something like:
- Preserve the perfectly fine behavior at 100%
- Ensure that there are no unintended gaps under/over the highlight line at other zooms and ensure that they have a consistent appearance (consistent height for bottom-/top-placed highlights) at each zoom

With the current state, I see this on 100% (left is existing, right is with this PR), so changed behavior for 100%:
image

In particular, the behavior is also inconsistent, as you can see in the highlight on top being larger than the highlight at the bottom: 
image

ShahzaibIbrahim:
After discussion, I have made changes for this iteration which will make highlight consistent for all kind of tabs (Rounded, Square and bottom). The base size will be 2pts and it will increase as per zoom (2pts * (zoom/100)). The bottom highlight do have a bit of issue of outline at zoom >= 175 but that's how it currently is. So no change for that. Here are the result of each tab highlight in different zooms. 

image

HeikoKlare (commented):
The current proposal seems to work decently fine when tab headers are placed at the top. The only limitation we see is that the highlights that are draw at the bottom end of a tab header (at the side of the tab contents) are a bit off on some zoom.

Unfortunately, the current implementation is fully incompatible with CTabFolders that have their tab headers placed at the bottom. I simply tested by adapting the CTabFolder constructor to always use `onBottom = true` and get this:
image
image

Even though the current state is not that great either (in terms of the highlights not having consistent heights), this PR would currently break the highlighting for such tab folders completely (top/bottom is swapped, placement is wrong and sometimes even the whole header is highlighted).

HeikoKlare:
@ShahzaibIbrahim I have created #3656 to overcome the current limitations (in particular with headers placed at the bottom) of this proposal without messing up this PR. Can you please validate the PR in itself and against this one?

ShahzaibIbrahim:
> @ShahzaibIbrahim I have created #3656 to overcome the current limitations (in particular with headers placed at the bottom) of this proposal without messing up this PR. Can you please validate the PR in itself and against this one?

I have overcome the current problems with onBottom = true, and adapted few changes from your PR. Here are the results.

image

HeikoKlare (commented):
With the latest changes, `onBottom` looks like this:

100%:
image

150%:
image

Both don't look in any way as intended.

Note that the screenshots in the previous post don't show any `onBottom` CTabFolder. They just show folders that are "on the bottom" of a workbench window.

ShahzaibIbrahim:
> With the latest changes, `onBottom` looks like this:
> 
> 100%: image
> 
> 150%: image
> 
> Both don't look in any way as intended.
> 
> Note that the screenshots in the previous post don't show any `onBottom` CTabFolder. They just show folders that are "on the bottom" of a workbench window.

I misunderstood how onBottom works. So this time I have added a commit fixing the onBottom case. Here are the result

image

i have kept the commit separate for now just for you to see what I have done to fix it without affecting the case when onBottom==false. Hopefully now all the cases are covered correctly. The issue initially arose due to gc.drawPolyline cutting through the highlight and you see that gap.

HeikoKlare (commented):
I've tested this again and so far found two issues:

First, in case the highlight is not on top of a header, the highlight looks slightly cut off at the left end on zoom higher than 100%:
image

This is caused by the highlight starting at the same x pixel coordinate than the outline, but since the highlight is drawn one pixel too low, the outline does not cover the highlight at than one pixel below the outline.
Since we found that we cannot easily fix this issue of being one pixel too low, we may just also start a bit more to the right (like done now) via something like:
```java
// When the rectangle is drawn at the bottom, the outline may not fully cover
// the left pixel such that we should start one point to the right (even though
// we produce a slight gap then)
int xOffset = highlightOnTop ? 0 : 1;
gc.fillRectangle(outlineBoundsForOutline.x + xOffset, highlightY,
		outlineBoundsForOutline.width - xOffset, highlightHeight);
```

One more severe observation is that when using `onBottom` the highlight on top "flickers" (i.e., the line become one pixel smaller/larger) on 125% when changing the position of the header by dragging the according part larger and smaller:
highlight_onbottom_flicker

Edit: the latter one is also present in the current `master` state, so even though we do not improve that, I would not make anything worse. So we can probably ignore that for now.

ShahzaibIbrahim:
@HeikoKlare the mentioned issue with xOffset has been resolved. This PR is now ready for merging.

akoch-yatta (approved):
I tested it on Windows with and without rounded corner as well as with onBottom=true and false on the standard zooms in windows. Except the mentioned (pre-existing) issue, the behavior looks consistent over all zooms for me.

PR #3534, opened 2025-11-11
In Project Explorer, ensure "Select All" (Ctrl+A and menu) works reliably irrespective of first or subsequent selections.

### LINKED ISSUE #2739: Edit/"Select All" Menu item disabled with single project selected in Project Explorer

## Steps to reproduce

1. Create two or more projects
2. In Project Explorer, click on a project
3. Click on "Edit" menu: the "Select All" item is disabled (it should be enabled), and CTRL+A has no effect (it should select all projects)
4. In Project Explorer, click on another project (or click again on the same project)
5. Click again on the "Edit" menu: the "Select All" item is enabled, and CTRL+A works as expected (but it should have worked at step 2)

* OS & version: Windows 10
* Eclipse IDE/Platform version (as shown in *Help > About*): Eclipse IDE for Enterprise Java and Web Developers (includes Incubating components)
Version: 2024-12 (4.34.0)
Build id: 20241128-0757

## Community

deepika-u:
Fixes #2739 

Description: 
In the Project Explorer, the "Edit > Select All" menu item and Ctrl+A shortcut did not work reliably when a single project was selected. Specifically, pressing Ctrl+A after selecting a project for the first time had no effect. Only after clicking another project did Ctrl+A begin to work as expected.

Details of the fix:
- Added a KeyListener to the Tree widget to handle Ctrl+A directly, ensuring selection works even without command activation.
- Registered a IPartListener2 to activate the EDIT_SELECT_ALL command handler when the Project Explorer view becomes active, enabling the menu item.
- Deactivated the handler when the view is no longer active to avoid stale references.

Note : No-ops - These methods are required by the IPartListener2 interface though unused but must be implemented. Yeah - i just tried even if removed it works(not sure if it added some warnings).

Open eclipse freshly, open Project Explorer view. Select a project and ctrl+A to select all the projects.
Before fix(unable to select, next click on 2nd project then ctrl+A will be enabled)
image

After fix(able to select)
image

### DIFF
--- a/bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
+++ b/bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
@@ -16,8 +16,6 @@
  *******************************************************************************/
 package org.eclipse.ui.navigator;
 
-import org.eclipse.core.commands.AbstractHandler;
-import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.PerformanceStats;
 import org.eclipse.core.runtime.SafeRunner;
@@ -31,25 +29,18 @@ import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.jface.viewers.ViewerFilter;
 import org.eclipse.swt.SWT;
-import org.eclipse.swt.events.KeyAdapter;
-import org.eclipse.swt.events.KeyEvent;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IMemento;
-import org.eclipse.ui.IPartListener2;
 import org.eclipse.ui.ISaveablePart;
 import org.eclipse.ui.ISaveablesLifecycleListener;
 import org.eclipse.ui.ISaveablesSource;
 import org.eclipse.ui.IViewSite;
-import org.eclipse.ui.IWorkbenchCommandConstants;
-import org.eclipse.ui.IWorkbenchPartReference;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.Saveable;
 import org.eclipse.ui.SaveablesLifecycleEvent;
 import org.eclipse.ui.actions.ActionGroup;
-import org.eclipse.ui.handlers.IHandlerActivation;
-import org.eclipse.ui.handlers.IHandlerService;
 import org.eclipse.ui.internal.navigator.CommonNavigatorActionGroup;
 import org.eclipse.ui.internal.navigator.NavigatorContentService;
 import org.eclipse.ui.internal.navigator.NavigatorPlugin;
@@ -169,8 +160,6 @@ public class CommonNavigator extends ViewPart implements ISetSelectionTarget, IS
 
 	private LinkHelperService linkService;
 
-	private IPartListener2 partListener;
-
 	public CommonNavigator() {
 		super();
 	}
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
```diff
@@ -255,6 +266,72 @@ public void handleLifecycleEvent(SaveablesLifecycleEvent event) {
...
+			public void partDeactivated(IWorkbenchPartReference ref) {
+				if (ref.getPart(false) == CommonNavigator.this && activation != null) {
+					IHandlerService hs = getSite().getService(IHandlerService.class);
+					hs.deactivateHandler(activation);
+					activation = null;
+				}
+			}
+
+			// no-ops for the other methods:
+			@Override
```
merks: All the methods have default implementations so these stubs are just useless noise.

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
```diff
@@ -255,6 +266,72 @@ public void handleLifecycleEvent(SaveablesLifecycleEvent event) {
 			ColumnViewerToolTipSupport.enableFor(commonViewer);
 		}
 
+		// Immediate fallback: handle Ctrl+A at the Tree level
+		commonViewer.getTree().addKeyListener(new KeyAdapter() {
+			@Override
+			public void keyPressed(KeyEvent e) {
```
merks: Somehow it seems like a hack to handle this one key directly, especially given the key binding appears to be configurable:

Image

merks (changes requested):

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
iloveeclipse: Note, CommonViewer is a base class in CNF and can be extended. Adding key and part listeners that consumes key or register handlers here directly might affect others code. So please carefully change the code here, it is not only about Project Explorer.

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
deepika-u: As suggested, removed them.

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
deepika-u: > Somehow it seems like a hack to handle this one key directly, especially given the key binding appears to be configurable:

Let me check on it.

bundles/org.eclipse.ui.navigator/src/org/eclipse/ui/navigator/CommonNavigator.java
```diff
@@ -255,6 +266,47 @@ public void handleLifecycleEvent(SaveablesLifecycleEvent event) {
...
+				// MOD1 = Ctrl on Win/Linux, Command on macOS
+				if ((e.stateMask & SWT.MOD1) != 0 && (e.keyCode == 'a' || e.keyCode == 'A')) {
```
merks: This continues to hard code a key binding that is configurable doesn't it?

merks (commented):
It continues to feel like we are treating a symptom and not fixing an underlying problem.

deepika-u:
@merks 
Sorry for the delayed reply on your comments. 
I have revised the fix as per your suggestion. When you get some time, can you take a look at this please? 
Thanks alot.

deepika-u:
@merks 
Could you take a look when you get some time please?

iloveeclipse:
> I have revised the fix as per your suggestion

Unfortunately cutrent fix doesn't change anything regarding original behavior - the same issue is still there.
I didn't found an explanation of the root cause - is it understood, why such weird behavior exist at all?
I would assume fixing the root cause would fix the problem without adding new code.

Looking at the current behnavior (**two** selection events needed before Ctrl+A starts to work), I guess there is something like (too) lazy init of some selection providers in CNF responsible for the "select all" code.

PR #3538, opened 2025-11-12
Removal of deprecation details due to warning caused in #2268

deepika-u:


akurtakov:
@deepika-u Have you looked into what would cause to not use this method in console? It is what I would rather see be done rather than this push/revert.

deepika-u:
@akurtakov 
Instead of calling the deprecated getVisibleLinesInViewport method, the line count can be computed inline within the caller, as I have access to the fTextWidget object. However, the limitation noted in the deprecation still applies - results may be inaccurate when the widget uses variable line heights. I will continue investigating potential approaches. Do you have any suggestions on how this could be achieved?

akurtakov:
I haven't found the time to look into it at all. It's the general approach how things should happen in this case that I pointed out.

vogella:
Is this change still relevant?

deepika-u:
@vogella 
From my further investigation, there is no replacement yet for getVisibleLinesInViewport() so better remove the deprecation details is a meaningful way to proceed till we have the equivalent ready.

Thanks for checking.

vogella:
I would like to verify the change. Where do I see this deprecation warning? If I open /org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewer.java I do not see a related warning.

deepika-u:
@vogella In this pr itself, we are not seeing the deprecation warning but via #2268 we saw this warning generated as a result of marking getVisibleLinesInViewport for removal(when done in #3495) because it is being referred to by TextConsoleViewer() from platform code though this is in platform.ui.

vogella:
TextConsoleViewer shows the warning in my IDE but it does not matter for it if I the method is only deprecated or marked for deletion. 

I don't feel I'm the right person to merge this. Mayge @akurtakov can help here.

akurtakov:
I would rather see the code using this API being migrated to not use deprecated API than removing forRemoval here. Closing this one as it's the wrong path forward.

PR #3540, opened 2025-11-12
Fix alignment of onboarding area command labels

### LINKED ISSUE #525: Degraded text alignment in onboarding area
**Description**
The texts in the onboarding area are not right-aligned anymore. This is probably not specific to the onboarding area but may be a more general placement issue.

This is how it looks with the I-Build from 11th November:
Image

And this is how it looks with the 2025-09 release:
Image

In addition, the vertical spaces between the elements are increased

**Reproduction**
Start a fresh SDK 175% primary monitor.

**Expected Behavior**
Same alignment as in the 2025-09 release.

HeikoKlare:
The command labels in a perspective's onboarding area are placed in a right-aligned layout but the labels themselves are not right-aligned, such that they are not perfectly right aligned as they can be up to 1 point off (which can sum up to multiple pixels with monitor zoom values).

This change adapts the labels to be right aligned such that they properly fit.

### Before
At 175%:
image

### After
At 175%:
image


akoch-yatta (approved):
Having the text aligned right here definitely makes sense. I gave it a short test and it worked as expected for me.

BeckerWdf:
Thanks for fixing

PR #3542, opened 2025-11-14
Fix incorrect SubMonitor usage patterns

vogella:
This commit addresses SubMonitor anti-patterns as described in https://www.eclipse.org/articles/Article-Progress-Monitors/article.html

Changes:
1. Removed unnecessary done() calls on monitors after SubMonitor.convert()
   - SubMonitor automatically handles done() when work is complete
   - Affected files:
     * Refactoring.java
     * PerformChangeOperation.java * PerformRefactoringOperation.java * MultiStateTextFileChange.java * TextChange.java * SmartImportJob.java (2 instances) * SaveablesList.java * SaveableHelper.java * CheckConditionsContext.java

2. Removed redundant isCanceled() checks after split() calls
   - split() already includes implicit cancellation checks
   - Affected files:
     * ProcessorBasedRefactoring.java (4 instances)
     * CheckConditionsContext.java * SaveablesList.java * SaveableHelper.java

These changes improve code quality by following SubMonitor best practices and removing redundant code that could mask issues.

### DIFF
--- a/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/TextChange.java
+++ b/bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/TextChange.java
@@ -247,7 +247,6 @@ public abstract class TextChange extends TextEditBasedChange {
 			throw Changes.asCoreException(e);
 		} finally {
 			releaseDocument(document, subMon.newChild(1));
-			subMon.done();
 		}
 	}
 
--- a/bundles/org.eclipse.ltk.core.refactoring/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ltk.core.refactoring/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.ltk.core.refactoring
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ltk.core.refactoring; singleton:=true
-Bundle-Version: 3.15.100.qualifier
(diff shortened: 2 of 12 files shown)


bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/TextChange.java
```diff
@@ -247,7 +247,6 @@ public Change perform(IProgressMonitor pm) throws CoreException {
 			throw Changes.asCoreException(e);
 		} finally {
 			releaseDocument(document, subMon.newChild(1));
-			subMon.done();
```
ptziegler, line 250: Wouldn't this cause https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1833 to reappear? Or is this why the call to `done()`  in `getCurrentDocument(IProgressMonitor)` remains?

My understanding is that the call to `done()` is only irrelevant, if the progress monitor is always used correctly. But I have yet to see a project where that is actually the case.

I.e. I would expect test failures in other components, which then also need to be updated.

vogella: Good point. So should we leave it as is?

ptziegler: If we can already expect test failures in other components then I would definitely leave it out. For the rest, I'd have to take a closer look to see which are definitely save and which might cause side-effects.
Perhaps it makes sense to split this PR up? Have one for the removal of the `isCanceled()` check (which at first glance looks safe) and one for the removal of `done()`.

Issue #3517, opened 2025-11-07
More clean-up actions


vogella:
Somewhere (in the unmanageable amount of PR's) @laeubi suggested that I should activate more JDT clean-ups.

Happy to do this next week as I have a bit of time on my hand for Eclipse releated activities.

@jjohnstn or @laeubi which performance or other clean-ups can you recommend to activate?

<img width="1261" height="1244" alt="Image" src="https://github.com/user-attachments/assets/373cd735-8738-49d0-a953-c78f6b3d0f1d" />

<img width="1256" height="1243" alt="Image" src="https://github.com/user-attachments/assets/e2302775-9457-475b-916a-8e801f7ff6d7" />

laeubi:
@vogella I'm not sure M3 week is the best to activate new cleanups (as there is less time to fix issues then), maybe it is better to wait for start of next release master opens? But of course it all depends a bit how dedicated you plan to work on these things.

What one usually can already do is to activate something that sounds useful, apply it to a few projects and then review the outcome to guess if it helps or if one can already discover some bugs.

vogella:
I agree with M3, the activation of new clean-ups must not be merged, only because we have a new PR for this. 

No particular suggestion for activation of clean-ups in the next release?

vogella:
No suggestion at the moment

laeubi:
> No suggestion at the moment

I'm not sure what to suggest, last time you opened a PR with some cleanups applied and I suggested to do it automatically instead. So that would then be something to enable ...

vogella:
I'm not aware of such a clean-up. If we do please let me know which one it is.

laeubi:
> I'm not aware of such a clean-up. If we do please let me know which one it is.

See for example here:

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3515

in general I think we should prefer enabling cleanups when we stumble over it instead of making it manual/ (or AI semi automatic).

laeubi:
Another example:

- https://github.com/eclipse-pde/eclipse.pde/pull/2058

vogella:
Simplify lambda expression is not yet enabled? This one makes sense IHMO. Configuration for the clean-ups were in the aggregator repo, right?

laeubi:
> Simplify lambda expression is not yet enabled?

You get a textual representation of whats enabled when you look at any of the cleanup PRs

[...]

PR #3544, opened 2025-11-14
JDT Clean-up Use String.replace() instead of String.replaceAll() when

vogella:
possible

In https://github.com/eclipse-platform/eclipse.platform.ui/issues/3517 we discuss which automatic clean-ups we should add. This is the result of a manual run in eclipse-platform.ui to see if this clean-up has issues. Also by running it manually before enabling the clean-ups we avoid creating PR per bundle.

### DIFF
--- a/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/internal/forms/widgets/FormTextModel.java
+++ b/bundles/org.eclipse.ui.forms/src/org/eclipse/ui/internal/forms/widgets/FormTextModel.java
@@ -122,11 +122,11 @@ public class FormTextModel {
 
 	private String processAmpersandEscapes(String pTaggedText) {
 		try {
-			String taggedText = pTaggedText.replaceAll("&quot;", "&#034;"); //$NON-NLS-1$//$NON-NLS-2$
-			taggedText = taggedText.replaceAll("&apos;", "&#039;"); //$NON-NLS-1$//$NON-NLS-2$
-			taggedText = taggedText.replaceAll("&lt;", "&#060;"); //$NON-NLS-1$//$NON-NLS-2$
-			taggedText = taggedText.replaceAll("&gt;", "&#062;"); //$NON-NLS-1$//$NON-NLS-2$
-			taggedText = taggedText.replaceAll("&amp;", "&#038;"); //$NON-NLS-1$//$NON-NLS-2$
+			String taggedText = pTaggedText.replace("&quot;", "&#034;"); //$NON-NLS-1$//$NON-NLS-2$
+			taggedText = taggedText.replace("&apos;", "&#039;"); //$NON-NLS-1$//$NON-NLS-2$
+			taggedText = taggedText.replace("&lt;", "&#060;"); //$NON-NLS-1$//$NON-NLS-2$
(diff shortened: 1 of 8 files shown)


laeubi:
So do you see any issues with that or not?

vogella:
@jjohnstn I think the change in AbstractTemplatePage are incorrect:

---------
  // Before (CORRECT):
  pattern.replaceAll("\\$", "\\$\\$")

  // After (INCORRECT):
  pattern.replace("$", "$$")

  Problem: The original code was escaping $ for regex replacement strings, not for regex patterns. In replaceAll(), the replacement string has special meaning for $ (backreferences like $1, $2). To insert a literal $,
  you need \\$\\$ (which becomes $$ after Java string escaping).

  However, String.replace() doesn't interpret $ specially in the replacement string. So while $$ will insert two literal dollar signs, the original intent was to escape one dollar sign for template syntax.

  Impact: This changes behavior - templates with $ will now have $$ instead of being properly escaped for the template engine.

---------

If you agree, I can open an issue in JDT UI

vogella:
Converting to draft so that this is not merged.

merks:
I'm a big confused....

This method
```
		for (var pattern : List.of("$", "$$", "${}")) {
			System.err.println(pattern.replaceAll("\\$", "\\$\\$"));
			System.err.println(pattern.replace("$", "$$"));
			System.err.println();
		}
	}
```
prints this result
```
$$
$$

$$$$
$$$$

$${}
$${}
```

So I'm tying to imagine a string for which they would print out two  different results...

vogella:
> I'm a big confused....
> 
> This method
> 
> ```
> 	public static void main(String[] args) {
> 		for (var pattern : List.of("$", "$$", "${}")) {
> 			System.err.println(pattern.replaceAll("\\$", "\\$\\$"));
> 			System.err.println(pattern.replace("$", "$$"));
> 			System.err.println();
> 		}
> 	}
> ```
> 
> prints this result
> 
> ```
> $$
> $$
> 
> $$$$
> $$$$
> 
> $${}
> $${}
> ```
> 
> So I'm tying to imagine a string for which they would print out two different results...

You're absolutely right to be confused - I made an error! Let me verify this myself:
 ```
 import java.util.List;

     public class TestStringReplace {
             System.out.println("Testing different patterns:");
             for (var pattern : List.of("$", "$$", "${}", "$1", "test$middle", "$start", "end$")) {
                 String replaceAllResult = pattern.replaceAll("\\$", "\\$\\$");
                 String replaceResult = pattern.replace("$", "$$");
                 System.out.println("Input: '" + pattern + "'");
                 System.out.println("  replaceAll: '" + replaceAllResult + "'");
                 System.out.println("  replace:    '" + replaceResult + "'");
                 System.out.println("  Same? " + replaceAllResult.equals(replaceResult));
                 System.out.println();
             }
         }
     }
 ```
I was wrong. @jjohnstn please ignore my request.

I would still wait this converting this to a real PR again, as IMHO we should not apply such clean-ups shortly before a release.

akurtakov:
Please wait for https://github.com/eclipse-platform/eclipse.platform.ui/pull/3594 to land as it was quite a work and I would appreciate less conflicts in it.

PR #3549, opened 2025-11-18
Add scrollbars to wizard dialog via ScrolledComposite

### LINKED ISSUE #526: Text cut-off in new class dialog
> Text cutt-off in new class dialog, the text after `Do you want to add comments is cut-off`, screenshot from 250% zoom, content even cut off at 100% zoom, Issue was not present in  in ~~2025-09~~ 2025-06
> 
> Image 

 _Originally posted by @arunjose696 in [#514](https://github.com/vi-eclipse/Eclipse-Platform/issues/514#issuecomment-3480250111)_

This issue is reproducible in the latest ibuild, the text is cut off with the setup "primary 250(left), secondary 175, theme dark"

arunjose696:
Previously, JFace wizard dialogs could have their content cut off without any scrollbars. This occurred when users resized the dialog or moved it between monitors, where size adjustments caused by rounding could make the dialog smaller than its content. As a result, parts of the UI could become inaccessible without any visual indication.

This change wraps the content in a ScrolledComposite, with vertical and horizontal scrollbars so users can still access all content even when the dialog becomes smaller.

Steps to reproduce

1)Open a new class dialog in eclipse IDE
2)Resize it to make it smaller to see the scrollbars appear with this change

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
@@ -52,6 +52,7 @@ import org.eclipse.swt.SWT;
 import org.eclipse.swt.accessibility.AccessibleAdapter;
 import org.eclipse.swt.accessibility.AccessibleEvent;
 import org.eclipse.swt.custom.BusyIndicator;
+import org.eclipse.swt.custom.ScrolledComposite;
 import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.graphics.Cursor;
 import org.eclipse.swt.graphics.Point;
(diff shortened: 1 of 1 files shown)


laeubi (changes requested):
@arunjose696 I don't think this is a valid change. This will suddenly affect all users of WizardDialog (and even break if they depend on internal behavior).

If anything needs to be done (resizing / relayout / ...) then we need to register a zoom change listener and perform the adjustment in a new protected methods so consumers can possibly override that method.

arunjose696:
> @arunjose696 I don't think this is a valid change. This will suddenly affect all users of WizardDialog (and even break if they depend on internal behavior).
> 


The design of the wizard dialog seems pretty flawed as one can resize (and in particular shrink) it without a scrollbar being provided for the contents. wouldnt the dialogs function as before if the content fits right and The scrollbars only show up when something gets clipped, wouldnt this be a reasonable default for all wizard dialogs?

> If anything needs to be done (resizing / relayout / ...) then we need to register a zoom change listener and perform the adjustment in a new protected methods so consumers can possibly override that method.

I can move this to a protected method, but Would it be OK to make the scrolling behavior opt-out? That way anyone who relies on the old behavior can override it, while everyone else gets a scrollbars as default?

laeubi:
> The design of the wizard dialog seems pretty flawed as one can resize (and in particular shrink) it without a scrollbar being provided for the contents. 

This is for sure but we can't just change it. Also the question is more is the wizard dialog flawed or the implementation that puts content into the dialog. e.g. if I carefully design my controls to have some content that do scrolling (e.g. a table) I won't see the issue but now possibly getting additional scrollbars.

What JFace (and SWT in general) mostly miss is the concept of a min / max size of controls.

> I can move this to a protected method, but Would it be OK to make the scrolling behavior opt-out? 

No see above, only opt-in would be compatible but I doubt we should do it. If I want such behavior I can create s scrolled composite on my own (what has its own limitations and quirks).

> while everyone else gets a scrollbars as default?

Scrollbars of the full content are really a wore UI design on its own and should really never be used without care or other options considered before.

arunjose696:
> 
> This is for sure but we can't just change it. Also the question is more is the wizard dialog flawed or the implementation that puts content into the dialog. e.g. if I carefully design my controls to have some content that do scrolling (e.g. a table) I won't see the issue 

To fix the issue in implementatation to fix issues with new class dialog I have opened the below PR in JDT so closing this effort

- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2653

PR #3550, opened 2025-11-18
Revert workaround for unexpected text wrapping in ExpandableComposite

### LINKED ISSUE #539: Revert workaround in ExpandableComposite
Tracking ticket for:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3550

HeikoKlare:
A workaround for unexpected text wrapping in ExpandableComposite had been introduced to deal with imprecise size calculation of controls for zooms like 125% and 175%. With recent changes to SWT, the size calculation for controls has been improved such that they always allocate sufficient space to fit their contents in. That makes the workaround in ExpandableComposite obsolete.

This change thus removes the obsolete, temporary workaround introduced with:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2986

### DIFF
--- a/bundles/org.eclipse.ui.forms/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.forms/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.ui.forms;singleton:=true
-Bundle-Version: 3.13.700.qualifier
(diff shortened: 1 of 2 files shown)


arunjose696 (approved):
Tested with zoom levels 125, 150, 175, and 250. The issue that previously required a workaround is no longer present. It can be observed that using `pixelToPointAsSufficientlyLargeSize` to calculate sizes which performs rounding up resolves the issue.
If `RoundingMode.Round` is used in `pixelToPointAsSufficientlyLargeSize`, the issue can reappear without the workaround. This confirms that the current approach fixes the problem. So approving this revert.

PR #3552, opened 2025-11-21
Fix traverse in multi page editors

Christopher-Hermann:
control.traverse on a CTabFolder in the MultiPageEditor is navigation up the the outer CTabFolder itself.

This is causing several issues:
Setting the page for example to the second-to-last is changing the page in the multipage editor. Then the traverse direction is changed causing the outer tab to move in the wrong direction. Furthermore on the outer tab you cannot jump from the last tab to the very first tab.

### How to reproduce

1. Open two .java editor and one multi page editor, for example the manifest.
2. On the java editors, traverse to the next tab until you are in the manifest editor
3. In the manifest editor, traverse next tab will not work
4. Traverse to previous tab will jump to the `plugin.xml` inner tab
5. Traverse to previous again will jump to the java editor

See 

https://github.com/user-attachments/assets/3af76624-462c-40e6-85e3-bc5c8b9d56ea

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/handlers/TraversePageHandler.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/handlers/TraversePageHandler.java
@@ -41,6 +41,15 @@ public class TraversePageHandler extends WidgetMethodHandler {
 			int traversalDirection = translateToTraversalDirection(forward);
 			Control control = focusControl;
 			do {
+				if (control instanceof CTabFolder folder && !isRootCTabFolder(folder)) {
+					// For multi-page editors, we want to skip inner CTabFolders
+					// If not skipped there are two issues:
+					// 1. loopToFirstOrLastItem would change the selection of the inner CTabFolder
+					// 2. control.traverse itself would navigate up to the outer CTabFolder and
+					// navigate there
+					control = control.getParent();
+					continue;
+				}
 				if (control instanceof CTabFolder folder && isFinalItemInCTabFolder(folder, forward)
 						&& !hasHiddenItem(folder)) {
 					loopToFirstOrLastItem(folder, forward);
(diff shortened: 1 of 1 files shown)


Christopher-Hermann:
@fedejeanne I saw that you introduced the fix that enables to navigate from the last to the first tab:
```
loopToFirstOrLastItem(folder, forward);
traversalDirection = translateToTraversalDirection(!forward); 
```

Maybe you can have a look on my fix. I'm not really sure if this is really the best solution for the problem. It somehow feels like the wrong place here since `control.traverse(traversalDirection)` is looping up to the parent controls as well. So now we somehow skip the traverse in multi page editors. Not sure if this is a problem.

Maybe you have an opinion on that, but I guess you are not an expert here as well :D

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/handlers/TraversePageHandler.java
```diff
@@ -41,6 +41,15 @@ public final Object execute(final ExecutionEvent event) {
 			int traversalDirection = translateToTraversalDirection(forward);
 			Control control = focusControl;
 			do {
+				if (control instanceof CTabFolder folder && !isRootCTabFolder(folder)) {
+					// For multi-page editors, we want to skip inner CTabFolders
+					// If not skipped there are two issues:
+					// 1. loopToFirstOrLastItem would change the selection of the inner CTabFolder
+					// 2. control.traverse itself would navigate up to the outer CTabFolder and
+					// navigate there
+					control = control.getParent();
```
fedejeanne, line 50: Just a guess, but I think you should have navigated all the way to the top with a loop, just to be sure you're navigating on the "Editor Area" and not inside a multi-page editor.

In any case, I plan to merge #3558 and added you and Matthias as co-authors :-)

fedejeanne (commented):
Just a comment... I'll proceed with #3558 as a joint effort 💪

PR #3558, opened 2025-11-25
Fix traversal of CTabFolders with PgUp/PgDown #2864

### LINKED ISSUE #2864: Eclipse 2025-03 Next Tab navigation broken with POM XML editor.
This bug is occurring in Eclipse EE 2025-03 on Windows. (I haven't used Eclipse in over a year, so I don't know if it was introduced in an earlier version.)

`Ctrl+PgUp` is mapped to "Previous Tab", and `Ctrl+PgDown` is mapped to "Next Tab". This has been working for decades, allowing me to navigate among tabs as expected in the Eclipse IDE.

1. In Eclipse EE 2025-03, open multiple files.
2. Additionally open a Maven POM file. The POM editor shows tabs at the bottom: "Overview", "Dependencies", "Dependency Hierarchy", "Effective POM", and "pom.xml".
3. Open one more file so that its file tab shows up _after_ the POM file tab.
4. Go back to the POM file tab and select the "pom.xml" POM view tab at the bottom. (I have this set up to the default.)
5. Select `Ctrl+PgUp` to go to the previous file tab.
6. **The selected file is changed to the previous open file tab as expected.**
7. Go back to the POM file tab.
5. Select `Ctrl+PgDown` to go to the next file tab.
6. **Rather than going to the next open file tab, the editor switches to the _previous_ open file tab.** (Additionally, if you manually return to the POM file tab, you'll see that the editor view has changed from "pom.xml" to "Dependencies".

As you can imagine, this completely breaks navigating among file tabs when one of them is a Maven POM file. Perhaps the editor view tab hotkey mapping is overriding the global Eclipse IDE file editor tab. In any case, it is a new bug.

fedejeanne:
When traversing from an editor with tabs (e.g. a POM or MANIFEST editor), which is a CTabFolder itself, I made sure that the traversal happens from the top-level CTabFolder i.e. from the one that contains all open editors instead of trying to traverse inside the editor itself.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2864

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/handlers/TraversePageHandler.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/handlers/TraversePageHandler.java
@@ -41,11 +41,12 @@ public class TraversePageHandler extends WidgetMethodHandler {
 			int traversalDirection = translateToTraversalDirection(forward);
 			Control control = focusControl;
 			do {
-				if (control instanceof CTabFolder folder && isFinalItemInCTabFolder(folder, forward)
-						&& !hasHiddenItem(folder)) {
-					loopToFirstOrLastItem(folder, forward);
-					traversalDirection = translateToTraversalDirection(!forward); // we are in the second-to-last item in the given
-					// direction. Now, use the Traverse-event to move back by one
+				if (control instanceof CTabFolder) {
+					CTabFolder folder = getTopLevelCTabFolderInParentHierarchy(control);
+					if (isFinalItemInCTabFolder(folder, forward) && !hasHiddenItem(folder)) {
+						loopToFirstOrLastItem(folder, forward);
+						traversalDirection = translateToTraversalDirection(!forward);
+					}
 				}
 				if (control.traverse(traversalDirection)) {
 					return null;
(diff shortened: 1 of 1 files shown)


BeckerWdf:
is this the same issue as reported / fixed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3552 ?

BeckerWdf:
@Christopher-Hermann

Christopher-Hermann:
> is this the same issue as reported / fixed in #3552 ?

Ah yes, damn. I did not see/find the initial issue. 
It's definitely the same issue and even almost the same fix/code change we did :D

Christopher-Hermann (approved):
Looks good. Retested and everything works like expected.

fedejeanne:
Thank you for looking into it, @Christopher-Hermann and @BeckerWdf ! If that's OK with you, I would like to merge my PR since I think getting the "top-level" `CTabFolder` is a bit safer than getting "the next `CTabFolder` (just a guess, not a hard-proven fact).

... but since you worked on it too ...
> It's definitely the same issue and even almost the same fix/code change we did :D

... I added you both as Co-Authors :-)

I plan to merge as soon as 4.39 is open for development.

fedejeanne:
Failing test is unrelated:
- #2708

PR #3573, opened 2025-12-11
 Guard clipping to avoid resize artifacts in tree viewer

### LINKED ISSUE #548: Artifacts when resizing column of TableViewer with StyledCellLabelProvider
**Description**
When resizing the columns of a TableViewer with StyledCellLabelProviders on a 175% monitor, rendering artifacts appear. It seems like the point/pixel values for some of the involed areas (clipping or the like) are not rounded correctly.

**Reproduction**
The issue can be reproduced with the following snippet and a 175% monitor monitor.
[Main2.java](https://github.com/user-attachments/files/23820226/Main2.java)

Image

**Expected Behavior**
No rendering artifacts occur.

**Additional knowledge**
So this is probably related to the issue we see in #518.

arunjose696:
Setting the clipping region temporarily to the textBounds   was originally added (https://bugs.eclipse.org/bugs/show_bug.cgi?id=291245 comment 25) to prevent images  from being overdrawn when implementting  RIGHT or CENTER column alignment support in StyledCellLabelProvider. However, enabling  clipping causes artifact lines during tree resizing, as a temporary workaround the  clipping call is  now guarded and applied only in those alignment cases to avoid issues with resizing of columns.

**Steps to reproduce**
Step 0:Set monitor zoom to 175
1. Set monitor zoom to 175%.
2. Launch a runtime workspace.
3. In that runtime workspace, open the **Debug Configurations** dialog.
4. Go to the **Plug-ins** tab for another runtime workspace / Eclipse application.
5. Resize the **ID** column in the *Features* view. Artifact lines appear **without** this workaround.

### Behavior *without* the fix
https://github.com/user-attachments/assets/1525680f-e9e4-459a-a8e1-0860b62a5cd6

### Behavior *with* the fix

https://github.com/user-attachments/assets/99223cd7-545d-41f5-883e-25bcf854fe00

Additionally this issue can also be reproduced with the snippet provided in https://github.com/vi-eclipse/Eclipse-Platform/issues/548#issue-3674207659

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StyledCellLabelProvider.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StyledCellLabelProvider.java
@@ -389,7 +389,7 @@ public abstract class StyledCellLabelProvider extends OwnerDrawLabelProvider {
 			int y = textBounds.y
 					+ Math.max(0, (textBounds.height - layoutBounds.height) / 2);
 
-			if (gc.isClipped()) {
+			if (gc.isClipped() && ((style & SWT.RIGHT) != 0 || (style & SWT.CENTER) != 0)) {
 				Rectangle saveClipping = gc.getClipping();
 				gc.setClipping(textBounds);
 				textLayout.draw(gc, x, y);


akoch-yatta (approved):
Tested with the Runtime and the mentioned Snippet. Adapted the Snippet to be able to test with TreeViewer and TableViewer. This PR works with both scenarios. TreeViewer doesn't properly work with SWT.RIGHT and SWT.CENTER when the first column is resized (at least in windows), but this is a pre-existing and not caused or worsened with this PR

PR #3576, opened 2025-12-12
Refactor: Improve refresh action performance with parallel processing

vogella:
Replaced WorkspaceModifyOperation with IRunnableWithProgress and direct Job usage in RefreshAction and ResourceMgmtActionProvider to enable parallel refreshing of resources and improve performance.

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/RefreshAction.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/RefreshAction.java
@@ -18,14 +18,14 @@ package org.eclipse.ui.actions;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


iloveeclipse (changes requested):
The original code was executed as `WorkspaceModifyOperation` with a rule, not it seem that you explicitly allows multiple threads to run multiple workspace related tasks without holding any rule.

Even as a draft, this makes no sense.

merks:
> Even as a draft, this makes no sense.

I was about to post the same comment for the same reason.

iloveeclipse:
Workspace operations in general can't be parallelized, there is always a central workspace lock anyway.

vogella:
> Workspace operations in general can't be parallelized, there is always a central workspace lock anyway.

The WorkspaceJob Javadoc says the opposite :

  // org.eclipse.core.resources/WorkspaceJob.java (Starting line 47)

````

 * Note that the workspace is not locked against other threads during the execution
 * of a workspace job. Other threads can be modifying the workspace concurrently
 * with a workspace job.  To obtain exclusive access to a portion of the workspace,
 * set the scheduling rule on the job to be a resource scheduling rule.  The
 * interface {@link IResourceRuleFactory} is used to create a  scheduling rule
 * for a particular workspace modification operation.
````

PR #3578, opened 2025-12-12
Remove redundant isCanceled checks in SaveablesList and SaveableHelper

vogella:
This addresses the proposal by ptziegler in PR #3542 to split the removal of isCanceled checks.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/SaveableHelper.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/SaveableHelper.java
@@ -193,9 +193,6 @@ public class SaveableHelper {
 						continue;
 					}
 					doSaveModel(model, subMonitor.split(1), window, confirm);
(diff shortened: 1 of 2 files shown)


ptziegler (approved):
The call of `newChild()` has been changed to `split()` as part of '[Bug 479879](https://bugs.eclipse.org/bugs/show_bug.cgi?id=479879) - Adopt SubMonitor.split in platform UI'. For those classes specifically in c7cb5c40a42eeab359a7e7de3f08a714ead2fa08.

These methods throw an `OperationCanceledException` if the monitor has been canceled by the user. If there are any side-effects from throwing an exception, they would've become apparent by now.

I would even go so far and say that simply breaking out of the loop is an anti-pattern, because the program could now continue in a partially saved state, rather than properly handling the cancellation.

ptziegler:
The test case has been sporadically failing and is tracked via https://github.com/eclipse-platform/eclipse.platform.ui/issues/3581. It should be irrelevant to this change.

vogella:
Thanks @ptziegler

PR #3587, opened 2025-12-16
Move bundles to Java 21

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/2824 ) so this change makes it to state it directly and silence the PDE warnings.


laeubi:
@akurtakov but then please also raise the lower bound to the once from the dependencies, otherwise you silence the warnings but it makes no sense as then the resolver could still resolve the old bundles.

Otherwise this statement

> They already transitively require it (due to https://github.com/eclipse-platform/eclipse.platform.swt/pull/2824 )

is not true... because they only require it transitively if resolved against never SWT versions but not in general!

akurtakov:
New warnings are URL constructors being deprecated but these need extra care so not doing it now.

PR #3588, opened 2025-12-16
Fix invalid parameter name in BindingManager#getActiveBindingsFor1 Javadoc

danthe1st:
The `BindingManager#getActiveBindingsFor1` method has a parameter named `command` but the `@param` in the Javadoc comment references the name `commandId`. This slight inconsistency is there since https://github.com/eclipse-platform/eclipse.platform.ui/commit/82271bc612f6b4eee954d66d3434b5e1cb646a5a where the signature was updated without the Javadoc and causes an error/warning (I think it's shown as an error due to the project's properties) when trying to [compile Eclipse Platform with javac](https://github.com/eclipse-jdtls/eclipse.jdt.javac) (doing that also causes a few other warnings).


danthe1st:
The warnings seem to be due to https://github.com/eclipse-platform/eclipse.platform.ui/commit/01831525ef6fe115c8d9f21b22ef9bd7a6ce0c7b which is in master.
> The constructor URL(String) is deprecated since version 20

iloveeclipse (approved):

PR #3590, opened 2025-12-16
TreeViewer.collapseToLevel() consistency test

iloveeclipse:
Test to verify JFace / SWT consistency if using
TreeViewer.collapseToLevel() with ALL_LEVELS vs specific level argument

This test will fail on GTK without change in SWT below.

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/2888 
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2887


iloveeclipse:
As expected, new `testExpandCollapseToLevel()` test fails on Linux.

iloveeclipse:
testDirtyingWhenClean (JFaceTextTestSuite FastAbstractReconcilerTest) failure on Mac is unrelated.

PR #3592, opened 2025-12-16
Move bundles to Java 21 - take 2

akurtakov:
eclipse-platform/eclipse.platform.swt#2824 ) so this change makes it to state it directly and silence the PDE warnings.


akurtakov:
Java 21 deprecations to be handled separately. Merging despite them.

PR #3593, opened 2025-12-17
Move bundles to Java 21 - take 3

akurtakov:
eclipse-platform/eclipse.platform.swt#2824 ) so this change makes it to state it directly and silence the PDE warnings.


iloveeclipse:
org.eclipse.jface.text.tests/.classpath still has Java 17 while MANIFEST needs 21. 
Not sure if this is fixed in this PR fixed or not, but also please double check other bundles.

akurtakov:
> org.eclipse.jface.text.tests/.classpath still has Java 17 while MANIFEST needs 21. Not sure if this is fixed in this PR fixed or not, but also please double check other bundles.

Thanks for spotting it, fixed now. PDE warns about many things but it's impossible to spot new warning in the current 1600+ . I'm going through them and will eventually spot such things too . Help would be more than appreciated though.

akurtakov:
New warnings are Java 21 deprecations. These are to be handled later as especially the URL ones have to be treated extra carefully. Merging.

PR #3597, opened 2025-12-18
Use OfFloat Point values to create precise highlight brackets

### LINKED ISSUE #521: Bracket selection square too small in some zooms
**Description**
The bracket selection square seems to be too small in some zooms. Some value is probably not scaled up correctly. In my opinion it is always t

Screenshot was made on  125% zoom. Lower end of the square overlaps with the bracker

Image

Screenshot with big font on 100%

Image

**Reproduction**
1. Open Java editor
2.  Select a bracket to have the counterpart highlighted

**Expected Behavior**
The bracket highlighter should look consistent on all zooms

ShahzaibIbrahim:
Because of rounding errors the bracket selection highlight is often inconsistent. Making use of residual values provides precise rectangle values and highlight is consistent across all zooms

Problem: At 250% (the problem is also there for all zooms, at 250 its just better visible), After every three line the fourth line x axis is bit too much on the left. (rounding error)

line 29: the width of highlight is 17px

image

line 30: the width of highlight is 16px

image

### How to test

- Run the runtime workspace with monitor-specific scaling turned on
- Add some brackets in Java file { }
- Select the opening bracket, the closing bracket will be highlighted. 
- See if the rectangle around highlighted bracket is surrounded the bracket correctly (no-cutoffs)

### Result 

**Before Fix:**
image

**After Fix:**
image

**Requires:**

- [x] https://github.com/eclipse-platform/eclipse.platform.swt/pull/2907

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/MatchingCharacterPainter.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/MatchingCharacterPainter.java
@@ -257,17 +257,15 @@ public final class MatchingCharacterPainter implements IPainter, PaintListener {
 		if (gc != null) {
 			gc.setForeground(fColor);
 
-			final Rectangle bounds= fTextWidget.getTextBounds(offset, offset);
+			Point offsetLocation= fTextWidget.getLocationAtOffset(offset);
 
-			// determine the character width separately, because the getTextBounds above
-			// will also include any in-line annotations (e.g. codemining annotations) in the width
 			final String matchingCharacter= fTextWidget.getText(offset, offset);
-			final int width= gc.textExtent(matchingCharacter).x;
-
-			final int height= fTextWidget.getCaret().getSize().y;
+			Point characterBounds= gc.textExtent(matchingCharacter);
+			characterBounds.y-= 1;
+			Rectangle hightlightingArea= Rectangle.of(offsetLocation, characterBounds);
 
 			// draw box around line segment
-			gc.drawRectangle(bounds.x, bounds.y + bounds.height - height, width, height - 1);
+			gc.drawRectangle(hightlightingArea);
 		} else {
 			fTextWidget.redrawRange(offset, 1, true);
 		}


bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/MatchingCharacterPainter.java
```diff
@@ -262,12 +263,12 @@ private void draw(final GC gc, final int offset) {
 			// determine the character width separately, because the getTextBounds above
 			// will also include any in-line annotations (e.g. codemining annotations) in the width
 			final String matchingCharacter= fTextWidget.getText(offset, offset);
-			final int width= gc.textExtent(matchingCharacter).x;
+			final Point sizePoints= gc.textExtent(matchingCharacter);
 
-			final int height= fTextWidget.getCaret().getSize().y;
```
HeikoKlare, line 267: Removing the incorporation of the caret size seems to be wrong. As indicated by the comment, the `bounds` include, e.g., in-line annotations. So with this change you get such results:
image

HeikoKlare (dismissed):

HeikoKlare (commented):
Is it intended that with removing the `-1` from the height calculation, the highlight is now too high and exceeds the line?
image

Before:
image

ShahzaibIbrahim:
> Is it intended that with removing the `-1` from the height calculation, the highlight is now too high and exceeds the line? image
> 
> Before: image

Yes, it is intended. We either cutoff the highlighted bracket or we go over the line.

HeikoKlare:
So if I understand correctly, we have two issues with two solutions here:
1. The highlight is drawn over the bracket, which is solved by removing the `-1`.
2. The highlight is too small on higher zooms because of bad rounding, which is solved by using `Rectangle.of()`.

If propose to make two separate PRs, one for each of the issues, out of this. While the usage of `Rectangle.of()` to solve the scaling issue should be without any discusionss, the `-1` has been added intentionally to make the highlight fit to the line, so changing that behavior might be more of a discussion.

ShahzaibIbrahim:
> So if I understand correctly, we have two issues with two solutions here:
> 
> 1. The highlight is drawn over the bracket, which is solved by removing the `-1`.
> 2. The highlight is too small on higher zooms because of bad rounding, which is solved by using `Rectangle.of()`.
> 
> If propose to make two separate PRs, one for each of the issues, out of this. While the usage of `Rectangle.of()` to solve the scaling issue should be without any discusionss, the `-1` has been added intentionally to make the highlight fit to the line, so changing that behavior might be more of a discussion.

This PR now keep the height as it was before. The other PR that increases the height is here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3653

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/MatchingCharacterPainter.java
```diff
@@ -262,12 +262,13 @@ private void draw(final GC gc, final int offset) {
 			// determine the character width separately, because the getTextBounds above
 			// will also include any in-line annotations (e.g. codemining annotations) in the width
 			final String matchingCharacter= fTextWidget.getText(offset, offset);
-			final int width= gc.textExtent(matchingCharacter).x;
-
+			Point sizePoints= gc.textExtent(matchingCharacter);
```
HeikoKlare: Please use proper variable name such as `characterBounds` or the like.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/MatchingCharacterPainter.java
```diff
@@ -262,12 +262,13 @@ private void draw(final GC gc, final int offset) {
 			// determine the character width separately, because the getTextBounds above
 			// will also include any in-line annotations (e.g. codemining annotations) in the width
 			final String matchingCharacter= fTextWidget.getText(offset, offset);
-			final int width= gc.textExtent(matchingCharacter).x;
-
+			Point characterBounds= gc.textExtent(matchingCharacter);
+			characterBounds.y= characterBounds.y - 1;
 			final int height= fTextWidget.getCaret().getSize().y;
+			Rectangle rectangleOfFloat= Rectangle.of(new Point(bounds.x, bounds.y + bounds.height - height), characterBounds);
```
akoch-yatta: rectangleOfFloat is not a good name here, rather something like hightlightingArea (?), or just inline it below.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/MatchingCharacterPainter.java
```diff
@@ -262,12 +262,13 @@ private void draw(final GC gc, final int offset) {
 			// determine the character width separately, because the getTextBounds above
 			// will also include any in-line annotations (e.g. codemining annotations) in the width
 			final String matchingCharacter= fTextWidget.getText(offset, offset);
-			final int width= gc.textExtent(matchingCharacter).x;
-
+			Point characterBounds= gc.textExtent(matchingCharacter);
+			characterBounds.y= characterBounds.y - 1;
 			final int height= fTextWidget.getCaret().getSize().y;
```
akoch-yatta: Before the height used for drawRectangle was `height - 1`, now you are using `characterBounds.y - 1`. I would have expected something like `characterBounds.y= height - 1;` Is that on purpose?

ShahzaibIbrahim: Yes, you are correct. I should have used height -  1 since we are not changing anything related to height in this PR.

vogella:
Can this be merged? @HeikoKlare you did removed your stale review, does that mean you approve the change?

akoch-yatta:
> Can this be merged? @HeikoKlare you did removed your stale review, does that mean you approve the change?

I will have a look at it today, so hopefully, yes

akoch-yatta (changes requested):
Won't merge this one today. While it fixes some scenarios it break other scenarios now, e.g. overlapping the line in the bottom in some scenarios. We will revise if we can sync the highlight bracket calculation better with the line bounds calculation in StyledText overall.

ShahzaibIbrahim:
@akoch-yatta with the latest commit, I got rid of textWidget.getBoundsAtOffset() and use getLocationAtOffset instead since we only need location from styledText and the latter gives us the location without text mining (more accurate). Then we have inconsistency in width and height if you move across different lines of code, for that I use Rectangle.of to so it uses the point OfFloat values and round them accordingly to give us consistent results. 

I used this code to test the outline on brackets on different lines:

```
package test.pack;

public class TestClass {
		 
		private void book(String a) {
			
		}
		
		@Test
		private void bigbook() {
			book("dadadad");
		}
		
		{ <- thats where i start
			{
				{
					{
						{
							{
								{
									{
										{
											
										}
									}
								}
							}
						}
					}
				}
			}
		}
}
```
and printed out results between master and after my latest changes here. The results are as follow at 175%:

**Master:** 
drawRectangleInPixels: x=110 y=840 width=13 height=25
drawRectangleInPixels: x=163 y=814 width=12 height=24
drawRectangleInPixels: x=215 y=788 width=13 height=24
drawRectangleInPixels: x=268 y=761 width=12 height=25
drawRectangleInPixels: x=322 y=735 width=12 height=25
drawRectangleInPixels: x=375 y=709 width=12 height=24
drawRectangleInPixels: x=427 y=683 width=12 height=24
drawRectangleInPixels: x=481 y=656 width=13 height=25
drawRectangleInPixels: x=534 y=630 width=12 height=25

**After Changes:**

drawRectangleInPixels: x=110 y=866 width=13 height=27
drawRectangleInPixels: x=163 y=840 width=13 height=27
drawRectangleInPixels: x=216 y=814 width=13 height=27
drawRectangleInPixels: x=269 y=788 width=13 height=27
drawRectangleInPixels: x=322 y=761 width=13 height=27
drawRectangleInPixels: x=375 y=735 width=13 height=27
drawRectangleInPixels: x=428 y=709 width=13 height=27
drawRectangleInPixels: x=481 y=683 width=13 height=27
drawRectangleInPixels: x=534 y=656 width=13 height=27

akoch-yatta (approved):
I tested the latest changes with multiple zoom and it looks more consistent to me now. The implementation seems more straight forward as the previous one as well.

I did check the history of this method and having bigger line height (e.g. caused by code mining) were the reason for some of the changes. This still works as expected due to my tests.

HeikoKlare (approved):
This seems to improve the highlight on certain zooms (such as 175%). I did not find any regressions on other zooms / in other scenarios (especially including code minings).

On 175%, it used to look like this:
image

And now looks like this:
image

akoch-yatta:
Compiler warning is unrelated, I will merge this one

akurtakov:
@iloveeclipse Does the JDT warning makes sense to you? https://github.com/eclipse-platform/eclipse.platform.ui/runs/61975614845

PR #3605, opened 2025-12-19
Remove beeping from FilteredTree

vogella:
beeping is annoying and beeping in a FilteredTree in certain internal situations does not add value for the user.


merks:
Somehow discussions about beeps seem to ring a bells, so to speak.  There are many calls:

image

And indeed there is this thread:

https://github.com/eclipse-platform/eclipse.platform.swt/issues/2613

Maybe there are visually impaired people that will be unhappy?  I have no clue.  I don't like beeps personally, but that's just a personal opinion...

vogella:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/2613 triggered this change, the agreement seems to be that we should remove the beep from our code and once this is done, we should deprecate this method.

Issue #3104, opened 2025-07-11
[Edge] Eclipse freezes when opening more than 2 png files in internal web browser via Ctrl+Shift+R


arunjose696:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

Setup eclipse and all the eclipse platform related github repositories using oomph https://www.eclipse.org/setups/installer/?url=https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.releng.aggregator/master/oomph/PlatformSDKConfiguration.setup&show=true

Open 2 or more *.png files(Images) with ctrl+shift+r.  This causes eclipse to hang.

I expected: No hangs

But got: Eclipse hangs

## Here is jstack of eclipse process id 

[Jstack eclipse.txt](https://github.com/user-attachments/files/21177426/Jstack.eclipse.txt)

## Tested under this environment:
* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): Version: 2025-09 (4.37)
Build id: I20250710-1800

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

sratz:
Important excerpt of the stack:

```
"main" #1 [34668] prio=6 os_prio=0 cpu=77593.75ms elapsed=1087.31s tid=0x0000021089970e30 nid=34668 runnable  [0x0000003865dda000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.swt.internal.win32.OS.PeekMessage(Native Method)
        at org.eclipse.swt.browser.Edge.processOSMessagesUntil(Edge.java:530)
        at org.eclipse.swt.browser.Edge$WebViewProvider.getWebViewWrapper(Edge.java:435)
        at org.eclipse.swt.browser.Edge$WebViewProvider.getWebViewWrapper(Edge.java:424)
        at org.eclipse.swt.browser.Edge$WebViewProvider.getWebView(Edge.java:443)
        at org.eclipse.swt.browser.Edge.isBackEnabled(Edge.java:1528)
[...]

PR #3614, opened 2025-12-29
Process KeyBinding Command Asyncly for Browser

### LINKED ISSUE #547: Freeze on Edge instantiation via BrowserViewer

### LINKED ISSUE #1801: "Unspecified error Waiting for Edge operation to terminate timed out"
Windows verification build failed on https://github.com/eclipse-platform/eclipse.platform/actions/runs/13118265699/job/36597841318?pr=1720
```
Error: Exception in thread "WorkbenchTestable" org.eclipse.swt.SWTException: Failed to execute runnable (org.eclipse.swt.SWTError: Unspecified error Waiting for Edge operation to terminate timed out)
	at org.eclipse.swt.SWT.error(SWT.java:4932)
	at org.eclipse.swt.SWT.error(SWT.java:4847)
	at org.eclipse.swt.widgets.Synchronizer.syncExec(Synchronizer.java:209)
	at org.eclipse.ui.internal.UISynchronizer.syncExec(UISynchronizer.java:133)
	at org.eclipse.swt.widgets.Display.syncExec(Display.java:4836)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Testable.testingFinished(E4Testable.java:157)
	at org.eclipse.tycho.surefire.osgibooter.AbstractUITestApplication.runTests(AbstractUITestApplication.java:57)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Testable.lambda$0(E4Testable.java:79)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: org.eclipse.swt.SWTError: Unspecified error Waiting for Edge operation to terminate timed out
	at org.eclipse.swt.SWT.error(SWT.java:4957)
	at org.eclipse.swt.browser.Edge.processOSMessagesUntil(Edge.java:474)
	at org.eclipse.swt.browser.Edge$WebViewProvider.getWebViewWrapper(Edge.java:370)
	at org.eclipse.swt.browser.Edge$WebViewProvider.getWebView(Edge.java:381)
	at org.eclipse.swt.browser.Edge.getUrl(Edge.java:893)
	at org.eclipse.swt.browser.Browser.getUrl(Browser.java:771)
	at org.eclipse.ui.internal.intro.impl.presentations.BrowserIntroPartImplementation.saveCurrentPage(BrowserIntroPartImplementation.java:645)
	at org.eclipse.ui.internal.intro.impl.model.AbstractIntroPartImplementation.saveState(AbstractIntroPartImplementation.java:302)
	at org.eclipse.ui.internal.intro.impl.model.IntroPartPresentation.saveState(IntroPartPresentation.java:427)
	at org.eclipse.ui.intro.config.CustomizableIntroPart.saveState(CustomizableIntroPart.java:405)
	at org.eclipse.ui.internal.ViewIntroAdapterPart.saveState(ViewIntroAdapterPart.java:198)
	at org.eclipse.ui.internal.ViewReference.persist(ViewReference.java:60)
	at org.eclipse.ui.internal.e4.compatibility.CompatibilityView.persistState(CompatibilityView.java:226)
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:56)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invokeUsingClass(InjectorImpl.java:299)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invoke(InjectorImpl.java:227)
	at org.eclipse.e4.core.contexts.ContextInjectionFactory.invoke(ContextInjectionFactory.java:148)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.safeRemoveGui(PartRenderingEngine.java:925)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$4.run(PartRenderingEngine.java:857)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.removeGui(PartRenderingEngine.java:841)
	at org.eclipse.e4.ui.workbench.renderers.swt.ElementReferenceRenderer.disposeWidget(ElementReferenceRenderer.java:113)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.safeRemoveGui(PartRenderingEngine.java:934)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$4.run(PartRenderingEngine.java:857)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.removeGui(PartRenderingEngine.java:841)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.safeRemoveGui(PartRenderingEngine.java:896)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$4.run(PartRenderingEngine.java:857)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.removeGui(PartRenderingEngine.java:841)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.safeRemoveGui(PartRenderingEngine.java:896)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$4.run(PartRenderingEngine.java:857)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.removeGui(PartRenderingEngine.java:841)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.safeRemoveGui(PartRenderingEngine.java:896)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$4.run(PartRenderingEngine.java:857)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.removeGui(PartRenderingEngine.java:841)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.stop(PartRenderingEngine.java:1219)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.close(E4Workbench.java:190)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Testable.lambda$3(E4Testable.java:157)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4077)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3693)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:668)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:576)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:178)
	at org.eclipse.tycho.surefire.osgibooter.UITestApplication.runApplication(UITestApplication.java:34)
	at org.eclipse.tycho.surefire.osgibooter.AbstractUITestApplication.run(AbstractUITestApplication.java:129)
	at org.eclipse.tycho.surefire.osgibooter.UITestApplication.start(UITestApplication.java:44)
Error: The operation was canceled.
```

amartya4256:
This PR adapts how KeyBindingDispatcher processed Key events in case of Browser. If the trigger of the event is a browser, it processes the command asynchronously to avoid any possible deadlocks.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/3104

**Problem Explained**
KeyBindings are registered globally with the Display. Inside the browser when a key combination is pressed which might open up another browser and use it synchronously, the browser needs to wait for the callback to finish and since it is not possible in case of Edge browser, it goes in a deadlock.

To achieve this, we must process the commands asynchronously. However, we need to tell the browser if the event eats the key event (sets event.doit = false) as the key is handled by SWT and not the browser natively. Hence, we need to proactively check if a command for the key event is going to be executed. If yes, it eats the key otherwise the browser can execute native action on the key event.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1801

### DIFF
--- a/bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
+++ b/bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
@@ -295,32 +295,42 @@ public class KeyBindingDispatcher {
 		final EHandlerService handlerService = getHandlerService();
 		final Command command = parameterizedCommand.getCommand();
 
-		final IEclipseContext staticContext = createContext(trigger);
-
 		final boolean commandDefined = command.isDefined();
 		// boolean commandEnabled;
 		boolean commandHandled = false;
 
+		// commandEnabled = handlerService.canExecute(parameterizedCommand,
+		// staticContext);
+		Object obj = HandlerServiceImpl.lookUpHandler(context, command.getId());
+		if (obj != null) {
+			if (obj instanceof IHandler handler) {
+				commandHandled = command.isEnabled() && handler.isHandled();
+			} else {
+				commandHandled = true;
+			}
+		}
+
+		if (isTracingEnabled()) {
+			logger.trace("Command " + parameterizedCommand + ", defined: " + commandDefined + ", handled: " //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+					+ commandHandled + " in " + describe(context)); //$NON-NLS-1$
+		}
+
+		if (trigger != null && trigger.widget instanceof Browser) {
+			getDisplay()
+					.asyncExec(() -> handleCommandExecution(parameterizedCommand, handlerService, trigger, obj));
+		} else {
+			handleCommandExecution(parameterizedCommand, handlerService, trigger, obj);
+		}
+		return (commandDefined && commandHandled);
+	}
+
+	private void handleCommandExecution(final ParameterizedCommand parameterizedCommand,
+			final EHandlerService handlerService, final Event trigger, Object handler) {
+		final IEclipseContext staticContext = createContext(trigger);
 		try {
-			// commandEnabled = handlerService.canExecute(parameterizedCommand, staticContext);
-			Object obj = HandlerServiceImpl.lookUpHandler(context, command.getId());
-			if (obj != null) {
-				if (obj instanceof IHandler) {
-					commandHandled = ((IHandler) obj).isHandled();
-				} else {
-					commandHandled = true;
-				}
-			}
-
-			if (isTracingEnabled()) {
-				logger.trace("Command " + parameterizedCommand + ", defined: " + commandDefined + ", handled: " //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-						+ commandHandled + " in " + describe(context)); //$NON-NLS-1$
-			}
-
 			handlerService.executeHandler(parameterizedCommand, staticContext);
 			final Object commandException = staticContext.get(HandlerServiceImpl.HANDLER_EXCEPTION);
 			if (commandException instanceof CommandException) {
-				commandHandled = false;
 				if (commandException instanceof ExecutionException) {
 					if (logger != null) {
 						logger.error((Throwable) commandException,
@@ -328,8 +338,7 @@ public class KeyBindingDispatcher {
 										+ describe(context));
 					}
 				} else if (isTracingEnabled()) {
-					logger.trace((Throwable) commandException,
-							"Command exception for: " + parameterizedCommand + " in " //$NON-NLS-1$ //$NON-NLS-2$
+					logger.trace((Throwable) commandException, "Command exception for: " + parameterizedCommand + " in " //$NON-NLS-1$ //$NON-NLS-2$
 									+ describe(context));
 					if (handlerService instanceof HandlerServiceImpl serviceImpl) {
 						IEclipseContext serviceContext = serviceImpl.getContext();
@@ -337,7 +346,7 @@ public class KeyBindingDispatcher {
 							StringBuilder sb = new StringBuilder("\n\tExecution context: "); //$NON-NLS-1$
 							sb.append(describe(serviceContext));
 							sb.append("\n\tHandler: "); //$NON-NLS-1$
-							sb.append(obj);
+							sb.append(handler);
 							logger.trace(sb.toString());
 						}
 					}
@@ -352,17 +361,16 @@ public class KeyBindingDispatcher {
 					}
 				}
 			}
-			/*
-			 * Now that the command has executed (and had the opportunity to use the remembered
-			 * state of the dialog), it is safe to delete that information.
-			 */
-			if (keyAssistDialog != null) {
-				keyAssistDialog.clearRememberedState();
-			}
 		} finally {
 			staticContext.dispose();
 		}
-		return (commandDefined && commandHandled);
+		/*
+		 * Now that the command has executed (and had the opportunity to use the
+		 * remembered state of the dialog), it is safe to delete that information.
+		 */
+		if (keyAssistDialog != null) {
+			keyAssistDialog.clearRememberedState();
+		}
 	}
 
 	private boolean isTracingEnabled() {
--- a/bundles/org.eclipse.e4.ui.bindings/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.bindings/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.ui.bindings;singleton:=true
-Bundle-Version: 0.15.0.qualifier
+Bundle-Version: 0.15.100.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin


bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
```diff
@@ -295,41 +295,58 @@ public final boolean executeCommand(final ParameterizedCommand parameterizedComm
...
+		} else {
+			handleCommandExecution(parameterizedCommand, handlerService, trigger, obj);
+		}
+		/*
+		 * Now that the command has executed (and had the opportunity to use the
+		 * remembered state of the dialog), it is safe to delete that information.
+		 */
+		if (keyAssistDialog != null) {
+			keyAssistDialog.clearRememberedState();
+		}
```
HeikoKlare: This may clear the state _before_  the command has been executed (in case the command is executed asynchronously). Is that intended? I am not sure what is meant by 
> had the oppportunity to use the remembered state of the dialog

and if clearing that state potentially too early might be wrong now.

Doing this inside the `handleCommandExecution` would also be wrong, as it can take arbitrary long to execute the command via asyncExec, so that a state may be cleared that belongs to a later multi-key assistance.

bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
```diff
@@ -295,41 +295,58 @@ public final boolean executeCommand(final ParameterizedCommand parameterizedComm
...
-					commandHandled = ((IHandler) obj).isHandled();
-				} else {
-					commandHandled = true;
-				}
+		// commandEnabled = handlerService.canExecute(parameterizedCommand,
+		// staticContext);
+		Object obj = HandlerServiceImpl.lookUpHandler(context, command.getId());
+		if (obj != null) {
+			if (obj instanceof IHandler) {
+				commandHandled = ((IHandler) obj).isHandled();
```
HeikoKlare: When refactoring this code anyway, it would be nice to apply instanceof pattern matching.

bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
```diff
@@ -295,41 +295,58 @@ public final boolean executeCommand(final ParameterizedCommand parameterizedComm
...
+		 * remembered state of the dialog), it is safe to delete that information.
+		 */
+		if (keyAssistDialog != null) {
+			keyAssistDialog.clearRememberedState();
+		}
+		return (commandDefined && commandHandled);
+	}
+
+	private void handleCommandExecution(final ParameterizedCommand parameterizedCommand,
+			final EHandlerService handlerService, Event trigger, Object obj) {
```
HeikoKlare: Can you use a more specific parameter name than `obj`?

HeikoKlare: Why are two out of three parameters final?
And couldn't you pass a `commandExecutor` that incorporates the `handlerService` and `trigger`/`staticContext` to avoid so many parameters? I.e., passing a lambda `command -> handlerService.executeHandler(command, staticContext)` instead of `handlerService` and `trigger` separately.

HeikoKlare (commented):
At a first glance, this looks like a reasonable workaround according to the proposal in https://github.com/eclipse-platform/eclipse.platform.ui/issues/3104#issuecomment-3061931734.

One concern regarding behavior is the clearance of the `keyAssistDialog` state. And note that a bunch of related tests is failing now.
My other current comments are rather regarding style.

bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
amartya4256: we need to dispose the context after the command has been processed. That's why I passed the event and created the context at the time of execution. Can also pass staticContext instead of trigger.

bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
amartya4256:
```
public void clearRememberedState() {
		previousWidth = NO_REMEMBERED_WIDTH;
	}
```

This just resets the previousRememberedState and since it can execute asynchronously, we want it to always have the previous width. Even though a task executed before it, if the width was changed and memorized, we will use that and reset the width asynchronously.

Hence, it is okay to move this block inside the method `handleCommandExecution`

bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
HeikoKlare: I see. Then it's probably best to keep the `trigger` parameter as it's otherwise confusing that you pass over the lifecycle handling for `staticContext` from the caller to the callee. Instead, the caller would have to dispose it.

amartya4256: Implemented

bundles/org.eclipse.e4.ui.bindings/src/org/eclipse/e4/ui/bindings/keys/KeyBindingDispatcher.java
```diff
@@ -295,49 +295,58 @@ public final boolean executeCommand(final ParameterizedCommand parameterizedComm
...
-					commandHandled = ((IHandler) obj).isHandled();
-				} else {
-					commandHandled = true;
-				}
+		// commandEnabled = handlerService.canExecute(parameterizedCommand,
+		// staticContext);
+		Object obj = HandlerServiceImpl.lookUpHandler(context, command.getId());
+		if (obj != null) {
+			if (obj instanceof IHandler handler) {
+				commandHandled = command.isEnabled() && handler.isHandled();
```
HeikoKlare, line 307: Why is the `command.isEnabled()` check added here?

amartya4256: if the command is not enabled, the code simply registers an execption and proceeds. Later this leads to commandHandled being set to false. Note that this needed to decide if the current handler eats the key. If the eatKey is false then the next handler in the notifyListener handles this event. e.g. CTRL + A can have multiple handler and if the first handler is not enabled and the second one should perform the command. The first one should set the commandHandled to false and not eatKey so that it would be propagated to the second handler. Since we can check this proactively before `handlerService#executeHandler`, we leverage this to already check if it can execute. Similar check is also performed inside `Command#execute`.

HeikoKlare: Great, thank you for the detailed explanation!

HeikoKlare (approved):
The change now looks good to me. It also properly resolves the issue reported in https://github.com/eclipse-platform/eclipse.platform.ui/issues/3104 for me.

@sratz can you please also have a look as you have been involved into the issue and made proposals for how to fit it as well?

HeikoKlare:
Since we are at the very beginning of a new development cycle, it would be a good point in time to merge a change like this. @sratz could you please have a look at this fix as well as it's Edge-related?

HeikoKlare:
I am in favor of merging this as early as possible for M1, so that we have a long period of testing before 2026-06 release. I will not be available until next Tuesday, thus I will not merge before that day as I would not be able to react to potential regressions. I plan to merge next Tuesday in case there are no objections by then.

sratz:
> The change now looks good to me. It also properly resolves the issue reported in #3104 for me.
> 
> @sratz can you please also have a look as you have been involved into the issue and made proposals for how to fit it as well?

Sorry for the delay, this somehow slipped my attention.

The changes look good, also in favor of getting this into M1 👍

sratz (approved):

vogella:
@sratz please merge in this case.

PR #3656, opened 2026-01-10
Improve CTabFolder highlight appearance on higher zooms

### LINKED ISSUE #517: CTabFolder highlight not positioned/sized correctly on zoom != 100%

HeikoKlare:
The highlight of CTabFolders implemented in CTabRendering does not properly deal with higher zooms on Windows. The outline of the tabs is drawn with a width of one pixel whereas the highlight has an offset of one point, producing a gap at higher zooms. In addition, the highlight for rounded tabs exceeds the intended bounds.

This change adapts the rendering as follows:
- When the highlight is drawn rounded at the top, the shape of the outline is taken and adapted to the intended highlight height to draw the highlight.
- In all other cases the highlight is drawn as a simple line of according width with an offset of half the line width with respect to its bounds. To compensate for the 1px border that is drawn on top of the highlight when it is placed at the top of a tab, another line of according size is drawn next to the highlight to make its height fit again.

One problem that is not solved by this are inconsistent heights of the highlights when using a CTabFolder with headers being placed at the bottom at 125% and 175%. The root cause is an already misplaced/miscalculated header region, at it exceeds the intended size to the bottom. So in order to improve that behavior, the header region calculation has to be fixed. 

This is a proposal to overcome the current limitations of:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3533

### Windows

CTabFolders with headers on top:
highlights_top

CTabFolders with headers at top:
highlights_bottom

### Linux
150% used fractional scaling mode, which is why the screenshot has the same size as 200% (the OS scales down a faked 200% canvas):
highlights_linux

### MacOS
highlights_macos

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java
@@ -18,6 +18,7 @@
 package org.eclipse.e4.ui.workbench.renderers.swt;
 
 import java.lang.reflect.Field;
+import java.util.Arrays;
 import java.util.Objects;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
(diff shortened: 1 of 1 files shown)


ShahzaibIbrahim (commented):
At 100%, there is a behavior change from previous implementation: 

The bottom highlight has a gap from bottom, the same issue that was mentioned here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3533#pullrequestreview-3635453709

Left: Current implementation
Right: Your change.

Image


At 250%: I see highlight color bleeding out to left tab when hovered. See, (its a gif, please open it in other tab to see it running)

HeikoKlare:
> At 100%, there is a behavior change from previous implementation:
> 
> The bottom highlight has a gap from bottom, the same issue that was mentioned here: [#3533 (review)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3533#pullrequestreview-3635453709)

It's interesting that I did not have that when testing (as can be seen in my screenshots). I don't understand why, as I can also reproduce it.

@ShahzaibIbrahim can you maybe update and simplify your PR to support the `onBottom` case based on insights from this PR? Maybe just adding the -1 back would already solve the placement issue you mentioned.

HeikoKlare:
Closing this as superceded by:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3533

Issue #3570, opened 2025-12-10
[Dark mode] Tree widgets have a hard to see black foreground in dark mode


Andr2ea:
Description

Tree widgets across multiple Eclipse views do not properly render selection highlighting in dark theme mode. This issue affects various Eclipse trees including the Preference Dialog tree, Import wizards, Breakpoints view, and other views - but notably NOT the Project Explorer which renders correctly. When items are selected in these affected trees, the selection background and foreground colors do not adapt to dark theme, resulting in poor visibility and contrast issues. A custom workaround was to manually override the SWT.EraseItem event and repaint selections with appropriate colors for dark themes. This issue also happens in S32Ds.

To Reproduce

Enable dark theme in Eclipse (Window -> Preferences -> General -> Appearance -> Theme -> Select "Dark")
Open any of the following:
Preferences dialog (Window -> Preferences) - click items in the left tree
Import wizard (File -> Import) - click items in the import source tree
Breakpoints view (Window -> Show View -> Debug -> Breakpoints) - select breakpoints in the tree
Click on any item in these "Tree" widgets
Observe that selected items have poor contrast/visibility with incorrect selection colors that don't match the dark theme
Compare with Project Explorer tree selection which renders correctly with proper dark theme colors

Expected behavior

All tree widget selections should automatically adapt to the current theme consistently across Eclipse. In dark theme:

Selected items should have a clearly visible selection background (similar to Project Explorer's selection rendering)
Selected text should be white or light colored for proper contrast
The selection rendering should be consistent across ALL tree widgets in Eclipse, not just Project Explorer

Screenshots

<img width="178" height="661" alt="Image" src="https://github.com/user-attachments/assets/2fee68de-8daf-4efb-b314-8831ff4d8257" />

<img width="686" height="525" alt="Image" src="https://github.com/user-attachments/assets/be56b746-9b84-45b2-a858-6742a7690921" />

<img width="446" height="406" alt="Image" src="https://github.com/user-attachments/assets/a0c12b71-cbf7-45f6-8fc4-3dc3b9dbdc38" />

<img width="682" height="514" alt="Image" src="https://github.com/user-attachments/assets/ea732072-c6e8-4c78-afd3-4d844d9de05a" />

Image Image Image Image
Version Information:

OS and OS Version/extra details: Windows 10
Eclipse Version 2023-12

HeikoKlare:
This is related to:
[...]

Issue #2780, opened 2025-02-07
[Dark theme] In explorer, highlight color is similar to the normal color


haoting-chen:
Eclipse Platform 4.34.0
In explorer, the highlight color is similar to the normal color, making highlighted items difficult to see.
Below is an example, first item is highlighted one.
![Image](https://github.com/user-attachments/assets/2e9aedde-0b1a-43c5-bd30-5d4ecc85611e)

HeikoKlare:
Related to:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/811
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1690

PR #3659, opened 2026-01-12
Proper selection coloring in JFace Structured Viewers

Christopher-Hermann:
Resumption of #1690

Highlighting problem when using the dark theme on Windows https://github.com/eclipse-platform/eclipse.platform.swt/issues/811

JFace viewer are using OS selection color to highlight the selected item. On some OS this is not accessible. In particular when the eclipse is used with dark theme and the OS is used with light theme, this can cause really bad UI/UX experience. With this change, the selection color can be changed via color preference in the settings of eclipse.

**Example**
Selection color on Windows Server in dark theme looks really bad:
272533915-9b63234a-6644-4174-9f2a-42b610eb5d66

**Testing**
On start up of eclipse, the selection colors in column viewers (table and tree viewer) should look like before.
On Windows, the dark theme should predefine the selection colors to fix the bad coloring on some Windows platforms.

In the preferences (Preferences > General > Appearance > Colors and Fonts), there are 4 new properties for the coloring of the viewer selection:

- Basic > Viewer selection background (focused)
- Basic > Viewer selection background (no focus)
- Basic > Viewer selection foreground (focused)
- Basic > Viewer selection foreground (no focus)
Changing these colors should affect the coloring of the selection in viewers. The coloring should immediately be applied as soon as the apply button is confirmed. The reset button should reset the coloring to the OS specific highlight coloring, expect for Windows dark theme, where the predefined colors should be restored.

**OS Information**
The fix is implemented for Windows and Mac only (no GTK).
On GTK the coloring via `SWT.EaraseItem` is not working properly.

Fixes:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/811
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1688
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1956
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2780
https://github.com/eclipse-platform/eclipse.platform.ui/issues/3570

See:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1690
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2793


bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,253 @@
...
+		Listener[] listeners = control.getListeners(SWT.EraseItem);
+
+		// Get the OwnerDrawListener if it exists
+		Object ownerDrawListener = control.getData(OWNER_DRAW_LISTENER_KEY);
+
+		// Count non-framework listeners (excluding this listener and OwnerDrawListener)
+		int customListenerCount = 0;
+		for (Listener listener : listeners) {
+			if (listener != this && listener != ownerDrawListener) {
+				customListenerCount++;
```
merks: You could return immediately without counting all of them...

laeubi:
@Christopher-Hermann please use descriptive title and PR description and also adapt the commit message. It is impossible to understand without looking at the code what is actually done here.

Beside that, its always good to show before/after screenshot for visual changes.

Christopher-Hermann:
Added PR title & description. In the current state there are most probably some things left that needs to be cleaned up, tests are missing, etc.

But in the first place I wanted to share the coding with @fedejeanne to get the fix running on all operating systems.
In the last try there were a lot of issue especially for Linux that we need to solve...

fedejeanne:
@Christopher-Hermann thank you for retrying this!

As agreed, I tested on Linux and I found some of the problems that were already mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1956, namely:

1. The **Viewer selection foreground (no focus)** does not apply in the _package explorer_ (`MANIFEST.MF`) and in the _Preferences_ (`Colors and Fonts` item)

Screenshot from 2026-01-13 13-45-24

This situation is worse with the default values since it's then white over white for non-focused elements (`MANIFEST.MF` in the _package explorer_)

image

---

2. Icons go missing, both for the focused (orange) and the non-focused (light blue) elements

Screenshot from 2026-01-13 13-56-43

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,253 @@
...
+ * <p>
+ * Colors are managed through the JFace {@link ColorRegistry} and can be
+ * customized by putting new values into the registry with the following keys:
+ * </p>
+ *
+ * @see org.eclipse.jface.viewers.OwnerDrawLabelProvider
+ * @see org.eclipse.jface.viewers.StyledCellLabelProvider
+ * @see org.eclipse.jface.viewers.FocusCellOwnerDrawHighlighter
+ */
+public class ColumnViewerSelectionColorListener implements Listener {
```
fedejeanne, line 53:
```suggestion
 * @see org.eclipse.jface.viewers.FocusCellOwnerDrawHighlighter
 * @since 3.39
 */
public class ColumnViewerSelectionColorListener implements Listener {
```

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColumnViewerSelectionColorFactory.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Color factory for viewer selection colors that adapts to the OS/desktop theme.
+ * Provides default colors based on system colors for focused and unfocused selections.
+ * <p>
+ * The default colors are based on system title bar colors which automatically adapt
+ * to light/dark themes and high contrast modes. Themes can override these defaults
+ * to provide custom styling.
+ * </p>
+ *
+ * @since 3.32
```
fedejeanne:
```suggestion
 * @since 3.39
```

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/StructuredViewer.java
```diff
@@ -1210,6 +1210,7 @@ public void widgetDefaultSelected(SelectionEvent e) {
 		});
 		handler.addPostSelectionListener(widgetSelectedAdapter(this::handlePostSelect));
 		handler.addOpenListener(StructuredViewer.this::handleOpen);
+		ColumnViewerSelectionColorListener.addListenerToViewer(this);
```
fedejeanne: This line is causing the missing icons (problem Nr. 2 in my review)

Christopher-Hermann: Mh damn, no clue why this is happening on Linux :(
This line is needed to get the `ColumnViewerSelectionColorListener` linked to the viewer. This listener is handling the drawing of the selection colors, so this is needed.

A really bad option would be to register the listener only on Windows. This would at least solve the issue on Windows and I guess on Linux we don't have the issue.

But I will have a look if I can figure out why the icons are missing on Linux

Christopher-Hermann: @fedejeanne I adjusted the bounds of the back ground for better debugging. The background is now drawn with a offset of some pixel.
Maybe the `fillBackground` is drawing over the icons.
In case you have time, could you post a new screenshot?

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -181,11 +181,20 @@ public static void drawSelection(Event event, org.eclipse.swt.graphics.Device de
...
 		// Fill selection rectangle
-		gc.fillRectangle(0, event.y, width, event.height);
+		// Use event.x and event.width to respect the bounds provided by SWT,
+		// which on some platforms (like GTK/Linux) excludes the image area
+		gc.fillRectangle(x, event.y, width, event.height);
 
 		// Remove SELECTED and BACKGROUND flags to prevent native drawing from overwriting our custom colors
-		event.detail &= ~(SWT.SELECTED | SWT.BACKGROUND);
+		// Note: On Linux/GTK, we must keep the SELECTED flag so that icons are still drawn
+		// We only remove BACKGROUND to prevent the native background from overwriting our custom colors
```
fedejeanne: The comment is confusing, you are removing `SELECTED` in the line below

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -181,11 +181,20 @@ public static void drawSelection(Event event, org.eclipse.swt.graphics.Device de
 			width = scrollable.getClientArea().width;
 		}
 
+		int x = 0;
+		if (width > 50) {
+			x = 50;
+		}
```
fedejeanne: Why 50? Magic number?

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
fedejeanne: ... or even rewrite it to use a more declarative idiom like:

```java
return Arrays.stream(listeners).anyMatch(l -> l != this && l != ownerDrawListener);
```

fedejeanne (commented):
@Christopher-Hermann I now have some other odd behavior: I see half painted selections :-\

_EDIT: I see the icons though, so that's good :-)_

Here I changed the _Viewer selection background (focused)_ from orange to green and  _Viewer selection background (no focus)_ to yellow and you can see that the selections still have a half-orange part, both the focused and the non-focused ones.

Image

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
fedejeanne: Ach,silly me, it was right in the commit-text: `Debugging: Draw background with offset`. So yes, it's a magic "debug" number.

btw. I played with it and reducing it to 15 produces a half painted selection without icon so this is the problematic code spot:

image

Christopher-Hermann: Yes, was for debugging removed it again. See my comment for information about the current/final state.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
Christopher-Hermann: Comment updated

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColumnViewerSelectionColorFactory.java
Christopher-Hermann: done

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
Christopher-Hermann: done

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
Christopher-Hermann: Done

Christopher-Hermann:
I tried a lot on a Linux VM and followed the Eclipse article where the `SWT.EaraseItem` is described as: "SWT.EraseItem: allows a client to custom draw a cell’s background and/or selection, and to influence whether the cell’s foreground should be drawn"
But this does not appear to work correctly on Linux (GTK) environments. This leads me to believe there might be a bug in its implementation for this platform.

I’ve implemented a fix/option that influences selection colors, but explicitly only for Windows and macOS, effectively skipping GTK.
I know this is not nice, but according to the amount of issues (all on windows) I guess this solution is better than doing nothing.

Any opinions?

fedejeanne:
@Christopher-Hermann thank you for looking into it and for going into so much trouble for Linux too! I have to say my opinion is somewhat partial since I only develop on Eclipse on Linux in my own free time (we use Windows at work, where my colleagues periodically remind me about this very annoying issue). That being said: I'd be OK to compromise and skip this feature for Linux.

Maybe @akurtakov / @jonahgraham / @iloveeclipse can chip in here since they work on Linux?

@Christopher-Hermann in case your proposal is accepted and you deliver the functionality for Windows and Mac only, you would have to remove the new properties under `General > Appearance > Colors and Fonts` for Linux though since they don't apply

image

Christopher-Hermann:
> @Christopher-Hermann in case your proposal is accepted and you deliver the functionality for Windows and Mac only, you would have to remove the new properties under `General > Appearance > Colors and Fonts` for Linux though since they don't apply

I tried to do this, but I didn't find a way to define a color definition only for Windows and Mac, or to hide it on Linux. So if someone has an idea how this could be done please let me know.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColumnViewerSelectionColorFactory.java
```diff
@@ -0,0 +1,67 @@
...
+		} else if ("SELECTED_CELL_BACKGROUND_NO_FOCUS".equals(color)) { //$NON-NLS-1$
+			return display.getSystemColor(SWT.COLOR_TITLE_INACTIVE_BACKGROUND).getRGB();
+
+		} else if ("SELECTED_CELL_FOREGROUND_NO_FOCUS".equals(color)) { //$NON-NLS-1$
+			return display.getSystemColor(SWT.COLOR_TITLE_INACTIVE_FOREGROUND).getRGB();
+
+		} else {
+			return new RGB(0, 0, 0);
+		}
+	}
```
fedejeanne, line 41: How about something more concise like ...
```suggestion
	   @Override
        public RGB createColor() {
                return Display.getDefault().getSystemColor(idToColor(color)).getRGB();
        }

        private int idToColor(String id) {
                return switch (id) {
                case "SELECTED_CELL_BACKGROUND" -> SWT.COLOR_TITLE_BACKGROUND; //$NON-NLS-1$
                case "SELECTED_CELL_FOREGROUND" -> SWT.COLOR_WHITE; //$NON-NLS-1$
                case "SELECTED_CELL_BACKGROUND_NO_FOCUS" -> SWT.COLOR_TITLE_INACTIVE_BACKGROUND; //$NON-NLS-1$
                case "SELECTED_CELL_FOREGROUND_NO_FOCUS" -> SWT.COLOR_TITLE_INACTIVE_FOREGROUND; //$NON-NLS-1$
                default -> SWT.COLOR_BLACK;
                };
        }
```

?

With one change: `default` (or `null`) now yield `SWT.COLOR_BLACK`, which should also work as expected **(please double check this)**.

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,199 @@
...
+		Color backgroundColor;
+		Color foregroundColor;
+
+		if (control.isFocusControl()) {
+			backgroundColor = getSelectionColor(COLOR_SELECTION_BG_FOCUS, event.display);
+			foregroundColor = getSelectionColor(COLOR_SELECTION_FG_FOCUS, event.display);
+		} else {
+			backgroundColor = getSelectionColor(COLOR_SELECTION_BG_NO_FOCUS, event.display);
+			foregroundColor = getSelectionColor(COLOR_SELECTION_FG_NO_FOCUS, event.display);
+		}
```
fedejeanne:
```suggestion
		final Color backgroundColor = getSelectionColor(
                                control.isFocusControl() ? COLOR_SELECTION_BG_FOCUS : COLOR_SELECTION_BG_NO_FOCUS, event.display);
                final Color foregroundColor = getSelectionColor(
                                control.isFocusControl() ? COLOR_SELECTION_FG_FOCUS : COLOR_SELECTION_FG_NO_FOCUS, event.display);
```

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,199 @@
...
+			systemColor = device.getSystemColor(SWT.COLOR_TITLE_INACTIVE_FOREGROUND).getRGB();
+			break;
+		default:
+			systemColor = device.getSystemColor(SWT.COLOR_LIST_SELECTION).getRGB();
+			break;
+		}
+
+		registry.put(key, systemColor);
+		return registry.get(key);
+	}
```
fedejeanne:
```suggestion
	private static Color getSelectionColor(String key, org.eclipse.swt.graphics.Device device) {
                ColorRegistry registry = JFaceResources.getColorRegistry();

                if (!registry.hasValueFor(key)) {
                        RGB systemColor = device.getSystemColor(idToColor(key)).getRGB();
                        registry.put(key, systemColor);
                }

                return registry.get(key);
        }

        private static int idToColor(String id) {
                return switch (id) {
                case COLOR_SELECTION_BG_FOCUS -> SWT.COLOR_TITLE_BACKGROUND;
                case COLOR_SELECTION_FG_FOCUS -> SWT.COLOR_WHITE;
                case COLOR_SELECTION_BG_NO_FOCUS -> SWT.COLOR_TITLE_INACTIVE_BACKGROUND;
                case COLOR_SELECTION_FG_NO_FOCUS -> SWT.COLOR_TITLE_INACTIVE_FOREGROUND;
                default -> SWT.COLOR_LIST_SELECTION;
                };
        }
```

fedejeanne (commented):
I'm looking into this again and trying to find the cause for the misbehavior in Linux, no luck.

I do have some minor comments regarding the writing style though.

Christopher-Hermann:
> I'm looking into this again and trying to find the cause for the misbehavior in Linux, no luck.
> 
> I do have some minor comments regarding the writing style though.

Thanks for checking it again and for your comments. I did not have time to look into it yet, hopefully I find some time the next days 👍

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColumnViewerSelectionColorFactory.java
```diff
@@ -0,0 +1,67 @@
...
+
+		} else if ("SELECTED_CELL_FOREGROUND".equals(color)) { //$NON-NLS-1$
+			return display.getSystemColor(SWT.COLOR_WHITE).getRGB();
+
+		} else if ("SELECTED_CELL_BACKGROUND_NO_FOCUS".equals(color)) { //$NON-NLS-1$
+			return display.getSystemColor(SWT.COLOR_TITLE_INACTIVE_BACKGROUND).getRGB();
+
+		} else if ("SELECTED_CELL_FOREGROUND_NO_FOCUS".equals(color)) { //$NON-NLS-1$
+			return display.getSystemColor(SWT.COLOR_TITLE_INACTIVE_FOREGROUND).getRGB();
+
```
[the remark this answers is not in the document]
fedejeanne: Done, it's now inside `org.eclipse.jface.viewers.ColumnViewerSelectionColorListener.idToDefaultSWTSystemColor(String)`

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,199 @@
...
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Listener;
+import org.eclipse.swt.widgets.Scrollable;
+
+/**
+ * EraseItem event listener that provides custom selection coloring for JFace
+ * viewers. This listener only activates when no custom owner draw label
+ * provider is registered, ensuring it doesn't conflict with existing custom
+ * drawing implementations.
+ * <p>
```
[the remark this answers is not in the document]
fedejeanne: @Christopher-Hermann we need to revisit this. Originally, this class was only listening to `SWT.EraseItem` in the past, which is (partly) why the icons were missing in Linux. It now listens to `SWT.EraseItem` and `SWT.PaintItem` and it separates the responsibilities properly i.e. when handling `SWT.EraseItem`, it merely draws the background (deleting the text and the icons) and when handling `SWT.PainItem` it draws the icon and the text.

So this raises 2 questions for me:

1. Do we still need to have the logic to avoid installing this listener when `isListenerRegistered(control)` is `true`?
2. Is it correct to skip the "erasing" part when there are other listeners that do that? (see usage of `hasAdditionalEraseItemListeners(event)` in `handleEraseItem(event)`)

WDYT?

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,199 @@
...
+	 */
+	private boolean hasAdditionalEraseItemListeners(Event event) {
+		if (!(event.widget instanceof Control control)) {
+			return false;
+		}
+
+		Listener[] listeners = control.getListeners(SWT.EraseItem);
+		Object ownerDrawListener = control.getData(OWNER_DRAW_LISTENER_KEY);
+		return Arrays.stream(listeners).anyMatch(l -> l != this && l != ownerDrawListener);
+	}
```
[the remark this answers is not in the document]
fedejeanne: Fixed, it now skips `this` and everything before it and searches for something other than the `OwnerDrawListener`.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColumnViewerSelectionColorFactory.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @since 3.39
+ */
+public class ColumnViewerSelectionColorFactory implements IColorFactory, IExecutableExtension {
+
+	private String color = null;
+
+	@Override
+	public RGB createColor() {
+		Display display = Display.getDefault();
```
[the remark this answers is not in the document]
fedejeanne: I'd be fine with using `getCurrent()` too. WDYT @Christopher-Hermann ?

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,199 @@
...
+		if (event.widget instanceof Scrollable scrollable) {
+			width = scrollable.getClientArea().width;
+		}
+
+		gc.fillRectangle(0, event.y, width, event.height);
+
+		// Remove SELECTED and BACKGROUND flags to prevent native drawing from
+		// overwriting our custom colors
+		event.detail &= ~(SWT.SELECTED | SWT.BACKGROUND);
+	}
```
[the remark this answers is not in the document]
fedejeanne: Not present anymore

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
```diff
@@ -0,0 +1,199 @@
...
+		int width = event.width;
+		if (event.widget instanceof Scrollable scrollable) {
+			width = scrollable.getClientArea().width;
+		}
+
+		gc.fillRectangle(0, event.y, width, event.height);
+
+		// Remove SELECTED and BACKGROUND flags to prevent native drawing from
+		// overwriting our custom colors
+		event.detail &= ~(SWT.SELECTED | SWT.BACKGROUND);
```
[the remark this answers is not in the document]
fedejeanne: Unnecessary

bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/TableOwnerDrawSupport.java
```diff
@@ -83,7 +85,7 @@ public void handleEvent(Event event) {
 				measureItem(event);
 				break;
 			case SWT.EraseItem:
-				event.detail &= ~SWT.FOREGROUND;
+				event.detail&= ~SWT.FOREGROUND;
```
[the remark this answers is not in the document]
fedejeanne: Fixed

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColumnViewerSelectionColorFactory.java
fedejeanne: Fixed and also I moved the whole logic to `org.eclipse.jface.viewers.ColumnViewerSelectionColorListener`. Notice that black is not the fallback anymore, I use `SWT.COLOR_LIST_SELECTION` instead (which is a system color). Its JavaDoc says

```
System color used to paint list selection background areas (value is 26).
```

... which seems to me that it is so arbitrary as `new RGB(0, 0, 0)` but it doesn't create any new instances.

fedejeanne:
Aside from the failing checks, this PR is ready to be reviewed. 

@Christopher-Hermann I made lots of changes that still need to be squashed. I thought it would be easier for you to review if there are several smaller, more descriptive, changes than only 1 big commit. I tried it on Windows and Linux and it works better than before, at least no icons go missing in Linux.

There is still one thing that doesn't really work as I expect it to: the **foreground** color in the preferences does not affect views like the package explorer, the outline, (quick) hierarchy view and others. It also doesn't affect lists inside the manifest editor. For what I saw, it only affects the code-completion popup, which is achieved by `CompletionProposalDrawSupport` (adapted from #2793). Do you have any idea where could we tweak the code so the missing places also honor the foreground color set in the preferences?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColumnViewerSelectionColorFactory.java
Christopher-Hermann: Yes looks good 👍

bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewerSelectionColorListener.java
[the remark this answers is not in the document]
Christopher-Hermann: I find it kind of hard to get a proper overview about all the table drawing that is available in the eclipse stack. 

1. I tried to be as minimal inversive as possible. I'm not sure what happens if there are other listeners registered. I retested with your changes and at least all the tests seems to be okay. But it's hard to tell if there are any other use cases where we "break" coloring. With the actual coding it breaks for example the "Abstract Syntax Tree"

2. Yes, I guess it is correct to skip if there are other listeners. We only want to change colors for viewers where no custom coloring is done

Christopher-Hermann (commented):
With the current state it breaks the Abstract Syntax Tree:

Image

Left side without the change, right side with the change (same type is selected)
To retest you can run the command `Show in (Abstract Syntax Tree)`

Next finding:
Foreground colors are not considered anymore on macOS:

Image

I'm not sure how we can proceed here, I already spend a lot of time into this topic and it feels like there is no proper solution working on all platforms for this problem :(

PR #3663, opened 2026-01-12
Add ConvertLineDelemiterTest to FileBuffersTestSuite

vogella:
Adds `ConvertLineDelemiterTest` to `FileBuffersTestSuite` so it is run as part of the test suite.


vogella:
I assume ConvertLineDelemiterTest was just forgotten to add to the test suite, works fine if I run it locally. Lets see that CI thinks

PR #3666, opened 2026-01-12
Add missing descriptions to Colours and Fonts theme entries in Preferences

elsazac:
**Consistent use of descriptions in Colours and Fonts preferences**

Some entries in the Colours and Fonts preference page already provide descriptions, while others leave the description area empty. This change makes use of the existing descriptions and applies them to theme entries so that the description box is consistently populated and available descriptions are properly utilised.This improves clarity and aligns the presentation with the approach used with the subtree elements in Eclipse preference pages and gives more clarity on what each entity is used for.

The Colours and Fonts preference page shows a description panel which is empty now. There are already descriptions available for some and not available for rest and they are not being used so as to display in the box. This PR make use of the description. This change adds the existing description usage to theme entries so that users always see meaningful information when navigating these settings.

for eg: `CSS bridge demo` theme and `view and editor folders` have descriptions available but it isn't displayed in the box; This PR makes changes to utilise the description available

before : 
Screenshot 2026-01-09 at 8 02 23 PM

after : 
Screenshot 2026-01-09 at 7 35 17 PM

similarly, with entities `Basic` and `View and Editor folders`

Screenshot 2026-01-09 at 7 35 08 PM
Screenshot 2026-01-09 at 7 35 32 PM

There are few more entries coming from `eclipse.platform` that needs similar updates. Once this change goes in, I plan to follow up on those as well.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
@@ -1959,6 +1959,17 @@ public final class ColorsAndFontsPreferencePage extends PreferencePage implement
 		}
 	}
 
+	private ThemeElementCategory getCategory() {
+		IStructuredSelection item = (IStructuredSelection) tree.getViewer().getSelection();
+		if (item != null && !item.isEmpty()) {
+			Object object = item.getFirstElement();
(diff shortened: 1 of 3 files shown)


deepika-u:
Good one, thanks for improving UI experience with such small improvements.

elsazac:
@vogella would you be able to review and approve this? and I would like to merge this if there are no outstanding concerns on this.

vogella:
@BeckerWdf can you review?

vogella:
Thanks @elsazac

PR #3670, opened 2026-01-13
Fix regression in ProjectionViewer.setVisibleRegion for empty ranges

vogella:
Commit 59dbd43a19dd introduced a regression where setVisibleRegion(0, 0) would expand to include the full line, causing SegmentedModeTest.testShowNothing to fail. This change ensures that empty ranges (length 0) are not expanded to the end of the line, restoring the previous behavior for this case while keeping the fix for partial lines (length > 0).

### DIFF
--- a/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
+++ b/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
@@ -846,6 +846,9 @@ public class ProjectionViewer extends SourceViewer implements ITextViewerExtensi
 	private static int computeEndOfVisibleRegion(int start, int length, IDocument document) throws BadLocationException {
 		int documentLength= document.getLength();
 		int end= start + length;
+		if (length == 0) {
+			return end + 1;
+		}
 		// ensure that the last line is fully included because projections cannot include partial lines
 		while (end < documentLength && !isLineBreak(document.getChar(end))) {
 			end++;


vogella:
@danthe1st WDYT?

danthe1st:
> @danthe1st WDYT?

I think it looks good as a bugfix in general but I think it would make sense to do more.
If the selected end is directly after the line break, I guess it doesn't really make sense to include the last line as well so I guess something like the following might make more sense:
```java
		if (length == 0 || isLineBreak(document.getChar(end - 1))) {
			return end + 1;
		}
```

That could be tested as follows (the test in this example is parameterized to show that the result is the same both with and without projections):
```java
	@ParameterizedTest
	@CsvSource({ "true", "false" })
	void testEndWithStartOfLine(boolean enableProjections) {
		Shell shell= new Shell();
		shell.setLayout(new FillLayout());
		TestProjectionViewer viewer= new TestProjectionViewer(shell, null, null, false, SWT.NONE);
		String documentContent= """
				Hello
				World
				123
				456
				""";
		Document document= new Document(documentContent);
		viewer.setDocument(document, new AnnotationModel());
		if (enableProjections) {
			viewer.enableProjection();
		}

		viewer.setVisibleRegion(documentContent.indexOf("World"), documentContent.indexOf("123") - documentContent.indexOf("World"));
		shell.setVisible(true);
		try {
			assertEquals("World\n", viewer.getVisibleDocument().get());
		} finally {
			shell.dispose();
		}
	}
```

I also ran some relevant tests I know in other projects (`org.eclipse.xtend.ide.tests.autoedit.AutoEditTest` in xtext, `FoldingTestSuite` in JDT-UI, `DocumentLinkReconcilingTest` in lsp4e) and at least they don't seem to break either.

vogella:
@danthe1st can you push a new PR for your suggestion (or update mine, not sure if that will work)?

danthe1st:
> @danthe1st can you push a new PR for your suggestion (or update mine, not sure if that will work)?

I can't update your PR but I created #3671

PR #3671, opened 2026-01-13
Fix regression in ProjectionViewer.setVisibleRegion for empty ranges

### LINKED ISSUE #3667: SegmentedModeTest.testShowNothing fails since I20260112-1800
See https://download.eclipse.org/eclipse/downloads/drops4/I20260112-1800/testresults/html/org.eclipse.ui.editors.tests_ep439I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

```
org.eclipse.ui.editors.tests.SegmentedModeTest	testShowNothing	Failure	expected:<[]> but was:<[this ]>

org.junit.ComparisonFailure: expected:<[]> but was:<[this
]>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.ui.editors.tests.SegmentedModeTest.testShowNothing(SegmentedModeTest.java:128)
```

This is a regression from 59dbd43a19dd75fb1f9caf6262294d2f35b1a682 / https://github.com/eclipse-platform/eclipse.platform.ui/pull/3585.

Note, the test fails locally only if launch config contains most of SDK plugins (via "Restore defaults" button). Executed with "Select required" plugins it passes (which explains why it works on jenkins and doesn't work in "official" SDK builds with all SDK bundles).

danthe1st:
Commit 59dbd43a19dd introduced a regression where `setVisibleRegion(0, 0)` would expand to include the full line, causing `SegmentedModeTest.testShowNothing` to fail. This change ensures that empty ranges (length 0) are not expanded to the end of the line, restoring the previous behavior for this case while keeping the fix for partial lines (length > 0).

In addition to that, this PR prevents the last line from being included/shown when the end character is directly after the line break.

This is an extended version of #3670 (I also copied most of the description from there)
Fixes #3667

### DIFF
--- a/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
+++ b/bundles/org.eclipse.jface.text/projection/org/eclipse/jface/text/source/projection/ProjectionViewer.java
@@ -846,6 +846,9 @@ public class ProjectionViewer extends SourceViewer implements ITextViewerExtensi
 	private static int computeEndOfVisibleRegion(int start, int length, IDocument document) throws BadLocationException {
 		int documentLength= document.getLength();
 		int end= start + length;
+		if (length == 0 || isLineBreak(document.getChar(end - 1))) {
+			return end + 1;
+		}
 		// ensure that the last line is fully included because projections cannot include partial lines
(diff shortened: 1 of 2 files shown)


vogella:
Thanks @danthe1st

PR #3672, opened 2026-01-14
Remove ISharedImages.*HOVER

akurtakov:
They are deprecated since 3.0 and marked for removal since 2024-03.
Reduces needlessly created images by the workbench.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/.settings/.api_filters
+++ b/bundles/org.eclipse.ui.workbench/.settings/.api_filters
@@ -102,6 +102,68 @@
             </message_arguments>
         </filter>
     </resource>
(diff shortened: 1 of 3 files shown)


akurtakov:
Content has been identical to the key without _HOVER since ages e.g. IMG_TOOL_UNDO_HOVER -> IMG_TOOL_UNDO

HeikoKlare:
Shouldn't we also remove the keys as well for the sake of consistency? Since they are marked for removal, we should either remove them together with the values they refer to or not at all, shouldn't we?
Is the rationale behind not removing the keys/APIs preservation of compatibility for old plugins?

akurtakov:
I was driven by effort to reduce needless lost memory for loading same images one more time. I agree with you about consistency and will remove the keys now too.

HeikoKlare (approved):
Great to see this cleaned up, thank you!

PR #3673, opened 2026-01-14
Fix IllegalArgumentException in PartServiceImpl.addPart when editor area is missing

vogella:
Fixes an IllegalArgumentException in PartServiceImpl.addPart that occurs when the shared editor area ('org.eclipse.ui.editorss') is not found.

The `modelService.findElements` method was called with a null `searchRoot` (area) when the editor area could not be located, triggering an assertion failure. This change adds a null check to handle this case gracefully.

Verified by running `OpenCloseTest` which now passes.


vogella:
IMHO this is reasonable behavior the PartService

PR #3675, opened 2026-01-15
Use varargs in FilteredList and AbstractElementListSelectionDialog

ptziegler:
This adapts the signature of the following methods to accept a vararg parameter instead of a plain array:

FilteredList:
- setElements(Object[]) -> setElements(Object...)
- setSelection(int[]) -> setSelection(int...)
- setSelection(Object[]) -> setSelection(Object...)

AbstractElementListSelectionDialog:
- setListElements(Object[]) -> setListElements(Object...)
- setSelection(Object[]) -> setSelection(Object...)

This stops callers from having to explicitly create an array when e.g. selecting a single element. For example `filteredList.setSelection(new int[]{0})` can be simplified to `filteredList.setSelection(0)`.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/AbstractElementListSelectionDialog.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/AbstractElementListSelectionDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -186,7 +186,7 @@ public abstract class AbstractElementListSelectionDialog extends SelectionStatus
 	 *
 	 * @param elements the elements of the list.
 	 */
-	protected void setListElements(Object[] elements) {
+	protected void setListElements(Object... elements) {
 		Assert.isNotNull(fFilteredList);
 		fFilteredList.setElements(elements);
 		handleElementsChanged();
@@ -263,7 +263,7 @@ public abstract class AbstractElementListSelectionDialog extends SelectionStatus
 	 *
 	 * @param selection the indices of the selection.
 	 */
-	protected void setSelection(Object[] selection) {
+	protected void setSelection(Object... selection) {
 		Assert.isNotNull(fFilteredList);
 		fFilteredList.setSelection(selection);
 	}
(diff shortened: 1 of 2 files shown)


akurtakov:
Thanks, these small and still compatible refinements of Eclipse API are smth I am very happy to see coming.

HeikoKlare (commented):
> these small and still compatible refinements of Eclipse API

I am not sure if these changes were actually compatible. Calling the changed methods with `null` now leads to a compiler error (as you can pass null as an array pointer but not as a varargs argument). The `AbstractElementListSelectionDialog.setSelection()` method even has a specification for a `null` argument in its contract:
> Empty or null array removes selection.

@akurtakov should we still keep this change? It will probably not affect many consumers (we have a single place in our codebase that does not compile anymore), but still it appears to be a breaking change to me.

ptziegler:
> as you can pass null as an array pointer but not as a varargs argument.

Yes you can...

image

If you mean the compiler _warning_ then yes, this is something that can be discussed.

akurtakov:
I would say - let's keep it. It's binary compatible change that introduces new compiler warnings in some cases but not 100% source compatible - documented at https://github.com/eclipse-platform/eclipse.platform/blob/808cabce9ffd5a0a14e882be4105d100d2ff8f25/docs/Evolving-Java-based-APIs-2.md#evolving-api-classes---api-methods-and-constructors . I understand how it can be inconvenient for projects that have raised some warnings to errors but that has always been the case with varargs adoption. Please let me know if I am missing something here and we have to think on it more.

HeikoKlare:
I see, we had changed the severity level for that case. I would still consider it a bit strange to have a method that explicitly accepts a `null` argument but leads to a warning when calling the method in that way. Maybe we should adapt the documentation to encourage using an empty list instead of null when one wants to clear the selection?

akurtakov:
It can still accept null argument but it has to be explicitly casted e.g. setListElements((Object[]) null); . Also if a variable of the Object[] type ends up null and is passed that would be handled fine. It's all about syntax sugar here IMO.

HeikoKlare:
Sure, that works. It just does not feel like that clean API. With the change to varargs you can now even simply do `setSelection()`.
But I see your arguments and with them collected, I am fine just keeping it as is. Thank you for your input!

ptziegler:
Being able to call `setSelection()` doesn't sound very nice... Then I'll just revert this and https://github.com/eclipse-platform/eclipse.platform/commit/da4f877759cf04c9f11d9862b3cdb63f973d8b86

akurtakov:
If you revert that will cause compile failures if code elsewhere has been adjusted and you'll make sure to revert all these changes too. setSelection() is somethin I read as "set no selection" . 
With the above said I believe reverting will be counter productive.

HeikoKlare:
I agree, With the above discussion, I would be in favor of keeping the change as is. I even think `setSelection()` is fine to clear a selection, but that seems to be a matter of taste 🙂

PR #3677, opened 2026-01-15
Don't do nested resize operations in TableResizeHelper

### LINKED ISSUE #3676: Quick Search repeatedly refreshes results and UI becomes “Not Responding” (STS 5/4 - Eclipse)
**Describe the bug**
Sometimes Quick Search dialog starts updating the “matching items” list repeatedly (multiple times per second) after opening it and changing the search string.
When this happens, the dialog becomes hard/impossible to close, and the window title shows (Not Responding). Only restarting the application helps.

Image

Image

**To Reproduce**
No reliable steps. Happens sporadically; usually works fine.

**Sample**
It doesn't depends on code.

[threaddump.txt](https://github.com/user-attachments/files/24639356/threaddump.txt)

The bug was initially reported to [spring-tools](https://github.com/spring-projects/spring-tools) project https://github.com/spring-projects/spring-tools/issues/1760

iloveeclipse:
Avoids endless resize operations & UI hangs in case resize listener/handler code itself triggers some OS specific resize events.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3676

### DIFF
--- a/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/TableResizeHelper.java
+++ b/bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/TableResizeHelper.java
@@ -28,6 +28,7 @@ import org.eclipse.swt.widgets.TableColumn;
 public class TableResizeHelper {
 
 	private final TableViewer tableViewer;
+	boolean resizeInProgress;
 
 	public TableResizeHelper(TableViewer tableViewer) {
 		this.tableViewer = tableViewer;
@@ -37,7 +38,9 @@ public class TableResizeHelper {
 		ControlListener resizeListener = new ControlListener() {
 			@Override
 			public void controlResized(ControlEvent e) {
-				resizeTable();
+				if (!resizeInProgress) {
+					resizeTable();
+				}
 			}
 			@Override
 			public void controlMoved(ControlEvent e) {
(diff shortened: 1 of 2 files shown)


iloveeclipse:
@HeikoKlare : I can't reproduce on Linux, so it looks like it is Windows specific. Could you please check if you can reproduce?

bundles/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/ui/TableResizeHelper.java
```diff
@@ -28,6 +28,7 @@
 public class TableResizeHelper {
 
 	private final TableViewer tableViewer;
+	boolean resizeInProgress;
```
iloveeclipse, line 31: package protected to avoid extra synthetic accessor method being generated

HeikoKlare (approved):
> I can't reproduce on Linux, so it looks like it is Windows specific. Could you please check if you can reproduce?

I tried different setups and inputs, but I was not able to reproduce the originally reported issue. Since the report mentions that the issue is sporadic, it seems not so easy to provoke that situation.
Since we quite some changes to control-sizing-related functionality because of the HiDPI functionality on Windows, I also checked if there might be anything involved here (which might have also been resolved recently), but I didn't find anything related.

Given the information we have from the error stack trace, the proposed solution seems fine to avoid a freeze. Since we don't know why the resize triggered itself repeatedly, there is a slight risk of a side effect/regression by the table potentially not reaching a "stable" resized state. But would still be better than running into a freeze.

iloveeclipse:
Thanks Heiko.

PR #3692, opened 2026-01-21
Remove redundant GC.dispose() calls for Image objects

### LINKED ISSUE #425: Remove unnecessary GC.dispose() calls for GCs on Image
When creating a GC for an image to programmatically draw that image, it is benefical to not dispose the GC before the image, as it allows to repaint the image at whatever zoom it is required as long as the GC is still active. In addition, the image disposes the GC upon its disposal anyway.

The code of the Eclipse SDK (Equinox, Platform, PDE, JDT) and GEF/Draw2D shall be checked for places to improve by removing the GC.dispose() call or by wrapping the implementation into an ImageGcWrapper. I propose to search for these places by looking for callers of the Image constructors `Image(Device, Rectangle)` and `Image(Device, int, int)`.

I scanned some of the projects already (Platform and GEF/Draw2D) and found the following:
- SWT
  - Image#ImageGcDrawerWrapper
  - Composite#WM_PAINT
- Platform UI:
  - TimeoutDumpTimer
  - ColorSelector#getImageData (JFace)
  - BreadcrumItemDropDown#AccessibleArrowImage#drawCompositeImage
  - ColorsAndFontsPreferencePage#PresentationLabelProvider
- GEF/Draw2D:
  - GhostImageFigure
  - ImageUtilities

Note that the list is incomplete and particlarly misses parts of the SDK (Equinox, PDE, JDT), so fone action item of this ticket should be the search for further places.

_FTR:_ We also discussed to extend the GC/Image implementation to either not dispose a GC on an image at all or to add a (strict) check that an GC on an image is not disposed before the image. The former has the drawback that it would implicitly und unnoticably keep resources/handles alive (such as images copied by the GC operations to be reapplicable). The latter is difficult to implement as a general check because there may be valid cases where you create multiple GCs (one after another) on the same image, so they would need to be adapted in some way to conform to the check. A strict check has the same drawback (but needs to be activated explicitly) but even then you need to have specific knowledge when the printed warning is actually useful/relevant and when it is not. To in my opinion, we should leave it up to the actual detection of blurry images to fix the implementation instead of proactively adding and automatisms or checks here. @akoch-yatta do you agree?

arunjose696:
This minor refactoring removes unnecessary calls to GC.dispose(), as disposing an Image automatically disposes its associated GC.

The images affected by this change are the color-rectangle visuals used in the color selector and syntax coloring. I tested the behavior both before and after the change and observed no visual differences.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ColorSelector.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ColorSelector.java
@@ -192,7 +192,6 @@ public class ColorSelector extends EventManager {
 		}
 		gc.setLineWidth(2);
 		gc.drawRectangle(image.getBounds());
-		gc.dispose();
 
 		ImageData data = image.getImageData(zoom);
 		image.dispose();


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
```diff
@@ -457,7 +457,6 @@ public Image getImage(Object element) {
 					gc.drawRectangle(offset, offset, usableImageSize - offset, usableImageSize - offset);
 					gc.fillRectangle(offset + 1, offset + 1, usableImageSize - offset - 1,
 							usableImageSize - offset - 1);
-					gc.dispose();
```
HeikoKlare, line 460: When will the image of this GC be disposed? It's not that obvious so it might be that this GC lives for a very long time. Couldn't the image creation be changed to an ImageGcDrawer instead?

arunjose696: replaced with ImageGcDrawer

arunjose696: Moved it to seperate PR as the commit message does not fit with the change to ImageGcDrawer

akoch-yatta (approved):
LGTM, image is disposed right after
The second place was converted to a ImageGcDrawer and into a separate PR

PR #3693, opened 2026-01-22
Use proper autoscale value for splash screen scaling

### LINKED ISSUE #596: Use proper autoscale value for splash screen scaling
Tracking ticket for:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3693

HeikoKlare:
The splash screen is initialized by the Equinox native launcher such that when the Workbench takes it over, the autoscaling settings used for processing it must fit to avoid a wrongly sized background image and controls.
Since the autoscaling default used to be "integer" and any other value was only defined by a system property, reading that system property was sufficient. With the adaptation of SWT and Equinox to default to autoscaling mode "quarter" on Windows when monitor-specific scaling is used, reading the system property is not sufficient anymore as it also depends on whether monitor-specific scaling is enabled.

This change adapts the identification of the proper autoscaling mode by just reading it from the SWT configuration. This also ensure that it will be resilient against potential future changes of default settings.

Without this change, starting an application based on the SWT and Equinox state of 2026-01-22 or newer and deactivating monitor-specific scaling via the Eclipse preference would result in unintended results on monitors with fractional scaling. The following shows the result on a 150% primary monitor:
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/Workbench.java
@@ -3666,14 +3666,12 @@ public final class Workbench extends EventManager implements IWorkbench, org.ecl
 	}
 
 	private static class AutoscaleAdaptation {
-		private static final String SWT_AUTOSCALE = "swt.autoScale"; //$NON-NLS-1$
-
 		private boolean incompatibleMonitorSpecificScalingDisabled;
 
 		private final String initialAutoScaleValue;
 
 		public AutoscaleAdaptation() {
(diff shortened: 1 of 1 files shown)


akoch-yatta (approved):
LGTM, small change, but improves the behavior as described

vogella:
LGTM, please merge


For clarification swt.autoScale will not work anymore in the future for testing DPI changes? Or will this only affect the spash?

akoch-yatta:
> LGTM, please merge
> 
> For clarification swt.autoScale will not work anymore in the future for testing DPI changes? Or will this only affect the spash?

This is just a specific fix for the splash screen behavior to align better with the active auto scale mode than before.

PR #3694, opened 2026-01-22
Correct wrongly scaled images in colors/fonts preference page

### LINKED ISSUE #425: Remove unnecessary GC.dispose() calls for GCs on Image

arunjose696:
The colors/fonts preference page uses images for the items shown in a tree viewer that preview the selected color. The current implementation has two limitations:
- It does not properly scale with the current monitor zoom
- It uses a wrong image size based on the tree item height instead of the tree item image size, such that the rendered image is scaled and may loose information

This change removes all custom extraction of the image size as the image simply needs to have the same size as all other images default to (16x16) to fit. It also wraps the implementation info an ImageGcDrawer to be monitor-scale-aware.

### Before
Opening on 125%:
image

Moving from 100% to 125%:
image

### After
125%:
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
@@ -63,6 +63,7 @@ import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.graphics.FontMetrics;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.ImageGcDrawer;
 import org.eclipse.swt.graphics.RGB;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.layout.FillLayout;
@@ -333,10 +334,6 @@ public final class ColorsAndFontsPreferencePage extends PreferencePage implement
 
 		private final HashMap<Color, Image> images = new HashMap<>();
 
-		private int imageSize = -1;
-
-		private int usableImageSize = -1;
-
 		private static final int REFRESH_INTERVAL_IN_MS = 300;
 
 		private final Runnable updateControlsAndRefreshTreeRunnable = () -> {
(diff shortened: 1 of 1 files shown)


amartya4256 (approved):
Looks good. The build failing is unrelated to the changes. Approving.

akoch-yatta (approved):
Looks good overall

When testing I notived a white background in dark mode, but this is already there, so not caused by this PR.
Image

PR #3698, opened 2026-01-23
Suppress discouraged access and missingJavadocComments warnings

HannesWell:
The affected classes access internal methods intentionally.


HannesWell:
Interestingly, multiple of the discouraged access warnings shown in the IDE (and which make sense), are not considered in the build and therefore new warnings are added as well (but net, more warnings are fixed). These warnings also show up in the I-build compiler reports.

I tried to find out where this configuration mismatch originates, but I didn't found it yet.
Therefore I'm progressing here and leave that investigation for later.

PR #3702, opened 2026-01-26
Fix invalid elements in SVGs for replace icons

HeikoKlare:
The icons contain a <g> element inside <clipPath>, which the JSVG rasterizer reports to be invalid or unsupported. Since the clipPath is not necessary at all, this change removes it.

Fixes these types or warnings:
```
Jan. 26, 2026 4:18:06 PM com.github.weisj.jsvg.nodes.container.BaseContainerNode isAcceptableType
WARNUNG: Element <g> not allowed in <clippath> (or not implemented)
```


akurtakov:
The icons look identical so should be fine to merge. Where/how do you see the warnings?

HeikoKlare:
You see the warnings in the logs when opening the replace bar of the find/replace overlay (CTRL+F + CTRL+R, in case you did not switch back to the find/replace dialog UI).

akurtakov:
I see. Thanks for this one.

PR #3703, opened 2026-01-26
Adapt HistoryTextWrapper#notifyListeners

### LINKED ISSUE #569: Ctrl+f search box is not relayouted on DPI change
When a file editor has the Ctrl+F search box open and the editor window is moved across screens with different DPI/scaling settings, the search box does not scale properly.

Steps to Reproduce:

1. Open a file in the editor.
2. Press Ctrl+F to open the search box.
3. Move the editor window from one monitor to another with a different zoom/DPI setting.
4. Observe that the search box remains the same size and does not adjust to the new DPI.

Image

amartya4256:
This PR adapts the notifyListeners method in HistoryTextWrapper to call super#notifyListeners since not calling it changes the underlying default behaviour of its superclasses leading to faulty behavior, e.g. it breaks the chain of DPIChange processing leading to a broken drop down button in the HistoryTextWraper on DPIChange Event.

#### Bug Explained
When a file editor has the Ctrl+F search box open and the editor window is moved across screens with different DPI/scaling settings, the search box does not scale properly.

Steps to Reproduce:

1. Open a file in the editor.
2. Press Ctrl+F to open the search box.
3. Move the editor window from one monitor to another with a different zoom/DPI setting.
4. Observe that the search box remains the same size and does not adjust to the new DPI.

Image

#### Reason
The class `HistoryTextWrapper ` overrides the method `notifyListeners`:

```
@Override
public void notifyListeners(int eventType, Event event) {
	textBar.notifyListeners(eventType, event);
}
```

This method doesn't call `super.notifyListeners(eventType, event)`, hence, the chain of DPIChange processing is broken, leading to the toolbar under the `HistoryTextWrapper` never scaling.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/HistoryTextWrapper.java
@@ -200,6 +200,7 @@ public class HistoryTextWrapper extends Composite {
 
 	@Override
 	public void notifyListeners(int eventType, Event event) {
+		super.notifyListeners(eventType, event);
 		textBar.notifyListeners(eventType, event);
 	}
 


HeikoKlare (approved):
The missing super call obviously breaks proper notification handling of the composite. Every other control overwriting that method does as super call as well. It seems to have been an unnoticed oversight in the original implementation of this class in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1990. At least there is no mention why exactly this kind of implementation is necessary.

PR #3705, opened 2026-01-27
StackRenderer updates correctly when part is moved

vogella:
Fixes Eclipse Bug 576186.

 The StackRenderer now correctly updates its UI when a part within the stack is moved. This was achieved by handling UIEvents.isMOVE events for MPartStack children in StackRenderer.java. CTabItems, which represent tabs, cannot be reordered directly and must be disposed of and recreated at the new index. The methods calcIndexFor and findItemForPart in StackRenderer were updated to accept a more generic MElementContainer<? extends MUIElement> parameter to resolve type compatibility issues with MPartStack.

Also a test is added for this behaivor

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
@@ -438,6 +438,8 @@ public class StackRenderer extends LazyStackRenderer {
 		Font font = item.getFont();
 		Object data = item.getData();
 
+		boolean wasSelected = tabFolder.getSelection() == item;
+
 		item.dispose();
 
 		CTabItem newItem = new CTabItem(tabFolder, (showClose ? SWT.CLOSE : SWT.NONE), newIndex);
@@ -448,6 +450,9 @@ public class StackRenderer extends LazyStackRenderer {
 		newItem.setData(data);
 		newItem.setData(OWNING_ME, movedElement);
 		newItem.setControl(control);
+		if (wasSelected) {
+			tabFolder.setSelection(newItem);
+		}
 	}
 
 	@Inject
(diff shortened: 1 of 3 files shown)


vogella:
Does not work, in case I use the example with from the Bug 576186 the UI gets correctly updated BUT if you start interacting with the UI you get additional tabs. We would not see that in the IDE because move alone (without additonal code)cannot be used due to the bug. 

Example code for moving a part around.

@Execute
	public void execute(MPart part, EModelService modelService, MWindow window) {
		MPartStack find = (MPartStack) modelService.find("atest.partstack.sample", window);
		modelService.move(part, find, 2);
	}
}


image

bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -395,6 +395,61 @@ void subscribeTopicChildrenChanged(@UIEventTopic(UIEvents.ElementContainer.TOPIC
...
+		item.dispose();
+
+		CTabItem newItem = new CTabItem(tabFolder, (showClose ? SWT.CLOSE : SWT.NONE), newIndex);
+		newItem.setText(text);
+		newItem.setImage(image);
+		newItem.setToolTipText(toolTipText);
+		newItem.setFont(font);
+		newItem.setData(data);
+		newItem.setData(OWNING_ME, movedElement);
+		newItem.setControl(control);
```
vogella, line 452: When the currently selected tab is moved, the selection is not preserved. I should try to save it

java   boolean wasSelected = tabFolder.getSelection() == item;
   // ... dispose and recreate ...
   if (wasSelected) {
       tabFolder.setSelection(newItem);
   }

tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRendererMoveTest.java
```diff
@@ -0,0 +1,97 @@
...
+		assertEquals(part1, children.get(1));
+
+		// Verify UI update
+		assertEquals(2, tabFolder.getItemCount());
+		CTabItem newItem1 = tabFolder.getItem(1);
+		CTabItem newItem2 = tabFolder.getItem(0);
+
+		// The old item1 should be disposed
+		assertTrue(item1.isDisposed(), "Old item for part1 should be disposed");
+
```
vogella: Also check    assertFalse(item2.isDisposed(), "Item2 should not be disposed");

vogella (commented):
Needs adjustments

vogella:
FYI: EModelService.move is not used at all in platform accoding to call hierachy

vogella:
> Needs adjustments

Done

BeckerWdf:
This is the link to the bug in good old bugzilla: https://bugs.eclipse.org/bugs/show_bug.cgi?id=576186

PR #3708, opened 2026-01-27
Migrate org.eclipse.jface.tests to JUnit 5

vogella:
Migrate remaining org.eclipse.jface.tests to JUnit 5


vogella:
New compiler warning looks unrelated to me but I wait until someone else confirms.

Internal inconsistency: Inappropriate operand stack size encountered during translation

features/.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
akurtakov: This file shouldn't be committed

vogella: Done

tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/fieldassist/FieldAssistAPITests.java
```diff
@@ -32,45 +32,45 @@ public class FieldAssistAPITests extends AbstractFieldAssistTestCase {
 	@Test
 	public void testSimpleContentProposal() {
 		proposal = new ContentProposal(content);
-		assertEquals("1.0", content, proposal.getContent());
```
akurtakov, line 35: These messages "1.0" are totally useless. It's better to remove them. Here and in many other places.

vogella: I agree but is unrelated to the migration. Should be done in a follow up commit.

vogella: I agree but is unrelated to the migration. Should be done in a follow up commit.

akurtakov: Agreed.

PR #3732, opened 2026-02-17
Adapt placement test in StatusDialogManagerTest for fractional scaling

### LINKED ISSUE #3717: StatusDialogManagerTest.testBug275867() flaky in Windows I-Builds
The test case `org.eclipse.ui.tests.statushandlers.StatusDialogManagerTest.testBug275867()` sporadically fails in I-Builds on Windows:

```java
Dialog is not centered correctly expected:<Point {254, 102}> but was:<Point {254, 103}>

java.lang.AssertionError: Dialog is not centered correctly expected:<Point {254, 102}> but was:<Point {254, 103}>
at org.junit.Assert.assertEquals(Assert.java:120)
at org.eclipse.ui.tests.statushandlers.StatusDialogManagerTest.testBug275867(StatusDialogManagerTest.java:777)
```

From the values, it might be some rounding issue related to HiDPI support / monitor-specific scaling.
As an initial thought if we cannot reproduce the fail locally or have any other clue: we might add debug output with the actual and expected values to the test, such that we know how the values differ in case the test succeeds. It would be strange if in case of a success the values are the exact same as we see in case of a failure (except for the off-by-one), as there should be no non-determinism in how values are calculated/rounded. That's why I would expect the values to be completely different in case of a success, which could be verified by debug output.

HeikoKlare:
The StatusDialogManagerTest.testBug275867() sporadically fails on Windows because of off-by-one errors between the expected and actual location of the shell, probably caused by rounding on fractional monitor zooms.
To be more resilient against slight offsets in the position of the shell, this change adds a tolerance to the position validation. It also adds some more debug information to the failure description.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3717

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/statushandlers/StatusDialogManagerTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/statushandlers/StatusDialogManagerTest.java
@@ -774,7 +774,16 @@ public class StatusDialogManagerTest {
 		wsdm.addStatusAdapter(statusAdapter, false);
 		assertNotNull(StatusDialogUtil.getStatusShell());
 		Shell shell = StatusDialogUtil.getStatusShell();
-		assertEquals("Dialog is not centered correctly",getInitialLocation(shell), shell.getLocation());
+
+		Point expected = getInitialLocation(shell);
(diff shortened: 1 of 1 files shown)


HeikoKlare:
The addressed failure happened in the I-Build last night again, so let's try to fix it to remove noise from I-Build test results.

Since this is a test-only fix, we can safely merge within freeze period.

PR #3734, opened 2026-02-17
Replace GIF icons with SVGs in example.navigator project

vogella:
- Copied delete_obj.svg, filenav_nav.svg, and prop_ps.svg from workbench bundles.
- Updated plugin.xml to reference the new SVG icons.
- Removed the original GIF files.
- Updated build.properties to include the icons directory in the binary build.


BeckerWdf:
but (at least some) icons are completely different. Is that by intention?

vogella:
> but (at least some) icons are completely different. Is that by intention?

I searched for fitting icons with the same name, as these are examples and are not visible in an Eclipse installation  I did not care too much if some of them look different.

BeckerWdf:
can we get a before vs. after screenshot

vogella:
> can we get a before vs. after screenshot

Sure, I will split the PR into individual PR per project and add before and after screenshos. 
I will do this after the freeze, not time critical as the current gifs have been ugly since a decade on a darker background.

nettozahler:
Would be very nice if we can get prettier icons on a dark background.

vogella:
> Would be very nice if we can get prettier icons on a dark background.

See https://github.com/eclipse-platform/ui-best-practices, and the script which let you exchange the icons for the IDE https://github.com/eclipse-platform/ui-best-practices/blob/main/scripts/replace_icons.sh see readme for  its usage

PR #3748, opened 2026-02-27
Fix sticky scroll height calculation for variable line heights

utafrali:
When emoji or other content affects line heights, sticky scroll was calculating the total height wrong by assuming all lines were the same. Now it gets the actual height for each line instead. Also fixes the height check for limiting visible lines.

### DIFF
--- a/tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControlTest.java
+++ b/tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControlTest.java
@@ -17,6 +17,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.greaterThan;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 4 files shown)


BeckerWdf:
@Christopher-Hermann can you pls. review?

utafrali:
Closing — can't sign the ECA.

iloveeclipse:
> Closing — can't sign the ECA.

You mean, there is a technical problem or you won't sign?

utafrali:
Honestly I'm not comfortable with the obligations that come with signing the ECA — I feel like it goes against the open source spirit. Closing this one out.

iloveeclipse:
> I feel like it goes against the open source spirit.

Could you please give some example? Eclipse contributors agreement is nowhere against open source. Whole Eclipse Foundation and all open source projects hosted here are based on work of people signed ECA.

merks:
Yes I too am genuinely interested in the concrete concerns!  I’m on the Board of Directors IP Advisory Committee as an elected committer representative so I can raise any concerns at the highest levels. I’ve never heard anyone express concerns about signing the agreement coming with obligations. Mostly it is intended as an assertion that you authored the content and agree to make it available under the open source license.  I’d like to understand how you see it differently. We want everyone to be included and valued in our community with no obligations beyond personal integrity.

utafrali:
Yeah, you're right. I think I just got spooked by the word "agreement" without actually looking into it. I've done a bit more reading and I'm happy to sign the ECA. Could you point me in the right direction for linking it to my GitHub account?

merks:
Start here

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#setting-up-your-eclipse-and-github-account

We’re here to help you. Just ask.

tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControlTest.java
```diff
@@ -236,6 +236,26 @@ void testStyling() {
...
+		stickyScrollingControl.setStickyLines(stickyLines);
+
+		Canvas stickyControlCanvas = getStickyControlCanvas(shell);
+		StyledText stickyLineText = getStickyLineText();
+		Composite stickyLineSeparator = getStickyLineSeparator();
+
+		int expectedHeight = 0;
+		for (int i = 0; i < 2; i++) {
+			expectedHeight += stickyLineText.getLineHeight(stickyLineText.getOffsetAtLine(i));
+		}
```
Christopher-Hermann: Not sure about the test. I remember that it was kind of ugly to write proper assertions on the size since OS, rendering, etc. all effects the unit test. So having hard coded size will not work. But I guess this test would be green even without the fix, right?

What do you think about this idea:
1. Set the 2 sticky lines as you did, get the `stickyControlCanvas.getBounds().height`
2. Replace the second sticky line with something that requires more space. As far as I understand the description emoji for example? get the new height
3. Assert that the new height (with emoji) is greater than the old height (without emoji)

I guess this would also be a good documentation of the actual bug, what do you think?

Christopher-Hermann (commented):
Thanks for providing a fix 👍 

Productive coding lokos good. I added one comment about the test

tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/internal/texteditor/stickyscroll/StickyScrollingControlTest.java
utafrali: Thanks for the great suggestion! You're right that the original test would have passed even without the fix, and hardcoded sizes are fragile across different OS/rendering environments.

I've implemented your approach: the test now sets plain-text sticky lines first, records the canvas height, then replaces the second line with emoji text and asserts that the height increases. This makes the test directly document the bug behavior - before the fix, the canvas height wouldn't adjust when emoji required more vertical space.

iloveeclipse:
@utafrali : please sign ECA. Without that PR can't be accepted.

Christopher-Hermann:
@utafrali Sorry to hear that you want sign the ECA, but okay for me 👍
Anyway, thanks for your contribution to the community and for providing the fix. Is it okay if I implement the fix myself? Thanks & enjoy your time!

PR #3753, opened 2026-02-28
Search filter showing only deepest match should't filter closed projects

merks:
- A closed project's content cannot be searched so matches attributed to a closed nested project should not filter the corresponding match from the outer project otherwise you can get very confusing results like `showing 0 of 1 matches`.
- Fix action label which is shown on a menu to be title case.


merks:
In this simple case, without the fix, the only match is filtered from the view

image

With the fix, including the title case action label it works properly:

image

PR #3755, opened 2026-03-02
Fix dark theme black toolbar background in inactive view stacks

vogella:
Two related bugs caused the toolbar composite (top-right area of a CTabFolder) to show a black background in the dark theme for any view stack that was never activated by the user:

1. CSS cascade timing (StackRenderer): bindWidget() was called before addTopRight(), so the initial CSS cascade triggered by bindWidget never reached trComp (it didn't exist yet). Only stacks later activated by the user got a subsequent cascade that corrected the background. Fix: call addTopRight() before bindWidget().

2. Wrong CSS selector type (e4-dark_tabstyle.css): the rule used "Composite#ToolbarComposite" (ID selector) but ToolBarManagerRenderer sets a CSS class, not a CSS ID, on the composite. The rule therefore never matched. Fix: change "#ToolbarComposite" to ".ToolbarComposite".


vogella:
Before: 

image

After:

image

PR #3756, opened 2026-03-02
Demo 1111

flexsiddu:
Fixed issue with quotation marks for content-types in file associations


iloveeclipse:
Don't get me wrong, but please your private forks for learning purposes, or at least mark your PR as a draft.

PR #3759, opened 2026-03-03
Dispose bold/italic fonts in FontRegistry upon display disposal

### LINKED ISSUE #615: Non-disposed bold/italic fonts in FontRegistry upon display disposal
Tracking ticket for:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3759

HeikoKlare:
The FontRegistry implementation uses hooks to dispose registered fonts upon disposal of the display they have been created for. This implementation does, however, only cover the base fonts but not the bold/italic versions of them. In case only a bold/italic font of a record is created on a separate display, no dispose hook for that display will be created, such that those fonts will not be disposed when the display becomes disposed. In consequence, fonts for non-disposed devices may be returned by the registry.

This change improves the FontRegistry implementation to create a display dispose hook also upon creation of a bold/italic font.


BeckerWdf (commented):
just read the code - did not test.
But code looks good to me.

PR #3760, opened 2026-03-04
Configurably improve disabled icons appearance

HeikoKlare:
This consists of two subsequent changes:
- Allow to ignore explicitly defined disabled icons (as SWT provides a [proper way to do it on-the-fly since 2025-06](https://eclipse.dev/eclipse/markdown/?f=news/4.36/platform.md#improved-disabled-icons-generation))
- Allow to use a different/enhanced algorithm for calculating disabled icons (based on desaturation rather than grayscaling).

The changes are put into separate commits and could be processed independently, but are mostly useful together, thus I put them into a single PR.

### Description

The algorithm for generating disabled icons on-the-fly has recently been enhanced. Most explicit disabled icons that have been embedded into bundles conform to what now can be generated on-the-fly. In addition, the algorithm is interchangeable, allowing custom stylings of disabled icons. However, when there are still bundles, such as extensions out of own control, that still explicitly define disabled icons, exchanging the algorithm for disabled icons will lead to inconsistent appearance as only the on-the-fly generated icons will adhere to that.

This change allows to ignore explicitly defined disabled icons, such that even if some bundles defined disabled icons, they will be ignored and on-the-fly generated disabled icons according to the selected algorithm will be used instead. An according preference that can be configured via the appearance tab is added.

This also adds a preference which enables the use of desaturated disabled icons, which replaces the existing algorithm to create grayscaled version with an algorithm that generates a desaturated version.

This alternative algorithm for disabled icons has been discussed and agreed on here:
- https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1741
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1936

### How to test

Both enhancements are configurable via preferences, thus can be tested by enabling/disabling them:
image

As an example, this is what the main toolbar looks like with different configurations:

#### Existing
image

#### No pregenerated disabled icons
image

#### Desaturated disabled icons
image

### Feedback

Feedback on the proposal is welcome as always. We can of course decide if we want to have both/either of these options enabled by default or not. Using the on-the-fly algorithm for calculating disabled icons by default makes sense to me to ensure a unified look and feel of them. Whether the alternative algorithm (desaturated vs. grayscaled) should be default may be more up to individual preference. But note that the new algorithm better aligns with common disablement visualizations used nowadays by Windows and GTK.

@HannesWell pinging you for feedback as we had already discussed about adding such options after we had implemented the new algorithm

@BeckerWdf pinging you for feedback as an expert on UX topics and as you were also involved into the discussion about the disabled icon look & feel back then


iloveeclipse:
I'm not sure if that is used somewhere, that "disabled" icon can show a semantically different icon as the "enabled" one? Let say "+" is enabled, "-" is disabled icon? This use case would be broken? Or is there a way to "force" some predefined disabled icons to be still used?

Beside this, I personally feel "Desaturated" variant not so good, the contrast for me is too low, so the first "auto-generated" variant looks better.

HeikoKlare:
> I'm not sure if that is used somewhere, that "disabled" icon can show a semantically different icon as the "enabled" one? Let say "+" is enabled, "-" is disabled icon? This use case would be broken? Or is there a way to "force" some predefined disabled icons to be still used?

I am not aware of any such use case. In all SDK projects and some around, we have removed the references to all predefined disabled icons last year and we did not see such a use case there. Still, it is of course theoretically possible that someone uses a disabled icon that is "independent" from the enabled one. We could think of adding some "force" option/parameter to allow enforcement of a specific disabled icon even when not using the predefined icons by default. But if no one has that use case, it would introduce unnecessary complexity. As an alternative, we may not provide such an option for now and wait if someone reports a need for it. Since the behavior of not using any predefined disabled icons is configurable, there would always be a workaround if someone needs specific disabled icons.

> Beside this, I personally feel "Desaturated" variant not so good, the contrast for me is too low, so the first "auto-generated" variant looks better.

We could even adapt the algorithm for the desaturated version as it was only introduced as "preview" so far: https://eclipse.dev/eclipse/markdown/?f=news/4.36/platform.md#improved-disabled-icons-generation

Or we may even provide a more sophisticated option list than just grayscaled vs. desaturated, as there is already a third algorithm available (emulating the GTK-way of disabling icons, which might be interesting for Linux users). If there is a need for it, there could be even more implementations and maybe even a registry for them, as it's just an interface to implement. But without according need it's probably not necessary to introduce too much configuration complexity.

PR #3762, opened 2026-03-06
Set compiler.problem.forbiddenReference to error for all projects

HannesWell:
If it's not already an error.

See
- https://github.com/eclipse-pde/eclipse.pde/issues/2244#issuecomment-4013458419


tests/org.eclipse.ui.tests.navigator/META-INF/MANIFEST.MF
```diff
@@ -18,6 +18,7 @@ Require-Bundle: org.eclipse.core.resources,
 Bundle-Vendor: Eclipse.org
 Import-Package: org.junit.jupiter.api;version="[5.14.0,6.0.0)",
  org.junit.jupiter.api.function;version="[5.14.0,6.0.0)",
+ org.junit.jupiter.api.io;version="[5.14.0,6.0.0)",
```
HannesWell, line 21: And we have our first missing dependency.
Without it, the `@TempDir` annotation cannot be used in:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/c5f3a58b40dda04da4e5e3bfe992b25150adee4f/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/resources/FoldersAsProjectsContributionTest.java#L56

The reason why this didn't surface at runtime, is probably because Annotations whose class cannot be loaded, are just silently ignored.
But the creation/clean-up of that directory certainly wouldn't have worked.
'Luckily' the referencing tests are disabled.

PR #3763, opened 2026-03-07
Minor improvements to o.e.core.filebuffers.tests

akurtakov:
Various cleanups and modernizations provoked by unstable test as can be seen at https://download.eclipse.org/eclipse/downloads/drops4/I20260306-1800/testresults/html/org.eclipse.core.filebuffers.tests_ep440I-unit-linux-x86_64-java25_linux.gtk.x86_64_25.html

### DIFF
--- a/tests/org.eclipse.core.filebuffers.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.core.filebuffers.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.core.filebuffers.tests;singleton:=true
-Bundle-Version: 3.14.0.qualifier
+Bundle-Version: 3.14.100.qualifier
 Bundle-Activator: org.eclipse.core.filebuffers.tests.FileBuffersTestPlugin
 Bundle-ActivationPolicy: lazy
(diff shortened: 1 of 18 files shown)


akurtakov:
Failed test is not being modified by this change aka unstable one. Merging.

trancexpress:
@akurtakov I'm seeing this compile error now:

```
Description	Resource	Location	Path	Type
The method createLinkedProject(String, JdtTextTestPlugin, Path) is undefined for the type ResourceHelper	CompilationUnitDocumentProviderTest.java	line 107	/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests	Java Problem
```

Am I missing some other change?

See: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2841

merks:
I see this is removed:

image

But that's what JDT is calling...

Perhaps there was an assumption that this test helper class is not used outside of the repository/project that hosts it.

merks:
Given @akurtakov seems to be unavailable, maybe it's best to put that method back?

trancexpress:
I've grepped through what Eclipse projects I have checked out, it seems that one JDT UI test is the only caller of the removed method. I've set: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2841

From my POV we can also re-add the method, though maybe someone removes it again in future since it looks to be unused.

merks:
Yes.  One might instead create a copy in the only place where it's used...

PR #3774, opened 2026-03-12
Don't init field to default value in ZeroWidthCharactersLineContentCodeMiningProvider

iloveeclipse:


iloveeclipse:
Change is trivial, builds are broken not because of the change (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3713). Merging.

PR #3779, opened 2026-03-16
Fix flaky AsyncContentAssistTest.testCompleteActivationChar

### LINKED ISSUE #890: JUnit tests in org.eclipse.jface.text.tests.contentassist are failing on Windows and Mac
This is a duplicate of [Bug 571316](https://bugs.eclipse.org/bugs/show_bug.cgi?id=571316) - JUnit tests in org.eclipse.jface.text.tests.contentassist are failing on Windows and Mac

I will close the bug in bugzilla and this bug will be tracked here.

The JFace text tests in org.eclipse.jface.text.tests.contentassist fail in the nightly build as below

Windows
1) testCompleteActivationChar

Both Windows and Mac
1) testSyncFailureNPE

vogella:
## Summary
- Replace `Display.post()` with `control.notifyListeners()` for reliable event delivery in headless CI environments — `Display.post()` sends native OS events that may not be processed, causing the content assist popup to never appear
- Fix race condition by capturing `beforeShells` before posting key events, and process events after `shell.open()`
- Increase `findNewShell` timeout from 1s to 5s for slower CI environments
- Remove Windows-only `assumeFalse` skip — the test now works on all platforms

## Test plan
- [x] Ran `AsyncContentAssistTest` 5 times consecutively — all passes
- [x] Ran full `org.eclipse.jface.text.tests` suite — no regressions (only pre-existing `CodeMiningTest` failures)

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/890

### DIFF
--- a/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/AsyncContentAssistTest.java
+++ b/tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/AsyncContentAssistTest.java
@@ -17,7 +17,6 @@ package org.eclipse.jface.text.tests.contentassist;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assumptions.assumeFalse;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -40,8 +39,6 @@ import org.eclipse.swt.widgets.Widget;
 import org.eclipse.core.runtime.ILogListener;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
-
-import org.eclipse.jface.util.Util;
 
 import org.eclipse.jface.text.Document;
 import org.eclipse.jface.text.IDocument;
@@ -134,7 +131,6 @@ public class AsyncContentAssistTest {
 
 	@Test
 	public void testCompleteActivationChar() {
-		assumeFalse(Util.isWindows(), "test fails on Windows, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/890");
 		shell.setLayout(new FillLayout());
 		shell.setSize(500, 300);
 		SourceViewer viewer= new SourceViewer(shell, null, SWT.NONE);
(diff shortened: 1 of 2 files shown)


vogella:
@trancexpress please review

trancexpress:
> @trancexpress please review

@laeubi can you review the change? I'm not sure what the intent is of sending SWT events. Sending the event directly to a widget seems odd to me, but so does sending SWT events.

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/AsyncContentAssistTest.java
```diff
@@ -151,17 +147,20 @@ public void testCompleteActivationChar() {
...
 		Display display= shell.getDisplay();
-		Event keyEvent= new Event();
+		DisplayHelper.runEventLoop(display, 0);
 		Control control= viewer.getTextWidget();
+		control.forceFocus();
+		DisplayHelper.runEventLoop(display, 0);
+		final Collection<Shell> beforeShells= AbstractContentAssistTest.getCurrentShells();
+		// Use notifyListeners instead of Display.post() for reliable event delivery
+		// Display.post() sends native OS events that may not be processed reliably
+		// in headless CI environments
```
trancexpress, line 157: Any idea why that would be? For better or worse we have a lot of tests that sent SWT events and we don't see sporadic fails with them.

Is something buggy in the the test or in the Eclipse CI environment, when it comes to UI event handling?

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/AsyncContentAssistTest.java
```diff
@@ -151,17 +147,20 @@ public void testCompleteActivationChar() {
...
 		keyEvent.widget= control;
 		keyEvent.type= SWT.KeyDown;
 		keyEvent.character= 'b';
 		keyEvent.keyCode= 'b';
-		control.getShell().forceActive();
-		control.getDisplay().post(keyEvent);
-		keyEvent.type= SWT.KeyUp;
-		control.getDisplay().post(keyEvent);
-		final Collection<Shell> beforeShells= AbstractContentAssistTest.getCurrentShells();
+		control.notifyListeners(SWT.KeyDown, keyEvent);
```
trancexpress, line 163: I would start with adding a listener to the widget and recording if the sent event is received. Then a check can validate that the event was received, prior to the timeout assert fail.

That way we would be certain that the problem really is event handling and not something else. If we do know event processing is buggy, a change like this still seems odd but it would be fine from my POV - if the test still tests what it should (Christoph can comment on this, I cannot).

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/AsyncContentAssistTest.java
vogella: Display.post() injects events into the native OS event queue, which requires a focused window to receive them. In headless CI environments  (Xvfb/virtual framebuffer), there's no window manager (reliably managing focus). Other tests using Display.post() may succeed because they don't depend on content assist popup detection within a tight timing window.

tests/org.eclipse.jface.text.tests/src/org/eclipse/jface/text/tests/contentassist/AsyncContentAssistTest.java
vogella: That's a fair point. You could add a diagnostic version first that adds a KeyDown listener on the widget, records whether the event  arrives, and logs/asserts on that before the timeout. This would confirm the hypothesis. However, note that notifyListeners() still triggers the same SWT  listener chain, ContentAssistant installs a VerifyKeyListener on the StyledText widget, and notifyListeners(SWT.KeyDown, ...) triggers that same path. The test still validates that typing an activation character triggers content assist and that the popup appears with the correct proposals. The only difference is the event delivery mechanism, direct widget notification vs. native OS queue.

PR #3781, opened 2026-03-16
Remove beep from content assist when no proposals found

vogella:
## Summary
- Remove `Display.beep()` call in `CompletionProposalPopup.hideWhenNoProposals()` that fires every time content assist is manually invoked but has no proposals to show

## Test plan
- [x] Manually verified content assist no longer beeps when no proposals are available


vogella:
This beep was driving me almost crazy during the work on https://github.com/eclipse-platform/eclipse.platform.ui/pull/3779

vogella:
@xpomul WDYT? Or do you think the beep is helpful?

xpomul:
> @xpomul WDYT? Or do you think the beep is helpful?

No, I don't think it's helpful.

If at all, some small non-disturbing visual indicator (such as a 🚫 popup that is shown for a second or so, and maybe even deactivateable for advanced users) might be nice. But that is a separate enhancement IMO.

PR #3783, opened 2026-03-16
Use correct font for highlighting matches in "Open Resource" dialog

### LINKED ISSUE #3782: Substring match highlighting in "Open Resource" dialog uses wrong font

## Steps to reproduce

* in the preferences, change the Dialog Font to a different font and/or size
* Create or import a project, so that at least one resource is available in the workspace
* Invoke "Open Resource" to open the dialog
* type in the filter text field so that it matches a part of a resource in the workspace

I expected: The highlight font to be the same as in the rest of the list items

But got: a different font (and size) for the bold font highlighting:

Image

* OS & version: MacOS
* Eclipse IDE/Platform Version: 2026-03 (4.39.0) Build id: 20260305-0817

## Community

xpomul:
# Summary

Fixes #3782 by using the dialog font as base for the highlighting style instead of the default font.

# Test Plan

(See also #3782)

* change dialog font to a bigger size
* open "Open Resource" dialog
* type in the filter box
* verify that the highlighting of matching substrings uses the dialog font size

### DIFF
--- a/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.ide/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.ui.ide; singleton:=true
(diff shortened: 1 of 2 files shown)


vogella:
LGTM

PR #3789, opened 2026-03-17
Implement Cut action for resources in Project Explorer

vogella:
This PR implements the 'Cut' action for resources in the Common Navigator-based Project Explorer.

Previously, only Copy, Paste, and Delete were supported. This implementation adds the ability to cut resources to the clipboard and move them to a target location upon pasting.

### Key Changes:
- **CutAction.java**: New package-private action that handles clipboard 'Cut' operations and tracks cut state via stored resource references (not a fragile boolean flag).
- **PasteAction.java**: Updated to perform a move (using `MoveFilesAndFoldersOperation`) if a cut is pending, otherwise continuing with a copy. Verifies clipboard contents match stored cut resources for robustness.
- **EditActionGroup.java**: Wired the new action into the context menu and global action bars.
- **CopyAction.java**: Updated to clear cut state when a copy operation is performed.
- **Localization**: Added necessary strings.
- **Tests**: Expanded `CopyPasteActionTest` with 14 test cases covering enablement, clipboard state, and move verification.

### Cut State Tracking:
The cut/paste state is tracked by storing the actual `IResource[]` that were cut (in `CutAction.cutResources`). On paste, the clipboard contents are compared element-by-element against the stored resources. This approach is more robust than a simple boolean flag because:
- It naturally handles the case where the clipboard is modified between cut and paste
- A copy operation after a cut correctly invalidates the cut state
- It is resilient against interference from other action providers


vogella:
Cut does not work across projects.

laeubi:
> Cut does not work across projects.

In what regards? As it works to move one file from one project to another with drag&drop I do not see a reason why a cut/paste can not be mapped to that operation?

vogella:
I don't remember why anymore, one of the reasons why I closed this pr

PR #3794, opened 2026-03-18
Fix race condition in AbstractReconciler.reset() event ordering

### LINKED ISSUE #2708: Random failing org.eclipse.jface.text.tests.reconciler.FastAbstractReconcilerTest#testReplacingDocumentWhenClean()

vogella:
## Summary

- Fixes flaky `FastAbstractReconcilerTest.testReplacingDocumentWhenClean` by reordering `BackgroundWorker.reset()` so that the `reconcilerReset()` hook executes before `fIsDirty` is set to true and before any queue notifications wake the background thread
- Adds `ReconcilerResetOrderingTest` that deterministically reproduces the race condition by widening the race window with a 50ms delay — this test always fails without the fix and always passes with it

## Root Cause

In `BackgroundWorker.reset()`, `reconcilerReset()` was called after `fIsDirty` was set and after `notifyAll()`/`informNotFinished()`. Two race windows existed:

1. `aboutToBeReconciledInternal()` (called before `reset()` in both `documentChanged` and `inputDocumentChanged`) triggers `signalWaitForFinish()` → `notifyAll()`, waking the background thread. It could then see `fIsDirty=true` (set at the start of `reset()`), consume the `fReset` flag, skip delay (`waitFinish=true`), and reach `process()` before `reconcilerReset()` ran.

2. Within `reset()` itself, `notifyAll()` and `informNotFinished()` were called before `reconcilerReset()`, providing additional wake-up points.

## Fix

Move `reconcilerReset()` to the very beginning of `reset()`, before `fIsDirty` is set. The background thread cannot find work (`isDirty()` returns false) and blocks in `delay()` until after the hook completes and state flags are set.

## User-visible impact

The `AbstractReconciler` drives background processing in text editors (syntax validation, spell checking, error/warning markers, code analysis). The race occurs when the document changes or is replaced (e.g., switching files, reverting). Without the reset hook firing first, `process()` can run with stale state, causing brief flickers of incorrect error markers or squiggles from a previous document. Being a race condition, this is intermittent and self-corrects on the next reconciling pass.

## Test plan

- [x] `AbstractReconcilerTest` (7 tests) — base reconciler tests, all pass
- [x] `FastAbstractReconcilerTest` (7 tests) — fast-mode tests including previously flaky `testReplacingDocumentWhenClean`, all pass
- [x] `ReconcilerResetOrderingTest` (7 tests) — new regression test with 50ms delay, deterministically fails without fix, passes with fix
- [x] Verified test fails without fix by temporarily reverting production code

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2708

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
@@ -140,14 +140,13 @@ abstract public class AbstractReconciler implements IReconciler {
 		 */
 		public void reset() {
 
+			reconcilerReset();
+
 			if (fDelay > 0) {
 
 				synchronized (this) {
 					fIsDirty= true;
 					fReset= true;
-				}
-				synchronized (fDirtyRegionQueue) {
-					fDirtyRegionQueue.notifyAll(); // wake up wait(fDelay);
 				}
 
 			} else {
@@ -155,14 +154,13 @@ abstract public class AbstractReconciler implements IReconciler {
 				synchronized (this) {
 					fIsDirty= true;
 				}
-
-				synchronized (fDirtyRegionQueue) {
-					fDirtyRegionQueue.notifyAll();
-				}
 			}
 
 			informNotFinished();
-			reconcilerReset();
+
+			synchronized (fDirtyRegionQueue) {
+				fDirtyRegionQueue.notifyAll();
+			}
 		}
 
 		/**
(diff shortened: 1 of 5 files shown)


vogella:
Test failure is reported here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3796

vogella:
Merging so that this gets in early and we can spot issue with it.

PR #3804, opened 2026-03-20
Add null-check to AbstractTableViewer and AbstractTreeViewer

fedejeanne:
Check for null before using the result of getViewerColumn(int). The JavaDoc explicitly says that the method may return null and other places in the code are already checking for null.


fedejeanne:
FTR here's the JavaDoc that I mentioned before:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/968d4cefd41e3cbf6eb3f70bc85aacdd283cf066/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/ColumnViewer.java#L192-L200

PR #3809, opened 2026-03-20
Match Section title bar color to background in dark theme

vogella:
## Summary
- Set `background-color-titlebar`, `background-color-gradient-titlebar`, and `border-color-titlebar` to `#2f2f2f` to match the Section `background-color` in the dark theme
- Previously these were `#4F5355`, creating a visible lighter band in the Section title bar area (e.g., "Unstaged Changes" in the Git Staging view) even though gradient start/end colors were identical

The Section's `background-color` was being inherited from its parent Composite as `#2f2f2f`, but the title bar properties remained at `#4F5355`, causing a color mismatch.

See #3803

## Test plan
- [ ] Verify Section title bars in dark theme (e.g., Git Staging view) render flat without a lighter band
- [ ] Verify Section title bar text and toggle icons remain visible against the new background

### DIFF
--- a/bundles/org.eclipse.ui.themes/css/dark/e4-dark_globalstyle.css
+++ b/bundles/org.eclipse.ui.themes/css/dark/e4-dark_globalstyle.css
@@ -213,11 +213,11 @@ Form {
 }
 
 Section {
-    background-color: #4F5355;
+    background-color: #2f2f2f;
     color: #AEBED0;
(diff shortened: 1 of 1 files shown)


vogella:
image

vogella:
-1 this will loose the visibility of that section header for drag and drop, we should drop the gradient but give the header a different solid color

PR #3816, opened 2026-03-24
Add preference for disabling file search on specific files

### LINKED ISSUE #3815: Provide options for disabling file search on some files
We have use cases where we would like to disable the search (main menu -> Search -> File Search) on some files.

We would like to add a way for this, similar to how derived files can be excluded from the search. We don't require UIs, since our product determines what needs to be hidden from the search - not the user.

trancexpress:
This change adds a new preference to `org.eclipse.search.core`: `search_exclusion_property`
The preference can be set to a session property name, e.g. `search_excluded_file`.
The search will then skip files for which this session property is set.

For linked files, the target file is skipped only if the session property is set for all link resources.
Each link without the session property set will result in matches, if there are any in the target file.

Example preference for product customization:
```
org.eclipse.search.core/search_exclusion_property=search_excluded_file
```

The session property can be set with:
```
file.setSessionProperty(new QualifiedName(null, "search_excluded_file"), "true");
```

Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3815

### DIFF
--- a/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -38,12 +38,14 @@ import org.eclipse.core.runtime.MultiStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.QualifiedName;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.content.IContentType;
 import org.eclipse.core.runtime.content.IContentTypeManager;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.jobs.JobGroup;
+import org.eclipse.core.runtime.preferences.IPreferencesService;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IResourceStatus;
@@ -70,6 +72,7 @@ public class TextSearchVisitor {
 
 	public static final boolean TRACING= "true".equalsIgnoreCase(Platform.getDebugOption("org.eclipse.search/perf")); //$NON-NLS-1$ //$NON-NLS-2$
 	private static final int NUMBER_OF_LOGICAL_THREADS= Runtime.getRuntime().availableProcessors();
+	private static final String EXCLUSION_PREFERENCE_NAME = "search_exclusion_property"; //$NON-NLS-1$
 
 	/**
 	 * Queue of files to be searched. IFile pointing to the same local file are
@@ -185,8 +188,6 @@ public class TextSearchVisitor {
 		}
 
 		public IStatus processFile(List<IFile> sameFiles, IProgressMonitor monitor) {
-			// A natural cleanup after the change to use JobGroups is accepted would be to move these
-			// methods to the TextSearchJob class.
 			Matcher matcher= fSearchPattern.pattern().isEmpty() ? null : fSearchPattern.matcher(""); //$NON-NLS-1$
 			IFile file = sameFiles.remove(0);
 			monitor.setTaskName(file.getFullPath().toString());
@@ -260,9 +261,7 @@ public class TextSearchVisitor {
 				if (fIsLightweightAutoRefresh && IResourceStatus.RESOURCE_NOT_FOUND == e.getStatus().getCode()) {
 					return monitor.isCanceled() ? Status.CANCEL_STATUS : Status.OK_STATUS;
 				}
-				Object[] args= { getExceptionMessage(e), file.getFullPath().makeRelative().toString() };
-				String message = MessageFormat.format(SearchCoreMessages.TextSearchVisitor_error, args);
-				return new Status(IStatus.ERROR, SearchCorePlugin.PLUGIN_ID, IStatus.ERROR, message, e);
+				return errorStatusForFile(file, e);
 			} catch (StackOverflowError e) {
 				fFatalError= true;
 				String message= SearchCoreMessages.TextSearchVisitor_patterntoocomplex0;
@@ -302,6 +301,8 @@ public class TextSearchVisitor {
 	private volatile boolean fIsLightweightAutoRefresh;
 	private final DirtyFileProvider fDirtyDiscovery;
 
+	private final QualifiedName fExclusionProperty;
+
 	public TextSearchVisitor(TextSearchRequestor collector, Pattern searchPattern, DirtyFileProvider dirtyDiscovery) {
 		fCollector= collector;
 		fDirtyDiscovery = dirtyDiscovery;
@@ -312,6 +313,10 @@ public class TextSearchVisitor {
 
 		fIsLightweightAutoRefresh= Platform.getPreferencesService().getBoolean(ResourcesPlugin.PI_RESOURCES, ResourcesPlugin.PREF_LIGHTWEIGHT_AUTO_REFRESH, false, null);
 		fileBatches = new ConcurrentLinkedQueue<>();
+
+		IPreferencesService prefs = Platform.getPreferencesService();
+		String exclusionPropertyName = prefs.getString(SearchCorePlugin.PLUGIN_ID, EXCLUSION_PREFERENCE_NAME, "", null); //$NON-NLS-1$
+		fExclusionProperty = exclusionPropertyName.isEmpty() ? null : new QualifiedName(null, exclusionPropertyName);
 	}
 
 	public IStatus search(IFile[] files, IProgressMonitor monitor) {
@@ -345,6 +350,9 @@ public class TextSearchVisitor {
 				Map<String, List<IFile>> remoteFilesByLocation = new LinkedHashMap<>();
 
 				for (IFile file : files) {
+					if (excluded(file)) {
+						continue;
+					}
 					IPath path = file.getLocation();
 					String key = path == null ? file.getLocationURI().toString() : path.toString();
 					Map<String, List<IFile>> filesByLocation = (path != null) ? localFilesByLocation
@@ -513,8 +521,14 @@ public class TextSearchVisitor {
 		return occurences;
 	}
 
+	private static Status errorStatusForFile(IFile file, CoreException e) {
+		Object[] args = { getExceptionMessage(e), file.getFullPath().makeRelative().toString() };
+		String message = MessageFormat.format(SearchCoreMessages.TextSearchVisitor_error, args);
+		Status status = new Status(IStatus.ERROR, SearchCorePlugin.PLUGIN_ID, IStatus.ERROR, message, e);
+		return status;
+	}
 
-	private String getExceptionMessage(Exception e) {
+	private static String getExceptionMessage(Exception e) {
 		String message= e.getLocalizedMessage();
 		if (message == null) {
 			return e.getClass().getName();
@@ -542,4 +556,22 @@ public class TextSearchVisitor {
 		}
 	}
 
+	private boolean excluded(IFile file) {
+		if (fExclusionProperty != null) {
+			try {
+				return file.getSessionProperty(fExclusionProperty) != null;
+			} catch (CoreException e) {
+				/*
+				 * The preference 'search_exclusion_property' indicates we
+				 * should skip files with the respective session property, but
+				 * we ran into an exception while reading the properties of the
+				 * file. Skip the file from the search, since we don't know if
+				 * the property is set or not.
+				 */
+				fStatus.add(errorStatusForFile(file, e));
+				return true;
+			}
+		}
+		return false;
+	}
 }
--- a/bundles/org.eclipse.search.core/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.search.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.search.core;singleton:=true
-Bundle-Version: 3.16.600.qualifier
+Bundle-Version: 3.16.700.qualifier
 Bundle-Activator: org.eclipse.search.internal.core.SearchCorePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/AllFileSearchTests.java
+++ b/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/AllFileSearchTests.java
@@ -28,6 +28,7 @@ import org.eclipse.search.tests.TextSearchResultTest;
 		SearchResultPageTest.class,
 		SortingTest.class,
 		TextSearchResultTest.class,
+		ExcludedFilesSearchTest.class,
 })
 public class AllFileSearchTests {
 }
--- a/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/ExcludedFilesSearchTest.java
+++ b/tests/org.eclipse.search.tests/src/org/eclipse/search/tests/filesearch/ExcludedFilesSearchTest.java
@@ -1,0 +1,284 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Simeon Andreev and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.search.tests.filesearch;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Pattern;
+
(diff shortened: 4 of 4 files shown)


trancexpress:
The test still needs adjustment, currently its only a snippet for performance testing.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -71,6 +72,11 @@ public class TextSearchVisitor {
 	public static final boolean TRACING= "true".equalsIgnoreCase(Platform.getDebugOption("org.eclipse.search/perf")); //$NON-NLS-1$ //$NON-NLS-2$
 	private static final int NUMBER_OF_LOGICAL_THREADS= Runtime.getRuntime().availableProcessors();
 
+	private static final boolean ENABLE_FORBIDDEN_FILES = Platform.getPreferencesService()
+			.getBoolean(SearchCorePlugin.PLUGIN_ID, "enable_search_forbidden_files", false, null); //$NON-NLS-1$
+	public static final QualifiedName FORBIDDEN_KEY = new QualifiedName(SearchCorePlugin.PLUGIN_ID,
+			"search_forbidden_file"); //$NON-NLS-1$
```
merks: Somehow this name seems a bit awkward starting with a verb.  I'm not sure users see it.   Something like `forbid_search` or `forbidden_search` is maybe better.    Does `_file` suggest this can only be applied to files?  Maybe I can use it to block a whole folder?  E.g., the target folder generated by an external Maven build?

trancexpress: > I'm not sure users see it.

Its not user visible.

>  Does `_file` suggest this can only be applied to files? Maybe I can use it to block a whole folder? E.g., the target folder generated by an external Maven build?

It only applies to files. I've not checked if I can apply the change also for folders, since for files it seems sufficient for our use case.

>  Something like `forbid_search` or `forbidden_search` is maybe better.

I'm not either, we are not disabling the entire search, only for some files. Maybe exclude is a better word than forbidden here, I'm not sure. I'll think about this.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -542,4 +548,7 @@ private String getCharSetName(IFile file) {
 		}
 	}
 
+	private static boolean forbidden(IFile file) throws CoreException {
+		return ENABLE_FORBIDDEN_FILES && file.getSessionProperty(FORBIDDEN_KEY) != null;
```
iloveeclipse: Looking at https://github.com/eclipse-platform/eclipse.platform.ui/issues/3817 there will be a second place where exact same mechanism can be used.

With the current proposal (session property with the "search" key) the client would be forced to set **two** session properties per file, which increases runtime overhead.

Therefore I would propose to add something like this to `IResource` (javadoc is missing)

```
	/**
	 * @since 3.24
	 */
	QualifiedName ACCESS_RESTRICTED = new QualifiedName(ResourcesPlugin.PI_RESOURCES, "restrictedAccess"); //$NON-NLS-1$

	/**
	 * @since 3.24
	 */
	default boolean isAcceessRestricted() throws CoreException {
		return getSessionProperty(ACCESS_RESTRICTED) != null;
	}

	/**
	 * @since 3.24
	 */
	default void setAccessRestricted(boolean restricted) throws CoreException {
		setSessionProperty(ACCESS_RESTRICTED, Boolean.toString(restricted));
	}
```

The name is generic enough and both local history in `FileSystemResourceManager.storeHistory(IResource)` / search here can use `file.isAccessRestricted()`, and client can use one API to set/unset the single property.

trancexpress: For 100 000 files setting the session property took 100 ms on my laptop. What exactly is the performance concern that justifies adding API for this and #3817 ?

iloveeclipse: - *two* different properties instead of one is one too much.
- doesn't work if we need to use it elsewhere
- code duplication everywhere, also at client code

trancexpress: When I see this on `IResource`, what exactly do I do with this method then?

It restricts access how? We document that file search is disabled if the `IResource` is actually `IFile` and that we don't store history, again if its an `IFile`? And potentially we keep adding documentation as we discover more use cases?

IMO this is not API, it has no clear purpose.

I can look for a place where to define a single session property, though again its not clear what the single property will be doing in future.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -71,6 +72,11 @@ public class TextSearchVisitor {
 	public static final boolean TRACING= "true".equalsIgnoreCase(Platform.getDebugOption("org.eclipse.search/perf")); //$NON-NLS-1$ //$NON-NLS-2$
 	private static final int NUMBER_OF_LOGICAL_THREADS= Runtime.getRuntime().availableProcessors();
 
+	private static final boolean ENABLE_FORBIDDEN_FILES = Platform.getPreferencesService()
```
trancexpress: This should be moved to the constructor I think, in case the preference changes at runtime. Which is done e.g. during the test.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
iloveeclipse: > And potentially we keep adding documentation as we discover more use cases?

The API doesn't need to explicitly point to the possible implementation. Javadoc can be like "The file will be accessible in terms of `isAccessible()` and visible in terms of `isHidden()` but file search, local history or other functionality for this file would be not available or limited".

BTW, I think this also better be on an `IFile` as the property is not supposed to be hierarchical.

> IMO this is not API, it has no clear purpose.

The purpose is to allow clients & IDE to limit **regular** functionality available for `IFile` in various cases. Similar API are `isHidden()` or `isDerived()` which also have no "clear" purpose, they are there to provide generic functionality that can be used.

trancexpress: What about using the preference value for the session property name?

I.e. if `org.eclipse.search.core/enable_search_forbidden_files` (or however we want to name it) is empty, the search works as usual.

If its not empty, the files are checked for the session property specified by the preference, e.g.: `org.eclipse.search.core/enable_search_forbidden_files=very_special_property`

If the session property is present for the file, the file is skipped.

I assume we'll have several preferences (one for search, one for history, who knows what else we need)? Preference behavior is fairly simple in that case, specific bits in the platform/JDT can be disabled per preference and per file. The platform code doesn't have to define property names, that can be up to the client code.

Then the client code can name the property however they want, it can use the same property or different properties for the different preferences and so on?

iloveeclipse: This sounds good. Please update PR to see ow it would look like.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
trancexpress: Preference is now called:

```
org.eclipse.search.core/search_exclusion_property=some_property_name
```

The property name can be chosen by the client code.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -191,7 +194,7 @@ public IStatus processFile(List<IFile> sameFiles, IProgressMonitor monitor) {
 			IFile file = sameFiles.remove(0);
 			monitor.setTaskName(file.getFullPath().toString());
 			try {
-				if (!fCollector.acceptFile(file) || matcher == null) {
+				if (!fCollector.acceptFile(file) || matcher == null || excluded(file)) {
 					return Status.OK_STATUS;
 				}
```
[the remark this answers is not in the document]
trancexpress: @iloveeclipse WDYT here? I'm using the same place where derived files are filtered... I think.

Change would probably look like this:

```
diff --git a/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java b/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
index 985a6fc1b3..46fb48c95e 100644
--- a/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -228,6 +228,9 @@ public class TextSearchVisitor {
                                fCollector.flushMatches(file);
 
                                for (IFile duplicateFiles : sameFiles) {
+                                       if (excluded(duplicateFiles)) {
+                                               continue;
+                                       }
                                        // reuse previous result
                                        ReusableMatchAccess matchAccess= new ReusableMatchAccess();
                                        for (TextSearchMatchAccess occurence : occurences) {
```

We'd have to add a test for it though.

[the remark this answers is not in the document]
trancexpress: I've adjusted the code to filter the files prior to any of the original code. If one file is the target of many links, and some of the links don't have the session property set, the target file will be searched.

trancexpress:
Please review, from my POV the change is ready.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -542,4 +568,20 @@ private String getCharSetName(IFile file) {
 		}
 	}
 
+	private boolean excluded(IFile file) throws CoreException {
+		return fExclusionProperty != null && file.getSessionProperty(fExclusionProperty) != null;
+	}
+
+	private static QualifiedName toQualifedName(String name) {
```
trancexpress: This code and reading the preference we could use in multiple places. For the next change I'll check if there is a good location for it... I didn't find one while grepping the repo.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -70,6 +72,7 @@ public class TextSearchVisitor {
 
 	public static final boolean TRACING= "true".equalsIgnoreCase(Platform.getDebugOption("org.eclipse.search/perf")); //$NON-NLS-1$ //$NON-NLS-2$
 	private static final int NUMBER_OF_LOGICAL_THREADS= Runtime.getRuntime().availableProcessors();
+	private static final String SEARCH_EXCLUSION_PREFERENCE = "search_exclusion_property"; //$NON-NLS-1$
```
iloveeclipse: Inconsistent wording: `SEARCH_EXCLUSION_PREFERENCE` vs `search_exclusion_property`

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -187,8 +190,21 @@ protected IStatus run(IProgressMonitor inner) {
 		public IStatus processFile(List<IFile> sameFiles, IProgressMonitor monitor) {
 			// A natural cleanup after the change to use JobGroups is accepted would be to move these
 			// methods to the TextSearchJob class.
+			List<IFile> files = new ArrayList<>(sameFiles.size());
```
iloveeclipse: By default we would create ArrayList which is not needed in most cases, so in the workspace search over 100.000 files we will have 100.000 extra objects created & GC'd for nothing. I would lazy create this list or guard the list inspection with `if fExclusionProperty != null`

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
iloveeclipse: Not sure why do we need to **parse** the value? Is `new QualifiedName(null, name)` not enough for now? Who would need the "qualifier" part? I would just return `new QualifiedName(null, name)` and if there will be any request for qualifier, it could be added later. That would simplify code and runtime.

trancexpress:
```
Continuous Integration / Test Results (pull_request) — 1 fail, 243 skipped, 7 628 pass in 58m 56s
```

I have no clue where this fail is.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -187,8 +190,21 @@ protected IStatus run(IProgressMonitor inner) {
 		public IStatus processFile(List<IFile> sameFiles, IProgressMonitor monitor) {
```
iloveeclipse, line 190: Have you considered to filter files at the very beginning, **before** the search qeues will be populated and `TextSearchJob` created & started?

https://github.com/eclipse-platform/eclipse.platform.ui/blob/b4e388feec421e125b7502b1162884a892a7a74f/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java#L347-L354

The point is: we can skip all the queue creation / jobs starting / extra cals to `processFile()` resulting in doing nothing if we filter the files earlier.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
trancexpress: `SEARCH_EXCLUSION_PREFERENCE` is the variable with which we refer to the preference key. `search_exclusion_property` is the actual name of the preference.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
trancexpress: I'm not sure this is a good idea, since reading the session property can throw an exception.

With the current code, that error is added to a `MultiStatus`.

With your suggestion we would either have to ignore the file silently (maybe with a logged error) or cancel the entire search.

The goal of the change is to exclude some files from the search, not try to improve its performance in rare cases?

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
iloveeclipse: The one is named `PREFERENCE` and another one is `property`, this is inconsistent.

trancexpress: Right now this enables the preference:

```
org.eclipse.search.core/search_exclusion_property=search_excluded_file
```

The value of the preference is a property, hinted by the key name.

And this sets the preference:

```
prefs.getString(SearchCorePlugin.PLUGIN_ID, SEARCH_EXCLUSION_PREFERENCE, "", null);
```

But the key for the preference is not a property itself.

Why should the two be "consistent", when its two different things? One is a preference key. The other is the actual preference.

Should I call it `SEARCH_EXCLUSION_PREFERENCE_NAME` or `SEARCH_EXCLUSION_PREFERENCE_KEY`? Resp. `EXCLUSION_PREFERENCE_NAME` or `EXCLUSION_PREFERENCE_KEY`?

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
iloveeclipse: I'm just thinking about 100.000 files we might want to skip. Ideally we would not fill in queues with them.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
iloveeclipse: > Why should the two be "consistent", when its two different things?

Because people that read the code don't know all that. They see seemengly "similar" but yet different names and start ask questions "why"?

> EXCLUSION_PREFERENCE_NAME or EXCLUSION_PREFERENCE_KEY

This is better.

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
trancexpress: How about now?

bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -345,6 +350,14 @@ public IStatus search(IFile[] files, IProgressMonitor monitor) {
 				Map<String, List<IFile>> remoteFilesByLocation = new LinkedHashMap<>();
 
 				for (IFile file : files) {
+					try {
+						if (excluded(file)) {
+							continue;
+						}
+					} catch (CoreException e) {
+						fStatus.add(errorStatusForFile(file, e));
+						continue;
```
trancexpress, line 354: This means, we have the new preference set, but were unable to read session properties for the file.

We don't know if we should skip it, but we don't want to risk not skipping it. There is no change from the previous proposal in this regard.

Should I add a comment for this, when I think about this?

iloveeclipse:
Looks OK for me, just a small thing I would like to ask for all future commits:

**please** add **full** url of the ticket in the commit message, not just a number, so that later on one could directly jump from the history view into github ticket or PR.

So instead of this:
image

do this:

image

You can't believe how this small change can help in my daily work.

iloveeclipse (approved):

PR #3819, opened 2026-03-25
Allow tests to wait on reconciler threads to finish

### LINKED ISSUE #3818: Allow tests to wait on reconciler threads to finish
While looking into various JDT UI test fails, e.g. https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2287, I noticed that the we have no option to wait on the thread started here: `org.eclipse.jface.text.reconciler.AbstractReconciler.BackgroundWorker`

E.g. I can get the addition of the project info at this stack trace:

```
"org.eclipse.jdt.internal.ui.text.JavaReconciler" #86 [165284] daemon prio=1 os_prio=0 cpu=0.86ms elapsed=66.00s tid=0x00007fe978058760 nid=165284 at breakpoint [0x00007fe9037f9000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:585)
        at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:284)
        at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:270)
        at org.eclipse.jdt.internal.core.JavaProject.getJavaProjectElementInfo(JavaProject.java:2088)
        at org.eclipse.jdt.internal.core.JavaProject.newNameLookup(JavaProject.java:2823)
        at org.eclipse.jdt.internal.core.SearchableEnvironment.<init>(SearchableEnvironment.java:117)
        at org.eclipse.jdt.internal.core.SearchableEnvironment.<init>(SearchableEnvironment.java:158)
        at org.eclipse.jdt.internal.core.CancelableNameEnvironment.<init>(CancelableNameEnvironment.java:32)
        at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:276)
        at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:375)
        at org.eclipse.jdt.internal.core.ReconcileWorkingCopyOperation.makeConsistent(ReconcileWorkingCopyOperation.java:244)
        at org.eclipse.jdt.internal.core.ReconcileWorkingCopyOperation.executeOperation(ReconcileWorkingCopyOperation.java:108)
        at org.eclipse.jdt.internal.core.JavaModelOperation.run(JavaModelOperation.java:751)
        at org.eclipse.jdt.internal.core.JavaModelOperation.runOperation(JavaModelOperation.java:816)
        at org.eclipse.jdt.internal.core.CompilationUnit.reconcile(CompilationUnit.java:1431)
        at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:132)
        at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy$1.run(JavaReconcilingStrategy.java:94)
        at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:91)
        at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:158)
        at org.eclipse.jdt.internal.ui.text.CompositeReconcilingStrategy.reconcile(CompositeReconcilingStrategy.java:94)
        at org.eclipse.jdt.internal.ui.text.JavaCompositeReconcilingStrategy.reconcile(JavaCompositeReconcilingStrategy.java:107)
        at org.eclipse.jface.text.reconciler.MonoReconciler.process(MonoReconciler.java:77)
        at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundWorker.run(AbstractReconciler.java:205)
        at java.lang.Thread.runWith(java.base@21.0.10/Thread.java:1596)
        at java.lang.Thread.run(java.base@21.0.10/Thread.java:1583)
```

While the entire `org.eclipse.jdt.ui.tests.core.ImportOrganizeTest` finishes.

We should have some option to be able to wait on this thread to exit, to ensure tests don't run with parallel tasks from previous tests.

trancexpress:
This change adds a job with a queue of `AbstractReconciler.BackgroundWorker` threads. Whenever a reconciler is uninstalled, its `BackgroundWorker` thread is added to the queue and the job is scheduled (if not running). The job will wait until the background threads finish.

The new job has the same family as other reconciler jobs:

```
ReconcilerJobFamilies.FAMILY_RECONCILER
```

Clients (mainly tests) that open an editor now have the option to wait on this job family, ensuring reconciler work is done before proceeding.

Fixes: #3818

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
@@ -13,6 +13,10 @@
  *******************************************************************************/
 package org.eclipse.jface.text.reconciler;
 
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
@@ -74,12 +78,16 @@ abstract public class AbstractReconciler implements IReconciler {
 
 		private final String fName;
 
-		private boolean fIsAlive;
+		private volatile boolean fIsAlive;
 
 		private volatile Thread fThread;
 
+		/** A latch to wait on until the background thread finishes. */
+		private final CountDownLatch fFinishLatch;
+
 		public BackgroundWorker(String name) {
 			fName= name;
+			fFinishLatch= new CountDownLatch(1);
 		}
 
 		/**
@@ -116,8 +124,8 @@ abstract public class AbstractReconciler implements IReconciler {
 		}
 
 		/**
-		 * Suspends the caller of this method until this background thread has
-		 * emptied the dirty region queue.
+         * Suspends the caller of this method until this background thread has
+         * emptied the dirty region queue.
 		 */
 		public void suspendCallerWhileDirty() {
 			AbstractReconciler.this.signalWaitForFinish();
@@ -216,6 +224,7 @@ abstract public class AbstractReconciler implements IReconciler {
 				}
 			} finally {
 				fIsAlive= false;
+				fFinishLatch.countDown();
 			}
 		}
 
@@ -339,6 +348,8 @@ abstract public class AbstractReconciler implements IReconciler {
 		}
 	}
 
+	private static final WaitForThreadsJob waitForFinishJob= new WaitForThreadsJob();
+
 	/** Queue to manage the changes applied to the text viewer. */
 	private DirtyRegionQueue fDirtyRegionQueue;
 	/** The background thread. */
@@ -353,6 +364,7 @@ abstract public class AbstractReconciler implements IReconciler {
 	private boolean fIsIncrementalReconciler= true;
 	/** The progress monitor used by this reconciler. */
 	private IProgressMonitor fProgressMonitor;
+
 	/**
 	 * Tells whether this reconciler is allowed to modify the document.
 	 * @since 3.2
@@ -531,6 +543,8 @@ abstract public class AbstractReconciler implements IReconciler {
 				BackgroundWorker bt= fWorker;
 				fWorker= null;
 				bt.cancel();
+				// allow clients (e.g. tests) to wait on the reconciler thread to finish
+				waitForFinishJob.addTask(bt);
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


trancexpress:
Lets see if something breaks during tests.

Technically all we need is just a list of threads or something similar, and a method which waits until that list is empty... The job is an overkill, since all it does is wait. But a system job that waits is probably simplest...

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java
```diff
@@ -531,6 +543,8 @@ public void uninstall() {
 				BackgroundWorker bt= fWorker;
 				fWorker= null;
 				bt.cancel();
+				// allow clients (e.g. tests) to wait on the reconciler thread to finish
+				waitForFinishJob.addTask(bt);
```
fedejeanne, line 547: I think this will _force_ (not allow) clients to wait. That seems excessive, especially because the reconciler explicitly wanted the thread to be a daemon i.e. not block the shutdown operation

https://github.com/trancexpress/eclipse.platform.ui/blob/1ec8bc3f5ce623220f6cce214d0ebb7917de6bd4/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/AbstractReconciler.java#L257

Maybe this could be made an opt-in feature?

trancexpress: Technically we only need this during tests... I can imagine other code that might need this, considering the code can corrupt the JDT model... But it should be very rare.

So opt-in would be fine.

I want to experiment some more though. Maybe we just go over all threads and wait for the reconciler thread to be gone... Its tests after all...

fedejeanne: One further improvement I can think of is trying to make the reconciler more aware and let it stop working once the thread has been asked to stop.
I didn't work specifically with the reconciler so I'm not an expert (take these hints always with a pinch of salt) but I did work with the `CoreASTProvider` & Co., more specifically I looked at this method in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2406: `org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(ITypeRoot, WAIT_FLAG, IProgressMonitor)`. Several use-cases go through it and IIRC, so does the reconciler. The method is a bit brittle and it sometimes leads to long waits (and UI freezes, if the use-case just happens to run in the main thread). Since the reconciler also uses that method, it may also fall victim of a freeze i.e. wait 30+ seconds in this line:

```java
fWaitLock.wait(30000); // XXX: The 30 seconds timeout is an attempt to at least avoid a deadlock. See https://bugs.eclipse.org/366048#c21
```

... which may indicate why it was set to run as a _daemon_ thread in the first place.

What I'm trying to say is: please thread lightly with the reconciler and don't let it affect the shutdown session (especially in production code) because you may inadvertently delay shutting down the whole workbench.

HTH!

trancexpress: The shutdown is fairly simple, if this job is cancelled it can just exit. It can add a `IWorkbenchListener.preShutdown(IWorkbench, boolean)` and cancel itself.

I'm more concerned that this is not really what a job is for. But if its not a job we are inventing new ways to wait in tests... I.e. new code that needs to be added to failing tests.

I'll write some proposals on https://github.com/eclipse-platform/eclipse.platform.ui/issues/3818 when I have a few.

fedejeanne: 👍 
I see that you already thought of some alternatives :-)
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2897

Issue #3800, opened 2026-03-20
Add tracing to text stores


trancexpress:
See: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/79

And in particular:

```
org.eclipse.text.edits.MalformedTreeException: End position lies outside document range
	at org.eclipse.text.edits.TextEditProcessor.checkIntegrityDo(TextEditProcessor.java:182)
	at org.eclipse.text.edits.TextEdit.dispatchCheckIntegrity(TextEdit.java:757)
	at org.eclipse.text.edits.TextEditProcessor.performEdits(TextEditProcessor.java:154)
	at org.eclipse.ltk.core.refactoring.TextChange.performEdits(TextChange.java:278)
	at org.eclipse.ltk.core.refactoring.TextFileChange.access$0(TextFileChange.java:1)
	at org.eclipse.ltk.core.refactoring.TextFileChange.lambda$0(TextFileChange.java:284)
	at org.eclipse.ui.internal.editors.text.UISynchronizationContext.run(UISynchronizationContext.java:35)
	at org.eclipse.core.internal.filebuffers.TextFileBufferManager.execute(TextFileBufferManager.java:614)
	at org.eclipse.ltk.core.refactoring.TextFileChange.performEdits(TextFileChange.java:295)
	at org.eclipse.ltk.core.refactoring.TextChange.perform(TextChange.java:239)
	at org.eclipse.jdt.internal.corext.fix.CleanUpRefactoring$CleanUpChange.perform(CleanUpRefactoring.java:168)
	at org.eclipse.ltk.core.refactoring.PerformChangeOperation.lambda$0(PerformChangeOperation.java:250)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2505)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2533)
	at org.eclipse.ltk.core.refactoring.PerformChangeOperation.executeChange(PerformChangeOperation.java:289)
	at org.eclipse.ltk.core.refactoring.PerformChangeOperation.run(PerformChangeOperation.java:214)
	at org.eclipse.jdt.internal.corext.fix.CleanUpPostSaveListener.saved(CleanUpPostSaveListener.java:412)
	at org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider$4.run(CompilationUnitDocumentProvider.java:1610)
[...]

Issue #3796, opened 2026-03-19
SearchResultPageTest.testTableNavigation fail sporadically


trancexpress:
See: https://download.eclipse.org/eclipse/downloads/drops4/I20260318-1800/reports.html#tests

https://download.eclipse.org/eclipse/downloads/drops4/I20260318-1800/testresults/html/org.eclipse.search.tests_ep440I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html

```
expected: <1> but was: <0>

org.opentest4j.AssertionFailedError: expected: <1> but was: <0>
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:531)
at org.eclipse.search.tests.filesearch.SearchResultPageTest.testTableNavigation(SearchResultPageTest.java:156)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596) 
```

Workspace log: https://download.eclipse.org/eclipse/downloads/drops4/I20260318-1800/testresults/ep440I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21/org.eclipse.search.tests.AllSearchTests.txt

```
!SESSION 2026-03-18 19:21:23.463 -----------------------------------------------
eclipse.buildId=4.40.0.I20260318-1800
java.version=21.0.5
java.vendor=Eclipse Adoptium
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_US
[...]

PR #3821, opened 2026-03-26
Fix debug tracing typos in CopyOnWriteTextStore

trancexpress:
Removes unnecessary "{0}", "{1}" and "{2}" from debug traces in CopyOnWriteTextStore.

See: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3800


trancexpress:
The one fail is from here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3796

PR #3823, opened 2026-03-26
Support new SWT dirty indicator in StackRenderer

vogella:
This change enables the new bullet-style dirty indicator for tabs in StackRenderer, which overlays the close button.

It implements the following:
- Added a preference to enable this functionality in the Views preference page.
- StackRenderer now calls setDirtyIndicatorStyle at the folder level.
- setShowDirty is called on individual tab items when dirty state changes.
- Preference changes at runtime are handled, updating all visible stacks.
- Traditional '*' prefix is skipped when the new bullet-style indicator is active.

This PR works with the changes introduced in https://github.com/eclipse-platform/eclipse.platform.swt/pull/3141.


merks:
Is it intentional that this PR includes 2 commits one of which is about rounded corners?  It's kind of confusing...

vogella:
> Is it intentional that this PR includes 2 commits one of which is about rounded corners? It's kind of confusing...

fixed

vogella:
@schneidermic0 any feedback here? Can you help with testing this change?

vogella:
I used this with a local build and it looks good.

image

Needs to be activated for now, maybe we want to change the default later.

image

vogella:
Light theme (with hover over close)

image

vogella:
I personally think the black is a bit to strong for the light theme but that is currently hard-coded by SWT. Planning to merge tomorrow so that others can easily test. 

This is not yet the default, hence no  change in the default L&F (without setting the preference).

azoitl:
@vogella I saw the screenshot for this in the new and noteworthy and noticed a difference between light and dark theme. In the dark theme the inactive tab the dot is greyed out in the light theme it is the same black for active and inactive tab. Any chance that we can improve that?

vogella:
Yes. Can you create a short issue with some screenshots which I can use to understand your requirement? Implemention will be trivial.

azoitl:
> Yes. Can you create a short issue with some screenshots which I can use to understand your requirement? Implemention will be trivial.

Done in eclipse-platform/eclipse.platform.swt#3278.

PR #3824, opened 2026-03-27
Revert "Add preference for disabling file search on specific files"

trancexpress:
Design of the new preference is questioned by platform committers, reverting for now.

This reverts commit f04b19d4aa6b3c3d8276966baf64eb72de9a56f8.


akurtakov:
Please hold off reverting until the issue is understood and a path forward is decided to reduce noise.

PR #3826, opened 2026-03-27
Add inline filter bar to Project Explorer

vogella:
## Summary

- Adds a collapsible filter text field to the Project Explorer, toggled by a new toolbar button (checkbox-style, sits next to the existing filter-dialog button).
- When active, the field uses the e4 `FilteredTree` / `PatternFilter` infrastructure to narrow the visible tree to files and folders matching the typed pattern — with the existing 200 ms debounce and soft expansion time-limit to protect performance on large workspaces.
- The filter bar is **hidden by default** (`setShowFilterControls(false)`) and restores focus to the tree when dismissed.

## Architecture

| Component | Change |
|-----------|--------|
| `ProjectExplorerFilteredTree` (new) | `e4 FilteredTree` subclass; overrides `doCreateTreeViewer()` to return a `CommonViewer` so all Common Navigator content/label/filter services work unchanged |
| `EmptyWorkspaceHelper` | New `setNonEmptyStackControl(Control)` method so the `StackLayout` top-control is the `FilteredTree` composite while `SWT.EmptinessChanged` still tracks the inner `Tree` widget |
| `ProjectExplorer` | Overrides `createCommonViewerObject()` to create the `FilteredTree` and return its embedded `CommonViewer` |
| `ProjectExplorerFilterActionGroup` | New `AS_CHECK_BOX` toggle action wired to `FilteredTree.setShowFilterControls()` |

## Test plan

- [ ] Open Project Explorer — filter bar must **not** be visible on startup
- [ ] Click the toggle button — filter bar appears and receives focus
- [ ] Type a pattern (e.g. `*.java`) — tree narrows to matching resources
- [ ] Clear the text / click toggle again — full tree is restored
- [ ] Open Project Explorer with an **empty** workspace — the "no projects" helper screen still appears correctly
- [ ] Verify the existing filter-dialog button still works alongside the new toggle

## Notes

This is a draft PR to invite early feedback on the approach before finalising icons and adding keyboard shortcut (Ctrl+F) support.


vogella:
I think filter would not be helpful here, as the user most likely want to expand the filtered element and in this case the filter should not hide elements.

What would be nice is a Search field, maybe customizable for Projects or other elements.

PR #3833, opened 2026-03-31
Remove Classic theme

vogella:
## Summary

- Removes the Classic theme (`org.eclipse.e4.ui.css.theme.e4_classic`) including its CSS file, `plugin.xml` registration, and `plugin.properties` label
- Removes the backward-compatibility mapping from the old Windows Classic theme ID to `e4_classic` (added for bugs 562794/563601) since it is no longer needed
- Updates the schema documentation example to reference the Light theme instead

## Fallback behavior for existing users

Users who had the Classic theme selected will be silently migrated to the Light theme on next startup. `ThemeEngine.restore()` handles this: when the stored preference ID no longer matches any registered theme and does not contain `"dark"`, it falls back to the caller-supplied alternate theme (the Light theme).

## Test plan

- [x] Start Eclipse with a workspace that previously used the Classic theme — verify it starts with the Light theme
- [x] Verify the Classic theme no longer appears in Preferences > General > Appearance
- [x] Verify Dark theme selection and OS dark mode detection still work correctly


vogella:
See discussion in https://github.com/eclipse-platform/eclipse.platform.swt/issues/3150

vogella:
Test plan finished, ready to be merged after build verification

vogella:
Resource test failure is unrelated, see https://github.com/eclipse-platform/eclipse.platform.ui/pull/3831

akurtakov:
I would appreciate a follow up commit that bumps the minor version for this and other cases where UX changes.

PR #3837, opened 2026-03-31
Remove unwanted single quote from links

Neethika-sumesh12:
This pr removes unwanted single quotes from the label links shown in Preference pages and make links in preference page consistent

before : 
before

after :
after


vogella:
LGTM, please merge after build is green

PR #3841, opened 2026-03-31
Use IFile.isContentRestricted() when restricting file search

### LINKED ISSUE #3815: Provide options for disabling file search on some files

trancexpress:
This change adjusts the initial change for file search restrictions, to use the newly added API: `IFile.isContentRestricted()`

Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3815

### DIFF
--- a/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/bundles/org.eclipse.search.core/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -38,7 +38,6 @@ import org.eclipse.core.runtime.MultiStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Platform;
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/resource/CompositeImageDescriptor.java
```diff
@@ -122,8 +122,8 @@ private final class CachedDescriptorImageDataProvider extends CachedImageDataPro
 		ImageData cached;
 		int cachedZoom;
 
-		private CachedDescriptorImageDataProvider(ImageDescriptor descriptor) {
-			this.descriptor = Objects.requireNonNull(descriptor);
+		private CachedDescriptorImageDataProvider(ImageDescriptor d) {
```
iloveeclipse: unrelated change

iloveeclipse (approved):
LGTM

PR #3844, opened 2026-03-31
Add support for line background to sticky scrolling

wahlbrink:
Some text editors, like EGit's unified diff viewer, uses line background to style the editor's text. This change adds support to specify the line background of sticky lines too, so that they can be styled like in the editor.

UnifiedDiffEditor-StickyLines

### DIFF
--- a/bundles/org.eclipse.ui.editors/.settings/.api_filters
+++ b/bundles/org.eclipse.ui.editors/.settings/.api_filters
@@ -17,4 +17,12 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="src/org/eclipse/ui/texteditor/stickyscroll/IStickyLine.java" type="org.eclipse.ui.texteditor.stickyscroll.IStickyLine">
+        <filter id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.ui.texteditor.stickyscroll.IStickyLine"/>
+                <message_argument value="getBackgroundColor()"/>
+            </message_arguments>
+        </filter>
+    </resource>
 </component>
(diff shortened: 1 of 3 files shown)


Christopher-Hermann (commented):
Coding looks good, thanks. But I guess you need to update the version in the manifest `org.eclipse.ui.editors` to `3.23.0` since the interface has changed

wahlbrink:
> Coding looks good, thanks. But I guess you need to update the version in the manifest `org.eclipse.ui.editors` to `3.23.0` since the interface has changed

In 2026-03/4.39 the version of `org.eclipse.ui.editors` was 3.21.0, so I think 3.22 is correct?

The APITools-check criticizes that an method was added to an interface (and wants to increment the major version -> 4.x). Does this rule also apply to added default methods?
If yes, I have to move the method to a new interface `IStickyLineExtension`.

iloveeclipse:
> In 2026-03/4.39 the version of `org.eclipse.ui.editors` was 3.21.0, so I think 3.22 is correct?

Yes. 

You can add default methods in interface, no major bump is needed, but please "mute" the error in the IDE by adding API filter (and add the changed filter to the PR).

merks:
> If yes, I have to move the method to a new interface `IStickyLineExtension`.

It’s also permissible to create a commented filter for such an error(using a quickfix in the IDE). I’ve not looked in detail at the changes.  To me that seems better, but it’s good to get a second opinion. Not sure who.  

@iloveeclipse 

do you have an opinion or know whom best to ask?

merks:
I was slower on my iPhone. It looks like @iloveeclipse and I agree on a filter. Thanks for tackling this issue!

iloveeclipse:
> do you have an opinion or know whom best to ask?

I believe I've answered already but you've missed my comment?

merks:
> you've missed my comment?

I think we need a synchronized block around the comment authoring method. 🤪

vogella:
@wahlbrink please update (rebase) so that we can get this in, would be nice if https://eclipse.gerrithub.io/c/eclipse-egit/egit/+/1237236 can use this.

wahlbrink:
@vogella ?

vogella:
> @vogella ?

This seem to break the hover highlight of the sticky lines. Please add tests for that and fix.

PR #3850, opened 2026-04-01
Fix test in response to SWT change in CTabFolder

vogella:
getSimple is now a no-opt

https://github.com/eclipse-platform/eclipse.platform.swt/pull/3168


vogella:
@HeikoKlare this needs to be merged, otherwise we have an test error in tomorrows build (OMG!!!! - just image that horrible scenario).

HeikoKlare (approved):
Good catch! But the test should now fail in the CI of this PR.

vogella:
> Good catch! But the test should now fail in the CI of this PR.

Moving away from keyboard, so you have to decide to merge or to wait with the merge

HeikoKlare:
The change adapts a test to modifications in SWT, which are not yet represented in TP and thus makes the test fail. But without this change, the test will fail in the coming I-Build test execution. Thus merging now despite the (expected) test failures in PR checks in order to have the I-Build tests succeed,

PR #3852, opened 2026-04-01
Remove swt-simple setting from CSS

vogella:
simple is no-opt since
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3168


vogella:
Is this wrong the classic theme has been removed and must be removed from this commit

PR #3854, opened 2026-04-02
Allow user to set default theme

vogella:
During theme switch the user can now set the selected theme as default This setting is saved as configuration scope and loaded for new workspaces unless the user has a theme set for this specific workspace. Add 'Use as default theme' checkbox to theme
restart dialog


vogella:
Configuration scope is hard to test with runtime Eclipse, as each runtime Eclipse uses its own "configuration scope". So to test this change, you have to build the IDE with the new bundles and start this new instance.

vogella:
To test, I needed to build the aggregator. Start Eclipse switch theme and select: use as default in the popup:

1.) Start Eclipse and change Theme

image

Now new workspaces or workspaces without configured theme will use the new default. Workspaces which have an explicit theme set will not be affected

vogella:
Plan to merge early M2, to late for M1.

PR #3855, opened 2026-04-02
Style workspace selection dialog with default theme setting

vogella:
This PR applies the user's preferred theme to the workspace selection dialog before the workbench and its ThemeEngine start up, so the dialog matches the IDE theme even at first launch.

**Implementation:**
- `initializeDefaultTheme()` reads the theme preference scoped to the current product or application ID (with no cross-product fallback — if no product ID is present, the generic preference is used)
- A display-level `SWT.Show` listener applies dark styles to any shell as it appears
- `createContents()` in the `ChooseWorkspaceDialog` subclass applies styles a second time after all controls are constructed, catching widgets not yet present at `SWT.Show` time
- Early styling is removed via `resetEarlyDarkTheme()` before the workbench starts so the ThemeEngine manages the theme from that point on
- `Color` objects use the modern display-free `Color(r, g, b)` constructor

### DIFF
--- a/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
+++ b/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
@@ -36,12 +36,15 @@ import java.util.Properties;
 
 import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.IExecutableExtension;
+import org.eclipse.core.runtime.IProduct;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.preferences.ConfigurationScope;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.UserScope;
 import org.eclipse.equinox.app.IApplication;
 import org.eclipse.equinox.app.IApplicationContext;
 import org.eclipse.jface.dialogs.IDialogConstants;
@@ -52,14 +55,18 @@ import org.eclipse.jface.window.Window;
 import org.eclipse.osgi.service.datalocation.Location;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.layout.FillLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Link;
+import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.forms.widgets.ExpandableComposite;
 import org.eclipse.ui.internal.Workbench;
 import org.eclipse.ui.internal.WorkbenchPlugin;
 import org.eclipse.ui.internal.WorkspaceLock;
@@ -93,6 +100,9 @@ public class IDEApplication implements IApplication, IExecutableExtension {
 
 
 	private static final String USER_NAME = "user.name"; //$NON-NLS-1$
+
+	private boolean isDark;
+	private Listener darkThemeShowListener;
 
 	// Use the branding plug-in of the platform feature since this is most likely
 	// to change on an update of the IDE.
@@ -145,6 +155,9 @@ public class IDEApplication implements IApplication, IExecutableExtension {
 		Job.getJobManager().suspend();
 
 		Display display = createDisplay();
+
+		initializeDefaultTheme(display);
+
 		// processor must be created before we start event loop
 		DelayedEventsProcessor processor = new DelayedEventsProcessor(display);
 
@@ -166,6 +179,10 @@ public class IDEApplication implements IApplication, IExecutableExtension {
 				WorkbenchPlugin.unsetSplashShell(display);
 				return instanceLocationCheck;
 			}
+
+			// Reset early dark theme styling before the workbench starts;
+			// the ThemeEngine will apply the correct theme from here on.
+			resetEarlyDarkTheme(display);
 
 			// create the workbench with this advisor and run it until it exits
 			// N.B. createWorkbench remembers the advisor, and also registers
@@ -587,6 +604,14 @@ public class IDEApplication implements IApplication, IExecutableExtension {
 				return null;
 			}
 
+			@Override
+			protected Control createContents(Composite parent) {
+				Control contents = super.createContents(parent);
+				if (isDark) {
+					applyDarkStyles(getShell());
+				}
+				return contents;
+			}
 		}.prompt(force);
 	}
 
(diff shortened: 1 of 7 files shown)


bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -145,6 +157,35 @@ public Object start(IApplicationContext appContext) throws Exception {
 		Job.getJobManager().suspend();
 
 		Display display = createDisplay();
+
```
iloveeclipse, line 158: Please once done, extract into a **protected** method, so whoever extends this (non API, I know) application can overwrite the code if needed.

vogella:
Fixed a crash introduced in the previous version of this commit.

The original code called `PartRenderingEngine.initializeStyling()` before the workspace picker is shown. That method internally accesses `InstanceScope` preferences, which requires the workspace location to be set — causing an `IllegalStateException`.

The fix uses `IThemeManager.getEngineForDisplay()` + `IThemeEngine.setTheme()` directly instead (the same pattern used by `BootstrapTheme3x`). This activates the CSS theme engine without touching instance-scoped preferences, so it works safely before workspace selection.

Also fixed: OSGi service reference now released via `ungetService()` in a `finally` block, and the `org.eclipse.e4.ui.workbench.swt` bundle dependency (which exposed internal API) is removed.

vogella:
Updated approach: dropped the CSS theme engine for the workspace dialog entirely.

The dark theme CSS uses symbolic color variables (`#org-eclipse-ui-workbench-DARK_BACKGROUND` etc.) that are resolved via `ColorAndFontProviderImpl`, which calls `Workbench.getInstance().getThemeManager()`. Since `Workbench.getInstance()` is `null` before the workbench starts, these colors can never resolve at this stage — so CSS styling of the pre-workspace dialog is not feasible.

New approach: if the configured theme ID contains "dark", directly set SWT colors on the dialog's widget tree using the literal RGB values from the dark theme CSS (`#48484c` background, `#eeeeee` foreground). Colors are allocated once on the shell and disposed via a `DisposeListener`.

vogella:
First iteration: 

image

vogella:
Getter better: 

image

vogella:
image

Browse also uses  the default theme: 

image

vogella:
Test in a local Eclipse SDK build and I do not see any issues with it. Planning to merge today

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -587,6 +604,14 @@ protected Shell getParentShell() {
...
 			}
 
+			@Override
+			protected Control createContents(Composite parent) {
+				Control contents = super.createContents(parent);
+				if (isDark) {
+					applyDarkStyles(getShell());
+				}
+				return contents;
+			}
```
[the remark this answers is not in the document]
vogella: Fixed. Removed the explicit `applyDarkStyles(getShell())` call from `createContents` — the global `SWT.Show` listener already handles it.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -852,6 +877,90 @@ protected static Version toMajorMinorVersion(Version version) {
...
+	private void applyDarkStyles(Shell shell) {
+		Color bg = new Color(shell.getDisplay(), 72, 72, 76); // #48484c
+		Color fg = new Color(shell.getDisplay(), 238, 238, 238); // #eeeeee
+		Color linkColor = new Color(shell.getDisplay(), 111, 197, 238); // #6FC5EE
+		shell.setBackground(bg);
+		shell.setForeground(fg);
+		applyStylesRecursive(shell, bg, fg, linkColor);
+		shell.addDisposeListener(e -> {
+			bg.dispose();
+			fg.dispose();
```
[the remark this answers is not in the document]
vogella: SWT Colors do not need to be disposed — removed all dispose calls.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -852,6 +877,90 @@ protected static Version toMajorMinorVersion(Version version) {
...
+		});
+	}
+
+	private void applyStylesRecursive(Control control, Color bg, Color fg, Color linkColor) {
+		control.setBackground(bg);
+		if (control instanceof Link link) {
+			link.setLinkForeground(linkColor);
+		} else {
+			control.setForeground(fg);
+
```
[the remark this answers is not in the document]
vogella: Not applicable here — our Link controls only contain anchor text, so there is no unlinked text that could become unreadable.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
vogella: Already done — `initializeDefaultTheme` and `resetEarlyDarkTheme` are both `protected`.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -852,6 +873,86 @@ protected static Version toMajorMinorVersion(Version version) {
...
+	private static String getProductOrApplicationId() {
+		IProduct product = Platform.getProduct();
+		if (product != null) {
+			return product.getId();
+		}
+		return System.getProperty("eclipse.application"); //$NON-NLS-1$
+	}
+
+	private void applyDarkStyles(Shell shell) {
+		Color bg = new Color(shell.getDisplay(), 72, 72, 76); // #48484c
```
iloveeclipse: Please use "modern" Color constructors like `new Color(72, 72, 76)`

vogella: https://github.com/eclipse-platform/eclipse.platform.swt/issues/3232

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -852,6 +873,86 @@ protected static Version toMajorMinorVersion(Version version) {
...
 	}
 
+	protected void initializeDefaultTheme(Display display) {
+		IEclipsePreferences themeNode = UserScope.INSTANCE.getNode("org.eclipse.e4.ui.css.swt.theme"); //$NON-NLS-1$
+		String productOrAppId = getProductOrApplicationId();
+		String defaultThemeId = null;
+		if (productOrAppId != null) {
+			defaultThemeId = themeNode.node(productOrAppId).get("themeid", null); //$NON-NLS-1$
+		}
+		if (defaultThemeId == null) {
```
iloveeclipse: should `if (defaultThemeId == null)` be inside the "else" branch of the product?
Means, if product never defined the theme, now it can change the theme used because some other Eclipse based application defined the theme?

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
```diff
@@ -852,6 +873,86 @@ protected static Version toMajorMinorVersion(Version version) {
...
+			defaultThemeId = themeNode.get("themeid", null); //$NON-NLS-1$
+		}
+		isDark = defaultThemeId != null && defaultThemeId.contains("dark"); //$NON-NLS-1$
```
iloveeclipse, line 889: `defaultThemeId.contains("dark")` seem to match only default theme id.
If product defines theme "com.my.product.themeid", there is not necessarily "dark" in the id but the theme can be "dark".

Ideally this could be extracted to a `protected boolean isDarkTheme(String themeId)`

iloveeclipse:
> Test in a local Eclipse SDK build and I do not see any issues with it. Planning to merge today

Given the fact that that "theming" is different on different platforms, it would be good if at least Windows & Mac behavior could be verified for this change upfront.

I've checked Linux with no themes and it seem to not break this use case. For the rest I don't have time.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
vogella: That is currently the way the theme engine detects a dark theme. We can make this more flexible here but the rest of theming engine will just look for the "dark" string so that would make this not aligned with the rest of the theming.

iloveeclipse: > That is currently the way the theme engine detects a dark theme

OMG

vogella: The theming code is a pleasure to look at. Tons of unused abstraction in lots of places and lots of hard coded values in others.

vogella:
Looks also good on Windows:

image

image

Unfortunately no access to a Mac.

PR #3860, opened 2026-04-03
Skip dot-folders (e.g. .git) when scanning for projects to import

### LINKED ISSUE #3858: Skip git folder when scanning for projects to import

## Suggestion

Import existing project into workspace, select the path with many plugin projects and a `.git` folder
Now, also the `.git` folder is scanned and for me this takes a while.
I find that always annoying. Could a `.git` folder be detects as to be skipped to find projects? 

## Community

vogella:
Adds a **Skip folders starting with '.'** checkbox (default: true) to both import wizards:

- **Smart Import** (SmartImportRootWizardPage / SmartImportJob)
- **Import Existing Projects** (WizardProjectsImportPage)

When enabled, directories like `.git`, `.svn`, `.hg` etc. are skipped during recursive project scanning, significantly improving import performance for repositories with large `.git` folders.

The setting is persisted in dialog settings and defaults to `true` for new installations.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3858


merks:
FYI I expect generally all builds will fail due to 502 errors from download.eclipse.org.

vogella:
> FYI I expect generally all builds will fail due to 502 errors from download.eclipse.org.

Thanks for the info.

vogella:
Sorry @merks missed the comment in the issue: There should always be a comma after ‘e.g.’

Follow-up PR coming

PR #3862, opened 2026-04-03
Fix vanishing markers when no resource is selected in Problems view

vogella:
## Problem

When the filter scope in the Problems view is set to a selection-dependent value (e.g., 'Selected resource', 'Same project', 'Same container') and no resource is currently selected in the Project Explorer, `selectByScope()` returns `false` for every marker. This makes the entire Problems view appear empty — users must re-select something in the explorer to see markers again.

## Fix

In `MarkerFieldFilterGroup.selectByScope()`, when the `resources` array is empty and the scope depends on selection (`ON_SELECTED_ONLY`, `ON_SELECTED_AND_CHILDREN`, `ON_ANY_IN_SAME_CONTAINER`), fall back to `ON_ANY` behavior (show all) instead of returning `false` (show nothing).

## Files Changed
- `MarkerFieldFilterGroup.java` — 4-line guard clause in `selectByScope()`
- `MarkerFieldFilterGroupTest.java` — New test class with 8 tests covering all scope + empty-resources combinations
- `InternalTestSuite.java` — Register new test class

## Testing
- New `MarkerFieldFilterGroupTest` verifies all scope values with both empty and non-empty resource arrays
- Verified compilation with `mvn clean compile -pl :org.eclipse.ui.ide -Pbuild-individual-bundles`

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerFieldFilterGroup.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/views/markers/MarkerFieldFilterGroup.java
@@ -649,6 +649,11 @@ public class MarkerFieldFilterGroup {
 
 	public boolean selectByScope(MarkerEntry entry, IResource[] resources) {
 		int scopeVal = getScope();
+		// When no resources are selected and scope depends on selection,
+		// fall back to showing all markers to avoid confusing empty results
+		if (resources.length == 0 && scopeVal != ON_ANY && scopeVal != ON_WORKING_SET) {
+			return true;
+		}
 		switch (scopeVal) {
 		case MarkerFieldFilterGroup.ON_ANY: {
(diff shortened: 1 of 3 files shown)


vogella:
Old:

image

No indicator that the detailed selection does not count:

image

New:

Show all indicates that the other UI elements are disabled:

image

Scope is drop-down box to save space

image


Types can be extended as drop-down.

image

vogella:
Latest UI

image

image

iloveeclipse:
> and no resource is currently selected in the Project Explorer, `selectByScope()` returns `false` for every marker. This makes the entire Problems view appear empty — users must re-select something in the explorer to see markers again.

This is correct behavior, where is the problem?
Now assume the workspace has 1000 projects, each 100 markers, according to this proposal, if nothing is selected, one would see 100.000 problems in the Problems view? Why should it be better, for which use case should it be helpful?

PR #3870, opened 2026-04-07
Fix disabled wizard buttons appearing enabled in dark theme

vogella:
In the dark theme, the CSS sets a foreground color on all Button elements, which overrides the native OS disabled button rendering. This causes disabled buttons (like 'Finish') to look identical to enabled ones.

## Approach

Sets a `disabled` CSS class on wizard buttons when they are disabled via `setData("org.eclipse.e4.ui.css.CssClassName", "disabled")`. The dark theme CSS includes a new `Button.disabled` rule with a dimmed foreground color. A `reskin(SWT.NONE)` call triggers the CSS engine to re-apply styles.

This avoids enabling global dynamic CSS pseudo-class support (`org.eclipse.e4.ui.css.dynamic`), which is disabled for performance reasons.

## Changes

- **`WizardDialog.java`**: Added `setButtonEnabled()` helper used in all button state change paths (`updateButtons`, `saveEnableStateAndSet`, `restoreEnableState`)
- **`e4-dark_globalstyle.css`**: Added `Button.disabled { color: #777777; }` rule

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java
@@ -973,7 +973,7 @@ public class WizardDialog extends TitleAreaDialog implements IWizardContainer2, 
 		if (control != null) {
 			Boolean b = (Boolean) saveState.get(key);
 			if (b != null) {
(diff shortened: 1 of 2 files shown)


vogella:
New

image

Old

image

PR #3871, opened 2026-04-07
Defer expensive project reference computation during selection changes

### LINKED ISSUE #2636: [performance] CloseUnrelatedProjectsAction.buildConnectedComponents(IProject[]) blocks UI startup

vogella:
CloseUnrelatedProjectsAction.getSelectedResources() calls computeRelated() which triggers buildConnectedComponents() and getReferencedProjects() for every project in the workspace. This is expensive because it can trigger classpath container resolution.

Previously, the inherited updateSelection() called getSelectedResources() on every selection change, causing this expensive computation to run on the UI thread during startup and navigation. This blocked the IDE.

Override updateSelection() to perform only a cheap enablement check: verify the selection contains at least one open IProject. The expensive computeRelated() call is deferred to getSelectedResources(), which is only invoked when the action is actually executed via run(). This is correct because getSelectedResources() already handles the selectionDirty flag and recomputes projectsToClose lazily when needed.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2636

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java
@@ -133,7 +133,7 @@ public class CloseUnrelatedProjectsAction extends CloseResourceAction {
 
 	/**
 	 * Overrides to avoid calling the expensive
-	 * {@link #computeRelated(List)} during selection changes. Uses only
+	 * {@code computeRelated(List)} during selection changes. Uses only
 	 * the raw selection to determine enablement.
 	 */
 	@Override
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/actions/CloseUnrelatedProjectsAction.java
@@ -30,6 +30,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.dialogs.MessageDialogWithToggle;
 import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.jface.viewers.IStructuredSelection;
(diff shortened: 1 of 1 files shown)


vogella:
I also looked for other actions but I only found BuildAction which does work only on selection projects and this is empty during startup.  The CloseUnrelatedProjectsAction was different because getSelectedResources() was overridden to compute the full workspace project graph regardless of what was selected

basilevs:
@vogella this causes the menu action to stay enabled after the first run.
Test steps:
- create two linked projects A, B
- create an unrelated project C
- close unrelated for A
- close unrelated for B

Old behavior:
menu item is disabled for the last step

New behavior:
menu item is always enabled.

Is this intentional? What does the phrase

> This is correct because getSelectedResources() already handles the selectionDirty flag and recomputes projectsToClose lazily when needed.

mean?

Why menu item stays in an obsolete state?

First RCPTT [test](https://github.com/eclipse-rcptt/org.eclipse.rcptt/blob/master/rcpttTests/rcptt_ide/platform_tests/q7Explorer/CloseUnrelatedProjectsInQ7Explorer.test) failure is [observed on platform staging on April 24th](https://ci.eclipse.org/rcptt/view/active/job/master/2419/).

vogella:
I have a look....

vogella:
Thanks for reporting @basilevs 

I could reproduce with new unit tests. Please test https://github.com/eclipse-platform/eclipse.platform.ui/pull/3941 with your scenario and let me know if the issue is fixed for you with it.

The new tests are failing before the change, and run fine with the change in the PR.

PR #3872, opened 2026-04-07
Add 'Manage default...' button to theme preference page

vogella:
Adds a **Manage default...** button next to the theme selection combo on the Appearance preference page.

Clicking it opens a dialog that:
- Explains that the default theme is used for new workspaces and workspaces without an explicit theme
- Shows the currently configured default theme (if any)
- Allows setting the selected theme as the default
- Allows removing the current default (button only shown when a default is configured)

This builds on the previous commit that added the ability to set a default theme via configuration scope during theme switch.


vogella:
image

image

image

PR #3874, opened 2026-04-08
Remove redundant platform-specific dark theme processors

### LINKED ISSUE #3857: [Proposal] Use new SWT API for native dark mode preference to simplify startup styling
### Description
Currently, the Eclipse IDE and RCP applications handle native dark mode (e.g., dark title bars, scrollbars, and native dialogs) through a fragmented set of platform-specific processors. These processors often rely on reflection, internal access, or fragile hooks into the OS.

Examples of current platform-specific implementations:
- `org.eclipse.e4.ui.swt.internal.cocoa.CocoaDarkThemeProcessor`
- `org.eclipse.e4.ui.swt.internal.win32.DarkThemeProcessor`
- `org.eclipse.e4.ui.swt.internal.gtk.DarkThemeProcessor`

These are often invoked late in the application lifecycle or require complex setup (like the CSS engine) to be fully effective.

### The Problem
One specific area of difficulty is the **early startup phase**. Elements like the Splash screen and the Workspace Selection Dialog (`ChooseWorkspaceDialog`) are created before the full E4 workbench and CSS engine are initialized. To style these with the user's "default theme" preference (e.g., a dark theme), we currently have to manually apply styles and listen for shell creation in `IDEApplication`.

Even then, native components like the `DirectoryDialog` (Browse button in the workspace launcher) do not respect these styles because the OS is not aware that the application "prefers" a dark look.

### Proposed Improvement
We propose the introduction of a cross-platform API in SWT (e.g., `Display.setDarkThemePreferred(boolean)`) that signals the dark mode preference directly to the OS.

**Resulting Simplifications:**
1. **Startup UI:** In `IDEApplication`, we could simply call `display.setDarkThemePreferred(true)` based on the saved default theme preference. This would immediately make the Splash screen and the Workspace Launcher (including its native DirectoryDialog) look correct without manual CSS application or complex listeners.
2. **Native Consistency:** Title bars on Windows, macOS, and Linux would automatically switch to their immersive dark variants.
3. **Code Cleanup:** We can eventually phase out the fragile, platform-specific processors in `org.eclipse.e4.ui.workbench.swt` and related bundles, replacing them with a single, supported SWT call.

### Benefits
- **Better UX:** A consistent dark experience from the first frame.
- **Maintainability:** Reduced reliance on internal platform-specific hacks.
- **Standards-based:** Leverages native OS capabilities (DWM on Windows, NSAppearance on macOS, GTK settings on Linux).

vogella:
The cross-platform SWT API `Display.setDarkThemePreferred(boolean)` already drives native dark mode (title bars, scrollbars, native dialogs), and the `ThemeEngine` calls it whenever the CSS theme changes.
That makes the platform-specific dark theme processors for GTK, Win32 and Cocoa redundant, so they are removed along with the fragile internal/reflection-based hooks they relied on.
The GTK and Win32 fragments only existed to host their processor and are deleted together with their entries in the `org.eclipse.e4.rcp` feature; the Cocoa fragment is kept because it still provides other processors.
The early `Display.setDarkThemePreferred(true)` call and the manual styling of the workspace selection dialog in `IDEApplication` are intentionally left in place, because the CSS engine is not yet running at that point in startup.

Fixes #3857


vogella:
This will already apply the default theme to the workspace launcher. But it seems that the launcher has some hard coded colors.

image

merks:
It would seem better if you experimented locally in your workspace until it’s working well rather than in a public draft PR that generates yet more email about interim states of progress that are, in my opinion, best managed by you privately until it’s ready for prime time. I don’t think verification builds are providing value in this case, do they? Perhaps such builds only contribute to global warming and to email generation. 

I’m tempted to ask about all the things I see being deleted, but it’s a draft so maybe pointless it’s a pointless question.

vogella:
> I’m tempted to ask about all the things I see being deleted, but it’s a draft so maybe pointless it’s a pointless question.

By using the dark  API from SWT, we can delete the processors in platform.ui. The processors were doing it platform specific. Once this is in we can delete the fragment projects.

merks:
So much mail. Please can this be reduced? Are the repeated builds and cascading notifications necessary? It just seems to be getting worse by the day. So many activities in parallel in public, driving endless builds and around the clock notifications.  

Please, please reconsider your work patterns.  Complete the design and testing before feeling the urge to create yet another draft on yet another topic in parallel to several others.

vogella:
> So much mail. Please can this be reduced? Are the repeated builds and cascading notifications necessary? It just seems to be getting worse by the day. So many activities in parallel in public, driving endless builds and around the clock notifications.
> 
> Please, please reconsider your work patterns. Complete the design and testing before feeling the urge to create yet another draft on yet another topic in parallel to several others.

Sure, I think you said the same in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3881#issuecomment-4226285677.

vogella:
Testing this now in a custom aggregator build.

vogella:
Tested on Linux and Windows (I have no access to Mac) in an custom aggregator build and styling still works and OS components, like scrollbars in Windows are still styled.

@HannesWell any concerns from the releng side? AFAICS these fragments are not part of a pom.xml and I removed the reference to them in the feature.xml

I plan to merge early next week if I do not hear concerns.

HannesWell (commented):
It's nice to have the build and platform a bit simpler with this. Thanks for that.
And considering the theme earlier is also good. 👍🏽 

> any concerns from the releng side? AFAICS these fragments are not part of a pom.xml and I removed the reference to them in the feature.xml

From RelEng POV this change should be fine. I didn't find anymore relevant references.

I just see that in SWT there seem to be code, that could be obsolete now:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/cb31ff97da4f842ad7c65af7dff65460178880c6/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/gtk/org/eclipse/swt/internal/gtk/OS.java#L2030-L2041

vogella:
> It's nice to have the build and platform a bit simpler with this. Thanks for that. And considering the theme earlier is also good. 👍🏽
> 
> > any concerns from the releng side? AFAICS these fragments are not part of a pom.xml and I removed the reference to them in the feature.xml
> 
> From RelEng POV this change should be fine. I didn't find anymore relevant references.
> 
> I just see that in SWT there seem to be code, that could be obsolete now: https://github.com/eclipse-platform/eclipse.platform.swt/blob/cb31ff97da4f842ad7c65af7dff65460178880c6/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/gtk/org/eclipse/swt/internal/gtk/OS.java#L2030-L2041

CORRECTED: AFAICS the method is now called internally via OS.setTheme so it is not obsolete.

PR #3875, opened 2026-04-08
Use UserScope instead of ConfigurationScope for default theme

vogella:
Changes the default theme preference storage from ConfigurationScope to UserScope.

UserScope is more appropriate as it represents a per-user setting that roams with the user across installations, while ConfigurationScope is shared across all users of an installation.

### Changes
- **E4Application.java**: Read default theme from UserScope instead of ConfigurationScope
- **ViewsPreferencePage.java**: Write default theme to UserScope instead of ConfigurationScope

Based on commit a7e91a0fe0cca3b15ab347ee380656242f42e823 which introduced the default theme feature.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
+++ b/bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
@@ -38,8 +38,8 @@ import org.eclipse.core.runtime.IExtensionRegistry;
 import org.eclipse.core.runtime.IProduct;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.RegistryFactory;
-import org.eclipse.core.runtime.preferences.ConfigurationScope;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.UserScope;
 import org.eclipse.e4.core.contexts.ContextFunction;
 import org.eclipse.e4.core.contexts.ContextInjectionFactory;
 import org.eclipse.e4.core.contexts.EclipseContextFactory;
@@ -317,9 +317,9 @@ public class E4Application implements IApplication {
 				: getArgValue(E4Application.THEME_ID, applicationContext, false);
 
 		if (!themeId.isPresent() && !cssURI.isPresent()) {
-			IEclipsePreferences configurationScopeNode = ConfigurationScope.INSTANCE
+			IEclipsePreferences userScopeNode = UserScope.INSTANCE
 					.getNode("org.eclipse.e4.ui.css.swt.theme");
-			String defaultThemeId = configurationScopeNode.get("themeid", DEFAULT_THEME_ID);
+			String defaultThemeId = userScopeNode.get("themeid", DEFAULT_THEME_ID);
 			context.set(E4Application.THEME_ID, defaultThemeId);
 		} else {
 			context.set(E4Application.THEME_ID, themeId.orElseGet(() -> null));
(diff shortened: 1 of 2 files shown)


merks:
So all different rcp and ide applications and all installations of those can only have one theme and there is no possibility to specify otherwise? Is that really a good idea? It seems like a loss of flexibility and a source of conflicting needs doesn’t it?

vogella:
> So all different rcp and ide applications and all installations of those can only have one theme and there is no possibility to specify otherwise? Is that really a good idea? It seems like a loss of flexibility and a source of conflicting needs doesn’t it?

No, the override is possible per workspace, this has not changed.

iloveeclipse:
> No, the override is possible per workspace, this has not changed.

Could you elaborate how this override works? Changed code seem to use only user scope?

vogella:
The logic to use either a workspace specific theme or a default theme was added in  https://github.com/eclipse-platform/eclipse.platform.ui/pull/3854/changes/75599e07c5a4da532d4eeacfd4c62e26786b34db

During theme switch the user can now set the selected theme as default This setting is saved as configuration scope and loaded for new workspaces unless the user has a theme set for this specific workspace. Add 'Use as default theme' checkbox to theme restart dialog

And on commit https://github.com/eclipse-platform/eclipse.platform.ui/commit/a7e91a0fe0cca3b15ab347ee380656242f42e823 which introduced the default theme feature but I assume that was clear as it was in the PR description,

merks:
> but I assume that was clear as it was in the PR description,

Imagine being on vacation getting 100 emails about all kinds of changes. After a while I think things are only clear to the email generators, which seem to produce higher volume with ever increasing frequency on so many different themes it feels like being on a freeway with oncoming traffic. 

It still feels like having a theme preference shared across all different applications is not necessarily ideal, but I suppose users affected could choose not to use it, though it’s not as if one could unset the default once set. I just imagined that different products/applications could choose different defaults. I was not aware a default was not possible previously. I thought products could specify default. Hard to investigate from an iPhone.

vogella:
@merks to give the full picture in short words:

Current situation: Theme can be set by workspace and via product customization (second one I guess, never used it).

New situation: During theme switch user can decide that the selected theme becomes the default theme for workspaces without an explicit theme set. Preference has a way to set or remove the default via a button (will be in tomorrows I-build) beside the theme selection. 

image


image

Examples: 

10 workspace never set a theme, user selects dark theme as default -> all go dark
10 workspace with 5 of them set to rainbow theme, the other ones never touched, switch default to dark theme you end up with 5 rainbow themes and 5 dark themes)
New workspace would use default if set. Otherwise light theme which is selected by default if nothing is set as default (like always light theme is selected if nothing else is set, no change here)

merks:
In principle it has useful aspects, but I keep mentioning all kinds of diverse eclipse applications and your answers always sound like talking about **the** eclipse IDE.

vogella:
Copilot suggest to add migration code, as this feature was only implemented yesterday I merge it now to avoid this migration need.

vogella:
> In principle it has useful aspects, but I keep mentioning all kinds of diverse eclipse applications and your answers always sound like talking about **the** eclipse IDE.

Ah, I did not see that comment. Must be the 100 email issue you mentioned. I think that is a valid concern. While the user may want to use the same theme for his Java IDE he may use a different theme for the C++ IDE.

Shall this theme default set by product or application?

HannesWell:
> Shall this theme default set by product or application?

Per product is probably better since it reflects the 'branding' and users might not know that the product/app they are using is based on Eclipse.
And one application can be used by many products, e.g. the IDE-apllication.

vogella:
Change is now product (or if product is missing application scoped) with https://github.com/eclipse-platform/eclipse.platform.ui/pull/3878

PR #3895, opened 2026-04-16
Style workspace version dialog with dark theme

vogella:
Apply the same dark theme styling used for the workspace selection dialog to the "Older/Newer Workspace Version" dialog. This ensures the version warning dialog is rendered consistently when the user's preferred theme is dark.


vogella:
Dark default (Windows)

image

Light default (Windows)

image

Issue #3784, opened 2026-03-17
Empty area after closing Welcome Screen


trancexpress:
See:

<img width="1133" height="800" alt="Image" src="https://github.com/user-attachments/assets/df80ccba-5307-4627-9a38-91a29742c88b" />

I can reproduce this with:

```
Eclipse SDK
Version: 2026-06 (4.40)
Build id: I20260316-1800
```

I cannot reproduce with:

```
Eclipse SDK
Version: 2026-06 (4.40)
Build id: I20260315-1800
```

With everything except Copilot installed from: https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/releng/org.eclipse.ui.releng/platformUiTools.p2f

Starting in a new workspace and closing the welcome screen:

https://github.com/user-attachments/assets/8ad1b1b5-9cf2-4ac3-ad52-9679a3a4814e

trancexpress:
When running from Eclipse, I cannot reproduce the fails we see in the latest I-build:

https://download.eclipse.org/eclipse/downloads/drops4/I20260316-1800/testresults/html/org.eclipse.ui.tests_ep440I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

```
	expected null, but was:<org.eclipse.ui.internal.ViewIntroAdapterPart@78997308>

java.lang.AssertionError: expected null, but was:<org.eclipse.ui.internal.ViewIntroAdapterPart@78997308>
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.failNotNull(Assert.java:756)
at org.junit.Assert.assertNull(Assert.java:738)
at org.junit.Assert.assertNull(Assert.java:748)
at org.eclipse.ui.tests.propertysheet.NewPropertySheetHandlerTest.hideAndAssertNoParts(NewPropertySheetHandlerTest.java:173)
at org.eclipse.ui.tests.propertysheet.NewPropertySheetHandlerTest.hideAndAssertNoParts(NewPropertySheetHandlerTest.java:162)
at org.eclipse.ui.tests.propertysheet.NewPropertySheetHandlerTest.testGetShowInContextWithNoActivePart(NewPropertySheetHandlerTest.java:183)
at java.base/java.lang.reflect.Method.invoke(Method.java:580) 
```

[...]

PR #3896, opened 2026-04-16
Fix flaky ensureCleanUpAddonCleansUp by holding CleanupAddon strongly

### LINKED ISSUE #3581: PartRenderingEngineTests.ensureCleanUpAddonCleansUp still fails randomly
https://github.com/eclipse-platform/eclipse.platform.ui/runs/57884315250

```
CleanupAddon should ensure that partStack is not rendered anymore, as all childs have been removed
java.lang.AssertionError: CleanupAddon should ensure that partStack is not rendered anymore, as all childs have been removed
	at org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests.ensureCleanUpAddonCleansUp(PartRenderingEngineTests.java:2472)
	at org.eclipse.e4.ui.tests.rules.WorkbenchContextRule$1.evaluate(WorkbenchContextRule.java:62)
	Suppressed: java.lang.Throwable: Screenshot written to C:\Users\RUNNER~1\AppData\Local\Temp\org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests.ensureCleanUpAddonCleansUp.png
		at org.eclipse.test.Screenshots$ScreenshotOnFailure.failed(Screenshots.java:41)
AWT screenshot saved to: C:\Users\RUNNER~1\AppData\Local\Temp\org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests.ensureCleanUpAddonCleansUp.png
```

The test failure can be reproduced locally (on Linux at least) after executing the test ~30-70 times.

See:
- https://github.com/eclipse-platform/eclipse.platform.ui/commit/64500cdcaf3233b8f368674d01b39b010db7af67
- https://github.com/eclipse-platform/eclipse.platform.ui/commit/7a24db1a216ef1e9940840c7c1156e1f9a82fc93
- https://github.com/eclipse-platform/eclipse.platform.ui/commit/a3fa06583905e66550c0b4e8a3961320cfb4ca43
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/751

vogella:
## Summary

`PartRenderingEngineTests.ensureCleanUpAddonCleansUp` has been flaking for a long time (#3581). Prior fixes (synchronous cleanup; extra `spinEventLoop()` between hides) addressed wrong root causes — the synchronous fix had to be reverted because it broke the Welcome screen (#3784), and the event-loop fix did not stop the flake.

## Root cause

The test creates the addon via `ContextInjectionFactory.make(CleanupAddon.class, appContext)` and discards the returned reference. The DI framework holds injected objects only via `WeakReference` (see `InjectorImpl.injectedObjects`). When GC collects the addon between event registration and event firing, the `@UIEventTopic` dispatcher (`UIEventObjectSupplier.UIEventHandler.handleEvent`) observes `requestor.isValid() == false` and silently unsubscribes — the addon method is never invoked, the `asyncExec` that would have unrendered the part stack is never queued, and the assertion fails.

Reproduced locally on Linux with the test trace clearly showing the `TBR partStackBC -> false` event missing on every failure.

## Fix

Store the addon in the application context. The context holds it strongly for the test lifetime; this also matches how addons are referenced in real applications (via the application model). The same pattern is applied to `testBug332463` and `CleanupAddonTest`, which had the same latent issue.

## CI verification

A previous revision of this PR wrapped the test in `@RepeatedTest(500)` to run it 500 times across all CI platforms — all green. Locally (Linux) before the fix: ~1.2% failure rate over 500 iterations; with the fix: 500/500 passed. The `@RepeatedTest(500)` annotation has now been removed and the test is back to a single `@Test`, ready for merge.

Fixes #3581

### DIFF
--- a/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
+++ b/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/workbench/PartRenderingEngineTests.java
@@ -2435,7 +2435,11 @@ public class PartRenderingEngineTests {
 		partStackForEditor.getChildren().add(editor);
 		partStackForEditor.setSelectedElement(editor);
 
-		ContextInjectionFactory.make(CleanupAddon.class, appContext);
+		// Store the addon in the context so it isn't garbage collected before
+		// its event handlers fire. The DI framework holds injected objects
+		// only via WeakReference, so an addon discarded here can be collected
+		// before its asyncExec runs, silently invalidating the requestor.
+		appContext.set(CleanupAddon.class, ContextInjectionFactory.make(CleanupAddon.class, appContext));
 
 		contextRule.createAndRunWorkbench(window);
 
@@ -2445,8 +2449,6 @@ public class PartRenderingEngineTests {
 
 		assertTrue(partStackForPartBPartC.isToBeRendered(), " PartStack with children should be rendered");
 		partService.hidePart(partB);
-		// Drain pending events between hides so that the event queue is clean
-		// when the second hidePart queues CleanupAddon's asyncExec for visCount==0
 		contextRule.spinEventLoop();
 		partService.hidePart(partC);
 		contextRule.spinEventLoop();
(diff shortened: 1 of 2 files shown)


vogella:
@ptziegler looks like we finally nailed it. For tip that you could reproduce the test failure if you run the test 100 times was the golden nugget which helped to solve this.

PR #3901, opened 2026-04-18
Fix sorting when using a TreePathViewerSorter in AbstractTreeViewer

### LINKED ISSUE #3900: CommonViewerSorter broken since 2026-03
The CommonViewerSorter has been deprecated in favor of CommonViewerComparator in #3621.

The above PR added additional cases for TreePathViewerComparator in most locations, but here, the existing check for TreePathViewerSorter was replaced with TreePathViewerComparator.
https://github.com/eclipse-platform/eclipse.platform.ui/blob/9b991c8cc7c6f613558e294558315712de204ec9/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java#L665-L691

So when using CommonViewerSorter, the above code falls trough to the else case where some default sorting is applied. I think instead of replacing the case for TreePathViewerSorter, the case should have been duplicated for TreePathViewerComparator.

This surfaced as a test-failure in LSP4E (see https://github.com/eclipse-lsp4e/lsp4e/pull/1512), where it was solved by using the new CommonViewerComparator. However, other Plugins still using CommonViewerSorter will be broken on 2026-03+

FYI: @iloveeclipse @akurtakov 

Edit: Sorry, meant to post this with my work account (@FlorianKroiss)

FlorianKroiss:
Fixes #3900 

I tried to minimize code duplication by making the outer if clause include both cases and then checking for the actual type again

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -662,12 +662,13 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 		return comparator.compare(this, e1, e2);
 	}
 
+	@SuppressWarnings("removal")
 	@Override
 	protected Object[] getSortedChildren(Object parentElementOrTreePath) {
 		Object[] result = null;
 		ViewerComparator comparator = getComparator();
 		if (parentElementOrTreePath != null
-				&& comparator instanceof TreePathViewerComparator tpvs) {
+				&& (comparator instanceof TreePathViewerSorter || comparator instanceof TreePathViewerComparator)) {
 			result = getFilteredChildren(parentElementOrTreePath);
 			// be sure we're not modifying the original array from the model
 			result = result.clone();
@@ -682,7 +683,11 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 					path = internalGetSorterParentPath(w, comparator);
 				}
 			}
-			tpvs.sort(this, path, result);
+			if (comparator instanceof TreePathViewerSorter tpvs) {
+				tpvs.sort(this, path, result);
+			} else if (comparator instanceof TreePathViewerComparator tpvc) {
+				tpvc.sort(this, path, result);
+			}
 			result = applyItemsLimit(parentElementOrTreePath, result);
 		} else {
 			return super.getSortedChildren(parentElementOrTreePath);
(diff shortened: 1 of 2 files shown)


FlorianKroiss:
@akurtakov I tested this change locally with the affected LSP4E version and the test passed again when using CommonViewerSorter.

merks:
The baseline errors look kind of bogus and seem to be related to signed versus unsigned.

akurtakov:
Looks good to me but I'll not merge it now to give sometime for the verification issue to be investigated.

merks:
Is this PR based on master?  @laeubi suggested this signing problem looks like one caused by being out of sync with master.  Or maybe @HannesWell made build changes.  Something is obviously fishy here...

HannesWell:
> Or maybe @HannesWell made build changes. Something is obviously fishy here...

Not explicitly, but that it only affects tests is noticeable.
```
Error:  Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.3-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.e4.core.commands.tests: Baseline and reactor have the same fully qualified version, but different content
Error:  different
Error:     META-INF/ECLIPSE_.RSA: present in baseline only
Error:     META-INF/ECLIPSE_.SF: present in baseline only
```

Maybe it's because of the recent removal of tests from the p2 repository. But as far as I know artifacts missing in the baseline are ignored.
But because that removal was only very recently, it could be that there are still older builds in the I-build composite that still contain the tests. And thus the baseline comparison sees an older version of the tests?

HannesWell:
> made build changes. Something is obviously fishy here...

Since this happens in almost all repositories I've created a dedicated issue for that now:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3784

vogella:
Nit: the switch with an unreachable default -> throw new IllegalStateException() is awkward. The sibling method internalCompare at 
  AbstractTreeViewer.java:654-663 already handles the same dual-type case cleanly with two instanceof checks. Mirroring that pattern here   
  would be more consistent and drop the dead branch:

````                                                                                                                                            
  if (comparator instanceof TreePathViewerSorter sorter) {                                                                                  
      sorter.sort(this, path, result);                                                                                                      
  } else if (comparator instanceof TreePathViewerComparator tpvc) {                                                                         
      tpvc.sort(this, path, result);                                                                                                        
  }  

````

FlorianKroiss:
Thanks @vogella for taking a look. I adjusted the code accordingly.

I also added tests to make sure both cases work as expected.

iloveeclipse (approved):
Thanks!

PR #3903, opened 2026-04-18
Relabel Project Explorer filter dialog OK as Apply, add search icon

vogella:
## Summary
- Rename the OK button in the Project Explorer "Filters and Customization" dialog (`CommonFilterSelectionDialog`) to "Apply", which better reflects that the action applies filter and content-extension changes to the view.
- Add `SWT.ICON_SEARCH` to the filter pattern `Text` in `CommonFiltersTab` so the search icon is shown next to the existing cancel icon.

## Test plan
- [ ] Open the Project Explorer, run "Filters and Customization..." and verify the primary button is labeled "Apply"; confirm it still commits filter and content-extension selections and closes the dialog.
- [ ] Verify the filter text field in the "Pre-set filters" tab shows both a search icon (leading) and the cancel icon (trailing), and that typing still filters the list.


vogella:
image

PR #3916, opened 2026-04-21
QuickAccessDialog: render filter as native search field

vogella:
Use SWT.SEARCH | SWT.ICON_SEARCH | SWT.ICON_CANCEL for the filter text so the dialog shows the platform magnifier icon and a cancel (X) button to clear the query, matching user expectations for a command palette.


vogella:
New 

image


Old

image

Issue #3905, opened 2026-04-20
Dark-Buttons on Workspace Selection Dialog Look Different Then on Other Dialogs


BeckerWdf:
In the dark theme the button on the workspace selection dialog (at start up of the IDE) look differently then on other dialogs of the IDE or on the same workspace selection dialog started from "File > Switch Workspace".

For screenshots and more details see https://github.com/eclipse-platform/eclipse.platform.swt/issues/3186#issuecomment-4278620942.

BeckerWdf:
@vogella can you have a look?

vogella:
> [@vogella](https://github.com/vogella) can you have a look?

Yes

vogella:
@BeckerWdf please test the test commit on Mac. If that works I would have to test it on Linux and Windows and might have to guard it for the Mac OS.

BeckerWdf:
How can I test this? As the issue only happens when starting the IDE. Does is also happen when I start a new IDE from a Development-IDE?

vogella:
If you start a runtime IDE without location you should be able to check

<img width="1239" height="978" alt="Image" src="https://github.com/user-attachments/assets/a436b7be-8f76-4d2f-9657-c5b1c3a0d5e5" />

BeckerWdf:
Without your changes:
<img width="705" height="370" alt="Image" src="https://github.com/user-attachments/assets/b3f8c348-b46e-411e-b72f-48e505dd577f" />

With your changes:
<img width="716" height="330" alt="Image" src="https://github.com/user-attachments/assets/404ed513-332b-4f35-a60e-70323d30d529" />

But from the running IDE it looks like this:
<img width="617" height="292" alt="Image" src="https://github.com/user-attachments/assets/2363e2d2-d1fe-4e60-a8aa-0943310dc5ab" />
See that the "Launch" button is highlighted

PR #3918, opened 2026-04-21
Better button styling in workspace selection dialog

vogella:
The workspace selection dialog at IDE startup mimics dark-theme styling because the ThemeEngine is not yet available. On macOS this produces buttons that look different from the buttons on other dialogs (e.g. `File > Switch Workspace`).

This change skips the hard-coded dark styling for `Button` controls only, so they get the native macOS dark appearance. All other widgets (Shell, Composite, Label, Link) continue to be styled as before.

On Windows the check button needs styling otherwise it appears white.

### DIFF
--- a/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
+++ b/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEApplication.java
@@ -58,6 +58,7 @@ import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.layout.FillLayout;
+import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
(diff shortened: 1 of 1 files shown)


vogella:
@BeckerWdf can you test this?

BeckerWdf:
> @BeckerWdf can you test this?

Didn't I already last week?

vogella:
> > @BeckerWdf can you test this?
> 
> Didn't I already last week?

This one is for your remark here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3905#issuecomment-4296936265 and AFAICS you did not give me feedback on this PR (which might solve your button styling).

BeckerWdf:
Ah sorry. I did test it. This PR solves it. Somehow I did not write that last week.

vogella:
> Ah sorry. I did test it. This PR solves it. Somehow I did not write that last week.

Can you post a screenshot, please?

BeckerWdf:
With your change:
image

Without your change:
image

vogella:
> With your change: image
> 
> Without your change: image

That looks like the same to me? The new one is better? Can you explain why?

BeckerWdf:
> > With your change: image
> > Without your change: image
> 
> That looks like the same to me? The new one is better? Can you explain why?

Look at the buttons. They are very different. The first one is correct (as all buttons in the IDE look this way). The second one is wrong. I already highlighted the difference in https://github.com/eclipse-platform/eclipse.platform.ui/issues/3905#issuecomment-4296936265.

vogella:
I assume you expected a blue button like in the original example you posted. If the above is sufficient I start testing it on Linux and Windows. Windows testing usually takes a bit longer for me.

BeckerWdf:
> I assume you expected a blue button like in the original example you posted. If the above is sufficient I start testing it on Linux and Windows. Windows testing usually takes a bit longer for me.

Sure. Having the button also in blue would be even better. But already this one is fetter then this strange "gradient" styled buttons.

vogella:
Linux Dark looks also really good:

image


Linux Light (unchanged to last years release):

image

vogella:
Windows looks bad:

image

vogella:
With latest patch set (special fix only for Windows):

image

vogella:
Also with "recent workspaces" fine on windows with the latest patch

image

vogella:
Tested on all platform, planning to merge this tomorrow.

PR #3921, opened 2026-04-22
Improve import dialog with column sorting and better visibility

SougandhS:
Improves the import dialog usability by adding column-based sorting for the “Folder” and “Import As” columns (with ascending/descending toggle on header click) and increasing the proposals list height to display more entries by default. This aligns sorting with the displayed column values and makes it easier to scan and organize import proposals, especially when dealing with many entries, reducing the need for manual resizing.

https://github.com/user-attachments/assets/78cd542f-92db-4c43-8b63-828855cfe474

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/SmartImportRootWizardPage.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/SmartImportRootWizardPage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2020 Red Hat Inc., and others
+ * Copyright (c) 2014, 2026 Red Hat Inc., and others
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


vogella:
Nice improvement, thanks for working on this. Two small things worth addressing before merge:                                                                                                                                                 
                                                                                                                                                                                                                                                
 1. Clicking the already sorted "Folder" column flips the direction on the very first click, because the comparator is initialized with `column = 0` and then `setColumn(0)` treats it as a repeat click. Users usually expect the first click 
  to sort ascending. Consider starting the comparator in an "unset" state (for example `column = -1`) so the first click lands on ascending.                                                                                                    
                                                                                                                                                                                                                                                
2. `ImportsComparator` can be `private static`. It does not capture any state from `SmartImportRootWizardPage` (the label provider is passed in via the constructor), so there is no need for the implicit outer reference.

SougandhS:
> Nice improvement, thanks for working on this. 

Thanks !

>Two small things worth addressing before merge:

Done 👍

vogella:
New observations:

1. Default-sort regression — column = -1 means the comparator returns 0 on open, so the tree is unsorted while the Folder column still shows a sort indicator (because setSortColumn(folderCol) runs unconditionally during setup).
2. heightHint = 190 — more than 2× the previous value; worth verifying on minimum screen size.
3. NPE risk — projectLP.getText(f1) in case 1 isn't null-guarded.

SougandhS:
> 1. Default-sort regression — column = -1 means the comparator returns 0 on open, so the tree is unsorted while the Folder column still shows a sort indicator (because setSortColumn(folderCol) runs unconditionally during setup).

Made few tweaks - Fixed 👍 

> 2. heightHint = 190 — more than 2× the previous value; worth verifying on minimum screen size.

Adjusted it to 150 - I hope its good. with 90 I can see only 4 or 5 items by default.

> 3. NPE risk — projectLP.getText(f1) in case 1 isn't null-guarded.

 Added a guard

SougandhS:
Thank you @vogella for the review and merging 😊

PR #3925, opened 2026-04-22
PreferenceDialog: render filter as native search field

vogella:
## Summary
- Use `SWT.SEARCH | SWT.ICON_SEARCH | SWT.ICON_CANCEL` for the filter field in `FilteredPreferenceDialog`'s `PreferenceFilteredTree`, so it is rendered as a native search field with magnifier and cancel icons.
- Override `setInitialText` to expose the filter hint via `Text#setMessage` only, rather than inserting it as real field content. The dialog auto-focuses the filter field on open, which made the old initial-text fallback behave poorly: users saw the hint as actual text and had to delete it before typing.

## Test plan
- [x] Open Window > Preferences and confirm the filter field renders with a magnifier icon and a cancel (x) icon once text is typed.
- [x] Verify the "type filter text" hint is shown as a placeholder and disappears when focus/typing begins, without needing to be deleted.
- [x] Filter by typing; verify the cancel icon clears the filter and restores the full tree.


vogella:
Old in Light theme

image


New in Dark theme

image

PR #3929, opened 2026-04-23
Avoid flashing shutdown dialog when save is fast

### LINKED ISSUE #1269: Closing Eclipse IDE shows flashing popup if closing is really fast

vogella:
Fixes #1269. The IDE's shutdown save dialog currently pops up and vanishes in a blink on fast machines. This change replaces the eager `CancelableProgressMonitorJobsDialog` with a plain `ProgressMonitorJobsDialog` that opens only after `IProgressService#getLongOperationTime` has elapsed; fast shutdowns now just show a busy cursor, slow ones still show a real progress dialog.

The cancel path that went away was effectively broken anyway. Its button only cleared a sub-task label, and the `CancelableProgressMonitorWrapper` used hard-coded work-amount math (`Math.abs(total - 4.0)`) to toggle cancellation around history pruning, an assumption that had drifted out of sync with `SaveManager`. Removing it also drops three no-longer-referenced message keys.

A new `IDEWorkbenchAdvisorTest#testShutdownWithSlowSave` exercises the delayed-open path via a save participant that sleeps past the long-operation threshold, asserting that the save still completes and hooks fire.

### DIFF
--- a/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
+++ b/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
@@ -61,6 +61,7 @@ import org.eclipse.swt.custom.BusyIndicator;
 import org.eclipse.swt.graphics.Resource;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Listener;
+import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.application.IWorkbenchConfigurer;
 import org.eclipse.ui.application.IWorkbenchWindowConfigurer;
@@ -79,7 +80,6 @@ import org.eclipse.ui.internal.ide.IDEWorkbenchErrorHandler;
 import org.eclipse.ui.internal.ide.IDEWorkbenchMessages;
 import org.eclipse.ui.internal.ide.IDEWorkbenchPlugin;
 import org.eclipse.ui.internal.ide.undo.WorkspaceUndoMonitor;
-import org.eclipse.ui.internal.progress.ProgressManagerUtil;
 import org.eclipse.ui.internal.progress.ProgressMonitorJobsDialog;
 import org.eclipse.ui.progress.IProgressService;
 import org.eclipse.ui.statushandlers.AbstractStatusHandler;
@@ -498,12 +498,12 @@ public class IDEWorkbenchAdvisor extends WorkbenchAdvisor {
 
 		final Display display = Display.getCurrent();
 		final int longOperationTime = savedLongOperationTime;
-		final Runnable openDialog = () -> {
-			if (ProgressManagerUtil.safeToOpen(dialog, null)) {
+		final Runnable openDialogLater = () -> {
+			if (!display.isDisposed() && !hasModalShell(display)) {
 				dialog.open();
 			}
 		};
-		display.timerExec(longOperationTime, openDialog);
+		display.timerExec(longOperationTime, openDialogLater);
 
 		try {
 			BusyIndicator.showWhile(display, () -> {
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
```diff
@@ -464,119 +461,61 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
+				status.merge(e.getStatus());
+			}
+		};
 
-					status.merge(((Workspace) ResourcesPlugin.getWorkspace()).save(true, true, monitor));
-				} catch (CoreException e) {
-					status.merge(e.getStatus());
-				}
-			};
+		final Display display = PlatformUI.getWorkbench().getDisplay();
```
iloveeclipse: Are you sure workbench is still there at this time?

vogella: ack

vogella:
Using this now in a custom SDK builds, it is please to close the IDE with this "non-flashing" patch. Thanks @HeikoKlare for the implementation suggestions.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
```diff
@@ -464,119 +470,61 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
+				} catch (InvocationTargetException e) {
+					status.merge(new Status(IStatus.ERROR, IDEWorkbenchPlugin.IDE_WORKBENCH, 1,
+							IDEWorkbenchMessages.InternalError, e.getTargetException()));
+				} catch (InterruptedException e) {
+					status.merge(new Status(IStatus.ERROR, IDEWorkbenchPlugin.IDE_WORKBENCH, 1,
+							IDEWorkbenchMessages.InternalError, e));
+				}
+			});
+		} finally {
+			display.timerExec(-1, openDialog);
```
iloveeclipse: `display.timerExec(-1, openDialog);`

Why `timerExec` and why "-1"? The code will be not executed on Linux!

iloveeclipse (dismissed):

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
vogella: `timerExec(-1, openDialogLater)` is the documented SWT idiom for cancelling a previously scheduled runnable, and it behaves the same on all three platforms. The javadoc states "If milliseconds is less than zero, the runnable is not executed," and each `Display` implementation matches the `Runnable` by reference in `timerList` and cancels the native timer before returning: GTK calls `g_source_remove`, win32 calls `OS.KillTimer`, and cocoa calls `timer.invalidate()` / `release()`. That is exactly the intent here: schedule `openDialogLater` to open the dialog after the long-operation threshold, and cancel it in `finally` if the save already finished, so no dialog appears on fast shutdowns. I pushed a small rename (`openDialog` -> `openDialogLater`) and a one-line comment on the cancel to make the schedule/cancel pairing explicit.

bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java
```diff
@@ -464,119 +470,61 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
-					status.merge(e.getStatus());
-				}
-			};
+		final Display display = Display.getCurrent();
+		final int longOperationTime = savedLongOperationTime;
+		final Runnable openDialog = () -> {
+			if (ProgressManagerUtil.safeToOpen(dialog, null)) {
+				dialog.open();
+			}
+		};
```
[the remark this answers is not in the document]
vogella: done

PR #3935, opened 2026-04-24
Drop dead log(Exception) helper in WorkbenchActivitySupport

### LINKED ISSUE #3915: WorkbenchActivitySupport.log(Exception) has never actually logged — 22-year-old bug
## Bug

`bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/activities/ws/WorkbenchActivitySupport.java:245`

```java
private void log(Exception e) {
    StatusUtil.newStatus(IStatus.ERROR, "Could not update contribution managers", e);
}
```

The method is named `log` and its Javadoc says *"Logs an error message to the workbench log"*, but it only constructs an `IStatus` and throws it away. No actual logging call.

Caller at line 232 swallows `InvocationTargetException` / `InterruptedException` from `ProgressMonitorDialog.run()` via this broken helper, so failures updating contribution managers after activity-state changes leave no trace.

## History

Introduced 2004-06-03 by Kim Horne (`4dd90155ddf`, "Fix for 65120"). Broken from day one; later touches were whitespace-only reformats. ~22 years, no reports: the catch block effectively never fires.

## Options

**1. Actually log it** via `StatusManager` (not `ILog` directly, so products with custom status handlers stay in control).

**2. Honest deletion.** Drop `log(Exception)` and let the caller swallow explicitly:
```java
} catch (InvocationTargetException | InterruptedException e2) {
    // ignored: ModalContext.run throws InterruptedException on cancel,
    // not real thread interruption; no caller has depended on this
    // being logged in 22 years
}
```
Note: do **not** call `Thread.currentThread().interrupt()` here. Per @tomaswolf, `ModalContext.run()` uses `InterruptedException` to signal cancellation rather than actual thread interruption, so re-setting the flag would propagate a false signal.

**3. Leave as-is.**

Preference: **option 2**. 22 years of silence is evidence nobody depends on the log output; cleaning up the dead helper is the honest fix.

vogella:
The private `log(Exception)` helper built an `IStatus` and immediately discarded it, never actually logging anything. It has been broken since it was introduced in 2004. Inline the exception swallow at the single call site with a comment explaining why the interrupt flag is not re-set (per @tomaswolf, `ModalContext.run` uses `InterruptedException` to signal cancellation, not actual thread interruption).

Fixes #3915

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/activities/ws/WorkbenchActivitySupport.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/activities/ws/WorkbenchActivitySupport.java
@@ -21,7 +21,6 @@ import org.eclipse.core.runtime.IExtension;
 import org.eclipse.core.runtime.IExtensionPoint;
 import org.eclipse.core.runtime.IProduct;
 import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.dynamichelpers.ExtensionTracker;
 import org.eclipse.core.runtime.dynamichelpers.IExtensionChangeHandler;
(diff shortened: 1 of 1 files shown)


tomaswolf (approved):

vogella:
Thanks @tomaswolf

PR #3938, opened 2026-04-27
Highlight matched text in editor drop-down list

### LINKED ISSUE #194: Support pattern matching in the editor drop-down list

vogella:
The chevron drop-down on editor stacks supports filtering but rendered matched characters in plain weight, so it was hard to see why a row qualified. This change switches `BasicPartList` to a `StyledCellLabelProvider` that bolds matched substrings using the same `StyledStringHighlighter` the Quick Switch Editor (Ctrl+E) dialog already uses, giving both lists a consistent look. The dirty-prefix asterisk is excluded from highlighting so only the filename portion is bolded.

Closes https://github.com/eclipse-platform/eclipse.platform.ui/issues/194


iloveeclipse:
Screenshots?

PR #3940, opened 2026-04-28
Use dialog font in editor drop-down

vogella:
The chevron drop-down on editor stacks inherited the SWT system font, which on many platforms looks out of place compared to the rest of the IDE chrome. This change switches both the filter text and the result table to `JFaceResources.getDialogFont()` so the list matches the Quick Switch Editor (Ctrl+E) dialog and other workbench dialogs.

The bold-when-hidden marker that this PR previously removed has been kept, following review feedback that it is useful for spotting which tabs are currently visible.

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/AbstractTableInformationControl.java
@@ -134,6 +134,7 @@ public abstract class AbstractTableInformationControl {
 		fTableViewer = createTableViewer(fComposite, controlStyle);
 
 		final Table table = fTableViewer.getTable();
+		table.setFont(JFaceResources.getDialogFont());
 		table.addKeyListener(KeyListener.keyPressedAdapter(e -> {
 			switch (e.keyCode) {
 			case SWT.ESC:
@@ -314,10 +315,11 @@ public abstract class AbstractTableInformationControl {
 
 	protected Text createFilterText(Composite parent) {
 		fFilterText = new Text(parent, SWT.NONE);
+		fFilterText.setFont(JFaceResources.getDialogFont());
 
 		GridData data = new GridData();
 		GC gc = new GC(parent);
-		gc.setFont(parent.getFont());
+		gc.setFont(fFilterText.getFont());
 		FontMetrics fontMetrics = gc.getFontMetrics();
 		gc.dispose();
 


vogella:
Old:

image


New (light):

Screenshot From 2026-04-28 10-53-28

New (dark):

image

The existing CTRL+E style which is unchanged would also in line with the new font

image

BeckerWdf:
You are mixing two changes. The font size and the bold-stuff. I think the one thing (size) is clear. The other one can be discussed. Should this be split into two PRs?

vogella:
> You are mixing two changes. The font size and the bold-stuff. I think the one thing (size) is clear. The other one can be discussed. Should this be split into two PRs?

I do not change the font size in this PR, please check the code. It looks bigger because it is bold (and not the dialog font)

BeckerWdf:
But your PR text say:

> The chevron drop-down on editor stacks inherited the SWT system font, which on many platforms looks out of place compared to the rest of the IDE chrome. **This change switches both the filter text and the result table to JFaceResources.getDialogFont()** so the list matches the Quick Switch Editor (Ctrl+E) dialog.

vogella:
> But your PR text say:
> 
> > The chevron drop-down on editor stacks inherited the SWT system font, which on many platforms looks out of place compared to the rest of the IDE chrome. **This change switches both the filter text and the result table to JFaceResources.getDialogFont()** so the list matches the Quick Switch Editor (Ctrl+E) dialog.

Yes, that is also what I said here: It looks bigger because it is bold (and not the dialog font)
So you are suggesting to split the PR by the usage of the dialog font from the bold removal?

iloveeclipse:
I can't spot a difference because every old/new pair of screenshots shows different dialogs, and the light one shows even dark mode??? Sorry, it just wastes everyone time. Can you please provide **clean** side by side examples before / after?

tomaswolf:
Leave the bolding. It is important information, and I rely on it quite frequently. I find it _very_ useful to see immediately which of the items is in a visible tab.

iloveeclipse:
> Leave the bolding. It is important information, and I rely on it quite frequently.

Correct. If this is not in Ctrl+E it should be added there, not otherwise.

vogella:
> Leave the bolding. It is important information, and I rely on it quite frequently. I find it _very_ useful to see immediately which of the items is in a visible tab.

So you need a visual distinction between open and not open. I see if I can come up with something that serves this purpose and does not look misplaced in the current ui (or even ugly) to me. I will copy you in the PR if I create one and will wait for your feedback.

vogella:
This change now only contains the usage of the dialog font. The "bold display" logic is unchanged.

BeckerWdf:
Can you pls again highlight what changed? With this big screenshots when you need to scroll up and down it's hard to see a difrerence.

PR #3942, opened 2026-04-29
Revert "Defer expensive project reference computation during selection changes"

vogella:
Reverts PR #3871 (commit 3aaea37e05) and the follow-up Javadoc fix in 3b23ac5263.

After #3871 the cheap `updateSelection()` override stopped consulting the project graph, so `CloseUnrelatedProjectsAction` stayed enabled after the last unrelated project was closed (also visible as an RCPTT regression on platform staging). Reverting restores the previous correct enablement; the original UI-thread cost addressed in #3871 can be tackled separately without breaking the action.

Closes #3941.

Refs https://github.com/eclipse-platform/eclipse.platform.ui/issues/2636


vogella:
Revert restores the old (blocking / delaying startup) behavior, merging

basilevs:
Has this reached simrel staging? RCPTT still fails.

merks:
The staging repository won’t change until the platform contributes M3 in two weeks.  But, given M2 is complete, the platform could contribute something sooner as well.

PR #3946, opened 2026-04-30
Restore Section title bar background lost after #3808

### LINKED ISSUE #3945: Broken/missing Forms UI header colors
Forms UI, as used in the Manifest editor, used to look like this:

image

It looks as follows since
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3808

image

@vogella is that an intended effect of that PR? I have to admit that I did not expect such a change from the PR description and, personally, do not really like the missing highlight in the header. The tool icons for the ordering also look quite free floating now without the clear relation to the header that was previously obvious because of the background connecting them.

vogella:
The flat-look optimization in #3808 added an early return in `Section.updateHeaderImage` whenever the title-bar gradient color matched the title-bar background color. That fired in the common case where the title-bar background is set distinct from the section body (e.g. `#eaeaea` on a `#ffffff` body), and the image it skipped was what painted that contrast across the title-bar area, so the title bar visually collapsed into the body.

Tighten the short-circuit to also require `bg.equals(getBackground())`, so the image is only skipped when title-bar gradient, title-bar background and section body background are all the same color. The optimization is preserved for genuinely flat sections; non-flat title bars get their image back.

Fixes #3945


BeckerWdf:
This fixes the issue. Thanks for providing a fix so fast.

PR #3954, opened 2026-04-30
Tests for CloseUnrelatedProjectsAction #2636

basilevs:
@vogella and me have attempted and failed fixing  #2636. These are passing tests that failed after our optimizations.

They should pass on the current version, potentially preventing regressions from the next optimization attempt.


vogella:
Good idea, thanks @basilevs

PR #3955, opened 2026-05-01
Fix dark styling for popup shells in workspace selection

vogella:
## Summary
- Use `display.addFilter` instead of `addListener` for `SWT.Show` so the early dark-theme listener actually fires for child popup shells (e.g. the content-proposal popup in the workspace selection combo). `Display.addListener` only catches events sent to the Display widget itself, not events on child shells — so the listener was previously a no-op for popups.
- Re-apply styles asynchronously to cover widgets that finish initializing after `SWT.Show` (e.g. the `SWT.VIRTUAL` proposal `Table` whose rows are populated lazily).

The two existing dark dialogs (workspace selection, version warning) appeared dark only because their `createContents()` overrides call `applyDarkStyles(getShell())` directly. The content-proposal popup is created by JFace's `ContentProposalAdapter` where we have no such hook, so it stayed light.

Reproducible on Windows; not reproducible on Linux.


vogella:
With this fix the completion window in windows is dark otherwise white:

image

Issue #3948, opened 2026-04-30
Upper version constraint for jakarta.annotation-api


ptziegler:
The Eclipse Orbit repository now contains the jakarta.annotation-api bundle in version 3.0.0. This means that products built against the simrel update site and which reference this plugin in version 0.0.0 now receive a validation error because 3.0.0 is pulled instead of 2.1.1.

<img width="1068" height="1170" alt="Image" src="https://github.com/user-attachments/assets/fa271121-81c7-44b5-9a05-70d62e5024c1" />

I don't think there is any breaking change between the 2.x and the 3.x (at least as far as the Eclipse IDE is concerned). Should we simply raise the upper bound so that consumers can use either one or the other?

akurtakov:
We should change these to [2.1.0,4.0.0) and even include 3.0.0 in the SDK. 
FWIW, this happens due to docker plugin -> docker-client ->jersey 4.x which mandates jakarta-annotation-api 3.0 (https://mvnrepository.com/artifact/org.glassfish.jersey.core/jersey-common/4.0.2/dependencies )

merks:
I'm not sure if it's a helpful/related suggestion, but a target platform can specify version ranges such that both version are included without specifying an exact version to include:

```
      <unit id="jakarta.annotation-api" version="[2.0.0,3.0.0)"/>
      <unit id="jakarta.annotation-api" version="[3.0.0,4.0.0)"/>
```

Tycho supports this as well...

merks:
Note I'm not suggesting we shouldn't relax the upper bound, just suggesting how one might deal with the current situation downstream.

merks:
FYI, this affects 3 different git repositories, eclipse.platform, eclipse.platform.ui, and eclipse.pde:

<img width="556" height="863" alt="Image" src="https://github.com/user-attachments/assets/c4c87bea-325d-4511-9e32-3f1fa15b170b" />

Shall I change them all or is someone else planning or already doing that?

ptziegler:
I wanted to start with this once I'm back home. But that shouldn't stop anyone if they happen to have some free time. :)

merks:
Yes, I have a long all day meeting.    

I think the only change is the removal of a class we aren't using:

<img width="1711" height="657" alt="Image" src="https://github.com/user-attachments/assets/5d430afc-1fc7-4d54-8c2d-b97f825d20fa" />

HannesWell:
[...]

PR #3956, opened 2026-05-01
Expand version range of jakarta.annotation to include 3.x.

merks:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/3948


merks:
Ignoring infrastructure problems with ci build.

PR #3959, opened 2026-05-03
[#3958] fix key command action handler gets deleted by content assist

### LINKED ISSUE #3958: Key command action handler gets deleted by content assist
## Summary
This CDT-LSP editor error has been reported: [CDT-LSP Editor smart caret issue at line start](https://github.com/eclipse-cdt/cdt-lsp/issues/611). The user reported that some editor actions like goto-line-start does not work anymore when the file gets edited. 
I found out that the text editor action handlers for goto-to-line-start, -end, etc. are set to `null` when a content assist has been triggered.

**Scroll down for TL;DR**

## Steps to reproduce
1. Open a file in an editor which is based on the extension based text editor and is backed via a language server (like cdt-lsp editor in CDT)
2. Write an indented line. Open CTRL-Space for content assist. Close the content assist pop-up window
3. Press HOME/pos1 to go to the start of the line. The caret position is now in column 0 and not right ahead of the first non-whitespace character.

## Detailed root cause
A content assist triggers a call of the `org.eclipse.jface.text.contentassist.ContentAssistant.fireSessionBeginEvent(boolean)` method.
 The `fireSessionBeginEvent(boolean)` method iterates over content assist processors and calls for each processor the listeners:
```java
if (fContentAssistSubjectControlAdapter != null && !isProposalPopupActive()) {
	Set<IContentAssistProcessor> processors= getProcessors(fContentAssistSubjectControlAdapter, fContentAssistSubjectControlAdapter.getSelectedRange().x);
	if (processors != null) {
		processors.forEach(processor -> {
			ContentAssistEvent event= new ContentAssistEvent(this, processor, isAutoActivated);
			for (ICompletionListener listener : fCompletionListeners) {
				listener.assistSessionStarted(event);
			}
		});
	}
}
```
One listener is the `org.eclipse.ui.texteditor.KeyBindingSupportForAssistant` class. The proplem is, that the callee is not aware to be called multiple times:
```java
@Override
public void assistSessionStarted(ContentAssistEvent event) {
	ICommandService commandService= PlatformUI.getWorkbench().getService(ICommandService.class);
	IHandler handler= getHandler(ContentAssistant.SELECT_NEXT_PROPOSAL_COMMAND_ID);
	fReplacedCommands= new ArrayList<>(10);
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.LINE_DOWN, handler, commandService));
	handler= getHandler(ContentAssistant.SELECT_PREVIOUS_PROPOSAL_COMMAND_ID);
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.LINE_UP, handler, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.LINE_START, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.LINE_END, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.PAGE_UP, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.PAGE_DOWN, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.TEXT_START, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.TEXT_END, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.SCROLL_LINE_UP, commandService));
	fReplacedCommands.add(new ReplacedCommand(ITextEditorActionDefinitionIds.SCROLL_LINE_DOWN, commandService));
}
```
The command handlers are replaced by content assist specific ones ore set to null in the `ReplacedCommand` helper class:
```
ReplacedCommand(String commandId, ICommandService commandService) {
	this.command= commandService.getCommand(commandId);
	replaceWith(null);
}
```
The `replaceWith` method stores the handler temporarily until the content assist has been finished:
```java
private void replaceWith(IHandler newHandler) {
	if (command.isHandled()) {
		handler= command.getHandler();
		command.setHandler(newHandler);
	}
}
```
**All `ReplacedCommand`objects gets deleted when the `assistSessionStarted(bool)` gets called a second time by `org.eclipse.jface.text.contentassist.ContentAssistant.fireSessionBeginEvent(boolean)`:**
`fReplacedCommands= new ArrayList<>(10);`
**Unfortunately now the call of `command.getHandler();` returns `null` in `replaceWith(Handler newHandler)`, as we have set it to null in the call before.
When the `org.eclipse.ui.texteditor.KeyBindingSupportForAssistant.assistSessionEnded(ContentAssistEvent)`method gets called when the content assist window has been closed, the old handler is lost, since its overriden in the second call of `assistSessionStarted(ContentAssistEvent event)`**

## TL;DR
A bug in the Eclipse text editor/content assist integration causes navigation keys like **HOME / go-to-line-start** to stop behaving correctly after content assist is opened in CDT-LSP or similar language-server-backed editors.

**What happens**

When content assist **(Ctrl+Space)** starts, Eclipse temporarily replaces several editor command handlers (line start, line end, page up/down, etc.) so the popup can use those keys.

**Root cause**

The content assist startup event is triggered **multiple times**, but the listener class `KeyBindingSupportForAssistant` assumes it runs only once.

On the second call:

- it recreates its internal list of saved handlers (`fReplacedCommands`)
- previous original handlers are discarded
- some commands were already set to `null`
- so the second save stores `null` instead of the real handler

When content assist closes, Eclipse restores the saved handlers — but now it restores `null`, meaning normal editor actions are lost.

**Visible result**

Keys like:

- **HOME** (smart line start)
- line end
- page navigation
- text start/end

may stop working correctly after using content assist.

**In short**

**Repeated content assist initialization overwrites the backup of command handlers, so original editor keybindings are never restored.**

## Solution 

I can provide a PR to fix that.

* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): Eclipse 2026-06

ghentschke:
fixes #3958 by handling multiple calls on assistSessionStarted(ContentAssistEvent)

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/KeyBindingSupportForAssistant.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/KeyBindingSupportForAssistant.java
@@ -136,6 +136,11 @@ public final class KeyBindingSupportForAssistant implements ICompletionListener 
 
 	@Override
 	public void assistSessionStarted(ContentAssistEvent event) {
+		// Do not replace commands twice if several processors are registered. This
+		// method gets called for each processor.
+		if (fReplacedCommands != null && !fReplacedCommands.isEmpty()) {
+			return;
+		}
 		ICommandService commandService= PlatformUI.getWorkbench().getService(ICommandService.class);
 		IHandler handler= getHandler(ContentAssistant.SELECT_NEXT_PROPOSAL_COMMAND_ID);
 		fReplacedCommands= new ArrayList<>(10);


ghentschke:
@mickaelistria can you please take a look at this?

mickaelistria:
I think this patch is good and is the simplest way to address the reported issue.
Sure, one could argue that this method should not be called multiple times, but anyway, there is no real way to prevent a method from being called multiple times, so making it multiple-calls-able seems more robust that working on the caller to call it only once.

merks (approved):
It looks fine to me as well.  Evidently `assistSessionEnded` is already guarded for being called multiple times or for being called without `assistSessionStarted` having been called, so guarding multiple calls to `assistSessionEnded` without a subsequent call to `assistSessionEnded` seems safe enough.

ghentschke:
Thank  for your fast replies @mickaelistria  and @merks 
IMO it would be great if we can merge this soon, as it the editor behavior is quite annoying for users. 
Can it be part of 2026-06 or is it too late?

merks:
I think others have had enough time to review if they so desired.

ghentschke:
Thank you

Issue #3961, opened 2026-05-04
We should not use the term "dirty indicator" in user facing tools documentaiton


BeckerWdf:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/3823 introduced a new setting:

<img width="445" height="77" alt="Image" src="https://github.com/user-attachments/assets/9480b221-87eb-455b-9cba-056fed97a074" />

The word "dirty" should not be used here. It sounds negative. The editor is not dirty it has "unsaved changes".

The documentation e.g. says:

Tabs in the editor area indicate the names of resources that are currently open for editing. An asterisk (*) indicates that an editor has unsaved changes.
The [documentation](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Fconcepts%2Fconcepts-6.htm&cp%3D0_2_3) says: 
"An asterisk indicates that an editor has unsaved changes."

On the other hand https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Freference%2Fref-12.htm&resultof%3D%2522%2564%2569%2572%2574%2579%2522%2520%2522%2564%2569%2572%2574%2569%2522%2520 
uses the term "dirty editor". I would propose to change this to "unsaved changes".

And also this:
<img width="953" height="523" alt="Image" src="https://github.com/user-attachments/assets/ac4955e1-4fed-4a42-a409-439f567f01de" />

and this:
<img width="955" height="327" alt="Image" src="https://github.com/user-attachments/assets/4cc2d9cc-0e21-4b55-b937-2eaad78d17ea" />

should be changed.

@vogella and others: What do you think?

BeckerWdf:
> On the other hand https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Freference%2Fref-12.htm&resultof%3D%2522%2564%2569%2572%2574%2579%2522%2520%2522%2564%2569%2572%2574%2569%2522%2520
> uses the term "dirty editor". I would propose to change this to "unsaved changes".

See: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3802 for this.

vogella:
Fine for me, please provide PR for the change

merks:
Dirty is arguably a dirty word. 😁  But, it's used in other places so I think we should be consistent across those places. Including this one:

[...]

PR #3962, opened 2026-05-04
Don't use "dirty" as user facing term

BeckerWdf:
Relates to:https://github.com/eclipse-platform/eclipse.platform.ui/issues/3961


vogella:
Just go ahead with merging such changes, I feel that I don't have time for text only changes which we already agreed upon.

PR #3963, opened 2026-05-04
Paint circular dirty indicator in editor list popups

### LINKED ISSUE #3960: "*" still used as "dirty indicator" even though is should be drawn as circle
https://github.com/eclipse-platform/eclipse.platform.ui/pull/3823 introduced a possibility to show the "dirty indicator" as a circle overlaying the close button instead of the "*" char.

This works nicely and I like it a lot. But in the Ctrl-E dialog we still see the "*" char at the beginning.
This should be changed to also show the circle at the end.

This is how it currently looks like:
Image

@vogella: Can you pls. have a look at this?

vogella:
The Ctrl+E editor switcher and the tab chevron drop-down kept showing the legacy `*` prefix even when the new tab dirty-indicator style was enabled, which #3960 reported. They now paint the same filled 8 px circle next to the title that `CTabFolderRenderer.drawDirtyIndicator` draws on tabs, so the visual style stays consistent with the tabs the user is switching between.

The drawing lives in a shared `DirtyIndicatorPainter` so the two popups, and any future list views that want the same style, call into one place. When the preference is off, the legacy `*` prefix is unchanged.

Closes #3960

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/BasicPartList.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/internal/workbench/renderers/swt/BasicPartList.java
@@ -72,11 +72,12 @@ public class BasicPartList extends AbstractTableInformationControl {
 
 		@Override
 		public String getText(Object element) {
-			if (element instanceof MDirtyable
-					&& ((MDirtyable) element).isDirty()) {
-				return "*" + ((MUILabel) element).getLocalizedLabel(); //$NON-NLS-1$
+			String label = ((MUILabel) element).getLocalizedLabel();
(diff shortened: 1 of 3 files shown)


BeckerWdf:
I tested this and one little thing should be improved:

image

The circle does not have enough space...

sratz:
Could we right-align the dots in a column? Otherwise it will be very hard to visually scan the list for dirty editors.

vogella:
image

image

vogella:
I plan to merge this tomorrow morning so that this is available in the next release.

BeckerWdf:
> I plan to merge this tomorrow morning so that this is available in the next release.

Looks good on macOS. it's no longer cut off.
image

BeckerWdf (approved):

vogella:
Thanks @BeckerWdf and @sratz for the improvement suggestions

BeckerWdf:
> Thanks @BeckerWdf and @sratz for the improvement suggestions

Thanks for fixing...

PR #3966, opened 2026-05-05
Copilot setup improvements

akurtakov:


akurtakov:
Merging as this will not be verified.

PR #3968, opened 2026-05-05
Modernize layout of Fonts & Colors preference page for better alignment and usability

deepika-u:
Refactored the layout of the Fonts & Colors preference page to improve alignment and usability.

- Aligned filter text field with tree viewer width
- Moved expand actions to a toolbar in the right control column next to filter text in the form of an icon.
- Added way to collapse actions similar to expand to the toolbar itself in the form of an icon.
- Ensured icons are right aligned.
- Improved overall spacing and visual structure to match Eclipse UI conventions similar to Preferences -> Java -> Compiler -> Errors/Warnings page.

Before this pr -> 
image

Now with the pr -> 
image
on clicking the `collapse all` icon, all the sections which are shown expanded gets collapsed.

To maintain consistency across other preference pages like ->
image

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2016 IBM Corporation and others.
+ * Copyright (c) 2003, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -78,6 +78,9 @@ import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.FontDialog;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.Text;
+import org.eclipse.swt.widgets.ToolBar;
+import org.eclipse.swt.widgets.ToolItem;
+import org.eclipse.ui.ISharedImages;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPreferencePage;
 import org.eclipse.ui.PlatformUI;
@@ -614,11 +617,14 @@ public final class ColorsAndFontsPreferencePage extends PreferencePage implement
 	private Button goToDefaultButton;
 
 	/**
-	 * The button to expand the tree.
-	 *
-	 * @since 4.5
+	 * The toolitem to expand the tree.
 	 */
-	private Button expandAllButton;
+	private ToolItem expandAllItem;
+
+	/**
+	 * The toolitem to collapse the tree.
+	 */
+	private ToolItem collapseAllItem;
 
 	/**
 	 * Map of definition FontDefinition-&gt;FontData[] capturing the changes explicitly
@@ -832,40 +838,44 @@ public final class ColorsAndFontsPreferencePage extends PreferencePage implement
 		});
 
 		final SashForm advancedComposite = new SashForm(parent, SWT.VERTICAL);
-		GridData sashData = new GridData(SWT.FILL, SWT.FILL, true, true);
-		advancedComposite.setLayoutData(sashData);
+		advancedComposite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
 
 		Composite mainColumn = new Composite(advancedComposite, SWT.NONE);
-		GridLayout layout = new GridLayout();
-		layout.numColumns = 2;
+		GridLayout layout = new GridLayout(2, false);
 		layout.marginWidth = 0;
 		layout.marginHeight = 0;
 		mainColumn.setFont(parent.getFont());
 		mainColumn.setLayout(layout);
 
-		GridData data = new GridData(GridData.BEGINNING);
-		data.horizontalSpan = 2;
 		Label label = new Label(mainColumn, SWT.LEFT);
 		label.setText(RESOURCE_BUNDLE.getString("colorsAndFonts")); //$NON-NLS-1$
 		myApplyDialogFont(label);
-		label.setLayoutData(data);
+		label.setLayoutData(new GridData(SWT.BEGINNING, SWT.CENTER, false, false, 2, 1));
 
-		createTree(mainColumn);
+		// LEFT COLUMN (tree + filter)
+		Composite leftColumn = new Composite(mainColumn, SWT.NONE);
+		leftColumn.setLayout(new GridLayout(1, false));
+		leftColumn.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true, 1, 1));
 
-		// --- buttons
+		createTree(leftColumn);
+
+		// RIGHT COLUMN (toolbar + buttons)
 		Composite controlColumn = new Composite(mainColumn, SWT.NONE);
-		data = new GridData(GridData.FILL_VERTICAL);
-		controlColumn.setLayoutData(data);
-		layout = new GridLayout();
-		layout.marginHeight = 0;
-		layout.marginWidth = 0;
-		controlColumn.setLayout(layout);
+		controlColumn.setLayout(new GridLayout());
+		controlColumn.setLayoutData(new GridData(SWT.FILL, SWT.FILL, false, true));
 
-		// we need placeholder to offset the filter control of the table
-		Label placeholder = new Label(controlColumn, SWT.NONE);
-		GridData placeholderData = new GridData(SWT.TOP);
-		placeholderData.heightHint = convertVerticalDLUsToPixels(12);
-		placeholder.setLayoutData(placeholderData);
+		// Toolbar at TOP (aligned with filter text)
+		ToolBar treeToolbar = new ToolBar(controlColumn, SWT.FLAT);
+		treeToolbar.setLayoutData(new GridData(SWT.END, SWT.BEGINNING, false, false));
+
+		expandAllItem = new ToolItem(treeToolbar, SWT.PUSH);
+		expandAllItem.setImage(PlatformUI.getWorkbench().getSharedImages().getImage(ISharedImages.IMG_ELCL_EXPANDALL));
+		expandAllItem.setToolTipText(RESOURCE_BUNDLE.getString("expandAll")); //$NON-NLS-1$
+
+		collapseAllItem = new ToolItem(treeToolbar, SWT.PUSH);
+		collapseAllItem
+				.setImage(PlatformUI.getWorkbench().getSharedImages().getImage(ISharedImages.IMG_ELCL_COLLAPSEALL));
+		collapseAllItem.setToolTipText("collapseAll"); //$NON-NLS-1$
 
 		fontChangeButton = createButton(controlColumn, RESOURCE_BUNDLE.getString("openChange")); //$NON-NLS-1$
 		fontSystemButton = createButton(controlColumn, WorkbenchMessages.FontsPreference_useSystemFont);
@@ -874,41 +884,28 @@ public final class ColorsAndFontsPreferencePage extends PreferencePage implement
 		editDefaultButton = createButton(controlColumn, RESOURCE_BUNDLE.getString("editDefault")); //$NON-NLS-1$
 		goToDefaultButton = createButton(controlColumn, RESOURCE_BUNDLE.getString("goToDefault")); //$NON-NLS-1$
 		createSeparator(controlColumn);
-		expandAllButton = createButton(controlColumn, RESOURCE_BUNDLE.getString("expandAll")); //$NON-NLS-1$
-		expandAllButton.setEnabled(true);
-		// --- end of buttons
+		// end of buttons
 
 		createDescriptionControl(mainColumn);
 
 		Composite previewColumn = new Composite(advancedComposite, SWT.NONE);
-		GridLayout previewLayout = new GridLayout();
-		previewLayout.marginTop = 7;
-		previewLayout.marginWidth = 0;
-		previewLayout.marginHeight = 0;
-		previewColumn.setFont(parent.getFont());
-		previewColumn.setLayout(previewLayout);
+		previewColumn.setLayout(new GridLayout());
+		previewColumn.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
 
-		// --- create preview control
+		// create preview control
 		Composite composite = new Composite(previewColumn, SWT.NONE);
-
-		GridData data2 = new GridData(GridData.FILL_BOTH);
-		composite.setLayoutData(data2);
-		GridLayout layout2 = new GridLayout(1, true);
-		layout2.marginHeight = 0;
-		layout2.marginWidth = 0;
-		composite.setLayout(layout2);
+		composite.setLayout(new GridLayout());
+		composite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
 
 		Label label2 = new Label(composite, SWT.LEFT);
 		label2.setText(RESOURCE_BUNDLE.getString("preview")); //$NON-NLS-1$
 		myApplyDialogFont(label2);
 
 		previewComposite = new Composite(composite, SWT.NONE);
-		previewComposite.setLayoutData(new GridData(GridData.FILL_BOTH));
+		previewComposite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
 		stackLayout = new StackLayout();
-		stackLayout.marginHeight = 0;
-		stackLayout.marginWidth = 0;
 		previewComposite.setLayout(stackLayout);
-		// -- end of preview control
+		// end of preview control
 
 		defaultFontPreview = createFontPreviewControl();
 		defaultColorPreview = createColorPreviewControl();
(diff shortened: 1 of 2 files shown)


deepika-u:
@vogella - Can you take a look at this when you get some time please?

vogella:
@deepika-u I have the problem that the expand buttons are IMHO the ugliest buttons we have in Eclipse. Everytime I see these ugly button gradients I feel the desire to remove them.

So I think I 'm not the right person to review this

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
```diff
@@ -614,11 +618,21 @@ public Color getBackground(Object element) {
...
 	 * @since 4.5
 	 */
-	private Button expandAllButton;
+	private ToolItem expandAllItem;
+
+	/**
+	 * The toolitem to collapse the tree.
+	 *
+	 * @since 4.5
+	 */
```
merks: It's not API so no need for `@since`.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
```diff
@@ -829,43 +843,69 @@ protected Control createContents(Composite parent) {
 			if (appliedDialogFont != null) {
 				appliedDialogFont.dispose();
 			}
+	        // Dispose custom images
+	        if (expandImage != null && !expandImage.isDisposed()) {
```
merks: I believe that you don't need an isDisposed guard just as there isn't one on the appliedDialogFont.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
```diff
@@ -829,43 +843,69 @@ protected Control createContents(Composite parent) {
...
 
 		final SashForm advancedComposite = new SashForm(parent, SWT.VERTICAL);
-		GridData sashData = new GridData(SWT.FILL, SWT.FILL, true, true);
-		advancedComposite.setLayoutData(sashData);
+		advancedComposite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
 
 		Composite mainColumn = new Composite(advancedComposite, SWT.NONE);
-		GridLayout layout = new GridLayout();
-		layout.numColumns = 2;
+		GridLayout layout = new GridLayout(2, false);
```
merks, line 844: You're inlining, which is fine, but as a reviewer it's not easy to see that this is identical.  I trust that it is.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
```diff
@@ -829,43 +843,69 @@ protected Control createContents(Composite parent) {
...
 		layout.marginWidth = 0;
 		layout.marginHeight = 0;
-		mainColumn.setFont(parent.getFont());
```
merks, line 843: Why this change?  This might have an impact that I don't notice.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
```diff
@@ -829,43 +843,69 @@ protected Control createContents(Composite parent) {
...
+		controlColumn.setLayoutData(new GridData(SWT.FILL, SWT.FILL, false, true));
+
+		// Toolbar at TOP (aligned with filter text)
+		ToolBar treeToolbar = new ToolBar(controlColumn, SWT.FLAT);
+		treeToolbar.setLayoutData(new GridData(SWT.END, SWT.BEGINNING, false, false));
+
+		// LOAD IMAGES ONCE
+	    ImageDescriptor expandDesc = ResourceLocator
+				.imageDescriptorFromBundle("org.eclipse.ui", "icons/full/elcl16/expandall.svg") //$NON-NLS-1$ //$NON-NLS-2$
+	        .orElse(null);
```
merks: If this were to fail, the button will be useless (horribly ugly) won't it?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
```diff
@@ -974,7 +1001,8 @@ private String getText(Object element) {
 
 		tree = new FilteredTree(parent, SWT.SINGLE | SWT.H_SCROLL | SWT.V_SCROLL | SWT.BORDER, filter, true);
```
merks: It might be nice to fix this given so many other changes here.

merks (changes requested):
Overall it looks good, but I've only looked at the changes here, not actually running in the IDE.  

The setFont changes concern me a bit because I don't know the impact and you've not shown pictures of the impact.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
deepika-u: oh ok removed them, thank you.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
deepika-u: updated as per appliedDialogFont.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
deepika-u: Thanks for calling that out. Yes, this is functionally identical  - new GridLayout(2, false) is equivalent to setting numColumns = 2 with the default makeColumnsEqualWidth = false.

I agree that the previous form may be easier to compare in a diff. I can revert to the expanded version if you prefer for readability.

deepika-u:
> @deepika-u I have the problem that the expand buttons are IMHO the ugliest buttons we have in Eclipse. Everytime I see these ugly button gradients I feel the desire to remove them.
> 
> So I think I 'm not the right person to review this

@Vogella
Thanks for the feedback — I understand your concern about the visual style of those buttons.

In this change, the goal wasn’t to introduce any new visual design, but to improve layout consistency and usability (alignment of the filter, tree, and action controls). The icons used are existing Eclipse resources, so this mainly repositions current actions and adds the complementary expand/collapse functionality that was missing.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
merks: Actually it's more an observation.  I like it simpler the way you wrote it!  I trust that you are properly testing all the layout and appearance that the simpler expression produces good behavior.

merks:
> I understand your concern about the visual style of those buttons.

Indeed.  We are all aware that other folks are busily creating a new dual tone icon pack so one might imagine a pretty icon instead so as better to focus on the design and functionality.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
deepika-u: Good point — thanks for calling that out.

This line was removed during layout refactoring, but it’s not strictly required since SWT controls inherit the parent font by default.

That said, I agree that keeping it makes the intent explicit and avoids any platform-specific inconsistencies. Shall i restore it to be safe and keep behavior identical to the original implementation.

vogella:
@deepika-u all good, just wanted to explain why I don't review this one

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
deepika-u:
```
		ImageDescriptor collapseDesc = ResourceLocator
				.imageDescriptorFromBundle("org.eclipse.ui", "icons/full/elcl16/collapseall.svg") //$NON-NLS-1$ //$NON-NLS-2$
			.orElse(null);

		collapseAllItem = new ToolItem(treeToolbar, SWT.PUSH);

		if (collapseDesc != null) {
			collapseImage = collapseDesc.createImage();
			collapseAllItem.setImage(collapseImage);
		} else {
			// Fallback to shared image
			collapseAllItem.setImage(
				PlatformUI.getWorkbench().getSharedImages()
					.getImage(org.eclipse.ui.ISharedImages.IMG_ELCL_COLLAPSEALL));
		}

		collapseAllItem.setToolTipText("Collapse All"); //$NON-NLS-1$
```
Thought of this way. But in the similar lines, icon for Expand All is not readily available in org.eclipse.ui.ISharedImages, how should this be handled in your opinion?	can i use org.eclipse.ui.ISharedImages.IMG_ELCL_COLLAPSEALL or IMG_ELCL_REMOVE?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
merks: Yes, better safe than sorry I think.

deepika-u:
> all good, just wanted to explain why I don't review this one

Got it, thanks for clarifying.
Appreciate the context!

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
merks: I think in the past the tendency was to ensure that the icon was available locally in the bundle rather than to use fragile programmatic access.

Let's ask @HeikoKlare @BeckerWdf how one can best do something here that's not fragile.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
deepika-u: Updated it to `tree = new FilteredTree(parent, SWT.SINGLE | SWT.H_SCROLL | SWT.V_SCROLL | SWT.BORDER, filter, true, true);` 
I’ve updated the deprecated constructor to the newer variant using the additional parameter for fast lookup. This keeps the behavior consistent while removing the warning.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
deepika-u: Done.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/themes/ColorsAndFontsPreferencePage.java
BeckerWdf: you could simply copy the PNG files into the bundle.
Alternatively add  "Expand All" as org.eclipse.ui.ISharedImages as API and use it.

deepika-u: Sorry, let me correct my update and put it across again in some time.

deepika-u: Thanks for the insights @BeckerWdf and @merks
I tried exploring that way too but was unsure how to achieve it correct and complete.

My findings are : icons/images are already existing in below paths.
1) ..\git\eclipse.platform.ui\bundles\org.eclipse.ui.ide\icons\full\elcl16 but here expandall.svg/png are not available.
2) ..\git\eclipse.platform.ui\bundles\org.eclipse.ui.editors\icons\full\elcl16 has both.
3) ..\git\eclipse.platform.ui\bundles\org.eclipse.ui\icons\full\elcl16 has both.
4) ..\git\eclipse.platform.ui\bundles\org.eclipse.ui.navigator\icons\full\elcl16 but here expandall.svg/ png are not available.

So for this pr the respective java file is at "..\git\eclipse.platform.ui\bundles\org.eclipse.ui.workbench\eclipseui\org\eclipse\ui\internal\themes\ColorsAndFontsPreferencePage.java"

Option 1 - Add respective expandall, collapseall icons svg/png at freshly created path for this plugin "..\git\eclipse.platform.ui\bundles\org.eclipse.ui.workbench\icons\full\elcl16" and use like below

ImageDescriptor expandDesc = AbstractUIPlugin.imageDescriptorFromPlugin(
    "org.eclipse.ui.workbench",
    "icons/full/elcl16/expandall.svg"
);

expandImage = expandDesc.createImage();
expandAllItem.setImage(expandImage);

Option 2 - For API way, if i add "icons/full/elcl16/expandall.svg" corresponding entries(inline with collapseall.svg which is already accessible) in ISharedImages.java and WorkbenchImages.java. Are these the only 2 steps i need to perform or do i have to add anything else which i might have missed.

Another observation as well.
I also see entries in plugin.xml at "..\git\eclipse.platform.ui\bundles\org.eclipse.ui" as below
```
      <image
            commandId="org.eclipse.ui.navigate.collapseAll"
            icon="$nl$/icons/full/elcl16/collapseall.svg">
      </image>
      <image
            commandId="org.eclipse.ui.navigate.expandAll"
            icon="$nl$/icons/full/elcl16/expandall.svg">
      </image>
```

So anything in similar lines needs to be done? But I do understand these are used by the command framework and are not directly applicable for ToolItem usage but better to do, so that it is available as a generic solution if anywhere else - it can also be used(since collapseall is already available but not expandall)
Please let me know if my understanding is correct. Thanks in advance.

BeckerWdf: > Are these the only steps i need to perform or do i have to add anywhere else which i missed.

I think yes.

merks: The expand-all and collapse-all I think are heavily reused so it would seem a good thing for them to be shared across those usages, e.g.,

image

deepika-u: To keep this as a generic one for everyone to use -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/3981

deepika-u: Updated as per #3981 merge.

merks:
At a glance it looks better.   Please rebase and force push a single commit.  I'll try to find time to actually pull the changes and test them locally.

deepika-u:
> At a glance it looks better. Please rebase and force push a single commit. I'll try to find time to actually pull the changes and test them locally.

All done :) Thanks aton in advance.

merks (approved):
Functionally it's fine.

I'm not sure if it's slightly awkward that the buttons are next to the search field while they act upon the tree?

@BeckerWdf 

Do you have an opinion on that.  Otherwise work fine for me.

BeckerWdf:
> Functionally it's fine.
> 
> I'm not sure if it's slightly awkward that the buttons are next to the search field while they act upon the tree?
> 
> @BeckerWdf
> 
> Do you have an opinion on that. Otherwise work fine for me.

I also had this reaction initially. But on the other hand on the "Errors / Warnings" page it's similar - maybe similarly awkward. So I don't have a clear opinion. I can try to ask some colleagues of mine what they think.

BeckerWdf:
I just thought about it again and also looked at other places in the IDE. I think the icons are a improvement compared to the current situation as the "Expand All" button is really fare away from the tree.

I also don't have a better idea where to place the icons. And as we have the same pattern on the "Warnings / Errors" page it's ok to have that also here. Consistency over different UIs in the same application is always good to have.

If we find a better layout we anyway should adapt all the places.

So I would say: Let's merge this.

merks:
Thank you for the nice improvement(s)!

deepika-u:
Thanks alot for the timely support and happy to hear various perspectives.

BeckerWdf:
@deepika-u: Should we add this to the New & Noteworthy document?

deepika-u:
Thanks for asking @BeckerWdf 
I think it could be worth a small entry in New & Noteworthy since the change improves both usability and UI consistency of the Fonts & Colors preference page.

In particular:
- added a dedicated collapse action (previously there was no direct way to collapse the full tree),
- replaced the older-style expand button with toolbar-based actions,
- improved alignment and layout consistency with other Eclipse preference pages.

Although the functionality itself is small, it is a visible user-facing improvement in a commonly used preferences area, so a brief N&N mention may help highlight the modernization effort.

BeckerWdf:
@deepika-u: So then I would propose that you provide a PR for the N&Ns.

deepika-u:
@BeckerWdf - added one as discussed -> https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/550

Thanks alot.

PR #3971, opened 2026-05-06
Fix sticky scroll height calculation for variable line heights

Christopher-Hermann:
**Description:**  
`StickyScrollingControl` was computing the sticky area height by multiplying a single line height by the number of sticky lines. This assumption breaks when content like emoji causes individual lines to render taller than the default — the canvas ends up too short, clipping the sticky lines visually.

`calculateCanvasBounds` now sums the actual height of each sticky line using `getLineHeight(offset)` instead of using a uniform `getLineHeight()`. 

**Reproduce:**  
Create an XML file with the content below. Enable sticky scrolling and scroll a few lines:
                                                                                                                       
```xml
<?xml version="1.0" encoding="UTF-8"?>
<test>
	<test2 name="👍">
		<test3 name="👎" />
		<!--
    Extended comment section 
    to demonstrate 
    sticky scrolling 
    functionality
  -->
	</test2>
</test>
```                                                                                                                 

image

Follow up of #3748

### DIFF
--- a/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.editors; singleton:=true
-Bundle-Version: 3.22.0.qualifier
+Bundle-Version: 3.22.100.qualifier
 Bundle-Activator: org.eclipse.ui.internal.editors.text.EditorsPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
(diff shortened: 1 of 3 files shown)


iloveeclipse:
There is a test failure in `testCanvasBoundsHeightAdjustsForVariableLineHeights (org.eclipse.ui.internal.texteditor.stickyscroll.StickyScrollingControlTest)`

Christopher-Hermann:
> There is a test failure in `testCanvasBoundsHeightAdjustsForVariableLineHeights (org.eclipse.ui.internal.texteditor.stickyscroll.StickyScrollingControlTest)`

Fixed in the latest version 👍

BeckerWdf:
@HeikoKlare: We missed to merge this last week for M3 - thanks to the bridge day.
The change looks straightforward.
What do you think? Is this something we still can merge for RC1?

BeckerWdf (approved):

PR #3972, opened 2026-05-06
Fix E4 model perspectives being converted to orphan local copies

akurtakov:
E4 model perspectives are being made into local copies as it counted them as "orphan" with log like:
```
!ENTRY org.eclipse.ui 1 1 2026-05-06 12:00:28.236
!MESSAGE Perspective with name 'Some name' and id 'org.some.id' has been
made into a local copy
```

Perspectives with contributorURI in fragments are fully defined and not orphan thus add them directly without going through orphan and local copy.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2020 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -3908,8 +3908,19 @@ public class WorkbenchPage implements IWorkbenchPage {
 			PerspectiveDescriptor perspectiveDesc = (PerspectiveDescriptor) getPerspectiveDesc(
 					mperspective.getElementId());
 			if (perspectiveDesc == null) {
-				fixedPerspective = true;
-				perspectiveDesc = fixOrphanPerspective(mperspective);
+				if (mperspective.getContributorURI() != null) {
+					PerspectiveRegistry reg = (PerspectiveRegistry) PlatformUI.getWorkbench().getPerspectiveRegistry();
+					// Registers an E4 perspective that is contributed directly via the application
+					// model (e.g. via a model fragment), without renaming it or marking it as a
+					// local copy. Unlike PerspectiveRegistry#addPerspective(MPerspective),
+					// this method does not add the perspective to the application snippets.
+					reg.createDescriptor(mperspective);
+					perspectiveDesc = (PerspectiveDescriptor) getPerspectiveDesc(mperspective.getElementId());
+					sortedPerspectives.add(perspectiveDesc);
+				} else {
+					fixedPerspective = true;
+					perspectiveDesc = fixOrphanPerspective(mperspective);
+				}
 			}
 			Perspective p = new Perspective(perspectiveDesc, mperspective, this);
 			modelToPerspectiveMapping.put(mperspective, p);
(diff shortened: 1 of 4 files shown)


akurtakov:
Problem can be seen with IWorkbenchPageTeset.testE4PerspectiveNotConvertedToLocalCopy and WorkbenchPage changes are reverted.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IWorkbenchPageTest.java
```diff
@@ -3349,6 +3354,43 @@ public void testGetViewStackWithSecondaryId() throws PartInitException {
...
+		WorkbenchWindow workbenchWindow = (WorkbenchWindow) window;
+		List<MPerspectiveStack> stacks = modelService.findElements(workbenchWindow.getModel(), null,
+				MPerspectiveStack.class);
+		assertFalse("Perspective stack should exist in the window model", stacks.isEmpty());
+		MPerspectiveStack perspectiveStack = stacks.get(0);
+
+		String e4Id = "org.eclipse.ui.tests.internal.e4.perspective.model";
+		String e4Label = "E4ModelPerspective";
+
+		assertNull("e4 perspective ID should not be inperspective registry",
```
merks:
```suggestion
		assertNull("e4 perspective ID should not be in perspective registry",
```

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IWorkbenchPageTest.java
```diff
@@ -3349,6 +3354,43 @@ public void testGetViewStackWithSecondaryId() throws PartInitException {
...
+
+		assertEquals("Element ID should not have changed", e4Id, e4Perspective.getElementId());
+
+		IPerspectiveDescriptor descriptor = PlatformUI.getWorkbench().getPerspectiveRegistry()
+				.findPerspectiveWithId(e4Id);
+		assertNotNull(descriptor);
+
+		assertFalse(((PerspectiveDescriptor) descriptor).hasCustomDefinition());
+
+		PlatformUI.getWorkbench().getPerspectiveRegistry().deletePerspective(descriptor);
```
merks, line 3391: Is this line cleanup for after the test to avoid this leaking into other tests?  If the test fails, might this leak anyway and then cause other tests to fail too?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchPage.java
```diff
@@ -3908,8 +3908,19 @@ private Perspective getPerspective(MPerspective mperspective) {
...
 					mperspective.getElementId());
 			if (perspectiveDesc == null) {
-				fixedPerspective = true;
-				perspectiveDesc = fixOrphanPerspective(mperspective);
+				if (mperspective.getContributorURI() != null) {
+					PerspectiveRegistry reg = (PerspectiveRegistry) PlatformUI.getWorkbench().getPerspectiveRegistry();
+					// Registers an E4 perspective that is contributed directly via the application
+					// model (e.g. via a model fragment), without renaming it or marking it as a
+					// local copy. Unlike {@link #addPerspective(MPerspective)}, this method does
+					// not add the perspective to the application snippets.
```
merks: It's nice that you add a description so that future developers understand the reason for the guard and the special handling.

merks (approved):
Just one typo and one question about the test cleanup.  Otherwise it looks fine.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IWorkbenchPageTest.java
akurtakov: Failures in other tests can occur only if the same ids are used or number of elements are checked inside the same suite. As every test suite creates new workbench it shouldn't be possible to leak into other test suites are these model changes should be gone on new startup as they are not backed by any permanent source (e.g. fragment.e4xmi). With all of the above, it is still nice to clean up when we can to prevent time lost in the future.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/IWorkbenchPageTest.java
akurtakov: Fixed.

vogella:
LGTM

PR #3973, opened 2026-05-06
QuickAccessDialog: defer built-in provider construction until first use

vogella:
The dialog constructor previously instantiated all nine built-in providers (Editor, View, Perspective, Command, Action, Wizard, Preference, Properties, HelpSearch) eagerly inside the open-time BusyIndicator, even when the user closed the dialog without typing. This wraps each in a LazyQuickAccessProvider that exposes the provider id and requiresUiAccess flag as static metadata and resolves the delegate only on the first getElements/findElement call, so identity and threading queries during dialog setup do not force construction.

As a side benefit, it should marginally help stabilize QuickAccessDialogTest.testPreviousChoicesAvailableForExtension on Windows by reducing constructor work and the initial-layout shell resize events that can preempt the compute job.


vogella:
Marking as draft so it is clear that this is NOT planned for 4.40 (it is planned for 4.41)

PR #3975, opened 2026-05-06
Remove unused BootstrapTheme3x and its dummy.css

vogella:
`BootstrapTheme3x` was a reflective bootstrap helper added in 2010 (bug 310153, commit 99e054cf44) so `org.eclipse.ui.internal.Workbench` could activate the e4 styling engine without taking a hard dependency on `org.eclipse.e4.ui.css.swt.theme`. The Workbench call site was removed in 2011 by commit ad3c59ac43 (the big "Merge in the e4.ui stuff" merge that brought the e4 UI stack into the platform), when styling initialization moved into the e4-based workbench. Nothing in the platform tree has referenced the class since, and the `dummy.css` it registered as the default theme is empty. This drops the class, the placeholder CSS, the now-empty `css/` entry from `build.properties`, and bumps the bundle micro version.


merks:
I think you will have major problems with PRs not based on the current master.

merks (approved):
Off course it looks fine. 😀

PR #3976, opened 2026-05-07
Drop unreachable IOException from CSSEngine String overloads

vogella:
parseSelectors(String), parsePropertyValue(String), and parseStyleDeclaration(String) wrap the input in a StringReader and delegate to the Reader-based overload. StringReader.read() never throws IOException, but the signature of the Reader overload propagates IOException up from the SAC parser API, so the String forms used to declare a checked exception that could not actually fire. Drop throws IOException from the three String overloads on the internal CSSEngine interface and wrap the inner Reader call in AbstractCSSEngine in try/catch -> IllegalStateException to preserve the impossible case as a defensive guard. Bundle is x-friends so the signature change is internal-only.

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.core/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.css.core/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-SymbolicName: org.eclipse.e4.ui.css.core;singleton:=true
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
-Bundle-Version: 0.14.600.qualifier
+Bundle-Version: 0.14.700.qualifier
 Export-Package: org.eclipse.e4.ui.css.core;x-internal:=true,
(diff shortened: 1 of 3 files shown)


iloveeclipse:
This breaks PDE compile.
```
Description	Resource	Path	Location	Type
Unreachable catch block for IOException. This exception is never thrown from the try statement body	CssSpyPart.java	/org.eclipse.pde.spy.css/src/org/eclipse/pde/spy/css	line 977	Java Problem
```

vogella:
I have a look

vogella:
> This breaks PDE compile.
> 
> ```
> Description	Resource	Path	Location	Type
> Unreachable catch block for IOException. This exception is never thrown from the try statement body	CssSpyPart.java	/org.eclipse.pde.spy.css/src/org/eclipse/pde/spy/css	line 977	Java Problem
> ```

Thanks. Fixed with https://github.com/eclipse-pde/eclipse.pde/pull/2327

vogella:
Validated with a local aggregator build containing [8ccf90a](https://github.com/eclipse-platform/eclipse.platform.ui/commit/8ccf90a4af1766246651634c326763cdd0a7522d) and the PDE fix.

PR #3978, opened 2026-05-07
Remove unused CSS serializer classes and dead color converter config

vogella:
The CSSSerializer / CSSSerializerConfiguration / CSSHTMLSerializerConfiguration trio in org.eclipse.e4.ui.css.core.serializers and CSSSWTSerializerConfiguration in org.eclipse.e4.ui.css.swt.serializers have no callers in the production tree or in any built test bundle. Their only references are in tests/.../tkuiTestsToRefactor, the quarantined refactor backlog that is not part of any test build. Drop the four classes and the matching x-friends/x-internal exports for the two serializers packages from the bundle manifests.

Drop ICSSValueConverterColorConfig and CSSValueConverterConfigColorImpl in the same pass: the COLOR_HEXA_FORMAT_CONFIG / COLOR_NAME_FORMAT_CONFIG / COLOR_RGB_FORMAT_CONFIG instances those classes provide are never constructed by any caller, so the corresponding instanceof branch in CSS2ColorHelper.getColorStringValue is unreachable and the helper always returns the hex form. Simplify the helper to drop its ICSSValueConverterConfig parameter and update the two SWT color converters that called it.

All bundles internal (x-friends/x-internal), no API surface affected.


vogella:
A grep across eclipse.pde for any of those names or the two *.serializers package paths returns zero matches, and the PR description notes the affected packages are x-friends/x-internal with no API surface change. So this PR should not affect the eclipse.pde repo.

PR #3981, opened 2026-05-07
Expose Expand All icon through ISharedImages API

deepika-u:
This change exposes the existing Expand All toolbar icon through the shared workbench image API, similar to the already available Collapse All image.

Currently, collapseall is accessible via ISharedImages, while expandall requires bundle-specific image loading or direct resource access. This leads to duplicated image handling and inconsistent usage across the workbench.

This PR:
- Adds shared image constants and registrations for the existing expandall icon
- Makes the icon accessible through PlatformUI.getWorkbench().getSharedImages()
- Aligns Expand All usage with the existing Collapse All API pattern
- Avoids fragile programmatic bundle resource lookups in downstream consumers

The intent is to provide a reusable, centralized API for any component that needs standard Expand All toolbar imagery.

This arises from https://github.com/eclipse-platform/eclipse.platform.ui/pull/3968#discussion_r3199104404

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -244,7 +244,7 @@ public abstract class FilteredItemsSelectionDialog extends SelectionStatusDialog
 	 * to deterministically wait for the filter/refresh pipeline to settle.
 	 *
 	 * @noreference This field is not intended to be referenced by clients.
-	 * @since 3.138
+	 * @since 3.139
 	 */
 	public static final Object JOB_FAMILY = new Object();
 
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchImages.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchImages.java
@@ -189,8 +189,8 @@ public/* final */class WorkbenchImages {
 		declareImage(ISharedImages.IMG_ELCL_SYNCED, ISharedImages.IMG_ELCL_SYNCED_DISABLED, //
 				PATH_ELOCALTOOL + "synced.svg", true); //$NON-NLS-1$
 
-		declareImage(ISharedImages.IMG_ELCL_COLLAPSEALL, ISharedImages.IMG_ELCL_COLLAPSEALL_DISABLED, //
-				PATH_ELOCALTOOL + "collapseall.svg", true); //$NON-NLS-1$
+		declareImage(ISharedImages.IMG_ELCL_EXPANDALL, ISharedImages.IMG_ELCL_EXPANDALL_DISABLED, //
+				PATH_ELOCALTOOL + "expandall.svg", true); //$NON-NLS-1$
 
 		declareImage(ISharedImages.IMG_ELCL_REMOVE, ISharedImages.IMG_ELCL_REMOVE_DISABLED, //
 				PATH_ELOCALTOOL + "remove.svg", true); //$NON-NLS-1$
--- a/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench; singleton:=true
-Bundle-Version: 3.138.100.qualifier
+Bundle-Version: 3.139.0.qualifier
 Bundle-Activator: org.eclipse.ui.internal.WorkbenchPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/ISharedImages.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/ISharedImages.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -70,6 +70,20 @@ public interface ISharedImages {
 	 * @since 3.4
 	 */
 	String IMG_ELCL_COLLAPSEALL_DISABLED = "IMG_ELCL_COLLAPSEALL_DISABLED"; //$NON-NLS-1$
+
+	/**
+	 * Identifies the expand all image in the enabled state.
+	 *
+	 * @since 3.139
+	 */
+	String IMG_ELCL_EXPANDALL = "IMG_ELCL_EXPANDALL"; //$NON-NLS-1$
+
+	/**
+	 * Identifies the expand all image in the disabled state.
+	 *
+	 * @since 3.139
+	 */
+	String IMG_ELCL_EXPANDALL_DISABLED = "IMG_ELCL_EXPANDALL_DISABLED"; //$NON-NLS-1$
 
 	/**
 	 * Identifies the remove image in the enabled state.


deepika-u:
@BeckerWdf : Can you take a look at it when you get some time please?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
```diff
@@ -244,7 +244,7 @@ public abstract class FilteredItemsSelectionDialog extends SelectionStatusDialog
 	 * to deterministically wait for the filter/refresh pipeline to settle.
 	 *
 	 * @noreference This field is not intended to be referenced by clients.
-	 * @since 3.138
+	 * @since 3.139
```
merks, line 247: What's changed here? When was this added?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchImages.java
```diff
@@ -189,8 +189,8 @@ private static final void declareImages() {
 		declareImage(ISharedImages.IMG_ELCL_SYNCED, ISharedImages.IMG_ELCL_SYNCED_DISABLED, //
 				PATH_ELOCALTOOL + "synced.svg", true); //$NON-NLS-1$
 
-		declareImage(ISharedImages.IMG_ELCL_COLLAPSEALL, ISharedImages.IMG_ELCL_COLLAPSEALL_DISABLED, //
-				PATH_ELOCALTOOL + "collapseall.svg", true); //$NON-NLS-1$
```
merks, line 193: Is this really being removed?

merks (changes requested):
I think you removed a registration and you changes an @since for something you didn't add.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchImages.java
deepika-u: At line 201, there exists an entry for COLLAPSEALL and this is a duplicate of it, so updated the same for EXPANDALL. Sorry for causing the confusion from diff perspective.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
deepika-u: Without this change, it was showing an error in this file. Just the version update to resolve the error. Otherwise i dint have anything to update in this file.

deepika-u: This was the error i was seeing `Invalid @since 3.138 tag on JOB_FAMILY; expecting @since 3.139`, so had to update it.

merks: I see, this was added by this commit during this release cycle:

Image

With a filter to suppress the API error.

I'm not sure what @vogella has in mind here? 

I this case I think changing it to 3.139 makes sense.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchImages.java
deepika-u: There is already an entry at 201 for IMG_ELCL_COLLAPSEALL, So i updated other instance to IMG_ELCL_EXPANDALL as conveyed above.
https://github.com/deepika-u/eclipse.platform.ui/blob/ac6bfa98d4cc0887888ab89f022b3267dac6c961/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WorkbenchImages.java#L201-L202

merks: Oh!  Then cleanup is good.

merks (approved):
Given your comments, the changes look good!

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
HeikoKlare: But then the filter can/should be removed now. I seems strange to add something with `@since 3.138` which is effectively only present in `3.138.100` and above. In my opinion, there should have been a bump to 3.139 in the previous PR already.

merks: Yes, best to clean up the bogus filter!  And yes, the current state looks like a mistake that no one reviewed.

merks: @deepika-u 

Specifically, please delete this:

image

bundles/org.eclipse.ui.workbench/.settings/.api_filters
```diff
@@ -1,13 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <component id="org.eclipse.ui.workbench" version="2">
-    <resource path="eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java" type="org.eclipse.ui.dialogs.FilteredItemsSelectionDialog">
-        <filter id="336658481">
-            <message_arguments>
-                <message_argument value="org.eclipse.ui.dialogs.FilteredItemsSelectionDialog"/>
-                <message_argument value="JOB_FAMILY"/>
-            </message_arguments>
-        </filter>
-    </resource>
```
merks, line 10: Thank you!

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java
deepika-u: Done, thanks alot for your time.

bundles/org.eclipse.ui.workbench/.settings/.api_filters
deepika-u: Thanks for your time :)

deepika-u:
I am still confused in 1 context which is again conveyed by the UT failures, 

image
Here should it be not 3.5 or 4.0 and where from this 5 ideally routes from? And not 3.139(this comes from plugin.xml i agree) in this file.
Hope i am still not too late. Sorry to ask but still.

merks:
Sorry, sorry.  The API tools are really fussy and consider an addition a constant to an API class or interface an API breaking change.  But we don't want that so each of them needs to be suppressed, preferably with a commented suppression  All the new added things (public constants) should be marked `@since 3.139` including the thing Lars added. 

Do you see these errors in your IDE?  That's the best place to added the suppression via quick fixes.

I do see this one (after some refreshing and forced builds): 

image

I guess asking you to remove the filter was bad advice.   Sorry.  😢

merks:
I'm not a big API tools expert.  I'm not sure if both of these are needed
```
    <resource path="META-INF/MANIFEST.MF">
        <filter comment="See https://github.com/eclipse-platform/eclipse.platform.ui/pull/3981" id="923795461">
            <message_arguments>
                <message_argument value="3.139.0"/>
                <message_argument value="3.138.0"/>
            </message_arguments>
        </filter>
    </resource>
    <resource path="eclipseui/org/eclipse/ui/dialogs/FilteredItemsSelectionDialog.java" type="org.eclipse.ui.dialogs.FilteredItemsSelectionDialog">
        <filter comment="See https://github.com/eclipse-platform/eclipse.platform.ui/pull/3981" id="336658481">
            <message_arguments>
                <message_argument value="org.eclipse.ui.dialogs.FilteredItemsSelectionDialog"/>
                <message_argument value="JOB_FAMILY"/>
            </message_arguments>
        </filter>
    </resource>
```

image

merks:
I think this one is sufficient:

image

deepika-u:
> I think this one is sufficient:

Reverted this change. Let me see the UT how it goes.

merks:
Sorry we are all strapped for time and then give advice simply by looking at a web page which doesn't always tell the whole story.  But I'm glad you are doing this in the best way for future flexibility!  🏆

deepika-u:
> Sorry we are all strapped for time and then give advice simply by looking at a web page which doesn't always tell the whole story. But I'm glad you are doing this in the best way for future flexibility! 🏆

Thanks for the encouragement! I agree that the full context isn’t always obvious from a quick look, so I’m trying to optimize for future flexibility and fewer constraints later on.

akurtakov:
Fastest path forward IMO is having separate/dedicated PRs that do the cleanups (e.g. remove the duplicated initialization, fix apitools, etc.) . These can get merged fast/unquestioned making the later PR easier to review.

deepika-u:
Hope we are good to go to merge this pr, thanks for all the inputs from everyone :)

PR #3989, opened 2026-05-08
Evaluate visibleWhen expressions for "Show in" menu

### LINKED ISSUE #3969: Handled Menu Item Contributions to ShowInMenu do not evaluate Visible-When Expression
Handled Menu Items contributed to the "ShowIn"-Menu do not evaluate Visible-When Expressions. Because of this problem, entries in the "ShowIn"-Menu can be disabled, but cannot be made invisible.

## Steps to reproduce

Reproducing this problem inside the workspace requires additional PDE bundles (in order to create and modify a new E4 model fragment).

Since this is a E4 problem, the easiest reproduction I found for this is by modifying the "org.eclipse.e4.ui.examples.e4editor" test bundle and adding this bundle to my run configuration:

1. Open the E4 fragment editor for /org.eclipse.e4.ui.examples.e4editor/fragment.e4xmi
2. Add a new menu contribution with parent-ID org.eclipse.ui.menus.showInMenu
3. Add a new handled menu item with a matching command and handler and connect them to each other
4. Add a imperative expression for the menu item
5. Create a new handler class which implements (at)Evaluate, (at)CanExecute and (at)Execute as follows (both the handled menu item and the imperative expression must point to this class):

Image

6. Add the "org.eclipse.e4.ui.examples.e4editor" bundle to a run configuration and start the application
7. Inside the runtime application, create a new project and open the "ShowIn"-menu (via the context menu) -> The problem occurs: The added menu item is show as disabled although the imperative expression we've added for this item should make it invisible (since the "Evaluate" method of our handler always returns false)

Image

For comparison: Change the parent-ID of the handled menu item to "popup" and restart the application -> The item is now an invisible entry in the popup menu (can be tested via breakpoints or by changing the "Evaluate" method to return true).

Note: I've tried fixing this issue as a PR but stopped, since it looks like fixing this would require a bigger change in the way the "ShowIn"-Menu is filled. As far as I can tell the root cause for this problem seems to be inside org.eclipse.ui.internal.ShowInMenu.fillMenu(IMenuManager). Here, the handled menu items get converted to CommandContributionItem without considering if a visibleWhen-Expression is present on the item.

Image

## Community

r-mennig:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3969

Currently, visibleWhen expressions are not evaluated for the "ShowIn"-menu. Because of this problem, entries in the "ShowIn"-menu can be disabled, but cannot be made invisible.

This PR fixes the issue by evaluating visibleWhen expressions before creating the command contribution items for the menu.

This also fixes an existing problem for the "Show in" -> "System explorer" entry. According to the definition of the menu contribution, this entry should be hidden when selecting multiple projects. Currently, it is shown as disabled because the visibleWhen expression of the menu contribution is not evaluated (only the enabledWhen expression of the handler gets evaluated correctly which disables the entry).

image

Menu contribution (in /org.eclipse.ui.ide/plugin.xml):
image

"org.eclipse.ui.ide.showInDefinition" definition (in /org.eclipse.ui.ide/plugin.xml):
image

Handler (in /org.eclipse.ui.ide/plugin.xml):
image

This PR also adds a test case documenting the changed behavior of the  "Show in" -> "System explorer" entry

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/ShowInMenu.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/ShowInMenu.java
@@ -27,6 +27,7 @@ import org.eclipse.e4.ui.internal.workbench.ContributionsAnalyzer;
 import org.eclipse.e4.ui.model.application.MApplication;
 import org.eclipse.e4.ui.model.application.commands.MCommand;
 import org.eclipse.e4.ui.model.application.commands.MParameter;
+import org.eclipse.e4.ui.model.application.ui.MExpression;
 import org.eclipse.e4.ui.model.application.ui.menu.MHandledMenuItem;
 import org.eclipse.e4.ui.model.application.ui.menu.MMenu;
 import org.eclipse.e4.ui.model.application.ui.menu.MMenuContribution;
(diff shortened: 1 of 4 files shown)


laeubi:
@r-mennig thanks for the fix, looks good to me.

merks:
Great that you added at too!  🏆

PR #3990, opened 2026-05-08
Use java.text.BreakIterator in DefaultTextDoubleClickStrategy

vogella:
Replaces `com.ibm.icu.text.BreakIterator` with `java.text.BreakIterator` in `DefaultTextDoubleClickStrategy` and drops the matching `Import-Package: com.ibm.icu.text` from the bundle manifest. The JDK `BreakIterator` exposes the same API (`getWordInstance`, `preceding`, `following`, `isBoundary`, `setText`, `DONE`) and the existing POSIX-locale workaround for `.` not being treated as a word boundary continues to work. This removes the last `com.ibm.icu` reference from `org.eclipse.jface.text`.

**Planned for 4.41**

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultTextDoubleClickStrategy.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultTextDoubleClickStrategy.java
@@ -14,11 +14,9 @@
 
 package org.eclipse.jface.text;
 
+import java.text.BreakIterator;
 import java.text.CharacterIterator;
 import java.util.Locale;
-
-import com.ibm.icu.text.BreakIterator;
-
 
 
 /**
@@ -223,7 +221,52 @@ public class DefaultTextDoubleClickStrategy implements ITextDoubleClickStrategy 
 	 * @since 3.5
 	 */
 	protected IRegion findWord(IDocument document, int offset) {
+		IRegion identifier= findIdentifierAt(document, offset);
+		if (identifier != null) {
+			return identifier;
+		}
 		return findWord(document, offset, getWordBreakIterator());
+	}
+
+	/**
+	 * If the offset lies on an ASCII identifier character ({@code [A-Za-z0-9_]}), or
+	 * just after one, returns the maximal contiguous identifier run. Otherwise
+	 * returns {@code null} so the caller falls back to the locale-aware
+	 * {@link BreakIterator}. This handles identifier-style words containing runs
+	 * of {@code '_'} (e.g. {@code foo__bar}, {@code __aaaa}) consistently across
+	 * JDK versions, since {@link BreakIterator#getWordInstance()} places word
+	 * boundaries between consecutive underscores while users expect such tokens
+	 * to be selected as a single word.
+	 */
+	private static IRegion findIdentifierAt(IDocument document, int offset) {
+		try {
+			IRegion line= document.getLineInformationOfOffset(offset);
+			int lineStart= line.getOffset();
+			int lineEnd= lineStart + line.getLength();
+			int probe;
+			if (offset < lineEnd && isIdentifierPart(document.getChar(offset))) {
+				probe= offset;
+			} else if (offset > lineStart && isIdentifierPart(document.getChar(offset - 1))) {
+				probe= offset - 1;
+			} else {
+				return null;
+			}
+			int start= probe;
+			while (start > lineStart && isIdentifierPart(document.getChar(start - 1))) {
+				start--;
+			}
+			int end= probe + 1;
+			while (end < lineEnd && isIdentifierPart(document.getChar(end))) {
+				end++;
+			}
+			return new Region(start, end - start);
+		} catch (BadLocationException e) {
+			return null;
+		}
+	}
+
+	private static boolean isIdentifierPart(char c) {
+		return c == '_' || (c < 128 && (c >= '0' && c <= '9' || c >= 'A' && c <= 'Z' || c >= 'a' && c <= 'z'));
 	}
 
 	/**
--- a/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.text
-Bundle-Version: 3.31.0.qualifier
+Bundle-Version: 3.31.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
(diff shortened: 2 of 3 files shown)


HannesWell:
I'm looking forward to have icu entirely removed.
But IIRC in the past difficulties to fully replace it have been mentioned due to different behavior. I searched for corresponding bugs/source but didn't find them. Do you have them at hand?

vogella:
I don't have a reference but the behavior is partially covered by unit tests which I extended. This was therefore not a drop in replacement.

HannesWell:
> I don't have a reference but the behavior is partially covered by unit tests which I extended. This was therefore not a drop in replacement.

I'm not deeply familiar with that part of the code, but it might be that The ICU `BreakIterator` handles non Roman languages better?
I asked Copilot about it and it provided the following answer:
```
Can com.ibm.icu.text.BreakIterator savely be replaced by java.text.BreakIterator, even for non euro-centric languages?


Not safely in general.

`com.ibm.icu.text.BreakIterator` and `java.text.BreakIterator` are related in purpose, but ICU’s implementation is usually more complete and more up to date for international text segmentation, especially for non-European scripts and edge cases like emoji, grapheme clusters, dictionary-based word breaking, and evolving Unicode rules.

A practical rule of thumb:

- **If you only handle simple Latin-script text** and basic sentence/word boundaries, replacement may work.
- **If you care about correctness across many languages**, especially Thai, Lao, Khmer, Burmese, CJK, Indic scripts, mixed-script text, or modern emoji behavior, **do not assume they are interchangeable**.
- **If your code currently depends on ICU**, it often does so for a reason.

Key differences to watch for:

1. **Unicode data freshness**
   - ICU is generally updated faster and more regularly for Unicode segmentation behavior.
   - JDK behavior depends on the Java version and may lag.

2. **Dictionary-based segmentation**
   - Languages without spaces between words, like **Thai** or **Japanese** in some contexts, are more sensitive.
   - ICU tends to perform better here.

3. **Grapheme cluster handling**
   - User-perceived characters are tricky:
     - emoji with modifiers
     - ZWJ emoji sequences
     - combining marks
     - regional-indicator flags
   - ICU is usually safer if cursor movement, deletion, highlighting, or UI text operations matter.

4. **Locale behavior**
   - Even when both APIs expose similar methods, actual boundary results can differ by locale and JDK/ICU version.

5. **API/behavior mismatch**
   - Even if code compiles after a swap, tests may fail because break positions differ.

So the honest answer is:

- **For euro-centric languages only:** maybe.
- **For non euro-centric languages:** **not reliably/safely without testing**.

What I’d recommend:
- Build a regression corpus with text in:
  - English, German, French
  - Arabic, Hebrew
  - Hindi
  - Thai
  - Chinese, Japanese, Korean
  - emoji-heavy social text
- Compare actual break positions for:
  - character boundaries
  - word boundaries
  - sentence boundaries
  - line boundaries if used
- Only replace ICU if your tests show identical or acceptable behavior for your supported locales and Java version.

If you want, I can also give you:
1. a **risk matrix by language/script**, or  
2. a **small Java test harness** to compare ICU vs JDK `BreakIterator` on sample strings.
```

Of course I'm not sure how correct that is. But if it's true, we risk breaking the existing translation e.g. for Asian languages.
Maybe some of our fellow developers from India can provide test text in Hindi?

vogella:
The Copilot dump you pasted is generic and over-broad for this specific change.
The platform already uses java.text.BreakIterator for text handling:

- org.eclipse.ui.editors → AbstractDecoratedTextEditor
- org.eclipse.ui.forms → FormUtil, TextSegment

So the JDK iterator is already the de-facto choice in the editor stack. After this PR, DefaultTextDoubleClickStrategy is the last com.ibm.icu.text.BreakIterator reference in org.eclipse.jface.text.

Happy to add a CJK/Thai sample to the test to document the behavior if that helps.

vogella:
Also note the scope here: this change only affects double-click word selection

HannesWell:
> The Copilot dump you pasted is generic and over-broad for this specific change.

Yes, of course. But my intend was to understand if it generally is a safe replacement or if one has to pay more attention.

> So the JDK iterator is already the de-facto choice in the editor stack. After this PR, DefaultTextDoubleClickStrategy is the last com.ibm.icu.text.BreakIterator reference in org.eclipse.jface.text.

Don't get me wrong, I'm glad when ICU is finally removed as dependency, but we should take care if that severely breaks existing users or not and should assert the risk and then decide if it's acceptable or not.


> Happy to add a CJK/Thai sample to the test to document the behavior if that helps.

Sure, that would help. But I don't have one at hand, do you? If not I can ask around if somebody else can provide one.

vogella:
I will use these:

こんにちは世界 Japanese   "Hello, world"
 foo 我是 bar  Chinese  我是 = "I am"
 foo ไทย bar  Thai  ไทย = "Thai"

vogella:
The scope here is narrow: this only uses getWordInstance() for double-click word selection, and the new identifier fast-path means the locale-aware iterator is only a fallback for non-ASCII text.

I added CJK/Thai tests to make it concrete.

merks:
It's nice to see the added test cases!

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultTextDoubleClickStrategy.java
```diff
@@ -223,9 +221,54 @@ protected IRegion findExtendedDoubleClickSelection(IDocument document, int offse
...
+			}
+			return new Region(start, end - start);
+		} catch (BadLocationException e) {
+			return null;
+		}
+	}
+
+	private static boolean isIdentifierPart(char c) {
+		return c == '_' || (c < 128 && (c >= '0' && c <= '9' || c >= 'A' && c <= 'Z' || c >= 'a' && c <= 'z'));
+	}
```
HannesWell, line 270: Is this related to `Character.isJavaIdentifierPart(char)` ? Or would it make sense to reuse it, although it's not identical atm.

HannesWell (commented):
> The scope here is narrow: this only uses getWordInstance() for double-click word selection, and the new identifier fast-path means the locale-aware iterator is only a fallback for non-ASCII text.
> 
> I added CJK/Thai tests to make it concrete.

Thanks for the update. Then this is probably fine or at least save enough from my POV.

bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/DefaultTextDoubleClickStrategy.java
vogella: It's conceptually related (both answer "is this an identifier character?"), but Character.isJavaIdentifierPart should not be reused here. It must stay ASCII-only by design. This method is the guard for a fast-path that only fires for [A-Za-z0-9_]. For anything else it returns null so the caller falls back to BreakIterator.

vogella:
Thanks for the feedback @HannesWell and @merks, I assume that your feedback is adressed and plan to merge later today or tomorrow morning.

Issue #3591, opened 2025-12-16
Error while running the Windows Defender exclusion check


azoitl:
We got the feedback for 4diac IDE that on sum windows machines the _Windows Defender exclusion check_ throws an exception (https://github.com/eclipse-4diac/4diac-ide/issues/1949). To check if it is with 4diac IDE or with the Eclipse Platform we downloaded a fresh Eclipse IDE 2025-12 and we could reproduce the problem with that exception (see below). 

In the mean time we see this happening with more users. Some of them have a different anti virus software, some run windows defender. Unfortunately we have not yet an understanding what causes the problem. You can see more in this matrix thread: https://chat.eclipse.org/#/room/!yVjgkwgBPbjMVDziCd:matrix.eclipse.org/$1a4JH_uhQAX7h5dcovdeq9vfNXVFWgrOT3oVl5_wP0g?via=matrix.eclipse.org&via=matrix.org

However I looked into the place that is throwing the exception and my question would be should we fail silent in cases where the windows defender check is returning unexpected values? 

If yes, I'm happy to provide a PR for that.

## Here is some relevant log output

From `<workspace>/.metadata/.log`
```
org.eclipse.core.runtime.CoreException: Failed to retrieve Microsoft Defender status.
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.isWindowsDefenderActive(WindowsDefenderConfigurator.java:332)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.runExclusionCheck(WindowsDefenderConfigurator.java:159)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.runCheckEnforced(WindowsDefenderConfigurator.java:144)
	at org.eclipse.ui.internal.dialogs.StartupPreferencePage.lambda$0(StartupPreferencePage.java:123)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:83)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:290)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4364)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1217)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4162)
[...]

PR #3994, opened 2026-05-10
Fix Windows Defender status retrieval under Java-25 or later

HannesWell:
In recent Java versions `Process.isAlive()` seems to report the termination of a process a little bit later then before, relative to the time when the process output reader is closed, which leads to false errors about process time-out.

Fixes errors like:
```
!MESSAGE Failed to obtain 'WinDefend' service state
!STACK 0
java.io.IOException: Process timed-out and it was attempted to forcefully terminate it
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.runProcess(WindowsDefenderConfigurator.java:395)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.runPowershell(WindowsDefenderConfigurator.java:369)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.isWindowsDefenderServiceRunning(WindowsDefenderConfigurator.java:302)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.runExclusionCheck(WindowsDefenderConfigurator.java:159)
	at org.eclipse.ui.internal.WindowsDefenderConfigurator.lambda$0(WindowsDefenderConfigurator.java:106)
```

I cannot yet tell why, because I haven’t investigated the details yet.

CC @merks and @azoitl 
Does this also help for 
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/3591

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java
@@ -31,6 +31,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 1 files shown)


merks (approved):
It's not so easy to test without a build,  I see zero risk in this producing worse behavior given `If the process has already terminated then this method returns immediately with the value {@code true}. `  So it should exactly deal with the case of it taking a while for isAlive to return false, and in the worst case taking 5 second longer to pop up a failure message...

merks:
@HannesWell 

If we merge this today, I can more easily test tomorrow's I-build...

HannesWell:
> It's not so easy to test without a build, 

The check is performed even in a secondary Eclipse, if you remove the `inDevelopmentMode()` test at

https://github.com/eclipse-platform/eclipse.platform.ui/blob/195df72dfebf1fd50469286e041c2c77fcb95303/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/WindowsDefenderConfigurator.java#L114

> I see zero risk in this producing worse behavior given `If the process has already terminated then this method returns immediately with the value {@code true}. ` So it should exactly deal with the case of it taking a while for isAlive to return false, and in the worst case taking 5 second longer to pop up a failure message...

Absolutely agree. And since it runs asynchronously in a job, its almost free.
So yes lets submit this now so that more I-build users can test this tomorrow.

azoitl:
Thx for looking into this @HannesWell.

merks:
FYI, it's working well for me when I tested it just now.

PR #4002, opened 2026-05-14
Allow to ignore explicitly defined disabled icons

HeikoKlare:
This change allows to ignore explicitly defined disabled icons (as SWT provides a [proper way to do it on-the-fly since 2025-06](https://eclipse.dev/eclipse/markdown/?f=news/4.36/platform.md#improved-disabled-icons-generation)) and enabled that behavior by default.

This is an extract of:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3760

In that PR, no objections have been raised on always using the on-the-fly generation of disabled icons. I thus intend to merge this at the beginning of 4.41 development in case no objections are raised until then.

@HannesWell we had discussed the extraction of this rather non-controversial change offline. Maybe you want to have a quick a look.

### Description

The algorithm for generating disabled icons on-the-fly has recently been enhanced. Most explicit disabled icons that have been embedded into bundles conform to what now can be generated on-the-fly. In addition, the algorithm is interchangeable, allowing custom stylings of disabled icons. However, when there are still bundles, such as extensions out of our control, that still explicitly define disabled icons, exchanging the algorithm for disabled icons will lead to inconsistent appearance as only the on-the-fly generated icons will adhere to that.

This change allows to ignore explicitly defined disabled icons, such that even if some bundles defined disabled icons, they will be ignored and on-the-fly generated disabled icons according to the selected algorithm will be used instead. An according preference that can be configured via the appearance tab is added.

### How to test

The enhancement is configurable via preference, thus can be tested by enabling/disabling it:
image

As an example, this is what the main toolbar looks like with different configurations:

#### Existing
image

#### No pregenerated disabled icons
image

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/action/ActionContributionItem.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/action/ActionContributionItem.java
@@ -57,6 +57,12 @@ import org.eclipse.swt.widgets.Widget;
 public class ActionContributionItem extends ContributionItem {
 
 	/**
+	 * System property for specifying whether custom disabled icons for actions
+	 * shall be used or whether all disabled icons shall be generated on-the-fly.
+	 */
+	private static final String SYSTEM_PROPERTY_USE_DISABLED_ICONS = "org.eclipse.jface.action.useDisabledIcons"; //$NON-NLS-1$
+
+	/**
 	 * Mode bit: Show text on tool items or buttons, even if an image is
 	 * present. If this mode bit is not set, text is only shown on tool items if
 	 * there is no image present.
@@ -94,6 +100,32 @@ public class ActionContributionItem extends ContributionItem {
 	 */
 	public static void setUseColorIconsInToolbars(boolean useColorIcons) {
 		USE_COLOR_ICONS = useColorIcons;
+	}
+
+	private static boolean USE_DISABLED_ICONS = Boolean.getBoolean(SYSTEM_PROPERTY_USE_DISABLED_ICONS);
+
+	/**
+	 * Returns whether disabled icons assigned to actions shall be used or whether
+	 * all such disabled icons shall be generated on-the-fly.
+	 *
+	 * @return <code>true</code> if disabled icons assigned to actions shall be
+	 *         used, <code>false</code> otherwise
+	 * @since 3.40
+	 */
+	public static boolean getUseDisabledIcons() {
+		return USE_DISABLED_ICONS;
+	}
+
+	/**
+	 * Sets whether disabled icons assigned to actions shall be used or whether all
+	 * such disabled icons shall be generated on-the-fly.
+	 *
+	 * @param useDisabledIcons <code>true</code> if disabled icons set to actions
+	 *                         shall be used, <code>false</code> otherwise
+	 * @noreference This method is not intended to be referenced by clients.
+	 */
+	public static void setUseDisabledIcons(boolean useDisabledIcons) {
+		USE_DISABLED_ICONS = useDisabledIcons;
 	}
 
 	/**
@@ -988,7 +1020,10 @@ public class ActionContributionItem extends ContributionItem {
 		if (widget instanceof ToolItem) {
 			ImageDescriptor image = action.getImageDescriptor();
 			ImageDescriptor hoverImage = action.getHoverImageDescriptor();
-			ImageDescriptor disabledImage = action.getDisabledImageDescriptor();
+			ImageDescriptor disabledImage = null;
+			if (getUseDisabledIcons()) {
+				disabledImage = action.getDisabledImageDescriptor();
+			}
 			// Make sure there is a valid image in case images are forced.
 			if (image == null && forceImage) {
 				image = ImageDescriptor.getMissingImageDescriptor();
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/AbstractContributionItem.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/AbstractContributionItem.java
@@ -38,6 +38,7 @@ import org.eclipse.e4.ui.workbench.IResourceUtilities;
 import org.eclipse.e4.ui.workbench.modeling.EModelService;
 import org.eclipse.e4.ui.workbench.swt.util.ISWTResourceUtilities;
 import org.eclipse.emf.common.util.URI;
+import org.eclipse.jface.action.ActionContributionItem;
 import org.eclipse.jface.action.ContributionItem;
 import org.eclipse.jface.action.IContributionManager;
 import org.eclipse.jface.action.IMenuCreator;
@@ -211,6 +212,9 @@ public abstract class AbstractContributionItem extends ContributionItem {
 	}
 
 	private String getDisabledIconURI(MItem toolItem) {
+		if (!ActionContributionItem.getUseDisabledIcons()) {
+			return ""; //$NON-NLS-1$
+		}
 		Object obj = toolItem.getTransientData().get(IPresentationEngine.DISABLED_ICON_IMAGE_KEY);
 		return obj instanceof String s ? s : ""; //$NON-NLS-1$
 	}
--- a/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.jface/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface;singleton:=true
-Bundle-Version: 3.39.100.qualifier
+Bundle-Version: 3.40.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.jface,
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/menus/CommandContributionItem.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/menus/CommandContributionItem.java
@@ -28,6 +28,7 @@ import org.eclipse.core.commands.ParameterizedCommand;
 import org.eclipse.core.commands.common.NotDefinedException;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.jface.action.ActionContributionItem;
 import org.eclipse.jface.action.ContributionItem;
 import org.eclipse.jface.action.IContributionManager;
 import org.eclipse.jface.action.IMenuListener2;
@@ -892,7 +893,9 @@ public class CommandContributionItem extends ContributionItem {
 			localResourceManager = m;
 		} else if (widget instanceof ToolItem item) {
 			LocalResourceManager m = new LocalResourceManager(JFaceResources.getResources());
-			item.setDisabledImage(disabledIcon == null ? null : m.create(disabledIcon));
+			if (ActionContributionItem.getUseDisabledIcons()) {
+				item.setDisabledImage(disabledIcon == null ? null : m.create(disabledIcon));
+			}
 			item.setHotImage(hoverIcon == null ? null : m.create(hoverIcon));
 			item.setImage(icon == null ? null : m.create(icon));
 			disposeOldImages();
--- a/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/action/ToolBarManagerTest.java
+++ b/tests/org.eclipse.jface.tests/src/org/eclipse/jface/tests/action/ToolBarManagerTest.java
@@ -141,8 +141,10 @@ public class ToolBarManagerTest {
 	@Test
 	public void testDefaultImageIsGray() {
 		boolean oldState = ActionContributionItem.getUseColorIconsInToolbars();
+		boolean oldStateUseDisabled = ActionContributionItem.getUseDisabledIcons();
 		try {
 			ActionContributionItem.setUseColorIconsInToolbars(false);
+			ActionContributionItem.setUseDisabledIcons(true);
 			ToolBarManager manager = new ToolBarManager();
 			Action action = new Action("Button with Hover") {
 			};
@@ -165,14 +167,17 @@ public class ToolBarManagerTest {
 			assertImageEqualsDescriptor(ImageDescriptor.createWithFlags(descriptor, SWT.IMAGE_GRAY), item.getImage());
 		} finally {
 			ActionContributionItem.setUseColorIconsInToolbars(oldState);
+			ActionContributionItem.setUseDisabledIcons(oldStateUseDisabled);
 		}
 	}
 
 	@Test
 	public void testActionImagesAreSet() {
 		boolean oldState = ActionContributionItem.getUseColorIconsInToolbars();
+		boolean oldStateUseDisabled = ActionContributionItem.getUseDisabledIcons();
 		try {
 			ActionContributionItem.setUseColorIconsInToolbars(true);
+			ActionContributionItem.setUseDisabledIcons(true);
 			ToolBarManager manager = new ToolBarManager();
 			Action action = new Action("Button with Hover") {
 			};
@@ -195,6 +200,7 @@ public class ToolBarManagerTest {
 			assertImageEqualsDescriptor(disabledDescriptor, item.getDisabledImage());
 		} finally {
 			ActionContributionItem.setUseColorIconsInToolbars(oldState);
+			ActionContributionItem.setUseDisabledIcons(oldStateUseDisabled);
 		}
 	}
 


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
```diff
@@ -436,6 +436,9 @@ ViewsPreference_hideIconsForViewTabs = Hide icons
 ViewsPreference_viewTabs_icons_and_titles_label = Tab icons and titles in view areas:
 ViewsPreference_viewTabs_dirty_indicator_label = Indicator for unsaved changes in view and editor tabs:
 ViewsPreference_showDirtyIndicatorForTabs = Indicate unsaved changes by overlaying the close button
+ViewsPreference_disabledIcons_description = Icons for disabled actions:
+ViewsPreference_ignoreDisabledIcons = Ignore pre-generated disabled icons
+ViewsPreference_ignoreDisabledIcons_tooltip = When enabled ignores pre-generated disabled icons in favor of generating consistently styled disabled icons on the fly
```
HannesWell: What about labling these icons as `explicit` disabled icons or `dedicated` disabled icons?
```suggestion
ViewsPreference_ignoreDisabledIcons = Ignore explicit disabled icons
ViewsPreference_ignoreDisabledIcons_tooltip = When enabled ignores explicit disabled icons in favor of generating consistently styled disabled icons on the fly
```
But I'm not sure if `explicit` could be read with other meaning.

HannesWell (approved):
Thanks for this. I think this totally makes sense and helps to fade out explicit/dedicated disabled icons.

I just have one suggestions for their label/description.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
nettozahler: Regarding the suggestion: maybe we should clarify with a native English speaker if it must be named "explicitly disabled".

laeubi:
For me a radio button style option would be more appropriate than a checkbox ... the current terms seem to technical for anyone to understand that is not familiar with the technical details.

Something similar to:

`Generate disabled images (*)  always   ( ) only if not provided  ( ) never`

(where the last option would e.g. use the "native" disabled image appearance if we want to offer such option)

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/messages.properties
HeikoKlare: > But I'm not sure if `explicit` could be read with other meaning.

That's why I tried to avoid the term in user-facing descriptions. I am sure I would not have understood the option when using that word if I was not involved in the development.

> maybe we should clarify with a native English speaker if it must be named "explicitly disabled"

I think that's a good example, why we should avoid the term "explicit", because "explicitly disabled" would be incorrect in terms of semantics.

I have no really good proposal for an alternative either. Just though about something like "manually generated".

HeikoKlare:
> For me a radio button style option would be more appropriate than a checkbox ... the current terms seem to technical for anyone to understand that is not familiar with the technical details.

I agree that the description is currently too technical. That's something we need to solve by proper naming. A radio button alone will not solve that, in particular because there are only the two options "always generate" or "consider pre-generated/explicit/manually generated/...". "never" is not a possible option without starting a new feature for supporting that option.

Maybe just taking your proposed text for the radio buttons would be fine as something like: \
**Option:** Always generate disabled icons on the fly \
**Description:** When enabled, ignores any provided resources for disabled icons in favor of generating consistently styled disabled out of their enabled version icons on the fly

laeubi:
I don't thing "generating" is a good therm at all, but the whole preference page lacks a bit the user perspective - I must confess most of this is already unclear what it really does and this pseudo headlines are really horrible.

So first think I would ask is does it really only affect *actions* (or is it more buttons that trigger an action?) - and only toolbar or also menus? Second I think, that it is more that we *derive* a disabled icon from the original one.
Then the question would be what is a *provided* one )same for the current pre-generated - generated from what?) - and why would it be inconsistent?

So in the end I would possibly still be better to have a radio option to make clear what are the choices here. One might also ask if this should really be a user setable option (e.g. how likely is it a user will ever change this?)

merks:
Maybe even more fundamental question.  Should the user decide this at all?  I imagine product owners will be better to judge on this front...

HeikoKlare:
> One might also ask if this should really be a user setable option (e.g. how likely is it a user will ever change this?)

> Maybe even more fundamental question. Should the user decide this at all? I imagine product owners will be better to judge on this front...

Those are absolutely valid points. In my opinion, the goal should be that the automated mechanism for generating disabled images on the fly is so good / configurable that no one not only needs to but not even wants to provide and maintain custom-generated disabled versions of icons. I see two potential reasons to consider regarind configurability:
1. There might be products that have specific versions of disabled icons. Or there may be some disabled icons that should look different from what the auto-generation would produce (not judging how reasonable that is from a UX POV), at least for now and it make take time to adapt to the auto-generation.
2. There might be existing (non-maintained) RCP extensions that contain icons which become unusable in their auto-generated disabled version. In that case, a user may want to be able to turn the usage of custom disabled icons on to make the extension usable.

While in my opinion (1) needs to be supported, I am not sure if (2) will really be a relevant use case. And with that insights, I agree that providing a configuration via preference may not be the best way of configuration here. Instead, we may simply properly share the information about this change via release N&N and provide a fallback via system property. But then we should be consistent and also deprecate all API for defining disabled icons (like was done for hover icons).

HeikoKlare:
I have now simplified the change to not introduce any preference but only a system property to re-enable usage of disabled icons. I hope that resolves the raised concerns.

bundles/org.eclipse.jface/src/org/eclipse/jface/action/ActionContributionItem.java
```diff
@@ -56,6 +56,12 @@
 @NoExtend
 public class ActionContributionItem extends ContributionItem {
 
+	/**
+	 * System property for specifying whether custom disabled icons for actions
+	 * shall be used or whether all disabled icons shall be generated on-the-fly.
+	 */
+	private static final String SYSTEM_PROPERTY_USE_DISABLED_ICONS = "org.eclipse.jface.action.useDisabledIcons"; //$NON-NLS-1$
+
```
HannesWell, line 64: Couldn't this be inlined? I only see one reference and the comment can also be added to the boolean field?

bundles/org.eclipse.jface/src/org/eclipse/jface/action/ActionContributionItem.java
```diff
@@ -96,6 +102,32 @@ public static void setUseColorIconsInToolbars(boolean useColorIcons) {
...
+	 * such disabled icons shall be generated on-the-fly.
+	 *
+	 * @param useDisabledIcons <code>true</code> if disabled icons set to actions
+	 *                         shall be used, <code>false</code> otherwise
+	 * @since 3.40
+	 */
+	public static void setUseDisabledIcons(boolean useDisabledIcons) {
+		USE_DISABLED_ICONS = useDisabledIcons;
+	}
+
```
HannesWell, line 130: Is this just for testing or should this also be used by down-stream consumers?

HannesWell (approved):
> Maybe even more fundamental question. Should the user decide this at all? I imagine product owners will be better to judge on this front...

That's indeed a good question and I share your opinion respectively what Heiko answered to it.
With that I'm again fine with this, just have two minor suggestions/questions below.

bundles/org.eclipse.jface/src/org/eclipse/jface/action/ActionContributionItem.java
HeikoKlare: Currently it's only used for testing, but it may also validly be used to programmatically configure the behavior instead of using the system property. I would be fine with only making it available for tests but properly doing it would not be that easy with our test setup. We could mark it as `@noreference` though. What do you think?

bundles/org.eclipse.jface/src/org/eclipse/jface/action/ActionContributionItem.java
HeikoKlare: We could inline it of course. I just think it's beneficial to have the property definition and its usage(s) separated. That's also conforms to the common pattern in other classes where system properties are often defined at the top of the class. Do you see any benefit in inlining it?

HannesWell: Not a significant benefit except for slightly more compact code.
But using the same style like in other classes is also good. So I'm fine with it either way.

bundles/org.eclipse.jface/src/org/eclipse/jface/action/ActionContributionItem.java
HannesWell: > Currently it's only used for testing, but it may also validly be used to programmatically configure the behavior instead of using the system property. 

If we definitively want to support that, then I think it makes sense to have this public. But if we don't or are not sure, then I suggest to not make it an 'API' and only elevate it when desired/necessary. I think it just should be a clear decision, but I'm not for one or the other.

> doing it would not be that easy with our test setup. We could mark it as `@noreference` though.

Yes, in that case it we could do that or make it (package) private and just call it reflectively or set the field reflectively. But that's of course not very elegant.

HeikoKlare: I've updated the PR with the added `@noreference` annotation in Javadoc. Reflective access would be rather ugly and doing it properly (making the tests actual unit tests via test fragment or test integrated into the production bundle or putting the setter into some internal package with the tests defined as x-friends) is, in my opinion, not worth the effort. Then you could rather think about if it's worth to keep the part of the test case for disabled images at all.

tomaswolf:
When I as a developer set a disabled icon, I have a reason for doing so, and I want that icon to be used. Period. I don't see the need to remove this functionality at all.

HeikoKlare:
> When I as a developer set a disabled icon, I have a reason for doing so, and I want that icon to be used. Period. I don't see the need to remove this functionality at all.

The main reasons for doing that so far was that disabled icons could not be properly generated on the fly. That has changed with recent improvements to SWT, as now on-the-fly generation is sufficient for generating consistent results. Do you have uses cases where you explicitly want to provide disabled icons that derivate from what the on-the-fly generation delivers?

akurtakov:
The only potential problem I see is if someone deliberately sets the widget to some entirely different image - it's all hypothetical (as I don't have such case) but maybe it's worth thinking/searching of such cases.

HeikoKlare:
> The only potential problem I see is if someone deliberately sets the widget to some entirely different image - it's all hypothetical (as I don't have such case) but maybe it's worth thinking/searching of such cases.

That's correct. I have listed the types of cases I see in this comment already: https://github.com/eclipse-platform/eclipse.platform.ui/pull/4002#issuecomment-4462774653

Also note that there is no plan to remove the functionality overall, it's just that by default explicitly provided disabled icons will be ignored. You can still use the provided system property to keep using such icons. Without this, we will never be able to change (or rather allow to change) the overall appearance of disabled icons because there may be extensions that still deploy such icons which will then not fit to the other icons anymore.

In addition, providing an entirely different image as disabled image is, to put it mildly, a questionable idea in terms of UX.

tomaswolf:
> Do you have uses cases where you explicitly want to provide disabled icons that derivate from what the on-the-fly generation delivers?

Yes. Consider command contributions to a toolbar in a view. If the handler is set to be active only when the view is active (to avoid handler conflicts), you may get the strange situation that the button shows disabled, but when clicked performs the command anyway since on click the view is first activated and then the button receives the click event.

Case is question is the refresh button in the Git Repositories view. It has a disabledIcon that is the same as the enabled icon (i.e., _not_ a grayscale version). However, I just noticed that this doesn't work quite as expected; somehow the platform sometimes still shows only a faded version of that icon.

We hit the same case with the toolbar top right in the Commit Editor in EGit (the one that opens when you select a commit in the history view and then choose "Open in Commit Viewer" from the context menu). See for instance [commit 5ad9f043 in EGit](https://github.com/eclipse-egit/egit/commit/5ad9f043); the commit message explains the problem. (Though that commit uses a different work-around. But that work-around doesn't work if you populate the toolbar via contributions from plugin.xml.)

As a plugin developer I have no control over what system properties might be set when the product using the plugin is started. So system properties are not really a solution.

HeikoKlare:
Thank you for those pointers. That's very interesting input.

> Consider command contributions to a toolbar in a view. If the handler is set to be active only when the view is active (to avoid handler conflicts), you may get the strange situation that the button shows disabled, but when clicked performs the command anyway since on click the view is first activated and then the button receives the click event.

That's an interesting case. But it actually rather sounds like a framework bug or a usage mistake. On the framework side, from a UX POV, the click on a disabled button should not work. On the usage side (at the example of the Refresh button in the Git repositories view): isn't the platform's refresh action retargettable, so shouldn't the general refresh action be retargeted when the repositories view is active rather defining another handler for an EGit-specific refresh? I have to admit that I am not completely sure about these mechanisms, but since refresh actions are available at other places as well (e.g., in the "Call Hierarchy View"), I wonder if EGit couldn't solve that in the same way. Is it because the action is not specific to the selected element but refreshes the view (I have to admit that I do not even understand what it is supposed to do)? Then the behavior is different from other places where F5 does a refresh of the currently selected element, so a behavior that may be questioned from UX POV. 

So in my opinion, solving these issues by overwriting the disabled icon with an enabled one is a quite hacky workaround and not a proper solution to the issue. Also note that the UX of this approach is quite bad, as the button does not give hover feedback (as it is disabled) even though it looks enabled. I remember that I was once confused why some button was shown enabled but provided no feedback when hovering (maybe it was the Refresh button of the Git repositories view) but forgot to follow-up on that.

So I think we have some other issues to adress here, but I still do not see a reasonable use case for "overwriting" disabled icons.

tomaswolf:
> But it actually rather sounds like a framework bug or a usage mistake. On the framework side, from a UX POV, the click on a disabled button should not work.

I don't agree for the view toolbar. Maybe you're right for the toolbar _inside_ the commit editor. But a button (or rather, ToolItem) in a view toolbar should not require two clicks to work (first click to activate the view, second click to execute the command). OTOH perhaps it's just not doable with a command contribution.

HeikoKlare:
> I don't agree for the view toolbar. Maybe you're right for the toolbar inside the commit editor. But a button (or rather, ToolItem) in a view toolbar should not require two clicks to work (first click to activate the view, second click to execute the command). OTOH perhaps it's just not doable with a command contribution.

But then the button should actually be enabled and not just partially mimic a enabled look while actually being disabled and only being enabled when clicking. Please consider my notes on the degraded UX with that approach as well. I.e., the framework would have to show that button as enabled even though the used handler is disabled when the view is not active.

So what is the proposal to proceed here? I don't expect that anyone disagrees that the cases mentioned so far are misusages of custom disabled icons, so I am not sure if that should block further progress. At least there was no disgreement to my last post on that.
Please note that I already kept the original proposal open for three months now (https://github.com/eclipse-platform/eclipse.platform.ui/pull/3760) and this extracted one for two weeks, so that we are able to merge it right at the beginning of a release cycle and then have as much implicit (also downstream) testing of the change as possible. So I would be in favor of a timely conclusion to avoid that we need to hold this back for another 3 months.

tomaswolf:
It will break the use case in the EGit "Git Repositories" view.

I don't understand why one would want to disable this feature at all. If the current disabled icons that just use a grayscale image are not good enough, one could simply remove these icons and get the auto-generated behavior. But if some plugin does set a disabled icon, why have extra code to ignore that? As I wrote above, the plugin developer had a reason for setting the icon, and wants that icon to be shown.

Incidentally I notice that #3760 got exactly one comment, and that one goes in the same direction.

A system property is beyond the control of plugin developers and with the default of `false` the feature is essentially switched off. A plugin developer cannot rely on it any more. One could equally well pull out all the code related to disabled icons and deprecate or remove &lt;disabledIcon>.

BTW, I wrote:
>However, I just noticed that this doesn't work quite as expected; somehow the platform sometimes still shows only a faded version of that icon.

On Mac. On Windows it works as expected.

laeubi:
I agree with @HeikoKlare that showing a disabled button as enabled because it actually is not disabled sounds like a bug that must be fixed (independently). I also agree with @akurtakov and @iloveeclipse case that one *might* define a disabled icon that looks completely different but also hardly can find a good example where it would be useful, in contrast for example a toogle-button (where it as far as I know is not supported out of the box).

The most limiting factor I see is more that this only ever will work for action contributions, so there might because where people construct on toolbars that then look differently, so consequently this more has to be placed at the JFace level or even SWT level to be really consistent.

akurtakov:
Tbh, I don't quite understand the "showing a disabled button as enabled because it actually is not disabled" so I would appreciate a pure SWT snippet showing what is meant by that.

laeubi:
> Tbh, I don't quite understand the "showing a disabled button as enabled because it actually is not disabled" so I would appreciate a pure SWT snippet showing what is meant by that.

As this is a jface feature (or even platform.ui) I don't think that is really possible, SWT does has no notation of "disabled" image for buttons. A button can be enabled/disabled but this has no effect on the icon as otherwise one can not set an own icon anyways.

merks:
Of course we've digressed from the topic of the title, But to me also it seems that some JFace problem related to the proper initial state of a tool button is a separate problem that should be addressed separately...

HeikoKlare:
> I don't understand why one would want to disable this feature at all. 

To clarify this again: this contributes to the continuous efforts of enabling enhanced look and feel for Eclipse products. Without this change you are unable to properly exchange the algorithm for generating disabled icons (and potentially even exchanging icon themes https://github.com/eclipse-platform/eclipse.platform.ui/discussions/2870).
Disabled icons should be something that is automatically derived from the enabled version. That is what all OSes and their UI libraries also do. Since the algorithms for generating the disabled icons can be different depending on the context (you can already now configure the algorithm used by SWT), it is problematic to provide explicit disabled icons, as will can, by design, only fit to the icons for one specific algorithm.

>  As I wrote above, the plugin developer had a reason for setting the icon, and wants that icon to be shown.

As explained above, the common reason so far was that it was not possible to generate disabled icons on-the-fly, which is why _every_ icon had an explicit disabled verson of itself. The developer would have been totally happy if that icon was generated on-the-fly instead of investing manual effort to generate and set it (which is now possible).
My current hypothesis is: every disabled icon that has not been created for that reason is there to either work around some actual framework bug or because of some questionable UX decision. I have still not seen any example that disproves this hypothesis.

So to repeat my previous question: how should we proceed? Are there any concerns regarding fixing the actual underlying root causes for the reported issue(s) and proceeding with this proposal?
I have to admit that I do not really understand why someone would prefer to keep an existing workaround for a bug in favor of resolving that bug and moving forward with enhancements.

tomaswolf:
I don't see this as a work-around. But if the consensus here is that disabledIcons are not to be used, then implement that: rip out/deprecate/make no-ops anything related to disabled icons (except the auto-generation). Just drop the feature instead of adding code to make it "configurable".

PR #4004, opened 2026-05-15
Show theme-default checkbox only when workbench theme actually changed

### LINKED ISSUE #4003: Restart dialog when changing "Appearance" settings unexpectedly mentions theme changes
The restart dialog when changing something in the Appearance page of the preferences always seems to mention something about default themes, even if no theme-related changes are made.

This is likely a regression from:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3854

@vogella can you please have a look?

## Steps to reproduce

On Windows:
- Start an Eclipse product
- Open the "Appearance" page in the preferences
- (De-)select "Use monitor-specific UI scaling"

I expected: A dialog asking for restart of the changes to take effect

But got: A dialog asking about default themes, even though I did not change anything about theming

Image

* OS & version: Windows 11 Build 22631
* Eclipse IDE/Platform version (as shown in *Help > About*): I20260513-0341

## Community

vogella:
Fixes #4003. The restart dialog on the Appearance preferences page surfaced a "Use as default theme" checkbox whenever theming was enabled, even when the user only changed an unrelated setting like monitor-specific UI scaling. The checkbox is now gated on the workbench theme having actually changed, which also prevents the same misleading offer when only the colors-and-fonts theme was changed.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/ViewsPreferencePage.java
@@ -508,10 +508,11 @@ public class ViewsPreferencePage extends PreferencePage implements IWorkbenchPre
 		boolean showRestartDialog = false;
 		String restartDialogTitle = null;
 		String restartDialogMessage = null;
+		boolean themeChanged = false;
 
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
Works fine. Tested different kinds of changes requiring restart, including the monitor-specific scaling setting. Does not contain the theme-related part anymore:
Image

Thank you for the immediate fix, @vogella!

PR #4005, opened 2026-05-15
Set proper default for monitor-specific scaling preference

HeikoKlare:
When restoring the default values for the "Appearance" preference page, the Windows-specific preference for enabling/disabling monitor-specific scaling is not updated. The default value for this preference is currently defined at every place accessing the preference redundantly.

With this change, a proper default value in the default preference scope is defined for the monitor-specific scaling preference. This value is used by all consumers of the preference via a ScopedPreferenceStore. The defaults restoration of the "Appearance" preference page is extended with a restore of the monitor-specific scaling preference.

### How to test

Open the "Apperance" page of the preference dialog and press "Restore Defaults". Without this change, the "Use monitor-specific UI scaling" option will not change. With this change, the default value (enabled) will be applied.

image


fedejeanne (approved):
Works as expected.

## Hint for testing
When testing this across restarts and using a launch configuration (i.e. when launching it from the IDE), make sure that you do **not** clear the configuration area before launching, otherwise the value is always reset to `true` (selected) upon restarts:

Image

PR #4008, opened 2026-05-15
Migrate OpenProjectExplorerFolderTest to JUnit 5

vogella:
Migrate this performance test from JUnit 4 to JUnit 5. The test now extends `PerformanceTestCaseJunit5` (the JUnit 5 counterpart of the dropped `PerformanceTestCaseJunit4`), which manages the `PerformanceMeter` via `@BeforeEach`/`@AfterEach`. Reusing the shared base class keeps the meter handling consistent with all other performance tests instead of reimplementing it locally.


HeikoKlare (commented):
Is there a specific reason to not migrate to `PerformanceTestCaseJunit5` and reimplement the contents of that class? In that case we shouldn't do that decision locally but align all implementations of `PerformanceTestCaseJunit5` accordingly.

vogella:
> PerformanceTestCaseJunit5

Looked at this class (again), it makes sense to use it. I will update PR

vogella:
The class has been updated to use PerformanceTestCaseJunit5

PR #4014, opened 2026-05-18
Remove unnecessary "Apply and Close" button from Properties dialog

elsazac:
The Properties dialog opened from `Installation Details → Properties` does not have any option to edit or make any modifications to the properties. Despite this it still displays both `Cancel` and `Apply and Close` buttons.

Since no changes can be made in this dialog, the `Apply and Close` button is basically redundant here and performs the same action as `Cancel`. This could be confusing to the users hence this commit removes the unnecessary `Apply and Close` button and retain only the `cancel` option.

Fixes :https://github.com/eclipse-equinox/p2/issues/1049

Before: 
Screenshot 2026-05-18 at 4 53 37 PM

After: 
Screenshot 2026-05-18 at 4 54 13 PM

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/PropertyDialog.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/PropertyDialog.java
@@ -17,6 +17,7 @@ package org.eclipse.ui.internal.dialogs;
 
 import java.util.Iterator;
 import org.eclipse.core.runtime.Adapters;
+import org.eclipse.jface.dialogs.IDialogConstants;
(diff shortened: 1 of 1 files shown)


BeckerWdf:
> After:
> Screenshot 2026-05-18 at 4 54 13 PM

Shown "Close" instead of "Cancel" would even be better from my point of view.

elsazac:
This is the updated view with close option :

image

elsazac:
build looks good; merging this now;

iloveeclipse:
This introduces regression elsewhere! 

Right click any project, see a property page for the project that now has no "Apply and Close" button, only separated "Apply" and "Close".  
Before:

image

After:

image

Please revert the change or provide a fix that doesn't affect other components.

elsazac:
I will revert this change for now;

Issue #4009, opened 2026-05-16
QuickAccessDialogTest.testPreviousChoicesAvailableForExtension fails


trancexpress:
See:

* https://download.eclipse.org/eclipse/downloads/drops4/I20260514-1800/testresults/html/org.eclipse.ui.tests_ep440I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html
* https://download.eclipse.org/eclipse/downloads/drops4/I20260515-1811/testresults/html/org.eclipse.ui.tests_ep440I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html
* https://download.eclipse.org/eclipse/downloads/drops4/I20260516-1800/testresults/html/org.eclipse.ui.tests_ep440I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html

All are the same fail:

```
expected: <true> but was: <false>

org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
at org.eclipse.ui.tests.quickaccess.QuickAccessDialogTest.testPreviousChoicesAvailableForExtension(QuickAccessDialogTest.java:302)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596) 
```

In all cases the test failed after ~8 seconds, i.e. the timeout occurred here:

```
		assertTrue(DisplayHelper.waitForCondition(text.getDisplay(), TIMEOUT,
				() -> dialogContains(dialog, TestQuickAccessComputer.TEST_QUICK_ACCESS_PROPOSAL_LABEL)));
```

The build before the last 2 fails has no fail:

[...]

PR #4017, opened 2026-05-19
Revert "Migrate QuickAccessDialogTest to JUnit 5"

vogella:
Reverts eclipse-platform/eclipse.platform.ui#3911

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/quickaccess/QuickAccessDialogTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/quickaccess/QuickAccessDialogTest.java
@@ -17,9 +17,9 @@ package org.eclipse.ui.tests.quickaccess;
 
 import static org.eclipse.ui.tests.harness.util.UITestUtil.openTestWindow;
 import static org.eclipse.ui.tests.harness.util.UITestUtil.processEventsUntil;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
 
 import java.io.File;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


akurtakov:
Please don't merge this one. I don't see any meaningful change in the assertions and migration to JUnit 5 is smth we have to complete.

vogella:
> Please don't merge this one. I don't see any meaningful change in the assertions and migration to JUnit 5 is smth we have to complete.

@HeikoKlare thinks that change might cause the flakyness and I currently do not have the time to investigate the real issue of the test flakyness, so I will restore the unit test to its original state so that the work of others is not blocked.

akurtakov:
I fail to see the reasoning. The assertions are exactly the same.

vogella:
a086677daf  "Improve fail message in QuickAccessDialogTest.testPreviousChoicesAvailableForExtension" added a String message argument to that very assertTrue, I will add the message to the revert.

vogella:
> I fail to see the reasoning. The assertions are exactly the same.

I also fail to see the reasoning but I do not have the time to convince others that this change is not the root.

akurtakov:
I even wonder whether  https://github.com/eclipse-platform/eclipse.platform.ui/pull/3916 exposes a limitation in MacOS SWT port . https://github.com/akurtakov/eclipse.platform.swt/blob/425c7b6c600689428ef9734822a54d4c920b4c61/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/widgets/Text.java#L510 - using different underlying native widgets .

HeikoKlare:
Can you shortly explain the intent of this PR? Is it supposed to be a temporary revert as a debugging means to see if the test would become less flaky again? Even if it did, I am not fully sure how that insight would help us.

> but I do not have the time to convince others that this change is not the root.

To be honest, it's sad to see that just posting analysis findings that may help us to find out what's going on is answered by some kind of defense action. I'd really much appreciate if we could just proceed with properly solving https://github.com/eclipse-platform/eclipse.platform.ui/issues/4009 and collecting whatever we need for analysis and resolution.

> I even wonder whether https://github.com/eclipse-platform/eclipse.platform.ui/pull/3916 exposes a limitation in MacOS SWT port .

Might be, but we should keep in mind that we also see the test failure happen on Windows (https://github.com/eclipse-platform/eclipse.platform.ui/issues/4009#issuecomment-4479378778) and for macOS we have only seen it happen on the x86 machine but not on the aarch64 one so far.

akurtakov:
Yeah, it's wild guessing for now. That's why I asked for these to not be merged . We have to understand the problem and do proper fix.

vogella:
Revert not desired.

PR #4018, opened 2026-05-19
Revert "QuickAccessDialog: render filter as native search field"

vogella:
Reverts eclipse-platform/eclipse.platform.ui#3916


vogella:
Shall this be merged? cc @HeikoKlare  and @akurtakov ?

HeikoKlare:
Could you please explain the intent of this PR? Without that it's impossible to answer that question.

akurtakov:
This seems to be due to my guess about SEARCH potentially exposing limitations on MacOS, which Heiko corrected me as being seen on Windows too. So I would dare to say - this shouldn't be merged.

Issue #2242, opened 2024-09-05
CompositeInformationControl replaces the original shell of its AbstractInformationControls


markdomeng:
at  [CompositeInformationControl](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/hover/CompositeInformationControl.java) : 111 (control.setParent(parent))
effectively replacing the original fShell of the AbstractInformationControl

this causes unexpected behaviors to its consumers. 
Some consumers I found while working on some LSP project (we are using GenericEditor): 
-  [MarkerInformationControl](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/markers/MarkerInformationControl.java)
   - at line 163, computeSizeHint() : this returns a very small Point(2,2) because the child controls are stolen. (Only if it is created by [CompositeInformationControlCreator](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/hover/CompositeInformationControlCreator.java) at line 123 as a CompositeInformationControl)
   - - is fixed by c8ddf0f27956ca2696db372d9e88c2bc3268ec23 but now the hover becomes very large
   - at line 149, setInput() : the shell is not getting disposed after applying the quickfix
   - proposed fixed in 7a51abb8f76b61069aa32191f4c958c64af0b1ab

- [BrowserInformationControl](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java)
   - at line 225, createContent() : I only observed this when using dark mode, but the background color of the BrowserInformationControl is different from the MarkerInformationControl when they are created as a CompositeInformationControl. I suspect this is the cause

also, the CompositeInformationControl always aggregate LSPHover even when there is no content on it, producing 
an extra status bar on the aggregated composite control
- proposed fix in 6e8dd5a0ef7d06de28378dc945ef6456aa45c193

_(I am not able to reproduce it in a regular eclipse, everything mentioned above were only observed in our project)_
## Tested under this environment:
* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): 
[...]

PR #4021, opened 2026-05-22
[#2242] fix wrong hover size of MarkerInformationControl

markdomeng:
the size of marker hover being set to a very small value (2,2) was fixed  in c8ddf0f27956ca2696db372d9e88c2bc3268ec23 but now it is getting the max size of the shell which is very large

solution is to pack the actual shell that contains the children of the marker control, and compute the size manually


iloveeclipse:
@markdomeng : could you please provide screenshots before / after and steps to reproduce?

PR #4023, opened 2026-05-22
Support concurrent access on ContextTypeRegistry.fContextTypes

### LINKED ISSUE #4020: ContributionContextTypeRegistry is not thread-safe
In our automated tests, we see such logged errors:

```
!ENTRY org.eclipse.egit.ui 4 0 2026-05-21 01:52:19.697
!MESSAGE Cannot register git support for Java templates
!STACK 0
java.util.ConcurrentModificationException
	at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:1024)
	at java.base/java.util.LinkedHashMap$LinkedValueIterator.next(LinkedHashMap.java:1053)
	at org.eclipse.egit.ui.internal.StartEventListener.lambda$2(StartEventListener.java:108)
	at org.eclipse.egit.ui.internal.StartEventListener.lambda$0(StartEventListener.java:70)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5084)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4549)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1147)
```

We started seeing the errors after activating the `org.eclipse.jdt.ls.core` early in our tests life-cycle.

JDT LS has initialization code that uses:

```
org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/PreferenceManager.java:                ContextTypeRegistry registry = JavaManipulation.getCodeTemplateContextRegistry();
``` 

EGit likewise, as seen in the stack trace above.

Since multiple clients want to initialize templates during their respective bundle activation, we need guidance on what needs to change in the respective bundles, or thread-safety in `ContributionContextTypeRegistry`.

trancexpress:
This change adjusts `ContextTypeRegistry` to use a `ConcurrentHashMap`.

NOTE: As a result, iteration order of `contextTypes()` is no longer deterministic, previously this was guaranteed by the underlying `LinkedHashMap`.

Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/4020

### DIFF
--- a/bundles/org.eclipse.text/src/org/eclipse/text/templates/ContextTypeRegistry.java
+++ b/bundles/org.eclipse.text/src/org/eclipse/text/templates/ContextTypeRegistry.java
@@ -14,8 +14,8 @@
 package org.eclipse.text.templates;
 
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.jface.text.templates.TemplateContextType;
 
@@ -32,7 +32,7 @@ import org.eclipse.jface.text.templates.TemplateContextType;
 public class ContextTypeRegistry {
 
 	/** all known context types */
-	private final Map<String, TemplateContextType> fContextTypes= new LinkedHashMap<>();
+	private final Map<String, TemplateContextType> fContextTypes= new ConcurrentHashMap<>();
 
 	/**
 	 * Adds a context type to the registry. If there already is a context type
--- a/bundles/org.eclipse.text/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.text
-Bundle-Version: 3.14.700.qualifier
+Bundle-Version: 3.14.800.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-Activator: org.eclipse.text.internal.Activator


bundles/org.eclipse.text/src/org/eclipse/text/templates/ContextTypeRegistry.java
```diff
@@ -60,6 +68,10 @@ public TemplateContextType getContextType(String id) {
 	 * @return an iterator over all registered context types
```
trancexpress, line 60: Should we adjust the JavaDoc to reflect that the iterator cannot be used to modify the context types?

iloveeclipse: Would make sense

trancexpress:
If we are merging this, does it need a N&N entry? We are potentially breaking client code.

merks:
Have you considered a concurrent map?  (I'm a little swamped so this is knee-jerk question.)

merks (changes requested):
We should not be doing this during the RC2 week.

trancexpress:
> Have you considered a concurrent map? (I'm a little swamped so this is knee-jerk question.)

Since we have to copy the map, we'd still have to use the map for a synchronized block in the iterator method. I'm not sure we gain much, but I guess it will reduce a few lines from the diff. Should I?

merks:
It's RC2 freeze week and @iloveeclipse mentioned M1 so I am explicitly blocking this for now.

Have you considered using java.util.concurrent.ConcurrentHashMap?  The iterator can make a copy of whatever it's iterating over.  But isn't this all easier if the map itself is thread safe?

trancexpress:
> Have you considered using java.util.concurrent.ConcurrentHashMap? The iterator can make a copy of whatever it's iterating over. But isn't this all easier if the map itself is thread safe?

We have a `LinkedHashMap` as a base, I don't think `ConcurrentHashMap` guarantees an order?

merks:
Does anything in the API suggest there is an order specified? Is order meaningful semantically?  Mostly it’s a wrapper around get and put I think. 🤔

iloveeclipse:
> Does anything in the API suggest there is an order specified? Is order meaningful semantically? Mostly it’s a wrapper around get and put I think. 🤔

Implementation uses LinkedHashMapy so order is at least desirable.

merks:
The implementation performs quickly without synchronization. 🤪 Now you want to synchronize everything rather than use a concurrent map, you want to change the semantics of the iterator to be a copy (and reject delete?) but you want to preserve an order that was never specified and has effectively no meaning with respect to multithreaded registrations.  Seems strange to me.  But that’s just my opinion and I don’t really care. Someone else might.

trancexpress:
> Now you want to synchronize everything rather than use a concurrent map

What is the difference in your view? Between concurrent and synchronized?

From my POV the concurrent case will likely perform better, but I'm not sure we are on a performance critical path in this code.

merks:
I just thought that a concurrent map had optimal performance for get, such that multiple threads doing a get did not block each other.  I’m no doubt splitting hairs, but preserving unspecified meaningless order because the current internals do that, albeit not in a thread safe way, is probably splitting a different hair.  I don’t really care so much and I’m pretty sure none of this is performance critical.

HeikoKlare:
Using a `ConcurrentHashMap` would not require the iterator to be explicitly made read only, would it? The reported problem is not an unintended state of the map being processes (which could be fixed with creating a copy for the iterator) but just that concurrent access via iterator on non-thread-safe data structures throws an exception (which is not the case for the `ConcurrentHashMap`).
So that solution would be least invasive (just removing entry order of the `LinkedHashMap` that does not seem to be contractually specified anyway), best in terms of performance (no `synchronized` overhead and no data structure copies) and less error prone on future changes (as with `synchronihzed` it's not obvious that the data structure is concurrently accessed, so that missing `synchronized` blocks on additional accesses may occur).

I would only propose to use the solution based on `synchronized` blocks in case there a known need for preserving the current order of entries in the map. In case just _any_ deterministic order shall be preserved, something like a `ConcurrentSkipListMap` may be an alternative.

trancexpress:
Adjusted to use a `ConcurrentHashMap`, as it was requested twice now. IMO non-linked hash maps lead to surprises on the long run, but I doubt we'll have issues here.

trancexpress:
> We should not be doing this during the RC2 week.

Would next week be a good time to merge this @merks ?

merks (approved):
The only use cases that seem kind of weird are these

This "randomly" uses the first item returned by the iterator.   Which item that is is of course determined by some indeterminate, undefined initialization order:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/53e3e920eef46855ec1e44cb2c950a30e57cabe1/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/templates/TemplatePreferencePage.java#L1207-L1211

This too at lines 1319 and 1336:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/53e3e920eef46855ec1e44cb2c950a30e57cabe1/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/templates/AbstractTemplatesPage.java#L1314-L1340

The rest of the uses iterate over the collection.  I doubt there  would be any concurrent modifications during the use cases I see in the Platform SDK itself.

Also, given the generally undefined nature of which item is first, even if addition order is preserved, because initialization order is not determinate, I'm not particularly concerned.

Given master is open now, I'm okay to merge this sooner rather than later.

PR #4027, opened 2026-05-25
Adjust wait for initial dialog contents in QuickAccessDialogTest

trancexpress:
Adjusted the wait for initial dialog contents in
`QuickAccessDialogTest.testPreviousChoicesAvailableForExtension()`, to rely on `QuickAccessDialog.setInfoText()` instead of relying on specific dialog contents set during the test.

See: https://github.com/eclipse-platform/eclipse.platform.ui/issues/4009


merks:
@trancexpress 

Feel free to merge changes intended to fix/improve flaky tests.

trancexpress:
Alright, though I have doubts this will help. I still see async code running despite the change, just less of it. The code in the job done listener is proving difficult to wait for... If that is even the cause for the I-build fails. So no rush from my side to merge this.

PR #4034, opened 2026-05-29
Use TestExecutionListener to log test start and end in org.eclipse.ui.tests

### LINKED ISSUE #4032: Use TestExecutionListener to log test start and end in org.eclipse.ui.tests
Tests in `org.eclipse.ui.tests` use `org.eclipse.ui.tests.harness.util.TestRunLogUtil.LOG_TESTRUN` to log/print when a test started and ended. This is a JUnit 4 rule which is present in some but not all tests. E.g. for https://github.com/eclipse-platform/eclipse.platform.ui/issues/4009, `QuickAccessDialogTest` doesn't have this rule and is a JUnit 5 test.

To add such log information for `QuickAccessDialogTest`, following the same pattern, we would have to define a JUnit 5 extension and add it to `QuickAccessDialogTest`.

We can instead do better and use `org.junit.platform.launcher.TestExecutionListener` to print start and end for all tests, not just JUnit 4 ones using `LOG_TESTRUN`. We would need to remove `LOG_TESTRUN` from the tests, to avoid double printing.

For an example of how to use `TestExecutionListener`, see e.g.: https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/accc76e590eb0dcb57f96e9f6924011620f8b77b

trancexpress:
Add a JUnit Jupiter `TestExecutionListener` to `org.eclipse.ui.tests`, which logs test start and end events via `UIPlugin`.
The listener is defined in `org.eclipse.ui.tests.harness`, so that it can be reused by other test bundles.

Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/4032

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/MultipleWindowsTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/MultipleWindowsTest.java
@@ -19,20 +19,15 @@ import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.WorkbenchException;
-import org.eclipse.ui.tests.harness.util.TestRunLogUtil;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Rule;
 import org.junit.Test;
-import org.junit.rules.TestWatcher;
 
 /**
  * A set of tests for multiple monitor situations that ensures interactions are
  * isolated to the respective window.
  */
 public class MultipleWindowsTest {
-	@Rule
-	public TestWatcher LOG_TESTRUN = TestRunLogUtil.LOG_TESTRUN;
 
 	IWorkbench wb;
 	IWorkbenchWindow win1;
--- a/tests/org.eclipse.ui.tests.harness/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.ui.tests.harness/META-INF/MANIFEST.MF
@@ -10,7 +10,9 @@ Require-Bundle: org.eclipse.ui;bundle-version="3.208.0",
  org.eclipse.core.resources
 Bundle-ActivationPolicy: lazy
 Import-Package: org.junit.jupiter.api;version="[5.0.0,6.0.0)",
- org.junit.jupiter.api.extension;version="[5.0.0,6.0.0)"
+ org.junit.jupiter.api.extension;version="[5.0.0,6.0.0)",
+ org.junit.platform.engine;version="[1.14.0,2.0.0)",
+ org.junit.platform.launcher;version="[1.14.0,2.0.0)"
 Export-Package: org.eclipse.ui.tests.harness,
  org.eclipse.ui.tests.harness.tests,
  org.eclipse.ui.tests.harness.util,
--- a/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/LogTestListener.java
+++ b/tests/org.eclipse.ui.tests.harness/src/org/eclipse/ui/tests/harness/util/LogTestListener.java
@@ -1,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Simeon Andreev and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.ui.tests.harness.util;
+
+import org.eclipse.core.runtime.Status;
+import org.eclipse.ui.internal.UIPlugin;
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+
+/**
+ * Logs tests start and end.
+ */
+public class LogTestListener implements TestExecutionListener {
+
+	@Override
+	public void executionStarted(TestIdentifier id) {
+		logInfo(id.getDisplayName() + " STARTING");
+	}
+
+	@Override
+	public void executionFinished(TestIdentifier id, TestExecutionResult result) {
+		logInfo(id.getDisplayName() + " DONE: " + result.getStatus());
+	}
+
+	private static void logInfo(String message) {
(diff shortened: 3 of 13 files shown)


tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/MultipleWindowsTest.java
```diff
@@ -19,20 +19,15 @@
...
 import org.junit.Test;
-import org.junit.rules.TestWatcher;
 
 /**
  * A set of tests for multiple monitor situations that ensures interactions are
  * isolated to the respective window.
  */
 public class MultipleWindowsTest {
-	@Rule
-	public TestWatcher LOG_TESTRUN = TestRunLogUtil.LOG_TESTRUN;
```
HeikoKlare, line 35: These rules should not be removed before the test class has been migrated to JUnit 5. Otherwise the log output is lost until the migration is done. You can, e.g., see that the logs for this PR do not contain logging for the `testIndependentWorkingSets` of this class anymore whereas previous builds contain it.

HeikoKlare (commented):
The change looks good to get the logging of execution orders for JUnit 5 UI tests again. I just don't think we should move the JUnit 4 rules until those tests have been migrated to JUnit 5.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/MultipleWindowsTest.java
trancexpress: Where can I see the logs actually? I checked the build job link but nothing was immediately obvious.

I thought JUnit 4 tests would be run with the vintage engine, but the listener would still be notified, resulting in double logging. If that is not the case, no issue to keep the rule.

trancexpress:
> The change looks good to get the logging of execution orders for JUnit 5 UI tests again. I just don't think we should move the JUnit 4 rules until those tests have been migrated to JUnit 5.

Sure, that makes things even more simple.

If we do see duplicated logged starts/ends in the first I-build with the changes, we can check where and why.

tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/api/MultipleWindowsTest.java
HeikoKlare: You can see the logs in the GHA runs that you reach when navigating to the details for the PR checks. For example, here is the Linux run for the latest change you made: https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/26677523134/job/78631997281?pr=4034

This one logs the `MultipleWindowTests` again:
image

This is the run with your previous changes: https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/26627134343/job/78466525141
Which does not log the `MultipleWindowTests`:
image

HeikoKlare (approved):

HeikoKlare (commented):
I just noted that the log outputs for the JUnit 4 tests are not missing anymore, but the outputs for the JUnit 5 tests are not present either. This seems to be caused by the `org.junit.platform.launcher` import being required by the actual test bundle ( `org.eclipse.tests.ui`) as well. Adding that package import to the test bundle leads to proper log output.
And then actually logging is also done for the JUnit 4 tests executed by the vintage engine, so your initial changes to remove the rules for those test swere actually correct. The issue was just that no logging by the `TestExecutionListener` was done at all.

trancexpress:
Anything else from your side @HeikoKlare ? If not I'll merge this.

HeikoKlare:
Is it working as expected now? I would have expected some `... STARTING` and `... DONE` logging for the tests in `org.eclipse.ui.tests`, which I also get during local execution, e.g.:
```
!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-01 08:51:19.238
!MESSAGE JUnit Platform Suite STARTING

!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-01 08:51:19.242
!MESSAGE JUnit Platform Suite DONE: SUCCESSFUL

!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-01 08:51:19.243
!MESSAGE JUnit Vintage STARTING

!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-01 08:51:19.258
!MESSAGE MultipleWindowsTest STARTING

!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-01 08:51:19.264
!MESSAGE testIndependentWorkingSets STARTING

!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-01 08:51:20.068
!MESSAGE testIndependentWorkingSets DONE: SUCCESSFUL
```

But in the Jenkins and GHA logs for this PR, I do not find anything like that.

trancexpress:
> Is it working as expected now? I would have expected some `... STARTING` and `... DONE` logging for the tests in `org.eclipse.ui.tests`, which I also get during local execution, e.g.:
> [..]
> But in the Jenkins and GHA logs for this PR, I do not find anything like that.

The start and end messages are logged in the workspace log, not printed to command line. I guess we can do both, if you require this, but the entries in the workspace log are the most useful for me. Its easy to tell what logged errors occurred during which test, which tracing belongs to which test and so on.

HeikoKlare:
I see. For me it's fine to not have verbose output to the command line for this. Do we have access to the workspace logs for GHA and Jenkins executions in PRs? I did not find those logs yet.

trancexpress:
> I see. For me it's fine to not have verbose output to the command line for this. Do we have access to the workspace logs for GHA and Jenkins executions in PRs? I did not find those logs yet.

I don't know where the logs are for GH jobs.

I check the I-build ones (e.g. https://download.eclipse.org/eclipse/downloads/drops4/I20260531-1800/testresults/logs.html), if the starts and ends are missing in the first I-build with the change here, I'll check what is going wrong.

trancexpress:
Anything else here @HeikoKlare ?

trancexpress:
Merging this in a few hours if there are no objections.

trancexpress:
The fails are https://github.com/eclipse-platform/eclipse.platform.ui/issues/4056, I'll wait for that to get fixed, just in case.

HeikoKlare (approved):

trancexpress:
I do see the log entries: https://download.eclipse.org/eclipse/downloads/drops4/I20260605-1800/testresults/ep441I-unit-linux-x86_64-java21_linux.gtk.x86_64_21/org.eclipse.ui.tests.UiTestSuite.txt

E.g.:

```
!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-06 00:43:12.943
!MESSAGE testGetLockInfoFile() STARTING

!ENTRY org.eclipse.ui.tests.harness 1 0 2026-06-06 00:43:12.953
!MESSAGE testGetLockInfoFile() DONE: SUCCESSFUL
```

Issue #4041, opened 2026-05-31
Preference pages not shown in filtered preference dialog (regression)


wahlbrink:
When opening a pre-filtered preference dialog, no pages are shown in the left tree.

## Steps to reproduce
- Open a Java file in the Java editor
- Open "Preferences..." from the context menu

In 2026-06 RC2 the dialog looks like this:
<img width="1354" height="1083" alt="Image" src="https://github.com/user-attachments/assets/1efaa075-080c-44cf-b14c-7a72001f082d" />
It seems that the entries are displayed for just a very brief moment, but then disappear.

## Expected behaviour:
The filtered preference pages are shown in the left tree, screenhot from 2025-12:
<img width="1354" height="1091" alt="Image" src="https://github.com/user-attachments/assets/40650f43-06b3-4730-a499-4bfb24db3e0f" />

## Tested under this environment:
Eclipse IDE: 2026-06 (4.40) I20260528-0011 (RC2)
OS: Windows 11

HeikoKlare:
At a first glance, this seems to be caused by "type filter text" not being embedded as a hint for the filter but as an actual filter string (which does not match any entries).

It is a regression from:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3925

Reverting that change restores the previous behavior (@vogella fyi). Not sure if the previous behavior was a perfect implementation, but at least it did not have the reported issue, so a quick immediate fix could be a revert of that change.

I am not yet sure how severe the issue is, i.e., whether we need to think about an RC2 respin.

HannesWell:
> I am not yet sure how severe the issue is, i.e., whether we need to think about an RC2 respin.

Until Friday, i.e. before the promotion of the GA artifacts, a RC2a respin is easily possible:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/RELEASE.md#rc2-respin

Since this is a clear bug and if the fix is very save, which I think a revert to the previous state is, I think a RC2 respin is justified.
I wouldn't do a 'forward' fix for the 4.40 maintenance branch at this stage. But on the master branch the latter could/should be done of course.

merks:
SimRel needs final contributions on Wednesday. There is sufficient runway for that so I would support that.

iloveeclipse:
I'm also for revert in RC2a, sooner is better.

HeikoKlare:
[...]

PR #4043, opened 2026-06-01
Revert "PreferenceDialog: render filter as native search field"

vogella:
Reverts #3925, the cause of #4041: the pre-filtered preference dialog (for example opening Preferences from a Java editor context menu) shows an empty page tree because the filter hint is applied as an actual filter pattern that matches no pages.

As agreed in the issue, this restores the previous working behavior for the 4.40 RC2a respin. A proper forward fix on master is handled separately.


HeikoKlare (approved):
+1 for merging this and doing an RC2 respin from my side

merks:
I tested that it restores the behavior with this dialog content:

image

merks:
I think the build failure is complaining just about this, which is not something that changed as part of this PR

https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-4043/1/eclipse/new/source.912b3483-e427-4033-afb9-9a5ac4805260/#163

So I think it's okay to ignore this and merge this change for a respin.

akurtakov:
As it has been tested and fixes the problem - GO.

PR #4044, opened 2026-06-01
Fix empty page tree in pre-filtered preference dialog

vogella:
Forward fix for #4041 on master (the 4.40 maintenance branch gets the plain revert in #4043).

The native search field rework in #3925 stopped maintaining the inherited initialText sentinel, so the hint that addFilter() wrote into the field was applied as a real filter pattern and hid every preference page. This keeps the sentinel in sync and no longer writes the hint into the field, so a pre-filtered dialog (for example Preferences from a Java editor context menu) shows its pages again. The native search field is preserved, and the "type filter text" hint is kept as the placeholder message so the field still signals that it filters the pages.

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/FilteredPreferenceDialog.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/FilteredPreferenceDialog.java
@@ -126,8 +126,12 @@ public abstract class FilteredPreferenceDialog extends PreferenceDialog implemen
 
 		@Override
 		public void setInitialText(String text) {
+			// Track initialText for the empty-filter check and accessible name, and show it
+			// as the field's placeholder hint. Unlike the base implementation we never write
+			// it into the field, so addFilter() cannot turn the hint into an active pattern.
+			initialText = text != null ? text : ""; //$NON-NLS-1$
 			if (filterText != null && !filterText.isDisposed()) {
-				filterText.setMessage(text != null ? text : ""); //$NON-NLS-1$
+				filterText.setMessage(initialText);
 			}
 		}
 
@@ -140,7 +144,8 @@ public abstract class FilteredPreferenceDialog extends PreferenceDialog implemen
 			getViewer().addFilter(filter);
 
 			if (filterText != null) {
-				setFilterText(WorkbenchMessages.FilteredTree_FilterMessage);
+				// Keep the field empty so the viewer filter narrows the pages; the
+				// hint text would otherwise be applied as a pattern and hide all pages.
 				textChanged();
 			}
 


vogella:
@wahlbrink FYI, I also tested your example and it works fine.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/FilteredPreferenceDialog.java
```diff
@@ -140,7 +140,8 @@ protected void addFilter(ViewerFilter filter) {
 			getViewer().addFilter(filter);
 
 			if (filterText != null) {
-				setFilterText(WorkbenchMessages.FilteredTree_FilterMessage);
```
HeikoKlare, line 143: Doesn't this removal make other places using that NLS obsolete or maybe kind of incompatible, such as ... ?
https://github.com/eclipse-platform/eclipse.platform.ui/blob/a6ae91ce2cc693e1c3e905737ca5fbd1c7abaacc/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/dialogs/FilteredTree.java#L619

HeikoKlare (commented):
With this change, the filterered trees in preference dialogs don't have any hint on the filter field at all anymore. Is that intended?

For the filtered preference dialog one might argue that it's okay because the filter text was only presented as ordinary filter text input when opening the dialog, so it might be somehow okay to remove that:
Image

But the ordinary preferences dialog now looks like this:
Image

While it used to contain a proper hint:
Image

Edit: just saw that there is the search symbol in front of the text now. Still I thought that there is a possibility to show a hint in such a search box, isn't it? I would rather expect that to be shown (like it was in the preferences dialog before). Otherwise it becomes less obvious that this is filter field rather than a search field.

vogella:
The search icon is the hint. I can add the text back if you prefer.

HeikoKlare:
In the current `master` state, we have the icon via hint and still a proper message (hint) about the filter functionality:
image

As said, in my opinion only having the search icon is wrong, as this is not a search field but a filter field. One may even argue that the search icon is not correct as it would need to be a filter icon.

vogella:
I restored the text hint

HeikoKlare:
Is there a specific reason why my review comment was left uncommented and the PR was merged without responding to it?

HeikoKlare:
Due to lack of response to my review comment, merging the PR despite the open review comment and ignoring my comment left after merging the PR, I plan to revert this change. @vogella fyi

vogella:
> In the current master state, we have the icon via hint and still a proper message (hint) about the filter functionality:


I thought I did adjust the pr based on your request. Is it now in sync with the master in the release, was that not your request @HeikoKlare ?

vogella:
> I restored the text hint

I assumed that was your suggestion.

HeikoKlare:
As said, my review comment is completely unanswered: https://github.com/eclipse-platform/eclipse.platform.ui/pull/4044#discussion_r3336823063

vogella:
> As said, my review comment is completely unanswered: [#4044 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/4044#discussion_r3336823063)

I restored the message based on your request, hence this feedback was outdated as we do not remove this text.

Technical we have this:

The message is not obsolete. FilteredTree_FilterMessage is still used at FilteredTree.java:294 and the comparison at line 619 still applies to the base FilteredTree and the subclasses that write the message into the field content. In FilteredPreferenceDialog we now keep the field empty instead of writing the message into it, but narrowingDown evaluates identically: with previousFilterText empty, getFilterString().startsWith("") is always true, which is the same result the equals(FilteredTree_FilterMessage) branch produced before. So no usage becomes dead and there is no behavioral change in that path.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/dialogs/FilteredPreferenceDialog.java
vogella: The message is not obsolete in the merge PR version. FilteredTree_FilterMessage is still used at FilteredTree.java:294 and the comparison at line 619 still applies to the base FilteredTree and the subclasses that write the message into the field content. In FilteredPreferenceDialog we now keep the field empty instead of writing the message into it, but narrowingDown evaluates identically: with previousFilterText empty, getFilterString().startsWith("") is always true, which is the same result the equals(FilteredTree_FilterMessage) branch produced before. So no usage becomes dead and there is no behavioral change in that path.

vogella:
I think you said two things in your feedback: "We should not remove the text" and "If we remove the text, this NLS becomes obsolete" so I assumed "I restored the text" answered both, I should have been more explicit about it, to avoid misunderstanding.

HeikoKlare:
> I restored the message based on your request, hence this feedback was outdated as we do not remove this text.

That's not correct. You restored the message, but not the text. The question I asked refers to the text being compared.

So with this change the filtered preference dialog behaves different in terms of UX for filtered trees than all other filtered trees (they still provide the message as initial text, which is why the code I asked for probably needs to stay). Is that intended? Don't we want to have consistent UX thoughout the usages of filtered trees?

vogella:
On the inline question: the FilteredTree_FilterMessage NLS isn't obsolete. It's still used at FilteredTree.java:294 and the comparison at line 619 still applies to the base FilteredTree and subclasses that write the message into the field value. In FilteredPreferenceDialog we now keep the field empty, but narrowingDown evaluates identically (with an empty previousFilterText, getFilterString().startsWith("") is always true, matching the old equals(FilterMessage) branch). So nothing becomes dead and there's no behavioral change.

On the hint itself: the intention of #3925 was to move the filter field away from the initial-text-in-value pattern, which is a recognized UX anti-pattern (e.g. https://www.nngroup.com/articles/form-design-placeholders/): screen readers can treat it as a real value, and it needs clear-on-focus handling. setMessage gives the same visible hint without those problems, and the hint is now restored that way.

I agree we should be consistent. My preference is to move the remaining places that still push the hint into the field value over to setMessage too, rather than reintroducing the old behavior here. Changing the global default is harder to land in this project, so I'd treat that separately. That said, if you feel strongly that the old initial-text behavior should be restored, I'm happy to do it. Let me know.

HeikoKlare:
> I agree we should be consistent. My preference is to move the remaining places that still push the hint into the field value over to setMessage too, rather than reintroducing the old behavior here. Changing the global default is harder to land in this project, so I'd treat that separately. That said, if you feel strongly that the old initial-text behavior should be restored, I'm happy to do it. Let me know.

No, I do not want to propose to move back. I fully agree that properly embedding the hint via `setMessage` is the right thing to do. My question was rather meat to ask if we shouldn't move forward with the `FilteredTree` implementation itself as well. I currently don't see why we need to keep the message as initial field text there and cannot remove it just like it was removed here.

vogella:
> I currently don't see why we need to keep the message as initial field text there and cannot remove it just like it was removed here.

I agree. I will create a PR for that and add you as reviewer to have the discussion there.

vogella:
FYI - AbstractFilteredViewerComposite is the place to do this, not  FilteredTree, will be visible in the upcoming PR.

vogella:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/4074

PR #4047, opened 2026-06-01
Add copy refactoring infrastructure to LTK

Felix-Schmid:
LTK is missing infrastructure for copy refactoring, e.g., a `CopyResourcesProcessor`. Eclipse RCP developers can define their own copy participants (similar to move/delete/rename participants), but they are never called.

This PR adds copy refactoring to LTK based on the existing implementation of move/delete refactoring, the existing refactoring for copying projects, and the copy processor implementation of JDT.

The changes were tested in Eclipse 4diac, where we could now perform refactoring operations (via participants) during copying of resources.


vogella:
I see no tests, please add them.

bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyResourcesDescriptor.java
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
```
BeckerWdf, line 12:
```suggestion

```
I assume the header is copied from somewhere. I did you put IBM in here by intention?

bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyResourceChange.java
```diff
@@ -0,0 +1,151 @@
...
+import org.eclipse.ltk.core.refactoring.Change;
+import org.eclipse.ltk.core.refactoring.ChangeDescriptor;
+import org.eclipse.ltk.core.refactoring.CompositeChange;
+import org.eclipse.ltk.core.refactoring.NullChange;
+import org.eclipse.ltk.core.refactoring.RefactoringStatus;
+import org.eclipse.ltk.core.refactoring.participants.ReorgExecutionLog;
+import org.eclipse.ltk.internal.core.refactoring.BasicElementLabels;
+import org.eclipse.ltk.internal.core.refactoring.RefactoringCoreMessages;
+
+public class CopyResourceChange extends ResourceChange {
```
iloveeclipse, line 44: If this is supposed to be a public API, you must bump bundle version to `3.16.0` and add `since 3.16` javadoc tags on every new API type

Felix-Schmid: Thanks for the heads-up, done.
Everything that was made public was done to align in with how the move/delete classes do it, a discussion about this was also had here https://github.com/eclipse-platform/eclipse.platform.ui/pull/2262#discussion_r1856718005.

bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyResourcesDescriptor.java
Felix-Schmid: This class needed to be very similar to the `DeleteResourcesDescriptor`, so I copied the file and made the necessary changes. I kept the original header to show that this was not all my creation. The same principle also applies to `CopyResourcesHandler` & `CopyResourcesRefactoringContribution`.
I can change the headers if this was incorrect or if there is a better way to do it.

merks: That’s okay. Increase the upper year to the current year.

Felix-Schmid: Done.

bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyResourceChange.java
iloveeclipse: API tooling complains, and I see now the package is not exported as API. Is this intentional?

merks:
Note that you're generally best to maintain a single commit which you amend and force push until it's done...

Felix-Schmid:
> I see no tests, please add them.

Tests have now been added.

> Note that you're generally best to maintain a single commit which you amend and force push until it's done...

Understood, I will do that with the next commit.

bundles/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/core/refactoring/resource/CopyResourceChange.java
Felix-Schmid: Should be fixed now.

vogella:
org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests.testBut336225 is unrelated, see https://github.com/eclipse-platform/eclipse.platform/issues/2700

vogella:
@Felix-Schmid can you think about a usage in platform or PDE for this new API? It would be really nice to have a user for this API in our own code  base

Felix-Schmid:
> can you think about a usage in platform or PDE for this new API? It would be really nice to have a user for this API in our own code base

I am not too familiar with the platform or PDE, but maybe copy refactoring could be used to update certain properties of MANIFEST, plugin, or *.product files? I am open to a more concrete suggestion here.

For an example of how this copy refactoring could be used, I can link to the [CopyTypeParticipant in 4diac](https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.typemanagement/src/org/eclipse/fordiac/ide/typemanagement/refactoring/copy/CopyTypeParticipant.java), which updates the package name for copied type files.
This infrastructure would also be needed elsewhere; e.g., [this PR for Xtext](https://github.com/eclipse-xtext/xtext/pull/3588#discussion_r2955882632) has to reuse the copy refactoring from JDT, since the platform does not provide it.

vogella:
IMHO the projects leads of platform should decide here if this API should be accepted without platform or pde or JDT consumers. 

cc @HeikoKlare @akurtakov @iloveeclipse

akurtakov:
While it's best to have Platform/PDE/JDT user if the API is generic enough and improves situation for multiple other simrel projects -  I can see this as good enough situation. Review should be stressing on "generic enough" solution.

vogella:
Given that one PL gave his OK and the missing review activity of other committers, I plan to help @Felix-Schmid to finish this one and merge it once I'm ok with it.

vogella:
Thanks again @Felix-Schmid, first of all this PR needs a rebase.

And here are some findings:

**Routing all IDE copies through LTK is too broad** 

CopyFilesAndFoldersOperation.performCopy / performCopyWithAutoRename now call LTKLauncher.copyResources(...) and return true before the existing logic runs. That short-circuit causes three problems:

1. "Create Link" / "Create Virtual Folder" is silently turned into a physical copy.
performCopy applies setCreateLinks / setVirtualFolders / setRelativeVariable after the new early return.
The Project Explorer drop handler (ResourceDropAdapterAssistant) sets createLinks = true and then calls the public copyResources(IResource[], IContainer), which reaches performCopy.
With the LTK UI bundle installed (the default IDE), the LTK path copies the files physically and the link / virtual folder the user requested is never created.

2. A failed or fatal LTK copy is reported as success and suppresses the fallback.
LTKLauncher.runCommand returns true whenever the command is defined and the handler does not throw, and CopyResourcesHandler.execute always returns null (it catches CoreException, shows a dialog, and does nothing when the selection/destination guard is false).
So on a fatal condition (destination is the workspace root, destination inside the copied resource, multiple destinations, etc.) copyResources returns true, performCopy returns true, and the real copy below is skipped: the resource is silently not copied, where the old path always copied.

Suggestion: gate the LTK route (skip it when createLinks || createVirtualFoldersAndLinks) and fall through to the legacy operation when the refactoring produced no executed change, rather than short-circuiting unconditionally.

**Folder overwrite replaces instead of merges (possible data loss)**

CopyResourceChange.deleteIfAlreadyExists deletes the whole existing destination folder and then recreates it via origin.copy(...).
The legacy CopyFilesAndFoldersOperation merges homogeneous folders file by file, so after the user accepts the overwrite prompt (expecting a merge) any files that existed only in the destination folder are lost.
CopyResourcesProcessor.checkFinalConditions returns an empty status, so unlike MoveResourcesProcessor there is no overwrite warning and no ResourceChangeChecker participation.
(Undo restores correctly via history.)

**The participant test does not run**

CopyRefactoringWithRefUpdateTest provides no coverage for two independent reasons:

1. It is not referenced by AllTests / ParticipantTests, and Tycho pins the root suite via testClass, so nothing discovers it.
2. It imports JUnit 4 org.junit.Test while using Jupiter @BeforeEach / @AfterEach, and testCopyRefactoringWithParticipants() is package-private; Jupiter ignores the JUnit 4 @Test and junit-vintage is not on the bundle's imports.

The sibling MoveRefactoringWithRefUpdateTest it was copied from uses org.junit.jupiter.api.Test and is registered in ParticipantTests.
(The additions to ResourceRefactoringTests do run.)

**Minor**

- CopyResourceChange.perform calls pm.beginTask(getName(), 2) and then two separate SubMonitor.convert(pm, 1) calls, each re-calling beginTask on the same monitor, so the progress reporting is wrong. MoveResourceChange uses one SubMonitor.convert(monitor, getName(), 4) plus newChild(...).
- CopyResourcesProcessor: destination = (IContainer) ...findMember(destPath) can throw ClassCastException for a malformed/replayed descriptor whose destination parent is a file. MoveResourcesProcessor avoids this by taking a typed IContainer.
- CopyResourceChange returns NullChange in the "identical resource already present" branch without calling log.markAsProcessed(origin), inconsistent with the normal path.
- CopyResourcesProcessor.createDescriptor() sets setProject(null) where Move uses the destination's project name (affects refactoring-history metadata only).

Felix-Schmid:
Thank you very much for the detailed review, I will get back with an updated PR.

Felix-Schmid:
I believe I have now addressed the review findings. For the `CopyResourcesHandler`, I was not sure if I should still show the dialogs to indicate problems with the LTK refactoring before falling back to the regular copy operation.

vogella:
@Felix-Schmid thanks. Just a quick feedback, as you increasing the version of the core and adding new API I believe you also have to adjust the dependency range in the ui plug-in.

vogella:
Also the FATAL case reported earlier is now correctly implemented but the RefactoringStatus.WARNING does seem to trigger a second copy:

````
PerformChangeOperation perform = new PerformChangeOperation(create); // create uses checkFailedSeverity = FATAL
perform.run(new NullProgressMonitor());          // (1) copy is actually performed here
if (perform.getConditionCheckingStatus().getSeverity() >= RefactoringStatus.WARNING) {
    openErrorDialog(shell, perform.getConditionCheckingStatus());
    throw new ExecutionException(...);           // (2) throws AFTER the copy already happened
}
````

The thrown ExecutionException is swallowed by LTKLauncher.runCommand (empty catch → return false), so tryPerformLTKcopy returns false and CopyFilesAndFoldersOperation falls through to the legacy CopyResourcesOperation, copying the same resources a second time into a destination LTK already populated.

Can this second copy be avoided?

Felix-Schmid:
> I believe you also have to adjust the dependency range in the ui plug-in

Done.
> the RefactoringStatus.WARNING does seem to trigger a second copy:

Good point, based on the severity descriptions, the copy is still performed for WARNING and ERROR. I changed it to only fall back in the FATAL case.

vogella:
Thanks @Felix-Schmid, AFAICS all feedback has been adressed and I plan to merge this tomorrow unless someone else identifies additional issues.

vogella:
Thanks @Felix-Schmid for this contribution. Please add an entry to the N&N for the next release so that people can discover this git@github.com:eclipse-platform/www.eclipse.org-eclipse.git. 

If time permits, please link issues in which you use this API in Eclipse projects for example in Eclipse 4diac to this ticket.

Issue #3980, opened 2026-05-07
CSS engine re-work


vogella:
The CSS engine is slow, bloated and overdesigned for its current usage. I'm the primary developer working on the CSS engine over the last 10 years, so I think I know what I'm taking about.

After increasing the test coverage in the 4.40 release to catch regression I plan to re-work the CSS engine in 4.41 to make it faster and easier to maintain.

Highlights of the planned changes:

* flatten engine/helper hierarchies (~-2,000 LOC, low risk)                                                                                                                                                                                                                                          
* replace W3C DOM mirror with POJOs/records (~-3,500 LOC, medium risk)                                                                                                                                                                                                                               
* replace SAC parser layer, keep Batik directly (~-2,000 LOC, medium risk)                                                                                                                                                                                                                           
* collapse property-handler dispatch, deprecate extension point (~-2,500 LOC, high risk)                                                                                                                                                                                                             
* merge css.swt.theme into css.swt (~-200 LOC, medium risk) 

The exsting CSS styling will remain the same just the internal handling will be re-worked. I expect also a much faster execution time of the re-worked CSS engine, which I also will measure. 

The above should reduce the CSS code by around 30% more if you remove the added lines for the additional tests. CSS engine is internal API, so no API breaking changes.

vogella:
The CSS engine rework has been in the SDK for a few weeks now without regressions reported, so here is where it stands.

Merged: mechanical cleanups, a selector and parser test safety net, flattened engine hierarchies, the SAC and Batik removal, the W3C DOM mirror replaced by a record based value model, and the property handler consolidation.

Size, over org.eclipse.e4.ui.css.core, org.eclipse.e4.ui.css.swt and org.eclipse.e4.ui.css.swt.theme:

|            | before | now    | delta         |
|------------|--------|--------|---------------|
| classes    | 262    | 175    | -87 (-33%)    |
| Java lines | 29,915 | 20,514 | -9,401 (-31%) |

Third party dependencies: org.apache.batik.css and the org.w3c.css.sac packages are gone from both bundles. org.eclipse.e4.ui.css.core now requires only org.eclipse.equinox.common and org.eclipse.core.runtime, with no Import-Package at all. Both libraries have since been dropped from the SDK target platform (eclipse-platform/eclipse.platform.releng.aggregator#3917 for Batik, eclipse-platform/eclipse.platform.releng.aggregator#3925 for SAC and xml-apis-ext).

[...]

PR #4051, opened 2026-06-02
Replace SAC selector layer with an internal Selector AST and matcher

vogella:
Replaces the vendored Batik/SAC selector tree with a small engine-internal selector model: a sealed Selector AST (records for type, class, id, attribute, pseudo, compound, and combinator forms) plus a static SelectorMatcher. A SAC to internal translator at the parser-output boundary feeds it, and CSSEngine.matches/parseSelectors now work on the internal AST, so 23 vendored impl/sac selector and condition wrappers fall away along with the dead ExtendedDocumentCSS query methods. Net around -1,450 lines; the bundles are internal (x-friends only) so there is no API surface change.

Specificity still follows CSS 2.1 and the static-pseudo-instance carve-out is preserved, so cascade behaviour does not shift. Sixteen new matcher unit tests cover the selector forms and specificity arithmetic, and CSSEngineTest and SelectorTest are rewritten on the internal AST. One deliberate behaviour note for reviewers: the adjacent-sibling matcher returns false (rather than throwing) on elements without sibling support. A local theme-swap benchmark on a busy workbench (20 editors, ~4000 widgets) measured a roughly 10% faster swap from removing the wrapper overhead; the larger matching and double-styling optimizations are separate later steps.

Three impl/sac classes remain as parser plumbing for a following step. Draft while the medium-risk matching path gets a closer look.

Contributes to #3980


akurtakov:
Is the faster swap coming from less SWT method calls?

vogella:
> Is the faster swap coming from less SWT method calls?

Less overhead in the framework , the less SWT method calls is still in the later optimization pipeline. As this is riskier I did not include that yet.

PR #4057, opened 2026-06-03
Trigger Show In for typed selection in Open Resource dialog

vogella:
In the Open Resource dialog, the Show In action for a non-file resource (for example a project) was gated on the OK button's enabled state. When the resource was reached by typing a pattern instead of clicking the row, that enabled state could lag behind the auto-selected first match, so pressing Enter or the Open button did nothing.

This acts on the selected item directly, the same way the dialog opens a typed file. A first-matched project now triggers Show In via Enter or the Open button, making the keyboard and mouse paths behave consistently.


vogella:
Test failure related to https://github.com/eclipse-platform/eclipse.platform/issues/2700

PR #4061, opened 2026-06-04
Update WorkbenchActionBuilder to use IEclipsePreferences

deepika-u:
IEclipsePreferences is used to modernize WorkbenchActionBuilder by updating references of org.eclipse.core.runtime.Preferences.

Partially fixes
https://github.com/eclipse-platform/eclipse.platform/issues/497

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/WorkbenchActionBuilder.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/ide/WorkbenchActionBuilder.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


deepika-u:
@HeikoKlare 
When you get some time, can you take a look at this please?

HeikoKlare (commented):
Thank you for this enhancement. The change looks reasonable to me.
Just as a minor comment: the change regarding usage of `CommandContributionItemParameter` seems unrelated to preferences. It would be cleaner to have that in a separate commit/PR to stay at one concern per commit (or otherwise at least explain that change in the commit message and PR as well).

deepika-u:
> Thank you for this enhancement. The change looks reasonable to me. Just as a minor comment: the change regarding usage of `CommandContributionItemParameter` seems unrelated to preferences. It would be cleaner to have that in a separate commit/PR to stay at one concern per commit (or otherwise at least explain that change in the commit message and PR as well).

Sure will separate it away. Actually that change was handling an illegal reference warning.

deepika-u:
@HeikoKlare 
Created new pr -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/4068 for the other change. 

Thanks for your time :)

HeikoKlare:
Thank you for the update. The change looks fine now.

PR #4062, opened 2026-06-04
Fix warnings in jface.text.tests

akurtakov:
Had to add forRemoval=true to DefaultUndoManager constructor as this is not inherited from the class.
Removed some useless comments and added more Deprecated to tests implementation to get the bundle clean.


akurtakov:
Initiated by the o.e.jface.text.tests bundle being reported as only qualifier bump in https://download.eclipse.org/eclipse/downloads/drops4/I20260603-1800/buildlogs/reporeports/reports/versionChecks.html

akurtakov:
8 warnings less. Test failures are unrelated. Merging.

PR #4066, opened 2026-06-05
Remove contents of IPageLayout that are deprecated for removal

HeikoKlare:
A fastview constant and two methods of IPageLayout have been deprecated for removal for almost 3 years. This change finally removes those contents.


vogella:
LGTM, API removals should be be merged in M1 so I recommend doing this soon

HeikoKlare:
Failing test already known and tracked via https://github.com/eclipse-platform/eclipse.platform.ui/issues/80

PR #4073, opened 2026-06-09
Fix toolbar items enabled state not updating automatically (#3953)

### LINKED ISSUE #3953: Toolbar items enabled state doesn't change automatically
We updated the dependencies of an Eclipse RCP app from `2018-09` to `2025-06`, and some toolbar items enabled state stopped getting updated automatically. The app is running on the compatibility layer and using custom source providers and `<enabledWhen>` expressions to control toolbar items handlers enabled state.

The following solutions didn't solve the problem:
- Storing the variables inside `IEclipseContext` directly instead of using `AbstractSourceProvider`
- Using E4 handler with `canExecute` method instead of E3 `<enabledWhen>` expressions

In all the cases above, the problem remain the same, toolbar items enabled state only changes when the focus change to another view regardless of the variable/data update.

Using `IEvaluationService.requestEvaluation("variable")` or `IEventBroker.send(UIEvents.REQUEST_ENABLEMENT_UPDATE_TOPIC, selector)` solves the problems but this is manual work.

A relevant question here is: what is the intended behavior of enabled the state? Should it get updated automatically when the data changes or not? If no, why was it working before the update?

Here is a simple project that illustrates the issue [sample.zip](https://github.com/user-attachments/files/27249292/sample.zip).

EdrilanBerisha:
this PR fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3953

when an AbstractSourceProvider fires fireSourceChanged(), the EvaluationService contextUpdater listener was updating the legacy expression context variable but never sending REQUEST_ENABLEMENT_UPDATE_TOPIC on the event broker.

as a result, ToolBarManagerRenderer (which listens for that topic) was never notified, so toolbar items relying on custom source providers with enabledWhen expressions stayed frozen until a focus change accidentally triggered an update

fix: send REQUEST_ENABLEMENT_UPDATE_TOPIC after each source change originating from a registered ISourceProvider (both single-variable and multi-variable variants). This makes the toolbar enablement update automatic and consistent with requestEvaluation() and the RAT updater, which already send this event.

added two regression tests to EvaluationServiceTest:
- testSourceProviderFiresEnablementUpdateEvent: verifies that a single-variable fireSourceChanged() sends REQUEST_ENABLEMENT_UPDATE_TOPIC
- testSourceProviderMultiVarFiresEnablementUpdateEvent: verifies the same for the multi-variable (Map) form of the source change

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/services/EvaluationService.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/services/EvaluationService.java
@@ -113,6 +113,10 @@ public final class EvaluationService implements IEvaluationService {
 			@Override
 			public void sourceChanged(int sourcePriority, String sourceName, Object sourceValue) {
 				changeVariable(sourceName, sourceValue);
+				// Notify toolbar/contribution items about the enablement state change
+				// so that items relying on custom source providers update automatically
+				// without requiring a focus change (bug 3953).
+				getEventBroker().send(UIEvents.REQUEST_ENABLEMENT_UPDATE_TOPIC, UIEvents.ALL_ELEMENT_ID);
 			}
 
 			@Override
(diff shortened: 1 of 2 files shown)


merks:
This needs to be a single commit rebased on master. Please force push a rebased commit.

EdrilanBerisha:
> This needs to be a single commit rebased on master. Please force push a rebased commit.

done

iloveeclipse:
Both added tests pass **without** the corresponding change in `EvaluationService`. Please double check & please if you update PR, rebase first on master.

vogella:
ratUpdater already sends REQUEST_ENABLEMENT_UPDATE_TOPIC/ALL_ELEMENT_ID on every re-run (EvaluationService.java:83), and HandlerProxy registers <enabledWhen> eagerly in its constructor via addEvaluationListener → addEvaluationReference, which puts username into ratVariables. changeVariable writes through legacyContext.addVariable → context.set(name, value), which is exactly what the RAT tracks with context.getActive(var).

That chain should already fire for your reported case. Adding a second unconditional send papers over whatever actually breaks it.

Can you explain why the ratUpdater is not triggered in your case?

Looking at your tests IEventBroker.send is synchronous so I don't think you are testing correctly (asychronous update failure).

PR #4075, opened 2026-06-09
Remove requirement of inlined o.e.jdt.launching.macosx fragment

HannesWell:
Part of
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/855


HannesWell:
Test failure is expected because the inlined code of the fragment is only available in the new target bundle after tonight's I-build.

PR #4084, opened 2026-06-11
Fix selection not updating when checking a tree item on Import window

tuanvtdeptre199:
I have found the issue when testing the Trace Compass in the Windows environment:
**Preproduce steps:**
1. Open the trace compass (My version is Version: 11.1.0) and try to import the archived trace package through **Trace Import** window:
2. I select the trace folder that contains the trace file in the trace package:

**Expected behaviour:**
It shows the trace file in the trace folder:
image

**Actual behaviour**
It does not show the trace file in the trace folder, and it also selects the wrong item:
image


tuanvtdeptre199:
Hi @jonahgraham -san, could you please help me review this code? Thank you!

PR #4092, opened 2026-06-12
Replace Batik with an internal hand-written CSS parser and drop SAC

vogella:
Replaces the Apache Batik CSS parser and the vendored W3C SAC layer with a small internal recursive-descent parser (tokenizer plus parser) scoped to the CSS subset the engine actually uses.
The engine parses through it end to end, the `org.apache.batik.css` Require-Bundle is removed, and the only remaining SAC type is the `LexicalUnit` interface that a later step drops.
Parsing in isolation measures about 22% faster and the engine sheds an external runtime dependency, with no change to styling behaviour: `@media`, `@font-face` and `@page` are still parsed and discarded as before.
Verified against the full css.core and css.swt test suites and by a differential comparison against the Batik output over all shipped Eclipse themes.

Contributes to #3980.


vogella:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/4091 should be merged before this one to catch potential regressions.

HeikoKlare:
I see a compilation error inside the IDE when having this change applied
image

vogella:
Thanks @HeikoKlare I have a look

vogella:
Will be fixed with https://github.com/eclipse-platform/eclipse.platform.ui/pull/4115

vogella:
Fixed @HeikoKlare sorry for the inconvenience

HeikoKlare:
Thank you and no problem. I just saw it when working on something else and decided to just post it here before others may run into the same situation.

PR #4094, opened 2026-06-13
Add method for providing disabled image descriptors

HeikoKlare:
Image descriptors are often shared and used with a `ResourceManager` to share an image created from the same descriptor. To do this with the disabled version of the image of a descriptor, a descriptor for the disabled image has to be managed in addition to the descriptor for the ordinary image. This is what is also done in shared workbench images where an image descriptor for the original and the disabled image are managed independently.

This change adds a method to an image descriptor for providing a shared image descriptor for the disabled version of the same image. This allows to only manage the original image's descriptor while also sharing the disabled version of that image. The disabled image descriptor is created lazy upon first access.
It is applied as a replacement at all places which used the `createWithFlags()` operation of an image descriptor for creating an image descriptor for the disabled version of the image to ensure that a shared image descriptor is used there.

As a further follow-up task, remaining usages of independently managed image descriptors for disabled image versions can be exchanged with usage of the new API.

This also contributes to:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3760

### Alternative Solutions

One alternative I have thought about is the extension of the `ResourceManager` to allow the creation and management of a disabled version of an image. But the `ResourceManager` interface is currently designed to just create/destroy/get a resource from a `DeviceResourceDescriptor` and all resource-type-specific methods (such as for creating images) are deprecated proposing the usage of the generic methods. The providing a disabled resource for a given ImageDescriptor would, however, require specific methods that allow to parameterize the request with the information about disablement. For that reason, I chose to not follow this approach but embed the functionality into the `ImageDescriptor` itself.

### Feedback

Like always, feedback is welcome, in particular since this will introduce new API at the `ImageDescriptor` class. In case you have other proposals for sharing the image descriptor for a disabled version of another image descriptor, please share it.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
@@ -20,6 +20,7 @@ import java.util.function.Supplier;
 
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.jface.util.Policy;
+import org.eclipse.swt.SWT;
 import org.eclipse.swt.SWTException;
 import org.eclipse.swt.graphics.Device;
 import org.eclipse.swt.graphics.Image;
@@ -67,6 +68,8 @@ public abstract class ImageDescriptor extends DeviceResourceDescriptor<Image> {
 	 * A small red square used to warn that an image cannot be created.
 	 */
 	protected static final ImageData DEFAULT_IMAGE_DATA = new ImageData(6, 6, 1, new PaletteData(new RGB(255, 0, 0)));
+
+	private ImageDescriptor disabledImageDescriptor;
 
 	/**
 	 * Constructs an image descriptor.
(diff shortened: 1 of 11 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/resource/ImageDescriptor.java
```diff
@@ -351,6 +347,22 @@ public Image createImage(Device device) {
...
+	/**
+	 * Returns an image descriptor representing a disabled version of the icon of
+	 * this descriptor. The descriptor is shared, i.e., calling this method multiple
+	 * times will return the same descriptor every time.
+	 *
+	 * @since 3.40
+	 * @return a shared image descriptor for a disabled version of the icon for this
+	 *         descriptor.
+	 */
+	public ImageDescriptor getDisabledDescriptor() {
```
laeubi: since there is no way to set it maybe it should be


```suggestion
	public ImageDescriptor asDisabledDescriptor() {
```

Also one might want to check if this is already a disabled one and then return `this`?

HeikoKlare: Thank you for this feedback. Both proposals make sense to me, so I have updated the PR accordingly. An image descriptor for a disabled image version is represented as `DerivedImageDescriptor` with the appropriate image flag set, which is why embedded the functionality to return `this` in case of calling `asDisabledDescriptor()` on such an instance to the `DerivedImageDescriptor`.

HeikoKlare:
I would like to merge this tomorrow evening to be able to proceed with follow-up tasks. If anyone has further opinions and concerns, I would appreciate it if you could share them by then.

laeubi (approved):
It looks sane to me.

HeikoKlare:
Failing test is known and documented in #80.

PR #4090, opened 2026-06-12
Skip Quick Access recompute on no-op resize

vogella:
During the Quick Access popup's initial layout, every shell resize re-triggered the proposal computation, and each recompute cancels the job still in flight. On Windows the native search field produces a burst of layout resizes, so the computation for freshly typed text kept getting cancelled and the result table was never populated, which is the intermittent failure of `testPreviousChoicesAvailableForExtension`.

A resize only affects results when it changes how many rows fit, so the dialog now records the row count it last computed for and skips the resize-triggered recompute when that count is unchanged, removing the cancellation path. This also drops the redundant recompute on cosmetic resizes during normal use. Finally it fixes an accidental ~83 minute timeout (`TIMEOUT * 1000`) in that test's dialog-init wait.

Fixes #4009

PR #4096, opened 2026-06-15
Restore bounded wait in QuickAccessDialogTest to avoid bundle timeout

vogella:
The dialog-initialization wait in `testPreviousChoicesAvailableForExtension` was set to `TIMEOUT * 1000`, roughly 83 minutes. When the dialog initialization stalls, which happens on the macOS CI runners, the wait blocks well past the 1200s tycho-surefire budget for the whole `org.eclipse.ui.tests` bundle, so the test runtime is killed with SIGTERM (exit code 143). Maven then reports a launch failure with no test results, which is why the macOS validation has been failing for every PR without pointing at a specific test.

Restoring the wait to `TIMEOUT` makes the test fail fast and locally when the dialog does not initialize, instead of hanging and taking the entire bundle down with it. The underlying race that leaves the proposals empty is addressed separately in #4090; this change only removes the accidental long wait so failures stay attributable and the bundle keeps reporting results. See #4009.

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/quickaccess/QuickAccessDialogTest.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/quickaccess/QuickAccessDialogTest.java
@@ -321,7 +321,7 @@ public class QuickAccessDialogTest {
 		 * wait for the dialog initialization, to avoid race conditions later on in the test, see:
 		 * https://github.com/eclipse-platform/eclipse.platform.ui/issues/4009
 		 */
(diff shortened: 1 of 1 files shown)


vogella:
@trancexpress the 83 min timeout which you places here, look like a typo to me and leads to process killing for the verification jobs. I suggest we change it back to the regular timeout to get a reasonable feedback (test errors) instead of a killed process.

trancexpress:
@vogella sorry! I must have committed a debug change without realizing it...

vogella:
> @vogella sorry! I must have committed a debug change without realizing it...

No worries, the timeout kill was just misleading for me, I assumed for a few days that we have an infrastructure problem.

PR #4097, opened 2026-06-15
Enhance EditingSupport/*CellEditor assertions

akurtakov:
When CellEditor used expects different object type, there is assertion message like:
```
org.eclipse.core.runtime.AssertionFailedException: assertion failed:
    at org.eclipse.core.runtime.Assert.isTrue(Assert.java:121)
    at org.eclipse.core.runtime.Assert.isTrue(Assert.java:106)
    at org.eclipse.jface.viewers.TextCellEditor.doSetValue(TextCellEditor.java:225)
    at org.eclipse.jface.viewers.CellEditor.setValue(CellEditor.java:853)
    at org.eclipse.jface.viewers.EditingSupport.initializeCellEditorValue(EditingSupport.java:102)
```

which is not really helpful in identifying the exact cause.
Enhance the assertions to give as much information as possible.


akurtakov:
Test failure unrelated. Merging.

PR #4105, opened 2026-06-15
Improve AGENTS.md: fix misleading and broken content

HeikoKlare:
## Summary

- Remove broken/malformed build command that concatenated two `mvn` invocations into one invalid command line
- Remove a `mvn clean test` example that doesn't work in Tycho projects (tests require `mvn verify`)
- Rename "Critical Limitation" section to "Building Individual Bundles" for clarity
- Add missing instruction on how to include changed source bundles alongside test bundles in the `-pl` parameter
- Remove every information about `-Pbuild-individual-bundles` as it's already defined via `.mvn/maven.config`
- Restructure sections: merge scattered Testing/Common Development Commands content into coherent Build System and Development Hints sections; move "Finding test bundles" and JUnit guidelines next to related content
- Remove incorrect statement that `Co-Authored-By` trailers fail the Eclipse license check

## Background

A key insight driving these changes: when running tests, agents were not including bundles containing modified classes in the `-pl` parameter alongside the test bundle. This caused tests to run against stale compiled code rather than the changes under test. The instructions now explicitly document the correct pattern:

```bash
mvn clean verify -pl :org.eclipse.jface,:org.eclipse.jface.tests
```

## Note

These changes were developed together with a Claude Code agent to verify that the instructions are actually useful and unambiguous from an agent's perspective.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,19 +61,15 @@ Each bundle contains:
 - `src/` or `eclipseui/` - Java source code
 - `.settings/` - Eclipse compiler settings
 
+
 ## Build System
 
-### Critical Limitation
-
-
-Use the `-Pbuild-individual-bundles` profile:
+### Building Individual Bundles
 
 ```bash
-# Compile single bundle
-mvn clean compile -pl :bundle-artifact-id -Pbuild-individual-bundles -q
-
-# Example for building a single bundle
-mvn clean verify -Pbuild-individual-bundles mvn clean verify -pl bundles/org.eclipse.ui -DskipTests
+# Compile single bundle such as org.eclipse.jface
+mvn clean compile -pl :org.eclipse.jface
+```
 
 ### Test Properties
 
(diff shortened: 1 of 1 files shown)


vogella:
Minor nits:

- Grammar: "their bundles needs to be added" → "their bundles need to be added."
- The placeholder-ization (:TEST_BUNDLE_NAME instead of org.eclipse.ui.tests) is a wash. You wisely kept inline org.eclipse.jface.tests examples, so the teaching value survives; concrete-only would have been fine too.

HeikoKlare:
Thank you for the hints. Both are fully valid. I have updated the PR accordingly, in particular removing the abstract command pattern and just preserving the examples (with proper prose where necessary).

vogella:
Thanks, especially the working examples for the build are important, abstraction here is only hurtful.

akurtakov:
I am not sure whether it is helpful or just adding noise but with https://github.com/eclipse-platform/eclipse.platform.ui/blob/b268df2c7ccc0324d3c67916c4e76a749055ce00/.mvn/maven.config#L1 -Pbuild-individual-bundles  parameter to most commands shouldn't be needed.

HeikoKlare:
> I am not sure whether it is helpful or just adding noise but with
> 
> https://github.com/eclipse-platform/eclipse.platform.ui/blob/b268df2c7ccc0324d3c67916c4e76a749055ce00/.mvn/maven.config#L1
> 
> -Pbuild-individual-bundles parameter to most commands shouldn't be needed.

Great hint, thank you. I wasn't aware of that configuration. I can confirm that passing the profile to maven commands is completely unnecessary. I have removed every information about that from the agents.md.

PR #4106, opened 2026-06-15
Run addons SWT tests via the AllTests suite in the Tycho build

vogella:
Instead of relying on
Tycho's class-name autodiscovery, use the standard Tycho approach via
build.properties.

Reintroduce the AllTests @Suite and point Tycho at it through the
pom.model.property.testClass build property. This disables the
additional
name-based scan, so each test class is executed exactly once via the
suite.

The restored suite also includes StackDropAgentTest, which the previous
AllTests suite had omitted, so every test class in the bundle is now
part
of the suite.

### DIFF
--- a/tests/org.eclipse.e4.ui.workbench.addons.swt.test/src/org/eclipse/e4/ui/workbench/addons/AllTests.java
+++ b/tests/org.eclipse.e4.ui.workbench.addons.swt.test/src/org/eclipse/e4/ui/workbench/addons/AllTests.java
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Heiko Klare - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.e4.ui.workbench.addons;
+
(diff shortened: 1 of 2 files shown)


tests/org.eclipse.e4.ui.workbench.addons.swt.test/src/org/eclipse/e4/ui/workbench/addons/AllTests.java
```diff
@@ -0,0 +1,21 @@
+package org.eclipse.e4.ui.workbench.addons;
```
akurtakov, line 14: Please add license header.

vogella: Funny, that was also missing there before. But makes sense to add it now. I will update the PR

akurtakov: I noticed that too. Still it makes sense to fix old problems.

vogella:
@HeikoKlare from Git history it looks like you created the suite in 79baeed7ac (without licence header), so I attributed it to you. If that is wrong or should be changed, please let me know.

HeikoKlare:
I did not create that suite but just moved it in the referenced commit https://github.com/eclipse-platform/eclipse.platform.ui/commit/79baeed7acfd63d782eb09ad1d11134005aaf31c. The suite was originally introduced in commit https://github.com/eclipse-platform/eclipse.platform.ui/commit/4291d9c3407e32555953e62905976d8a2fac58fd.

vogella:
So want should I put into the copyright header?

akurtakov:
I am merging this one as I don't care about archeology and this is good enough for me.

PR #4112, opened 2026-06-17
Remove DOM mirror rule classes and the dead CSS2Properties facade

vogella:
Removes two layers left dead by the internal-parser cutover in #4092: the W3C DOM mirror rule classes the hand-written parser no longer produces (CSSMediaRuleImpl, CSSPageRuleImpl, CSSFontFaceRuleImpl, CSSUnknownRuleImpl, CounterImpl, RectImpl, CSSPropertyListImpl), and the CSS2Properties facade behind CSSStylableElement.getStyle(), a ~1,300 line W3C compatibility shim that no workbench code calls. About 1,750 lines of internal (x-internal/x-friends) code go with no behavioural change.

It is the first slice of the DOM-mirror replacement and sits on top of #4092, so it stays in draft until that merges and will be rebased onto master then.

Contributes to #3980.


vogella:
I now compile the aggregator locally with this change to ensure that this does not break the PDE CSS spy

vogella:
PDE CSS spy builds fine with this change.

PR #4114, opened 2026-06-18
Increase surefire timeout for the org.eclipse.ui.tests bundle

vogella:
The org.eclipse.ui.tests suite runs in about 7 minutes on Linux but roughly three times slower on the GitHub macOS runners, where it exceeds the inherited 1200s surefire.timeout. The forked test VM is then killed (exit 143) while the suite is almost finished, which currently shows up as a recurring macOS CI failure on master unrelated to any individual test.

This raises surefire.timeout to 1800s for just this bundle via the Tycho pomless pom.model.property mechanism in build.properties, giving the slower macOS runner enough headroom to complete.


vogella:
Merging to prevent other PR validation to time out as for example https://github.com/eclipse-platform/eclipse.platform.ui/pull/4113

PR #4116, opened 2026-06-18
Remove OpenCloseTest from UiTestSuite

### LINKED ISSUE #4101: OpenCloseTest make validation very slow
PR validations on macOS sometimes fail with a killed process. OpenCloseTest hangs in testOpenClosePerspective: it prints its tear down marker and then produces no output for roughly 8 to 9 minutes, until the 1200 second per bundle timeout kills the whole JVM. In a healthy macOS run the whole test takes about 146 seconds (see the passing build in #4098), so this is a real stall and not just slowness. Linux and Windows are not affected.

We should make the test dump all thread stacks and fail fast when a method exceeds a deadline, instead of relying on the bundle timeout. A plain JUnit timeout will not work because these tests run on the UI thread, so this needs a watchdog thread that dumps Thread.getAllStackTraces() before aborting.

vogella:
OpenCloseTest is a stress test that repeatedly opens and closes editors, views, perspectives, workbench windows and the intro part. On the macOS CI runners it is the single most expensive test (about 290s, roughly 29% of the org.eclipse.ui.tests suite) and it intermittently stalls in testOpenClosePerspective for 8 to 9 minutes, killing the forked test VM through the per-bundle timeout. The watchdog added earlier only turns that stall into a fast failure; it removes neither the cost nor the flakiness.

Why this is safe: the open/close cycles it exercises are already covered functionally and for leaks elsewhere, for example LeakTests (testSimpleEditorLeak, testSimpleViewLeak, testSimpleWindowLeak), IWorkbenchPageTest, the zoom ShowView and OpenEditor tests and the intro tests. Removing it from UiTestSuite therefore loses no real coverage while reclaiming a large, variable chunk of macOS validation time. The test class is kept so it can still be run manually for stress testing.

Fixes #4101

### DIFF
--- a/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/UiTestSuite.java
+++ b/tests/org.eclipse.ui.tests/Eclipse UI Tests/org/eclipse/ui/tests/UiTestSuite.java
@@ -54,7 +54,6 @@ import org.eclipse.ui.tests.quickaccess.QuickAccessTestSuite;
 import org.eclipse.ui.tests.releng.PluginActivationTests;
 import org.eclipse.ui.tests.services.ServicesTestSuite;
 import org.eclipse.ui.tests.statushandlers.StatusHandlingTestSuite;
-import org.eclipse.ui.tests.stress.OpenCloseTest;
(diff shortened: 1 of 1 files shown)


vogella:
Again a time out https://github.com/eclipse-platform/eclipse.platform.ui/pull/4117/checks, merging this now

PR #4118, opened 2026-06-18
Unify JFace popup dialog menu icon with workbench view menu

### LINKED ISSUE #4089: Consistent icons for drop-downs
Recently @HeikoKlare and I discussed that our UI should be consistent. While start to work on the flaky QuickAccessTests I noticed that our drop-down icon in QuickAccess looks bad.

Image

Other icons use the kebap menu.

Image#

In isolation:

Image

Image

I suggest we use the kebap menu consistently.

vogella:
JFace `PopupDialog` rendered its top-right menu button with the old blue gradient chevron, while workbench view menus already use the modern three vertical dots glyph. This replaces the JFace `popup_menu` artwork (both the enabled and disabled variants) with copies of the workbench `view_menu` three-dots design, so the CTRL+3 Find Actions dialog and any other PopupDialog with a dialog menu now match nearby view menus. The disabled icon is a greyed version of the same glyph using the existing disabled palette. The registration key `popup_menu_image` is unchanged, so there is no API impact, and the third JFace-local copy follows the existing pattern since JFace cannot depend on the workbench bundle.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/4089


BeckerWdf:
Can you show screenshots how this now looks like in context?

vogella:
image

PR #4120, opened 2026-06-19
Migrate CSS value consumers to the internal record model

vogella:
Migrates the property handlers, converters and SWT helpers to read CSS values directly from the internal `CssValues` records (pattern matching on `CssNumber`, `CssDimension`, `CssColor`, ...) instead of going through the W3C `CSSValue` accessor methods.
With every consumer moved over, the bridge types that only existed to feed them (`CSS2PrimitiveValueImpl`, `CSS2RGBColorImpl`) are removed.

Behaviour is unchanged: the same values reach the same setters, just by a more direct route.

Stacked on #4117 (the value records this consumes), so the diff currently also shows that commit; it reduces to just the consumer migration once #4117 is merged and this branch is rebased.
Part of the CSS engine rework (#3980).


vogella:
1 out of 3 runs failed: testPreviousChoicesAvailableForExtension (org.eclipse.ui.tests.quickaccess.QuickAccessDialogTest) is known to be flaky

PR #4121, opened 2026-06-19
Add button to cancel job family to the JobsView (Part of the JFace's examples)

fedejeanne:
The `JobsView` can be opened with `Ctrl+3` -> "Job Factory"

Also change some comments that were copy-pasted and were wrong.

### DIFF
--- a/examples/org.eclipse.e4.ui.examples.job/src/org/eclipse/e4/ui/examples/jobs/views/JobsView.java
+++ b/examples/org.eclipse.e4.ui.examples.job/src/org/eclipse/e4/ui/examples/jobs/views/JobsView.java
@@ -247,7 +247,19 @@ public class JobsView {
 			}
 		});
 
-		// join the running test jobs
+		// Cancel the job family (all TestJobs have the same family)
+		Button cancelFamily = new Button(group, SWT.PUSH);
+		cancelFamily.setText("Cancel Job family"); //$NON-NLS-1$
+		cancelFamily.setToolTipText("IJobManager.cancel() on test jobs"); //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


fedejeanne:
Failing tests are unrelated:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/1737
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/80

examples/org.eclipse.e4.ui.examples.job/src/org/eclipse/e4/ui/examples/jobs/views/JobsView.java
```diff
@@ -540,6 +552,14 @@ protected void joinTestJobs() {
 		}
 	}
 
+	private void cancelTestJobs() {
+		try {
+			progressService.busyCursorWhile(__ -> Job.getJobManager().cancel(TestJob.FAMILY_TEST_JOB));
```
r-mennig: Even if the monitor is not used, I'd say it's better to still give the variable a useful name here (so a reader can easily see what's unused here)

```suggestion
			progressService.busyCursorWhile(monitor -> Job.getJobManager().cancel(TestJob.FAMILY_TEST_JOB));
```

r-mennig (approved):

examples/org.eclipse.e4.ui.examples.job/src/org/eclipse/e4/ui/examples/jobs/views/JobsView.java
fedejeanne: Good catch, and I had also called it `monitor` in the other class :-)

Done ✔️

fedejeanne:
I interrupted the builds since this is a change in an example, so the tests do not make sense. 🥬 💚

PR #4122, opened 2026-06-19
Replace the W3C computed-style cascade with internal types

vogella:
Folds the W3C ViewCSS / DocumentCSS computed-style machinery into the engine: `CSSEngineImpl` now holds the stylesheet list and a cached flat rule list directly, and the former `ViewCSSImpl.getComputedStyle` logic becomes `CSSEngine.computeStyle`.
`CSSStyleSheetImpl` becomes a plain list of `CssRule` (a new sealed interface), and `ViewCSSImpl`, `DocumentCSSImpl`, `StyleSheetListImpl`, `MediaListImpl` and the related W3C rule-mirror classes are deleted.

`CSSStyleDeclarationImpl` keeps implementing the W3C `CSSStyleDeclaration` interface, since `IStylingEngine.getStyle` / `IThemeEngine.getStyle` are frozen public API returning that type. A deprecated `getViewCSS()` default method remains as a binary-compatibility bridge.

Stacked on #4120 (the consumer migration it builds on), so the diff currently also shows that commit; it reduces to just this change once #4120 is merged and this branch is rebased.
Part of the CSS engine rework (#3980).


vogella:
This one requires to sync changes on the PDE side, so I plan to merge it early M2 together with the PDE changes.

PR #4131, opened 2026-06-22
Defer short-running jobs from the progress view

### LINKED ISSUE #4126: Progress monitor should not flash frequently with short term entries which cannot be read
If show system jobs is activated in the progress view, the progress view flashes regular (most frequently IOScheduler or something like that, to fast to read). This is not helpful as the entries cannot be read and the UI is constantly flickering. 

I suggest to surpress updates less than 300 ms or something similar that should remove most of the short term jobs.

vogella:
The progress view added a job the moment it was scheduled and removed it when it finished, so jobs that complete within a few hundred milliseconds flashed in and out and could not be read (very noticeable with system jobs enabled).

This adds a per-job grace period before a job is announced to the progress listeners. Jobs that finish, sleep or are cancelled before the grace period elapses are never shown, which removes the flicker for refresh, validation and decoration bursts. The grace period defaults to half of `getLongOperationTime()` with a 200ms floor and can be overridden through the `org.eclipse.ui.progress.viewGracePeriod` system property. Jobs kept after completion (KEEP_PROPERTY or an error result) are still captured, and running jobs remain queryable through `getJobInfos()`.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/4126

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
@@ -38,6 +38,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -175,6 +176,23 @@ public class ProgressManager extends ProgressProvider implements IProgressServic
 	private static final String IMAGE_KEY = "org.eclipse.ui.progress.images"; //$NON-NLS-1$
 
 	private final Throttler uiRefreshThrottler;
+
+	/**
+	 * System property overriding, in milliseconds, the grace period a job must run
+	 * before it is shown in the progress view; {@code 0} disables it and restores
+	 * the immediate display of jobs.
+	 */
+	public static final String GRACE_PERIOD_PROPERTY = "org.eclipse.ui.progress.viewGracePeriod"; //$NON-NLS-1$
+
+	private static final int MINIMUM_GRACE_PERIOD = 200;
+
+	/**
+	 * Scheduled jobs not yet announced, mapped to the {@link System#nanoTime()} at
+	 * which their grace period elapses. Jobs that finish, sleep or are cancelled
+	 * before then are removed here and never appear, which avoids flicker from very
+	 * short running jobs.
+	 */
+	private final Map<Job, Long> pendingJobAdditions = new ConcurrentHashMap<>();
 
 	/**
 	 * Returns the progress manager currently in use.
@@ -360,6 +378,7 @@ public class ProgressManager extends ProgressProvider implements IProgressServic
 	 * Send pending notifications to listeners.
 	 */
 	/* Visible for testing */ public void notifyListeners() {
+		promotePendingJobs();
 		Set<GroupInfo> localPendingGroupUpdates, localPendingGroupRemoval;
 		Map<JobInfo, Set<IJobProgressManagerListener>> localPendingJobUpdates, localPendingJobAddition,
 				localPendingJobRemoval;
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
```diff
@@ -704,14 +747,85 @@ public void addJobInfo(JobInfo info) {
...
+		Runnable graceTimer = new Runnable() {
+			@Override
+			public void run() {
+				if (pendingJobAdditions.remove(job, this)) {
+					notifyJobAdded(info);
+				}
+			}
+		};
+		pendingJobAdditions.put(job, graceTimer);
+		display.asyncExec(() -> {
```
iloveeclipse: This can flood UI thread with asyncExec / timerExec tasks for short living jobs. I would not misuse UI thread to dispatch tasks just to make sure progress view don't show jobs.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
```diff
@@ -704,14 +747,85 @@ public void addJobInfo(JobInfo info) {
...
+	 * @param job the job whose pending add should be cancelled
+	 * @return {@code true} if an add was pending and got cancelled
+	 */
+	private boolean cancelPendingAdd(Job job) {
+		Runnable pending = pendingJobAdditions.remove(job);
+		if (pending == null) {
+			return false;
+		}
+		if (display != null && !display.isDisposed()) {
+			display.asyncExec(() -> display.timerExec(-1, pending));
```
iloveeclipse: Same here

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
vogella: Reworked to avoid per-job UI dispatch: `pendingJobAdditions` is now a `Map<Job, Long>` of grace deadlines instead of per-job timers. Scheduling only puts the deadline in the map; `notifyListeners()` (the throttler's coalesced UI pass) drains and announces the expired entries, re-arming once via `throttledAsyncExec()` while any remain. Cancelling is a plain `ConcurrentHashMap.remove`, no dispatch.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
vogella: Same fix: cancelling is now just `pendingJobAdditions.remove(job)`, no `asyncExec`/`timerExec`.

vogella:
@iloveeclipse any additional feedback?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
```diff
@@ -395,6 +419,36 @@ public void setBlocked(IStatus reason) {
...
+	 * from {@link #pendingJobAdditions} and are therefore never announced.
+	 */
+	private void promotePendingJobs() {
+		if (pendingJobAdditions.isEmpty()) {
+			return;
+		}
+		long now = System.nanoTime();
+		for (Iterator<Entry<Job, Long>> it = pendingJobAdditions.entrySet().iterator(); it.hasNext();) {
+			Entry<Job, Long> entry = it.next();
+			if (now - entry.getValue().longValue() < 0) {
```
merks: Would this be auto-unwapped without the longValue() call?

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
```diff
@@ -704,12 +773,63 @@ public void addJobInfo(JobInfo info) {
...
+
+		int gracePeriod = getViewGracePeriod();
+		if (gracePeriod <= 0) {
+			rememberJobAddition(info);
+		} else {
+			// Defer announcing the job. It is announced from notifyListeners once
+			// its grace period elapses. Jobs that finish, sleep or are cancelled
+			// before that are removed from the map and never shown, which avoids
+			// flicker from very short running jobs.
+			pendingJobAdditions.put(job, Long.valueOf(System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(gracePeriod)));
```
merks: Would this be auto wrapped without the valueOf call?

merks (commented):
Looks okay.  Just nit questions.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
vogella: yes, will update PR

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/progress/ProgressManager.java
vogella: yes, will update PR

vogella:
@merks I assume M1 is already out or at least finished. Please let me know if that is wrong.

merks:
Yes, the Platform has contributed M1 so M2 is started and we have some breathing room again.  Thanks for asking.

Speaking of which, the following issue is why I wanted to complete the Platform target platform removals for M1 because I was pretty sure this problem would happen:

https://github.com/eclipse-simrel/simrel.build/issues/1432

Doing this for M1 gives much more runway for projects to address it., so I'm glad about that.

PR #4133, opened 2026-06-22
Fix configuration of the unsaved tab dirty indicator

### LINKED ISSUE #4130: Unsaved tab decoration configuration is broken
Two problems from https://github.com/eclipse-platform/eclipse.platform.ui/commit/233d793a48ced37549916097969e9c80b49f83e4 & https://github.com/eclipse-platform/eclipse.platform.ui/pull/4070

1) If CSS themes are disabled, there is no way to switch off the new default for the unsaved tab decoration, the preference is simply missing on page:

Image

Expected to see the following option below:

Image

2) The default preference value for the new tab decoration can't be customized via product customization file. The `org.eclipse.e4.ui.workbench.renderers.swt/SHOW_DIRTY_INDICATOR_ON_TABS=false` preference line is ignored.

This is inconvenient as the user can't switch back to old behavior and also breaks our internal tests for now because we can't even switch it off via product customization.

vogella:
The dirty indicator preference was only shown and stored while CSS theming was active, so it could not be turned off when Eclipse was started with `-cssTheme none` or in high-contrast mode. It is now created, stored and reset next to the "Show most recently used tabs" option and is therefore always available.

The renderer also read the value from the instance scope only, which ignored a default contributed through product customization (`SHOW_DIRTY_INDICATOR_ON_TABS` in `plugin_customization.ini`). It now reads via the preferences service so the default scope is honored, allowing products to restore the old asterisk behavior.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/4130

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
+++ b/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
@@ -34,6 +34,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.e4.core.contexts.IEclipseContext;
 import org.eclipse.e4.core.di.annotations.Optional;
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/StackRenderer.java
```diff
@@ -941,8 +942,10 @@ private boolean getMRUValueFromPreferences() {
 	}
 
 	private boolean getShowDirtyIndicatorForTabsFromPreferences() {
-		return preferences.getBoolean(CTabRendering.SHOW_DIRTY_INDICATOR_ON_TABS,
-				CTabRendering.SHOW_DIRTY_INDICATOR_ON_TABS_DEFAULT);
+		// Use the preferences service so product customization (default scope) is honored
+		return Platform.getPreferencesService().getBoolean(
```
iloveeclipse, line 946: Note, 5 lines above there is yet another wrong access to `preferences` to read the MRU value.

vogella: Agree, but unrelated to this change, please fix via separate PR, I don't want to mix unrelated things in this change.

iloveeclipse (approved):
LGTM

vogella:
Also ready from my side, feel free to merge.

PR #4134, opened 2026-06-22
Fix Ctrl+F7 view switcher showing same name for multi-instance views

### LINKED ISSUE #2774: Terminal views all show same name 'Terminal' in Ctrl+F7 view switcher popup
## Problem

When multiple Terminal views are open and you use **Ctrl+F7** to cycle through open views, the view switcher popup shows all terminal entries with the same name **"Terminal"**. This makes it impossible to distinguish which terminal to switch to.

Image

The Terminal tabs/views have the correct name: 

Image

## Expected Behavior

Each Terminal view should have a unique name in the popup, e.g. **"Terminal 1"**, **"Terminal 2"**, **"Terminal 3"**, matching the tab label shown in the view itself.

## Steps to Reproduce

1. Open 2 or more Terminal views (e.g. via `Window > Show View > Terminal` multiple times)
2. Press **Ctrl+F7** to open the view switcher popup
3. Observe that all terminal entries are listed as **"Terminal"** — no numbering to differentiate them

## Environment

- **View class:** `TerminalsView`
- **Contributing plug-in:** `org.eclipse.terminal.view.ui` (1.0.100.v20260410-1230)
- **View identifier:** `org.eclipse.terminal.view.ui.TerminalsView`
-   Terminal Feature	1.0.100.v20260410-1244	org.eclipse.terminal.feature.feature.group	Eclipse.org
- Eclipse Version: 2025-12 (4.38)

## Additional Context

The tab labels in the view stack itself correctly show numbered terminals (e.g. "Terminal 1", "Terminal 2"), but the `getPartName()` or equivalent method used for the Ctrl+F7 switcher does not reflect this numbering.

A fix would likely involve ensuring the view's `getPartName()` (or the secondary ID / content description) returns the numbered name so the switcher can display it distinctly.

Philipp0205:
**Change made:** Removed the `getPartName()` override from `ViewReference.java`.

Views like Terminal call `setPartName("Terminal " + secondaryId)` which triggers `CompatibilityPart` to update the E4 model via `part.setLabel(computeLabel())`. The parent class `WorkbenchPartReference.getPartName()` returns `part.getLocalizedLabel()` which reflects this dynamic name (e.g. Terminal 1, Terminal 2 instead of Terminal Terminal).

The removed override was bypassing this by always returning the static plugin.xml descriptor label. Generic fix: Works for all multi-instance views (Terminal, Console, Search).

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2774

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/ViewReference.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/ViewReference.java
@@ -69,11 +69,6 @@ public class ViewReference extends WorkbenchPartReference implements IViewRefere
 	}
 
 	@Override
-	public String getPartName() {
-		return descriptor == null ? "" : descriptor.getLabel(); //$NON-NLS-1$
-	}
-
-	@Override
 	public String getSecondaryId() {
 		MPart part = getModel();
 


iloveeclipse:
@Philipp0205 : I've rebased your branch on latest master state. Your state was very old.

Philipp0205:
@iloveeclipse thanks, I forgot to update my fork :upside_down_face:

iloveeclipse:
This works for Terminal views but doesn't work for Console / Search views - they also show exact same names in the tabs. I guess the way how Terminal (coming originally from CDT) manages its name differs from "regular" multi-instance views which were developed in the Platform.

Ideally we should investigate what is the difference and provide a fix that works consistently for all views.

Philipp0205:
okay, apparently I did not test this good enough. I also noticed that console and search views do not add numbers to their tab names if opened multiple times.

Let me have  another look at the other views.

ravnskjaer:
I have investigated this a bit further and I found that the problem can be reproduced for all these views:
- Problems
- Tasks
- Terminal
- Search
- Console

As you pointed out @Philipp0205 , the Search and Console views does not add numbers to the tabs. Your fix does however work for those too. I tried adding similar logic (based on _secondaryId_ if present as in TerminalsView::createPartControl) for updating the _partName_ in ConsoleView and SearchView, and then I verified with your fix.

I suggest that your fix should be accepted as it is and separate issues created if we want the names on the actual tab names  to be changed for Search and Console views. I consider it to be different, although related topics. 

I am new to this community so bear with me if I am doing this in the wrong way :)

PR #4137, opened 2026-06-25
Inherit host control colors in content proposal popups

vogella:
Content proposal popups now fall back to the host control's colors when
the JFace `CONTENT_ASSIST_*` preferences are unset, instead of defaulting
to white.

RCP applications that style their controls via SWT API get matching
popups for free, and the white completion popup in the dark workspace
selection dialog is fixed.


vogella:
This is just a fallback scenario in case the color registry is not yet set, so all regular behavior should be unchanged.

PR #4139, opened 2026-06-26
Retire the stale SAC dependency from the css.swt test bundle

vogella:
After the CSS value-record migration the production bundles no longer reference SAC. The only remaining `org.w3c.css.sac` trace was the `tkuiTestsToRefactor` legacy tests in `org.eclipse.e4.ui.tests.css.swt` plus a matching `Import-Package` in its manifest. That folder is not on the build path (only `src/` is compiled), so it never ran as tests.

This deletes the folder and drops the SAC `Import-Package`, which lets the `org.w3c.css.sac` library leave the target platform once no bundle outside this repo needs it. Independent of the cascade work in #4122.


vogella:
Before and during my work on the CSS engine I added lots of tests to ensure that I do not break existing functionality. These tests are outdated and test old API so I remove them. 

These are reference tests dead since **2008 (since 17 years)** only touched recently by my global "update color constructors" sweep.

PR #4141, opened 2026-06-26
Add JFace snippet for a notification popup that follows its shell

vogella:
Adds Snippet086 to the JFace snippets, showing how to keep a NotificationPopup anchored to a top-level shell while the window is moved or resized. The stock NotificationPopup only positions itself once when it opens; the snippet installs a ControlListener that re-anchors the popup to the shell's bottom-right corner on every move and resize, and removes the listener again when the popup closes. The JFaceSnippets documentation page gets a matching entry under the Notification section.

### DIFF
--- a/docs/JFaceSnippets.md
+++ b/docs/JFaceSnippets.md
@@ -67,6 +67,10 @@ Demonstrates a `NotificationPopup` with interactive content — a label and a bu
 
 ![Snippet085.png](images/Snippet085NotificationPopupWithUserInteraction.png)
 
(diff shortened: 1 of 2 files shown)


laeubi:
> The stock NotificationPopup only positions itself once when it opens

So if it is desired behavior that it keeps at that position maybe the `NotificationPopup` better is enhanced than letting users install custom code?

vogella:
> > The stock NotificationPopup only positions itself once when it opens
> 
> So if it is desired behavior that it keeps at that position maybe the `NotificationPopup` better is enhanced than letting users install custom code?

IIRC from reading the user guides during the notification API migration from Mylyn to platform the "desired" behavior is not to move the notification so I would leave it as an example.

PR #4146, opened 2026-06-26
Decompose checkInstanceLocation into focused helper methods

vogella:
Refactors the long `checkInstanceLocation` method into a short orchestrator by extracting the already-set lock handling (`lockExistingWorkspace`), the prompt-retry loop (`promptAndSetWorkspace`), and the inline "workspace in use" dialog (`openWorkspaceInUseDialog`). Pure extraction, no behavior change. Stacked on #4140 (the two overlap in the same method), so this currently includes that commit as well; kept as a draft until #4140 merges, after which it rebases down to a single-commit extraction diff.


iloveeclipse:
Please keep all methods extracted from protected methods also protected, so they could be overriden if needed.

vogella:
> Please keep all methods extracted from protected methods also protected, so they could be overriden if needed.

done

PR #4147, opened 2026-06-30
Apply the new popup menu icon to the loaded JFace icons/full copy

vogella:
`JFaceResources` registers the `PopupDialog` menu icon (`POPUP_IMG_MENU`) with `icons/full/popup_menu.svg` as the primary resource and only falls back to the class-relative `dialogs/images` copy when JFace runs outside OSGi. PR #4118 updated only the `dialogs/images` copy, so inside a running Eclipse the Quick Access dialog and other `PopupDialog`s still showed the old blue chevron. This copies the new three-dots artwork (enabled and disabled, in png, @2x and svg) into `icons/full` so the icon is actually visible in the IDE. This completes the change started in #4118 for issue #4089.


vogella:
Test failure is unrelated.

PR #4149, opened 2026-07-01
Keep OS high contrast colors instead of forcing the default theme

### LINKED ISSUE #3095: Retire High Contrast Theme
The High Contrast theme does look horrible (on Linxu). Maybe we should retire it?

Image

cc @BeckerWdf

vogella:
When the OS is in high contrast mode Eclipse requests the high-contrast theme, but that theme was removed in the 2025-09 release. Since #3854 the theme engine then forced the default light theme, and a product or workspace default could override the selection, so the native OS high-contrast colors were lost and the UI became mostly white and unusable. This keeps the OS colors when the requested theme is missing while the OS is in high contrast mode, and no longer lets a product or workspace default override the high-contrast theme. The fallback to the default theme stays unchanged outside high contrast mode.

Fixes #3095

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
+++ b/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
@@ -430,7 +430,12 @@ public class ThemeEngine implements IThemeEngine {
 				return;
 			}
 		}
-		// Theme not found (e.g. it was uninstalled); fall back to default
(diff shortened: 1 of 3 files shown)


vogella:
No auto-commit for the version increase, might be temporary not working. I add one manually.

HeikoKlare (approved):
The change looks sound to me and it works fine on Windows (i.e., restores the high contrast behavior from 2026-03 and below):
Image

Thank you for the immediate fix.

vogella:
> Thank you for the immediate fix.

Based on your analysis @HeikoKlare the fix was trivial to implement. Thank you for that. I plan to merge this tomorrow (if build is green and no one objects).

PR #4152, opened 2026-07-03
Force qualifier changes for o.e.ui.views and o.e.ui.workbench

merks:


merks:
Given we need a fixed M1 build, and these are only forced qualifier changes, I'll merge immediately.

vogella:
Missing version bump https://github.com/eclipse-platform/eclipse.platform.ui/pull/4157

PR #4156, opened 2026-07-03
Cap Quick Access providers at a fair share of the visible slots

### LINKED ISSUE #4155: Quick Search Behaves Poorly
I noticed immediately that the Quick Search behaves poorly after updating my SDK IDE just now.   I'm used to type "Error" to open the error log and now it's really not so good for that anymore.  I've never actually used it to open files but now it's lists lots of files.

It's a bit hard to make a video...

https://github.com/user-attachments/assets/cbe47373-4fb4-4b82-80ca-342842947a71

@vogella 

I imagine this is a regression from the many changes I saw related to the implementation details.

It feels bad to ship M1 like this...

vogella:
Since the relevance-ranking rework, the limited Quick Access table is filled with the globally highest-scored entries, so a single prolific provider can flood out every other category. Typing "Error" in an SDK workspace listed lots of workspace files instead of the Error Log view, because short file names outscore the category-suffixed view label. This keeps the relevance ranking but caps each provider at its fair share of the visible slots; slots a provider does not use go to the best remaining entries. That restores the provider diversity of the previous round-robin fill while relevance still orders entries within each category. The selection helper is pure and covered by new unit tests in QuickAccessMatchingTest, and QuickAccessDialogTest passes unchanged.

Fixes #4155

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/quickaccess/QuickAccessContents.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/quickaccess/QuickAccessContents.java
@@ -574,8 +574,7 @@ public abstract class QuickAccessContents {
 			}
 		} else {
 			int numberOfSlotsLeft = perfectMatch != null ? maxNumberOfItemsInTable - 1 : maxNumberOfItemsInTable;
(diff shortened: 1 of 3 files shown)


iloveeclipse:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/4157 was merged, I've just rebased.

PR #4163, opened 2026-07-04
Index style rules by their rightmost simple selector

vogella:
computeStyle tested every selector alternative of every rule against every element, the main cost of a theme swap. A new RuleIndex buckets each alternative by the id, class, or element type its rightmost compound requires (universal, attribute-only and pseudo-only alternatives go into a remainder bucket consulted for every element), and the matcher now runs only on the candidates an element can possibly match, in unchanged cascade order. Styling results are identical: match positions are preserved, and the new RuleIndexTest locks the invariant that the index never drops an alternative the matcher would accept, checked over a stylesheet covering every selector shape.

Measured with a workbench theme-swap benchmark (20 editors, about 4000 widgets, 5 warmup and 10 measured swap rounds per direction), engine applyStyles time falls from 177.4 to 155.8 microseconds per call and from 3018 to 2651 ms per swap, about 12 percent less styling work per theme change. The applyStyles call count is identical at 17013 per swap, which confirms the styling behaviour is unchanged.

An earlier revision of this description quoted 21 percent. That figure was measured against a baseline predating #4160, and master has since become about 20 percent faster on this benchmark on its own, so the remaining gain from rule indexing is smaller. The numbers above were re-measured against current master.

Follow-up to #3980.

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/engine/CSSEngineImpl.java
+++ b/bundles/org.eclipse.e4.ui.css.core/src/org/eclipse/e4/ui/css/core/impl/engine/CSSEngineImpl.java
@@ -61,6 +61,7 @@ import org.eclipse.e4.ui.css.core.impl.dom.CSSStyleRuleImpl;
 import org.eclipse.e4.ui.css.core.impl.dom.CSSStyleSheetImpl;
 import org.eclipse.e4.ui.css.core.impl.dom.CssRule;
 import org.eclipse.e4.ui.css.core.impl.dom.StyleWrapper;
+import org.eclipse.e4.ui.css.core.impl.engine.selector.RuleIndex;
 import org.eclipse.e4.ui.css.core.impl.engine.selector.SelectorMatcher;
 import org.eclipse.e4.ui.css.core.impl.engine.selector.Selectors;
 import org.eclipse.e4.ui.css.core.resources.IResourcesRegistry;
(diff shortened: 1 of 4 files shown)


vogella:
Re-measured against current master, since the figure in the original description was taken before #4160 and #4161 landed.

Three runs back to back in one session, same machine, same reactor, each tree differing only by the commit under test. Benchmark is a workbench theme swap with 20 editors and about 4000 widgets, 5 warmup and 10 measured rounds per direction.

| tree | engine applyStyles per call | per swap | applyStyles calls per swap |
|---|---|---|---|
| baseline predating #4160 | 222.61 us | 3787 ms | 17013 |
| current master | 177.38 us | 3018 ms | 17013 |
| current master + this PR | 155.80 us | 2651 ms | 17013 |

So this PR is worth about 12 percent on top of current master. The originally quoted 21 percent was measured from the older baseline; master itself got roughly 20 percent faster in between, so the same code now starts from a lower number.

Wall clock moves less, since paint and layout dominate a swap: dark swap median 3219.79 to 3094.57 ms, light swap median 3017.48 to 2927.00 ms, dark p95 3621.64 to 3175.38 ms.

The applyStyles call count is identical at 17013 per swap and 340260 in total across all three runs, which is a useful check that the index changes which rules are tested and not which ones match.

Two caveats for anyone reproducing this. Absolute microsecond values drift between sessions, the same pre-#4160 tree measured 196.60 us in an earlier run and 222.61 us now, so only within-session pairs are comparable. And the accumulated engine time counts nested applyStyles calls more than once, so the per-swap figure is a relative metric and not a share of wall-clock swap time.

PR #4166, opened 2026-07-06
Auto-clear persisted workbench state when Eclipse is launched with a different locale

### LINKED ISSUE #3108: Some ViewParts are not localized at startup when using -nl <language> parameter
When launching  application with the -nl <language_code> parameter, some UI components are not getting translated. Once we click on those button, It is getting translated. (like Console, Tasks, Problems, Progress, etc.) that means it do not show translated strings until the user manually activates or interacts with them.

1. Launch an Eclipse application with (-nl <language_code>  (e.g., -nl es) that includes multiple views in its perspective.
2. View the UI components strings like Tasks, Problems, Progress etc.
3. Then Manually click on this views it gets translated.

Eclipse Application startup: (English)

Image

After adding Translation -nl es : 

Image

After Click :

Image

Expected Behaviour: 

All ViewParts should respect the locale at application startup and show the translated UI immediately, even if the view is not currently active.

Workaround
Set the following in config.ini:
clearPersistedState=true

This forces all views to reinitialize and get properly localized, but it disables the persisted UI state across restarts.

deepika-u:
**The Problem**
When Eclipse is launched with -nl <locale> and the workspace already has a persisted workbench.xmi from a previous session in a different language, the UI starts up showing stale labels - view tabs like Declaration, Problems and Javadoc remain in the old language until the user manually clicks on each tab. The only workaround today is launching with -clearPersistedState, which throws away the entire workbench layout every time.

Fixes #3108 

**Root Cause**
Eclipse caches view labels and other UI strings directly in workbench.xmi at shutdown. On the next launch, these cached values are rendered before the NL bundles have a chance to supply the correct translated strings. Since nothing was comparing the current locale against the one used in the previous session, stale labels went undetected.

**Fix in-detail**
A single new private method - hasNlChanged() - is introduced in E4Application. On each startup it does three things:

- Reads the locale recorded from the last launch from a small properties file (last_nl.properties) stored alongside workbench.xmi
- Compares it against the current Platform.getNL()
- If they differ, signals the existing CLEAR_PERSISTED_STATE mechanism to rebuild the workbench model fresh from the correct NL bundles

If the locale is unchanged, or if this is the first ever launch, nothing happens - the user's layout is fully preserved.

**What Changes**
- E4Application.java - the only file modified
- last_nl.properties - a new small file written to `<workspace>/.metadata/.plugins/org.eclipse.e4.workbench/` at first launch, updated on each locale change

No new dependencies. No schema changes. No impact on the common single-locale case(except that locale is being saved now).

_Notes_
- The CLEAR_PERSISTED_STATE path used here is well-tested and already used in production via the -clearPersistedState launch argument. This change simply triggers it conditionally and automatically.
- Locale changes between Eclipse sessions are infrequent - the cost of a one-time layout reset in that scenario is acceptable and consistent with user expectations.
- First launch on a fresh workspace is handled gracefully - last_nl.properties is created silently with no side effects.
- Path resolution uses instanceLocation.getURL().toURI() rather than getFile() to correctly handle Windows paths, where getFile() returns a URL-style path with a leading /C:/ that Path.of() rejects on Windows.

**Without Fix** 
1. Locale is English -> 
image

2. Locale changed to French, immediately seen labels -> 
image

3. On refresh of the tabs, it then shows -> 
image

**With Fix**
1. Locale is English -> 
image

2. Locale changed to French, immediately seen labels -> 
image

### DIFF
--- a/bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
+++ b/bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
@@ -26,9 +26,13 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
@@ -120,6 +124,8 @@ public class E4Application implements IApplication {
 	private static final String SHOWLOCATION_ARG_NAME = "showLocation";
 	private static final String DEFAULT_THEME_ID = "org.eclipse.e4.ui.css.theme.e4_default";
 	public static final String HIGH_CONTRAST_THEME_ID = "org.eclipse.e4.ui.css.theme.high-contrast";
+	private static final String NL_STATE_FILENAME = "last_nl.properties"; //$NON-NLS-1$
+	private static final String NL_STATE_KEY = "eclipse.last.nl"; //$NON-NLS-1$
 
 	private String[] args;
 
@@ -371,6 +377,10 @@ public class E4Application implements IApplication {
 		// Persisted state
 		Boolean clearPersistedState = getArgValue(IWorkbench.CLEAR_PERSISTED_STATE, appContext, true)
 				.map(Boolean::parseBoolean).orElse(Boolean.FALSE);
+		// If not already clearing, check if NL locale changed since last run
+		if (!clearPersistedState) {
+			clearPersistedState = hasNlChanged(instanceLocation);
+		}
 		eclipseContext.set(IWorkbench.CLEAR_PERSISTED_STATE, clearPersistedState);
 
 		String resourceHandler = getArgValue(IWorkbench.MODEL_RESOURCE_HANDLER, appContext, false)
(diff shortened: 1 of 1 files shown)


deepika-u:
@merks : When you get some time, can you review this please?

bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
```diff
@@ -385,6 +391,52 @@ private MApplication loadApplicationModel(IApplicationContext appContext, IEclip
...
+		if (nlFile.exists()) {
+			try (FileInputStream in = new FileInputStream(nlFile)) {
+				props.load(in);
+				String savedNL = props.getProperty(NL_STATE_KEY);
+				if (!currentNL.equals(savedNL)) {
+					localeChanged = true;
+				}
+			} catch (IOException e) {
+				// Cannot read - safe to skip, don't trigger clear
+				return false;
```
merks: If this occurs for whatever reason that the file will never be written.

bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
```diff
@@ -385,6 +391,52 @@ private MApplication loadApplicationModel(IApplicationContext appContext, IEclip
...
+				+ ".plugins" + File.separator //$NON-NLS-1$
+				+ "org.eclipse.e4.workbench"); //$NON-NLS-1$
+		File nlFile = new File(stateDir, NL_STATE_FILENAME);
+
+		String currentNL = Platform.getNL();
+		Properties props = new Properties();
+		boolean localeChanged = false;
+
+		// Read previously saved locale if file exists
+		if (nlFile.exists()) {
```
merks: It's a nit, but if it's a directory it also exists.

bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
```diff
@@ -385,6 +391,52 @@ private MApplication loadApplicationModel(IApplicationContext appContext, IEclip
...
+	 * Persists the current locale in a properties file alongside workbench.xmi.
+	 *
+	 * @return true if the locale has changed since the last launch, false otherwise
+	 */
+	private boolean hasNlChanged(Location instanceLocation) {
+		if (instanceLocation == null || instanceLocation.getURL() == null) {
+			return false;
+		}
+
+		File stateDir = new File(instanceLocation.getURL().getFile(), ".metadata" + File.separator //$NON-NLS-1$
```
merks: If you used Path you could use resolve and wouldn't need to do all the file separator logic...

merks (commented):
I think the whole thing could be cleaner and more "modern" if you use Path/Files instead of File.

bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
deepika-u: Done. Replaced File with Path.of(instanceLocation.getURL().toURI()) and chained .resolve() calls for each path segment - no more manual File.separator concatenation.

bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
deepika-u: Fixed. Changed nlFile.exists() to Files.isRegularFile() which returns true only for actual regular files, correctly excluding directories and anything else that isn't a plain file.

bundles/org.eclipse.e4.ui.workbench.swt/src/org/eclipse/e4/ui/internal/workbench/swt/E4Application.java
deepika-u: Good catch. Replaced return false with localeChanged = true so that execution falls through to the write block regardless - meaning a corrupt or unreadable file gets repaired or recreated on the same launch rather than staying broken indefinitely.

deepika-u:
@merks : Updated the fix now, can you take a look once you get some time please.

merks (approved):
Yes, that looks nice!  Thank you!!

deepika-u:
> Yes, that looks nice! Thank you!!

Thanks alot for the quick review :)

PR #4170, opened 2026-07-08
Smart Import: offer to open/show already-imported projects

### LINKED ISSUE #1978: If importing a project failed because its exist, ask the user to open/jump to the project
Consider the following situation:

1. One has imported a project `X` in the project but closed it
2. No one uses the Wizard "Import existing workspace" or right click in the Git Repositories View and choose "Import Projects ...)
3. Try to import project `X` and get an error that it is already imported

This has bad user experience as the user obviously would not have tried to import it if he was aware of that fact and the dialog does not really help here.

Instead there should be the following options:

1. The project is open, then ask the user to jump to the project (e.g. in Package/Project Explorer)
2. If the project is closed, ask the user to open the project

deepika-u:
**Problem**
When Smart Import found a folder that was already imported as a project, it greyed out the row and silently ignored any attempt to check it - no explanation, no way forward. Users had to abandon the wizard just to go interact with a project that was already sitting right there in the workspace.
That row is still greyed out and still can't be selected for import - that part is correct and unchanged.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1978

**What's now**
Now that it's no longer a dead end:
- **Already open?** → **Show In Project Explorer** jumps you straight to the project.
- **Closed?** → **Open Project** opens it, then shows it.
- **Changed your mind?** → **Cancel** leaves everything exactly as it was.

_Additionally,_
The wizard also gets out of your way on its own: if the project you just acted on was the only thing selected, the wizard closes for you automatically. If other projects are still queued up for import, it stays open so that you can finish the job.

**What's different for the user**
- No more silent, unexplained dead end when a project already exists.
- One click to get to where you actually meant to go - open or reveal, instead of quitting the wizard and hunting for the project
  yourself manually.
- The wizard now senses when there's nothing left to do and closes itself, instead of sitting there disabled and waiting to be
  dismissed manually.
- The affected row is also highlighted the moment the popup appears, so it's obvious at a glance which project the dialog is talking about, even in a long list.

**Scope**
Only applies to projects matching by **location** - i.e. the exact project is already imported. Projects that merely collide by **name**  (different folder, same project name) are untouched and keep their existing silent-exclusion behavior.

**Implementation notes**
- `isExistingProject` now delegates to a new `findExistingProject`, which returns the actual matching `IProject` instead of just a boolean, so the checkbox handler has something to act on.
- The auto-close after a completed action is dispatched via `asyncExec` - this lets the current checkbox-event callback finish
  cleanly before the wizard container is torn down, rather than closing the shell out from under its own event loop.

**Before pr** 
1. Select a project which is already imported, unable to do anything on that specific project. -> 
image

2. Select a project which is already imported but closed, unable to do anything on that specific project. -> 
image

**After pr**
1. Select a project which is already imported. It gets highlighted and what next can be done is shown in the pop-up window  -> 
image
 
Once the user clicks on "Show In Project Explorer", focus is shifted to the new project.
image

2. Select a project which is already imported but closed. -> 
image

Once the user clicks on "Open Project", project is opened and focused on. 
image

3. Select a project which needs to be imported then another which is already imported. -> 
image

Once the user clicks on "Show In Project Explorer", focus is shifted to the project but wizard continues to retain so that other project can be imported when "Finish" is clicked on.
image

4. Select a project which needs to be imported then another which is already imported but closed. ->
image

Once the user clicks on "Open Project", focus is shifted to the opened project but wizard continues to retain so that other project can be imported when "Finish" is clicked on.
image


deepika-u:
@merks : When you get some time, can you review this please?

deepika-u:
Hi @merks, sorry to bother you again, could you try to look into this when you get some time please?

merks:
Sorry I don't have lots of time for testing.  But in general I have concerns.  What if I wanted to open 10 projects as well as import some of the other projects?  This would be pretty painful when this process that supports clicking one and then a dialog for that one.  Also, in the view that's displayed, can I see the difference between a project already exists and is already open versus the project is closed?   And then, if I select a project it seems to come up from that.  Is that just selection (that happens also with keyboard, or check state change (where a user really don't expect a disabled entry to be responsive to anything).   Given it's an import dialog, I might expect as a convenience is to offer as a enabled entry to open the project if it's current closed.   I just don't see the need to turn it into a navigator to open something in a particular view.   And what I've I'm using the package explorer, do I really want the selection in the project explorer?

@iloveeclipse 

Probably other opinions would be useful...

iloveeclipse:
> When Smart Import found a folder that was already imported as a project, it greyed out the row and silently ignored any attempt to check it - no explanation, no way forward. 

AI is mixing things here. **Of** course one can proceed - just without the disabled entry imported. So it is not "silently" ignored, it is **disabled** and visible that it will be not imported because it is already imported, and that is all on the screen. 

> Users had to abandon the wizard just to go interact with a project that was already sitting right there in the workspace.

Not sure why user has to remove "already sitting" project, if it is already there and just doesn't need to be imported, as the wizard properly indicates.

Beside this, proposed workflows with popups and selecting something in some views are strange and "unusual".

We are *in the wizard*, and typically, if there are some extra actions needed, wizard propose built-in functionality or a new page with those extra steps. 

In this case use could select action for each potential candidate, (offered in a new column on same page or in the extra "Import conflicts" page) like:
1) Don't import (already there, represented by unchecked checkbox)
2) Import new project with different project name (inline edit the new name)
3) Remove existing project from location XYZ
4) Rename existing project from location XYZ to ABC

deepika-u:
Thanks both for the detailed feedback - really appreciate you taking the time to dig into this, especially given the bandwidth constraints as mentioned.

There's a lot of good substance here (the disabled-row interaction concern, the open/closed visibility gap, and the broader question of whether this belongs in the wizard as a popup/navigation action at all vs. an inline conflict-resolution flow). I want to sit with these properly rather than reply in a rush, so I'll follow up with specific thoughts and questions in the next few days once I've worked through them in detail.

In the meantime, I've split out a small, non-behavioral refactor(#4222) from the underlying lookup logic here, since it's needed
regardless of which direction this discussion lands on and doesn't touch any of the discussion parts. Figured it's worth landing
independently while we figure out the rest.

More soon - thanks again for the thoughtful reviews.

PR #4174, opened 2026-07-09
Register model-contributed perspectives in PerspectiveRegistry

akurtakov:
Perspectives contributed directly into the application model (e.g. as children of a perspective stack via a model fragment) rather than through the org.eclipse.ui.perspectives extension point or as application snippets were unknown to the PerspectiveRegistry, and therefore invisible to the "Open Perspective" dialog and unresolvable by id.

Scan the model for such perspectives and register descriptors for them, both at startup and lazily from getPerspectives(). Add a test covering a model-contributed perspective.


merks (approved):
Looks reasonable to me.

Issue #3987, opened 2026-05-08
SortingTest.testSorted failed in I-build


trancexpress:
See: https://download.eclipse.org/eclipse/downloads/drops4/I20260507-1800/testresults/html/org.eclipse.search.tests_ep440I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html

```
	Test that all matches have been added again ==> expected: <892> but was: <891>

org.opentest4j.AssertionFailedError: Test that all matches have been added again ==> expected: <892> but was: <891>
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:563)
at org.eclipse.search.tests.filesearch.SortingTest.testSorted(SortingTest.java:70)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596) 
```

Log file is short, I see nothing useful in it: https://download.eclipse.org/eclipse/downloads/drops4/I20260507-1800/testresults/ep440I-unit-win32-x86_64-java21_win32.win32.x86_64_21/org.eclipse.search.tests.AllSearchTests.txt

```
!SESSION 2026-05-07 23:11:50.709 -----------------------------------------------
eclipse.buildId=4.40.0.I20260507-1800
java.version=21.0.5
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_US
[...]

PR #4177, opened 2026-07-10
Stabilize SortingTest match count

akurtakov:
Why mismatch: getMatchCount() can observe a different concurrent state than getElements()/getMatches() (cached count vs traversal), so the old comparison could race on Windows.

Fix: derive baseline count from the same collected match snapshot used for allMatches.

Should stabilize https://github.com/eclipse-platform/eclipse.platform.ui/issues/3987


akurtakov:
@trancexpress Do you have anything against merging this one ?

trancexpress:
No, no objections, lets hope the sporadic fails are gone after this. Thank you!

PR #4190, opened 2026-07-17
Fix Tab navigation in split editors (revert CTabFolder workaround, requires SWT fix)

### LINKED ISSUE #4135: 'Tab' navigatin Bug. Ctrl+PageUp -> Next Tab, Ctrl+PageDown -> Previous Tab.
## Summary

In Eclipse IDE **2026-06 (4.40)**, the `Ctrl+PageUp` and `Ctrl+PageDown` keyboard shortcuts navigate editor tabs in the **opposite direction** compared to all previous versions.

This shortcut pair has been stable for many years (confirmed in 2026-03 / 4.39 and earlier):

| Shortcut | Expected behaviour (≤ 2026-03) | Actual behaviour (2026-06) |
|---|---|---|
| `Ctrl+PageUp` | **Previous Tab** (← move left) | ~~Next Tab~~ (→ move right) ❌ |
| `Ctrl+PageDown` | **Next Tab** (→ move right) | ~~Previous Tab~~ (← move left) ❌ |

---

## Steps to Reproduce

1. Install **Eclipse IDE 2026-06 R (4.40)** — any package (JEE, Java, etc.)
2. Open **3 or more source files** in the editor area (e.g. `A.java`, `B.java`, `C.java`)
3. Click on `A.java` to make it the active editor
4. Press `Ctrl+PageDown`
   - **Expected:** moves to `B.java` (next tab — one step to the right)
   - **Actual:** moves to the tab on the **left** (previous tab direction)
5. Press `Ctrl+PageUp`
   - **Expected:** moves back to `A.java` (previous tab — one step to the left)
   - **Actual:** moves to the tab on the **right** (next tab direction)

> The two shortcuts are completely swapped in direction.

---

## Expected Behaviour

- `Ctrl+PageUp` → moves to the **previous** (left) editor tab
- `Ctrl+PageDown` → moves to the **next** (right) editor tab

This is the documented and long-established behaviour:  
https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.user/concepts/accessibility/navigation.htm

---

## Actual Behaviour

- `Ctrl+PageUp` → moves to the **next** (right) editor tab ← reversed
- `Ctrl+PageDown` → moves to the **previous** (left) editor tab ← reversed

---

## Environment

| Item | Value |
|---|---|
| Eclipse version | 2026-06 R (4.40.0) |
| Package | Eclipse IDE for Enterprise Java and Web Developers (JEE) |
| Last **working** version | 2026-03 R (4.39.0) |
| OS | Windows 10 / 11 (x86_64) |
| JVM | Eclipse-bundled JRE 21 |

---

## Additional Context

- This regression affects **plain Java editors** and is not limited to multi-page editors (e.g. Maven POM).
- The preference binding under **Window → Preferences → General → Keys** still shows `Ctrl+PageUp` mapped to `Previous Editor` and `Ctrl+PageDown` mapped to `Next Editor`, confirming the key-binding metadata is unchanged. The actual navigation direction of the underlying implementation appears to have been inverted.
- A related but narrower bug was previously reported for multi-page editors in 2025-03:  
  → [#2864 — Eclipse 2025-03 Next Tab navigation broken with POM XML editor](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2864)
- Historical Bugzilla reference:  
  → [Bug 199499 — Switching tabs by Ctrl+PageUp/PageDown must not be caught on the inner tab set](https://bugs.eclipse.org/bugs/show_bug.cgi?id=199499)

---

## Workaround

Until a fix is available, the direction can be manually corrected under:

**Window → Preferences → General → Keys**

1. Search for `Previous Editor` → reassign to `Ctrl+PageUp` (overwrite if needed)
2. Search for `Next Editor` → reassign to `Ctrl+PageDown` (overwrite if needed)

---

## Suggested Fix Area

The regression is likely in the `CTabFolder` traversal logic within the SWT/Platform UI layer — specifically the handler that maps `SWT.TRAVERSE_PAGE_NEXT` / `SWT.TRAVERSE_PAGE_PREVIOUS` events to `Ctrl+PageDown` / `Ctrl+PageUp`. A direction flag may have been inadvertently inverted in a recent commit targeting the 4.40 cycle.

fedejeanne:
## What
Fixes #4135.

This PR **reverts `TraversePageHandler` back to its standard, original traversal behavior**, removing the `CTabFolder`-specific special-casing that had been added as a workaround (loop-to-first/last-item logic, hidden-item detection, and the top-level `CTabFolder` parent-hierarchy lookup helper introduced by #2864 and the earlier attempt at fixing #4135).

That special-casing was only needed because `CTabFolder` itself did not loop over its own tabs when navigating with Ctrl+PageUp/PageDown, and it caused regressions such as bypassing the chevron drop-down list when tabs did not fit on screen (which is presumably the root cause of #4135, since the workaround interfered with focus traversal in split editors).

## Dependency
**This PR requires eclipse-platform/eclipse.platform.swt#3448 to be merged/available.** That SWT PR fixes `CTabFolder.onPageTraversal` directly, so `CTabFolder` now correctly wraps around from the last tab to the first (and vice versa) on its own — without needing any workaround at the `TraversePageHandler` level, and without breaking the chevron behavior when not all tabs fit.

## How to test
1. Make sure you have a build of SWT that includes eclipse-platform/eclipse.platform.swt#3448.
2. Open two editors side-by-side (split editors) so that Ctrl+PageUp/PageDown needs to traverse between the editor area's `CTabFolder`s correctly.
3. Confirm that tab navigation with Ctrl+PageUp/PageDown behaves as expected and no longer misbehaves in the split-editor scenario described in #4135.
4. Resize the editor area so that not all tabs fit and the chevron (drop-down list) appears; confirm that traversing past the last/first visible tab still opens the chevron drop-down as expected (this is the behavior that the old workaround used to bypass).

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/handlers/TraversePageHandler.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/handlers/TraversePageHandler.java
@@ -13,11 +13,8 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.handlers;
 import java.lang.reflect.Method;
-import java.util.Arrays;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.swt.SWT;
-import org.eclipse.swt.custom.CTabFolder;
-import org.eclipse.swt.custom.CTabItem;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
@@ -38,16 +35,9 @@ public class TraversePageHandler extends WidgetMethodHandler {
 		Control focusControl = Display.getCurrent().getFocusControl();
 		if (focusControl != null) {
 			boolean forward = "next".equals(methodName); //$NON-NLS-1$
-			int traversalDirection = translateToTraversalDirection(forward);
+			final int traversalDirection = translateToTraversalDirection(forward);
 			Control control = focusControl;
 			do {
-				if (control instanceof CTabFolder) {
-					CTabFolder folder = getTopLevelCTabFolderInParentHierarchy(control);
-					if (isFinalItemInCTabFolder(folder, forward) && !hasHiddenItem(folder)) {
-						loopToFirstOrLastItem(folder, forward);
-						traversalDirection = translateToTraversalDirection(!forward);
-					}
-				}
 				if (control.traverse(traversalDirection)) {
 					return null;
 				}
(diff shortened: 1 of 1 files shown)


fedejeanne (commented):
I tested on **Windows** and it works as expected:
- Circular navigation (forward and backwards) is fine
- Navigation does **not** happen _inside_ composed `CTabFolder`s (e.g. in the _Manifest Editor_) but in the outermost level (it moves between classes, manifest editor and other views)
- If the chevron is visible, it opens it instead of navigating circularly
- There is no _skipping an editor/view_ (as reported in https://github.com/eclipse-platform/eclipse.platform.ui/issues/4135#issuecomment-5003079046).

Here's the video:
 
<details><summary>Video (Windows)</summary>

Image

</details>

I couldn't test under Linux/Mac yet.

HeikoKlare (commented):
I've tested this together with https://github.com/eclipse-platform/eclipse.platform.swt/pull/3448 on MacOS and there it works fine for me (with ordinary editor stacks, MRU stacks, with all tabs being shown as well as some in the chevron menu).

HeikoKlare (approved):

PR #4201, opened 2026-07-23
Find/Replace overlay: fix content assist not available in Java files

### LINKED ISSUE #2651: Find/Replace Overlay: "Content Assist Available" ctrl+space does not work in .java files

HeikoKlare:
## Motivation

Fixes #2651. Pressing Ctrl+Space in the overlay's search/replace fields while regex mode is active shows the "content assist available" hint, but never actually opens content assist — most noticeably in `.java` files.

The cause: `ContentAssistCommandAdapter` (used by the overlay's regex content-assist fields) activates its handler on the workbench-root `IHandlerService` without an expression. The target editor (e.g. the Java editor) keeps its own handler for the same command permanently active via a part-scoped `ActivePartExpression`. Since Eclipse resolves conflicting handler activations for the same command purely by evaluated source priority — independent of context nesting — the editor's higher-priority activation always wins, regardless of where focus actually is.

## What changed

Register an ordinary `FindReplaceOverlayAction` for the content-assist command, the same way `EDIT_FIND_AND_REPLACE` (Ctrl+F) is already handled: through `registerAction()`'s `overlayFocusedExpression` (an `ACTIVE_FOCUS_CONTROL`-based expression). That expression alone already outranks the editor's `ActivePartExpression`-based activation, regardless of which `IHandlerService` the activation is registered on, so no separate, part-scoped override mechanism is needed for this.

## Screenshot

image

---

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlay.java
@@ -501,6 +501,8 @@ public class FindReplaceOverlay {
 		// Also close on Ctrl+F: otherwise it would reopen (a duplicate of) this very overlay.
 		commandSupport.registerAction(
 				new FindReplaceOverlayAction(this::close, IWorkbenchCommandConstants.EDIT_FIND_AND_REPLACE));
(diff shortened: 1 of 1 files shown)


HeikoKlare:
@nettozahler fyi since you asked about that fix here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/4164#issuecomment-4898014524

nettozahler:
This is great so the regex help (which I use quite often) will be usable again in Java files. Many thanks.

PR #4205, opened 2026-07-24
Modernize Java idioms in WizardNewFolderMainPage

deepika-u:
This change modernizes a few Java idioms in WizardNewFolderMainPage by:

- Removing redundant return statements.
- Replacing length() > 0 with isEmpty().
- Use pattern matching for instanceof to simplify type checks and eliminate explicit casts.
- Apply minor readability cleanups.

No functional or behavioral changes are intended.

### DIFF
--- a/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFolderMainPage.java
+++ b/bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFolderMainPage.java
@@ -200,16 +200,15 @@ public class WizardNewFolderMainPage extends WizardPage implements Listener {
 				if (path != null) {
 					IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
 					IResource resource = root.findMember(path);
-					if (resource != null && resource instanceof IContainer) {
+					if (resource instanceof IContainer container) {
 						String resourceName = resourceGroup.getResource();
-						if (resourceName.length() > 0) {
+						if (!resourceName.isEmpty()) {
 							try {
-								return ((IContainer) resource).getFolder(IPath.fromOSString(resourceName));
+								return container.getFolder(IPath.fromOSString(resourceName));
 							} catch (IllegalArgumentException e) {
 								// continue below.
 							}
 						}
-						return resource;
 					}
 					return resource;
 				}
@@ -238,7 +237,7 @@ public class WizardNewFolderMainPage extends WizardPage implements Listener {
 					} else {
 						IPath path = URIUtil.toPath(existingLink);
 						if (path != null) {
-							setDefaultLinkValue = path.toPortableString().length() > 0;
+							setDefaultLinkValue = !path.toPortableString().isEmpty();
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


deepika-u:
@HeikoKlare  - Could you take a look at this when you get some time please?

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFolderMainPage.java
```diff
@@ -232,13 +231,11 @@ private void setupLinkedResourceTarget() {
 			try {
 				if (isFilteredByParent()) {
 					URI existingLink = linkedResourceGroup.getLinkTargetURI();
-					boolean setDefaultLinkValue = false;
-					if (existingLink == null) {
-						setDefaultLinkValue = true;
-					} else {
+					boolean setDefaultLinkValue = existingLink == null;
```
BeckerWdf: I am not a big fan of this change - but that might be personal taste.

BeckerWdf (commented):
change looks good to me

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFolderMainPage.java
deepika-u: Thanks for the feedback! I agree this is largely a matter of personal preference. Since existingLink == null directly evaluates to a boolean, I felt this form was concise and clear, so I'd prefer to keep it unless there are concerns beyond coding style.
Thanks for your time and review :)

BeckerWdf: you can keep it.

merks:
In general I think we should do "cleanups" via automated cleanup actions as @laeubi  has automated them and not manually like this, because a manual approach looks to be a bottomless pit to me.

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFolderMainPage.java
HeikoKlare: I stumbled upon this change as well, as it makes the code more difficult to understand. Then I saw that there was already this comment about the change. Having two people challenging this change because they see a degradation of comprehensibility might be an indicator that it should actually not be done.

HeikoKlare (commented):
In general, the changes look sound (except for one, where I left a specific comment). But I agree with Ed that investing time into optimizing auto-cleanups than doing manual cleanup would probably be even more valuable (though not all of these changes may be achievable with auto-cleanups).

bundles/org.eclipse.ui.ide/extensions/org/eclipse/ui/dialogs/WizardNewFolderMainPage.java
deepika-u: Thanks for the feedback! Since multiple people found this version less readable, I don't think that specific section cleanup is worth it. I've reverted this part of the change and kept only the isEmpty() improvement.

deepika-u:
Thanks for the feedback! I agree that automated cleanups are the more scalable approach where applicable. I came across these changes while working on something else, but I'll keep this in mind for similar cleanup work going forward.

deepika-u:
Please let me know if there's anything else you'd like me to address.

deepika-u:
@BeckerWdf @HeikoKlare 
Please let me know if there's anything else I should address.

HeikoKlare (approved):
Looks good now, thank you.

deepika-u:
> Looks good now, thank you.

Thanks for your time :)

PR #4211, opened 2026-07-28
Remove the deprecated getViewCSS() compatibility bridge

vogella:
Removes the deprecated `getViewCSS()` default method from `CSSEngine`. It was added in #4122 as a bridge for the PDE CSS spy, which at the time still went through the W3C cascade accessor, and was always meant to be temporary.

Nothing in this repository calls it: the `org.eclipse.e4.ui.css.core.engine` package is not exported to any PDE bundle, so the spy reached it reflectively rather than at compile time.

The current spy no longer needs it either. `CssEngineCompat` in `org.eclipse.pde.spy.css` (eclipse-pde#2396, merged) gates on `hasMethod(CSSEngine.class, "computeStyle", Element.class, String.class)` and takes the `computeStyle` path on any engine that has it, so the `getViewCSS` branch is already unreachable on a reworked platform. eclipse-pde#2399 deletes that reflection entirely, but this removal does not depend on it.

Local verification: `org.eclipse.e4.ui.css.core` and `org.eclipse.e4.ui.css.swt` build clean and both suites are green (128 css.core, 210 css.swt with 9 skipped).

Kept as a draft for now so it can follow eclipse-pde#2399. 

The whole CSS engine is internal API

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.core/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.e4.ui.css.core/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-SymbolicName: org.eclipse.e4.ui.css.core;singleton:=true
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
-Bundle-Version: 0.14.800.qualifier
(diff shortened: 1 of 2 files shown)


vogella:
Something for early next release.

PR #4214, opened 2026-07-29
Remove duplicate Link with Editor from menu bar in project explorer

elsazac:
This change removes the `Link with Editor` option from the view menu in project explorer.

The same action is already available in the toolbar, so having it in both places is unnecessary. This also makes the view consistent with the other toolbar actions, which do not have duplicate entries in the view menu.

A similar duplicate entry is there in Package Explorer in JDT UI. I plan to raise for that, once this change is merged.

Before :
Screenshot 2026-07-29 at 7 04 31 PM

After :
Screenshot 2026-07-29 at 7 04 03 PM


SougandhS:
Always felt this was redundant 👍 

+1 to remove

PR #4222, opened 2026-08-04
Extract findExistingProject from isExistingProject

deepika-u:
isExistingProject(File) looked up a matching IProject by location but only returned whether a match was found, discarding the IProject itself. Extract that lookup into a new findExistingProject(File), returning the matching IProject or null, and have isExistingProject delegate to it.

All 7 existing references of isExistingProject(File) continue to receive the same boolean result for the same input as before; none are affected in behavior.

This is groundwork split out from the discussion on #4170, which proposed letting users act on already-imported projects from the Smart Import wizard. That interactive behavior is still under review; this change is limited to the underlying lookup and carries no UI or behavior change on its own.

Why this pr?
#4170  proposes letting users take action on a project that's already imported (open it if closed, navigate to it if open), which needs the actual `IProject`, not just a yes/no. Reviewer feedback on that PR raised open questions about the right interaction model for that (modal dialog vs. inline action vs. a different approach entirely), so that part is still being worked out.

This lookup, though, is needed regardless of which ever direction that discussion lands on. So splitting it out lets it be reviewed and merged on its own, rather than being tied up in a larger design discussion.

### DIFF
--- a/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/SmartImportRootWizardPage.java
+++ b/bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/SmartImportRootWizardPage.java
@@ -787,14 +787,19 @@ public class SmartImportRootWizardPage extends WizardPage {
 		}
 
 	}
-	protected boolean isExistingProject(File element) {
+
+	private IProject findExistingProject(File element) {
 		for (IProject project : ResourcesPlugin.getWorkspace().getRoot().getProjects()) {
 			IPath location = project.getLocation();
 			if (location != null && element.equals(location.toFile())) {
-				return true;
+				return project;
 			}
 		}
-		return false;
+		return null;
+	}
+
+	protected boolean isExistingProject(File element) {
+		return findExistingProject(element) != null;
 	}
 
 	protected boolean isExistingProjectName(File element) {


bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/SmartImportRootWizardPage.java
```diff
@@ -787,14 +787,19 @@ private String getValue(String s) {
 		}
 
 	}
-	protected boolean isExistingProject(File element) {
+
+	protected IProject findExistingProject(File element) {
```
merks, line 791: If the new `findExistingProject` is only called by the old `isExistingProject` method, what is the purpose of such a new method that is only called in one place and could be in-lined at that location?  The fact it's protected suggests that you might want to use it somewhere in a derived class...

deepika-u: Fair point, and you're right to question it in isolation - as it stands in this PR alone, there's no second caller, so the split doesn't pay for itself yet.

The motivation is forward-looking: this was split out of the discussion on #4170, which is exploring letting users act on an already-imported project (open it if closed, etc.), and that would need the actual `IProject`, not just the boolean `isExistingProject` returns today. Since that follow-up direction is still being worked out there, I wanted to land the lookup itself separately rather than bundle it with a still-undecided UI change.

That said, I take the concern seriously - if it's not going to be used elsewhere in the near term, inlining it back into
`isExistingProject` and reintroducing it later (alongside whatever change actually needs it) is a perfectly reasonable alternative, and probably cleaner than speculative API surface sitting unused. Happy to go either way - do you have a preference, or would you rather see this land together with its first real caller instead of ahead of it?

On `protected`: that was just matching the existing visibility of `isExistingProject` and `isExistingProjectName` in the same class, not a signal that subclassing was the intent - no derived class uses it today, so I don't think that reasoning holds up either. If we do keep the split, `private` is more honest until there's an actual external/derived caller.

deepika-u: @merks - can i go ahead making it private for now?

merks (approved):

bundles/org.eclipse.ui.ide/src/org/eclipse/ui/internal/wizards/datatransfer/SmartImportRootWizardPage.java
deepika-u: Made it to private now, thanks.

elsazac (approved):

PR #4226, opened 2026-08-04
Add opt-in block-end code mining

vogella:
Adds a language-agnostic code mining that echoes a block's opening line at its closing brace, so an isolated `}` tells you which block it terminates.
The echoed text is prefixed with `// ` so it reads like the closing brace comment people write by hand.

Braces are matched purely structurally, which keeps `org.eclipse.ui.workbench.texteditor` free of any dependency on JDT or a language model.
Braces in comments and string literals are skipped where the document has a partitioning that marks code as the default content type; a partitioning that uses content types of its own, as TextMate does for every text file, cannot say where the code is and is ignored.
A block is annotated once it spans at least a configurable number of lines, 20 by default.
Nothing in the platform recomputes minings on a document change (JDT ships its own reconciler for that), so the provider follows the document itself and refreshes shortly after typing stops, off the UI thread.

The feature is off by default behind a new checkbox on the Text Editors preference page.
Since that page is already crowded, the new setting and the existing problem annotation code mining settings now share a "Code Minings" group instead of adding two more rows to the flat option list.


vogella:
pr4226-preferences

pr4226-code-minings

PR #4232, opened 2026-08-10
Report per operation timings in the multi editor and perspective tests

vogella:
Continues #4230 for the two tests that cover the operations users feel most: opening many editors and switching perspectives. Both drop org.eclipse.test.performance, warm up before measuring, time the individual operations and print the distribution, so a change in one operation can be attributed instead of disappearing into a single number that nothing recorded.

OpenMultipleEditorTest also reports the first and the last ten editors separately, which turns it into the only test we have that shows how cost scales with the number of open editors. Opening the 90th editor currently costs 20 to 67 percent more than the 10th, and closing an editor while 90 are open costs about three times as much as when 10 remain. Because both ends come from the same run, that comparison is unaffected by machine load, which the absolute numbers very much are.

While adding assertions I found that the test never measured what it claimed: the workbench recycles the oldest editor once REUSE_EDITORS are open, defaulting to 99, so the hundredth open silently reused the first editor. The count is now 90 with an assertion explaining the threshold. PerspectiveSwitchTest had a related problem, silently returning and printing to stdout when a perspective is missing, which reported a pass for something never measured; it now skips through an assumption, so the three JDT cases show up as skipped rather than green.

Sample output of a local run:

```
openMultiple[perf_outline, closeEach] open, first 10  n=10  min= 35.52  p50= 43.01  p90= 67.45 (ms)
openMultiple[perf_outline, closeEach] open, last 10   n=10  min= 54.62  p50= 67.08  p90= 76.15 (ms)
openMultiple[perf_outline, closeEach] close, first 10 n=10  min= 25.48  p50= 30.01  p90= 39.69 (ms)
openMultiple[perf_outline, closeEach] close, last 10  n=10  min=  6.22  p50=  9.46  p90= 17.50 (ms)
openMultiple[perf_text, closeAll] closeAllEditors     n=1   1011.07 (ms, 90 editors)
PerspectiveSwitch to [org.eclipse.ui.tests.performancePerspective1] n=50 min= 16.74 p50= 18.60 (ms)
```

### DIFF
--- a/tests/org.eclipse.ui.tests.performance/src/org/eclipse/ui/tests/performance/OpenMultipleEditorTest.java
+++ b/tests/org.eclipse.ui.tests.performance/src/org/eclipse/ui/tests/performance/OpenMultipleEditorTest.java
@@ -55,8 +55,12 @@ import org.junit.runners.Parameterized.Parameters;
 @RunWith(Parameterized.class)
 public class OpenMultipleEditorTest {
 
-	/** Number of files the {@link UIPerformanceTestRule} creates per extension. */
-	private static final int EDITOR_COUNT = 100;
+	/**
+	 * Stays below the REUSE_EDITORS threshold, which defaults to 99. Above it the
+	 * workbench recycles the oldest editor, so further opens would measure a reuse
+	 * instead of an open.
+	 */
+	private static final int EDITOR_COUNT = 90;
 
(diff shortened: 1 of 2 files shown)


vogella:
Change in test code, performance tests are not part of the CI verification (only compiled)

PR #2320, opened 2024-09-24
Issue 2279 extra

iloveeclipse:
1) File search: show "results filtered" message also in the tree mode

This change unifies "filter" label handling for tree/table mode in the
file search, we show now that the matches are filtered in both cases.

Example:
image

2) File search: cache getMatchCount() result to avoid UI hangs

If the search view has matches for 100.000 files (like above), it would hang forever on updating the view via F5 because it asks getMatchCount() for every element and that traverses entire result again and again.

This change remembers last computed result and returns that value if there were no further changes on the search result itself. This avoids UI hang completely.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2279

PR #4233, opened 2026-08-11
Fix misleading "showing X of Y matches" label in File Search

### LINKED ISSUE #3720: File search incorrectly reports shown matches (shown file count vs shown match count)

- [x] I looked at the source code of the master branch (FileSearchPage.java) and the code seams to be identical to the code in my Eclipse 4.35 installation.

## Steps to reproduce

Create a text file with the following 3 lines:
ABC
ABCBC
ABCDCDC

Then do a file search on the file and search for "A". The label of the search view says "3 matches in ..." OK
Then repeat the search and search for "B". The label of the search view says "4 matches in ... (showing 3 of 4 matches)" which is NOT OK because all matches are shown.
If you search for "C" the label will be "6 matches in ... (showing 3 of 6 matches)" which is also NOT OK.

Since all matches are shown in the search view the view should not report not shown matches.

Locking at the source code of the label generation in FileSearchPage.getLabel() the problem seams to be an incorrect  comparison of lines (leaves in the tree) and matches in the first if branch that generates this label extension.

* OS & version: Windows 11
* Eclipse IDE/Platform version (as shown in *Help > About*): 4.35

## Community

deepika-u:
## Problem
Create a text file with the following 3 lines:
ABC
ABCBC
ABCDCDC

When a File Search match spans multiple occurrences on the same line (e.g. searching for "B" in a file containing "ABCBC"), the Search view label incorrectly reports matches as hidden:

    'B' - 4 matches in workspace (*.txt) (showing 3 of 4 matches)

even though every line is fully visible and no matches are actually left out/skipped. This happens because `FileSearchPage.getLabel()` compares the number of visible tree *leaves* (lines) against the total *match* count and one line can legitimately hold several matches, so `itemCount < matchCount` fires any time a line has more than one hit, regardless of real truncation.

## Fix
- Added a live, filter-independent recount of matches actually represented in the currently visible tree (respecting the existing
  element-limit truncation in `FileTreeContentProvider#getChildren`), using the already-existing   `LineElement#getNumberOfMatches(result)`.
- The "showing X of Y matches" label now only appears when this recount is genuinely lower than the total match count - i.e. when matches are truly missing from view (real truncation).
- When all matches are present but simply grouped across fewer lines, a new label is shown instead: `(found on N lines)`, avoiding the false implication that something is hidden/skipped/missed.

## Changes
- `FileSearchPage.java`: reworked the line-match branch of `getLabel()`; added `getVisibleMatchCount(...)` helpers.
- `SearchMessages.properties` / `SearchMessages.java`: added `FileSearchPage_matches_on_lines_format` message key.

## Testing
Verified against the repro from #3720 (search "A"/"B"/"C" in a filecontaining `ABC`, `ABCBC`, `ABCDCDC`):
- "A" → 3 matches, 3 lines → no qualifier shown (unchanged, correct).
- "B" → previously "showing 3 of 4 matches"; now "found on 3 lines".
- "C" → previously "showing 3 of 6 matches"; now "found on 3 lines".
- "BC" → previously "showing 3 of 4 matches"; now "found on 3 lines".

Fixes #3720

### DIFF
--- a/bundles/org.eclipse.search/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.search/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.search; singleton:=true
-Bundle-Version: 3.19.0.qualifier
+Bundle-Version: 3.19.100.qualifier
 Bundle-Activator: org.eclipse.search.internal.ui.SearchPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.java
+++ b/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -176,6 +176,7 @@ public final class SearchMessages extends NLS {
 	public static String FileSearchPage_sort_by_label;
 	public static String FileSearchPage_limited_format_files;
 	public static String FileSearchPage_limited_format_matches;
+	public static String FileSearchPage_matches_on_lines_format;
 	public static String FileSearchPage_filtered_message;
 	public static String FileSearchPage_filteredWithCount_message;
 	public static String WorkspaceScope;
--- a/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.properties
+++ b/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2000, 2023 IBM Corporation and others.
+# Copyright (c) 2000, 2026 IBM Corporation and others.
 #
 # This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License 2.0
@@ -182,6 +182,7 @@ FileSearchPage_error_marker=Could not create marker
 FileSearchPage_sort_by_label=Sort By
 FileSearchPage_limited_format_files={0} (showing {1} of {2} files)
 FileSearchPage_limited_format_matches={0} (showing {1} of {2} matches)
+FileSearchPage_matches_on_lines_format={0} (found on {1} lines)
 FileSearchPage_open_file_dialog_title=Open File
 FileSearchPage_filtered_message={0} (Filtered)
 FileSearchPage_filteredWithCount_message={0} ({1} matches filtered from view)
--- a/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchPage.java
+++ b/bundles/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchPage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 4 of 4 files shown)


iloveeclipse:
Yes, the lag is there, and it depends on file size - bigger files add delay now after search is completed.

I've used my lorem ipsum project generator, with 22 files, each 10_000 lines (see branch https://github.com/iloveeclipse/java-project-generator/tree/issue_3720) the search freezes UI at the end for some seconds. Bigger files (50_000 lines, ~5MB size each) lead to minute longer freezes. Just imagine searching for something in a big log file with few MB.

Beside this, the labels are still "strange" as soon as the file is bigger as few lines and there are more than few files.

Here is additionally what Claude Opus thinks (I've not checked how much from that is AI slop, but the direction is similar):

```
1. Performance — high severity (main concern)

LineElement.getNumberOfMatches(result) calls result.getMatchSet(parent), which does Collections.enumeration(matches) over the whole match set of the file and then linearly scans it, filtering by identity. The new getVisibleMatchCount calls this once per visible line element, so the cost per file is O(lines × matches) with an enumeration allocation per line.

Multiplied by:

	•  getLabel() is called from AbstractTextSearchViewPage.UpdateUIJob.runInUIThread(), i.e. every 500 ms while the query is running, plus on every element-limit change, layout change, clear, etc.
	•  When match filters are active, getLabel() now does two full tree walks (getVisibleMatchCount + getFilteredMatchCount).

On a large text search (100k+ matches) this is a noticeable UI-thread regression. Suggested fix — one pass over the matches per file instead of one pass per line:

// per file, built once:
Map<LineElement, Integer> counts = new HashMap<>();
for (Enumeration<Match> e = result.getMatchSet(file); e.hasMoreElements();) {
    counts.merge(((FileMatch) e.nextElement()).getLineElement(), 1, Integer::sum);
}

…or better, cache the count in LineElement / short-circuit: if fContentProvider.getLeafCount(result) equals the total number of line elements in the model (no truncation), skip the traversal entirely.

2. Duplicated traversal logic

getVisibleMatchCount(ITreeContentProvider, Object[], result) is nearly identical to the existing getMatchCount(ITreeContentProvider, Object[]) used by getFilteredMatchCount(). The only difference is lineElement.getNumberOfMatches(result) vs getDisplayedMatchCount(element). Consider parameterizing/reusing the existing method (one traversal, one code path) instead of adding a second recursive walker.

3. Filter semantics inconsistency

LineElement.getNumberOfMatches() counts all matches on the line, including filtered ones (the model keeps filtered matches; see AbstractTextSearchResult.setActiveMatchFilters javadoc). Consequences when filters are active:

	•  visibleMatchCount includes matches that are not shown ⇒ truncation can be under-reported;
	•  conversely, lines that disappeared entirely because all their matches were filtered make visibleMatchCount < matchCount ⇒ the label shows "showing X of Y matches" and appends "(N matches filtered from view)" — double/contradictory messaging.

Using getDisplayedMatchCount(element) (as the filtered-count path already does) would at least keep the two computations consistent.

4. Reads SWT widgets instead of the model

getVisibleMatchCount goes through getRootElements(treeViewer) → Tree.getItems(). This means:

	•  UI-thread only (pre-existing constraint of getFilteredMatchCount, but this path now runs on every label update in tree mode, not only when filters are active);
	•  if the tree is empty/not yet refreshed the result is 0, producing a transient, wrong "showing 0 of N matches". Note runBatchedClear() calls updateLabel() before clear(), and setElementLimit() is called from the constructor/restoreState(). A guard (if (roots.length == 0) return result.getMatchCount();) or using cp.getElements(result) would be safer and cheaper.

5. Dead branch

getVisibleMatchCount(AbstractTextSearchResult)'s non-TreeViewer fallback is unreachable: it is only called under showLineMatches(), which requires FLAG_LAYOUT_TREE. Harmless, but worth a comment or removal.

6. Message wording / i18n

FileSearchPage_matches_on_lines_format={0} (found on {1} lines) has no singular form: a file with 2 matches on 1 line yields "found on 1 lines". Also "found on N lines" reads oddly next to the base label ("'foo' - 12 matches in workspace (found on 5 lines)"); something like "{0} ({1} lines)" or "on {1} matching lines" may read better. Key naming/placement in both .java and .properties is consistent — good.

7. Missing tests

No test accompanies the fix, although tests/org.eclipse.search.tests (e.g. SearchResultPageTest) already sets up a search page/viewer and would allow asserting FileSearchPage.getLabel() for the three cases: no truncation, several matches per line, real truncation via element limit.
```

iloveeclipse:
While reviewing this, Opus proposed alternative patch: https://github.com/eclipse-platform/eclipse.platform.ui/pull/4236.

Feel free to review / reuse / update as you like. I will be in vacation from Friday till end of the month, so I would not spend more time on this topic.

deepika-u:
Thanks for surfacing the background on #2320 and #2279 - that context is genuinely useful and wasn't something I had going into this.

The line that connects everything: #2320 (merged Sep 24, 2024) unified "filter" label handling between tree and table mode in file search, and that's the change that took the showLineMatches() branch in getLabel() from dead code to live - which is exactly where the leaf-count-vs-match-count comparison started actually running and producing the misleading label reported in #3720.

To be upfront about scope: #4233 was built purely off #3720 as filed, with no visibility into the earlier UI-freeze history behind this code path. The intent was narrowly a message fix - the label was misleading users into thinking results were missing when nothing was ("showing 4 of 6 matches"), and the goal was simply to correct that and reliably flag the user only when real truncation happens. Nothing more than that was in view.

Seeing #4236 - and the history that motivated it - made it clear this needed to be addressed at the content-provider level, where the original performance issue lived, not just at the label layer. That's a more complete fix than #4233 was ever scoped to be, and I likely wouldn't have arrived at it without this background being pointed out.

Closing this in favor of #4236.

Thanks again for the context and the review time.

PR #4238, opened 2026-08-13
Keep the height reserved by a code mining in sync with the font

vogella:
A line header annotation reserves its vertical space while it is painted, so an annotation that is out of view keeps the space of the font it was last painted with. When the user changes the text size, only the lines still in view take the new size and everything else keeps the old spacing until it is scrolled into view. This is visible in any editor with multi-line line header minings, and it is what makes the experimental unified diff overlay update only partially on a zoom.

`InlinedAnnotationSupport` now refreshes the reserved height of every line header annotation that already reserves space when the font of the text widget changes. SWT reports no font change, so it is noticed on the redraw that the change triggers, and lines that never reserved space keep reserving it lazily on their first paint as before.

The new test in `CodeMiningTest` fails without the fix, with line 5 still reserving 19 pixels after the line height grew to 37.


tobiasmelcher (approved):
Thanks a lot Lars for providing the fix.

vogella:
Thanks @tobiasmelcher for the review

PR #4239, opened 2026-08-13
Report per operation timings in OpenMultipleEditorTest

vogella:
Applies the same treatment to `OpenMultipleEditorTest` that `OpenCloseEditorTest` and `EditorSwitchTest` already received: it drops `org.eclipse.test.performance`, whose measurements went nowhere because the performance database has not been configured for years, and times the individual operations instead.

Each of the 100 opens is now timed on its own and the close phase is timed per variant. Since this test is about the accumulation of editors, the first and last 25 opens are reported as separate distributions, so it is visible whether an open gets more expensive as the editor area fills up. The test warms up before measuring and is bounded by time rather than a single fixed round.

The practical gain is that the test can now fail and can now report something. It asserts that every editor really opened and that none are left behind, and a manual run prints min, median, 90th percentile and maximum per operation. The first numbers already show that closing 100 editors one by one costs noticeably more than `closeAllEditors`, and that the per open cost does not grow at the floor as editors accumulate.

Also fixes the parameterized display name, which labelled the extension as `closeAll` and the `closeAll` flag as `closeEach`.


vogella:
Change in test code, performance tests are not part of the CI verification (only compiled)

PR #4247, opened 2026-08-13
Hash tree paths incrementally when restoring the expansion state

vogella:
Restoring the expansion state walks every item of the tree and looks each one up in a table keyed by `TreePath`. A tree path hashes all of its segments, so an element whose `hashCode` recurses over its children, as an LSP `DocumentSymbol` does, was hashed once for every descendant of every ancestor on its path. The cost of the walk grew with the square of the tree rather than with its size, which is what makes a large outline take seconds to refresh.

A tree path hashes as the sum of its segments, and a sum composes, so `internalSetExpandedTreePaths` now carries the parent's hash down the recursion and derives each child's from it with a single element hash. Once the set of paths to expand is exhausted the remaining items can only be collapsed, which needs neither a path nor its hash, so the walk switches to a plain collapse. `CustomHashtable` keeps the hash in its entries, which lets a lookup rule out an entry without comparing the keys and lets the table grow without hashing the keys again.

Measured with `TreeViewerExpansionTest`, recursive hash visits and the minimum of 100 restores:

| shape | before | after |
|---|---|---|
| fully expanded, 156 of 1884 elements | 361056 visits, 8.3 ms | 44676 visits, 5.1 ms |
| 2 of 1803 elements, all materialized | 1087807 visits, 10.2 ms | 4206 visits, 0.2 ms |

The second shape, a large document with one container open, is where the old behaviour was worst: every item that could no longer match still paid a full path hash. A viewer with an `IElementComparer` was already cheap, so this helps the clients that have not set one, which is the default.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/viewers/AbstractTreeViewer.java
@@ -2202,19 +2202,34 @@ public abstract class AbstractTreeViewer extends ColumnViewer {
 	 *            which are expanded
 	 * @param widget
 	 *            the widget
+	 * @param currentHash
+	 *            the hash {@link TreePath#hashCode(IElementComparer)} would return
+	 *            for <code>currentPath</code>
 	 */
 	private void internalSetExpandedTreePaths(
 			CustomHashtable expandedTreePaths, Widget widget,
-			TreePath currentPath) {
+			TreePath currentPath, int currentHash, IElementComparer comparer) {
 		Item[] items = getChildren(widget);
 		for (Item item : items) {
+			if (expandedTreePaths.size() == 0) {
+				// Every path that had to be expanded has been found, so the rest of the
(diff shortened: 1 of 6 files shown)


vogella:
JUnit A/B (`TreeViewerExpansionTest` from master, run against a jface built from the PR head `8ca1d02592` and against the same tree with the two files reverted to master; 3/3 pass on both sides):

| shape | stock visits | PR visits | stock p50 | PR p50 |
|---|---|---|---|---|
| fully expanded, 156 of 1884 | 361,056 | 44,676 | 10.28 ms | 6.22 ms |
| 1 of 1803, all materialized | 1,086,605 | 1,803 | 13.77 ms | 0.23 ms |
| with `IElementComparer` | 0 | 0 | 5.66 ms | 5.36 ms |

vogella:
The [CustomHashtable.java](https://github.com/eclipse-platform/eclipse.platform.ui/pull/4247/changes#diff-ee552ee64b67a6dc6db77012ef47e4cf4ac5b72fbe897feeff1cf49823967ca0) is also not efficient but that would be another change.

PR #4251, opened 2026-08-15
Force qualifier update for split packages of o.e.text and o.e.jface.text

merks:

This is needed after the changes from https://github.com/eclipse-platform/eclipse.platform.ui/pull/4238


merks:
Windows test failures unrelated. Change is necessary for releasing.

PR #4254, opened 2026-08-19
Fix LTK copy causing change in move operation

### LINKED ISSUE #4253: Use of tryPerformLTKcopy changes behavior of Subclass MoveFilesAndFoldersOpertation
> getUndoableCopyOrMoveOperation is overwritten in MoveFilesAndFoldersOperation. 
> 
> But it will never be called because tryPerformLTKcopy returns true and is not overwritten, so a MoveFilesAndFoldersOpertation is turned into a CopyFilesAndFoldersOperation. 

 _Originally posted by @hd42 in [0aaee1a](https://github.com/eclipse-platform/eclipse.platform.ui/commit/0aaee1a5c92bdde1541b648273f0a52664df284c#r196788078)_

Felix-Schmid:
Fixes #4253.


merks:
I doubt the mac failure is related.  I try to re-run that.

merks:
FYI, I'll be out of office most of the day and need to run almost immediately.

merks (approved):
This looks safe and simple to me.

PR #4258, opened 2026-08-20
Combine the tree viewer performance tests into one class

vogella:
TreeAddTest, FastTreeTest, ShrinkingTreeTest and TreeViewerRefreshTest each measured one aspect of the same widget in a class of its own, which the old performance database needed because a class was its reporting unit. Now that these tests print their own timings, the split only spreads one subject over four files and a shared base class.

TreeViewerPerformanceTest carries all of the scenarios. The additions that FastTreeTest repeated ten times with a removal in between become a round count on the shared measuring method, and the two refresh scenarios keep their own content provider and comparator setup, so the measurements are the ones the separate classes took. TreeTest is gone and its content provider now sits with the test that uses it.


vogella:
Test code which is not affected by the release freeze rule

PR #4259, opened 2026-08-20
Move the Project Explorer performance test to the performance bundle

vogella:
The navigator test bundle carried a PerformanceTest that no suite ran: it was commented out of NavigatorTestSuite for bug 264449 and reported its timings with System.out.println, so it could neither fail nor be compared against anything. It belongs with the other performance tests, which report through UIPerformanceTestUtil since they were migrated off the performance database.

ProjectExplorerPerformanceTest keeps the project deletion and label provider mapping scenarios, measures each of them over several rounds rather than once, and no longer depends on NavigatorTestBase. It also adds testRemoveFilesOfCollapsedProject, which removes files that have no item in the tree because their project was never expanded. That is what a delete below a collapsed project comes down to, and it is the case where the viewer has to ask the content service for a parent, which nothing covered before.


vogella:
Test code which is not affected by the release freeze rule

PR #4261, opened 2026-08-20
Allow org.eclipse.ui.monitoring to run in pure E4 applications

vogella:
The UI freeze monitoring bundle was tied to the 3.x workbench only by a handful of `PlatformUI` calls. Its startup trigger is already the E4 `UILifeCycle.APP_STARTUP_COMPLETE` event, which a plain E4 application fires as well, so very little stood in the way of using it outside the IDE.

Since #4270 moved `ScopedPreferenceStore` to JFace, the bundle can create its preference store directly instead of going through `PlatformUI.createPreferenceStore`, and the monitor thread uses the default Display rather than the workbench display. The 3.x preference page is the only remaining consumer of `org.eclipse.ui`, which is therefore an optional requirement now.

The upside is that any E4 RCP application can add this bundle and get UI freeze reporting in its error log without dragging in the compatibility layer. In the IDE nothing changes and the existing tests still pass.

### DIFF
--- a/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringPlugin.java
+++ b/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringPlugin.java
@@ -20,7 +20,7 @@ import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.jface.preference.IPreferenceStore;
-import org.eclipse.ui.PlatformUI;
+import org.eclipse.jface.preference.ScopedPreferenceStore;
 import org.eclipse.ui.monitoring.PreferenceConstants;
 
 /**
@@ -43,7 +43,7 @@ public class MonitoringPlugin {
 
 	public static IPreferenceStore getPreferenceStore() {
 		if (store == null) {
-			store = PlatformUI.createPreferenceStore(MonitoringPlugin.class);
+			store = ScopedPreferenceStore.getInstance(MonitoringPlugin.class);
 		}
 		return store;
 	}
--- a/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/preferences/MonitoringPreferenceListener.java
+++ b/bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/preferences/MonitoringPreferenceListener.java
@@ -20,7 +20,6 @@ import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.util.IPropertyChangeListener;
 import org.eclipse.jface.util.PropertyChangeEvent;
 import org.eclipse.swt.widgets.Display;
-import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.internal.monitoring.EventLoopMonitorThread;
 import org.eclipse.ui.internal.monitoring.MonitoringPlugin;
 import org.eclipse.ui.internal.monitoring.MonitoringStartup;
@@ -66,7 +65,7 @@ public class MonitoringPreferenceListener implements IPropertyChangeListener {
 
 			monitorThreadRestartInProgress = true;
 
-			final Display display = PlatformUI.getWorkbench().getDisplay();
+			final Display display = Display.getDefault();
 			// Schedule the event to restart the thread after all preferences have had enough time
 			// to propagate.
 			display.asyncExec(this::refreshMonitoringThread);
@@ -86,7 +85,7 @@ public class MonitoringPreferenceListener implements IPropertyChangeListener {
 			// If thread is null, the newly-defined preferences are invalid.
 			if (thread == null) {
 				MessageDialog.openError(
-						PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
+						Display.getCurrent().getActiveShell(),
 						Messages.MonitoringPreferenceListener_preference_error_header,
 						Messages.MonitoringPreferenceListener_preference_error);
 				return;
--- a/bundles/org.eclipse.ui.monitoring/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.monitoring/META-INF/MANIFEST.MF
@@ -11,8 +11,8 @@ Export-Package: org.eclipse.ui.internal.monitoring;x-internal:=true,
  org.eclipse.ui.internal.monitoring.preferences;x-internal:=true,
  org.eclipse.ui.monitoring;x-internal:=true
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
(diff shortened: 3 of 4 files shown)


laeubi (commented):
The change look reasonable in general, but why we need to change JFace API > core runtime preference? JFace is not E3, so likely one only need to adapt getting the preference store what would result in a much smaller scoped PR.

ptziegler:
> JFace is not E3, so likely one only need to adapt getting the preference store what would result in a much smaller scoped PR.

AFAIK, there are only two public implementations of the `IPreferenceStore` interface. The `org.eclipse.jface.preference.PreferenceStore` (which works on _.properties_ files) and the `org.eclipse.ui.preferences.ScopedPreferenceStore` (which works on _.pref_ files).

If you want to create a scoped store, you'll therefore almost always end up with a dependency to the Eclipse 3.x API, unless you switch to the `IEclipsePreferences` interface.

Perhaps there is a way to move the `ScopedPreferenceStore` out of the 3.x UI plugins instead?

laeubi:
> Perhaps there is a way to move the `ScopedPreferenceStore` out of the 3.x UI plugins instead?

Or provide an E4 alternative... given the complexity here it seems not good to replicate the same over and over again. Given this is almost always bound to a bundle, one could even think about having this inject-able via a service instead like we have for the scopes already with `BundleStateScopeServiceFactory`

ptziegler:
The `ScopedPreferenceStore` doesn't use any 3.x-specific functionality, so you could copy the class to an 4.x plugin without any significant changes (or preferably move it to a "neutral" plugin).

Though regardless of whether it's done like this or via a service, migration would be the biggest obstacle. Because in our product, I already see 91 references to the 3.x `ScopedPreferenceStore` and 1193 references to `AbstractUIPlugin.getPreferenceStore()`. If there is now a different way of accessing this preference store then that would mean a lot of work.

laeubi:
> If there is now a different way of accessing this preference store then that would mean a lot of work.

I don't think any of these will change/go away, it would just allow to migrate new code.

vogella:
I like the idea of making ScopedPreferenceStore available outside 3.x. Maybe move it and wrap the old to use it, then we can refactor to update the references? If someone opens a GH issue for it, I can try to look into for the next release.

laeubi:
Is it really that hard to just click this link (at any comment you find useful enough) instead of repeatably asking **in a comment** others to create issues for your book keeping?

grafik

Don't get me wrong but I noticed this at least twice now and it makes me feel you think your time is more valuable than from others so you better delegate a task you can easily do on your own (I do not question here if it takes more time to write a comment waiting for others to do it compared to open an issue right away...).

Apart from that one would expect from committers that the do this kind of architectural design decision at best during / before working on such task. Just because it is easy today to let the AI copy/paste lots of code does not makes it right.

vogella:
> Is it really that hard to just click this link (at any comment you find useful enough) instead of repeatably asking **in a comment** others to create issues for your book keeping?

Nice I did not know this feature. Thanks. Opening issue for other ideas feels wrong to me, like "stealing" the idea of someone, so I prefer to give the honor of creating the idea to the person who had the idea.

laeubi:
> > Is it really that hard to just click this link (at any comment you find useful enough) instead of repeatably asking **in a comment** others to create issues for your book keeping?
> 
> Nice I did not know this feature. Thanks. Opening issue for other ideas feels wrong to me, like "stealing" the idea of someone, so I prefer to give the honor of creating the idea to the person who had the idea.

I think it is common understanding that quoting is the right way to giving note of the original author. In any case nothing buy me (or others) the honor for open a PR (we usually only given credits to the PR author in N&N if at all), so from my side it is fine to take the idea and do something good from it - it otherwise just creates more noise for everyone.

merks:
Yes, if you create a link back to the source issue/pr/discussion the information about the origin remains. typically in both directions.   Please steal"my ideas!

vogella:
Sounds good, thanks for letting me know. So in future if I see something from you two which I think is a good idea and  in case I want to work on it, I handle it directly. I hope that this will not result in double work. Now let me try the "create issue from comment" feature...

vogella:
@ptziegler please finish your your ScopedPreference change, so that I can update this one

bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringPlugin.java
```diff
@@ -43,7 +44,7 @@ public static void logWarning(String message) {
 
 	public static IPreferenceStore getPreferenceStore() {
 		if (store == null) {
-			store = PlatformUI.createPreferenceStore(MonitoringPlugin.class);
+			store = new ScopedPreferenceStore(InstanceScope.INSTANCE, PreferenceConstants.PLUGIN_ID);
```
ptziegler: If the use-case already shows up here, perhaps it really makes sense to also add a static `getInstance(String)` method to the `ScopedPreferenceStore`? Otherwise this can also be simplified to:
```suggestion
			store = ScopedPreferenceStore.getInstance(MonitoringPlugin.class);
```

bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/preferences/MonitoringPreferenceListener.java
```diff
@@ -86,7 +85,7 @@ private synchronized void refreshMonitoringThread() {
 			// If thread is null, the newly-defined preferences are invalid.
 			if (thread == null) {
 				MessageDialog.openError(
-						PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell(),
+						Display.getDefault().getActiveShell(),
```
ptziegler: Given that the call to `PlatformUI.getWorkbench().getActiveWorkbenchWindow()` only works when already inside the UI thread, would it be cleaner to call `Display.getCurrent()` instead? This way you avoid the "One is created if it did not already exist" side-effect, which I don't think is desirable here. You may get a `NullPointerException` instead, but at that point something has already gone fundamentally wrong.
```suggestion
						Display.getCurrent().getActiveShell(),
```

ptziegler (approved):
That looks much cleaner! I just have some minor remarks but nothing serious.

bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/MonitoringPlugin.java
vogella: ScopedPreferenceStore.getInstance(class) is nice. Thanks. I don't think we need the (String) method. I will update the pR

bundles/org.eclipse.ui.monitoring/src/org/eclipse/ui/internal/monitoring/preferences/MonitoringPreferenceListener.java
vogella: Sounds good, I will update the PR

vogella:
Updated, thanks @ptziegler

PR #4265, opened 2026-08-21
Avoid eagerly creating the default font as a side effect in FontRegistry

HeikoKlare:
When `put()` replaces an already-realized font, it obtains the current default font via `defaultFontRecord()` purely to compare it against the replaced font's instances for staleness. But `defaultFontRecord()` does not just look up an already-realized default font — it also creates and caches one if none exists yet.

So replacing an unrelated symbolic name's font allocates a native default-font handle earlier than needed, merely as a side effect of an identity comparison. And because `createFont()` registers the data it used, this is directly observable: `hasValueFor(JFaceResources.DEFAULT_FONT)` flips to `true` and `getKeySet()` starts reporting the default font, purely because some other name was re-`put`.

Look up the already-cached default font record directly instead, tolerating that it may not exist yet — in which case the replaced font's instances are unconditionally treated as stale, exactly as before, since they can never equal a still-unrealized default font.

Includes a regression test asserting that replacing a realized font leaves the default font unregistered.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
@@ -829,7 +829,9 @@ public class FontRegistry extends ResourceRegistry {
 		}
 
 		if (oldFont != null) {
-			oldFont.addAllocatedFontsToStale(defaultFontRecord().getBaseFont());
+			FontRecord defaultRecord = stringToFontRecord.get(JFaceResources.DEFAULT_FONT);
+			Font defaultFont = defaultRecord != null ? defaultRecord.getBaseFont() : null;
+			oldFont.addAllocatedFontsToStale(defaultFont);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


fedejeanne (commented):
Looks good, I would also remove the now unused method [`addAllocatedFontsToStale`](https://github.com/search?q=org%3Aeclipse-platform+addAllocatedFontsToStale&type=code) since it's part of `private class FontRecord`.

HeikoKlare:
> Looks good, I would also remove the now unused method [`addAllocatedFontsToStale`](https://github.com/search?q=org%3Aeclipse-platform+addAllocatedFontsToStale&type=code) since it's part of `private class FontRecord`.

@fedejeanne I think I do not understand that comment. The method belongs to the `FontRecord` class already now. This change just adapts the `defaultRecord` retrieval that is passed to the method. Do I miss something?

fedejeanne (approved):
Nevermind, I just noticed that the method is removed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/4269 so it's all good 👍

HeikoKlare:
Alright. The method is reframed in #4268 and then just taken over in #4269.

Issue #4262, opened 2026-08-21
Move the `ScopedPreferenceStore` out of the 3.x UI plugins


vogella:
> Perhaps there is a way to move the `ScopedPreferenceStore` out of the 3.x UI plugins instead?

Or provide an E4 alternative... given the complexity here it seems not good to replicate the same over and over again. Given this is almost always bound to a bundle, one could even think about having this inject-able via a service instead like we have for the scopes already with `BundleStateScopeServiceFactory`

_Originally posted by @laeubi in https://github.com/eclipse-platform/eclipse.platform.ui/issues/4261#issuecomment-5366264869_

vogella:
Comment from @ptziegler:

````
The ScopedPreferenceStore doesn't use any 3.x-specific functionality, so you could copy the class to an 4.x plugin without any significant changes (or preferably move it to a "neutral" plugin).

Though regardless of whether it's done like this or via a service, migration would be the biggest obstacle. Because in our product, I already see 91 references to the 3.x ScopedPreferenceStore and 1193 references to AbstractUIPlugin.getPreferenceStore(). If there is now a different way of accessing this preference store then that would mean a lot of work.
````

ptziegler:
I looked into this a little bit and I think that `org.eclipse.jface` would be the best location. _However_ this would reintroduce a dependency to `org.eclipse.core.runtime`, which was removed with 287d23a810b9b751654dc16508177c83e72c4bb2 and I therefore assume to be undesirable.

In principle, a dependency to `org.eclipse.equinox.preferences` should be sufficient, if there is an easy replacement for `Platform.getPreferenceService()`. In the worst case, it has to be loaded directly via the OSGi API.

Because this class is used everywhere, I would also keep a dummy implementation in the 3.x plugin. Therefore it probably makes sense to do this via two commits; One where the class is moved and one where the dummy class is created. Otherwise Git won't recognize the class as being moved and the history remains in the dummy class.

vogella:
@ptziegler send a PR once you are ready, I will have a look

vogella:
I also think jface is the correct place. I think the dependency to core.runtime can be avoided. The `PreferencesService.get` call can be inlined.

The consumer change to the moved API should be trival with my new Eclipse MCP server https://github.com/vogellacompany/eclipse-mcp-server and if not I fix it in my MCP server to make it trival. :-)

PR #4270, opened 2026-08-21
Move `ScopedPreferenceStore` to `org.eclipse.jface` #4262

ptziegler:
The `IPreferenceStore` interface has two public implementations; A JFace `PreferenceStore` based on _*.properties_ files and an Eclipse 3.x `ScopedPreferenceStore` based on _*.prefs_ files. Consumers who want to use scoped preferences therefore always carry a direct dependency to the 3.x framework, rendering this interface unusable in a pure 4.x application.

Consumers currently have to fall back to the `IEclipsePreferences` interface, which is not a drop-in replacement, however, due to a different behavior when it comes to e.g. default values.

Because the `ScopedPreferenceStore` doesn't depend on any 3.x-specific functionality, it can be easily moved to the JFace component. The dependency to the Eclipse Runtime bundle can be reduced to the Equinox Preferences bundle by inlining the call to `IPreferenceService.get`.

Note that this change has to be split up into two commits; One where the `ScopedPreferenceStore` is moved and one where a dummy implementation is added to the Eclipse Workbench bundle. Otherwise Git is unable to detect the class as being moved to the JFace component and thus doesn't carry over the file history.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/4262


ptziegler:
@vogella

vogella:
Thanks, @ptziegler 

I would leave 'AbstractUIPlugin.getPreferenceStore()' and 'PlatformUI.getPreferenceStore()' returning the old type for now to avoid class cast exceptions in client code which are casting to it. 

I also suggest to Deprecate the old type and mark it for deletion. After the deprecation has been around for at least a milestone we could IMHO change 'AbstractUIPlugin.getPreferenceStore()' and 'PlatformUI.getPreferenceStore()' but I would not do this immediately without warning to consumer.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/plugin/AbstractUIPlugin.java
```diff
@@ -23,6 +23,7 @@
 import org.eclipse.jface.dialogs.IDialogSettings;
 import org.eclipse.jface.dialogs.IDialogSettingsProvider;
 import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.jface.preference.ScopedPreferenceStore;
```
laeubi: I would also retain the old implementation here.

Instead we should add a static method `ScopedPreferenceStore#instance(String id)` that return `new ScopedPreferenceStore(InstanceScope.INSTANCE,
							id);` and mention it in the javadoc of getPreferenceStore as an alternative for consumer not wanting to use an activator.

To make it even more convenient we could additionally add  `ScopedPreferenceStore#instance(Class<?> context)` and use `FrameworkUtil#getBundle` to a story from the bundle ID itself (the most common case).

bundles/org.eclipse.jface/META-INF/MANIFEST.MF
```diff
@@ -37,7 +37,8 @@ Export-Package: org.eclipse.jface,
 Require-Bundle: org.eclipse.swt;bundle-version="[3.126.0,4.0.0)";visibility:=reexport,
  org.eclipse.core.commands;bundle-version="[3.4.0,4.0.0)";visibility:=reexport,
  org.eclipse.equinox.common;bundle-version="[3.18.0,4.0.0)",
- org.eclipse.equinox.bidi;bundle-version="[0.10.0,2.0.0)";resolution:=optional
+ org.eclipse.equinox.bidi;bundle-version="[0.10.0,2.0.0)";resolution:=optional,
+ org.eclipse.equinox.preferences;bundle-version="[3.12.100,4.0.0)"
```
laeubi: As `ScopedPreferenceStore` is the only consumer we can make this optional, even though I generally try to avoid this, it would not introduce any new dependency.

If preferences does not contain a split package, an import-package might be more suitable here.

laeubi (approved):
I would fetching the `IPreferencesService` through `org.eclipse.core.internal.preferences.PreferencesService.getDefault()` - this is actually internal API but I think in this case acceptable (and we already have other references in platform) and we likely never will change this anyways and would prevent us to duplicate the implementation. Alternatively we could add a method `IPreferencesService.get()` at Equinox side.

As a nit-pick, the `IPreferenceStore` javadoc should be updated as it currently mentions only `PreferenceStore` as the only implementation.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,674 @@
...
+ * </p>
+ * <p>
+ * The default scope is always included in the search scopes when searching for
+ * preference values.
+ * </p>
+ *
+ * @see org.eclipse.core.runtime.preferences
+ * @since 3.40
+ */
+public class ScopedPreferenceStore extends EventManager implements IPersistentPreferenceStore {
```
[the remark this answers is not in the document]
ptziegler: Good point. I've bumped `org.eclipse.ui` and updated the version range of all affected plugins.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,674 @@
...
+	 * search contexts set, then return this store's context.
+	 * <p>
+	 * Whether or not the default context should be included in the resulting list
+	 * is specified by the <code>includeDefault</code> parameter.
+	 * </p>
+	 *
+	 * @param includeDefault <code>true</code> if the default context should be
+	 *                       included and <code>false</code> otherwise
+	 * @return IEclipsePreferences[]
+	 * @since 3.4 public, was added in 3.1 as private method
```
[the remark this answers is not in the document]
ptziegler: Keeping the `@since 3.40` doesn't make much sense when that's the version the class was introduced is. So I removed that line instead.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,674 @@
...
+	 * Return the string value for the specified key. Look in the nodes which are
+	 * specified by this object's list of search scopes. If the value does not exist
+	 * then return <code>null</code>.
+	 *
+	 * @param key            the key to search with
+	 * @param includeDefault {@code true} if the default context should be included
+	 *                       and {@code true} otherwise
+	 * @return String or <code>null</code> if the value does not exist.
+	 */
+	private String internalGet(String key, boolean includeDefault) {
```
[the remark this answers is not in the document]
ptziegler: That sentence doesn't make much sense. First, the test cases already cover both `includeDefault` branches. And the custom search context are not relevant for the underlying change. Given that I've switched to `PreferencesService.getDefault()`, as suggested, I also see this as obsolete now.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/plugin/AbstractUIPlugin.java
ptziegler: > I would fetching the IPreferencesService through org.eclipse.core.internal.preferences.PreferencesService.getDefault()

I like that idea and have updated the PR.

> To make it even more convenient we could additionally add ScopedPreferenceStore#instance(Class<?> context) and use FrameworkUtil#getBundle to a story from the bundle ID itself (the most common case).

I don't want to overdo the static methods so I think the `instance(Class<?>)` method is sufficient. Personally I think that using a raw string is bad practice, because you won't get any notice if the bundle is ever renamed or just outright deleted.

For the name, I've decided to use `getInstance(...)`, because that's what's used in other JFace classes as well.

bundles/org.eclipse.jface/META-INF/MANIFEST.MF
ptziegler: Almost all problems I've ever encountered with OSGi can be traced back to some weird Import-Package shenanigans. So I'd like to avoid that, if possible. Marking this dependency as optional makes sense though, especially because the same is also done for the bidi bundle, which is referenced in the `BidiUtils` class.

ptziegler:
I've also marked the "old" `ScopedPreferenceStore` as "for removal" and supressed the warnings in the `PluginUI` and `AbstractUIPlugin` classes, as suggested by @vogella. A quick search indicated that they are still used in `org.eclipse.ui.console` and `org.eclipse.jdt.ui`. So I'll have to create PRs for those projects as well.

bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/plugin/AbstractUIPlugin.java
laeubi: > Personally I think that using a raw string is bad practice, because you won't get any notice if the bundle is ever renamed or just outright deleted.

Strings become handy if you do not want to bind directly to the bundle (e.g. it is not your own), but of course this is optional and can be added any time later if we find it becomes necessary.

bundles/org.eclipse.jface/META-INF/MANIFEST.MF
laeubi: > Almost all problems I've ever encountered with OSGi can be traced back to some weird Import-Package shenanigans.

`Import-Package` is the OSGi preferred way, just because people use it wrong do not mean it is flawed but Require-Bundle has some serve limitations see:

- https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3198951

I also think it is the responsibility of eclipse-platform to set good examples and "avoiding" something because of vague feelings seems more than wrong. So if in **this concrete case** there is no explicit reason to use `Require-Bundle` (e.g. a split package) we should import the API with a proper version range as a package.

laeubi:
> I've also marked the "old" `ScopedPreferenceStore` as "for removal" and supressed the warnings in the `PluginUI` and `AbstractUIPlugin` classes, as suggested by @vogella.

I must confess I'm not convinced we should deprecate any of this API - it does not bring any value or introduce any maintenance risk here but is *massively* used across the eco-system (as there was no real alternative anyways in the past) and **removing** it will likely break thousand of plugins for no benefit.

merks:
The behavior or deprecating and removing sometimes feels similar to death by 1,000 paper cuts.   While the Platform may have an very active vibrant community, the vast majority of the ecosystem is in maintenance mode.  Moreover, users who find that some cool bundle they use every day suddenly stops working will just blame "Eclipse".   They won't be impressed by how wonderfully clean code base nor by how much better e4, which they aren't using, works for whomever is using that.   So it's great to make things better, but maybe we can do that without making 1,000 people make changes just to keep it working as well as before, with no actual functional benefit to be gained for their use case.

bundles/org.eclipse.jface/META-INF/MANIFEST.MF
ptziegler: > `Import-Package` is the OSGi preferred way

That doesn't mean it's a _good_ way... I think it's the third or fourth time by now where a customer reported a bug of the form: "The application gets stuck during the splash screen, the CPU spins up to 100% and the program gobbles up all available memory", and that just within a few years. And every time, the culprit was an erroneous `Import-Package` header that caused the Equinox resolver to completely go out of control.

>  just because people use it wrong do not mean it is flawed

Isn't simply the fact that so many people get it wrong an indicator, that this part of the specification is just not very good?

> I also think it is the responsibility of eclipse-platform to set good examples

The Eclipse platform should use whatever is the least likely to _break_ applications. And Import-Package headers are right up there.

> because of vague feelings

What about years of experience with working on an imperfect application that depends on imperfect bundles? The vast majority of bundles I've seen don't have a clean manifest and as such continue to cause problems. In a perfect world, `Import-Packages` might be superior. But in reality, I have yet to see this being the case.

ptziegler:
I don't have any strong opinion on marking it as "for removal" or not. But I would still like to deprecate it so that it isn't used in new code.

merks:
Deprecation is merely annoying.  But then most projects ignore the 10,000 warnings in their projects (including the Platform's for its projects).  The sea of yellow is of little concern.  Clients are familiar with being annoyed.  Many are more focused on running on a wide range of dependencies than on making their frameworks work only on the latest version.  That is to the benefit of the end-users and their consumers...

Removal is catastrophic for the consumers who really don't spend every release cycle looking for new warnings that they can't spot among the 10,000 they are currently ignoring.  It's even worse for the users who might well choose something other than Eclipse.

ptziegler:
Yes, a deprecation is annoying. But just because something is deprecated doesn't mean it's going to be removed and as such wouldn't break client code. And to me that's an important distinction.

At the very least the documentation of the "UI" ScopedPreferenceStore should mention that the "JFace" ScopedPreferenceStore exists as a drop-in replacement...

merks:
Yes, deprecation is merely annoying, but, as you suggest, folks can live with annoying.  And the user users won't notice at all.  It's not unreasonable to encourage the use of new better ways, especially for new code.  But we also should be realistic at this point in Eclipse's life that the volume of old code far outweighs volume of new code that is likely to be created.  So how much annoyance do we wish to create for the existing clients in favor of the benefits for hose who will actually benefit?  It's a tradeoff and of course as the provider of a framework we want it to be pure, shinny, and perfect.  But that is not without its downside.  Removal basically has mostly only downsides.

---

For example, we might consider to deprecate the Path constructor because there is a better way:

image

But then I have to spend quite some time dealing with that:

image

EMF builds with very old target platforms that long predate the new and improved way so my code won't get any prettier as a result of the shiny jewel at the center of the ecosystem....

laeubi:
> I don't have any strong opinion on marking it as "for removal" or not. But I would still like to deprecate it so that it isn't used in new code.

I think its fair to deprecate the class itself (as one should use the interface in client code anyways), but I don't see why deprecating the method helps in any ways - a plugin using `AbstractUIPlugin` won't benefit in migration in any kind unless the plugin plans to move away from E3 entirely and then it's good to have a javadoc note there to inform about alternatives.

The problem with deprecation is more that it often describes *that* it is deprecated but not *why* and what are the alternatives. Also sooner or later then one will argue "it has be deprecated for so long" and mark it for removal again.

So I think any deprecation should be choose with whats the benefits for the user/developer and explicitly state this - in wich case there is a noteable benefit users might start to migrate anyways.

bundles/org.eclipse.jface/META-INF/MANIFEST.MF
vogella: I also have bad experience with import packages. Lots of my clients have to use -Dequinox.resolver.revision.batch.size=1 to run their unit tests, without that the resolver goes crazy in the import package resolver, IIRC this was analysed a few years ago in Equinox but we did not find a solution for that.

laeubi: > IIRC this was analysed a few years ago in Equinox but we did not find a solution for that.

This has nothing to do with import-package (and require bundle makes it even more worse especially bogus re-exports) but more with package-substitution (and `equinox.resolver.revision.batch.size=1` just tells the resolver not to try finding a valid solution and leave it to the user to resolve conflicts).

I analyzed this and worked on it already on this and its solvable ... but people better complain and using "hacks" and have bad feelings so no one very yet cared to pay me for fixing that part more deeply.

In any case this is **not** a problem here, so one should not transfer bad feelings from other areas to unrelated things, apart from that import-package support in PDE and platform was always improved over time.

> And every time, the culprit was an erroneous Import-Package header that caused the Equinox resolver to completely go out of control.

That does not mean a *proper* `Import-Package` does cause any problems, that really sounds a bit like "because others do it wrong we can't do it right" ...

ptziegler:
> I think its fair to deprecate the class itself (as one should use the interface in client code anyways), but I don't see why deprecating the method helps in any ways - a plugin using AbstractUIPlugin won't benefit in migration in any kind unless the plugin plans to move away from E3 entirely and then it's good to have a javadoc note there to inform about alternatives.

Which is exactly what I've done. And I wouldn't want to deprecate the utility methods for exactly this reason. I only suppress the deprecation _warnings_ in those methods, because they still use the "UI" store.

At least for the time being, I'll probably remove this annotation again, as this discussion is not relevant for the goal of this PR, which is to have a scoped preference store that doesn't add a dependency to an Eclipse 3.x plugin.

bundles/org.eclipse.jface/META-INF/MANIFEST.MF
merks: It would seem to me better to use a package import.  The package is versioned and is not a split package, so this seems like a best practice where it's generally good that the platform demonstrate best practices..  

But let's not start a religious war on the topic.   Others may feel differently for reasons that may seem misguided.  In my experience resolution problems come, as @laeubi suggests, more from bundles the import their (locally provided) exported packages and do so with range constraint that is too permissive (i.e., plain wrong).   That might lead one to conclude that package imports are bad when in fact importing your exports is something to be done only if there is a very good reason.  When in doubt don't do it.  But I think it's the default to do that, so Orbit generally does this to avoid that behavior:
```
Export-Package:         *;version="${version}";-noimport:=true
```

ptziegler: As said before, I still consider package-imports to be significantly inferior. But that's not a hill I plan to die on. Given that the majority seems to prefer this approach, I've updated the manifest again.

ptziegler:
I've also removed the deprecation annotation and replaced it with a simple comment about the JFace preference store and rebased it on the master.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,679 @@
...
+public class ScopedPreferenceStore extends EventManager implements IPersistentPreferenceStore {
+
+	/**
+	 * Creates a {@link IPreferenceStore} store for the bundle that loaded that
+	 * class
+	 *
+	 * @param clazz the class to use for determining the responsible bundle
+	 * @return the instance scoped preference store for the given class
+	 */
+	public static ScopedPreferenceStore getInstance(Class<?> clazz) {
```
laeubi: Is there any reason for a consumer to use the `ScopedPreferenceStore` directly (e.g. are there methods of interest not in the public API)? If generally not we should use


```suggestion
	public static IPersistentPreferenceStore getInstance(Class<?> clazz) {
```

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,679 @@
...
+ * </p>
+ * <p>
+ * The default scope is always included in the search scopes when searching for
+ * preference values.
+ * </p>
+ *
+ * @see org.eclipse.core.runtime.preferences
+ * @since 3.41
+ */
+public class ScopedPreferenceStore extends EventManager implements IPersistentPreferenceStore {
```
laeubi, line 57: Do we expect any extensions to this class? If not

 
```suggestion
public final class ScopedPreferenceStore extends EventManager implements IPersistentPreferenceStore {
```

would limit the scope of the public API.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,679 @@
...
+	/**
+	 * The searchContext is the array of contexts that will be used by the get
+	 * methods for searching for values.
+	 */
+	private IScopeContext[] searchContexts;
+
+	/**
+	 * A boolean to indicate the property changes should not be propagated.
+	 */
+	protected boolean silentRunning = false;
```
laeubi, line 86: same as above - if we do not expect extensions or package access.


```suggestion
	private boolean silentRunning = false;
```

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,679 @@
...
+	 * Does its best at determining the default value for the given key. Checks the
+	 * given object's type and then looks in the list of defaults to see if a value
+	 * exists. If not or if there is a problem converting the value, the default
+	 * default value for that type is returned.
+	 *
+	 * @param key the key to search
+	 * @param obj the object who default we are looking for
+	 * @return Object or <code>null</code>
+	 */
+	Object getDefault(String key, Object obj) {
```
laeubi: Who needs this package private method?


```suggestion
	private Object getDefault(String key, Object obj) {
```

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,679 @@
...
+			return null;
+		}
+	}
+
+	/**
+	 * Return the IEclipsePreferences node associated with this store.
+	 *
+	 * @return the preference node for this store
+	 */
+	IEclipsePreferences getStorePreferences() {
```
laeubi: Same here

```suggestion
	private IEclipsePreferences getStorePreferences() {
```

laeubi (approved):
It looks good in general just some remarks regarding visibility.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
```diff
@@ -0,0 +1,679 @@
...
+	/**
+	 * Boolean value indicating whether or not this store has changes to be saved.
+	 */
+	private boolean dirty;
+
+	/**
+	 * Create a new instance of the receiver. Store the values in context in the
+	 * node looked up by qualifier. <strong>NOTE:</strong> Any instance of
+	 * ScopedPreferenceStore should call
+	 *
```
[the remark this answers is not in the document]
ptziegler: The original text was `Any instance of ScopedPreferenceStore should have {@link #dispose()} called before it is dereferenced so that listeners can be cleaned up.`. Which doesn't exist anymore. The removal of this requirement was removed with ba247b7e5a6bca75e72629ade871083f42333e22, albeit not completely. I've therefore removed it altogether.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
ptziegler: You're right. There shouldn't be any need to know the explicit type returned by this method. So I've adapted the suggestion and also updated the JavaDoc, because it was still mentioning the `IPreferenceStore`.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
ptziegler: Yes, this class is extended by the `UI` ScopedPreferenceStore. But I've added the `@NoExtend` annotation + the required api filter.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
ptziegler: The "UI" ScopedPreferenceStore might've already been extended by clients. So this protected field has to stay for backwards compatibility.

bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ScopedPreferenceStore.java
ptziegler: Yes. The package-private methods and fields are not used anywhere else. I've changed the the visibility to "private".

ptziegler:
I think I've addressed all open topics now. If there are no objections, I'll merge this tomorrow so that #4261 can continue.

ptziegler:
The failure on MacOS seems to have been caused by the infrastructure:
> An unexpected error occurred while launching the test runtime (process returned error code 143)

Given that previous builds were successful and given that the other builds are green, I'll be merging it regardless.

PR #4271, opened 2026-08-22
The menu item `Show text filter` should use title case

merks:


merks:
@iloveeclipse @HeikoKlare 

I believe this to be completely uncontroversial.   This is new menu item and it doesn't conform to the way all other menu items are named.  I noticed this while reviewing the N&N documentation.

iloveeclipse:
I believe this menu text was "inspired" by the same menu in the Error log view. I have no objections to change it, but probably the Error log view menu should be changed as well.

merks:
Oh, you're right:

image

iloveeclipse (approved):

merks:
I will ignore the macOS failure. I looks unrelated.

PR #4275, opened 2026-08-24
Use a solid color for unselected tabs

### LINKED ISSUE #4274: swt-unselected-tabs-color always takes the gradient path, even for a flat colour
CSSPropertyUnselectedTabsSWTHandler.applyCSSProperty:
  - CssPrimitive → renderer.setUnselectedTabsColor(color) + folder.setBackground(color) — solid path
  - CssList → renderer.setUnselectedTabsColor(colors, percents) + folder.setBackground(colors, percents, true) — gradient path

The platform themes always write two stops (css/dark/e4-dark_tabstyle.css:34, css/light/e4-light_tabstyle.css:27 and :39), so they always take the gradient path, in the dark theme the stops are #49484C / #48484C, i.e. visually flat.

This leads to an image creation in SWT which is wasteful. We should use just one color in our CSS files and support that in the handler.

vogella:
The themes declared two identical gradient stops for `swt-unselected-tabs-color`, which left `CTabFolder.gradientColors` set and made the folder paint the controls in its tab row (min/max bar, chevron, topRight composite) with a rendered background image instead of a plain color. On GTK that image is not painted on a ToolBar, so those bars can fall back to the GTK theme color.

The themes now declare a single color, the CSS handler collapses a flat gradient to the solid path so custom themes get the same benefit, and the unused `*_UNSELECTED_TABS_COLOR_END` definitions are removed. No visual change in the shipped themes: all stops were identical or differed by one unit.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/4274


bundles/org.eclipse.ui.themes/css/dark/e4-dark_ide_colorextensions.css
```diff
@@ -68,12 +66,6 @@ ColorDefinition#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START {
 	label: url('platform:/plugin/org.eclipse.ui.themes?message=INACTIVE_UNSELECTED_TABS_COLOR_START');
 }
 
-ColorDefinition#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_END {
-	color: #464649;
```
BeckerWdf, line 72: bit this value is not the same as used in org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START (just above).

bundles/org.eclipse.ui.themes/css/dark/e4-dark_ide_colorextensions.css
```diff
@@ -128,12 +120,6 @@ ColorDefinition#org-eclipse-ui-workbench-ACTIVE_UNSELECTED_TABS_COLOR_START {
 	label: url('platform:/plugin/org.eclipse.ui.themes?message=ACTIVE_UNSELECTED_TABS_COLOR_START');
 }
 
-ColorDefinition#org-eclipse-ui-workbench-ACTIVE_UNSELECTED_TABS_COLOR_END {
-	color: #48484C;
```
BeckerWdf, line 132: same here. The color used here is not the same as used in ACTIVE_UNSELECTED_TABS_COLOR_START (just above)

bundles/org.eclipse.ui.themes/css/dark/e4-dark_ide_colorextensions.css
```diff
@@ -23,7 +23,6 @@ ThemesExtension { color-definition:
     '#org-eclipse-ui-workbench-DARK_BACKGROUND',
     '#org-eclipse-ui-workbench-DARK_FOREGROUND',
 	'#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START',
-	'#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_END',
```
BeckerWdf, line 26: So it's no longer possible to define a gradient on INACTIVE_UNSELECTED_TABS in a theme?

bundles/org.eclipse.ui.themes/css/dark/e4-dark_ide_colorextensions.css
vogella: This is correct, the end color is not used anymore.

bundles/org.eclipse.ui.themes/css/dark/e4-dark_ide_colorextensions.css
vogella: This is correct, the end color is not used anymore.

bundles/org.eclipse.ui.themes/css/dark/e4-dark_ide_colorextensions.css
vogella: It is still possible to define a theme with gradients. And now it is also possible to define a theme with only one color.

bundles/org.eclipse.ui.themes/css/dark/e4-dark_ide_colorextensions.css
BeckerWdf: So there *is* a visual difference with this change?

vogella: No, gradient was from Color A to Color A which is a solid Color. But the code path creates an image for this solid color as SWT does not understand that we are actually defining a solid color. And it is not a bug in SWT, we ask for a gradient so it created one.

BeckerWdf: thanks for explaining. so "color B" was basically unused?

vogella: Where was no Color B, both values were the same Color A. I think it was "we only support gradients so to get a solid color we use twice the same color".

BeckerWdf: see this:

```
ColorDefinition#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START {
	color: #515658;
	category: '#org-eclipse-ui-presentation-default';
	label: url('platform:/plugin/org.eclipse.ui.themes?message=INACTIVE_UNSELECTED_TABS_COLOR_START');
}

ColorDefinition#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_END {
	color: #464649;
	category: '#org-eclipse-ui-presentation-default';
	label: url('platform:/plugin/org.eclipse.ui.themes?message=INACTIVE_UNSELECTED_TABS_COLOR_END');
}
```

There *are* two different colors for start and end.

vogella:
I was thinking about renaming org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START but that would break existing themes which use this inofficial key. So I leave this like it it.

PR #4276, opened 2026-08-24
Unify test case names in FontRegistryTest

HeikoKlare:
Two test cases for the same behavior of retrieving bold/italic fonts had inconsistent names.


HeikoKlare:
Just test correction, thus merging despite freeze period.

PR #4277, opened 2026-08-24
Add tests for retrieving styled fonts from FontRegistry in non-UI threads

HeikoKlare:
`FontRegistry` supports retrieving a font from a thread that has no `Display` of its own, falling back to the default font instead of failing. That fallback is covered by a test for `get()`, but not for `getBold()` and `getItalic()`, whose variants still have to be realized when they are requested.

These tests close that gap. They pass on `master` — they document existing behaviour rather than change it.


HeikoKlare:
Just test extensions, thus merging even though we are effectively in freeze period.

PR #4278, opened 2026-08-24
Make FontRegistry reuse the main display's font across displays when already available

HeikoKlare:
> [!IMPORTANT]
> This PR is based on six preparatory PRs, which make up its first six commits. They should be reviewed and merged first — **only the last commit, `Reuse the main display's font across displays when already available`, is what this PR actually covers.**
>
> - #4266
> - #4267
> - #4265
> - #4263
> - #4268
> - #4269

### The problem

#4269 gives `FontRegistry` one font cache per `Display`, which fixes ownership and lifetime but gives up sharing: a symbolic name requested from several `Display`s now ends up with a separate native font handle per `Display`. That happens even for names whose data never varies between `Display`s — the JFace default font among them — so the registry allocates font handles that a single shared cache would not have needed.

Before the per-display split, those lookups did share one font. Keeping that sharing is worthwhile in its own right, and it is what the registry used to do.

### The solution

A lookup prefers a font already realized on the registry's **main display**, provided that font carries the exact style requested. Only when it does not does the requesting `Display` realize and cache a font of its own. Resource usage stays close to a single shared cache, while ownership and disposal remain per `Display` as introduced in #4269.

This is safe precisely because of the invariant #4269 relies on: the main display outlives every other `Display` the registry is used from, so a font borrowed from it cannot be disposed while another `Display` is still using it. The flip side is explicit — **the font returned for a symbolic name is no longer necessarily owned by the caller's `Display`**, so callers must not infer their `Display` from `Font#getDevice()`.

Two properties keep the behaviour predictable:

- A `Display` that already had to realize a font of its own keeps getting that one. Repeated lookups never silently switch instances once the main display catches up, and this costs nothing, since such a font only exists where it had to be created anyway.
- Only a record that *already* carries the requested style is reused. Lazily creating a styled font therefore stays confined to the `Display` owning the record, so it needs no synchronization of its own.

Callers without a `Display` of their own benefit as well: where they could previously only fall back to the default font, they now reuse any already realized font the same way.

Because reuse depends on which style is being asked for, the style is now known while a record is looked up, instead of being applied to the record afterwards.

For a single-display application this changes nothing: that `Display` is the main display, so every lookup already finds its own font.

### The verification

Four tests are added, on both sides of the line:

- **Two fail before this commit and pass after.** `multipleDisplay_reusesMainDisplayFont_whenStyleAlreadyCached` covers the reuse directly; `multipleDisplay_reusesMainDisplayFont_onceStyleBecomesCachedThere` covers reuse starting once the main display realizes the style later.
- **Two pass before and after.** They pin the boundary where reuse must *not* happen: `multipleDisplay_createsOwnFont_whenStyleNotYetCachedOnMainDisplay` asserts that a style the main display has not realized is created on — and disposed with — the requesting `Display`, and `multipleDisplay_keepsOwnFont_whenMainDisplayRealizesItLater` asserts that a `Display` holding its own font keeps it.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
@@ -64,6 +64,10 @@ import org.eclipse.swt.widgets.Display;
 @NoExtend
 public class FontRegistry extends ResourceRegistry {
 
+	private enum FontStyle {
+		NORMAL, BOLD, ITALIC
+	}
+
 	/**
 	 * FontRecord is a private helper class that holds onto a font
 	 * and can be used to generate its bold and italic version.
@@ -109,8 +113,21 @@ public class FontRegistry extends ResourceRegistry {
 		 * Return the base Font.
 		 * @return Font
 		 */
-		public Font getBaseFont() {
+		private Font getBaseFont() {
 			return baseFont;
+		}
+
+		/**
+		 * Return the font for the given style, creating it lazily if necessary.
+		 * @param style the requested style
+		 * @return the font
+		 */
+		Font get(FontStyle style) {
+			return switch (style) {
+				case NORMAL -> getBaseFont();
+				case BOLD -> getBoldFont();
+				case ITALIC -> getItalicFont();
+			};
 		}
 
 		/**
@@ -118,7 +135,7 @@ public class FontRegistry extends ResourceRegistry {
 		 * of the base font to get it.
 		 * @return Font
 		 */
-		public Font getBoldFont() {
+		private Font getBoldFont() {
 			if (boldFont != null) {
 				return boldFont;
 			}
@@ -130,6 +147,20 @@ public class FontRegistry extends ResourceRegistry {
 			FontData[] boldData = getModifiedFontData(SWT.BOLD);
 			boldFont = new Font(baseFont.getDevice(), boldData);
 			return boldFont;
+		}
+
+		/**
+		 * Returns whether the given style has already been realized for this
+		 * record.
+		 * @param style the style to check
+		 * @return whether the given style is already available
+		 */
+		boolean has(FontStyle style) {
+			return switch (style) {
+				case NORMAL -> baseFont != null;
+				case BOLD -> boldFont != null;
+				case ITALIC -> italicFont != null;
+			};
 		}
 
 		/**
@@ -158,7 +189,7 @@ public class FontRegistry extends ResourceRegistry {
 		 * base font to get it.
 		 * @return Font
 		 */
-		public Font getItalicFont() {
+		private Font getItalicFont() {
 			if (italicFont != null) {
 				return italicFont;
 			}
@@ -236,7 +267,7 @@ public class FontRegistry extends ResourceRegistry {
 				return;
 			}
 			FontRecord defaultRecord = records.get(JFaceResources.DEFAULT_FONT);
-			Font defaultFont = defaultRecord != null ? defaultRecord.getBaseFont() : null;
+			Font defaultFont = defaultRecord != null ? defaultRecord.get(FontStyle.NORMAL) : null;
 			replacedRecord.getAllocatedFonts().stream().filter(font -> font != defaultFont).forEach(staleFonts::add);
 		}
 
@@ -609,7 +640,16 @@ public class FontRegistry extends ResourceRegistry {
 	 * @return Font
 	 */
 	public Font defaultFont() {
-		return defaultFontRecord().getBaseFont();
+		return defaultFont(FontStyle.NORMAL);
+	}
+
+	/**
+	 * Return the default font in the given style, creating it if necessary.
+	 * @param style the requested style
+	 * @return the font
+	 */
+	private Font defaultFont(FontStyle style) {
+		return defaultFontRecord(style).get(style);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
```diff
@@ -582,15 +693,65 @@ record = createFont(JFaceResources.DEFAULT_FONT, fontData);
...
+	 * still using it.
+	 * </p>
+	 * <p>
+	 * Requiring the exact style to be present is also what keeps the lazy
+	 * creation of styled fonts single-threaded: a record is only ever handed to
+	 * a foreign thread once the style it asks for has been realized, so only
+	 * the record's own display ever reaches the creating branch of
+	 * {@link FontRecord#get(FontStyle)}. Note that this means the main
+	 * display's record always wins for {@link FontStyle#NORMAL}, whose font is
+	 * realized when the record is created.
```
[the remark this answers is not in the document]
HeikoKlare: Good catch, and the mismatch is broader than the `NORMAL` note. This Javadoc still described the main-display-first ordering the method originally had. The ordering was flipped to current-display-first during review of #4269, and the documentation was not updated along with it. Three statements were wrong:

- the precedence was stated inverted, since the current display is checked first;
- "the main display's record always wins for `NORMAL`" is in fact the opposite: for `NORMAL` every existing record matches, so the current display's own record wins whenever it has one;
- "returns `null` if neither has a matching record" is inaccurate, because the method returns the current display's record even when it lacks the requested style, and `null` only when the current display has no record for the name at all. That fallback is load-bearing: it is what makes the style get realized on the display owning the record.

Rewritten to describe the actual lookup order and that fallback.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
```diff
@@ -152,68 +189,118 @@ private FontData[] getModifiedFontData(int style) {
...
+	 * invalidates the records of <em>every</em> display from whichever thread it
+	 * is called on, and may do so while that display is disposing itself.
+	 * </p>
 	 */
-	private final Map<String, FontRecord> stringToFontRecord = new HashMap<>(7);
+	private static class DisplayFontRecords {
+
+		private final Map<String, FontRecord> records = new ConcurrentHashMap<>();
+
+		private final Collection<Font> staleFonts = new ConcurrentLinkedQueue<>();
```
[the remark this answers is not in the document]
HeikoKlare: Agreed, and the suppressed comment on `dispose()` names the actual consequence: `forEach` followed by `clear()` can drop a font enqueued by a concurrent `put()` without ever disposing it, leaking the handle.

`staleFonts` is now a `Queue<Font>`, and `dispose()` drains it with `poll()`, so every removed element is disposed exactly once and a font enqueued concurrently is either disposed here or still queued afterwards.

Note that this belongs to the "Scope FontRegistry font caching and disposal to the Display that created each font" commit, which is under review as #4269, so the fix was applied there rather than in this PR's own commit.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
```diff
@@ -667,44 +826,41 @@ public Font getBold(String symbolicName) {
...
-	private FontRecord getFontRecord(String symbolicName) {
+	private Font getFont(String symbolicName, FontStyle style) {
 		Assert.isNotNull(symbolicName);
-		Object result = stringToFontRecord.get(symbolicName);
-		if (result != null) {
-			return (FontRecord) result;
+		FontRecord existingRecord = getExistingFontRecord(symbolicName, style);
+		if (existingRecord != null) {
+			return existingRecord.get(style);
 		}
```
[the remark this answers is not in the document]
HeikoKlare: This path is intentional and unchanged by this PR. It is master's existing behaviour, pinned by tests merged in #4277.

`getBold_fromNonUIThread_realizesBoldFontOnTheRegistrysDisplay` and `getItalic_fromNonUIThread_realizesItalicFontOnTheRegistrysDisplay` set up exactly the state described (the plain default font realized, the styled variant not), call `getBold`/`getItalic` from a thread without a `Display`, and assert that the caller does get a styled font, realized on the registry's display. Both pass on this branch.

The suggested restriction, never creating styled fonts off the UI thread and falling back to `NORMAL`, would therefore change behaviour that master deliberately covers, and would fail those two tests. The `Assert` in `FontRecord#getBoldFont()`/`getItalicFont()` explicitly admits `Display.getCurrent() == null` for this reason: the font is realized on the display owning the record, not on the caller's.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
```diff
@@ -106,26 +113,56 @@ void dispose() {
...
+		private Font getBaseFont() {
 			return baseFont;
 		}
 
+		/**
+		 * Return the font for the given style, creating it lazily if necessary.
+		 * @param style the requested style
+		 * @return the font
+		 */
+		public Font get(FontStyle style) {
```
fedejeanne: This method could be `private`.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
```diff
@@ -566,8 +672,17 @@ public FontDescriptor getDescriptor(String symbolicName) {
 	/**
 	 * Returns the default font record.
 	 */
-	private FontRecord defaultFontRecord() {
-		FontRecord record = stringToFontRecord.get(JFaceResources.DEFAULT_FONT);
+	private FontRecord defaultFontRecord(FontStyle style) {
```
fedejeanne, line 681: You could update this Javadoc to document the new style parameter (and, ideally, the return value), similar to what was done for `defaultFont(FontStyle)` above and also mention that _it might create a new font_ (I see calls to `createFont(...)` in it).

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
```diff
@@ -132,6 +149,20 @@ public Font getBoldFont() {
 			return boldFont;
 		}
 
+		/**
+		 * Returns whether the given style has already been realized for this
+		 * record.
+		 * @param style the style to check
+		 * @return whether the given style is already available
+		 */
+		public boolean has(FontStyle style) {
```
fedejeanne: This method could be `private`.

fedejeanne (approved):
Looks good, just some nit comments.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
HeikoKlare: Makes sense. I have extended the Javadoc accordingly.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
HeikoKlare: Since this is a private inner class, the visibilities of members do not matter technically, because the outer class can access everything anyway. The visibility is rather used to indicate if the outer class is supposed to access that member or not (like it was done before this change). i.e., a private method is not supposed to be accessed by the outer class but a non-private one is. I adapted the change to an intermediate state (that I had also adopted for the `DisplayFontRecords`) that declares the methods to be used by the outer class with default visibility.

bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
HeikoKlare: Same as above

PR #4280, opened 2026-08-26
Derive heap status colors from background and foreground

vogella:
The heap status kept its paint colors in fields, seeded from hardcoded grays and system colors, and overrode `setBackground`, `setForeground` and `getForeground` to keep those fields in sync. As a result a theme could not restyle the control through the usual means, and `getForeground` returned a fixed gray rather than the control's actual foreground.

All painted colors except the low-memory red and the free-memory orange now derive from the control's own background and foreground, so the standard `background-color` and `color` properties are all a theme needs to set. The border is drawn as one flat rectangle in the derived color instead of a two-tone bevel built from `COLOR_WIDGET_HIGHLIGHT_SHADOW`.

This also drops ten color fields and the three overrides, and lets the widget pick up an inherited background from the trim, which the old fixed `COLOR_WIDGET_BACKGROUND` default prevented.


vogella:
heap-before-after

PR #4287, opened 2026-08-27
Wait for the decoration jobs instead of sleeping a fixed time

vogella:
`DecorationSchedulerRaceConditionTest` slept a fixed 500 ms after each decoration and assumed the update job had run by then. That ties the test to `DecorationScheduler.UPDATE_DELAY`, an internal constant it has no business knowing about.

It now waits for the `FAMILY_DECORATE` jobs to finish instead. A plain `join` on the family would deadlock, because the update job is a `WorkbenchJob` that needs the UI thread, so the wait spins the event loop through `DisplayHelper` and polls the job manager.

The timeout goes from 500 ms to 5 s since it is an upper bound now rather than an unconditional sleep. In practice the test gets faster (both cases run in 2.8 s), and a timeout fails with a clear message instead of an `assertEquals` on stale text.


vogella:
just a test change

Issue #2498, opened 2024-11-06
Find/Replace Overlay: Compare Editor


jukzi:
ctrl+f in Compare editor still opens modal search dialog. That feels inconsistent
![image](https://github.com/user-attachments/assets/bd003226-a880-4865-ae09-fe474d87dc77)

HeikoKlare:
This is actually not a bug but a known limitation: the overlay is currently only implemented for `StatusTextEditor`. This limitations comes from how the positioning happens, which requires specific information that is not generically available for all kinds of find targets, so some conceptual extension needs to be implemented for that.

Wittmaxi:
In fact, the same happens in the Console. As Heiko has pointed out, this is by design.
I am planning on +eventually+ coming around to at least implementing a concept that could help other vendors use the Find/Replace Overlay themselves.

PR #4292, opened 2026-08-28
Find/Replace overlay: add tests for its interplay with the host editor

HeikoKlare:
## Why

The Find/Replace overlay is drawn inside its host editor's widget tree, so focusing one of its input fields does not change the active workbench part. The editor stays active, and its key bindings and command handlers therefore compete for every keystroke with the field the user is typing into. Getting that wrong is very visible: Ctrl+Z undoing in the document instead of in the search field, Ctrl+A selecting the whole file, Delete editing the document.

This has gone wrong repeatedly since the overlay was introduced, and every time it was noticed by someone trying it out rather than by a test, because nothing covers it: `FindReplaceOverlayTest` drives the overlay on a bare `TextViewer`, which has no workbench part behind it, so the situation the whole mechanism exists for never arises there.

## What this adds

`FindReplaceOverlayInEditorTest`, an end-to-end test that opens a real editor and reaches the overlay through `FindReplaceAction`, the way a user does. It asserts only *where* a command ends up taking effect, never *how* that is arranged, so it stays valid across changes to the implementation.

It passes against unmodified production code: this is a regression guard, not a specification of new behaviour, and there is no production change in this PR.

## Why now

The overlay is currently restricted to `StatusTextEditor`: `FindReplaceAction#shouldUseOverlay()` only uses it for editors of that type, and `FindReplaceOverlay` itself branches on it in two places. Subsequent changes are intended to remove that restriction so the overlay can serve other editors, for instance the compare editor, which still opens the modal Find/Replace dialog (#2498).

Lifting that restriction means changing how the overlay and its host editor arbitrate commands, which is exactly the behaviour described above. Adding the guard first means those changes, and any later ones in this area, can be judged against behaviour that is already pinned down rather than re-verified by hand.

## Notes on the test setup

A real editor part is needed but the workspace and the IDE are not, so the test contributes a minimal editor of its own; marking the test bundle as a singleton is what makes its `plugin.xml` read at all. Key strokes are notified to the widget rather than posted as native events, which still runs the display filter the key binding dispatcher installs but does not depend on the operating system, so the test also works headless.

### DIFF
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
-Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor.tests
+Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor.tests;singleton:=true
 Bundle-Version: 3.15.0.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
--- a/tests/org.eclipse.ui.workbench.texteditor.tests/build.properties
+++ b/tests/org.eclipse.ui.workbench.texteditor.tests/build.properties
@@ -13,6 +13,7 @@
 #     Mickael Istria (Red Hat Inc.) - 419531 Get rid of nested jars
 ###############################################################################
 bin.includes = plugin.properties,\
+               plugin.xml,\
                test.xml,\
                about.html,\
                .,\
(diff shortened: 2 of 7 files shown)


iloveeclipse:
@HeikoKlare: there are test failures in the new `FindReplaceOverlayInEditorTest` on master on all platforms, see e.g. https://download.eclipse.org/eclipse/downloads/drops4/I20260831-2300/testresults/html/org.eclipse.ui.workbench.texteditor.tests_ep442I-unit-linux-x86_64-java25_linux.gtk.x86_64_25.html

HeikoKlare:
I'll have a look. Must be some mistake in the test setup as the tests run fine with Tycho and the similarly set up `FindReplaceOverlayTest`s also succeed on Jenkins.

HeikoKlare:
No test failures in today's I-Build anymore, so https://github.com/eclipse-platform/eclipse.platform.ui/pull/4330 seems to have fixed the issue in the Jenkins runs.

PR #4293, opened 2026-08-28
Find/Replace overlay: make the overlay the active part while focused

HeikoKlare:
## Rationale

The Find/Replace overlay is drawn inside the widget tree of the editor it searches in, so focusing one of its input fields does not change the active workbench part. The editor stays active, and its key bindings and command handlers therefore compete for every keystroke with the field the user is typing into. Getting that wrong is very visible: undo applying to the document instead of to the search field, Select All selecting the whole file, Delete editing the document.

So far this was handled by reflectively calling the private `AbstractTextEditor#setActionActivation(false)` while an input field has focus, and by nulling out the editor's global action handlers for multi-page editors. Both are workarounds rather than a mechanism: they reach into private API, they only work for editors derived from `AbstractTextEditor`, and they cover the six retargetable actions rather than the full set of commands that conflict with typing into a text field.

That coupling is also what keeps the overlay tied to one kind of editor. It is currently offered only for `StatusTextEditor`, and `FindReplaceOverlay` branches on that type in two further places.

## Concept

Instead of suppressing the editor's commands one at a time, the overlay is given an `IEclipseContext` of its own, which it activates while one of its input fields has focus and in which it publishes its own id as the active part id.

Two independent conditions decide whether one of the editor's handlers wins a command, and the editor's commands are spread over both. Its key binding scopes and part-level handlers are only reachable through the editor's part context, and activating a context that is a sibling of that context takes them off the active chain. Its retargetable actions are not registered there at all but in the window context, guarded by an expression over the active part id, which publishing the overlay's own id makes evaluate to false.

Together this leaves no editor handler in the command resolution path, so command enablement no longer depends on the editor type, on a list of commands to suppress, or on reflection. Keys that the platform does not otherwise handle reach the native text widget, and the workbench's own default handlers for cut, copy, paste and select all act on the focused input field, so those work from the Edit menu again. The overlay's own key binding scopes are activated in that same context, which also gives them their lifetime.

## What this enables

With command enablement independent of the editor type, the remaining `instanceof StatusTextEditor` checks can be removed in follow-up changes, which is what would let the overlay be used with editors other than text editors:

- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2498: Ctrl+F in the compare editor still opens the modal Find/Replace dialog
- https://github.com/eclipse-pde/eclipse.pde/issues/1749: the XMI tab of the E4 tools EMF editor could use the overlay instead of its own search

Neither is addressed here. This change removes the reason those checks exist.

With the simple removal of the `StatusTextEditor` limitation, immediately the `Console` view would have the find/replace overlay available:
image

## Alternatives

A number of architecturally quite different approaches were considered, and several were built as proofs of concept and measured against each other, among them suppressing the editor's commands individually, giving the overlay its own shell, and modelling it as a real workbench part. The approach proposed here turned out to be both the simplest and the most flexible: it needs no command list, no reflection, no knowledge of the editor, and it leaves the workbench's notion of the active part, the active editor and the selection untouched, so nothing else in the IDE observes a change while the overlay has focus.

Because the reasoning behind that comparison is not recoverable from the code, the insights and the decision are recorded as an architecture decision record in `docs/adr/0001-find-replace-overlay-key-handling.md`, including the alternatives, what was measured about them, and the consequences of the chosen one.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/plugin.xml
+++ b/bundles/org.eclipse.ui.workbench.texteditor/plugin.xml
@@ -1561,7 +1561,7 @@
       <context
             id="org.eclipse.ui.workbench.texteditor.findReplaceOverlay"
             name="%context.findReplaceOverlay.name"
-            parentId="org.eclipse.ui.textEditorScope">
+            parentId="org.eclipse.ui.contexts.window">
       </context>
       <context
             id="org.eclipse.ui.workbench.texteditor.findReplaceOverlay.searchFocused"
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayContextSupport.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayContextSupport.java
@@ -1,0 +1,244 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.ui.internal.findandreplace.overlay;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.eclipse.core.expressions.EvaluationResult;
+import org.eclipse.core.expressions.Expression;
+import org.eclipse.core.expressions.ExpressionInfo;
+import org.eclipse.core.expressions.IEvaluationContext;
+
+import org.eclipse.e4.core.contexts.IEclipseContext;
+import org.eclipse.e4.ui.model.application.ui.basic.MBasicFactory;
+import org.eclipse.e4.ui.model.application.ui.basic.MPart;
+import org.eclipse.e4.ui.services.EContextService;
+
+import org.eclipse.ui.ISources;
+import org.eclipse.ui.IWorkbenchPart;
+import org.eclipse.ui.IWorkbenchPartSite;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
+
+/**
+ * Owns the Find/Replace overlay's contexts, of which there are two kinds: an
+ * {@link IEclipseContext} that makes the overlay rather than its host editor the
+ * active part while an input field has focus, and the overlay's key binding
+ * scopes. The second kind lives inside the first, because scopes are collected
+ * along the chain between the active leaf and the root: a scope activated here is
+ * active exactly while this context is the active leaf, so the shared scope is
+ * activated once and only the per-field scope is switched.
+ * <p>
+ * Two conditions decide whether one of the editor's handlers wins, and the
+ * editor's commands are spread over both. Its key binding scopes and part-level
+ * handlers are only <em>reachable</em> through the editor's part context, which
+ * activating a sibling of that context takes off the chain. Its retargetable
+ * actions live in the window context instead, guarded by an <em>expression</em>
+ * over {@link ISources#ACTIVE_PART_ID_NAME}, which publishing this context's own
+ * id as the active part id makes false. Together they leave no editor handler in
+ * the resolution path, so no command has to be suppressed individually. The
+ * context sits below the window context rather than the application context,
+ * which would detach window-scoped commands and services as well.
+ * <p>
+ * The {@link MPart} looks superfluous, since nothing reads it, but removing it
+ * changes behaviour: it is what lets {@code ActivePartLookupFunction} resolve an
+ * active part here. Without it that lookup yields {@code null}, and
+ * {@code PartServiceImpl} answers a null active part by firing part deactivation
+ * and clearing the active selection. With it, the same code path finds a part
+ * outside the application model and returns early. That early return is also why
+ * the part is neither added to the model nor rendered nor activated through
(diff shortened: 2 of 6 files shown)


bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayContextSupport.java
```diff
@@ -0,0 +1,220 @@
...
+	 */
+	void fieldsLostFocus() {
+		if (!overlayContextActive) {
+			return;
+		}
+		overlayContextActive = false;
+		if (contextToRestore != null) {
+			contextToRestore.activateBranch();
+		}
+	}
```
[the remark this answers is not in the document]
HeikoKlare: Thanks. I applied the second half of this, clearing `contextToRestore` after restoring, but not the defensive wrapping, because I don't think `activateBranch()` can throw here.

`EclipseContext` keeps no disposal flag, and it throws in only two places, neither reachable from this call. `dispose()` ends with `localValues.clear()`, and `getParent()` reads the parent out of `localValues`, so a disposed context reports no parent. `activate()` returns immediately when the parent is null, and `activateBranch()` is a loop over `getParent()`, so restoring a disposed context does one no-op iteration and ends. A `try`/`catch` would guard against nothing while hiding real failures if this code changed later.

The chain cannot be left inconsistent either. The worst reachable case is restoring a context that is mid-disposal, which writes it back as the window's active child, and `EclipseContext.dispose()` clears exactly that itself via `if (this == parent.getActiveChild()) parent.set(ACTIVE_CHILD, null)`. `FindReplaceOverlayContextSupport.dispose()` also restores before disposing the overlay context, so the ordering is right.

Clearing the field is still worth doing, for a different reason: it makes the invariant explicit that `contextToRestore` is only meaningful while the overlay owns the active leaf, and it stops the overlay holding the host editor's part context after focus is handed back.

docs/adr/0001-find-replace-overlay-key-handling.md
```diff
@@ -0,0 +1,321 @@
...
+introduced the embedded composite for these reasons). **A separate shell is therefore out of
+scope**, which constrains the whole design space below.
+
+The consequence of being embedded is the root of the problem: focus in the overlay's fields does
+not change the active workbench part. The editor stays active, so its key-binding scopes and its
+command handlers stay in effect and compete with the overlay for every keystroke.
+
+### Goals
+
+1. Standard text editing keys (Ctrl+C/V/X/Z/A, Ctrl+Backspace, word navigation, Home/End,
```
r-mennig: Nitpicking: Isn't Ctrl+Y (Redo) missing here?

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/overlay/FindReplaceOverlayContextSupport.java
```diff
@@ -0,0 +1,244 @@
...
+ * activating a sibling of that context takes off the chain. Its retargetable
+ * actions live in the window context instead, guarded by an <em>expression</em>
+ * over {@link ISources#ACTIVE_PART_ID_NAME}, which publishing this context's own
+ * id as the active part id makes false. Together they leave no editor handler in
+ * the resolution path, so no command has to be suppressed individually. The
+ * context sits below the window context rather than the application context,
+ * which would detach window-scoped commands and services as well.
+ * <p>
+ * The {@link MPart} looks superfluous, since nothing reads it, but removing it
+ * changes behaviour: it is what lets {@code ActivePartLookupFunction} resolve an
```
r-mennig, line 52: Nitpicking: "Behaviour" (with an 'u') is the British spelling of the word. I think we should use the American spelling here, which is "behavior".

bundles/org.eclipse.ui.workbench.texteditor/plugin.xml
```diff
@@ -1561,7 +1561,7 @@
       <context
             id="org.eclipse.ui.workbench.texteditor.findReplaceOverlay"
             name="%context.findReplaceOverlay.name"
-            parentId="org.eclipse.ui.textEditorScope">
+            parentId="org.eclipse.ui.contexts.window">
```
r-mennig, line 1564: General question: Is there a fundamental restriction which prevents this overlay from being available in dialogs? Otherwise this could be changed to `org.eclipse.ui.contexts.dialogAndWindow` to support the overlay there too.

r-mennig (approved):
I've tested these changes inside the integrations of the FindReplaceOverlay in the `CompilationUnitEditor` (writeable) and the `ClassFileEditor` (readonly) and found no issues.

The code looks sound to me as well (I've only found some nitpickings and a potential improvement). With the restriction to the `AbstractTextEditor` removed, this now allows the overlay to be used in other editors as well. Thank you for this change!

bundles/org.eclipse.ui.workbench.texteditor/plugin.xml
HeikoKlare: Valid question and I first also thought that it might make sense to be more generic here. However, we want to inherit commands from the Window scope and while `FindReplaceOverlayContextSupport.createOverlayContext()` ensures this via its context creation, generalization to the `dialogsAndWindows` scope in the plugin.xml would lead to an kind of inconsistency between the hierarchy specified for the scope via plugin.xml (and visible in the Keys preferences) and what is later established by theimplementation.
That code is also the essential point why using the overlay inside dialogs is currently not possible anway: the overlay is tied to a workbench part and mechanisms around it, which does not apply to dialogs. So to make the overlay available in dialogs, a larger generalization or parameterization would be necessary anyway. So the `dialogAndWindow` may become the right choice only once the overlay is genuinely offered for dialog-hosted find/replace targets.

PR #4294, opened 2026-08-29
Paint the range indicator with alpha instead of a stipple image

vogella:
The range indicator in the vertical ruler was drawn from a 1-bit image with every other pixel set, a stipple that fakes a 50 percent tint. On scaled displays the pattern is scaled in blocks of zoom/100 pixels, so at 200 percent the ruler shows a coarse checkerboard in the indicator colour instead of a tint.

The range is now filled with the indicator colour at half alpha, which renders as a smooth tint at every zoom level and keeps the solid one pixel lines at the top and bottom of the range. The cached image, its dispose listener and the recreation on resize or colour change go away with it, so the class shrinks from 176 to 91 lines.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.workbench.texteditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.workbench.texteditor; singleton:=true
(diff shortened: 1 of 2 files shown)


vogella:
dark-theme-range-indicator-before-after


light-theme-range-indicator-before-after

Custom theme (Dracula)

dracula-theme-range-indicator-before-after

Custom theme (AI Neon)

ai-neon-theme-range-indicator-before-after

BeckerWdf:
Cool change. That's a good catch!

HeikoKlare:
I also really like this. I remember that we have adapted the code to produce proper patterns in HiDPI scenarios some time ago, but just making the indicator visually plain is so much better.

PR #4307, opened 2026-08-31
Remove the heap status border

### LINKED ISSUE #4306: Heap status shows dark gray borders
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/4280

Heap status shows now very prominent dark gray border around in the light theme on Linux. 

Before:

Image

Now:

Image

vogella:
Since #4280 the heap status border and used-memory bar are a 50/50 blend of foreground and background, which on the light GTK theme draws a prominent mid-gray frame around the control. Following the discussion in #4306 the border is removed entirely, and the bar and its separators mix 40% foreground into the background, close to the old hardcoded gray. The separators are painted with fillRectangle instead of drawLine, because cairo leaves the end pixels of a 1px line half covered, which showed as lighter corners once the border no longer hid them.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/4306

### DIFF
--- a/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/HeapStatus.java
+++ b/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/HeapStatus.java
@@ -331,10 +331,19 @@ public class HeapStatus extends Composite {
 		}
 	}
 
-	private static Color blend(Color c1, Color c2) {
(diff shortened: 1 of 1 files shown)


vogella:
heap-status-noborder-final

iloveeclipse:
Tested in IDE, looks good. Thanks.

vogella:
Also fine in my custom themes from https://github.com/vogellacompany/eclipse.themes

heap-status-noborder-vogella-themes

PR #4308, opened 2026-08-31
Remove unused internal classes from org.eclipse.ui.workbench

vogella:
Deletes seven internal classes in org.eclipse.ui.workbench that are referenced neither from Java nor from plugin.xml: CacheWrapper, ILegacyAttributeNames, LightweightActionDescriptor, NoKeysPreferencePage, PropertyMapUnion, QuickMenuHandler and ThemeAdapter. Most are leftovers of the pre-e4 presentation and legacy command layers.

Validated in the IDE workspace with no additional warnings or errors, and with a full aggregator build.


vogella:
These classes are unused in the top level Eclipse project, merging early so that consumer of this internal API find this fast and can notify us or fix their code.

PR #4309, opened 2026-08-31
Remove unused internal classes from org.eclipse.ui.ide

vogella:
Deletes six internal classes in org.eclipse.ui.ide that are referenced neither from Java nor from plugin.xml: DeltaMarkerEntry, MarkerSeverityField, NonUIThreadProgressMonitorPart, TaskMarker, TypeFieldGroup and TaskListMessages. TaskListMessages was the last type in org.eclipse.ui.internal.views.tasklist, so its messages.properties and the package export in the manifest are removed as well.

Validated in the IDE workspace with no additional warnings or errors, and with a full aggregator build.


vogella:
These classes are unused in the top level Eclipse project, merging early so that consumer of this internal API find this fast and can notify us or fix their code.

PR #4314, opened 2026-08-31
Relax LargeFileTest marker timing factor #2622

### LINKED ISSUE #2622: Random failing LargeFileTest
https://download.eclipse.org/eclipse/downloads/drops4/I20241211-1800/testresults/html/org.eclipse.ui.editors.tests_ep435I-unit-macosx-x86_64-java17_macosx.cocoa.x86_64_17.html

```
java.lang.AssertionError: Opening large file took too long: 644.8726ms with marker vs. 306.91605ms without
at org.eclipse.ui.editors.tests.LargeFileTest.openLargeFileWithAnnotation(LargeFileTest.java:135)
```

cc at author of that test: @tomaswolf 
maybe just increase timeout?

AzazelSensei:
LargeFileTest.openLargeFileWithAnnotation fails on macOS CI when opening with a marker takes a little over 2x the baseline (about 0.6-1.0s vs 0.3-0.5s). That is machine jitter, not a 15s OverviewRuler freeze.

I widened the relative limit from 2x to 5x so the test still flags a real freeze without flaking.

Fixes #2622

### DIFF
--- a/tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/editors/tests/LargeFileTest.java
+++ b/tests/org.eclipse.ui.editors.tests/src/org/eclipse/ui/editors/tests/LargeFileTest.java
@@ -130,9 +130,9 @@ public class LargeFileTest {
 		page.closeEditor(part, false);
 		TestUtil.runEventLoop();
 		// Be generous here; all this timing is approximate. Fail if the attempt
-		// with the marker takes more than twice as long. If the OverviewRuler
+		// with the marker takes more than five times as long. If the OverviewRuler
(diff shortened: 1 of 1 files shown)


vogella:
I think it has not been failing in a while and I'm currently making the marker handling faster. Why this PR at this point in time?

AzazelSensei:
I cannot run the macOS unit tests from this machine. If LargeFileTest is already stable on current I-builds I can close this.

vogella:
Yes, I think it is not necessary to increase the timeout at the moment. Thanks for the attempt.

PR #4318, opened 2026-08-31
Add an expand/collapse hover button in search results tree view

elsazac:
The search results view (File Search, Java Search, Git Search, etc.) in tree layout mode provides only two toolbar options called Expand All and Collapse All. This result in expand and collapse of entire tree at once. If one wants to expand or collapse fully, only a single project individually, it is not possible. This is really hard when there are lot of projects in workspace.

This PR adds an expand/collapse hover option to the search results tree view. when the user hovers over a top-level project node in the tree layout, a small button now appears immediately to the right of the project name. The button shows the expand icon if the project is collapsed, or the collapse icon if it is already expanded. clicking it fully expands or collapses that project. The button disappears  automatically when the cursor moves away.

https://github.com/user-attachments/assets/f799cdc3-f5b5-4455-9586-9b5c69070c6f


BeckerWdf:
> This PR adds an expand/collapse hover option to the search results tree view. when the user hovers over a top-level project node in the tree layout, a small button now appears immediately to the right of the project name.

I like it from a functional point of view. But I am very unsure about the UI interaction. For me this feels very hidden and I also don't know another place in Eclipse where we have this "icon appears on hover". Why not simply adding the expand collapse icons to the view's toolbar.

elsazac:
> I like it from a functional point of view. But I am very unsure about the UI interaction. For me this feels very hidden and I also don't know another place in Eclipse where we have this "icon appears on hover". Why not simply adding the expand collapse icons to the view's toolbar.

Expand All and Collapse All for the projects are already available in the toolbar, so adding the same icons again for each individual project could be confusing for users.Thats why separated it away from the toolbar. The  action was specifically to allow users to fully expand just one project, rather than the entire tree. so felt it would be convenient to have it somewhere near the project itself.  but I am open to suggestions if there’s a more better way to represent this within the existing view.

elsazac:
@BeckerWdf Just wanted to check if you have any concerns with this approach, or if you think it would be better to move the action into the project’s context menu instead?

iloveeclipse (changes requested):
I don't like two things, therefore -1 for merge:

1) There are collapse/expand buttons on the left side of every tree element. The current PR adds more code only to replicate the functionality already available. It is more code to maintain with little added benefit for users.
2) Main issue. **If** this should be considered a cool feature: for the IDE consistency, it should be available for all tree clients  in the IDE without clients code change (but with possibility to opt-out) and it should be controlled by a dedicated IDE preference so that it can be switched on/off if needed.

BeckerWdf:
> I don't like two things, therefore -1 for merge:
> 
> 1. There are collapse/expand buttons on the left side of every tree element. The current PR adds more code only to replicate the functionality already available. It is more code to maintain with little added benefit for users.
> 2. Main issue. **If** this should be considered a cool feature: for the IDE consistency, it should be available for all tree clients  in the IDE without clients code change (but with possibility to opt-out) and it should be controlled by a dedicated IDE preference so that it can be switched on/off if needed.

I agree..

trancexpress:
> There are collapse/expand buttons on the left side of every tree element. The current PR adds more code only to replicate the functionality already available. It is more code to maintain with little added benefit for users.

The new buttons actually expand the tree element fully, instead of only one level - which the existing buttons do?

iloveeclipse:
> The new buttons actually expand the tree element fully, instead of only one level - which the existing buttons do?

Correct, I've missed that the proposed "expand" button does more. My main issue is however is the consistency with the rest of the IDE.

SougandhS:
> I don't like two things, therefore -1 for merge:
> 
> 1. There are collapse/expand buttons on the left side of every tree element. The current PR adds more code only to replicate the functionality already available. It is more code to maintain with little added benefit for users.
> 2. Main issue. **If** this should be considered a cool feature: for the IDE consistency, it should be available for all tree clients  in the IDE without clients code change (but with possibility to opt-out) and it should be controlled by a dedicated IDE preference so that it can be switched on/off if needed.

this makes sense for cases with a lot of nested items - being able to fully expand just the parent you care about, instead of everything, saves a lot of scrolling to find what you actually wanted...

Would also be great to see this extended to Project Explorer and other tree views, not just Search. That'd probably address the consistency point too.

PR #4319, opened 2026-08-31
Style the progress bar in the dark theme

vogella:
The progress bar had no rule of its own in the dark theme: only the `ProgressIndicator` around it was coloured, so the bar fell to the `Composite > *` catch-all on `DARK_BACKGROUND` and did not separate from the chrome.

Gives it `#5A5F62` as the trough and `LINK_COLOR` as the fill. The `.MPart` copies live in `e4-dark_partstyle.css` because they only tie `.MPart Composite > * > *` on specificity and need to be imported after it.

Verified on Windows in the Progress view, the status line and the p2 install wizard: fill and trough resolve to the two colours at every fill level.


vogella:
Windows neu:

styled-progress-bars10

styled-progress-bars20

BeckerWdf:
how did it look before?

PR #4320, opened 2026-09-01
Broadcast the import delta when the creation operation ends

vogella:
Smart import creates the selected projects in one workspace operation. That operation passed `IWorkspace.AVOID_UPDATE`, which suppresses the resource change notification at the end of the operation and leaves it to the workspace notification job, up to 1.5 seconds later. Listeners such as Project Explorer therefore learned about the imported projects late, and `SmartImportTests.testProjectsAreCreatedInASingleWorkspaceOperation` failed on the I20260831-2300 Linux build because it saw no event at all when it asserted right after the job returned.

Dropping the flag broadcasts the delta synchronously when the operation ends. It is still a single delta covering all projects, which is what batching the creation loop was introduced for, so listeners are notified once and immediately instead of once and late.

Verified with `SmartImportTests` (14 tests, no failures) run headless under Xvfb with `org.eclipse.ui.ide` in the reactor.


vogella:
Fixes https://download.eclipse.org/eclipse/downloads/drops4/I20260831-2300/testresults/html/org.eclipse.ui.tests_ep442I-unit-linux-x86_64-java26_linux.gtk.x86_64_26.html

vogella:
Fixes an error reported in the integration build.

PR #4324, opened 2026-09-01
Log CSS engine errors through ILog instead of a pluggable handler

vogella:
CSSErrorHandler was an indirection with a single production implementor, ThemeEngineManager, which did nothing but forward to the platform log. Its default, CSSErrorHandlerImpl, called printStackTrace(), so any embedder that did not install a handler lost the errors to stderr or dropped them. Each site now logs through ILog directly, which puts the entry under the bundle it came from and lets the message name the property or the phase that failed.

CSSErrorHandler, CSSErrorHandlerImpl and the handleExceptions plumbing are gone. This removes API, but the package is x-friends only and nothing outside the engine and its tests referenced it. The SWT test engine and the doc sample drop their last callers.

Verified with the org.eclipse.e4.ui.tests.css.core and org.eclipse.e4.ui.tests.css.swt suites (217 tests, 9 skipped) on Linux after the rebase onto the merged #4321 and #4322.


akurtakov:
Thanks for this one! I'm really tired of "logging" generating so much trouble for really no benefit like it happened with jsvg (https://github.com/eclipse-platform/eclipse.platform.ui/issues/4328) to a disastrous result.

PR #4331, opened 2026-09-01
Style CTabFolder pages that are hidden when they are skinned

vogella:
The Search dialog opens unstyled in the dark theme and only picks up the theme once a tab is switched. A CTabFolder exposes only the page of its selected tab to the CSS engine, and applyStyles skips elements that are not visible. SearchDialog attaches its page before selecting the tab, so the page is hidden during the skin pass, and the programmatic setSelection afterwards fires no selection event. A page whose creation runs the event loop is skinned before it is attached at all, and attaching it to an already selected tab sends no show event either.

CSSSWTApplyStylesListener now watches a skipped CTabFolder child for the resize or show that makes it the selected tab's page and styles it then. CTabFolderElement is the only ChildVisibilityAwareElement, so this covers the whole class of the problem.

Two CTabFolderTest cases mirror the dialog's two sequences and fail without the fix. The swt-simple test is dropped, curved tabs are gone and getSimple() always returns true. Verified with CTabFolderTest (21 tests) on Linux.


vogella:
Before:

pr4331-before

After:
 
pr4331-after-screen

PR #4332, opened 2026-09-01
Ask before closing or opening projects nested below the selection

vogella:
Close Project and Open Project acted on exactly the selected projects, so a project that physically contains other projects, such as a Maven or Gradle root, left its modules open, and reopening the root brought back only the root. Open Project then asked about referenced projects from inside its job, a second dialog for the same action.

Both actions now report the nested projects in one dialog with an "Include nested projects" check box (off by default) and the usual Close or Open and Cancel buttons. Open Project asks about referenced projects in that same dialog with a second check box backed by the existing preference, so there is at most one question per action, and Cancel now opens nothing instead of leaving the first project open. A Remember my decision toggle stores each answer in its workspace preference, shown next to the existing settings. Nothing is asked when neither applies, and Close Unrelated Projects keeps its own confirmation.

This supersedes #4344, which is now part of this change. Verified against a workspace where one root project contains 757 nested projects shown in the hierarchical Project Explorer: closing the root closed all 758, opening it brought all 758 back. The tests answer the dialog programmatically and cover both check boxes, the remembered answers in both directions, and Cancel.


vogella:
nested-projects

PR #4334, opened 2026-09-02
Remove resource filters injected by the Java Language Server

vogella:
83 .project files carry the same generated resource filter excluding node_modules, .git and the __CREATED_BY_JAVA_LANGUAGE_SERVER__ sentinel. That filter is written by the JDT Language Server behind the VS Code Java extension into every project it imports, and it reached master accidentally in 2023 with 6af77694a28. Nobody asked for it and it does not describe anything about these projects.

It is also the worst shape a filter can have: inheritable, anchored at the project root, excluding both files and folders, so the workspace runs the regex matcher against every child of every directory it refreshes, in every one of those 83 projects, for no benefit. Removing it makes resource refresh cheaper and stops the files from drifting whenever someone opens the repo in another editor.

The hand-authored filter in org.eclipse.text.quicksearch, which excludes target, is deliberate and stays. Only XML metadata is removed, no code or build input is touched, and all 83 files still parse.

### DIFF
--- a/bundles/org.eclipse.core.commands/.project
+++ b/bundles/org.eclipse.core.commands/.project
@@ -37,15 +37,4 @@
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.pde.api.tools.apiAnalysisNature</nature>
 	</natures>
-	<filteredResources>
-		<filter>
-			<id>1676382308243</id>
-			<name></name>
-			<type>30</type>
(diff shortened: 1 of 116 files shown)


merks:
Seems like a fine idea!

vogella:
> Seems like a fine idea!

Definitely, also this creates a measured performance degration. Sorry for adding this stuff in 2023 during the merge of the Git repos.

PR #4335, opened 2026-09-02
Simplify FontRegistry's font lookup to direct fallbacks

HeikoKlare:
Looking up a font for a symbolic name assembled its result in a mutable record variable across a chain of null checks, which conflated the two distinct reasons for ending up with the default font: no font data being registered for the name at all, and no font being creatable from the data that is registered. Only the latter is an error worth logging, which the chained conditions made hard to see.

Each case now returns its font directly, and both fallbacks go through the accessor for the default font in a given style rather than resolving a record and applying the style afterwards.

The fallback for the error case is still resolved before the error is logged, but now includes realizing the requested style rather than only looking up the record providing it. As a result, the message announcing that the default font is used instead can no longer be logged on a path that fails to provide any font at all and thus ends in an exception.

### DIFF
--- a/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
+++ b/bundles/org.eclipse.jface/src/org/eclipse/jface/resource/FontRegistry.java
@@ -856,25 +856,27 @@ public class FontRegistry extends ResourceRegistry {
 			return existingRecord.get(style);
 		}
 
-		FontData[] existingFontData = stringToFontData.get(symbolicName);
-
-		FontRecord fontRecord;
-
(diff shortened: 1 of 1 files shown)


akurtakov:
Changes like https://github.com/eclipse-equinox/p2/commit/1d6050bba230efc09cb8c32eda17e9b50c7ab280 will have to be done to all repos to get jenkins build working again. I go through things as I face them and ensure that tycho-build and etc. are also current. Help is more than welcome though.

akurtakov:
Rebasing on https://github.com/eclipse-platform/eclipse.platform.ui/pull/4350 once it in should get the things back in shape.

PR #4354, opened 2026-09-07
Let themes declare that they are dark

vogella:
Themes can now declare their appearance with an `isDarkTheme` attribute on the theme extension point instead of the platform guessing it from the theme id. Themes that do not set it keep the old id based classification, so existing contributions behave as before.

`ITheme.isDark()` is added as a default method, so implementors are not broken. The theme engine uses it for the SWT appearance preference and when picking a dark theme to inherit the operating system setting, which means a product shipping its own dark theme is now handled instead of only `org.eclipse.e4.ui.css.theme.e4_dark`. The appearance preference page only recommends a restart when the switch actually crosses between light and dark, so switching between two dark themes no longer interrupts with a dialog. The flag is persisted next to the theme id, which lets `IDEApplication` style the workspace selection dialog from the recorded value, the only thing it has before the workbench and the theme engine exist.

This supersedes #2808 by @BeckerWdf, which had the same idea and where the `isDarkTheme` attribute name comes from. That PR no longer applies since the platform specific dark theme processors it patches have been removed in the meantime.

### DIFF
--- a/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
+++ b/bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
@@ -30,6 +30,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.FileLocator;
@@ -88,6 +89,13 @@ public class ThemeEngine implements IThemeEngine {
 	private final HashMap<String, List<IResourceLocator>> sourceLocators = new HashMap<>();
 
 	private static final String THEMEID_KEY = "themeid";
+
+	/**
+	 * Prefix of the key recording a theme's appearance. The theme id is part of the
+	 * key, so a record left behind by another installation sharing the preferences
+	 * cannot be mistaken for the appearance of the theme in {@link #THEMEID_KEY}.
+	 */
+	private static final String THEME_IS_DARK_KEY_PREFIX = "themeIsDark.";
 
 	public static final String THEME_PLUGIN_ID = "org.eclipse.e4.ui.css.swt.theme";
 
@@ -157,13 +165,14 @@ public class ThemeEngine implements IThemeEngine {
 						final String themeBaseId = id + version;
 						String themeId = themeBaseId;
 						String label = ce.getAttribute("label");
+						boolean dark = isDarkTheme(ce);
 						String originalCSSFile;
 						String basestylesheeturi = originalCSSFile = ce.getAttribute("basestylesheeturi");
 						if (!basestylesheeturi.startsWith("platform:/plugin/")) {
 							basestylesheeturi = "platform:/plugin/" + ce.getContributor().getName() + "/"
 									+ basestylesheeturi;
 						}
-						registerTheme(themeId, label, basestylesheeturi, version);
+						registerTheme(themeId, label, basestylesheeturi, version, dark);
 
 						//check for modified files
 						if (modifiedFiles != null) {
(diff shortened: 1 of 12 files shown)


bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
```diff
@@ -264,13 +287,18 @@ public synchronized ITheme registerTheme(String id, String label, String basesty
...
+	}
+
+	public synchronized ITheme registerTheme(String id, String label, String basestylesheetURI, String osVersion,
+			boolean dark) throws IllegalArgumentException {
 		for (Theme t : themes) {
 			if (t.getId().equals(id)) {
 				throw new IllegalArgumentException("A theme with the id '" + id
 						+ "' is already registered");
 			}
 		}
```
BeckerWdf: couldn't we use streams here just like some lines below?

`getThemes().stream().filter...`

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
BeckerWdf: also here we could use streams

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
```diff
@@ -654,12 +684,42 @@ public void restore(String alternateTheme) {
...
 
-		String themeToRestore = overrideWithDarkTheme ? E4_DARK_THEME_ID : alternateTheme;
+		String themeToRestore = overrideWithDarkTheme ? darkTheme.get().getId() : alternateTheme;
 		if (themeToRestore != null) {
 			setTheme(themeToRestore, false);
 		}
 	}
 
+	/**
+	 * Whether the theme recorded in the preferences is a dark one. Preferences
```
BeckerWdf: it looks like the method is not specific to an themeId recorded in preferences. So the comment could be more precise here or if it's specific to perfrenence keys then the method name could be more speaking.

bundles/org.eclipse.e4.ui.css.swt.theme/src/org/eclipse/e4/ui/css/swt/internal/theme/ThemeEngine.java
```diff
@@ -654,12 +684,42 @@ public void restore(String alternateTheme) {
...
+	 * written before themes declared their appearance only carry the theme id.
+	 */
+	private static String darkKey(String themeId) {
+		return THEME_IS_DARK_KEY_PREFIX + themeId;
+	}
+
+	private boolean isPreferenceThemeDark(String prefThemeId) {
+		boolean darkById = prefThemeId.contains("dark"); //$NON-NLS-1$
```
BeckerWdf, line 699: here we check for "contains" in "darkKey" we check for a prefix. Is this correct?

vogella: Different things: THEME_IS_DARK_KEY_PREFIX builds the preference key themeIsDark.<themeId>, while contains("dark") is the old id heuristic used as the default when no appearance was recorded yet. The misplaced Javadoc made them look related, that is fixed now.

vogella:
PR updated with the proposals, except one which was IMHO just a misunderstanding caused by wrong Javadoc, fixed now.

BeckerWdf (approved):

vogella:
Thanks @BeckerWdf for the review

BeckerWdf:
@vogella  in https://github.com/eclipse-platform/eclipse.platform.ui/pull/4356#issuecomment-5573221809 you said that with the PR merge a restart is not longer needed when switching between light and dark.
I just tested with yesterday's I-build (https://download.eclipse.org/eclipse/downloads/drops4/I20260913-2300).

But I still see the "restart" dialog when switching between dark and light:
image

vogella:
> @vogella in [#4356 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/4356#issuecomment-5573221809) you said that with the PR merge a restart is not longer needed when switching between light and dark. I just tested with yesterday's I-build (https://download.eclipse.org/eclipse/downloads/drops4/I20260913-2300).
> 
> But I still see the "restart" dialog when switching between dark and light: image dark) or different light themes (light -> light). You can test by installing additional themes, e.g. from https://github.com/vogellacompany/eclipse-themes

BeckerWdf:
> I said no restart is required for different dark themes (dark -> dark) or different light themes (light -> light). You can test by installing additional themes, e.g. from https://github.com/vogellacompany/eclipse-themes

Ah ok. So then I mis-interpreted your text: 
> switching between different light and dark themes

Thanks for clarification.

BeckerWdf:
Btw: What would be needed to be fixed so that we no longer need to trigger a restart after a switch from light to dark theme?

PR #4356, opened 2026-09-07
Add a Themes submenu to Window > Appearance

vogella:
Switching the theme so far meant opening the appearance preferences. This adds a Themes submenu under Window > Appearance that lists the installed themes as radio items, applies the pick right away and remembers it for the next start. A switch between a light and a dark theme offers a restart, and a last entry records the active theme as the default for new workspaces.

The items are built in the workbench, where the theme engine is visible, and instantiated from the IDE menu through `ExtensionFactory` like the Show In menu. In high contrast mode no items are offered, matching the appearance preferences.

Stacked on #4354: the first commit is that PR, only the second one belongs here. Please merge #4354 first, this branch will then rebase down to a single commit.


merks:
That's quite some additional complexity to maintain forever.  Should we ask ourselves if users really switch often enough that this needs to be more convenient for them?

sratz:
> That's quite some additional complexity to maintain forever. Should we ask ourselves if users really switch often enough that this needs to be more convenient for them?

I think most of the complexity is because this PR builds on #4354. Only the top commit is the actual menu entry.

But I agree: Since switching theme requires a restart, is this a feature a people would actually use in practice?

akurtakov:
IMO poll/vote is best in the cases like this one where there is no agreement about a change in general.

That aside, the PR still needs some work as the case of theming engine is not considered (the menu should not be shown in this case for sure) but with this PR it is and looks like 
image

vogella:
@sratz switching between different light and dark themes does not require a restart anymore after we merge the is dark pr. We only require a restart at the moment because we do not know if a theme wants the native dark styling or not.

I park this as draft for the moment.

vogella:
> That's quite some additional complexity to maintain forever. Should we ask ourselves if users really switch often enough that this needs to be more convenient for them?

With the rebase, the code change is relatively small

BeckerWdf:
on macOS this currently looks like this: 
image

vogella:
This is how to looks like with additional community themes (for example https://github.com/vogellacompany/eclipse-themes).

Screenshot From 2026-09-14 11-01-37

vogella:
Here is how it looks like with additional community themes, like https://github.com/vogellacompany/eclipse-themes

Screenshot From 2026-09-14 11-01-37

vogella:
> That aside, the PR still needs some work

If theming is disable, the menu shows now the single option to enable it. 

Screenshot From 2026-09-15 09-36-47

PR #4357, opened 2026-09-07
Find/Replace overlay: renamed tooltips

r-mennig:
Currently, the tooltips for the buttons in the Find/Replace Overlay to find the next / previous / all hits are shown as "Search forward" / "Search backward" / "Search all". The search view however uses the tooltips "Show Next Match" / "Show Previous Match".

To closer align both sets of tooltips, this PR changes the "Search forward", "Search backwards" and "Search all" tooltips to "Find Next", "Find Previous" and "Find All" ("Find" is used here to signal to the user that a search is being executed, unlike in the search view in which all matches are already known).

Additionally, this PR also updates the capitalization of all find/replace overlay tooltips according to the [Eclipse UI style guide](https://eclipse-platform.github.io/ui-best-practices/#capitalization) (see Guideline 1.5). The placeholder texts in the search / replace bar are not changed, as those are already capitalized correctly according to Guideline 1.6.


r-mennig:
Ping @HeikoKlare

bundles/org.eclipse.ui.workbench.texteditor/plugin.properties
```diff
@@ -206,8 +206,8 @@ command.joinLines.name = Join Lines
 command.openHyperlink.name= Open Hyperlink
 command.openHyperlink.description= Opens the hyperlink at the caret location or opens a chooser if more than one hyperlink is available
 
-command.findReplaceOverlay.searchForward.name= Find and Replace: Search Forward
-command.findReplaceOverlay.searchBackward.name= Find and Replace: Search Backward
+command.findReplaceOverlay.searchForward.name= Find and Replace: Find Next
```
HeikoKlare, line 209: It seems reasonable to preserve the existing command names and not rename them to `findNext` or the like as well, because then existing command configuration would not be matched anymore. I just want to point out that this should be an explicit decision and not left as is by accident.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceMessages.properties
```diff
@@ -47,20 +47,20 @@ FindReplace_CloseButton_label=Close
...
-FindReplaceOverlay_upSearchButton_toolTip=Search backward
-FindReplaceOverlay_downSearchButton_toolTip=Search forward
-FindReplaceOverlay_searchAllButton_toolTip=Search all
-FindReplaceOverlay_searchInSelectionButton_toolTip=Only search in selected area
-FindReplaceOverlay_regexSearchButton_toolTip=Match regular expression pattern
-FindReplaceOverlay_caseSensitiveButton_toolTip=Match case
-FindReplaceOverlay_wholeWordsButton_toolTip=Match whole word
+FindReplaceOverlay_upSearchButton_toolTip=Find Previous
+FindReplaceOverlay_downSearchButton_toolTip=Find Next
+FindReplaceOverlay_searchAllButton_toolTip=Find All
```
HeikoKlare, line 52: The command name is "Select All Matches". I think we should unify both. I have no strong opinion which term to use. While `Select All Matches` is more accurate and fits to the naming in the find/replace dialog, an ordinary find actually also does a selection, so consistently calling it "find" might be best unification.

HeikoKlare (commented):
Thank you for the proposal. I like the consistent naming of using "find next/previous" instead of "search forward/backward" and would be in favor of merging this renaming.

bundles/org.eclipse.ui.workbench.texteditor/plugin.properties
r-mennig: Yes, not changing the command identifiers was a deliberate decision to not break any existing configuration. Thanks for pointing this out.

bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/internal/findandreplace/FindReplaceMessages.properties
r-mennig: I agree, consistently calling this functionality "Find ..." in any user-facing text seems like the best option.

I've update the PR.

HeikoKlare (approved):
Looks good now, thank you!

PR #4358, opened 2026-09-07
Skip javadoc generation for test bundles

vogella:
The `javadoc` profile that the Jenkins job enables generates and jars API docs for every module in the reactor, including the 33 test bundles. Nobody publishes those docs and the Jenkinsfile archives only logs, so the jars are built and then thrown away.

Measured locally, javadoc for the test and example bundles takes 123 s of the 411 s the whole reactor spends on javadoc, roughly 5 to 6 minutes on the slower CI agent. That is a useful cut for a job running against an 80 minute timeout that has been hitting it lately.

Javadoc stays enabled for the published bundles. Its doclint HTML and syntax checks are the only validation of doc comment markup in CI, since the GitHub Actions builds do not pass `-Pjavadoc` at all.

### DIFF
--- a/tests/org.eclipse.core.filebuffers.tests/build.properties
+++ b/tests/org.eclipse.core.filebuffers.tests/build.properties
@@ -28,3 +28,4 @@ output.. = bin/
 
 # Maven/Tycho pom model adjustments
 pom.model.property.testClass = org.eclipse.core.filebuffers.tests.FileBuffersTestSuite
+pom.model.property.maven.javadoc.skip = true
--- a/tests/org.eclipse.e4.core.commands.tests/build.properties
+++ b/tests/org.eclipse.e4.core.commands.tests/build.properties
@@ -20,3 +20,4 @@ bin.includes = META-INF/,\
                about.html,\
                plugin.properties
 src.includes = about.html
+pom.model.property.maven.javadoc.skip = true
--- a/tests/org.eclipse.e4.emf.xpath.test/build.properties
+++ b/tests/org.eclipse.e4.emf.xpath.test/build.properties
@@ -25,3 +25,4 @@ bin.includes = .,\
 jars.compile.order = .
 source.. = src/
 output.. = bin/
+pom.model.property.maven.javadoc.skip = true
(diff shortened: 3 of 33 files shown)


vogella:
@akurtakov WDYT?

akurtakov:
Examples/snippets are smth I consider as end user visible content (via links like https://eclipse.dev/eclipse/swt/snippets/ or direct ones in some documentation) so I would rather keep them with good javadoc.

merks:
It sounds like a good idea.  Nothing speaks against this does it @HannesWell ?

vogella:
Thanks for the feedback @akurtakov, closing therefore.

merks:
I see. What about tests?

vogella:
> It sounds like a good idea. Nothing speaks against this does it @HannesWell ?

Ah, sorry, I think that is the opposite that @akurtakov suggested, please reopen if we should continue here. I also noticed that the aggregator configures zipping the Javadoc here but at least in CI the zips are not used, just thrown again, this might also be a simplification for speedup for the build

akurtakov:
Reopening and making it clear - skip javadoc in test bundles only.

vogella:
> Reopening and making it clear - skip javadoc in test bundles only.

PR is updated to cover only the tests.

HannesWell:
For tests I agree this is fine. Don't see anything that speaks against it.

> Examples/snippets are smth I consider as end user visible content (via links like https://eclipse.dev/eclipse/swt/snippets/ or direct ones in some documentation) so I would rather keep them with good javadoc.

For these kind of bundles, maybe only assembling the jars and generating the pages could be skipped but the javadoc could maybe still be checked. Not producing jars should be possible, but I don't know if there is an even lighter method to check the javadoc than to generate the pages.
And it would also be up investigate how much that actually saves and how much complexity it adds to the build.

akurtakov (approved):

akurtakov:
Merging this one as this should be the major improvement from this PR. Further enhancements can happen in follow up PRs .

HannesWell:
@vogella, @akurtakov, before you sweep through all SDK repositories: Have you considered adding a profile to the parent pom that activates automatically for `eclipse-test-plugin` and skipps javadoc generation once and for all (test-plugin) children?

vogella:
> @vogella, @akurtakov, before you sweep through all SDK repositories: Have you considered adding a profile to the parent pom that activates automatically for `eclipse-test-plugin` and skipps javadoc generation once and for all (test-plugin) children?

Good idea, https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4036

PR #4359, opened 2026-09-07
Re-add inlined annotations a client removed from the annotation model

vogella:
InlinedAnnotationSupport kept every annotation of its last render in `fInlinedAnnotations` and treated all of them as attached. Once a client removed a code mining annotation from the annotation model directly, that annotation object was reused for every later render at the same position but never put back into the model, so the minings there stayed invisible for the rest of the editor session. Annotations the model no longer knows are now added again; the removal side is unchanged. The fallback for annotation models without `IAnnotationModelExtension` now removes only the annotations that are actually gone, rather than dropping all of them and restoring only the newly added ones.

Found while working on the unified diff of `org.eclipse.compare` (https://github.com/eclipse-platform/eclipse.platform/pull/2791), whose side is already fixed. The framework still swallowed the minings for any other client doing the same, which the new `CodeMiningAnnotationReattachTest` covers for a synchronous and an asynchronous provider.

### DIFF
--- a/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
+++ b/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
@@ -16,7 +16,6 @@ package org.eclipse.jface.text.source.inlined;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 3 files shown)


tobiasmelcher (approved):
The new test has not been added to the test suite. Is this intentional?

Apart from that, the change looks good to me and makes sense. I successfully tested it with the Predictive Code Completion feature in ABAP Development Tools, which makes extensive use of inlined annotations, and did not encounter any issues.

Thank you for providing this improvement.

vogella:
> The new test has not been added to the test suite. Is this intentional?

No, I forgot to add it. Will do so now. 

> Apart from that, the change looks good to me and makes sense. I successfully tested it with the Predictive Code Completion feature in ABAP Development Tools, which makes extensive use of inlined annotations, and did not encounter any issues.
> 
> Thank you for providing this improvement.

Thanks.

tobiasmelcher (approved):

PR #4360, opened 2026-09-08
Coalesce editor re-styles caused by preference bursts

vogella:
A theme switch writes every preference key on its own, and each key made every open editor re-run its presentation reconciler: `AbstractTextEditor` invalidated the whole document, `SourceViewerDecorationSupport` repainted the annotation painter over the highlighted range. With dozens of keys per switch that cost adds up per editor.

Both listeners now schedule at most one re-style per UI event loop turn, so a burst of preference changes costs one whole-document and one annotation re-style per editor instead of one per key. The painter's colors are still applied synchronously, only the repaint is deferred, so nothing is displayed with stale colors.

Verified that `org.eclipse.ui.workbench.texteditor` compiles. Note that in the theme-switch profile I took, the platform path this changes is only a small share of the stall; the bulk sits in JDT's semantic highlighting, which is https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/3142.

### DIFF
--- a/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -2543,6 +2543,8 @@ public abstract class AbstractTextEditor extends EditorPart
 	private SourceViewerConfiguration fConfiguration;
 	/** The editor's source viewer. */
 	private ISourceViewer fSourceViewer;
+	/** Whether a full re-style of the text presentation is already scheduled. */
+	private boolean fInvalidateTextPresentationPending;
 	/**
 	 * The editor's selection provider.
 	 *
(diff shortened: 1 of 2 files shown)


BeckerWdf:
That sounds interesting. But isn't this a general issue that preference changes are "published" one by one? The change looks like it "only" fixes the issue for text editors. What about other parts of the UI? Don't they suffer from the same issue? If yes: Can't we fix it at the central place once for everything (e.g. by changing preferences / publishing them in one bulk-call)?

vogella:
> That sounds interesting. But isn't this a general issue that preference changes are "published" one by one? The change looks like it "only" fixes the issue for text editors. What about other parts of the UI? Don't they suffer from the same issue? If yes: Can't we fix it at the central place once for everything (e.g. by changing preferences / publishing them in one bulk-call)?

Maybe. This would need a detailed analysis, changing such a central behavior is very dangerous.

BeckerWdf:
> Maybe. This would need a detailed analysis

Ok. Let's see.

PR #4361, opened 2026-09-08
Show every line header code mining after a fold collapsed

vogella:
In a `ProjectionViewer` with collapsed folds only a single `LineHeaderCodeMining` was drawn, even though every mining was created and attached to the annotation model. On screen the line numbers jumped across a collapsed region with nothing in the gap, and a mining that came into view later never appeared at all.

`InlinedAnnotationSupport` caches the document offset range the view port shows and the drawing strategy skips annotations outside it. That cache is refreshed on a view port, resize or document event. Collapsing a fold is none of those, since the document is untouched and the view port does not move, and a programmatic scroll reports no view port event either. So the range kept an old value and every mining behind it was silently dropped.

The range is now recomputed on each use from the display thread, which is cheap and self correcting, and invalidated on a text change for the uses off the display thread.


BeckerWdf:
Hi Lars,
can you provide a screenshot or screencast showing the issue?

vogella:
> Hi Lars, can you provide a screenshot or screencast showing the issue?

See https://github.com/eclipse-platform/eclipse.platform/pull/2791, the folding does not re-appear after unfolding.

PR #4363, opened 2026-09-08
Remove the font properties the CSS engine never implemented

vogella:
`font-adjust`, `font-stretch` and `font-variant` were registered as handled properties, but every apply method threw `UnsupportedPropertyException` and every retrieve method returned `null`. The engine swallows that exception, so a style sheet using one of them was ignored without a warning.

SWT `FontData` has no stretch axis, no small-caps and no size-adjust, so none of the three can be implemented. This drops the registrations, the handler methods and the `CSS2FontProperties` fields, which were only ever set to `null`. No shipped style sheet used them, the affected package is `x-friends`, and there is no visual change.


BeckerWdf:
always good to get rid of not uses code.
I assume existing themes out there that declared them will silently ignore these (without having any effect) also after this PR.

vogella:
We throw at the moment an Unsupported Operation exception if someone tries to use it, so I doubt that this is used.

BeckerWdf:
> We throw at the moment an Unsupported Operation exception if someone tries to use it, so I doubt that this is used.

So then even better..

PR #4370, opened 2026-09-10
Update ActionSetDescriptor to use IEclipsePreferences

deepika-u:
Updates `ActionSetDescriptor` to read and write its action-set visibility preference using `IEclipsePreferences` instead of the
old `Preferences` API.

Changes
- `isInitiallyVisible()`: now reads the preference through `Platform.getPreferencesService()` instead of the instance-scope node directly, so the lookup resolves through the standard scope search order rather than only checking one scope.

- `setInitiallyVisible()`: now writes the preference via `InstanceScope.INSTANCE.getNode(...)` and flushes the change explicitly, logging a failure if the flush doesn't succeed.

- Replaced `WorkbenchPlugin.PI_WORKBENCH` with `PlatformUI.PLUGIN_ID` in both preference methods and the constructor's error status. `WorkbenchPlugin.PI_WORKBENCH` is only an alias for `PlatformUI.PLUGIN_ID`, and `WorkbenchPlugin` already marks it with a TODO to be replaced by direct use of `PlatformUI.PLUGIN_ID`.


deepika-u:
@merks : When you get some time, could you take a look at this please?

merks (approved):
I assume you've ensured that the same preference node is use such that existing preference values are preserved. I guess the old implementation didn't save/flush the preference until shutdown or did that also happen immediately and automaticlaly.

deepika-u:
> I assume you've ensured that the same preference node is use such that existing preference values are preserved. I guess the old implementation didn't save/flush the preference until shutdown or did that also happen immediately and automatically.

@merks : Thanks alot for your time. 
Confirmed - InstanceScope.INSTANCE.getNode(PlatformUI.PLUGIN_ID) is the same node the old getPluginPreferences() bridge used, so existing values are preserved.

On the save timing: the old implementation didn't flush per write. It relied on ActionSetDescriptor → WorkbenchPlugin → AbstractUIPlugin#stop() → savePluginPreferences(), which saves the whole plugin's preferences in one batch at shutdown. Our explicit flush() here saves this value immediately instead, so it also survives an abnormal shutdown. I searched across the source and found no internal callers of setInitiallyVisible at all - only the declaration and interface signature - so the extra flush is effectively free in this codebase. Happy to drop it and rely on the existing shutdown-time save if you'd prefer to keep behavior identical to before.

merks (approved):
I think it's fine as is.  I just wanted confirmation and I think it's find to flush the value immediately too.

Do you need someone to merge?

deepika-u:
> I think it's fine as is. I just wanted confirmation and I think it's find to flush the value immediately too.
> 
> Do you need someone to merge?

@merks  - Thanks alot for your time and inputs. Can you go ahead merging this please as i dont have merge access.

Issue #4291, opened 2026-08-28
Define colors in platform to be used in other areas


vogella:
During work on PDE styling via https://github.com/eclipse-pde/eclipse.pde/pull/2432 I noticed that PDE defines a lot of color constants (for syntax highlighting in target editor, Manifest and e4 editor). Also Ant does the same.

We should define some general constants and re-use them in these areas. 

We have: 

- org.eclipse.pde.ui (5 Keys)
- org.eclipse.pde.genericeditor.extension (6 colorDefinitions)
- org.eclipse.ant.ui (6 Keys)

And more

vogella:
| Component | Mechanism | Ids |
| --- | --- | --- |
| `org.eclipse.pde.ui` | preference keys | `editor.color.tag`, `editor.color.string`, `editor.color.xml_comment`, `editor.color.instr`, `editor.color.default` |
| `org.eclipse.pde.genericeditor.extension` | `org.eclipse.ui.themes` colorDefinitions | `…extension.editor.color.tag`, `.attribute`, `.quote`, `.comment`, `.header`, `.text` |
| `org.eclipse.ant.ui` | preference keys | `org.eclipse.ant.ui.tagsColor`, `.constantStringsColor`, `.commentsColor`, `.processingInstructionsColor`, `.textColor` |
| `org.eclipse.e4.tools.emf.ui` | hardcoded RGB in `OSGI-INF/resources.properties`, being converted to colorDefinitions in eclipse-pde/eclipse.pde#2432 | `org.eclipse.e4.tools.emf.ui.color.tag`, `.string`, `.comment`, `.procInstr` |

merks:
It's interesting to consider that SimRel has 50+ projects and that each of the EPP packages combines the bundles of a significant subset of the those project's contributions.  This necessarily happens without the coordination of the Platform providing reusable general constants.  This suggests that PDE is special merely because it's under the the same top-level project umbrella, i.e., it's special in a way that Mylyn, for example, is not special...

vogella:
Adaption in leaf projects will take long unless they want to constraint themself to latest platform

PR #4372, opened 2026-09-10
Define shared syntax coloring colors in the platform

vogella:
PDE, Ant and the e4 model tooling each define their own colors for syntax coloring of the same XML-ish content (issue #4291). This adds one shared set to `org.eclipse.ui.editors`: theme color definitions for tag, attribute name, attribute value/string, comment and directive, with dark theme values, a "Syntax coloring" category in Colors and Fonts and the public `SyntaxThemeConstants` class.

An editor reusing these no longer needs its own color definitions, its own dark stylesheet or its own entry in the Colors and Fonts tree. The ids are `org.eclipse.editors.syntax.*` rather than tied to the contributing bundle, and the token names avoid XML wording, so `directiveColor` covers an XML declaration, a doctype, a preprocessor statement or a shebang alike and tokens like `keywordColor` can be added later. There is no constant for unstyled text on purpose, the existing editor foreground color covers it.

https://claude.ai/code/session_01Qe7ZCgDF7UAGXk8VvqiPLz

### DIFF
--- a/bundles/org.eclipse.ui.editors/plugin.properties
+++ b/bundles/org.eclipse.ui.editors/plugin.properties
@@ -156,3 +156,11 @@ dummy=
 #--- color definition for code mining annotations
 TEXT_EDITOR_CODE_MINING_COLOR= Text editor code mining color
 TEXT_EDITOR_CODE_MINING_COLOR_DESCRIPTION= The default color used for text editor code mining annotations.
+
+SYNTAX_COLORS_CATEGORY= Syntax coloring
+SYNTAX_COLORS_CATEGORY_DESCRIPTION= Colors shared by editors for syntax coloring, for example the XML, target definition and Ant editors.
+SYNTAX_TAG_COLOR= Tag
+SYNTAX_ATTRIBUTE_NAME_COLOR= Attribute name
+SYNTAX_STRING_COLOR= Attribute value and string
+SYNTAX_COMMENT_COLOR= Comment
+SYNTAX_DIRECTIVE_COLOR= Directive
--- a/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.ui.editors/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.editors; singleton:=true
-Bundle-Version: 3.22.200.qualifier
+Bundle-Version: 3.23.0.qualifier
 Bundle-Activator: org.eclipse.ui.internal.editors.text.EditorsPlugin
 Bundle-ActivationPolicy: lazy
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.ui.editors/plugin.properties
BeckerWdf, line 1: This looks like this:

image


I find the names of the color definitions quite general. Shouldn't we make the color names a bit more specific?

bundles/org.eclipse.ui.editors/plugin.properties
```diff
@@ -156,3 +156,11 @@ dummy=
 #--- color definition for code mining annotations
 TEXT_EDITOR_CODE_MINING_COLOR= Text editor code mining color
 TEXT_EDITOR_CODE_MINING_COLOR_DESCRIPTION= The default color used for text editor code mining annotations.
+
+SYNTAX_COLORS_CATEGORY= Syntax coloring
+SYNTAX_COLORS_CATEGORY_DESCRIPTION= Colors shared by editors for syntax coloring, for example the XML, target definition and Ant editors.
+SYNTAX_TAG_COLOR= Tag
```
BeckerWdf, line 162:
```suggestion
SYNTAX_TAG_COLOR=XML Tag
```

bundles/org.eclipse.ui.editors/plugin.properties
```diff
@@ -156,3 +156,11 @@ dummy=
...
+SYNTAX_ATTRIBUTE_NAME_COLOR= Attribute name
```
BeckerWdf, line 163:
```suggestion
SYNTAX_ATTRIBUTE_NAME_COLOR=XML Attribute name
```

bundles/org.eclipse.ui.editors/plugin.properties
```diff
@@ -156,3 +156,11 @@ dummy=
...
+SYNTAX_STRING_COLOR= Attribute value and string
```
BeckerWdf, line 164:
```suggestion
SYNTAX_STRING_COLOR=XML Attribute value and string
```

bundles/org.eclipse.ui.editors/plugin.properties
```diff
@@ -156,3 +156,11 @@ dummy=
...
+SYNTAX_COMMENT_COLOR= Comment
```
BeckerWdf, line 165:
```suggestion
SYNTAX_COMMENT_COLOR=XML Comment
```

bundles/org.eclipse.ui.editors/plugin.properties
```diff
@@ -156,3 +156,11 @@ dummy=
...
+SYNTAX_DIRECTIVE_COLOR= Directive
```
BeckerWdf, line 166:
```suggestion
SYNTAX_DIRECTIVE_COLOR=XML Directive
```

vogella:
The whole point of this change is to define general colors which others editors can re-use for configuration files not necessary XML files.

PR #4381, opened 2026-09-12
Find/Replace tests: stop spinning the event queue where nothing is due

HeikoKlare:
The find/replace UI tests spent half a second dispatching events after every simulated key stroke, via `FindReplaceTestUtil.runEventQueue()`, which sleeps 10 x 50ms unconditionally. That dominated their runtime: about 35s of a 49s suite run.

Where a key stroke is delivered by notifying the widget, nothing asks for that wait. The listeners run while the event is being sent, key bindings included, since `KeyBindingDispatcher` executes the handler from the display filter, and neither the find/replace dialog nor the overlay schedules work with a delay along the paths these tests exercise. Carrying out what is pending and moving on is enough there. What does arrive late is the focus, handed over by the window manager, and that is waited for on the condition itself rather than for a fixed time; such a wait now carries out everything pending per attempt rather than a single event, so that it also covers opening an editor and its overlay.

Traversing for Enter in the dialog keeps its wait: it can end up activating the shell's default button, which GTK delivers on a timer of its own that flushing the queue does not cover. That is [bug 323272](https://bugs.eclipse.org/bugs/show_bug.cgi?id=323272), the case the wait was originally introduced for.

Measured on Windows with `mvn clean verify -pl :org.eclipse.ui.workbench.texteditor,:org.eclipse.ui.workbench.texteditor.tests`, 231 tests green before and after:

| Class | before | after |
| --- | --- | --- |
| `FindReplaceOverlayTest` | 16.76s | 4.35s |
| `FindReplaceOverlayInEditorTest` | 11.16s | 3.01s |
| suite total | 49.03s | 30.40s |

`FindReplaceDialogTest` is not listed because it is unaffected: its Enter keeps the wait, so it stays where it was.

Draft, because the interesting question is Linux: whether the dialog's wait is still needed there at all. `FindReplaceDialog` has a `TraverseListener` that handles `TRAVERSE_RETURN` by notifying the Find button synchronously and then sets `e.doit = false`, which would suppress the default-button activation and with it the GTK timer. If that listener also runs for a synthetic `traverse()` on a child widget, the wait can go and that class becomes as fast as the overlay ones. Input from someone who knows the GTK traversal path is welcome.


marcushoepfner (approved):
LGTM

PR #4385, opened 2026-09-14
Use Java editor colors for shared string and comment defaults

vogella:
The light defaults of the shared syntax colors from #4372 now use blue (42,0,255) for strings and green (63,127,95) for comments, matching the Java editor. Green strings read like comments in the target definition editor, and comments were indistinguishable from attribute names because both used 128,0,0. The dark theme values are unchanged, and no release ships these constants yet, so no preference migration is needed.


merks:
So when I change the colors for the ant editor do I also indirectly change colors of things used by Java?

image

I guess I'm confused by the impact of how each color is "reused" across different contexts.

vogella:
> So when I change the colors for the ant editor do I also indirectly change colors of things used by Java?

No but I'm using similar color constants as JDT.

PR #4388, opened 2026-09-15
Remove dead tab background colors from the light theme CSS

vogella:
The light theme's base stylesheet defined tab background colors that never took effect, because the GTK, Windows and macOS stylesheets importing it override all of them. The GTK stylesheet also set INACTIVE_NOFOCUS_TAB_BG_*, which is not a color definition the workbench knows. This removes both, so the remaining definitions are the ones actually in use. No visual change is expected.


vogella:
Before

before-light

After

after-light


They are similar (as they should be).

PR #4389, opened 2026-09-15
Name the stylesheet and theme when a theme stylesheet cannot be loaded

akurtakov:
When a bundle contributes a stylesheet to a theme but does not package the file, the error logged was only the bundle entry path, for example "/css/default.css". Neither the message nor the stack trace tells which bundle contributed the stylesheet, so finding the broken bundle meant searching every installed plug-in. Log the stylesheet URI, which carries the contributing bundle, and the theme id instead.

Assisted-by: Anthropic Claude Code (claude-opus-5[1m])


BeckerWdf (approved):
Nice improvement

PR #4394, opened 2026-09-18
Guard against a missing reconciler in computeStyleRanges()

vogella:
`SourceViewer#computeStyleRanges()` dereferenced `fPresentationReconciler` in the partition loop without a null check, so a viewer that was never configured threw a `NullPointerException`.
The existing early return for documents that are not `IDocumentExtension3` did not cover this, because a plain `Document` implements that interface.
The method now returns the empty list, consistent with the other cases in which no styles can be computed.

This removes the need for callers to catch `NullPointerException` around the call, which would also swallow NPEs raised by real presentation repairers.
The new test in `SourceViewerComputeStyleRangesTest` fails with exactly that NPE without the guard; the whole class passes with it.


tobiasmelcher (approved):
thanks a lot Lars

